Issue #59, opened 2018-09-24
Restructure folders somewhat


ralphtheninja:
My issue is this:

1. `electron-builder` needs a build folder for build resources, by default it will pick `build/` and it will not be included when `electron-builder` builds the app
2. this is why I changed configuration to set the build folder to `src/` https://github.com/deltachat/deltachat-desktop/blob/master/package.json#L88, which fixed the problem in 1. but it's not a good solution because then the icons can't be found

### solution 1

1. `buble` builds to another folder than `build/` (and we remove the `src/` config) (and `index.js` needs updating since it currently loads `build/main`)
2. we put icons etc in `build/`

### solution 2

1. we create a new folder, e.g. `resources` and change `buildResources` in configuration to use that
2. put icons etc inside that

I'll play with this a bit. At first glance I think second solution is the best.

Issue #58, opened 2018-09-24
.deb release problems on Ubuntu


sbadux:
Hello,
I've installed the last version on Ubuntu 18.04 and found these problems:

1) The .deb file is really big (183MB). Can you try to reduce it?

2) The icon is not shown correctly on GNOME:
![schermata da 2018-09-24 21-08-42](https://user-images.githubusercontent.com/28478027/45973386-9bd91280-c03e-11e8-9728-3d9dd11de177.png)

3) The app correctly chose the system-language (italina) but the app is still in english. Can you please add the translation? In transifex all the strings are translated.

karissa:
Thanks @sbadux ! Will get to this by the weekend.

hpk42:
On Mon, Sep 24, 2018 at 12:16 -0700, sbadux wrote:
> Hello,
> I've installed the last version on Ubuntu 18.04 and found these problems:
> 
> 1) The .deb file is really big (183MB). Can you try to reduce it?

true, see https://github.com/deltachat/deltachat-desktop/issues/41

> 3) The app correctly chose the system-language (italina) but the app is still in english. Can you please add the translation? In transifex all the strings are translated.

there is no translation integration yet ASFAIK, and it will likely last a while
till they arrive.

karissa:
@hpk42 there is translation integration.

karissa:
point 3 fixed in https://github.com/deltachat/deltachat-desktop/commit/fc4dda6f274b4cfc8fbc357ac70b492cd14a68ff

karissa:
since point 2 and 3 are fixed and point 1 is a duplicate, closing. thanks!

sbadux:
Thanks a lot for the quick fix :)

PR #62, opened 2018-09-24
Fix icons

ralphtheninja:
Solves https://github.com/deltachat/deltachat-desktop/issues/59 and solves point 2 in https://github.com/deltachat/deltachat-desktop/issues/58

screenshot from 2018-09-25 00-29-14
screenshot from 2018-09-25 00-29-44


package.json
```diff
@@ -84,9 +84,6 @@
       ]
     },
     "appId": "com.deltachat.desktop",
-    "directories": {
-      "buildResources": "src"
-    },
     "linux": {
       "target": [
         "AppImage",
```
ralphtheninja: wth is this red coloring?

package.json
```diff
@@ -84,9 +84,6 @@
       ]
     },
     "appId": "com.deltachat.desktop",
-    "directories": {
-      "buildResources": "src"
-    },
```
ralphtheninja: This hack is not needed anymore since we build to `es5/` folder.

package.json
karissa: missing an empty double quote?

package.json
karissa: woot

PR #73, opened 2018-09-25
disable login button

karissa:
half-done: needs to re-enable the login button when login fails.. @ralphtheninja


src/renderer/components/Login.js
```diff
@@ -68,7 +70,7 @@ class Login extends React.Component {
                 onChange={this.handleChange}
```
ralphtheninja: Can input field still accept enter?

Issue #82, opened 2018-09-27
contact request do not work


r10s:
contact requests are displayed in the chatlist and clicking on it shows the question "Chat with NAME - Yes - No - Never", however, clicking on yes, no chat with the contact is created.

creating the chat manually works then.

also, the question should include the email-address and not only the name, i would suggest to use [dc_get_contact_name_n_addr()](https://c.delta.chat/classdc__contact__t.html#a16056431c6926d327f5728ad3ddd4fdf) here which takes care eg. not showing the address twice if the name is unset.

karissa:
this shoudl be fixed now

PR #87, opened 2018-09-27
fix dead drops and use contact name and addr

karissa:
so now contact requests work and create a new chat, but for some reason the old dead drop is still there -- do we need to 'mark read' the chat? @r10s


karissa:
fixes #82

ralphtheninja:
@karissa Lets merge? Or do you still need feedback from @r10s?

PR #92, opened 2018-09-28
Tweak login page + fix this._render is not a function

ralphtheninja:
* Moves login button to left of cancel
* Fixes `this._render` is not a function (only set once init has completed)
* Reset email and password to empty strings rather than null


karissa:
cool!

Issue #78, opened 2018-09-26
listen for DC_EVENT_MSG_DELIVERED and _FAILED and update chats


karissa:
this will be done in src/main/deltachat.js

for good performance, we might want to have an easy way to lookup just the message that changed and update it's state.

PR #96, opened 2018-09-29
Update message status

ralphtheninja:
* Listen to `DC_EVENT_MSG_DELIVERED`, `DC_EVENT_MSG_FAILED` and
`DC_EVENT_MSG_READ`
* Be lazy in `ChatMessage` and don't cache `dc.getMessage()`, instead get message
object at time of calling `toJson()`

Closes https://github.com/deltachat/deltachat-desktop/issues/78

Demo: https://cloud.weho.st/index.php/s/sMsBCEGcfCHbjdJ (link valid for 30 days from post date)

### DIFF
--- a/src/main/deltachat.js
+++ b/src/main/deltachat.js
@@ -11,10 +11,9 @@ class ChatMessage {
   }
 
   toJson () {
-    var msg = this._dc.getMessage(this._messageId)
-    var fromId = msg && msg.getFromId()
-    var contact = fromId && this._dc.getContact(fromId)
-
+    const msg = this._dc.getMessage(this._messageId)
+    const fromId = msg && msg.getFromId()
+    const contact = fromId && this._dc.getContact(fromId)
     return {
       fromId,
       id: this._messageId,
--- a/src/main/deltachat.js
+++ b/src/main/deltachat.js
@@ -6,24 +6,22 @@ const config = require('../config')
 
 class ChatMessage {
   constructor (messageId, dc) {
-    var msg = dc.getMessage(messageId)
-    this.fromId = msg && msg.getFromId()
-    if (!this.fromId) return
-
-    this.msg = msg
-    this.messageId = messageId
-    this.isMe = this.fromId === 1
-    this.contact = dc.getContact(this.fromId)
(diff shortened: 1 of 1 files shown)


ralphtheninja:
Unfortunately, with the current design, this is how it has to be, since doing `dc.getMessage(id)` will cache the message state at time of creation, which means `dc.getMessage()` has to be called again.

Maybe later we should try and rethink the design a bit and render only visible messages etc. But it requires a lot of thought which right now seems out of scope for this PR. But we should keep this in mind.

src/main/deltachat.js
```diff
@@ -6,24 +6,22 @@ const config = require('../config')
 
 class ChatMessage {
   constructor (messageId, dc) {
-    var msg = dc.getMessage(messageId)
-    this.fromId = msg && msg.getFromId()
-    if (!this.fromId) return
```
ralphtheninja: I removed this one and made something similar in `toJson()`. I think the previous commit would have crashed on `this.contact.toJson()` since `this.contact` would have been `undefined` in case `fromId` was falsey (since we return in the constructor and never set `this.contact`).

src/main/deltachat.js
```diff
@@ -6,24 +6,22 @@ const config = require('../config')
...
-    this.msg = msg
-    this.messageId = messageId
-    this.isMe = this.fromId === 1
-    this.contact = dc.getContact(this.fromId)
+    this._messageId = messageId
+    this._dc = dc
   }
 
   toJson () {
+    var msg = this._dc.getMessage(this._messageId)
```
ralphtheninja: Since we call `dc.getMessage()` in `toJson()` we will always get the correct state when `render()` is called.

src/main/deltachat.js
```diff
@@ -6,24 +6,22 @@ const config = require('../config')
...
+    var fromId = msg && msg.getFromId()
+    var contact = fromId && this._dc.getContact(fromId)
```
ralphtheninja: Actually, these should be `const`.

ralphtheninja:
Gah, accidentally merged this.

Issue #99, opened 2018-09-29
Cache less and clean up in deltachat controller


ralphtheninja:
I think we should move closer to `deltachat-node`/`deltachat-core` and not cache as much as we do. My main concerns is that the delta chat controller code is yet another layer, storing the same message and chat data, which is already in core. This is partly due to the design of the poc cli tool. The code will be a lot simpler since we don't have to manage state as much, e.g. adding chat message to array, removing from array etc. This clean up will also make it easier to later write new code for being more lazy, e.g. rendering only seen or ranges of messages etc, as opposed to rendering everything all the time.

I'll pin some points down here as we go. These needs to be done in order to avoid big bang effect.

- [ ] Remove `ChatMessage` class. It's basically just a function around a `messageId` which converts to json for the browser
- [ ] Don't cache messages in `ChatPage#_messages`. Get the messages from node/core on the fly in `toJson()`. This means `.messages()`, `.append()`, `.clear()`, `.appendMessage()` and `.deleteMessage()` are no longer needed and `ChatPage` class becomes a function around `chatId` (just like for `ChatMessage`)
- [ ] Remove `ChatPage#messages()`
- [ ] Remove `ChatPage#append()`
- [ ] Remove `ChatPage#clear()`
- [ ] Remove `ChatPage#appendMessage()`
- [ ] Remove `ChatPage#deleteMessage()`

hpk42:
just wanted to chime in that in the python bindings i also 
removed caching from the higher level layer for now.  Already 
there are some caching-like issues on the c-layer itself, i.e. when you 
get a "dc_msg_t" object and ask about a message's state etc because
that struct only reflects the state at the time you got it. 
So i think it's a good idea to remove caching and only add it
later when it's really needed performance-wise.

PR #119, opened 2018-10-04
Refactor state

ralphtheninja:
Closes https://github.com/deltachat/deltachat-desktop/issues/99

This removes code caching messages and chats and renders everything on the fly (which we did before anyway).

### DIFF
--- a/src/main/deltachat.js
+++ b/src/main/deltachat.js
@@ -30,10 +30,6 @@ class ChatPage {
     return this._messages
   }
 
-  append (line) {
-    this._messages.push(line)
-  }
-
   clear () {
     this._messages = []
   }
@@ -49,8 +45,6 @@ class ChatPage {
     }
     return chat
   }
-
-  appendMessage (messageId) {}
 
   deleteMessage (messageId) {
     const index = this._messages.findIndex(m => {
@@ -124,18 +118,11 @@ class DeltaChatController {
 
     dc.on('DC_EVENT_MSGS_CHANGED', (chatId, msgId) => {
       log('event msgs changed', chatId, msgId)
-      const msg = dc.getMessage(msgId)
-      if (!msg) return
-
-      if (msg.getState().isPending() || msg.getState().isDelivered()) {
-        this.appendMessage(chatId, msgId)
-      }
-      render()
+      if (dc.getMessage(msgId)) render()
     })
 
     dc.on('DC_EVENT_INCOMING_MSG', (chatId, msgId) => {
       log('incoming message', chatId, msgId)
-      this.appendMessage(chatId, msgId)
       render()
     })
 
@@ -196,11 +183,7 @@ class DeltaChatController {
     chat.clear()
   }
 
-  loadMessages (chatId) {
-    const chat = this._loadChatPage(chatId)
-    const messageIds = this._dc.getChatMessages(chatId, 0, 0)
-    messageIds.forEach(chat.appendMessage.bind(chat))
-  }
+  loadMessages (chatId) {}
 
   sendMessage (...args) {
     return this._dc.sendTextMessage(...args)
@@ -241,10 +224,6 @@ class DeltaChatController {
     this._dc.blockContact(contactId, true)
     const name = contact.getNameAndAddress()
     this.warning(`Blocked contact ${name} (id = ${contactId})`)
-  }
-
-  appendMessage (chatId, messageId) {
-    this._loadChatPage(chatId).appendMessage(messageId)
   }
 
   _loadChatPage (chatId, opts) {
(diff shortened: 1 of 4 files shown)


ralphtheninja:
It's a rather large diff. Easiest way to review is to take one commit at a time, before looking at the complete diff. Each commit corresponds to one step in the refactoring process.

src/main/deltachat.js
```diff
@@ -4,24 +4,17 @@ const path = require('path')
...
+  const msg = dc.getMessage(messageId)
+  const fromId = msg && msg.getFromId()
+  const contact = fromId && dc.getContact(fromId)
+  return {
+    fromId,
+    id: messageId,
+    isMe: fromId === 1,
+    contact: contact ? contact.toJson() : {},
+    msg: msg.toJson(),
+    filemime: msg.getFilemime()
```
ralphtheninja: This commit replaces `ChatMessage` class with a function.

src/main/deltachat.js
```diff
@@ -26,10 +26,6 @@ class ChatPage {
     this.chat = this._dc.getChat(this.chatId)
   }
 
-  messages () {
```
ralphtheninja: Not used.

src/main/deltachat.js
```diff
@@ -274,10 +274,6 @@ class DeltaChatController {
     log.error(line)
   }
 
-  _getChats () {
```
ralphtheninja: Not used.

src/main/deltachat.js
```diff
@@ -210,7 +195,7 @@ class DeltaChatController {
       const msg = this._dc.getMessage(messageIds[0])
       page.fromId = msg.getFromId()
     }
-    page.summary = summary || this.getSummaryByChatId(chatId)
+    page.summary = summary
```
ralphtheninja: Since `summary` is always set to something, the other branch is dead code.

src/main/deltachat.js
```diff
@@ -18,24 +18,22 @@ function messageIdToJson (messageId, dc) {
...
+    const fromId = msg.getFromId()
+    if (fromId) {
+      const contact = dc.getContact(fromId)
+      if (contact) {
+        chat.contact = contact.toJson()
+      }
     }
-    return chat
   }
+  return chat
```
ralphtheninja: This replaces `ChatPage` class with a function.

src/renderer/components/ChatView.js
```diff
@@ -44,14 +44,10 @@ class ChatView extends React.Component {
...
 
   componentWillUnmount () {
-    var chatId = this.props.screenProps.chatId
-    ipcRenderer.send('dispatch', 'clearChatPage', chatId)
     this.observer.disconnect()
   }
 
   componentDidMount () {
-    var chatId = this.props.screenProps.chatId
-    ipcRenderer.send('dispatch', 'loadChats', chatId)
```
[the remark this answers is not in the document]
karissa: (maybe that got lost though because there was a bug, but i wonder if this refactor will make it easier to reintroduce such a performance increase!)

hpk42:
On Fri, Oct 05, 2018 at 14:24 +0000, Lars-Magnus Skog wrote:
> I'll try to break this down into smaller chunks instead.

Why?  Seemed to be you "just" needed to address Karissa's review comments ...
but maybe i missed something. 

i know it's considered good practise to have smaller chunks but 
it can also be a lot of effort to do that with a lot of review overhead etc.
which is not always warranted if it's 2-3 active developers.

karissa:
yeah I don't think the PR is too big @ralphtheninja , and it's a good start. I can take it from here too, if that feels better

ralphtheninja:
> yeah I don't think the PR is too big @ralphtheninja , and it's a good start. I can take it from here too, if that feels better

If you say so. Re-opening it and rebasing on latest master.

> Why? Seemed to be you "just" needed to address Karissa's review comments ... but maybe i missed something.

I couldn't address it since this PR complete changes the current design, since we're no longer caching messages (array of `ChatMessage` class) in chats (`ChatPage` class)

ralphtheninja:
@karissa I think further on we should try to re-think the architecture a bit. With that I don't mean changing model <-> ipc <-> view, but rather to split up the `'render'` message into more specific messages, so extending the ipc to get more fine grained control. My initial idea was that e.g. a `ChatView` could have a separate message for getting the messages for a particular chat, this way we don't have to transfer all the state. If that makes sense.

karissa:
@ralphtheninja yeah I could see this being useful for the specific case of getting messages for the `ChatView`. Let's start there, mostly for performance reasons. I think though overall the `render` method is nice for the state transfer as long as it's not huge (which it shouldn't be, as long as 'messages' are passed separately). React checks properties and only re-renders views that have properties that have changed. So we could reduce unnecessary re-rendering in the future if we want by only passing the state objects that are necessary, rather than the whole `deltachat` state object

ralphtheninja:
> Let's start there, mostly for performance reasons. I think though overall the `render` method is nice for the state transfer as long as it's not huge (which it shouldn't be, as long as 'messages' are passed separately).

:+1:

PR #163, opened 2018-10-15
(WIP) Implemented splitted ChatList/View

Jikstra:
Just a quick change/hack to get into desktop/react development. Some questions i have:
- where do we want to style the different components? Only in the conversations repo?
- If i have to change something in the conversations repo, i always have to make 2 pull requests and to review them, both have to get pulled which is annoying. What's the reason behind the conversations repo? To have it clearly seperated?
- Currently i have the height/width hardcoded but it should be the screen height/width. How can i access the screen height/width? 

I didn't do much with react, so in case i'm doing something wrong/in a bad way, please please tell me :)

### DIFF
--- a/src/renderer/ScreenController.js
+++ b/src/renderer/ScreenController.js
@@ -2,8 +2,8 @@ const React = require('react')
 const { ipcRenderer } = require('electron')
 
 const Login = require('./components/Login')
-const ChatList = require('./components/ChatList')
-const ChatView = require('./components/ChatView')
+const SplittedChatListAndView = require('./components/SplittedChatListAndView')
+//
 const CreateChat = require('./components/CreateChat')
 const CreateGroup = require('./components/CreateGroup')
 const CreateContact = require('./components/CreateContact')
@@ -59,11 +59,8 @@ class Home extends React.Component {
       case 'CreateGroup':
         Screen = CreateGroup
         break
-      case 'ChatView':
-        Screen = ChatView
-        break
       default:
-        Screen = ChatList
+        Screen = SplittedChatListAndView
         break
     }
 
--- a/src/renderer/ScreenController.js
+++ b/src/renderer/ScreenController.js
@@ -21,7 +21,7 @@ class Home extends React.Component {
   }
 
   changeScreen (screen, screenProps) {
-    if (!screen) screen = ChatList
+    if (!screen) screen = SplittedChatListAndView
     if (!screenProps) screenProps = {}
     this.setState({ screen, screenProps })
     ipcRenderer.send('render')
--- a/static/main.css
+++ b/static/main.css
@@ -57,9 +57,11 @@ body {
 .user-feedback.error {
   background-color: red;
 }
+
 .user-feedback.success {
   background-color: green;
 }
+
 img {
   max-width: 500px;
 }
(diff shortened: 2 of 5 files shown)


static/main.css
```diff
@@ -71,3 +73,15 @@ ul {
...
+
+.ChatList {
+  width: 30%;
+  float: left;
+}
+
+.ChatView {
+  width: 70%;
+  float: right;
+  height: 800px;
```
ralphtheninja: Maybe `height: 100%'?

Jikstra: Height 100% makes it just as big as it needs to be with a scroll bar but not fitting the window height. There's some property i can get from electron, i will give it a try

ralphtheninja:
> where do we want to style the different components? Only in the conversations repo?

We should be able to style in `static/main.css`?

>  If i have to change something in the conversations repo, i always have to make 2 pull requests and to review them, both have to get pulled which is annoying. What's the reason behind the conversations repo? To have it clearly seperated?

Yes, this is painful. Main reason is that `conversations` needs a completely different build step, since it's based on `TypeScript`. I suggest you hack on your fork of it and pick a branch from that in `package.json` for now. Then we don't have to handle both PRs at the same time to get something changed.

Jikstra:
> Yes, this is painful. Main reason is that conversations needs a completely different build step, since it's based on TypeScript.

Don't we want to use TS in deltachat-desktop too? I think it would be better if we would use consistent written code. I don't think it makes a lot of sense to maintain an extra package of components which maybe no one else will ever use besides us. Still we should seperate logic and view components completely.

src/renderer/ScreenController.js
```diff
@@ -59,11 +59,8 @@ class Home extends React.Component {
       case 'CreateGroup':
         Screen = CreateGroup
         break
-      case 'ChatView':
-        Screen = ChatView
-        break
       default:
-        Screen = ChatList
+        Screen = SplittedChatListAndView
```
ralphtheninja: Maybe just keep the name as `ChatView` and throw everything else away?

ralphtheninja:
> Don't we want to use TS in deltachat-desktop too?

I don't see the reason why. Mainly because I'm old and haven't used TypeScript :smile:

Jikstra:
> > Don't we want to use TS in deltachat-desktop too?
> 
> I don't see the reason why. Mainly because I'm old and haven't used TypeScript smile Otoh, if others want to use it, please go ahead and don't let me stop you.

Me neither. Don't see any benefits of using ts, also i don't really like the extra time which is needed for the the ts -> js transpile time but i never really used it and don't know it's benefits. Maybe @karissa can give us some insights.

static/main.css
ralphtheninja: > I don't really want to continue "monkey patching/overwriting" css stuff outside of conversations.

Yet another reason to move conversations into desktop :)

src/renderer/ScreenController.js
Jikstra: I don't have any hard opinions on naming, but maybe it makes more sense to use new new names for the components. I will think about it, maybe i have an idea.

hpk42:
closing this in favor of #178

PR #167, opened 2018-10-16
Merge in dependencies from conversation

ralphtheninja:


hpk42:
why does it make sense to remove the test code?

ralphtheninja:
> why does it make sense to remove the test code?

I just thought we wouldn't need it. Can revert.

hpk42:
I don't know if feasible but if the now-inline conversations evolves it'd be good to run/evolve the tests as well ...

Jikstra:
> why does it make sense to remove the test code?

Can keep them for now, when we looked at it for a glimpse, they didn't look very helpful BUT it looks like they aren't unhelpful on a second look. Should keep them.

@ralphtheninja LGTM

Issue #26, opened 2018-09-11
Group Management (Edit Group screen)


karissa:
When in the chat view for a group, you should be able to click the menu in the upper right and hit 'Edit Group' to go to the Edit Group screen.

To keep it simple, screen could be the same as the one that is shown when the group is created -- `createGroup.js`. Upon save, the changes in the members and name are then reflected in the group, and the screen returns to the chat window.

ralphtheninja:
You should be able to:

1. change group name
  * input field can be populated with current group name
  * create group button could be labeled with "Save" instead
1. add more members to the group
  * show list of users that aren't yet in the group
  * clicking on users will mark them green just like as in the Create Group use case
  * selected users are added to the group when "Save" is clicked

Apart from this, we should list current group members (which shouldn't be editable)

ralphtheninja:
I think instead of tweaking `CreateGroup.js` (which will handle both create and edit) we can copy `CreateGroup.js` to `EditGroup.js` and tweak that. Once that's done we can try refactor common code into a base class component.

PR #172, opened 2018-10-17
Edit group

ralphtheninja:
Closes https://github.com/deltachat/deltachat-desktop/issues/26

I started by copying functionality from `CreateGroup` to `EditGroup` and then refactored the common code into a `GroupBase` component.

The edit screen looks the same, but have different heading and different label on the button.

In edit mode, already existing group members cannot be de-selected, only contacts that haven't yet been added to the group can be toggled. When "Save" button is clicked, the group name is applied and the newly selected contacts are added to the group.

Also, I removed the error handling completely, disabling the button if no group name is set, or if no contacts have been selected. So the user can either add/modify name, and select contacts to enable the button or bail out with the arrow in the upper left corner.

### DIFF
--- a/_locales/en.json
+++ b/_locales/en.json
@@ -21,7 +21,23 @@
     },
     "createGroup": {
         "message": "Create Group",
-        "description": "Create group button label"
+        "description": "Create group menu item label"
+    },
+    "archiveGroup": {
+        "message": "Archive Group",
+        "description": "Archive group menu item label"
+    },
+    "deleteGroup": {
+        "message": "Delete Group",
+        "description": "Delete group menu item label"
+    },
+    "editGroup": {
+        "message": "Edit Group",
+        "description": "Edit group menu item label"
+    },
+    "save": {
+        "message": "Save",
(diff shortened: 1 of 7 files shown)


ralphtheninja:
This concludes the Q3 planning, except for the windows parts, which we should postpone since it's based on many other things working, e.g. core and node bindings working on windows.

ralphtheninja:
Screen from chat view:

screenshot from 2018-10-17 22-47-40

Edit screen:

screenshot from 2018-10-17 22-48-24

ralphtheninja:
Maybe `Apply` is better than `Save`?

r10s:
great :)

fun-fact: [just wrote an issue](https://github.com/deltachat/deltachat-android-ii/issues/92) to add this in android-dev :)

ralphtheninja:
@r10s So you should be able to remove contacts from a group? I had a look at the android app to spec this out.

ralphtheninja:
And group image as well.

Issue #36, opened 2018-09-12
Consolidate the Chat List screen and Chat View screen


karissa:
Chat List screen is a preview list of the chats, and Chat View is a view of a single chat. Right now, these are two separate screens. Instead, we could have (like Signal Desktop, or Facebook Messenger) have one screen with the Chat List on the left hand side and the current chat you are viewing in the rest of the window. Example:

![image](https://user-images.githubusercontent.com/633012/45445834-2ce3dd00-b680-11e8-8797-b588c68752c3.png)

r10s:
yes, i think this makes sense.

PR #178, opened 2018-10-18
Consolidate ChatList and ChatView

Jikstra:
Fixes https://github.com/deltachat/deltachat-desktop/issues/36

Some things i want to mention:
- for the `#root` element there was an padding-top: 50px which caused trouble. I switched to using calc + 100vh to have the complete window height instead of doing some weird css stuff. Please consider switching to vh for your components too. It's possible that this "breaks" other things.
- Adding css to conversations isn't really possible because we define the bigger components outside of conversations. That means we have no sass support & can't really do nice css preprocessing. 
- There's weird css code in conversations. For example the chatlist is a weirdly hacked flexbox (which isn't the right thing for this, a normal ul would be enough for this as we don't need wrapping or anything like this) which makes trouble on resizing. We maybe should replace this.

### DIFF
--- a/src/renderer/components/ChatView.js
+++ b/src/renderer/components/ChatView.js
@@ -42,13 +42,13 @@ class ChatView extends React.Component {
   }
 
   onArchiveChat () {
-    const chatId = this.props.screenProps.chatId
+    const chatId = this.props.chat.id
     ipcRenderer.send('dispatch', 'archiveChat', chatId)
     this.props.changeScreen()
   }
 
   onDeleteChat () {
-    const chatId = this.props.screenProps.chatId
+    const chatId = this.props.chat.id
     ipcRenderer.send('dispatch', 'deleteChat', chatId)
     this.props.changeScreen()
   }
@@ -59,7 +59,7 @@ class ChatView extends React.Component {
   }
 
   writeMessage (text) {
-    const chatId = this.props.screenProps.chatId
+    const chatId = this.props.chat.id
     ipcRenderer.send('dispatch', 'sendMessage', chatId, text)
   }
 
@@ -79,7 +79,7 @@ class ChatView extends React.Component {
 
   getChat () {
     const { deltachat } = this.props
-    const { chatId } = this.props.screenProps
+    const { chatId } = this.props.chat
     const index = deltachat.chats.findIndex(chat => {
       return chat.id === chatId
     })
@@ -87,7 +87,8 @@ class ChatView extends React.Component {
   }
 
   scrollToBottom (force) {
-    var doc = document.querySelector('html')
+    console.log('scroll')
+    var doc = document.querySelector('.ChatView')
     doc.scrollTop = doc.scrollHeight
   }
 
@@ -118,8 +119,8 @@ class ChatView extends React.Component {
 
   render () {
     const { attachmentMessage, setupMessage } = this.state
-    const chat = this.getChat()
-    if (!chat) return (<div />)
+    const chat = this.props.chat
+    if (!chat) return (<div className="ChatView"><p>No chat selected</p></div>)
 
     this.state.value = chat.textDraft
 
@@ -134,21 +135,7 @@ class ChatView extends React.Component {
     </Menu>)
 
     return (
-      <div>
-        <Navbar fixedToTop>
-          <NavbarGroup align={Alignment.LEFT}>
-            <Button className={Classes.MINIMAL} icon='undo' onClick={this.props.changeScreen} />
-            <img src={chat.profileImage} />
-            <NavbarHeading>{chat.name}</NavbarHeading>
-            <div>{chat.subtitle}</div>
-          </NavbarGroup>
-          <NavbarGroup align={Alignment.RIGHT}>
-            <Popover content={menu} position={Position.RIGHT_TOP}>
-              <Button className={Classes.MINIMAL} icon='menu' />
-            </Popover>
-          </NavbarGroup>
-        </Navbar>
-
+      <div className="ChatView">
         <SetupMessageDialog
           userFeedback={this.props.userFeedback}
           setupMessage={setupMessage}
@@ -173,7 +160,9 @@ class ChatView extends React.Component {
             })}
           </ConversationContext>
         </div>
-        <Composer onSubmit={this.writeMessage.bind(this)} />
+        <div className="InputMessage">
+          <Composer onSubmit={this.writeMessage.bind(this)} />
+        </div>
       </div>
     )
   }
--- a/src/renderer/components/ChatView.js
+++ b/src/renderer/components/ChatView.js
@@ -56,7 +56,7 @@ class ChatView extends React.Component {
 
   scrollToBottom (force) {
     var doc = document.querySelector('.ChatView')
-    if(doc) doc.scrollTop = doc.scrollHeight
+    if (doc) doc.scrollTop = doc.scrollHeight
   }
 
   onClickAttachment (attachmentMessage) {
--- a/src/renderer/components/ChatView.js
+++ b/src/renderer/components/ChatView.js
@@ -8,19 +8,6 @@ const { Overlay } = require('@blueprintjs/core')
 
 let MutationObserver = window.MutationObserver
 
-const {
-  Alignment,
-  Classes,
-  Navbar,
-  Position,
-  Menu,
-  MenuItem,
-  Popover,
-  NavbarGroup,
-  NavbarHeading,
-  Button
-} = require('@blueprintjs/core')
-
 const { ConversationContext, Message } = require('./conversations')
 
 var theme = 'light-theme' // user prefs?
@@ -37,8 +24,6 @@ class ChatView extends React.Component {
     this.scrollToBottom = this.scrollToBottom.bind(this)
     this.conversationDiv = React.createRef()
   }
-
-
 
   writeMessage (text) {
     const { chatId } = this.props
@@ -92,17 +77,15 @@ class ChatView extends React.Component {
     this.setState({ setupMessage: false })
   }
 
-
   render () {
     const { attachmentMessage, setupMessage } = this.state
     const chat = this.getChat()
-    if (!chat) return (<div className="ChatView"><p>No chat selected</p></div>)
+    if (!chat) return (<div className='ChatView'><p>No chat selected</p></div>)
 
     this.state.value = chat.textDraft
 
-
     return (
-      <div className="ChatView">
+      <div className='ChatView'>
         <SetupMessageDialog
           userFeedback={this.props.userFeedback}
           setupMessage={setupMessage}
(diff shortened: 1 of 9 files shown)


hpk42:
begings to work nicely!
two things: 
- missing refresh: the chatview does not show messages after i sent one, if i click another contact then return to the chatview, my typed message is shown as sent
- the lefthand chatlist view should have a larger minimum width -- maybe 60 chars or so -- and probably doesn't need to be much larger right now because nothing more is shown if you enlargen it

ralphtheninja:
Cool! A few things:

* The hamburger menu no longer contains alternatives for archive, delete chat/group and edit group
* There should probably be a border between the chatlist to the left and the chatview to the right

Also some indentations in css, console.log and standard issues, but maybe we can address that at the end.

src/renderer/components/ChatView.js
```diff
@@ -134,21 +135,7 @@ class ChatView extends React.Component {
...
-      <div>
-        <Navbar fixedToTop>
-          <NavbarGroup align={Alignment.LEFT}>
-            <Button className={Classes.MINIMAL} icon='undo' onClick={this.props.changeScreen} />
-            <img src={chat.profileImage} />
-            <NavbarHeading>{chat.name}</NavbarHeading>
-            <div>{chat.subtitle}</div>
-          </NavbarGroup>
-          <NavbarGroup align={Alignment.RIGHT}>
-            <Popover content={menu} position={Position.RIGHT_TOP}>
```
ralphtheninja: This menu is now gone. Maybe move it to the split component view?

ralphtheninja:
Did a quick write-up of the current render workflow:

https://gist.github.com/ralphtheninja/09311a546cc73c736d5abed5c8e31060

src/renderer/components/ChatView.js
```diff
@@ -118,8 +119,8 @@ class ChatView extends React.Component {
 
   render () {
     const { attachmentMessage, setupMessage } = this.state
-    const chat = this.getChat()
-    if (!chat) return (<div />)
+    const chat = this.props.chat
```
ralphtheninja: I think this is where we broke updates. It's using a cached value.

ralphtheninja: `chat` should be the chat object from the `deltachat` json and it has to be retrieved each time _or_ the calling component needs to update this prop.

Jikstra:
@ralphtheninja aaaah now i get it. We only passed down the cached chat even if we rerendered? 

But i'm not really happy with only passing down the chatid because we need the chatobject in the splitted component (for displaying the group info like title/image...) and in the chatview obviously. Getting it twice isn't needed and will most likely do weird things if the states differ.

ralphtheninja:
> @ralphtheninja aaaah now i get it. We only passed down the cached chat even if we rerendered?

Correct.

>  Getting it twice isn't needed and will most likely do weird things if the states differ.

Sure. It's not needed during the same render operation obviously. But now you understand the problem and can tweak it better than I did :)

ralphtheninja:
I guess the split-component should get the chat and pass it down to `ChatView`.

ralphtheninja:
> Getting it twice isn't needed and will most likely do weird things if the states differ.

Just a side note. The state shouldn't differ, since we're rendering because of an update and the `deltachat` prop doesn't change since it's single threaded.

hpk42:
Using current branch the chatlistview does not highlight the selected group, the group chatview shows no "membrers"  anymore (let alone edit) and the messages show no author: iow, it's hard to know at all in which group i am ...

src/renderer/components/SplittedChatListAndView.js
```diff
@@ -20,20 +20,20 @@ class SplittedChatListAndView extends React.Component {
...
     super(props)
     this.state = {
-      chatId: null
+      selectedChat: {}
     }
 
     this.onChatClick = this.onChatClick.bind(this)
   }
 
   onChatClick (chat) {
```
ralphtheninja: I think the main problem starts here. If `onChatClick()` called back with a `chatId` instead all caching problems would be gone :)

ralphtheninja: Nevermind my comments for now. I'll PR it later once this is merged.

ralphtheninja:
> Shouldn't be broken if we call getChat() on each render.

That's the thing, we aren't calling `getChat()` from within the splitted component (we're calling it in `ChatView`), which is using cached data. I think `getChat()` method should be moved to the split component and instead pass in chat object to `ChatView`. But I can make a PR later instead, maybe it gets clearer then.

ralphtheninja:
> Don't really know what's better, especially if we need the same stuff in different components.

I don't know either. Main thing right now is to get rid of cached data, which means getting data from `this.props.deltachat.chats[whatever]`. Later on we can figure out "what's best". :smile:

Jikstra:
>      That's the thing, we aren't calling getChat() from within the splitted component (we're calling it in ChatView), which is using cached data. I think getChat() method should be moved to the split component and instead pass in chat object to ChatView. But I can make a PR later instead, maybe it gets clearer then.

But ChatView is also getting rerendered if deltachat changes as we pass it down to ChatView too. At least that's how I understand it. But check out my last commit, this should fix it, no more caching :D

src/renderer/components/SplittedChatListAndView.js
```diff
@@ -20,28 +20,41 @@ class SplittedChatListAndView extends React.Component {
...
+    this.setState({ selectedChatId: chatId })
   }
 
   render () {
     const { deltachat } = this.props
-    const { selectedChat } = this.state
+    const { selectedChatId } = this.state
     const tx = window.translate
 
+    const selectedChat = this.getSelectedChat()
```
ralphtheninja: This isn't used.

src/renderer/components/SplittedChatListAndView.js
```diff
@@ -20,28 +20,41 @@ class SplittedChatListAndView extends React.Component {
...
     const { deltachat } = this.props
-    return deltachat.chats[0]
+    if(deltachat.chats.length == 0) return null
+    return deltachat.chats[0].id
   }
 
-  onChatClick (chat) {
-    console.log('xx', chat)
-    this.setState({ selectedChat: chat })
+  getSelectedChat() {
```
ralphtheninja: This is essentially the same method as `getChats()` in `ChatView`. Why not just move that method here?

ralphtheninja: Essentially I want to revert the commit I made and pass a chat object into `ChatView` instead of id, so `ChatView` doesn't need to do the lookup. The split component takes care of that.

Jikstra: We need it for both. It would be better to have a generic deltachat.getChat(id) method?

Jikstra:
@ralphtheninja i think most stuff is implemented and *could* work ^^ I hope everything is fixed now, but problems may occur. Maybe it's easier to have further conversations n irc?

ralphtheninja:
> @ralphtheninja i think most stuff is implemented and _could_ work ^^ I hope everything is fixed now, but problems may occur. Maybe it's easier to have further conversations on irc?

I'll spend some time testing it.



> Oh and I didn't focus on test/standard yet... Just wanted to get things working/looking better so sorry for pushing so many commits over the last couple of hours. I was just in the flow ^^

No worries what so ever. One thing at a time!

ralphtheninja:
Top most chat is selected in `ChatList` but it's hidden

screenshot from 2018-10-19 15-32-11

ralphtheninja:
Maybe also put the "Key transfer" at the bottom of the hamburger menu?

screenshot from 2018-10-19 15-33-37

Jikstra:
> Top most chat is selected in ChatList but it's hidden

Thanks, fixed.

> Maybe also put the "Key transfer" at the bottom of the hamburger menu?

Will do this for now, the plan for the long term is to have a menu on the left side for creating chats... and one on the right to modify the currently selected chat. Maybe even open a dialog when clicking on the group or a gear icon on the right side.

> when deleting a chat, maybe select the first chat again? now it says "no chat selected"

Same with creating a new chat. Would be some ipc the right thing for this? Otherwise i need to pass the onChatClick method around different screens or something like this which isn't really nice in my opinion. How can we implement this?

static/main.css
```diff
@@ -12,15 +12,18 @@ body {
...
   flex-grow: 1;
   overflow: auto;
   padding: 10px 10px;
 }
 
 
+.window, .below-navbar {
+  height: calc(100vh - 50px);
+  margin-top: 50px;
+}
```
ralphtheninja: :+1:

src/renderer/components/SplittedChatListAndView.js
```diff
@@ -145,10 +148,10 @@ class SplittedChatListAndView extends React.Component {
       <MenuItem icon='plus' text={tx('addContact')} onClick={this.onCreateContact} />
       <MenuItem icon='plus' text={tx('addChat')} onClick={this.onCreateChat} />
       <MenuItem icon='plus' text={tx('createGroup')} onClick={this.onCreateGroup} />
-      <MenuItem icon='exchange' text={tx('initiateKeyTransferTitle')} onClick={this.initiateKeyTransfer} />
-      <MenuItem icon='compressed' text={archiveMsg} onClick={this.onArchiveChat} />
-      <MenuItem icon='delete' text={deleteMsg} onClick={this.onDeleteChat} />
+      {selectedChat ? <MenuItem icon='compressed' text={archiveMsg} onClick={this.onArchiveChat} /> : null}
+      {selectedChat ? <MenuItem icon='delete' text={deleteMsg} onClick={this.onDeleteChat} /> : null}
       {isGroup ? <MenuItem icon='edit' text={tx('editGroup')} onClick={this.onEditGroup.bind(this, selectedChat)} /> : null}
```
ralphtheninja: `selectedChat && isGroup`?

ralphtheninja:
> Will do this for now, the plan for the long term is to have a menu on the left side for creating chats... and one on the right to modify the currently selected chat. Maybe even open a dialog when clicking on the group or a gear icon on the right side.

Sounds good. Also good that someone is making plans :wink: 

Lets get this merged now!

Jikstra:
>  Sounds good. Also good that someone is making plans wink

:D

> Lets get this merged now!

Finally!

ralphtheninja:
screenshot from 2018-10-19 16-32-20

src/renderer/components/SplittedChatListAndView.js
```diff
@@ -130,15 +130,18 @@ class SplittedChatListAndView extends React.Component {
     ].includes(selectedChat && selectedChat.type)
   }
 
+  componentWillReceiveProps() {
+    if(!this.getSelectedChat()) {
```
ralphtheninja: standard ;)

src/renderer/components/SplittedChatListAndView.js
```diff
@@ -130,15 +130,18 @@ class SplittedChatListAndView extends React.Component {
...
+      this.setState({ selectedChatId: this.getInitiallySelectedChatId()})
```
ralphtheninja: And here as well I guess

Issue #183, opened 2018-10-19
Auto scrolling sometimes not scrolling down in ChatView


ralphtheninja:

Jikstra:
How is it not working? What I experienced that it sometimes doesn't scroll down, but that's it.

ralphtheninja:
Yes, this is what I meant by not working :)

PR #188, opened 2018-10-20
We didn't select the correct element for the scroll down

Jikstra:
Fixes https://github.com/deltachat/deltachat-desktop/issues/183


ralphtheninja:
Nice!

Issue #180, opened 2018-10-18
Show sender/authors name in ChatView


Jikstra:
Didn't track it down completely but somewhere in the conversations stuff the profileAuthorName /DisplayName isn't getting pushed through correctly or something. But the profileAuthorName is given through the initial conversations stuff correctly or at least more is borken.

ralphtheninja:
Can you paste some links to where the problem is? And/or some screenshots.

ralphtheninja:
I think we should nail this one before releasing 1.5.0

r10s:
the android versions show the name of the senders directly above the text in a slightly different font/color.

on android, this is done _only_ for group chats, for normal chats, this is not required (skipped to save monitor space, however, on desktop maybe this optimization in not required)

PR #192, opened 2018-10-20
Set conversationType correctly (always set to direct)

ralphtheninja:
Closes https://github.com/deltachat/deltachat-desktop/issues/180

Result:

screenshot from 2018-10-20 15-35-25


ralphtheninja:
Hmm, let me re-adjust this a little bit. Better to pass in `conversationType` and not calculate that for each message.

Issue #103, opened 2018-09-30
"About" under "help" menu


hpk42:
No easy way to find out which version my desktop is -- maybe add an ``About`` entry with a bit of info (also which version of deltachat-core and maybe other libs this is running).

PR #197, opened 2018-10-21
[wip] Add About dialog

ralphtheninja:
Closes https://github.com/deltachat/deltachat-desktop/issues/103

Currently looks really crappy since long values doesn't break properly and are rendered outside the dialog. But at least we have the complete flow and data ready.

screenshot from 2018-10-21 04-27-26

What's left:

- [ ] fix layout (css, or replace table with something better, suggestions welcome, basically `hilfe!`)
- [ ] the user _must_ be logged in to get the info so the menu alternative must be disabled when logged out
- [ ] `the_underscore_keys` should be translated

cc @karissa @Jikstra @hpk42

### DIFF
--- a/src/main/ipc.js
+++ b/src/main/ipc.js
@@ -58,6 +58,10 @@ function init (cwd) {
     e.returnValue = ret
   })
 
+  ipc.on('dispatchSyncNoRender', (e, ...args) => {
+    e.returnValue = dispatch(...args)
+  })
+
   // Calls the function without returning the value (async)
   ipc.on('dispatch', (e, ...args) => {
     dispatch(...args)
--- a/_locales/en.json
+++ b/_locales/en.json
@@ -255,6 +255,14 @@
         "message": "Report an Issue...",
         "description": "Report an Issue... menu item label"
     },
+    "menu.help.about": {
+        "message": "About DeltaChat",
+        "description": "About menu item"
+    },
+    "aboutDeltaChat": {
+        "message": "About DeltaChat",
+        "description": "Window title on about dialog"
+    },
     "timestampFormat_M": {
         "message": "MMM D",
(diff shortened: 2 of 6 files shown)


src/main/ipc.js
```diff
@@ -58,6 +58,10 @@ function init (cwd) {
     e.returnValue = ret
   })
 
+  ipc.on('dispatchSyncNoRender', (e, ...args) => {
+    e.returnValue = dispatch(...args)
+  })
```
ralphtheninja: Needed a separate message to be able to get data from the backend _without_ rendering automatically, i.e. get data outside of the normal state pipeline.

Jikstra:
> the user must be logged in to get the info so the menu alternative must be disabled when logged out

Hmm, why not only show the keys which we can display if the user is logged in? Not having an about when not logged in seems a bit weird to me. What's blocking us from always showing it?

ralphtheninja:
The current design doesn't allow us to query the DC context because it's created in the login method. We must tweak that first.

@hpk requested this, hence why I put it in about dialog. I both agree and don't agree. Might as well have an advanced tab in the about dialog, many programs does this. Why do we have to follow exactly what signal does?

At the same time I don't really care either, as long as we don't have to bikeshed everything.

ralphtheninja:
I'll leave this hanging for a while if anyone else has comments before closing.

Jikstra:
> The current design doesn't allow us to query the DC context because it's created in the login method. We must tweak that first.

Makes sense :)

> Might as well have an advanced tab in the about dialog, many programs does this. 

Sounds like a good solution. Most people won't care about those infos, might be important to have a better understanding of the user environment on bug reports.

ralphtheninja:
Hmm, multiple accounts make querying the dc context a bit problematic. Since we need to know the account, to get the folder in order to query it. Gah.

Jikstra:
Why did you close/delete the branch?

Issue #200, opened 2018-10-22
hashtags are evil!


hpk42:
i like being able to start using desktop, thanks for all the good work!

however, hash tags are evil -- please disable/remove/replace them if possible :) 
![evil-hash](https://user-images.githubusercontent.com/73579/47284603-3c900300-d5e8-11e8-980b-f940ee7b9623.png)

r10s:
iiuc they come from the avatars that have the same background color as the "bubble" in this case so that only the hashtag is left.

however, i also think  that it makes sense to disable them until they are shown in some context (using dc_chat_get_profile_image()/dc_contact_get_profile_image() (the latter is not yet implemented) and/or a colored fallback.

r10s:
great, much, better now :)

PR #201, opened 2018-10-22
Move avatar out of Message and display it next to the message (left or

Jikstra:
right) in groups
Fixes https://github.com/deltachat/deltachat-desktop/issues/200


Jikstra:
Oh and I think we should also show the avatar in 1to1 chats, we have enough space for it. It's weird to only show them in groups. Better stay consistent.

ralphtheninja:
Much cleaner now. Nice!

hpk42:
very nice -- am also for adding the avatar to 1:1 chats on desktop.

PR #206, opened 2018-10-23
Add About dialog

ralphtheninja:
Closes https://github.com/deltachat/deltachat-desktop/issues/103

Will leave this hanging for someone else to tweak the dialog. I'm not skilled enough and/or don't have enough creativity to make something out of the dialog.

Currently my thoughts are:

* add some tabs
* default tab shows basic and clean About information (desktop version, license, yada yada)
* click the second tab to get more information

Also see old PR https://github.com/deltachat/deltachat-desktop/pull/197

### DIFF
--- a/src/main/deltachat.js
+++ b/src/main/deltachat.js
@@ -153,6 +153,17 @@ class DeltaChatController {
   }
 
   /**
+   * Dispatched when showing the AboutDialog (owned by ScreenController)
+   */
+  getInfo () {
+    if (this.ready === true) {
+      return this._dc.getInfo()
+    } else {
+      return DeltaChat.getSystemInfo()
+    }
+  }
+
+  /**
    * Dispatched when sending a message in ChatView
    */
   sendMessage (chatId, text) {
--- a/src/main/deltachat.js
+++ b/src/main/deltachat.js
@@ -153,7 +153,7 @@ class DeltaChatController {
   }
 
   /**
-   * Dispatched when showing the AboutDialog (owned by ScreenController)
+   * TODO: Currently not used
    */
   getInfo () {
     if (this.ready === true) {
--- a/src/main/ipc.js
+++ b/src/main/ipc.js
@@ -54,6 +54,10 @@ function init (cwd) {
     e.returnValue = dispatch(...args)
   })
 
+  ipc.on('dispatchSyncNoRender', (e, ...args) => {
+    e.returnValue = dispatch(...args)
+  })
+
   // Calls the function without returning the value (async)
   ipc.on('dispatch', (e, ...args) => {
     dispatch(...args)
--- a/src/main/ipc.js
+++ b/src/main/ipc.js
@@ -54,10 +54,6 @@ function init (cwd) {
     e.returnValue = dispatch(...args)
   })
 
-  ipc.on('dispatchSyncNoRender', (e, ...args) => {
-    e.returnValue = dispatch(...args)
-  })
-
   // Calls the function without returning the value (async)
   ipc.on('dispatch', (e, ...args) => {
     dispatch(...args)
--- a/src/renderer/ScreenController.js
+++ b/src/renderer/ScreenController.js
@@ -17,7 +17,6 @@ class Home extends React.Component {
       screen: 'SplittedChatListAndView',
       screenProps: {},
       showAbout: false,
-      aboutInfo: {},
       message: false
     }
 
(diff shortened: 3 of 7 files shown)


src/main/deltachat.js
```diff
@@ -145,6 +145,17 @@ class DeltaChatController {
     if (typeof this._render === 'function') this._render()
   }
 
+  /**
+   * Dispatched when showing the AboutDialog (owned by ScreenController)
```
ralphtheninja: It has to be `ScreenController` that owns this, if we want to show it without being logged in.

src/main/deltachat.js
```diff
@@ -145,6 +145,17 @@ class DeltaChatController {
...
+   */
+  getInfo () {
+    if (this.ready === true) {
+      return this._dc.getInfo()
+    } else {
+      return DeltaChat.getSystemInfo()
```
ralphtheninja: This returns basic system information, which is a stripped down version of `dc.getInfo()`

src/renderer/ScreenController.js
```diff
@@ -37,13 +44,27 @@ class Home extends React.Component {
...
+        'dispatchSyncNoRender',
+        'getInfo'
+      )
+    }
+    this.setState({ showAbout, aboutInfo })
   }
 
   render () {
-    // renderer/main.js polls every second and updates the deltachat
-    // property with current state of database.
```
ralphtheninja: Old comment, we're no longer polling for data.

src/main/ipc.js
```diff
@@ -54,6 +54,10 @@ function init (cwd) {
     e.returnValue = dispatch(...args)
   })
 
+  ipc.on('dispatchSyncNoRender', (e, ...args) => {
+    e.returnValue = dispatch(...args)
+  })
```
ralphtheninja: This should be removed now.

src/main/deltachat.js
karissa: Yeah I have an idea on how to refactor this with a sort of `DialogController`, can take a look.

ralphtheninja:
@Jikstra Have you tried opening the dialog and display the advanced stuff while logged in? I'm thinking maybe we should either limit the info or somehow tweak the ui so we can display long strings in a nicer way, e.g. via line breaks etc

Jikstra:
> @Jikstra Have you tried opening the dialog and display the advanced stuff while logged in? I'm thinking maybe we should either limit the info or somehow tweak the ui so we can display long strings in a nicer way, e.g. via line breaks etc

Yes did it. It's ugly but not really possible to do it differently with the current html. We would need a table or something. I didn't want to further work on it as i think it's not the right place to put those infos. It's just a hack until we have settings in my opinion.

src/main/deltachat.js
ralphtheninja: @Jikstra We could skip the top line and just do `return DeltaChat.getSystemInfo()`.

ralphtheninja:
I think we should do one out of the following two things to round this off:

* _remove_ the collapse thingie entirely and just display a simple/clean about dialog (and move this info elsewhere in another issue)
* _only_ do `DeltaChat.getSystemInfo()` (this will give more details about the software, which I think is nice and at the same time don't be _too_ ugly with enormously long strings)

Jikstra:
> remove the collapse thingie entirely and just display a simple/clean about dialog (and move this info elsewhere in another issue)

I'm pro this. Should be somewhere in Settings->Advanced->Debug Information or something

karissa:
Yeah I think removing the collapse thingy is a good idea. Removing complexity

ralphtheninja:
Anyone against merging this now?

Jikstra:
> Anyone against merging this now?

Nope. LGTM let's merge.

src/renderer/ScreenController.js
```diff
@@ -87,6 +100,7 @@ class Home extends React.Component {
             deltachat={deltachat}
           />
         }
+        { showAbout && (<AboutDialog isOpen={showAbout} onClose={this.onCloseAbout} />) }
```
karissa: you don't need showAbout && if you have an `isOpen` parameter.

src/renderer/components/dialogs/About.js
```diff
@@ -0,0 +1,28 @@
...
+    return (
+      <Dialog
+        isOpen={isOpen}
+        title={tx('aboutDeltaChat')}
+        icon='info-sign'
+        onClose={onClose}
+        canOutsideClickClose={false}>
+        <div className={Classes.DIALOG_BODY}>
+          <p style={{ color: 'grey' }}>{'version ' + version}</p>
+          <p>Official Delta Chat Desktop app.</p>
```
karissa: we'll have to translate this but I guess it's secondary

src/renderer/ScreenController.js
ralphtheninja: More specifically, the overlay div will be part of the dom once it has been rendered:

screenshot from 2018-11-01 10-47-55

Lets just remove it. Makes the code cleaner as well.

ralphtheninja: https://github.com/deltachat/deltachat-desktop/commit/a8a4710f15f94c8330dac5d815b51617bd6294e4

Issue #212, opened 2018-10-24
Unread counter


Jikstra:
- Show counter with unread messages in ChatList

PR #213, opened 2018-10-24
[WIP] Implement fresh message counter in ChatList

Jikstra:
Tackles #212

To make this work properly, we should als implement markNoticedChat(chatId) and maybe even markSeenMessages(). For this i need to make some IPC stuff with which i'm unsure in the moment. Maybe @ralphtheninja  can help me with this?

EDIT: Implemented ipc dispatch to mark chat as noticed, only marking messages as read is missing.
EDIT1:
Bugs:
- if we are in a chat and get new messages, we need to switch the selected chat to get rid of the fresh message counter. Maybe we can check for a onFocus event or something and also call markNoticedChatIfNeeded()

EDIT2: Telegram is cleaning the fresh message counter for the selected chat on window focus

### DIFF
--- a/src/main/deltachat.js
+++ b/src/main/deltachat.js
@@ -307,7 +307,9 @@ class DeltaChatController {
     for (let i = 0; i < count; i++) {
       const chatId = list.getChatId(i)
       const chat = chatIdToJson(chatId, this._dc)
+      const freshMessageCounter = this._dc.getFreshMessageCount(chatId)
       chat.summary = list.getSummary(i).toJson()
+      chat.freshMessageCounter = freshMessageCounter
       chats.push(chat)
     }
     return chats
--- a/src/main/deltachat.js
+++ b/src/main/deltachat.js
@@ -32,6 +32,7 @@ function chatIdToJson (chatId, dc) {
       }
     }
   }
+  chat.freshMessageCounter = this._dc.getFreshMessageCount(chatId)
   return chat
 }
 
@@ -312,9 +313,7 @@ class DeltaChatController {
     for (let i = 0; i < count; i++) {
       const chatId = list.getChatId(i)
       const chat = chatIdToJson(chatId, this._dc)
-      const freshMessageCounter = this._dc.getFreshMessageCount(chatId)
       chat.summary = list.getSummary(i).toJson()
-      chat.freshMessageCounter = freshMessageCounter
       chats.push(chat)
     }
     return chats
--- a/src/main/deltachat.js
+++ b/src/main/deltachat.js
@@ -282,6 +282,11 @@ class DeltaChatController {
     this._dc.removeContactFromChat(chatId, C.DC_CONTACT_ID_SELF)
   }
 
+  markNoticedChat(chatId) {
+    log('mark noticed chat', chatId)
+    this._dc.markNoticedChat(chatId)
+  }
+
   /**
    * Returns the state in json format
    */
--- a/src/main/deltachat.js
+++ b/src/main/deltachat.js
@@ -32,7 +32,7 @@ function chatIdToJson (chatId, dc) {
       }
     }
   }
-  chat.freshMessageCounter = this._dc.getFreshMessageCount(chatId)
+  chat.freshMessageCounter = dc.getFreshMessageCount(chatId)
   return chat
 }
 
--- a/src/main/deltachat.js
+++ b/src/main/deltachat.js
@@ -282,7 +282,7 @@ class DeltaChatController {
     this._dc.removeContactFromChat(chatId, C.DC_CONTACT_ID_SELF)
   }
 
-  markNoticedChat(chatId) {
+  markNoticedChat (chatId) {
     log('mark noticed chat', chatId)
     this._dc.markNoticedChat(chatId)
   }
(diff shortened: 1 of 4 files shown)


src/renderer/components/ChatList.js
```diff
@@ -20,6 +20,7 @@ class ChatList extends React.Component {
         <div className='ConversationList'>
           {deltachat.chats.map((chat) => {
             if (!chat) return
+            console.log(chat)
```
karissa: console.log

src/renderer/components/SplittedChatListAndView.js
```diff
@@ -41,6 +41,7 @@ class SplittedChatListAndView extends React.Component {
     this.onCreateContact = this.onCreateContact.bind(this)
     this.initiateKeyTransfer = this.initiateKeyTransfer.bind(this)
     this.onKeyTransferComplete = this.onKeyTransferComplete.bind(this)
+    this.markNoticedChatIfNeeded = this.markNoticedChatIfNeeded.bind(this)
```
karissa: this is only needed when the function is used as a button in a React element, so you can remove it

src/main/deltachat.js
```diff
@@ -307,7 +312,9 @@ class DeltaChatController {
     for (let i = 0; i < count; i++) {
       const chatId = list.getChatId(i)
       const chat = chatIdToJson(chatId, this._dc)
+      const freshMessageCounter = this._dc.getFreshMessageCount(chatId)
       chat.summary = list.getSummary(i).toJson()
+      chat.freshMessageCounter = freshMessageCounter
```
karissa: instead, let's put this in the chatIdToJson function. The only reason we have `chat.summary` here is because we need `list` in order to `getSummary` on the chat.

```
chat.freshMessageCounter = this._dc.getFreshMessageCount(chatId)
```

src/renderer/components/SplittedChatListAndView.js
```diff
@@ -41,6 +41,9 @@ class SplittedChatListAndView extends React.Component {
     this.onCreateContact = this.onCreateContact.bind(this)
     this.initiateKeyTransfer = this.initiateKeyTransfer.bind(this)
     this.onKeyTransferComplete = this.onKeyTransferComplete.bind(this)
+    this.onWindowFocus = this.onWindowFocus.bind(this)
+
+    window.onfocus = this.onWindowFocus
```
karissa: By using `onfocus`, is the message only marked noticed when you click on the ChatView window? The behavior of `onfocus` in the window seems a bit dubious, perhaps 

Maybe it would cover more edge cases if this happened instead on componentDidUpdate where chatId is different than the previous chatId? I think that will catch anytime the chatview is updated with a new chat id.

src/renderer/components/SplittedChatListAndView.js
```diff
@@ -27,7 +27,7 @@ class SplittedChatListAndView extends React.Component {
     this.state = {
       deadDropChat: false,
       keyTransfer: false,
-      selectedChatId: this.getInitiallySelectedChatId()
+      selectedChatId: null
```
ralphtheninja: Good! I also noticed it's not needed.

src/renderer/components/SplittedChatListAndView.js
Jikstra: onfocus is obsolete, changed the behaviour to allways select the chat as noticed if its the selected one, also on rerender. So if new messages come in the selected chat,they are always noticed

src/renderer/components/ChatView.js
```diff
@@ -29,6 +29,8 @@ class ChatView extends React.Component {
     this.focusInputMessage = this.focusInputMessage.bind(this)
     this.scrollToBottom = this.scrollToBottom.bind(this)
     this.conversationDiv = React.createRef()
+
+    console.log(this.conversationDiv)
   }
```
ralphtheninja: `console.log()`

ralphtheninja: `console.log()`

src/renderer/components/SplittedChatListAndView.js
ralphtheninja: Good! I also noticed it's not needed.

src/renderer/components/SplittedChatListAndView.js
```diff
@@ -140,6 +155,8 @@ class SplittedChatListAndView extends React.Component {
       selectedChatId = this.state.selectedChatId = this.getInitiallySelectedChatId()
     }
 
+    this.markNoticedChatIfNeeded(selectedChatId)
```
ralphtheninja: Is this needed or can we move this elsewhere? I'd rather not do ipc stuff inside a render.

ralphtheninja: Not the end of the world if it has to stay :)

Jikstra: it's needed, but only done if the freshMessagesCounter in the selectedChat is > 0. Maybe we can do this before render.

ralphtheninja: Leave it, with a TODO comments. Lets figure out later.

Issue #210, opened 2018-10-24
Window title changes from Delta.Chat to DeltaChat (BETA) after login


ralphtheninja:
![screenshot from 2018-10-24 17-28-56](https://user-images.githubusercontent.com/308049/47442276-6b59d500-d7b2-11e8-9bfc-c7c10d986f03.png)

![screenshot from 2018-10-24 17-29-06](https://user-images.githubusercontent.com/308049/47442289-70b71f80-d7b2-11e8-9958-702f5fec0c46.png)

ralphtheninja:
Can we agree on what it should be? I personally prefer `DeltaChat`. I don't really know why we use `Delta.Chat` sometimes and sometimes not.

ralphtheninja:
Side note. I'd like to put more information in the window title, maybe the logged in account? Or is that not a kosher thing to do? :smile:

karissa:
@ralphtheninja changing window title is fine! you can do this in windows/main.js, `BrowserWindow` object should have a `setTitle` command or something similar.

hpk42:
i think DeltaChat is the canonical string.  We should try to get 
rid of "Delta.Chat" -- it's hard to write in texts/reports/mails ...
and "deltachat.org" redirects to "delta.chat" FWIW.
Also "deltachat" in google and duckduckgo goes to the
correct domain ...

ralphtheninja:
Also go through translation files.

r10s:
> Side note. I'd like to put more information in the window title, maybe the logged in account? Or is that not a kosher thing to do? smile

i would suggest the same, just sth. as `Delta Chat - user@example.org`.
the title may be shown in many places eg. windows app bar.

i would not put more information to the title nor use square brackets or so. i think this looks too technical and always a bit messy when the title is truncated.

PR #222, opened 2018-10-25
Put logged in user in window title

ralphtheninja:
Closes https://github.com/deltachat/deltachat-desktop/issues/210

### DIFF
--- a/src/main/deltachat.js
+++ b/src/main/deltachat.js
@@ -45,26 +45,19 @@ class DeltaChatController {
    */
   constructor (cwd) {
     this.cwd = cwd
-    this.ready = false
-    this.credentials = {
-      email: null,
-      cwd: null
-    }
-    this._selectedChatId = null
-    this._showArchivedChats = false
+    this._resetState()
   }
 
   /**
    * Dispatched when logging in from Login
    */
-  login (credentials, render) {
+  login ({ email, password }, render) {
     // Creates a separate DB file for each login
-    const cwd = path.join(this.cwd, Buffer.from(credentials.email).toString('hex'))
+    const cwd = path.join(this.cwd, Buffer.from(email).toString('hex'))
     log('Using deltachat instance', cwd)
     this._dc = new DeltaChat()
     var dc = this._dc
-    this.credentials.email = credentials.email
-    this.credentials.cwd = cwd
+    this.credentials = { email }
     this._render = render
 
     dc.open(cwd, err => {
@@ -73,9 +66,8 @@ class DeltaChatController {
         log('Ready')
         this.ready = true
         this.configuring = false
-        const addr = credentials.email
-        if (!electron.app.logins.includes(addr)) {
-          electron.app.logins.push(addr)
+        if (!electron.app.logins.includes(email)) {
+          electron.app.logins.push(email)
         }
         render()
       }
@@ -83,8 +75,8 @@ class DeltaChatController {
         dc.once('ready', onReady)
         this.configuring = true
         dc.configure({
-          addr: credentials.email,
-          mail_pw: credentials.password
+          addr: email,
+          mail_pw: password
         })
         render()
       } else {
@@ -147,8 +139,9 @@ class DeltaChatController {
   logout () {
     this._dc.close()
     this._dc = null
-    this.configuring = false
-    this.ready = false
+
+    this._resetState()
+
     log('Logged out')
     if (typeof this._render === 'function') this._render()
   }
@@ -377,6 +370,18 @@ class DeltaChatController {
    */
   _warning (line) { log('WARNING', line) }
   _error (line) { log.error(line) }
+
+  /**
+   * Internal
+   * Reset state related to login
+   */
+  _resetState () {
+    this.ready = false
+    this.configuring = false
+    this.credentials = { email: '' }
+    this._selectedChatId = null
+    this._showArchivedChats = false
+  }
 }
 
 module.exports = DeltaChatController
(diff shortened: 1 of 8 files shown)


ralphtheninja:
Currently looks like:

screenshot from 2018-10-25 16-34-18

Maybe we should slide the chat title etc to the left now?

src/main/deltachat.js
```diff
@@ -45,26 +45,19 @@ class DeltaChatController {
...
   constructor (cwd) {
     this.cwd = cwd
-    this.ready = false
-    this.credentials = {
-      email: null,
-      cwd: null
-    }
-    this._selectedChatId = null
-    this._showArchivedChats = false
+    this._resetState()
```
ralphtheninja: Moved this code to `_resetState()` since we need to do this in `logout()` as well. Also `this.credentials.cwd` was not used.

src/main/ipc.js
```diff
@@ -28,7 +28,6 @@ function init (cwd) {
   ipc.on('setAspectRatio', (e, ...args) => main.setAspectRatio(...args))
   ipc.on('setBounds', (e, ...args) => main.setBounds(...args))
   ipc.on('setProgress', (e, ...args) => main.setProgress(...args))
-  ipc.on('setTitle', (e, ...args) => main.setTitle(...args))
```
ralphtheninja: Skipped this message, since we can take care of this directly in render. Also, the backend should decide the window title, just like any state.

src/main/ipc.js
```diff
@@ -92,6 +91,8 @@ function init (cwd) {
 
   function render () {
     log('RENDER')
-    windows.main.send('render', dc.render())
+    const json = dc.render()
+    windows.main.setTitle(json.credentials.email)
```
ralphtheninja: Here.

src/main/windows/main.js
```diff
@@ -136,7 +136,11 @@ function setProgress (progress) {
 
 function setTitle (title) {
   if (!main.win) return
-  main.win.setTitle(title)
+  if (title) {
+    main.win.setTitle(`${config.APP_WINDOW_TITLE} - ${title}`)
+  } else {
+    main.win.setTitle(config.APP_WINDOW_TITLE)
+  }
```
ralphtheninja: We force the title to always use the `APP_WINDOW_TITLE` with an optional email.

ralphtheninja: Maybe should rename `title` to `email`.

src/renderer/components/SplittedChatListAndView.js
```diff
@@ -158,7 +158,6 @@ class SplittedChatListAndView extends React.Component {
           <Navbar fixedToTop>
             <NavbarGroup align={Alignment.LEFT}>
               { showArchivedChats && (<Button className={Classes.MINIMAL} icon='undo' onClick={this.onHideArchivedChats} />) }
-              <NavbarHeading>{deltachat.credentials.email}</NavbarHeading>
```
ralphtheninja: Moved to window title.

src/renderer/lib/state.js
```diff
@@ -28,16 +28,10 @@ function getDefaultState () {
...
-      progress: -1,
-      badge: null
-    },
     window: {
       bounds: null, /* {x, y, width, height } */
       isFocused: true,
-      isFullScreen: false,
-      title: config.APP_WINDOW_TITLE
+      isFullScreen: false
     },
```
ralphtheninja: Threw this away.

src/renderer/main.js
```diff
@@ -76,12 +76,4 @@ function onWindowBoundsChanged (e, newBounds) {
...
   state.deltachat = deltachat
   app.setState(state)
-  updateElectron()
-}
-
-function updateElectron () {
-  if (state.window.title !== state.prev.title) {
-    state.prev.title = state.window.title
-    ipcRenderer.send('setTitle', state.window.title)
-  }
```
ralphtheninja: Also threw this away, since the renderer process never updates the window title.

hpk42:
could you for the time being also say "DeltaChat (preview-VERSION)" instead of "(Beta)"? 
The F-droid version is currently labeled  "Beta" so it's better to say for now
that desktop is a preview release, not really a beta :)

r10s:
> Maybe we should slide the chat title etc to the left now?

i would suggest to leave it above the chat only. it changes when sth. is selected on the left side, so it is not really the title of the whole view.

instead, we could make the chatlist a bit higher - or add the sandwich button here?
i am not really sure if it is placed correctly to the very right, functions as "Add chat" are related to the chatlist on the left where the button is most right.
i think sooner or later the title bar will also get colored, this should also be regarded ...

but these are questions i would not really target currently, maybe it is just the best to leave it as it is now :)

ralphtheninja:
@hpk42 Comme ca?

screenshot from 2018-10-25 18-28-54

ralphtheninja:
Think I'll make a new release after this PR. Quite a bit has happened the past week.

ralphtheninja:
@hpk42 Merge if you're happy with it.

hpk42:
On Thu, Oct 25, 2018 at 09:29 -0700, Lars-Magnus Skog wrote:
> @hpk42 Comme ca?
> 
> screenshot from 2018-10-25 18-28-54

nice, thanks :)

src/main/windows/main.js
karissa: at some point we might want to add other kinds of titles here so i think the abstraction is a good one

src/renderer/lib/state.js
karissa: if there's no title set what does it look like on startup (is there a second where there's no title?)

src/renderer/main.js
karissa: yet :P

PR #230, opened 2018-10-28
Make emojis work on incoming messages

ralphtheninja:
Below follows _all_ emojis that can be sent from `deltachat-android` (yes, I clicked through all of them):

Emoji size also depends on length of message. If only sending a few emojis, they will be larger. But if you send with text or a lot of emojis, the smaller size will be picked.

screenshot from 2018-10-27 13-47-46

screenshot from 2018-10-27 13-47-55

screenshot from 2018-10-28 10-54-14

### DIFF
--- a/_locales/en.json
+++ b/_locales/en.json
@@ -462,5 +462,9 @@
     "language.zh_CN": {
         "message": "Chinese",
         "description": "Menu alternative in Preferences -> Choose Language..."
+    },
+    "emojiAlt": {
+        "message": "Emoji of $1$",
+        "description": "Used with aria-label for emojis"
     }
 }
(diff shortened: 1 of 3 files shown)


ralphtheninja:
This increases build size somewhat. The `AppImage` ends up at 71Mb (from ~59Mb) so it's not too bad. We can also try to reduce the size of the emojis, maybe even make them 32x32 instead of 64x64. We can also try removing other stuff to reduce the total build size.

r10s:
well, we come to one of the most important functions of Delta Chat now :)

but, no kidding, using the same icons on different Delta Chat versions is very important imho.
also emoticons as such, i know some people which check the available icons and stickers _first of all_.

ralphtheninja:
Anyone against this PR or can we get it merged?

karissa:
lGTM

Issue #229, opened 2018-10-27
hangs at missing config file: creating new one


substack:
This is the output I see when I try to run `npm start`:

```
$ npm start

> deltachat-desktop@1.7.0 start /home/substack/projects/deltachat-desktop
> npm run build && electron .


> deltachat-desktop@1.7.0 build /home/substack/projects/deltachat-desktop
> npm run build-ts && npm run build-css && npm run build-es5


> deltachat-desktop@1.7.0 build-ts /home/substack/projects/deltachat-desktop
> cd conversations && tsc


> deltachat-desktop@1.7.0 build-css /home/substack/projects/deltachat-desktop
> cd conversations && node-sass stylesheets/manifest.scss build/manifest.css

Rendering Complete, saving .css file...
Wrote CSS to /home/substack/projects/deltachat-desktop/conversations/build/manifest.css

> deltachat-desktop@1.7.0 build-es5 /home/substack/projects/deltachat-desktop
> buble src --output es5

Missing config file: Creating new one
```

and then it hangs.

ralphtheninja:
Do you know if there's an `err` set?

https://github.com/deltachat/deltachat-desktop/blob/cd537333b7b395baa15e1c934c6ba1c85ca6d425/src/renderer/lib/state.js#L73-L81

substack:
The `err` is `null`.

Jikstra:
@substack  What's your system language?

substack:
This issue is fixed if I downgrade electron to @^2.

PR #232, opened 2018-10-29
downgrade to electron 2, fixes https://github.com/deltachat/deltachat…

substack:
fixes issue #229 on linux where the window isn't created


karissa:
great thanks!

PR #241, opened 2018-10-30
don't select recent chat by default, fixes #228

### LINKED ISSUE #228: Durign initial sync we select different chats and clean fresh message counter
This makes us lose overview over new messages. Generally the initial sync is bad in my opinion.

karissa:
screenshot from 2018-10-30 15-42-32


ralphtheninja:
nice!

PR #243, opened 2018-10-30
Block a contact/Unblock contacts

karissa:
* See currently blocked contacts
* Block a contact
* Refactor dialogs, add shared 'confirmation' dialog code
* Dialog for confirming blocking a contact
* Dialog for confirming unblocking a contact

### DIFF
--- a/src/renderer/components/dialogs/index.js
+++ b/src/renderer/components/dialogs/index.js
@@ -1,0 +1,21 @@
+const { remote } = require('electron')
+
+const KeyTransfer = require('./KeyTransfer')
+const SetupMessage = require('./SetupMessage')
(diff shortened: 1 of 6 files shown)


src/renderer/components/dialogs/index.js
```diff
@@ -0,0 +1,21 @@
...
+  SetupMessage,
+  DeadDrop,
+  KeyTransfer
+}
+
+function confirmation (message, cb) {
+  var opts = {
+    type: 'question',
+    message: message,
+    buttons: ['No', 'Yes']
```
ralphtheninja: Are these buttons translated?

karissa: No :P

Issue #239, opened 2018-10-29
Advanced login options doesn't set serverFlags yet


Jikstra:
Therefore imap and smtp security options have no effect

PR #244, opened 2018-10-30
 Fix setting server flags for send/mailSecurity and fix some react wa…

Jikstra:
…rnings. Fixes https://github.com/deltachat/deltachat-desktop/issues/239

### DIFF
--- a/src/renderer/components/Login.js
+++ b/src/renderer/components/Login.js
@@ -23,7 +23,7 @@ class Login extends React.Component {
     this.handleShowAdvanced = this.handleShowAdvanced.bind(this)
   }
 
-  _defaultState(showAdvanced) {
+  _defaultState (showAdvanced) {
     return {
       addr: process.env.DC_ADDR ? process.env.DC_ADDR : '',
       mailUser: '',
@@ -36,7 +36,7 @@ class Login extends React.Component {
       sendServer: '',
       sendPort: '',
       sendSecurity: '',
-      showAdvanced: showAdvanced,
+      showAdvanced: showAdvanced
     }
   }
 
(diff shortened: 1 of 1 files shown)


Jikstra:
wtf somehow the about dialog stuff got into it.

src/renderer/components/Login.js
```diff
@@ -51,7 +83,18 @@ class Login extends React.Component {
...
+        this.state.sendSecurity
+      )
+    }
+
+    // Remove all properties where value is an empty string, gets us around
+    // React not liking state properties where the value is null/undefined if
+    // bound to an input value
+    for (let key in credentials) {
+      if (credentials[key] === '') delete credentials[key]
     }
```
karissa: when would it be null/undefined?

src/renderer/components/Login.js
```diff
@@ -15,32 +16,63 @@ const {
...
     var state = {}
     state[event.target.id] = event.target.value
     this.setState(state)
   }
 
+  translateSecurityToServerFlags (mailSecurity, sendSecurity) {
+    let serverFlags = []
+    if (mailSecurity) {
+      let imapFlag
+      if (mailSecurity === 'ssl') {
```
karissa: perhaps we could create a constant dictionary and look up the value in the dictionary instead of having `if` statements.

karissa:
@ralphtheninja @Jikstra mind waiting until I can weigh in before merging PRs?

src/renderer/components/Login.js
Jikstra: React doesn't like having a null or undefined value bound to an input value, i get a warning and it should be an empty string. But empty strings aren't seen as nothing in the core/login procedure so i'm cleaning them out. Otherwise we're doing a set_config for all keys.

src/renderer/components/Login.js
Jikstra: May be better, will check it out.

PR #254, opened 2018-10-31
Hide password by default and add button to show it

Jikstra:
Fixes https://github.com/deltachat/deltachat-desktop/issues/207

### DIFF
--- a/src/renderer/components/Login.js
+++ b/src/renderer/components/Login.js
@@ -48,7 +48,8 @@ class Login extends React.Component {
   }
 
   handleCredentialsChange (event) {
-    this.setState({ credentials: { [event.target.id]: event.target.value } })
+    let stateCredentials = Object.assign(this.state.credentials, { [event.target.id]: event.target.value })
+    this.setState(stateCredentials)
   }
 
   translateSecurityToServerFlags (mailSecurity, sendSecurity) {
@@ -85,7 +86,8 @@ class Login extends React.Component {
   }
 
   handleUISwitchStateProperty (key) {
-    this.setState({ ui: { [key]: !this.state.ui[key] } })
+    let stateUi = Object.assign(this.state.ui, { [key]: !this.state.ui[key] })
+    this.setState(stateUi)
   }
 
   cancelClick (event) {
--- a/src/renderer/components/Login.js
+++ b/src/renderer/components/Login.js
@@ -18,14 +18,21 @@ const {
 class Login extends React.Component {
   constructor (props) {
     super(props)
-    this.state = this._defaultState(false, false, false)
+    this.state = {
+      credentials: this._defaultCredentials(),
+      ui: {
+        showAdvanced: false,
+        showPasswordMail: false,
+        showPasswordSend: false
+      }
+    }
 
-    this.handleChange = this.handleChange.bind(this)
+    this.handleCredentialsChange = this.handleCredentialsChange.bind(this)
     this.handleSubmit = this.handleSubmit.bind(this)
     this.renderPasswordInput = this.renderPasswordInput.bind(this)
   }
 
-  _defaultState (showAdvanced, showPasswordMail, showPasswordSend) {
+  _defaultCredentials () {
     return {
       addr: process.env.DC_ADDR ? process.env.DC_ADDR : '',
       mailUser: '',
@@ -38,17 +45,12 @@ class Login extends React.Component {
       sendServer: '',
       sendPort: '',
       sendSecurity: '',
-      showAdvanced: showAdvanced,
-      showPasswordMail: showPasswordMail,
-      showPasswordSend: showPasswordSend
     }
   }
 
-  handleChange (event) {
-    console.log('handleChange', event.target.id, event.target.value)
-    var state = {}
-    state[event.target.id] = event.target.value
-    this.setState(state)
+  handleCredentialsChange (event) {
+    console.log('handleCredentialsChange', event.target.id, event.target.value)
+    this.setState({credentials: {[event.target.id]: event.target.value}})
   }
 
   translateSecurityToServerFlags (mailSecurity, sendSecurity) {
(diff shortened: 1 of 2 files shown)


src/renderer/components/Login.js
```diff
@@ -10,20 +10,21 @@ const {
...
-    this.state = this._defaultState(false)
+    this.state = this._defaultState(false, false, false)
 
     this.handleChange = this.handleChange.bind(this)
     this.handleSubmit = this.handleSubmit.bind(this)
-    this.handleShowAdvanced = this.handleShowAdvanced.bind(this)
   }
 
-  _defaultState (showAdvanced) {
+  _defaultState (showAdvanced, showPasswordMail, showPasswordSend) {
```
karissa: instead of this, can _defaultState take an `opts` parameter which will override the default options using Object.assign?

ex: 
`
_defaultState (opts) {
  return Object.assign(theBigStateObject, opts || {})
}

src/renderer/components/Login.js
```diff
@@ -44,6 +45,7 @@ class Login extends React.Component {
   }
 
   handleChange (event) {
+    console.log('handleChange', event.target.id, event.target.value)
```
karissa: console.log

src/renderer/components/Login.js
```diff
@@ -35,15 +43,13 @@ class Login extends React.Component {
...
+      sendSecurity: ''
     }
   }
 
-  handleChange (event) {
-    var state = {}
-    state[event.target.id] = event.target.value
-    this.setState(state)
+  handleCredentialsChange (event) {
+    console.log('handleCredentialsChange', event.target.id, event.target.value)
```
ralphtheninja: `console.log`

src/renderer/components/Login.js
```diff
@@ -74,41 +80,19 @@ class Login extends React.Component {
...
     // TODO: Implement security
-    ipcRenderer.send('login', credentials)
+    ipcRenderer.send('login', this.state.credentials)
     event.preventDefault()
   }
 
-  handleShowAdvanced () {
-    this.setState({ showAdvanced: !this.state.showAdvanced })
+  handleUISwitchStateProperty (key) {
+    console.log('a', this.state.ui[key])
```
ralphtheninja: `console.log`

src/renderer/components/Login.js
```diff
@@ -117,12 +101,54 @@ class Login extends React.Component {
...
+        rightElement={lockButton}
+      />
+    )
+  }
+
   render () {
     const { logins, deltachat } = this.props
-    const { addr, mailUser, mailPw, mailServer, mailPort, mailSecurity, sendUser, sendPw, sendServer, sendPort, sendSecurity } = this.state
+
+    console.log(this.state)
```
ralphtheninja: `console.log`

src/renderer/components/Login.js
```diff
@@ -117,12 +101,54 @@ class Login extends React.Component {
     ipcRenderer.send('login', { addr: login, mailPw: true })
   }
 
+  renderPasswordInput (keyShowPassword, keyValue) {
+    const tx = window.translate
+
+    let lockButton = (
```
ralphtheninja: `const`

src/renderer/components/Login.js
```diff
@@ -117,12 +101,54 @@ class Login extends React.Component {
...
+      sendServer,
+      sendPort,
+      sendSecurity
+    } = this.state.credentials
+
+    const { showAdvanced } = this.state.ui
+
     const tx = window.translate
 
     var loading = deltachat.configuring
```
ralphtheninja: `const`

src/renderer/components/Login.js
Jikstra: you want a `let`? Help me :D

ralphtheninja: I always want `const` :D (it's also commenting just beneath the row)

Jikstra: But isn't const behaving differently on the scope? I use let always when its just temporary and only for this scope. But as soon as it's not changing you want a `const`?

ralphtheninja: > But isn't const behaving differently on the scope?

No, `const` and `let` behave the same.

> But as soon as it's not changing you want a `const`?

Yep.

src/renderer/components/Login.js
```diff
@@ -35,15 +43,12 @@ class Login extends React.Component {
...
+      sendSecurity: ''
     }
   }
 
-  handleChange (event) {
-    var state = {}
-    state[event.target.id] = event.target.value
-    this.setState(state)
+  handleCredentialsChange (event) {
+    this.setState({ credentials: { [event.target.id]: event.target.value } })
```
karissa: this makes me nervous for some reason, can we verify that these keys are within a range of accepted keys? Sorry to be nit=picky, but this seems like an important part of the process and i'd worry that somehow the event.target.id/value could be manipulated.. any thoughts?

ralphtheninja: If we assume they can be manipulated. Would it matter?

src/renderer/components/Login.js
```diff
@@ -144,28 +168,22 @@ class Login extends React.Component {
...
-                leftIcon='lock'
-                type='mailPw'
-                value={mailPw}
-                onChange={this.handleChange}
-              />
+              {this.renderPasswordInput('showPasswordMail', 'mailPw')}
             </FormGroup>
-            <Button onClick={this.handleShowAdvanced}>{(this.state.showAdvanced ? '-' : '+') + ' ' + tx('login.advanced') }</Button>
-            <Collapse isOpen={this.state.showAdvanced}>
+            <Button onClick={this.handleUISwitchStateProperty.bind(this, 'showAdvanced')}>{(showAdvanced ? '-' : '+') + ' ' + tx('login.advanced') }</Button>
```
karissa: I wonder if this also should be internationalized? but it's probably good enough for now

karissa:
LGTM!

Jikstra:
Sooo lets merge this? @karissa @ralphtheninja

ralphtheninja:
> Sooo lets merge this? @karissa @ralphtheninja

Sure!

Issue #110, opened 2018-10-01
missing libssl


gerroon:
Hi

I am trying to iuse it on Debian testing, I tried both the appimagea and the .deb, both complaints about missing  libssl.so.1.0.0: 

A JavaScript error occurred in the main process
Uncaught Exception:
Error: libssl.so.1.0.0: cannot open shared object file: No such file or directory
    at process.module.(anonymous function) [as dlopen] (ELECTRON_ASAR.js:166:20)
    at Object.Module._extensions..node (module.js:671:18)
    at Object.module.(anonymous function) [as .node] (ELECTRON_ASAR.js:180:18)
    at Module.load (module.js:561:32)
    at tryModuleLoad (module.js:504:12)
    at Function.Module._load (module.js:496:3)
    at Module.require (module.js:586:17)
    at require (internal/module.js:11:18)
    at load (/tmp/.mount_deltacPxA4uR/resources/app.asar/node_modules/node-gyp-build/index.js:13:10)

thanks

ralphtheninja:
@gerroon Thanks for input! Tried installing `libssl1.0.0`?

ralphtheninja:
@gerroon Could you try this release? https://github.com/deltachat/deltachat-desktop/releases/tag/v1.2.3-bundled-openssl

gerroon:
yeah it works (the ui opens0 with the bundled openssl. I cant test the account stuff though since my server requires manual settings.

gerroon:
Still an issue in ./deltachat-desktop-1.3.0-x86_64.AppImage

A JavaScript error occurred in the main process
Uncaught Exception:
Error: libssl.so.1.0.0: cannot open shared object file: No such file or directory
    at process.module.(anonymous function) [as dlopen] (ELECTRON_ASAR.js:166:20)
    at Object.Module._extensions..node (module.js:671:18)
    at Object.module.(anonymous function) [as .node] (ELECTRON_ASAR.js:180:18)
    at Module.load (module.js:561:32)
    at tryModuleLoad (module.js:504:12)
    at Function.Module._load (module.js:496:3)
    at Module.require (module.js:586:17)
    at require (internal/module.js:11:18)
    at load (/tmp/.mount_deltacN7yBtA/resources/app.asar/node_modules/node-gyp-build/index.js:13:10)
    at Object.<anonymous> (/tmp/.mount_deltacN7yBtA/resources/app.asar/node_modules/deltachat-node/binding.js:1:198)

[...]

PR #255, opened 2018-10-31
Install libssl-dev on Travis

ralphtheninja:
_Addresses_ https://github.com/deltachat/deltachat-desktop/issues/110 (we need to make a new release so e.g. @gerroon can test this)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ addons:
       - python3-pip
       - python3.5
       - g++-7
+      - libssl-dev
 
 before_install:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then export CC=gcc-7; fi


.travis.yml
```diff
@@ -16,6 +16,7 @@ addons:
       - python3-pip
       - python3.5
       - g++-7
+      - libssl-dev
```
ralphtheninja: @flub Maybe we should install `libssl1.1-dev`? (to solve #110)

ralphtheninja:
@gerroon I'll merge this and maybe you can test next release. If that doesn't cut it, lets try installing `libssl1.1-dev`.

.travis.yml
flub: Hum, on my debian testing machine there's only `libssl-dev` (installing 1.1.x) and `libssl1.0-dev` which kind of makes sense.  I think with 1.0 we do not get as many compile-time warnings as with 1.1 but it both work afaik.

Issue #112, opened 2018-10-03
Failed to load video message


ralphtheninja:
Platform: Linux

I sent a video message from my phone to desktop with an empty text message as a result.

karissa:
Is there a video screenshot @r10s ?

karissa:
@ralphtheninja can you be a bit more specific? What was displayed on the screen?

ralphtheninja:
> @ralphtheninja can you be a bit more specific? What was displayed on the screen?

An empty text message.

ralphtheninja:
Just to clarify. This is what I see on desktop when sending video from mobile:

![screenshot from 2018-10-31 21-47-28](https://user-images.githubusercontent.com/308049/47817697-a336be80-dd56-11e8-949b-f5f6af126655.png)

ralphtheninja:
It seems `conversations` needs `screenshot.url` to be set and we somehow need to get/generate a thumbnail for the video:

![screenshot from 2018-10-31 22-10-48](https://user-images.githubusercontent.com/308049/47819189-142ba580-dd5a-11e8-83d9-1b26e166cd18.png)

PR #256, opened 2018-10-31
Use <video> element for video attachments

ralphtheninja:
Closes https://github.com/deltachat/deltachat-desktop/issues/112

@#$%^@ the `` tag, lets just use `<video>` and be done with it?

Result:

screenshot from 2018-10-31 22-58-34

### DIFF
--- a/src/renderer/components/ChatView.js
+++ b/src/renderer/components/ChatView.js
@@ -213,7 +213,7 @@ class RenderMessage extends React.Component {
       props.text = msg.text
     }
 
-    return (<Message {...props} />)
+    return (<div className='MessageWrapper'><Message {...props} /></div>)
   }
 }
 
--- a/src/renderer/components/ChatView.js
+++ b/src/renderer/components/ChatView.js
@@ -204,7 +204,11 @@ class RenderMessage extends React.Component {
     }
 
     if (msg.file) {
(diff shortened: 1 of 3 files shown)


ralphtheninja:
Might need some css love though because the "lock" and timestamp are slightly hidden underneath.

ralphtheninja:
@Jikstra Probably faster if you add a commit with appropriate css changes on this branch.

ralphtheninja:
screenshot from 2018-11-02 10-16-16

src/renderer/components/ChatView.js
```diff
@@ -204,12 +204,16 @@ class RenderMessage extends React.Component {
...
+        url: msg.file,
+        contentType: convertContentType(message.filemime),
+        filename: msg.text
+      }
     } else {
       props.text = msg.text
     }
 
-    return (<Message {...props} />)
+    return (<div className='MessageWrapper'><Message {...props} /></div>)
```
karissa: this would also be a good case for `styled-components`

ralphtheninja: TIL

karissa:
lgtm, let's refactor to use styled-components (or something similar) soon though otherwise the css file will start getting really frustratingly long

PR #261, opened 2018-11-01
[wip] Load messages on demand

karissa:
this is not 100% working.

### DIFF
--- a/src/renderer/components/ChatView.js
+++ b/src/renderer/components/ChatView.js
@@ -9,6 +9,8 @@ const { Overlay } = require('@blueprintjs/core')
 const MutationObserver = window.MutationObserver
 const IntersectionObserver = window.IntersectionObserver
 
+const { List, InfiniteLoader } = require('react-virtualized')
+
 const { ConversationContext, Message } = require('./conversations')
 
 const GROUP_TYPES = [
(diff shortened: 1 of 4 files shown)


src/renderer/components/ChatView.js
```diff
@@ -6,6 +6,8 @@ const SetupMessageDialog = require('./dialogs/SetupMessage')
 const Composer = require('./Composer')
 const { Overlay } = require('@blueprintjs/core')
 
+const { TrackVisibility } = require('react-on-screen')
```
ralphtheninja: cool!

karissa:
this is hard

Jikstra:
Any ideas why the list is behaving weirdly (jumps around/renders again and again)?

karissa:
@Jikstra something to do with estimating properly (and updating) the vertical height of each message..

I'm suspecting this is why signal renders ~20-30 messages at a time, then loads the next 20-30 once the person has scrolled all the way to the top, rather than estimating the scroll bar height and simulating virtual scroll. It's a bit easier to reason about and implement. We wouldn't need this 'infinitescroll' component then I think..

karissa:
@Jikstra regarding height of each message, see: https://github.com/deltachat/deltachat-desktop/pull/261/commits/ab8660c5cc65a98ffc0f5591e343d78fa7cfafa2

Jikstra:
> I'm suspecting this is why signal renders ~20-30 messages at a time, then loads the next 20-30 once the person has scrolled all the way to the top, rather than estimating the scroll bar height and simulating virtual scroll. It's a bit easier to reason about and implement. We wouldn't need this 'infinitescroll' component then I think..

But i think signal keeps all the messages rendered and just applies more elements to the ul. I don't know how well this scales if we want to rerender on every prepend.

karissa:
closing in favor of #299

Issue #195, opened 2018-10-20
Tweaks to Autocrypt Key Transfer


ralphtheninja:
When starting the key transfer, the same message is used when initiating the key transfer and also when typing in the code on the other device, e.g.:

Perhaps we should tweak this a bit

![screenshot from 2018-10-20 22-23-08](https://user-images.githubusercontent.com/308049/47260209-099a2200-d4b7-11e8-9611-e67b59d53be8.png)

![screenshot from 2018-10-20 22-22-52](https://user-images.githubusercontent.com/308049/47260211-0dc63f80-d4b7-11e8-93e6-bbf27b7565b4.png)

Also the `'Transfer Key'` button should be moved a bit.

ralphtheninja:
Agreed. And the second screen on the other device should have a very brief text. "Please enter bla bla".

Jikstra:
Yup, and instead of having some 'Transfer key' button, an "OK" button would be enough.

PR #262, opened 2018-11-01
Fix autocrypt ui

Jikstra:
screenshot_20181102_004710
screenshot_20181102_004750

### DIFF
--- a/src/renderer/components/dialogs/KeyTransfer.js
+++ b/src/renderer/components/dialogs/KeyTransfer.js
@@ -10,13 +10,27 @@ const {
 } = require('@blueprintjs/core')
 
 class KeyViewPanel extends React.Component {
+  formatAutocryptKey(autocryptKey) {
+    const splittedAutocryptKey = autocryptKey.split('-')
+    return splittedAutocryptKey.map((e, i) => {
+      const isLast = (i + 1) === splittedAutocryptKey.length
+      const isNThird = (i + 1) % 3 === 0
+
+      const addLineBreak = !isLast && isNThird
+      const addDash = !isLast
+
+      return e + (addDash ? ' - ' : '') + (addLineBreak ? "\n" : '')
+    }).join('')
+  }
+
   render () {
-    const { autocryptKey } = this.props
+    const formattedAutocryptKey = this.formatAutocryptKey(this.props.autocryptKey)
+
     const tx = window.translate
     return (
       <div>
         <p>{tx('showKeyTransferMessage')}</p>
-        <p>{autocryptKey}</p>
+        <pre>{formattedAutocryptKey}</pre>
         <div className={Classes.DIALOG_FOOTER}>
           <div className={Classes.DIALOG_FOOTER_ACTIONS}>
             <ButtonGroup>
--- a/src/renderer/components/dialogs/KeyTransfer.js
+++ b/src/renderer/components/dialogs/KeyTransfer.js
@@ -10,7 +10,7 @@ const {
 } = require('@blueprintjs/core')
 
 class KeyViewPanel extends React.Component {
-  formatAutocryptKey(autocryptKey) {
+  formatAutocryptKey (autocryptKey) {
     const splittedAutocryptKey = autocryptKey.split('-')
     return splittedAutocryptKey.map((e, i) => {
       const isLast = (i + 1) === splittedAutocryptKey.length
@@ -19,7 +19,7 @@ class KeyViewPanel extends React.Component {
       const addLineBreak = !isLast && isNThird
       const addDash = !isLast
 
-      return e + (addDash ? ' - ' : '') + (addLineBreak ? "\n" : '')
+      return e + (addDash ? ' - ' : '') + (addLineBreak ? '\n' : '')
     }).join('')
   }
 
--- a/src/renderer/components/dialogs/SetupMessage.js
+++ b/src/renderer/components/dialogs/SetupMessage.js
@@ -3,12 +3,10 @@ const { ipcRenderer } = require('electron')
 
 const {
   Spinner,
-  InputGroup,
   Classes,
   Button,
   ButtonGroup,
-  Dialog,
-  NumericInput
+  Dialog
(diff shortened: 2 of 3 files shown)


Jikstra:
Fixes https://github.com/deltachat/deltachat-desktop/issues/195

src/renderer/components/dialogs/KeyTransfer.js
```diff
@@ -10,13 +10,27 @@ const {
...
+    return splittedAutocryptKey.map((e, i) => {
+      const isLast = (i + 1) === splittedAutocryptKey.length
+      const isNThird = (i + 1) % 3 === 0
+
+      const addLineBreak = !isLast && isNThird
+      const addDash = !isLast
+
+      return e + (addDash ? ' - ' : '') + (addLineBreak ? '\n' : '')
+    }).join('')
+  }
```
ralphtheninja: Can we make this a normal function? No reason to be a class method since it doesn't use any class state (i.e. not using `this`).

src/renderer/components/dialogs/SetupMessage.js
```diff
@@ -14,33 +13,60 @@ class SetupMessagePanel extends React.Component {
...
+          id={i}
+          onChange={this.handleChangeKey}
+          min='0'
+          max='9999'
+          size='4'
+        />
+      )
+
+      if (i !== 8) {
+        inputs.push(<div class='seperator' />)
```
ralphtheninja: `separator`

src/renderer/components/dialogs/SetupMessage.js
```diff
@@ -14,33 +13,60 @@ class SetupMessagePanel extends React.Component {
...
+      }
+    }
+    return inputs
   }
 
   render () {
     const tx = window.translate
+    console.log(this.state)
```
ralphtheninja: `console.log`

static/main.css
```diff
@@ -177,3 +177,18 @@ ul li button {
...
 }
+
+.InputTransferKey {
+  display: grid;
+  grid-template-columns: 100px 10px 100px 10px 100px 10px;
+  grid-gap: 10px;
+  padding-bottom: 10px;
+}
+
+.InputTransferKey .seperator {
```
ralphtheninja: `.separator`

static/main.css
```diff
@@ -177,3 +177,18 @@ ul li button {
...
+  text-align: center;
+}
+
+.InputTransferKey .seperator::after {
```
ralphtheninja: Ditto.

src/renderer/components/dialogs/KeyTransfer.js
Jikstra: I don't care but i like to have it in a class just to have it grouped with where it belongs to. Also i think i already saw similiar stuff.

src/renderer/components/dialogs/SetupMessage.js
```diff
@@ -14,33 +13,60 @@ class SetupMessagePanel extends React.Component {
...
     this.state = {
-      key: ''
+      key: Array(9).fill('')
     }
-    this.handleChange = this.handleChange.bind(this)
+    this.handleChangeKey = this.handleChangeKey.bind(this)
   }
 
-  handleChange (event) {
+  handleChangeKey (event) {
```
karissa: can remove this TODO now :)

static/main.css
```diff
@@ -177,3 +177,18 @@ ul li button {
...
+
+.InputTransferKey {
+  display: grid;
+  grid-template-columns: 100px 10px 100px 10px 100px 10px;
+  grid-gap: 10px;
+  padding-bottom: 10px;
+}
+
+.InputTransferKey .seperator {
+  text-align: center;
```
karissa: is this `text-align: center` needed? I wonder if we could create then a class `.centered` or use `styled-components` or something..

static/main.css
```diff
@@ -177,3 +177,18 @@ ul li button {
...
+}
+
+.InputTransferKey .seperator::after {
+  content: '-';
```
karissa: instead can we use `-` in the `div`? I find this `content` property is only really useful in cases where you don't have access to generating the dom directly, but in this case we do

src/renderer/components/dialogs/SetupMessage.js
```diff
@@ -14,33 +13,60 @@ class SetupMessagePanel extends React.Component {
...
     // TODO: insert - automatically in between every 4 characters
     // TODO: lint the value for correct setup message format
-    this.setState({ key: event.target.value, message: false })
+    if (event.target.value < 0 || event.target.value > 9999) {
```
karissa: What happens when the user inputs a string like `asdf`? 
If the value isn't a number, we should return false also. So here:
```js
var value = Number(event.target.value)
if (isNaN(value) || value < 0 || value > 9999)  return false
let updatedkey = this.state.key
updatedkey[Number(event.target.id)] = value
this.setState({ key: updatedkey })
```

src/renderer/components/dialogs/SetupMessage.js
```diff
@@ -14,33 +13,60 @@ class SetupMessagePanel extends React.Component {
...
     return (<div>
       {this.props.message && <h3>{this.props.message}</h3>}
       <p>
         {tx('showKeyTransferMessage')}
       </p>
-      <InputGroup
-        onChange={this.handleChange}
-        value={this.state.key}
-      />
+      <div class='InputTransferKey'>
```
karissa: what if we use `styled-components` here instead?

ralphtheninja: Can we make a separate issue for `styled-components` and gather up points in the code where it's useful?

karissa: Sure. all of the wrappers that can be packaged as components

ralphtheninja:
@Jikstra rebase + meeeeerrrrrge! :)

Jikstra:
Hmm github doesn't really notice that the conflicts are already solved... can't merge it :(

Issue #245, opened 2018-10-31
Make links clickable


Jikstra:
Currently `<a href...>` aren't clickable. Maybe we want all a tags to just be clickable if this is possible? Or some extra component for this?

karissa:
`shell.openExternal(URL)`

It's better to do this by hand for security reasons, any URLs that we're opening in the user's browser should be hard coded

PR #274, opened 2018-11-04
Wrap <a href>

ralphtheninja:
Closes https://github.com/deltachat/deltachat-desktop/issues/245


Jikstra:
LGTM, we should start thinking about where to put those helper components

PR #275, opened 2018-11-04
Main window and about dialog should show same version

ralphtheninja:
screenshot from 2018-11-04 15-31-41


src/renderer/components/dialogs/About.js
```diff
@@ -29,7 +29,7 @@ class AboutDialog extends React.Component {
         onClose={onClose}
         canOutsideClickClose={false}>
         <div className={Classes.DIALOG_BODY}>
-          <p style={{ color: 'grey' }}>{'version ' + version}</p>
+          <p style={{ color: 'grey' }}>{`Version ${APP_VERSION}`}</p>
```
karissa: wicked

Issue #271, opened 2018-11-03
App crashes on login because of missing .mail_pw <3


jankass:
```
throw new Error('Missing .mail_pw')
            ^

Error: Missing .mail_pw
    at DeltaChat.configure (/opt/DeltaChat/resources/app.asar/node_modules/deltachat-node/index.js:106:13)
    at /opt/DeltaChat/resources/app.asar/es5/main/deltachat.js:84:10
    at binding.dcn_open.err (/opt/DeltaChat/resources/app.asar/node_modules/deltachat-node/index.js:478:9)
```

ralphtheninja:
Seems we broke the code activating/deactivating the login button?

ralphtheninja:
I can't repro on latest version, so I'm guessing this is an old version.

Jikstra:
@ralphtheninja she uses `1.7.3`

jankass:
This is because we pass a credentials object to deltachat-node with keys in camelcase (eg: mailPw) instead of snakecase (eg: mail_pw).

ralphtheninja:
> This is because we pass a credentials object to deltachat-node with keys in camelcase (eg: mailPw) instead of snakecase (eg: mail_pw).

Oh, haha, sorry. ANY password fails. The reason "I couldn't repro" was because I thought you had to input an empty password :)

ralphtheninja:
How can _any_ parameter work at all?

PR #284, opened 2018-11-05
Snake case keys during configure

ralphtheninja:
Closes https://github.com/deltachat/deltachat-desktop/issues/271

### DIFF
--- a/src/main/deltachat.js
+++ b/src/main/deltachat.js
@@ -460,24 +460,20 @@ function snakeCaseKeys (obj) {
 function translateSecurityToServerFlags ({ mailSecurity, sendSecurity }) {
   const flags = []
 
-  if (mailSecurity) {
-    if (mailSecurity === 'ssl') {
-      flags.push(C.DC_LP_IMAP_SOCKET_SSL)
-    } else if (mailSecurity === 'starttls') {
-      flags.push(C.DC_LP_IMAP_SOCKET_STARTTLS)
-    } else if (mailSecurity === 'plain') {
-      flags.push(C.DC_LP_SMTP_SOCKET_PLAIN)
-    }
+  if (mailSecurity === 'ssl') {
+    flags.push(C.DC_LP_IMAP_SOCKET_SSL)
+  } else if (mailSecurity === 'starttls') {
+    flags.push(C.DC_LP_IMAP_SOCKET_STARTTLS)
+  } else if (mailSecurity === 'plain') {
+    flags.push(C.DC_LP_SMTP_SOCKET_PLAIN)
   }
 
-  if (sendSecurity) {
-    if (mailSecurity === 'ssl') {
-      flags.push(C.DC_LP_SMTP_SOCKET_SSL)
-    } else if (mailSecurity === 'starttls') {
-      flags.push(C.DC_LP_SMTP_SOCKET_STARTTLS)
-    } else if (mailSecurity === 'plain') {
-      flags.push(C.DC_MAX_GET_INFO_LEN)
-    }
+  if (sendSecurity === 'ssl') {
+    flags.push(C.DC_LP_SMTP_SOCKET_SSL)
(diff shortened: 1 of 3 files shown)


ralphtheninja:
We should get this one out ASAP.

src/main/deltachat.js
```diff
@@ -444,11 +443,17 @@ class DeltaChatController {
...
+  return {
+    addr: obj.addr,
+    mail_user: obj.mailUser,
+    mail_pw: obj.mailPw,
+    mail_server: obj.mailServer,
+    mail_port: obj.mailPort,
+    send_user: obj.sendUser,
+    send_pw: obj.sendPw,
+    send_server: obj.sendServer,
+    send_port: obj.sendPort
```
ralphtheninja: So this code is incomplete since I left out `mailSecurity` and `sendSecurity`. There are no corresponding properties with that name in core.

ralphtheninja:
@Jikstra How would you like to handle `mailSecurity` and `sendSecurity`? I'm guessing they should have effect on the `server_flags` property?

src/main/deltachat.js
Jikstra: yeah the corresponding key would be server_flags (we are translating that already with translate... something method)

Jikstra: we should introduce a serverFlag variable and also do the snakecase stuff for it.

ralphtheninja:
> Ahh okay i see you're applying the snakeCase stuff in the backend. So my approach would be to also pass `mailSecurity` and `sendSecurity` to the backend and do the `translateSecurityToServerFlags()` stuff there. It basically allows us to use a simple string in the frontend and translate it to the corresponding DeltaChat CONSTANT.

Cool. `translateSecurityToServerFlags()` wasn't used in `Login.js`.

src/main/deltachat.js
```diff
@@ -452,8 +452,53 @@ function snakeCaseKeys (obj) {
     send_user: obj.sendUser,
     send_pw: obj.sendPw,
     send_server: obj.sendServer,
-    send_port: obj.sendPort
+    send_port: obj.sendPort,
+    server_flags: translateSecurityToServerFlags(obj)
   }
 }
 
+function translateSecurityToServerFlags ({ mailSecurity, sendSecurity }) {
```
ralphtheninja: Another argument for not making this into a class method is that `standard` would have caught this not being used.

src/main/deltachat.js
```diff
@@ -452,8 +452,53 @@ function snakeCaseKeys (obj) {
...
+      flags.push(C.DC_LP_IMAP_SOCKET_SSL)
+    } else if (mailSecurity === 'starttls') {
+      flags.push(C.DC_LP_IMAP_SOCKET_STARTTLS)
+    } else if (mailSecurity === 'plain') {
+      flags.push(C.DC_LP_SMTP_SOCKET_PLAIN)
+    }
+  }
+
+  if (sendSecurity) {
+    if (mailSecurity === 'ssl') {
```
ralphtheninja: The original code had a bug and it came with copy paste. Fixing.

src/main/deltachat.js
```diff
@@ -442,4 +442,59 @@ class DeltaChatController {
   }
 }
 
+function snakeCaseKeys (obj) {
```
karissa: seems like this could be done on the frontend -- `<input>` accepts underscore variables, so you would then only need to translate the security flags to sever flags.

but if you think it's fine, feel free to merge it and we could simplify it later.

karissa: seems like this could be done on the frontend -- `<input>` accepts underscore variables, so you would then only need to translate the security flags to sever flags.

but if you think it's fine, feel free to merge it and we could simplify it later.

Jikstra: input accepts underscore variables, but initiating them in the constructor throws standard errors if i understood @ralphtheninja correctly.

Issue #283, opened 2018-11-05
Main window content doesn't get rendered because of config folder not found


jankass:
Reproducable by deleting `~/.config/DeltaChat` and `~/.config/Delta.Chat`.
When the program is started for the second time, it works. 

```
$ npm start

> deltachat-desktop@1.7.3 start /home/jankass/Workspace/deltachat-desktop
> electron . --debug --disable-http-cache

(node:26368) Warning: N-API is an experimental feature and could change at any time.
A JavaScript error occurred in the main process
Uncaught Exception:
Error: ENOENT: no such file or directory, scandir '/home/jankass/.config/DeltaChat'
Missing config file: Creating new one
```

ralphtheninja:
Nice find!

ralphtheninja:
It's also a bit strange that there are _two_ configurations folders. There should be only _one_.

http://www.lovethispic.com/uploaded_images/208687-There-Can-Be-Only-One.png

ralphtheninja:
It seems the `Delta.Chat` folder is an remnant. I believe it's related to `productName` (in `package.json`) used to be `Delta.Chat` but is now `DeltaChat`. This folder can be safely removed.

PR #287, opened 2018-11-06
Ensure CONFIG_PATH always exists at start up

ralphtheninja:
Closes https://github.com/deltachat/deltachat-desktop/issues/283

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "linkify-it": "^2.0.3",
     "lodash": "^4.17.11",
     "lodash.merge": "^4.6.1",
+    "mkdirp": "^0.5.1",
     "moment": "^2.22.2",
     "react": "^16.4.2",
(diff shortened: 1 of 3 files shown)


karissa:
Would this run `mkdirp.sync` every time the config file is required? 
What if instead we ensure the config folder exists on startup in the `src/main/index.js` (so it only mkdirp once)?
And I wonder if this will cause issues when trying to `require` the config.js file from the renderer

ralphtheninja:
> Would this run `mkdirp.sync` every time the config file is required?

No. The code is only run the first time a module is required and the result is cached in `module.exports`.

> What if instead we ensure the config folder exists on startup in the `src/main/index.js` (so it only mkdirp once)?

Totally fine. I only put it in `config.js` because it is the closest position to knowing the value of `CONFIG_PATH`. Probably better to keep it out of the renderer process.

PR #290, opened 2018-11-07
the back button should go back to CreateChat screen on both CreateGroup and AddContact

ralphtheninja:
Found this gem while going through `git log` :)


ralphtheninja:
Adding @Jikstra as a reviewer since this is @karissa 's code.

LGTM! (I can't review since I made the PR)

karissa:
nice catch @ralphtheninja ! tested, works fine. cheers!

Issue #175, opened 2018-10-17
Show group image in chat list / chat view


ralphtheninja:
Once https://github.com/deltachat/deltachat-desktop/issues/174 is fixed

PR #291, opened 2018-11-08
Add group image to chat list

ralphtheninja:
Closes https://github.com/deltachat/deltachat-desktop/issues/175

WDYT?

screenshot from 2018-11-08 02-07-34

### DIFF
--- a/conversations/ts/components/ConversationListItem.tsx
+++ b/conversations/ts/components/ConversationListItem.tsx
@@ -35,10 +35,7 @@ export class ConversationListItem extends React.Component<Props> {
     const {
       avatarPath,
       color,
-      i18n,
-      name,
-      phoneNumber,
-      profileName,
+      name
     } = this.props;
 
     if (!avatarPath) {
@@ -57,14 +54,9 @@ export class ConversationListItem extends React.Component<Props> {
       );
     }
 
-    const title = `${name || phoneNumber}${
-      !name && profileName ? ` ~${profileName}` : ''
-    }`;
-
     return (
       <img
         className="module-conversation-list-item__avatar"
-        alt={i18n('contactAvatarAlt', [title])}
         src={avatarPath}
       />
     );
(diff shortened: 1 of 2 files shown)


conversations/ts/components/ConversationListItem.tsx
```diff
@@ -57,14 +54,9 @@ export class ConversationListItem extends React.Component<Props> {
...
     }
 
-    const title = `${name || phoneNumber}${
-      !name && profileName ? ` ~${profileName}` : ''
-    }`;
-
     return (
       <img
         className="module-conversation-list-item__avatar"
-        alt={i18n('contactAvatarAlt', [title])}
```
ralphtheninja: I removed this since I don't really see any use of it since at this point we know we have an image. The other lines went away because not used (TypeScript compiler complains).

ralphtheninja:
Now this got me thinking that maybe we don't need the image + text above the chat? On mobile it makes a lot of sense because there isn't any split view, but on desktop we know which chat that is selected on the left, which also contains pretty much the same information with the exception of the `X members` (which we could add to the chat list item?).

Jikstra:
> Now this got me thinking that maybe we don't need the image + text above the chat?

We don't need it, but I would leave it in if it's not done in a couple of minutes for now. Otherwise it looks good to me, even if i couldn't test real group images (all groups with images didn't get synced to desktop for me ^^)

r10s:
> Now this got me thinking that maybe we don't need the image + text above the chat?

i would leave it. image+name+subtitle. 

- the subtitle is not show in the chatlist
- the chatlist may scroll away, may be filtered through a search and whatever - it's good to have a clear caption then
- the title in the chatlist may be truncated, the chat has more place
- due to the search-field on the left (great, btw :) and to the menu on the right, the place is there anyway and we do not waste much screenspace

ralphtheninja:
> i would leave it. image+name+subtitle.

Good arguments. Lets leave it.

ralphtheninja:
@karissa Does this look ok?

conversations/ts/components/ConversationListItem.tsx
karissa: yeah makes sense to me. Although I wonder if we would want to add back in some sort of 'address' in place of phone number at some point.

karissa:
LGTM!

Issue #30, opened 2018-09-11
Create verified groups.


karissa:
This should be similar functionality to the android version. Next to the 'create chat' button on the main window, you should have a mini QR code button. This button should open a screen and when a mobile device scans that QR code, they will be entered into the verified group.

This also should create a 'create verified group' button in the 'create chat' view

PR #293, opened 2018-11-08
Verified groups

ralphtheninja:
Closes https://github.com/deltachat/deltachat-desktop/issues/30

Most things work, but missing some tweaks and feedback. What's missing is (as I see it):

* some form of graphical indication that a group is verified or not (the blue check mark on android)
* some text in the qr dialog?
* other stuff?

To avoid too much bikeshedding, maybe we can come to some agreement on things we should adjust here and other things that should change but that we can create issues for?

Also, @hpk42 mentioned that the workflow for verified groups is about to change on android so we might not want to spend too much time on this, since we would have to change it later anyway.

Some screens:

screenshot from 2018-11-08 05-40-45
screenshot from 2018-11-08 05-41-03
screenshot from 2018-11-08 05-41-19
screenshot from 2018-11-08 05-41-36


src/renderer/components/GroupBase.js
```diff
@@ -79,12 +84,34 @@ class GroupBase extends React.Component {
...
+      qrCode: ipcRenderer.sendSync('dispatchSync', 'getQrCode')
+    })
+  }
+
+  onShowQrInviteCode () {
+    const { chatId } = this.state
+    this.setState({
+      qrCode: ipcRenderer.sendSync('dispatchSync', 'getQrCode', chatId)
+    })
+  }
```
ralphtheninja: **Note** There are _two_ different QR codes. One is for adding verified contacts, which is done by `onShowQrVerifyCode()` and the other is for displaying an invite code for a verified group, which is done by `onShowQrInviteCode()`.

src/renderer/components/GroupBase.js
```diff
@@ -26,7 +29,9 @@ class GroupBase extends React.Component {
     this.state.group = this.state.group || {}
     this.state.name = this.state.name || ''
     this.state.image = this.state.image || ''
+    this.state.qrCode = ''
```
ralphtheninja: The qr code dialog is not shown if `qrCode` is empty.

src/renderer/components/GroupBase.js
```diff
@@ -128,6 +158,14 @@ class GroupBase extends React.Component {
       </div>
     )
   }
+
+  _getContacts () {
+    const { contacts } = this.props.deltachat
+    if (this.state.showVerifiedContacts) {
+      return contacts.filter(c => c.isVerified === true)
```
ralphtheninja: When creating or editing a verified group, you're only allowed to select between verified contacts.

src/renderer/components/dialogs/QrCode.js
```diff
@@ -0,0 +1,33 @@
...
+    const isOpen = !!qrCode
+    const tx = window.translate
+
+    return (
+      <Dialog
+        isOpen={isOpen}
+        title={tx('qrCode')}
+        icon='barcode'
+        onClose={onClose}
+        style={{ width: 296 }}
```
ralphtheninja: 20px * 2 + 256

r10s:
wow :)

hpk42:
wow indeed! :)

r10s:
@ralphtheninja yeah, the image is fine. if needed, a higher-res image is also available at https://github.com/deltachat/interface/tree/master/icons

btw. in the android-port, we show the icon _right_ of the group title. if it is easy to change, i would go for that. if it is complicated in any way, it's definitely fine for now :)

ralphtheninja:
> btw. in the android-port, we show the icon _right_ of the group title. if it is easy to change, i would go for that. if it is complicated in any way, it's definitely fine for now :)

@r10s I tried putting it there at first but it gets cut off if the window is too small (but maybe that is _ok_). Updating so you can see:

screenshot from 2018-11-08 18-33-25

r10s:
@ralphtheninja if i would have to decide between verified-icon always vislite on the left or sometimes visible on the right, i would vote for the left one :) 
btw. android-dev has exactly the same problem: https://github.com/deltachat/deltachat-android-ii/issues/120 :)

ralphtheninja:
> @ralphtheninja if i would have to decide between verified-icon always vislite on the left or sometimes visible on the right, i would vote for the left one :) at least for now i would not put much effort in this.
> btw. android-dev has exactly the same problem: [deltachat/deltachat-android-ii#120](https://github.com/deltachat/deltachat-android-ii/issues/120) :)

Might be able to float it right next to the timestamp. Will experiment a bit more.

r10s:
> Might be able to float it right next to the timestamp. Will experiment a bit more.

experimenting is always fine 🔨🔧 however, i think it belongs more to the group name than to the timestamp.

ralphtheninja:
I'd personally vote for keeping it to the left, but have no problem leaving it to the right either.

ralphtheninja:
I could use some css hacker. Hilfe @Jikstra ! It would be nice if we could still have it to the right of the group name, but without the ellipsis/overflow thingie cutting off the image (or whatever it is).

r10s:
@ralphtheninja just leave it to the left of the group name an go for sth. else :)
apart from understanding css-technology it's not that important currently. might be, that the verified-group will look differently at all in the near future :)

ralphtheninja:
Good to merge?

src/renderer/components/dialogs/QrCode.js
karissa: hehe nice. *magic*

karissa:
Looks great to me @ralphtheninja !

PR #299, opened 2018-11-15
load messages on demand using a paging model

### LINKED ISSUE #6: request things only as needed
this is not an real issue, but more a reminder about how deltachat-core (and the deltachat-node bindings) should be used to request things.

@ralphtheninja @karissa feel free to close it :)

stumbling upon the [loadChats()](https://github.com/deltachat/deltachat-cli/blob/750c4c2470ac74fd09cb61618baed582b79e4822/state.js#L196) in IRC i have noted that this call would be an expensive call as it get the message lists for every chat. while this may be okay in an CLI program, for a desktop app, this may be too slow.

the idea is more to get the **list of all chats** via [dc_get_chatlist()](https://deltachat.github.io/api/classdc__context__t.html#a709a7b5b9b606d85f21e988e89d99fef) - without any listflags, see below - and to implement a "virtual list" or so (the count of chats is known).

only for the **items that are in view** (the list may have several hundreds chats), the UI should call [dc_chatlist_get_summary()](https://deltachat.github.io/api/classdc__chatlist__t.html#a7e155ba468613547743bf383600914d1) then. this function provides all elements needed for painting the item.

on a click of such an item, the UI should **change to the chat view** and get all messages from this view via [dc_get_chat_msgs()](https://deltachat.github.io/api/classdc__context__t.html#a51353ff6b85fa9278d2ec476c3c95eda). again, a "virtual list" is created (the count of messages is known) and for each messages that is scrolled into view, [dc_get_msg()](https://deltachat.github.io/api/classdc__context__t.html#a4fd6b4565081c558fcd6ff827f22cb01) is called then.

why no listflags? without listflags, dc_get_chatlist() adds the deaddrop and the archive "link" automatically as needed. the UI can just render these items differently then.
although the deaddrop link is currently always the first entry and only present on new messages, there is the rough idea that it can be optionally always present and sorted into the list by date. rendering the deaddrop in the described way would not add extra work in the UI then.

karissa:
fixes #6 loading messages on demand using a page model. 
this is similar to how Signal handles it -- the user scrolls to the top, and then it loads new messages.

### DIFF
--- a/src/renderer/components/ChatView.js
+++ b/src/renderer/components/ChatView.js
@@ -10,6 +10,8 @@ const MutationObserver = window.MutationObserver
 
 const { ConversationContext, Message } = require('./conversations')
 
+const SCROLL_BUFFER = 70
+
 const GROUP_TYPES = [
   C.DC_CHAT_TYPE_GROUP,
   C.DC_CHAT_TYPE_VERIFIED_GROUP
@@ -27,6 +29,11 @@ class ChatView extends React.Component {
     this.focusInputMessage = this.focusInputMessage.bind(this)
     this.scrollToBottom = this.scrollToBottom.bind(this)
     this.conversationDiv = React.createRef()
+    this.lastId = this.props.chat.id
+  }
+
+  componentWillUnmount () {
+    if (this.observer) this.observer.disconnect()
   }
 
   writeMessage (text) {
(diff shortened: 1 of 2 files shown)


karissa:
this still isn't perfect if you scroll really really far back, but it's pretty darn good

src/renderer/components/ChatView.js
```diff
@@ -27,34 +29,55 @@ class ChatView extends React.Component {
     this.focusInputMessage = this.focusInputMessage.bind(this)
     this.scrollToBottom = this.scrollToBottom.bind(this)
     this.conversationDiv = React.createRef()
+    this.lastId = this.props.chat.id
```
ralphtheninja: Just wondering about `this.lastId`

ralphtheninja:
It's friggin super slick!

> this still starts to slow down again if you scroll really really far back, but it's pretty darn good

What is causing that slowdown? Maybe we can start unloading stuff from the DOM if we get way back?

ralphtheninja:
@karissa I'm cleaning up the other branches if that's ok, e.g. `omgwhat` and `load-messages-on-demand`

PR #308, opened 2018-11-16
add image attachments

### LINKED ISSUE #27: Add attachments to chat messages
There is a little paperclip icon in the chat window. A user should be able to click that button and a file chooser appears. The user should be able to select multiple files from this picker, and send them.

karissa:
fixes #27

### DIFF
--- a/src/renderer/components/Composer.js
+++ b/src/renderer/components/Composer.js
@@ -1,12 +1,14 @@
 const React = require('react')
 
 const { ControlGroup, Button, InputGroup } = require('@blueprintjs/core')
+const { remote } = require('electron')
 
 class Composer extends React.Component {
   constructor (props) {
     super(props)
     this.state = {
-      value: '',
+      filename: null,
+      text: '',
       error: false
     }
     this.minimumHeight = 48
@@ -18,7 +20,7 @@ class Composer extends React.Component {
 
   onKeyDown (e) {
     if (e.keyCode === 13 && e.shiftKey) {
-      this.setState({ value: this.state.value + '\n' })
+      this.setState({ text: this.state.text + '\n' })
       e.preventDefault()
       e.stopPropagation()
     } else if (e.keyCode === 13 && !e.shiftKey) {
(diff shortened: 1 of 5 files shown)


src/renderer/components/RenderMedia.js
```diff
@@ -0,0 +1,29 @@
+const React = require('react')
+
+class RenderMedia extends React.Component {
+  render () {
+    var { className, url, contentType } = this.props
+    if (!contentType) contentType = 'image/jpg'
+    let elm = <div />
+    // TODO: there must be a stable external library for figuring out the right
+    // html element to render
+    if (contentType) {
```
ralphtheninja: This if statement can be removed since it's set to `'image/jpg'` above (or _that_ if statement should be removed).

src/renderer/components/Composer.js
```diff
@@ -33,39 +35,50 @@ class Composer extends React.Component {
...
   }
 
-  addAttachment () {
-    console.log('adding attachment')
+  addFilename () {
+    var self = this
+    var opts = {
+      properties: ['openFile']
+    }
+    remote.dialog.showOpenDialog(opts, function (filenames) {
```
ralphtheninja: Use arrow function and remove `self`.

src/renderer/components/Composer.js
```diff
@@ -33,39 +35,50 @@ class Composer extends React.Component {
...
   handleChange (e) {
-    this.setState({ value: e.target.value, error: false })
+    this.setState({ text: e.target.text, error: false })
   }
 
-  addAttachment () {
-    console.log('adding attachment')
+  addFilename () {
+    var self = this
+    var opts = {
```
ralphtheninja: `const`

src/renderer/components/RenderMedia.js
```diff
@@ -0,0 +1,29 @@
+const React = require('react')
+
+class RenderMedia extends React.Component {
+  render () {
+    var { className, url, contentType } = this.props
```
ralphtheninja: Maybe `const { className, url, contentType = 'image/jpg' } = this.props` and remove if-statement below?

karissa: tried something different instead

karissa:
thanks @ralphtheninja !

src/renderer/components/RenderMedia.js
```diff
@@ -0,0 +1,27 @@
+const React = require('react')
+
+class RenderMedia extends React.Component {
+  render () {
+    const { className, url } = this.props
+    const contentType = this.props.contentType || 'image/jpg'
```
ralphtheninja: :+1:

PR #309, opened 2018-11-16
mark messages seen when fetched

### LINKED ISSUE #214: Mark messages as read if they are rendered and user actually sees them on screen

karissa:
fixes #214  (mostly)

right now it'll mark them as seen if the user fetches them, but we don't have a reliable way right now to mark them when they're on screen -- feeling like one way to make this more reliable would be to reduce the 'messages per page size' (right now it is 20)

### DIFF
--- a/src/main/deltachat.js
+++ b/src/main/deltachat.js
@@ -445,6 +445,7 @@ class DeltaChatController {
     const ids = selectedChat.messageIds
     var messageIds = ids.splice(ids.length - this._pages * PAGE_SIZE, ids.length)
     selectedChat.messages = messageIds.map((id) => this._messageIdToJson(id))
+    this._dc.markSeenMessages(messageIds)
     return selectedChat
   }
 


Jikstra:
We should really only mark them if they are on the screen and not recude the 'messages per page size' (which is already too low now in my opinion). Marking 20 messages or 10 messages (if we reduce them even more) as read if they aren't actually read isn't a good user experience. Can we somehow hook on the onScroll event of the virtual list component?

src/main/deltachat.js
```diff
@@ -445,6 +445,7 @@ class DeltaChatController {
     const ids = selectedChat.messageIds
     var messageIds = ids.splice(ids.length - this._pages * PAGE_SIZE, ids.length)
     selectedChat.messages = messageIds.map((id) => this._messageIdToJson(id))
+    this._dc.markSeenMessages(messageIds, messageIds.length)
```
ralphtheninja: The second parameter is not needed.

karissa:
hey @Jikstra the second approach doesn't sound like a bad idea. We'd need to fetch each message asyncronously I think so as not to block the thread. Could be combined with refactoring the RPC stuff with npmjs.org/pauls-electron-api

@ralphtheninja is the second parameter generated by deltachat-node?

karissa:
@Jikstra @ralphtheninja i still think this approach is better than *not* marking messages as read, although I think we ought to implement a setting to turn this off, too..

ralphtheninja:
> @ralphtheninja is the second parameter generated by deltachat-node?

Yes https://github.com/deltachat/deltachat-node/blob/master/index.js#L439

karissa:
thanks !

Jikstra:
> @Jikstra @ralphtheninja i still think this approach is better than not marking messages as read, although I think we ought to implement a setting to turn this off, too..

I agree to this, and this is maybe fine for the moment. In the long run we should change this. But in the long run we also should build the virtual list thing and stuff ^^

PR #311, opened 2018-11-16
Back button for Edit group should go back to split view

### LINKED ISSUE #292: Back button for Edit group should go back to split view
Scenario:

1. login to split view
1. select a group chat
1. select `Edit Group` in hamburger menu
1. hit back button in upper left corner
1. you end up at `Add contact` page

Expected result: you should end up at the split view

alfaslash:
I like the idea of this messenger!

fixes #292


ralphtheninja:
Thank you!

PR #316, opened 2018-11-16
Tweak group image and chat name margin

ralphtheninja:
Group name + no image, before:

screenshot from 2018-11-16 22-44-04

Group name + no image, after:

screenshot from 2018-11-16 22-47-34

Group name + image, before:

screenshot from 2018-11-16 22-49-56

Group name + image, after:

screenshot from 2018-11-16 22-49-15

### DIFF
--- a/static/main.css
+++ b/static/main.css
@@ -127,7 +127,12 @@ ul li button {
 }
 
 .Navbar img {
-    height: 50px;
+  height: 40px;
+  margin-left: 5px;
+}
+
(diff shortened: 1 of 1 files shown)


karissa:
is there a common component that exists outside of the img tag or navbar heading? perhaps we can add a div around the entire bar or add a class to it that can be referenced as 'chatview-header' or something, so that margin-left doesn't need to be in there twice.

karissa:
just add this to the list of css crap to refactor

PR #317, opened 2018-11-16
Upgrade node bindings

ralphtheninja:
**Note** This will pull in the single folder core.


karissa:
cool we should also notify the user

ralphtheninja:
> cool we should also notify the user

At least before doing the next release! Got any ideas?

PR #323, opened 2018-11-17
allow user to save a file attachment to disk

### LINKED ISSUE #310: save images/attachments to disk
to implement some very basic workflows, it should be possible to save received attachments by just clicking "save" or "export" beside the image or so.
i would expect the app showing a "save as" dialog then so that i can select the destination where the image/attachment should be saved to.

there is already a "download" button when the mouse is over an image, however, currently this is not working. but i think, making this functional would be sufficient for now.

karissa:
fixes #310

### DIFF
--- a/src/main/ipc.js
+++ b/src/main/ipc.js
@@ -71,6 +71,12 @@ function init (cwd) {
     })
   })
 
+  ipc.on('saveFile', (e, source, target) => {
+    fs.copyFile(source, target, function (err) {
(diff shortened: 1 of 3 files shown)


ralphtheninja:
Will merge after release.

ralphtheninja:
Red cross because I stopped build on Travis.

r10s:
great :+1:

PR #328, opened 2018-11-18
Add Jenkinsfile

ralphtheninja:


ralphtheninja:
First commit only to make it trigger. Currently using a plain `node:10.13-jessie` docker image. As we can see on https://ci.delta.chat/blue/organizations/jenkins/deltachat-desktop/detail/jenkins/1/pipeline it fails because meson is missing.

Next step: Create a docker image based on `node:10.13-jessie` with goodies already installed.

ralphtheninja:
Ok, _some_ progress. We can now do `npm install` and also build the application. Unit tests work but can't run integration tests. I think it's the same error we have on Travis if it's not set up correctly for `electron`. Will verify and fix this now.

ralphtheninja:
Time to figure out how to pass `GH_TOKEN` to the build container.

ralphtheninja:
Seems I can pass global environment variables to the build container, but that's not good enough, because anyone can make a PR and just print them out. Need something more secure.

ralphtheninja:
This article seems like the way to go https://support.cloudbees.com/hc/en-us/articles/203802500-Injecting-Secrets-into-Jenkins-Build-Jobs

ralphtheninja:
Ok. Merging this now so we get ci running on Jenkins for real.

package.json
```diff
@@ -9,7 +9,7 @@
     "test-integration": "node test/integration",
     "start": "electron . --debug --disable-http-cache",
     "pack": "npm run build && electron-builder --dir",
-    "dist": "npm run build && electron-builder --publish=onTagOrDraft",
+    "dist": "electron-builder --publish=onTagOrDraft",
```
ralphtheninja: `npm run build` is already made in the pipeline and no reason to do it again, since `npm run dist` is primarily a script which should be run in a ci environment.

karissa: sounds good

PR #329, opened 2018-11-19
use webpack instead of buble

karissa:

PR #330, opened 2018-11-19
set content security policy to script:self -- fixes #198

### LINKED ISSUE #198: Electron Security Warning
screenshot from 2018-10-21 04-43-28

Just something to check out.

--------------------------------------------------

Checklist (from https://electronjs.org/docs/tutorial/security):

- [x] Only load secure content

We're doing `.loadURL()` on `file://` and `main.html` is not loading any remote content.

- [x] Disable the Node.js integration in all renderers that display remote content

We can't do `{ webPreferences: { nodeIntegration: false } }` since we're using `require` directly in `main.html`, but we're not loading any remote content so we should be good?

- [x] Enable context isolation in all renderers that display remote content

- [x] Use `ses.setPermissionRequestHandler()` in all sessions that load remote content

- [x] Do not disable `webSecurity`

- [x] Define a `Content-Security-Policy` and use restrictive rules (i.e. `script-src 'self'`)

- [x] Override and disable `eval`, which allows strings to be executed as code. https://github.com/deltachat/deltachat-desktop/commit/1cf6750c0ce5ac454968c37c547c90c29b708f63

- [x] Do not set `allowRunningInsecureContent` to `true`

- [x] Do not enable experimental features

- [x] Do not use `enableBlinkFeatures`

- [x] `<webview>`: Do not use `allowpopups`

- [x] `<webview>`: Verify options and params

- [x] Disable or limit navigation https://github.com/deltachat/deltachat-desktop/commit/5026b502292d88f2de6217703be3cd24ab68269d

- [x] Disable or limit creation of new windows https://github.com/deltachat/deltachat-desktop/commit/303380568ba75816e055c26fabc058690b7a0567

karissa:
depends on moving to webpack merging #329

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -113,3 +113,12 @@ function sliceArgv (argv) {
     : config.IS_TEST ? 4
       : 2)
 }
+
+electron.session.defaultSession.webRequest.onHeadersReceived((details, fun) => {
+  fun({
+    responseHeaders: {
+      ...details.responseHeaders,
+      'Content-Security-Policy': ['default-src \'none\'']
+    }
+  })
+})
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -118,7 +118,7 @@ app.once('ready', () => {
   electron.session.defaultSession.webRequest.onHeadersReceived((details, fun) => {
     fun({
       responseHeaders: {
-          ...details.responseHeaders,
+        ...details.responseHeaders,
         'Content-Security-Policy': ['default-src \'none\'']
       }
     })
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -114,11 +114,13 @@ function sliceArgv (argv) {
       : 2)
 }
 
-electron.session.defaultSession.webRequest.onHeadersReceived((details, fun) => {
-  fun({
-    responseHeaders: {
-      ...details.responseHeaders,
-      'Content-Security-Policy': ['default-src \'none\'']
-    }
+app.once('ready', () => {
+  electron.session.defaultSession.webRequest.onHeadersReceived((details, fun) => {
+    fun({
+      responseHeaders: {
+          ...details.responseHeaders,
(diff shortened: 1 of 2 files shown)


ralphtheninja:
I get these errors in the console:

screenshot from 2018-11-19 23-18-46

karissa:
ok @ralphtheninja , I'm copying signal here:
https://github.com/signalapp/Signal-Desktop/blob/090397a59b14a6abb6e096b629a066f85e93eae8/background.html#L9

it works now, but not exactly sure if this is where we should go..

ralphtheninja:
> ok @ralphtheninja , I'm copying signal here:
> https://github.com/signalapp/Signal-Desktop/blob/090397a59b14a6abb6e096b629a066f85e93eae8/background.html#L9

Maybe we could ask Lilia or someone else?

Jikstra:
I think this should be fine but i think it would be better to set those headers on the electron side, just to be more clean. I don't really like to have it in the html markup, even if it's not a big thing or security issue as far as i know. 

Are we using inlined styles anywhere? Otherwise we should get rid of that. https://stackoverflow.com/questions/30653698/csp-style-src-unsafe-inline-is-it-worth-it

ralphtheninja:
@Jikstra Do you want to modify this before doing 2.0.0 or do you want to keep it around for after 2.0.0?

ralphtheninja:
Will rebase on latest master to start with.

ralphtheninja:
I get the following error when starting the app:

screenshot from 2018-11-21 17-32-31

ralphtheninja:
> it works now, but not exactly sure if this is where we should go..

@karissa To my limited knowledge on this it looks ok. If Signal does it this way, it's likely ok. But, this doesn't have to be perfected right now. Lets let it out there and we can have this reviewed/audited by someone that _does_ know this inside/out.

Issue #298, opened 2018-11-13
[Discussion] Modularize CSS/Preprocessor


Jikstra:
How do we want to achieve modularized css?
I think goals are:
- Keep react components and css somehow together
- Reusable css
- Maybe a preprocessor

Do we want to do styled components? What I don't like about them is that the js & css is in the same file (as far as i understand it).

Currently we have node-sass in the dependencies, but i don't like it because it depends on some node-gyp c/c++ binding stuff which often breaks. Otherwise i'm fine with it, but there are pure js preprocessors like less/stylus.

------------------------------------------------------------------------------

Things to clean up / refactor:

- [ ] Group image and group header, see https://github.com/deltachat/deltachat-desktop/pull/316#issuecomment-439557187

ralphtheninja:
@Jikstra Maybe you can suggest something you _do_ like and lets just go with that.

Jikstra:
I would go for less and put the `.less` files just next to the components. In the main.css (than less) i would just import all the component .less files. Clean and no big trouble, only need to transpile one file and we're done.

karissa:
Hey Jikstra, thanks for opening this.

Often these less files will be quite small, and creates a bit of overhead. Wondering what benefit we gain by doing this instead of using `styled-components` inside the js file.

The benefit of styled-components is that the css is right next to the component, but is still css so you can copy/paste it out if necessary.

Bundling into a single css file is nice for browsers (minifiaction etc) but on electron we don't really have the problem of a large js file needing to be downloaded before the site renders.

ralphtheninja:
Does `styled-components` mean we can still have classes etc? If so, where do they end up?

r10s:
i do not know much about node and all this, however, i general, i would suggest to keep things as simple as possible. even if this means eg. to add a color on several places to a css (a typical issue ..)

also because we otherwise would raise the skill requirement of $person who does the initial stying. eg. i know a bit of css - but i have no idea of frameworks or so.

when styled components or whatever (currently they are also "styled", or not?) are really needed, i would add it just then.

ralphtheninja:
> when styled components or whatever (currently they are also "styled", or not?) are really needed, we can add them probably later.

I think @karissa and @Jikstra are a bit worried that we do too much work with the current setup and end up with a gazillion css styles to change later which will be a nightmare. So we want to come with something better that's more easily maintained.

karissa:
[...]

PR #331, opened 2018-11-19
Start using styled-components

Jikstra:
We're now using styled components for half of the css rules in main.css. Fixes https://github.com/deltachat/deltachat-desktop/issues/298


ralphtheninja:
Borked on Jenkins because I cleaned up too much. Ignore.

ralphtheninja:
This needs rebasing.

PR #335, opened 2018-11-20
Tweak jenkins

ralphtheninja:
* renamed docker image
* only build `dist` on master branch (will speed up PRs as well, roughly ~2 mins per commit)
* add clean up step after build is done (regardless of state)


ralphtheninja:
Will squash.

PR #338, opened 2018-11-21
Simplify dialogs

karissa:
This is one step in the direction of refactoring the dialog code to all be at the top level.

ideally, dialogs would be handled in a DialogController, also required in `Home.js`

it also adds a contact detail dialog (for viewing safety numbers), but doesn't implement it quite yet


src/renderer/components/dialogs/ContactDetail.js
```diff
@@ -0,0 +1,33 @@
...
+
+const {
+  Classes,
+  Dialog
+} = require('@blueprintjs/core')
+
+class ContactDetailDialog extends React.Component {
+  render () {
+    var { contactId, onClose } = this.props
+    var isOpen = !!contactId
```
alfaslash: Why not to use `const`?

ralphtheninja: @alfaslash No reason :) @karissa been hacking hard and in the flow

src/renderer/components/Message.js
```diff
@@ -54,7 +54,7 @@ class RenderMessage extends React.Component {
 
     const contact = {
       onSendMessage: () => console.log('send a message to', fromId),
-      onClick: () => console.log('clicking contact', fromId)
+      onClick: () => { console.log('click contact') }
```
ralphtheninja: Is this unfinished? cc @Jikstra

ralphtheninja: Afaict, this `contact` is passed into `<Message>` but ends up in `renderEmbeddedContact()` (in `conversations/`) but that method doesn't seem to be used anywhere.

src/renderer/components/Message.js
```diff
@@ -54,7 +54,7 @@ class RenderMessage extends React.Component {
 
     const contact = {
       onSendMessage: () => console.log('send a message to', fromId),
```
ralphtheninja: This is never called either since `renderSendMessageButton()` renders nothing due to `contact.hasSignalAccount` is not set.

https://github.com/deltachat/deltachat-desktop/blob/4aceefe4306d611f12d8d92178ee4b334edc616d/conversations/ts/components/conversation/Message.tsx#L625-L627

ralphtheninja:
Lets merge this. There's already an issue on fixing contact details, which this PR isn't about.

PR #339, opened 2018-11-21
Deaddrop

karissa:
@r10s i wonder if this makes more sense -- we should be getting the *last* message received from the chat and displaying that contact as the 'dead drop', not the *first* message

### DIFF
--- a/src/main/deltachat.js
+++ b/src/main/deltachat.js
@@ -386,7 +386,7 @@ class DeltaChatController {
       return dc.getContact(id).toJson()
     })
     if (chatId === C.DC_CHAT_ID_DEADDROP) {
-      const msg = dc.getMessage(chat.messageIds[0])
+      const msg = dc.getMessage(chat.messageIds[chat.messageIds.length - 1])
       const fromId = msg && msg.getFromId()
 
       if (!fromId) {


src/main/deltachat.js
```diff
@@ -386,7 +386,7 @@ class DeltaChatController {
       return dc.getContact(id).toJson()
     })
     if (chatId === C.DC_CHAT_ID_DEADDROP) {
-      const msg = dc.getMessage(chat.messageIds[0])
+      const msg = dc.getMessage(chat.messageIds[chat.messageIds.length - 1])
```
r10s: i think this is wrong. 

what we need is the message-id from the message displayed in the corresponding chatlist entry.
we can get this message-id by [dc|dcn_chatlist_get_msg_id(chatlist, index)](https://c.delta.chat/classdc__chatlist__t.html#a0d089b4a45faa0e29d4350ef29049fd3) when we set the index to the index of the chatlist entry.

ralphtheninja: See https://github.com/deltachat/deltachat-node#listgetmessageidindex

karissa: I'm confused because the chat is the dead drop, but how do you know which message to get?

r10s: somewhere in the code, you render the deaddrop item in the chatlist using a get-summary-function, i assume you save some information together with the chatlist-view-item, eg. the index in the chatlist-object.

with this index you can then get all needed information from the chatlist using dc_chatlist_chat_msg_id() (or the correct wrapper-function).

r10s: dc_get_chat_msgs() resp. dc_chat_get_chat.messageIds[] for the deaddrop is only needed if you want to render _all_ pending contact requests in an overview (android does this with settings > contact requests), however, for the deaddrop in the chatlist, this is _not_ needed

Jikstra: @r10s we already have the complete messageIds array in our state, so no need to get the last messageId differently. Is the messageIds[-1] id any different from getting something from dc_chatlist_chat_get_msg_id() ?

r10s: well, not 100% sure, just compared the sql-commands ([here](https://github.com/deltachat/deltachat-core/blob/master/src/dc_chat.c#L1000) and [here](https://github.com/deltachat/deltachat-core/blob/master/src/dc_chatlist.c#L250)), probably, this will work most times, however, i would not guarantee this wrt blocking, fresh-state (the deaddrop-chat (not yet implemented in desktop) shows noticed and fresh messages)
also, there may always be the question what happens when the state changes between display of the chatlist and the click.

so, i would really recommend to use dc|dcn_chatlist_get_msg_id(chatlist, index)

> we already have the complete messageIds array in our state

what is the use of them? i think for rendering the chatlist items, this is not needed.

Jikstra: @r10s the dc_chatlist_chat_msg_id() gets which message of a chat. The first or the last?

PR #341, opened 2018-11-21
refactor menu

karissa:
:file_cabinet: :factory: :recycle: :relaxed: cleaning house

still didn't find a broom emoji though, what's with that?

### DIFF
--- a/src/renderer/components/Menu.js
+++ b/src/renderer/components/Menu.js
@@ -1,0 +1,151 @@
+const React = require('react')
+const C = require('deltachat-node/constants')
+const { ipcRenderer } = require('electron')
+const autobind = require('class-autobind').default
+
+const dialogs = require('./dialogs')
+const {
+  Menu,
+  MenuItem
+} = require('@blueprintjs/core')
+
+class Controller {
+  constructor (props) {
+    this.props = props
+    autobind(this)
+  }
+
+  onLeaveGroup () {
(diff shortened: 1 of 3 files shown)


src/renderer/components/Menu.js
```diff
@@ -0,0 +1,151 @@
...
+const dialogs = require('./dialogs')
+const {
+  Menu,
+  MenuItem
+} = require('@blueprintjs/core')
+
+class Controller {
+  constructor (props) {
+    this.props = props
+    autobind(this)
```
alfaslash: A good solution.
But how about use [React-autobind](https://github.com/cassiozen/React-autobind) instead [class-autobind](https://github.com/kodefox/class-autobind)?

`react-autobind` allow specify the only methods that should be auto-bound. And in other components  you can use auto-bound without create new class (ex. `Controller`).

ralphtheninja: @alfaslash @karissa is on vacation right now and we want to move as fast as possible atm, so things aren't perfect. If you think this important, make an issue and link to this PR _or_ make a PR :)

karissa: @alfaslash in this case we wanted all of the methods to be bound to the controller so this solution seemed to work well.

karissa: class-autobind uses decorators which requires more babel transforms

Issue #343, opened 2018-11-21
Content Security Policy: Refused to load the image 


Jikstra:
With latest master pictures aren't displayed at all and i'm getting a shi*load of those errors:

```
inputGroup.js:55 Refused to load the image 'file:///home/kerle/Workspace/deltachat-desktop-jikstra/images/ellipsis.svg' because it violates the following Content Security Policy directive: "default-src 'none'". Note that 'img-src' was not explicitly set, so 'default-src' is used as a fallback.

module.exports../node_modules/@blueprintjs/core/lib/esm/components/forms/inputGroup.js.InputGroup.updateInputWidth @ inputGroup.js:55
module.exports../node_modules/@blueprintjs/core/lib/esm/components/forms/inputGroup.js.InputGroup.componentDidUpdate @ inputGroup.js:44
commitLifeCycles @ react-dom.development.js:15976
commitAllLifeCycles @ react-dom.development.js:17262
callCallback @ react-dom.development.js:149
invokeGuardedCallbackDev @ react-dom.development.js:199
invokeGuardedCallback @ react-dom.development.js:256
commitRoot @ react-dom.development.js:17458
completeRoot @ react-dom.development.js:18912
performWorkOnRoot @ react-dom.development.js:18841
performWork @ react-dom.development.js:18749
performSyncWork @ react-dom.development.js:18723
requestWork @ react-dom.development.js:18592
scheduleWork @ react-dom.development.js:18401
enqueueSetState @ react-dom.development.js:12313
module.exports../node_modules/react/cjs/react.development.js.Component.setState @ react.development.js:372
update @ main.js:78
ipcRenderer.on @ main.js:49
emitTwo @ events.js:126
emit @ events.js:214
inputGroup.js:55 Refused to load the image 'file:///home/kerle/Workspace/deltachat-desktop-jikstra/images/padlock.svg' because it violates the following Content Security Policy directive: "default-src 'none'". Note that 'img-src' was not explicitly set, so 'default-src' is used as a fallback.

module.exports../node_modules/@blueprintjs/core/lib/esm/components/forms/inputGroup.js.InputGroup.updateInputWidth @ inputGroup.js:55
module.exports../node_modules/@blueprintjs/core/lib/esm/components/forms/inputGroup.js.InputGroup.componentDidUpdate @ inputGroup.js:44
commitLifeCycles @ react-dom.development.js:15976
commitAllLifeCycles @ react-dom.development.js:17262
[...]

Issue #344, opened 2018-11-21
When sending a message the scrollbar is disappearing or jumping up


Jikstra:
Not sure why, but with latest master this happens for me.

ralphtheninja:
Yes, same for me.

PR #345, opened 2018-11-21
add img-src: 'self' to Content-Security-Policy fixes https://github.com/deltachat/deltachat-desktop/issues/343

Jikstra:
fixes https://github.com/deltachat/deltachat-desktop/issues/343


Jikstra:
Nice, this magically also fixes https://github.com/deltachat/deltachat-desktop/issues/344

ralphtheninja:
Good catch!

Issue #325, opened 2018-11-17
URLs containing more than one dash aren't clickable


hpk42:
When URLs like https://deploy-preview-110--deltachat.netlify.com/en/2018-11-17-deltaxi are posted it doesn't result in clickable links.  Other URLs do -- some regex/heuristics issue that can be fixed easily?  Seems in Android it's consistently turned into a clickable link.

ralphtheninja:
It's due to https://github.com/markdown-it/linkify-it , just tested their demo as well https://markdown-it.github.io/linkify-it/ and it doesn't work there either.

ralphtheninja:
It's the double dash `--` that's causing it.

ralphtheninja:
It seems to be disabled by design:

https://github.com/markdown-it/linkify-it/blob/master/test/fixtures/not_links.txt#L12

ralphtheninja:
For follow up https://github.com/markdown-it/linkify-it/issues/73

ralphtheninja:
Either we patch `linkify-it` or replace it with another module, or write the damn regexp ourselves.

hpk42:
On Sat, Nov 17, 2018 at 04:50 -0800, Lars-Magnus Skog wrote:

> Either we patch `linkify-it` or replace it with another module, or write the damn regexp ourselves.

Might be a good idea to drop the dep -- there is no activity on this linkify-it repo it seems. 

FWIW here is a discussion of JS-regexes with URLS:
https://stackoverflow.com/questions/8188645/javascript-regex-to-match-a-url-in-a-field-of-text#8218223

karissa:
this doesn't seem high priority, it's a pretty rare link type `--` and you can otherwise copy and paste it into your browser. If you agree @ralphtheninja given the other stuff we have, i'd vote to table this for now

ralphtheninja:
> this doesn't seem high priority, it's a pretty rare link type `--` and you can otherwise copy and paste it into your browser. If you agree @ralphtheninja given the other stuff we have, i'd vote to table this for now

Agreed!

Jikstra:
Pending PR: https://github.com/markdown-it/linkify-it/pull/74

PR #348, opened 2018-11-22
Use linkify fork wich allows domains to contain multiple dashses

Jikstra:
Fixes https://github.com/deltachat/deltachat-desktop/issues/325

Thanks @ralphtheninja  :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,6 @@
     "@types/filesize": "^3.6.0",
     "@types/google-libphonenumber": "^7.4.17",
     "@types/jquery": "^3.3.19",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -74,7 +74,6 @@
     "@types/filesize": "^3.6.0",
     "@types/google-libphonenumber": "^7.4.17",
     "@types/jquery": "^3.3.19",
-    "@types/linkify-it": "^2.0.3",
```
ralphtheninja: :+1:

ralphtheninja: Hmm, maybe we can do the same trick for other things as well?

Issue #269, opened 2018-11-02
point users to android-dev


r10s:
as soon as the [single-folder milestone](https://github.com/deltachat/deltachat-core/milestone/2) gets merged into master and this version is used for deltachat-desktop, users that want to use a multi-client-setup with android and desktop should be pointed to use android-dev.

the reason for this as the current "stable" android-f-droid version moved messages from the INBOX to the DeltaChat folder that is no longer watched by single-folder-deltachat-desktop then.

android-dev is available at https://github.com/deltachat/deltachat-android-ii/releases and already uses the single-folder branch today.

ralphtheninja:
I was thinking we could have some sorts of "release banner" at the top with a X-button to dismiss?

r10s:
> I was thinking we could have some sorts of "release banner" at the top with a X-button to dismiss?

whatever is not too complicated ;)

r10s:
just merged the single-folder milestone to master, the next released version will contain this change then. (@ralphtheninja afaik node only uses tagged versions)

karissa:
@r10s yes, deltachat-node is in a beta state (`0.24.1`) so we can publish the single-folder milestone to `0.25` and it won't get used on desktop until desktop is ready and manually upgrades to `0.25.x`

r10s:
@karissa cool - i'd like to play around first here a bit with the "staged" commits and will release 0.25 then probably later this week. i think, it's not urgent in any way.

r10s:
0.25.0 is just released. 
so, if you use this new core, it would be great to point multi-device-users to android-dev

PR #349, opened 2018-11-22
Implement single folder warning banner

Jikstra:
@r10s  @hpk42 I'm not totally sure with the text, was just a quick write up. Also do we want to translate it?
single folder banner

Fixes https://github.com/deltachat/deltachat-desktop/issues/269

### DIFF
--- a/src/renderer/components/Login.js
+++ b/src/renderer/components/Login.js
@@ -10,8 +10,11 @@ const {
   InputGroup,
   FormGroup,
   Collapse,
-  Intent
+  Intent,
+  Callout
 } = require('@blueprintjs/core')
+
+const ClickableLink = require('./helpers/ClickableLink')
 
 const styled = require('styled-components').default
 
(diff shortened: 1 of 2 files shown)


src/renderer/components/Login.js
```diff
@@ -135,6 +138,9 @@ class Login extends React.Component {
           </NavbarGroup>
         </Navbar>
         <div className='window'>
+          <Callout intent='danger' title='Single folder incompatibility'>
+            To improve the user experience with Delta.Chat using multiple devices, we had to change the behaviour of Delta.Chat. Therefore, beginning with this release, we aren't compatible with the <b>old android client</b> which you can currently find in the f-droid store. Please use the <b>new development</b> version. You can find it <ClickableLink href='https://github.com/deltachat/deltachat-android-ii/releases'>here</ClickableLink>.
```
ralphtheninja: Just `Delta Chat` without the dot.

hpk42:
On Thu, Nov 22, 2018 at 11:02 -0800, Lars-Magnus Skog wrote:
> ralphtheninja requested changes on this pull request.
> 
> 
> 
> > @@ -135,6 +138,9 @@ class Login extends React.Component {
>            </NavbarGroup>
>          </Navbar>
>          <div className='window'>
> +          <Callout intent='danger' title='Single folder incompatibility'>
> +            To improve the user experience with Delta.Chat using multiple devices, we had to change the behaviour of Delta.Chat. Therefore, beginning with this release, we aren't compatible with the <b>old android client</b> which you can currently find in the f-droid store. Please use the <b>new development</b> version. You can find it <ClickableLink href='https://github.com/deltachat/deltachat-android-ii/releases'>here</ClickableLink>.
> 
> Just `Delta Chat` without the dot.


maybe better than "... we had to change the behaviour of Delta.Chat:" is something like
"... we experimentally changed the behaviour of DeltaChat:"

PR #356, opened 2018-11-25
Try preminor release of node bindings

ralphtheninja:
Reduce install time from like 45 to 25 seconds.


ralphtheninja:
Lets use this a little while and see how it works in the wild.

Jikstra:
Nice, prebuilds :) :+1:

PR #358, opened 2018-11-26
Try a cleaner approach using xvfb-run

ralphtheninja:


jenkins/Jenkinsfile.linux
```diff
@@ -24,10 +24,7 @@ pipeline {
     }
     stage('Integration Tests') {
       steps {
-        sh '/usr/bin/xvfb-chromium &'
-        sh 'sleep 5'
-        sh 'export DISPLAY=:99'
-        sh 'npm run test-integration'
+        sh 'xvfb-run npm run test-integration'
```
ralphtheninja: `xvfb-run` runs `Xvfb` server, sets up `DISPLAY` environment flag and when the command after (`npm run test-integration`) ends, it shuts down the server. Pretty cool!

ralphtheninja: Also, we don't have to rely on potentially flakey sleep stuff.

ralphtheninja:
Need to get this into master now since the docker image has changed.

PR #359, opened 2018-11-26
Use version unchanged

ralphtheninja:
This should make all builds on master as fast as PR builds. Unless the version has changed :)


karissa:
cool!

PR #365, opened 2018-11-26
properly set viewType when sending an attachment image

karissa:


karissa:
@Jikstra the `convertContentType` function doesn't handle all cases -- for example, a `pdf` file should have a display mimetype of pdf but right now it is still `application/octet-stream`, which is also a downloadable file, just doesn't say PDF in the viewport.

src/renderer/components/Message.js
```diff
@@ -147,9 +147,20 @@ function convertMessageStatus (s) {
...
 }
 
-function convertContentType (filemime) {
+function convertContentType (message) {
+  var filemime = message.filemime
   if (!filemime) return 'image/jpg'
-  if (filemime === 'application/octet-stream') return 'audio/ogg'
+  if (filemime === 'application/octet-stream') {
+    var s = message.msg.viewType
+    switch (s) {
```
Jikstra: Only a very minor thing, why doing the reference to `s`? I would just do an `switch (message.msg.viewType), i think it's clean enough.

Jikstra:
@karissa makes sense. Let's open an issue on that one so we don't forget it? Or is it obvious?

Issue #273, opened 2018-11-03
Group image not updated if someone else changes it


ralphtheninja:
How to reproduce:

1. create a group with someone
1. change group image
1. let the other user change image
1. image not updated

Below I have a chat with "Dave". I first set the `level.png` group image (and Dave's android client updates the group image accordingly). However, when Dave _changes_ the group image on his side (to the `delta.png` image) the desktop client doesn't update (it renders a normal message with an attached image) and keeps the old `level.png`.

![screenshot from 2018-11-03 15-52-53](https://user-images.githubusercontent.com/308049/47953654-f3916480-df80-11e8-9ec5-2a90eed537a8.png)

Shouldn't this be automatically handled by core? Or do you need to do some magic on the specific client to make this work?

cc @r10s 

----------------------------------------------------

TODO:

- [x] Handle this in `desktop` (or wait for core to do that)
- [x] Update message style (centered `$Person changed Group Image) (fixed in https://github.com/deltachat/deltachat-desktop/pull/366)

r10s:
> Shouldn't this be automatically handled by core? Or do you need to do some magic on the specific client to make this work?

this should be handled automatically by the core.

the core parses theses messages and you get the new image from [dc_chat_get_profile_image()](https://c.delta.chat/classdc__chat__t.html#ad96106c6b88f7705d8ac263133a61c43). ui-wise it should not make a difference if the group image was changed by the local or by a remote user - the event [DC_EVENT_CHAT_MODIFIED](https://c.delta.chat/group__DC__EVENT.html#ga27512e465c573fcf295014f8e0075adf) is emitted in both cases.

however, please note that the dc_chat_get_profile_image() will just point to another file; the files as such won't be changed and the old file will also still be present (as the old message is probably also still there).

to get a better chat overview, i would suggest to check if a given message is an "info" message and render them differently - eg. only text, centered. to check this state, you can use [dc_msg_is_info()](https://c.delta.chat/classdc__msg__t.html#a7272ddc7b2da2605463fa3ce0a00edc7).

ralphtheninja:
> this should be handled automatically by the core.

My point is that core does _not_ handle this. The profile image for the chat isn't updated. I.e. if I call `dc_chat_get_profile_image()` again, it's the same file as the previous one.

r10s:
i will have a closer look at this and check if it really works as expected on cmdline/android-dev.

r10s:
[...]

PR #366, opened 2018-11-26
display info text differently than regular messages

karissa:
screenshot from 2018-11-26 14-51-47
addresses this comment by @r10s https://github.com/deltachat/deltachat-desktop/issues/273#issuecomment-441241757

### DIFF
--- a/src/renderer/components/Message.js
+++ b/src/renderer/components/Message.js
@@ -79,8 +79,12 @@ class RenderMessage extends React.Component {
     }
 
     if (msg.attachment && !msg.isSetupmessage) props.attachment = msg.attachment
+    let body
+    if (message.isInfo) {
+      body = <div className='infoText'>{msg.text}</div>
+    } else body = <Message {...props} />
 
-    return (<div ref={this.el} className='MessageWrapper'><Message {...props} /></div>)
+    return (<div ref={this.el} className='MessageWrapper'>{body}</div>)
   }
 }
 
--- a/src/renderer/components/Message.js
+++ b/src/renderer/components/Message.js
@@ -1,6 +1,7 @@
 const React = require('react')
 const path = require('path')
 const C = require('deltachat-node/constants')
+const styled = require('styled-components').default
 const { Message } = require('./conversations')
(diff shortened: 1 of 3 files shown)


src/renderer/components/Message.js
```diff
@@ -9,16 +10,53 @@ const GROUP_TYPES = [
...
+    background-color: #757575;
+    color: #ffffff;
+    font-size: 25px;
+    line-height: 36px;
+  }
+
+  .module-message__author-default-avatar {
+    position: static;
+    margin-right: 8px;
+  }
```
Jikstra: :+1:  for switching more stuff to styled components

src/renderer/components/Message.js
```diff
@@ -79,8 +118,9 @@ class RenderMessage extends React.Component {
     }
 
     if (msg.attachment && !msg.isSetupmessage) props.attachment = msg.attachment
+    if (message.isInfo) return <InfoMessage>{msg.text}</InfoMessage>
```
Jikstra: I would like to have the distinction between the different message rendering types at one place. Currently we're doing it in render() and RenderMessage.render() method. If it's not too much trouble i would prefer to have it at the same place. Otherwise looks good to me :)

static/main.css
```diff
@@ -72,29 +72,6 @@ ul li button {
...
-.module-message__author-default-avatar {
-  position: static;
-  margin-right: 8px;
-}
-
-.MessageWrapper .module-message__metadata {
-  margin-top: 10px;
-  margin-bottom: -7px;
-}
-
```
Jikstra: :+1:  shrink that main.css ^^

src/renderer/components/Message.js
karissa: I think this would need some refactoring and/or rethinking UI, because the whole setup message now is expected to be clickable (outside of the message), otherwise they have to specifically click a box within the text or the text itself to open the setup message.

Jikstra: Okay then let's get this merged

PR #370, opened 2018-11-28
Update linkify-it dependency to 2.1.0 which should include the fix fo…

Jikstra:
…r #325


ralphtheninja:
Ok, so we pushed the fix completely to that repo. Cool.

PR #371, opened 2018-11-29
Backups

### LINKED ISSUE #267: Import/Export backups

karissa:
fixes #267 

Still needs:
- [x] UI work in the settings
- [x] translation for the export confirmation dialog
- [x] listen to events for progress bar and give user feedback when finished for both import and export

### DIFF
--- a/src/renderer/components/dialogs/index.js
+++ b/src/renderer/components/dialogs/index.js
@@ -7,6 +7,8 @@ const ContactDetail = require('./ContactDetail')
 const DeadDrop = require('./DeadDrop')
 const KeyTransfer = require('./KeyTransfer')
 const QrCode = require('./QrCode')
+const ImexProgress = require('./ImexProgress')
+const About = require('./About')
 
 module.exports = {
   confirmation,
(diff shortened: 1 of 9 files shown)


ralphtheninja:
Is it a work in progress?

Jikstra:
LGTM how do we want to tackle the nicer transition between settings screens? Currently it seems like it just upons another dialog on top of the settings one.

karissa:
@Jikstra good question. It works fine, but yeah having some sort of transition and putting them in the same dialog might be nice. Would take some refactoring

karissa:
@ralphtheninja yes it is a WIP, those 3 things still need to be done

src/renderer/components/dialogs/index.js
```diff
@@ -19,14 +19,16 @@ module.exports = {
...
+  if (!opts) opts = {}
   const tx = window.translate
-  const opts = {
+  var defaultOpts = {
     type: 'question',
     message: message,
     buttons: [tx('dialogs.confirmation.no'), tx('dialogs.confirmation.yes')]
   }
-  remote.dialog.showMessageBox(opts, response => {
+  remote.dialog.showMessageBox(Object.assign(defaultOpts, opts, {}), response => {
```
ralphtheninja: Why the last `{}`?

ralphtheninja: Did you mean to do `opts || {}`?

karissa: creates a copy I think although perhaps not necessary? does Object.assign mutate?

ralphtheninja: Yes it mutates first argument.

karissa:
ready for review @Jikstra @ralphtheninja

PR #376, opened 2018-12-04
unselectable helper component

Simon-Laux:
To make preventing stuff from being selected easier.
It doesn't feel professional if you can select stuff you normally couldn't select in an desktop app.


src/renderer/components/dialogs/About.js
```diff
@@ -30,9 +31,11 @@ class AboutDialog extends React.Component {
         canOutsideClickClose={false}>
         <div className={Classes.DIALOG_BODY}>
           <p style={{ color: 'grey' }}>{`Version ${APP_VERSION}`}</p>
-          <p>Official Delta Chat Desktop app.</p>
```
ralphtheninja: Maybe wrap this `<p>` as well?

Jikstra:
Looks good to me :+1:

Issue #394, opened 2018-12-07
Paragraphs, the possibility of a new line in the desktop version


EvgenyKabanov:
Debian GNU\Linux amd64
Delta Chat Desktop Version: 2.1.0
In the desktop client (2.0.1) you really want to be able to break the text into paragraphs, start new paragraphs with a new line. This will allow you to abandon the mail client without losing the quality of correspondence. Ideally, you want to connect an external editor, but this is how it will turn out. But without paragraphs it's sad.

PR #399, opened 2018-12-08
Feature: multiline composer, fixes https://github.com/deltachat/deltachat-desktop/issues/394

Jikstra:
Allows multiline messages, still needs ui/ux improvement, but we can maybe do this later.

EDIT: ui improvments are done, to make the input bigger in height while typing, we need to do some ugly hacks or refactor the whole chat/view/navbar/composer skeleton (which we should definetly do)


r10s:
\o/

Jikstra:
@ralphtheninja let's get this merged?

PR #411, opened 2018-12-11
Notifications, fixes #246

### LINKED ISSUE #246: notify on new message
the desktop app, while running, should notify the user when it receives a new message.
https://electronjs.org/docs/api/notification

karissa:
This implements notifications when you receive a new message.

* It doesn't notify when the browser window is focused.
* It only notifies you for one message at a time. When the notification closes, it then allows for more notifications to go through. I wanted to implement this in a way that didn't spam users when you first load the app and there are potentially tons of fresh messages. Open to feedback.
* The icon is really tiny for the notifications on my linux build! Is there one of proper size somewhere? 
* Users should be able to choose in the settings whether or not to allow notifications, or allow showing message bodies in the notifications, but right now everything is set to `true`. Will open another PR to address that in the settings.


src/main/notifications.js
```diff
@@ -0,0 +1,40 @@
...
+
+  function getMsgBody (msgId) {
+    const tx = app.translate
+    if (!settings.showNotificationContent) return tx('newMessageNotification')
+    var json = dc.messageIdToJson(msgId)
+    var summary = json.msg.summary
+    return `${summary.text1 || json.contact.displayName}: ${summary.text2}`
+  }
+
+  if (Notification.isSupported()) {
```
ralphtheninja: Put this as a guard at the top instead.

```js
if (!Notification.isSupported()) return
```

ralphtheninja:
@karissa I rebased it on master to solve conflict in `_locales/en.json`.

karissa:
great thanks!

PR #412, opened 2018-12-11
Forwarding messages

karissa:
This implements forwarding messages when clicking the context menu in a message and hitting 'forward message.'

A dialog window pops up with the list of contacts. When a contact is clicked, the dialog closes and the message is forwarded to that person.

This also implements a Search input box when selecting a contact, for filtering the contact list based on a search term

* :relaxed:   refactor: Contacts are retrieved using a synchronous model so we don't have to pass contacts and re-render everything every time someone types in the contact search box
* :tada: bonus!: This search function is abstracted out into it's own `ContactList` component and reused for searching for contacts when creating a new group and creating a new chat 
* :see_no_evil:  hack!: The contact list was getting multiple contacts with the same id when using get_contacts with a search query. This was causing many contacts with the same name to show up when searching for a contact. I [implemented a hack on our side to fix this](https://github.com/deltachat/deltachat-desktop/pull/412/commits/b150fd103d92f9d99903f1190557dedf89bfe1f0) but ideally this would be fixed by making core return only distinct queries. What do you think @r10s ?

### DIFF
--- a/_locales/en.json
+++ b/_locales/en.json
@@ -823,5 +823,9 @@
   "newMessageNotification": {
     "message": "New Message!",
     "description": "Shown as a placeholder when notification details are turned off"
+  },
+  "forwardMessageTitle": {
+    "message": "Forward Message",
+    "description": "Title in the dialog for choosing a contact to forward to"
   }
 }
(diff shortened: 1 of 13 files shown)


karissa:
@ralphtheninja @Jikstra can you review?

ralphtheninja:
@karissa I'm on the move atm. Will review/test in a bit.

ralphtheninja:
Nice work!

PR #413, opened 2018-12-11
Add imap folder configuration

### LINKED ISSUE #301: Add imapFolder configuration
Once https://github.com/deltachat/deltachat-node/issues/181 is done and released

ralphtheninja:
Closes #301

### DIFF
--- a/_locales/en.json
+++ b/_locales/en.json
@@ -213,7 +213,7 @@
     },
     "login.enterPassword": {
       "message": "Enter your password...",
-      "description": "Text telling user to enter his password"      
+      "description": "Text telling user to enter his password"
     },
     "login.mailPw": {
         "message": "Password",
@@ -242,6 +242,14 @@
     "login.mailPort": {
         "message": "IMAP-Port",
         "description": "Label above input field for imap server port"
+    },
+    "login.imapFolder": {
+        "message": "IMAP-Folder",
+        "description": "Label above input field for imap folder"
+    },
+    "default": {
+        "message": "default",
+        "description": "Label to the right of imap folder input describing the default value for imap folder"
     },
     "login.mailSecurity": {
         "message": "Security",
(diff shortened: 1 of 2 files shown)


hpk42:
hum, not sure if we actually make the folder configurable for the upcoming release.  Maybe we just make "watch INBOX" and "watch DeltaChat" two boolean options in advanced settings (and both would be on by default). @r10s what's your current take?

karissa:
yeah let's wait on this until we have some UX input from @r10s , also curious what xenia and vadym think

r10s:
yeah, i would also wait a bit until the shared-account-ships comes to calmer water :)

in fact, imap_folder option is currently deprecated (it will still work, but i do not think the option will stay under this name)

webratte:
Since you have in Android Client (dev-version)

- Watch inbox folder

- Watch sent folder

- Watch deltachat folder

you also put this options in the desktop client.
I use only the deltachat folder (mails to a DC aliasadress will moved there via server side filter) to watch.
The current desktop client (v. 0.97.0-preview) seems only watch the inbox. So I can't keep my messages in sync between Android and Desktop.

webratte:
Hi,
will you not implement this?
So the desktop client will be useless for users like me who use a Alias adress for DC and only the DC folder :-(

See also my last post.

hpk42:
i thikn this was closed by mistake -- the android "imap options" should also be in desktop.

hpk42:
oh, but this PR is about a specific variant of imap folder options (setting the name of the folder i think).  so actually it is fine to close this -- the PR for desktop to implement imap options needs to look differently.

webratte:
> the android "imap options" should also be in desktop.

OK, sounds good to me and I think it's amust have 👍

hpk42:
see also https://github.com/deltachat/deltachat-desktop/issues/508

Issue #415, opened 2018-12-11
New Feature: Drag and Drop Files


Simon-Laux:
Drag and Drop in both directions:

- [x] Drag a file out of the window
- [x] Drag a file in order to send it

PR #419, opened 2018-12-11
Feature drag file out

Simon-Laux:
#415 
Basic Drag out files (only the ones that get the file icon).
To drag the file out, start draging on the file icon and drag it to the place you want the file to be saved to.


karissa:
Works for me.

Would be nice to be able to push these changes upstream also to Signal (they also use electron).. not sure if that would be possible or desirable, but in any case, would be more difficult if we're going to incorporate 'conversations' as js directly

karissa:
thanks @Simon-Laux ! great work

PR #422, opened 2018-12-12
Android translations

karissa:
Now the translation strings need to be updated across the desktop app, including plurals which require a new signature now: `tx('string to translate', 1, {quantity: 'one'})`

Missing translation strings:

**Note** Do not copy paste another generated list over the current list since it has been manually modified with strings not found by the script.

- [x] addContact
- [x] autocryptCorrect
- [x] autocryptEnterMessage
- [x] autocryptIncorrect
- [x] autocryptKeyTransfer
- [x] backupConfirmationMessage
- [x] chatWith
- [x] contactDetailTitle
- [x] contactRequestTitle
- [x] createChatError
- [ ] DC_STR_E2E_AVAILABLE
- [ ] DC_STR_ENCR_NONE
- [ ] DC_STR_ENCR_TRANSP
- [ ] DC_STR_STARREDMSGS
- [x] deleteMessage
- [x] deleteThisMessage
- [x] dialogs.blockContact
- [x] dialogs.confirmation.no
- [x] dialogs.confirmation.yes
- [x] dialogs.deleteChat
- [x] dialogs.leaveGroup
- [x] dialogs.unblockContact
- [x] editGroup
- [x] ~~editVerifiedGroup~~ (replaced with corresponding editGroup)
- [x] emailValidationFailed
- [x] ~~emojiAlt~~ (removed :fire:)
- [x] encryptionInfoMenu
- [x] encryptionInfoTitle
- [x] exportBackup
- [x] forgetLoginConfirmation
- [x] forwardMessage
- [x] forwardMessageTitle
- [x] groupName
- [x] imexProgressTitle
- [x] importBackup
- [x] initiateKeyTransfer
- [x] initiateKeyTransferTitle
- [x] language_be
- [x] login.email
- [x] login.required
- [x] logout
- [x] me (when watching encryption info)
- [x] messageDetailTitle
- [x] menu.view.floatontop
- [x] moreInfo
- [x] name
- [x] newChat
- [x] newGroup
- [x] newMessageFrom
- [x] newMessageNotification
- [x] newVerifiedGroup
- [x] noChats
- [x] noArchivedChats
- [x] qrCode
- [x] remove
- [x] replyToMessage
- [x] save
- [x] ~~searchAriaLabel~~ (removed :fire:)
- [x] selectGroupImage
- [x] showKeyTransferMessage
- [x] showQrInviteCode
- [x] showQrVerifyCode
- [x] ~~this.state.buttonLabel~~ (not a string, it's a variable)
- [x] ~~this.state.heading~~ (not a string, it's a variable)
- [x] timestampFormat_M
- [x] transferKey
- [x] unblockContacts
- [x] unblockContacts.noneBlocked
- [x] utocrypt_prefer_e2ee (this is due to a typo bug, it should be autocrypt_prefer_e2ee)
- [x] verified
- [x] writeMessage
- [x] ~~writeMessageAriaLabel~~ (removed :boom:)
- [x] ~~newIdentity~~ (false positive, it's from a markdown file)


Jikstra:
So I looked into this, there are some problems:
- There are strings which we need in desktop which just aren't there in android, not sure how we get them into transifex (not signed up there yet) 
- The desktop ui/ux is different from android, so to use the same translations we maybe need to slightly adjust things more to the android side of things

This is of course all doable, but maybe more work then just replacing strings.

ralphtheninja:
@Jikstra Make a PR to https://github.com/deltachat/deltachat-android-ii/blob/master/res/values/strings.xml and then @r10s can upload it to Transifex.

r10s:
just in case i am not available for pushing strings to transifex, this could also be done by everyone with transifex-admin access by a simple `tx push -s` from the android repo (some care is needed with the push command to not overwrite new translations, however).

in general, i would try to formulate the strings for the same actions in a way that they fit for both, android and desktop.

eg. for the setup message, i've replaced 'tap on the message' by 'tap or click on the message', so that no two strings are needed. also this makes sense as android could also use a mouse (anbox...) and desktop could also have a touchscreen.

when this approach is not possible, i would suggest to posfix the strings with `_desktop` or even `_mac`.

but i would not add this prefix for strings that are just not used currently in android - eg. the account settings strings currently used in desktop may be needed for android just a but later.

r10s:
maybe we should add a howto-localize readme to the interface repo. i'm currently not up for that but i will add an issue and link this page from there.

src/localize.js
```diff
@@ -11,19 +12,30 @@ const log = require('./main/log')
...
+      let c = 0
+      return val.replace(/(?:%\d\$[\w\d])|(?:%[\w\d])/g, () => {
+        if (typeof substitutions[c] === 'undefined') {
+          console.error(
+            `translation: Missing ${c}th argument for key '${key}'`
+          )
+        }
+        let replacement = substitutions[c].toString()
+        c++
+        return replacement
```
ralphtheninja:
```js
return substitutions[c++].toString()
```

Jikstra:
Missing translation strings
- [ ] autocryptCorrect
- [ ] autocryptEnterMessage
- [ ] autocryptIncorrect
- [ ] autocryptKeyTransfer
- [ ] backupConfirmationMessage
- [ ] chatWith
- [ ] contactDetailTitle
- [ ] contactRequestTitle
- [ ] createChatError
- [ ] DC_STR_E2E_AVAILABLE
- [ ] DC_STR_ENCR_NONE
- [ ] DC_STR_ENCR_TRANSP
- [ ] DC_STR_STARREDMSGS
- [ ] dialogs.blockContact
- [ ] dialogs.confirmation.yes
- [ ] dialogs.deleteChat
- [ ] dialogs.leaveGroup
- [ ] dialogs.unblockContact
- [ ] emailValidationFailed
- [ ] encryptionInfoMenu
- [ ] encryptionInfoTitle
- [ ] forgetLoginConfirmation
- [ ] forwardMessageTitle
jq: error: id/0 is not defined at <top-level>, line 1:
.GROUPS[id].key
jq: 1 compile error
- [ ] groupName
- [ ] imexProgressTitle
- [ ] importBackup
- [ ] initiateKeyTransfer
- [ ] initiateKeyTransferTitle
- [ ] login.required
- [ ] logout
- [ ] messageDetailTitle
- [ ] name
- [ ] newChat
- [ ] newGroup
- [ ] newMessageFrom
- [ ] newMessageNotification
- [ ] newVerifiedGroup
- [ ] qrCode
- [ ] remove
- [ ] searchAriaLabel
jq: error: syntax error, unexpected IDENT, expecting $end (Unix shell quoting issues?) at <top-level>, line 1:
.showArchivedChats ? noArchivedChats : noChats
jq: 1 compile error
- [ ] selectGroupImage
- [ ] showKeyTransferMessage
- [ ] showQrInviteCode
- [ ] showQrVerifyCode
- [ ] this.state.buttonLabel
- [ ] this.state.heading
- [ ] transferKey
- [ ] unblockContacts
- [ ] unblockContacts.noneBlocked
- [ ] utocrypt_prefer_e2ee
- [ ] writeMessage
- [ ] writeMessageAriaLabel
- [ ] newIdentity

pazkero:
While running the desktop version (built from git, no patches applied), I got the following error:

```Could not load messages for pt-BR Error: ENOENT: no such file or directory, open '/home/jesusalva/projetos/deltachat/_locales/pt_BR .json'```

Just like english, where we have en_US vs en_GB along other minor variants, so we have with portuguese (pt_PT vs pt_BR, along other minor variants)

My advise is to add a edge case, that if it tries to load a language and it fails, but there is the underline char, to try to load only the first part of the string.

So if you get 'pt_BR' and can't find 'pt_BR.json', to trim everything after the underline.
This way it'll try to load pt.json which actually exists.


Best regards from Brazil,
Jesusalva

**Comment by ralphtheninja** [comment]
This has become rather messy now. What's the process of fixing the missing resources?

File: src/main/ipc.js [hunk 2]
```diff
@@ -204,16 +204,17 @@ function txCoreStrings () {
   strings[C.DC_STR_MSGADDMEMBER] = tx('systemmsg_member_added')
   strings[C.DC_STR_MSGDELMEMBER] = tx('systemmsg_member_removed')
   strings[C.DC_STR_MSGGROUPLEFT] = tx('systemmsg_group_left')
-  //strings[C.DC_STR_E2E_AVAILABLE] = tx('DC_STR_E2E_AVAILABLE')
-  //strings[C.DC_STR_ENCR_TRANSP] = tx('DC_STR_ENCR_TRANSP')
-  //strings[C.DC_STR_ENCR_NONE] = tx('DC_STR_ENCR_NONE')
+  // TODO why are these commented out?
+  // strings[C.DC_STR_E2E_AVAILABLE] = tx('DC_STR_E2E_AVAILABLE')
+  // strings[C.DC_STR_ENCR_TRANSP] = tx('DC_STR_ENCR_TRANSP')
+  // strings[C.DC_STR_ENCR_NONE] = tx('DC_STR_ENCR_NONE')
```
ralphtheninja: What's the reason for commenting these out?

ralphtheninja: Nevermind, just noticed the `TODO` at the top.

ralphtheninja:
Only four strings left:

* DC_STR_E2E_AVAILABLE
* DC_STR_ENCR_NONE
* DC_STR_ENCR_TRANSP
* DC_STR_STARREDMSGS

@r10s Are these needed to translate? (@Jikstra why did you uncomment these four?)

ralphtheninja:
I've been thinking through how we're doing this and it's still quite not an ideal situation. At the moment we edit the `_locales/_experimental_en.json` as a workaround for not editing `_locales/en.json` which would be overwritten if we do `npm run translations`. But then you run into the problem how how to sync this file with upstream together with the fact that the application _should_ use `_locales/en.json`.

I propose the following:

1. remove `_locales/*.json` from git
1. add `_locales/*.xml` to git
1. when modifying/adding/removing language resources we do this in `_locales/en.xml`, which will be the _canonical_ translation file (just like `_locales/en.json` used to be)
1. run `npm run translations` as a build step when building/deploying the application

Jikstra:
I like your solution a lot more than what we currently have.
It's a lot cleaner, we stay in the same syntax, it's easier to push changes. I like it.

Jikstra:
> @r10s Are these needed to translate? (@Jikstra why did you comment these four?)

They aren't translated at all in the android strings, so i think they are deprecated/not used anymore. @r10s any info?

ralphtheninja:
> What would be the workflow for a new feature where we don't want to push the language strings yet?
> The problem i see is that we don't want to immediatly push new language strings if they change in the next days to not annoy the translators.

Really good question. We just keep the changes in git until we decide they should go live. I think when we release a new version of desktop would be a good time to sync with upstream.

ralphtheninja:
Oh and another thing, `npm run translations` should be `npm run build-translations` and it shouldn't do `tx pull`, it should only build on the current xml files. `tx pull` and `tx push` should be done separately.

r10s:
> Only four strings left:
> 
>     * DC_STR_E2E_AVAILABLE
> 
>     * DC_STR_ENCR_NONE
> 
>     * DC_STR_ENCR_TRANSP
> 
>     * DC_STR_STARREDMSGS
> 
> 
> @r10s Are these needed to translate? (@Jikstra why did you comment these four?)

the strings are all used in places that are about subject to change, so i think it's fine to just ignore them for now.

Issue #372, opened 2018-11-29
Group icon missing


ralphtheninja:
In android there's a small group icon, just before the group name, which is missing on desktop, see below:

![android-screenshot1](https://user-images.githubusercontent.com/308049/49225946-981e8f00-f3e5-11e8-9c69-fb687b45ceca.png)

@r10s Mind searching for the correct image resource and adding it here?

ralphtheninja:
Ping @r10s

r10s:
@ralphtheninja icon added to interface repo at https://github.com/deltachat/interface/blob/master/icons/ic_group_white_24dp.png , the icon should be colored in the same color as the text right of it. 

android-dev uses the same icon in the greenish title (currently right, but i would prefer it left of the name in the same way as in the chatlist)

PR #428, opened 2018-12-12
Render group icon in ChatList

ralphtheninja:
Closes https://github.com/deltachat/deltachat-desktop/issues/372


karissa:
nice @ralphtheninja !

For code organization purposes, wondering if we can have a `Chat.js` module or somehow incorporate it into the same file as `ChatView.js` -- thinking similar to what we had/have for Message, which contains some helper functions in there: and exposes `render` (for creating the React element) and `convert` for transferring the delta-api to api needed for renderer..

ralphtheninja:
> For code organization purposes, wondering if we can have a `Chat.js` module or somehow incorporate it into the same file as `ChatView.js` -- thinking similar to what we had/have for Message, which contains some helper functions in there: and exposes `render` (for creating the React element) and `convert` for transferring the delta-api to api needed for renderer..

I trust your judgement :)

karissa:
Yeah, this way we can just pass `selectedChat` and there is a `isGroup` property on that, rather than passing `isGroup` as a property of the ChatView, or as an extra call/require in the Menu file.

karissa:
(similar to `chat.isVerified`)

ralphtheninja:
> (similar to `chat.isVerified`)

Aah think I know what you mean. I'll tweak it a bit.

src/renderer/components/Menu.js
```diff
@@ -77,7 +76,7 @@ class DeltaMenu extends React.Component {
 
     const tx = window.translate
 
-    const isGroup = isGroupChat(selectedChat)
+    const isGroup = selectedChat.isGroup
```
ralphtheninja: @karissa Something like this?

karissa: yeah better!

ralphtheninja:
@karissa Good to merge?

Issue #404, opened 2018-12-10
Sent webm video isn't watchable


Jikstra:
Someone sent me a webm video but i can't play it from dc-desktop. On android it works. Downloading it and playing it works too.

PR #430, opened 2018-12-13
Lookup mime type using filename

ralphtheninja:
Closes https://github.com/deltachat/deltachat-desktop/issues/404

### DIFF
--- a/src/renderer/components/MessageWrapper.js
+++ b/src/renderer/components/MessageWrapper.js
@@ -6,6 +6,7 @@ const Message = require('./Message')
 const { remote, ipcRenderer } = require('electron')
 const StyleVariables = require('./style-variables')
 const moment = require('moment')
+const mime = require('mime-types')
 
 const GROUP_TYPES = [
   C.DC_CHAT_TYPE_GROUP,
@@ -241,16 +242,20 @@ function convertMessageStatus (s) {
 }
 
 function convertContentType (message) {
-  var filemime = message.filemime
-  if (!filemime) return 'image/jpg'
+  const filemime = message.filemime
+  if (!filemime) return 'application/octet-stream'
   if (filemime === 'application/octet-stream') {
     switch (message.msg.viewType) {
       case C.DC_MSG_IMAGE:
         return 'image/jpg'
       case C.DC_MSG_VOICE:
         return 'audio/ogg'
+      case C.DC_MSG_FILE:
+        const type = mime.lookup(message.msg.file)
+        if (type) return type
+        else return 'application/octet-stream'
       default:
-        return 'application/octect-stream'
+        return 'application/octet-stream'
     }
   }
   return filemime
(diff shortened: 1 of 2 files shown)


src/renderer/components/MessageWrapper.js
```diff
@@ -241,16 +242,19 @@ function convertMessageStatus (s) {
 }
 
 function convertContentType (message) {
-  var filemime = message.filemime
-  if (!filemime) return 'image/jpg'
```
ralphtheninja: I don't think we should return `image/jpg`. It should fallback to `application/octet-stream`.

src/renderer/components/MessageWrapper.js
```diff
@@ -241,16 +242,19 @@ function convertMessageStatus (s) {
...
       case C.DC_MSG_IMAGE:
         return 'image/jpg'
       case C.DC_MSG_VOICE:
         return 'audio/ogg'
+      case C.DC_MSG_FILE:
+        const type = mime.lookup(message.msg.file)
+        if (type) return type
       default:
-        return 'application/octect-stream'
+        return 'application/octet-stream'
```
ralphtheninja: Typo.

src/renderer/components/MessageWrapper.js
```diff
@@ -241,16 +242,19 @@ function convertMessageStatus (s) {
...
+  if (!filemime) return 'application/octet-stream'
   if (filemime === 'application/octet-stream') {
     switch (message.msg.viewType) {
       case C.DC_MSG_IMAGE:
         return 'image/jpg'
       case C.DC_MSG_VOICE:
         return 'audio/ogg'
+      case C.DC_MSG_FILE:
+        const type = mime.lookup(message.msg.file)
+        if (type) return type
```
ralphtheninja: Lookup mime type using file name. For `.webm` files this will be `video/webm`. If `type` can't be fount it will be `false`, which then will fall through to `application/octet-stream` below.

src/renderer/components/MessageWrapper.js
karissa: yeah makes sense, this was a hacky first pass, your solution is nice!

karissa:
LGTM, thanks!

src/renderer/components/MessageWrapper.js
```diff
@@ -241,19 +242,23 @@ function convertMessageStatus (s) {
...
+  const filemime = message.filemime
+
+  if (!filemime) return 'application/octet-stream'
+  if (filemime !== 'application/octet-stream') return filemime
+
+  switch (message.msg.viewType) {
+    case C.DC_MSG_IMAGE:
+      return 'image/jpg'
+    case C.DC_MSG_VOICE:
+      return 'audio/ogg'
```
ralphtheninja: We can probably simplify this switch statement to a simple `mime.lookup()` which should catch `'image/jpg'` and `'audio/ogg'` as well but it's fine for now I guess.

PR #437, opened 2018-12-13
 Advanced account settings

### LINKED ISSUE #249: Settings->Advanced->Account settings
Change email/password/advanced settings for config options.

karissa:
This allows you to edit your account settings (like password, email address, and SMTP/IMAP) from the settings view.
fixes #249 

One TODO left:
@ralphtheninja @Jikstra one last thing to do: what should we do when the user hits 'update settings'? Do we need to just set the new config options? or do we login also? 

screenshot from 2018-12-13 13-29-37
screenshot from 2018-12-13 13-29-29
screenshot from 2018-12-13 13-29-25

EDIT:
Work left:
- [x] Don't need to enter password to update advanced settings (maybe fill password with some stars?)
- [x] Don't allow to change the e-mail
- [x] Get filled in settings and show them
- [x] actually save the submitted changes
- [x] check login/show error if something failed

### DIFF
--- a/src/main/ipc.js
+++ b/src/main/ipc.js
@@ -126,6 +126,7 @@ function init (cwd, state) {
 
   ipc.on('updateCredentials', (e, credentials) => {
     var dir = path.join(os.tmpdir(), Date.now().toString())
+    if (!credentials.mailPw) credentials.mailPw = dc.getConfig('mail_pw')
     var tmp = new DeltaChat(dir, state.saved)
 
     tmp.on('DC_EVENT_CONFIGURE_PROGRESS', function (data1) {
--- a/src/main/ipc.js
+++ b/src/main/ipc.js
@@ -22,6 +22,7 @@ function init (cwd, state) {
 
   const ipc = ipcMain
   const main = windows.main
+  console.log(state)
   const dc = new DeltaChat(cwd, state.saved)
 
   ipc.once('ipcReady', function (e) {
@@ -114,6 +115,10 @@ function init (cwd, state) {
 
   ipc.on('updateSettings', (e, saved) => {
     dc.updateSettings(saved)
+  })
+
+  ipc.on('retrieveAdvancedSettingsObj', (e, ...args) => {
+    windows.main.send('retrieveAdvancedSettingsObjResp', dc.getAdvancedSettings())
   })
 
   function dispatch (name, ...args) {
--- a/src/main/ipc.js
+++ b/src/main/ipc.js
@@ -131,6 +131,7 @@ function init (cwd, state) {
     tmp.on('DC_EVENT_CONFIGURE_PROGRESS', function (data1) {
       if (Number(data1) === 0) { // login failed
         windows.main.send('error', 'Login failed')
+        tmp.close()
       }
     })
 
@@ -142,6 +143,7 @@ function init (cwd, state) {
       if (tmpJson.ready) {
         dc.login(credentials, render, txCoreStrings())
         windows.main.send('success', 'Configuration success!')
+        tmp.close()
       }
     }
 
--- a/src/main/ipc.js
+++ b/src/main/ipc.js
@@ -7,7 +7,9 @@ const {
   ipcMain
 } = require('electron')
 
+const path = require('path')
 const fs = require('fs')
+const os = require('os')
 
 const localize = require('../localize')
 const menu = require('./menu')
@@ -22,7 +24,6 @@ function init (cwd, state) {
 
   const ipc = ipcMain
   const main = windows.main
-  console.log(state)
   const dc = new DeltaChat(cwd, state.saved)
 
   ipc.once('ipcReady', function (e) {
@@ -63,11 +64,17 @@ function init (cwd, state) {
   })
 
   dc.on('DC_EVENT_ERROR', (error) => {
-    windows.main.send('error', error.toString())
+    windows.main.send('error', error)
   })
 
-  dc.on('DC_EVENT_NETWORK_ERROR', (first, error) => {
-    windows.main.send('error', error.toString())
+  dc.on('DC_EVENT_ERROR_NETWORK', (first, error) => {
+    windows.main.send('error', error)
+  })
+
+  dc.on('DC_EVENT_CONFIGURE_PROGRESS', function (data1) {
+    if (Number(data1) === 0) { // login failed
+      windows.main.send('error', 'Login failed!')
+    }
   })
 
   // Calls a function directly in the deltachat-node instance and returns the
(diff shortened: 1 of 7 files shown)


src/renderer/components/dialogs/Settings.js
```diff
@@ -84,9 +87,15 @@ class Settings extends React.Component {
     ipcRenderer.send('updateSettings', this.state.saved)
   }
 
+  onLoginSubmit (event) {
+    console.log('what do we do here??')
```
karissa: @ralphtheninja @Jikstra @r10s what do we do when they update their credentials while already logged in?

ralphtheninja: It would be nice to be able to remove the `'configured'` state, so next time the user logs in, the check for `dc.isConfigured()` fails and it will try to reconfigure again.

karissa: @ralphtheninja i guess that gives me the question: should we test their account credentials here while they're in this screen and re-log them in?

ralphtheninja:
@karissa Cool. Maybe the window title (top most image) should be `Account Settings` instead?

src/renderer/components/dialogs/Settings.js
ralphtheninja: Come to think of it, I'm not sure I like the concept of `Advanced`. Who defines what is advanced or not? It's all settings, right?

ralphtheninja: Sorry, last comment was a general comment, not about this PR at all.

karissa: Most people don't have these imap settings and it can confuse users who just want to login with email/password. This is how it's done in android and it seems to be working for folks

ralphtheninja: Anyway, back to what we were talking about. How is this handled in android?

karissa:
from @r10s : it's basically the same as account creation, you'll be informed about success/error by DC_EVENT_CONFIGURE_PROGRESS. however, on errors, the last working configuraton stays active (the core always has a set or "working" paramters, you do not overwite then eg. by dc_set_config(mail_pw) you should take care on _failures_ if you use dc_get_config() _afterwards_, you should re-set the parameters on dialog exit to the ones before the dialog was entered. there was a longer discussion about this at https://github.com/deltachat/deltachat-core/issues/246

karissa:
@Jikstra @ralphtheninja  do you have any time to finish up this PR? I don't have an account that requires advanced settings

src/renderer/components/dialogs/Settings.js
Jikstra: > what do we do when they update their credentials while already logged in?

Try to login again, if it fails, stay in the advanced tab somehow with an error?

src/renderer/components/dialogs/Settings.js
```diff
@@ -114,7 +114,8 @@ class Settings extends React.Component {
               <Login
                 addr={deltachat.credentials.addr}
                 onSubmit={this.onLoginSubmit}
-                loading={deltachat.configuring}>
+                loading={deltachat.configuring}
+                addrDisabled>
```
ralphtheninja: Aha, is this the same thing as doing `addrDisabled={addrDisabled}`?

ralphtheninja:
I get a white screen. The logs say:

screenshot from 2018-12-18 01-30-02

Jikstra:
> I get a white screen. The logs say:

Fixed with the "Pass deltachat..." commit.

I implemented some basic but important mechanism to show the set config stuff.

src/renderer/components/dialogs/Settings.js
```diff
@@ -93,26 +103,37 @@ class Settings extends React.Component {
...
+          isOpen={advancedSettings !== false}
           title={tx('settingsAccountTitle')}
           icon='settings'
           onClose={() => this.setState({ advancedSettings: false })}>
           <SettingsDialog className={Classes.DIALOG_BODY}>
             <Card elevation={Elevation.ONE}>
               <H5>{tx('settingsAccountTitle')}</H5>
               <Login
-                addr={deltachat.credentials.addr}
+                {...advancedSettings}
```
ralphtheninja: I still don't get what the result of this is.

src/renderer/components/dialogs/Settings.js
```diff
@@ -93,26 +103,37 @@ class Settings extends React.Component {
     this.props.userFeedback({ type: 'success', text: 'Account updated successfully' })
   }
 
+  onAdvancedSettingsOpen () {
+    ipcRenderer.send('retrieveAdvancedSettingsObj')
+  }
+
+  retrieveAdvancedSettingsObjResp (e, advancedSettingsObj) {
```
ralphtheninja: Why not name this `advancedSettings`?

src/renderer/components/dialogs/Settings.js
```diff
@@ -93,26 +103,37 @@ class Settings extends React.Component {
...
+    console.log('retrieveAdvancedSettingsObjResp', advancedSettingsObj)
+    this.setState({ advancedSettings: advancedSettingsObj })
```
ralphtheninja: And do `this.setState({ advancedSettings })`

src/main/ipc.js
```diff
@@ -116,6 +117,10 @@ function init (cwd, state) {
     dc.updateSettings(saved)
   })
 
+  ipc.on('retrieveAdvancedSettingsObj', (e, ...args) => {
+    windows.main.send('retrieveAdvancedSettingsObjResp', dc.getAdvancedSettings())
```
ralphtheninja: Does it have to be this complicated and handled in a special way?

src/renderer/components/dialogs/Settings.js
karissa: this expands the object into it's parts

src/renderer/components/dialogs/Settings.js
karissa: +1

src/renderer/components/dialogs/Settings.js
karissa: no this is the same thing as `addrDisabled={true}`, its annoying but think of it like HTML dom properties

src/main/ipc.js
karissa: we can do instead ipcRenderer(`dispatchSync`, `getAdvancedSettings`)

src/renderer/components/dialogs/Settings.js
ralphtheninja: So if `var obj = { foo: true, bar: false }` then `<Component {...obj} />` would be the same as `<Component obj={true} bar={false} />` ?

ralphtheninja: Think I got it now. It would be the same as `<Component foo={obj.foo} bar={obj.bar} /> `

Jikstra: Yup would be the same :)

src/renderer/components/dialogs/Settings.js
Jikstra: advancedSettings seems to be too generic to me. It's not really explaining what this method is there for.

src/main/ipc.js
```diff
@@ -117,8 +124,28 @@ function init (cwd, state) {
...
+      var tmpJson = tmp.render()
+      json.configuring = tmpJson.configuring
+      windows.main.send('render', json)
+      if (tmpJson.ready) {
+        dc.login(credentials, render, txCoreStrings())
+        windows.main.send('success', 'Configuration success!')
+      }
+    }
+
+    tmp.login(credentials, fakeRender, txCoreStrings())
```
ralphtheninja: I think we need some cleanup as well, the dc context inside the temporary `DeltaChat()` controller should be closed before (or after) `window.main.send()` has been called.

karissa: ok

PR #438, opened 2018-12-13
handle and show errors, including DC_EVENT_ERROR_NETWORK

### LINKED ISSUE #302: handle DC_EVENT_ERROR_NETWORK
up to 0.25.0, network connection errors were reported by a DC_EVENT_ERROR|WARNING plus a string. the core has decided whether to report this as an ERROR or a WARNING depending on DC_EVENT_IS_OFFLINE. 

in the upcoming 0.26.0 we removed the event DC_EVENT_IS_OFFLINE and it's now up to the UI to show warning in this case.

the simplest approach would be to handle DC_EVENT_ERROR_NETWORK just the same as DC_EVENT_ERROR, however, it may make sense to do some optimizations as described in the [documentation](https://c.delta.chat/group__DC__EVENT.html#ga055e6bfcba292678fc5f0d687f6230a1).

related issue on android: https://github.com/deltachat/deltachat-android-ii/issues/129

karissa:
this handles network errors and shows them raw (as we get them from core) in the user interface. Do we need to do any translation for this? @r10s 

fixes #302

### DIFF
--- a/src/main/deltachat.js
+++ b/src/main/deltachat.js
@@ -115,6 +115,12 @@ class DeltaChatController extends events.EventEmitter {
     })
 
     dc.on('DC_EVENT_ERROR', (error) => {
+      this.emit('DC_EVENT_ERROR', error)
+      log.error(error)
+    })
+
+    dc.on('DC_EVENT_NETWORK_ERROR', (error) => {
+      this.emit('DC_EVENT_NETWORK_ERROR', error)
       log.error(error)
     })
   }
--- a/src/main/ipc.js
+++ b/src/main/ipc.js
@@ -61,6 +61,14 @@ function init (cwd, state) {
     windows.main.send('DC_EVENT_IMEX_PROGRESS', progress)
   })
 
+  dc.on('DC_EVENT_ERROR', (error) => {
+    windows.main.send('error', error.toString())
+  })
+
+  dc.on('DC_EVENT_NETWORK_ERROR', (error) => {
+    windows.main.send('error', error.toString())
+  })
+
   // Calls a function directly in the deltachat-node instance and returns the
   // value (sync)
   ipc.on('dispatchSync', (e, ...args) => {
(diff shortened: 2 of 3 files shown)


src/main/deltachat.js
```diff
@@ -115,6 +115,12 @@ class DeltaChatController extends events.EventEmitter {
     })
 
     dc.on('DC_EVENT_ERROR', (error) => {
+      this.emit('DC_EVENT_ERROR', error)
+      log.error(error)
+    })
+
+    dc.on('DC_EVENT_NETWORK_ERROR', (error) => {
+      this.emit('DC_EVENT_NETWORK_ERROR', error)
```
ralphtheninja: This is wrong. The callback has the signature `(first, error)`, see https://c.delta.chat/group__DC__EVENT.html#ga055e6bfcba292678fc5f0d687f6230a1

src/main/ipc.js
```diff
@@ -61,6 +61,14 @@ function init (cwd, state) {
     windows.main.send('DC_EVENT_IMEX_PROGRESS', progress)
   })
 
+  dc.on('DC_EVENT_ERROR', (error) => {
+    windows.main.send('error', error.toString())
+  })
+
+  dc.on('DC_EVENT_NETWORK_ERROR', (error) => {
+    windows.main.send('error', error.toString())
+  })
```
ralphtheninja: We could merge these ones if the delta chat controller emitted the same event.

src/main/deltachat.js
ralphtheninja: We could do `this.emit('error', error)` for both cases.

src/main/ipc.js
karissa: I think it's ok to keep them different for now

src/main/deltachat.js
karissa: awesome thanks

src/main/ipc.js
ralphtheninja: It's fine :)

ralphtheninja: The `.toString()` shouldn't be needed though.

PR #440, opened 2018-12-14
add encryption info for chats

### LINKED ISSUE #31: Contact details
When clicking a contact from within a chat screen, you should be able to see more details about them, including verification numbers

karissa:
TODO:
- [x] implement encryption info for groups
- [x] should you be able to select/copy the encryption info?
- [x] might be a good time to refactor our dialogs..
- [x] finish translations

fixes #31 
screenshot from 2018-12-14 11-41-40

### DIFF
--- a/src/renderer/ScreenController.js
+++ b/src/renderer/ScreenController.js
@@ -17,18 +17,16 @@ class Home extends React.Component {
     this.state = {
       screen: 'SplittedChatListAndView',
       screenProps: {},
-      showAbout: false,
-      showSettings: false,
       message: false
     }
 
     this.changeScreen = this.changeScreen.bind(this)
     this.onError = this.onError.bind(this)
     this.userFeedback = this.userFeedback.bind(this)
+    this.openDialog = this.openDialog.bind(this)
+    this.closeDialog = this.closeDialog.bind(this)
     this.onShowAbout = this.showAbout.bind(this, true)
-    this.onShowSettings = this.showSettings.bind(this, true)
-    this.onCloseSettings = this.showSettings.bind(this, false)
-    this.onCloseAbout = this.showAbout.bind(this, false)
+    this.dialogs = React.createRef()
   }
 
   changeScreen (screen = 'SplittedChatListAndView', screenProps = {}) {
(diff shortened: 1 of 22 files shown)


src/renderer/components/Menu.js
```diff
@@ -104,6 +110,10 @@ class DeltaMenu extends React.Component {
           icon='delete'
           text={deleteMsg}
           onClick={controller.onDeleteChat} />
+        <MenuItem
+          icon='lock'
+          text='Show Encryption Info'
```
ralphtheninja: Needs translation. Make a TODO.

karissa:
okay ready for review @Jikstra @ralphtheninja 

it needs some more styling, but is ok for this pass. see #442

this PR needs some more intensive testing because I refactored all of the dialog code

src/renderer/ScreenController.js
```diff
@@ -54,7 +52,7 @@ class Home extends React.Component {
 
   componentWillUnmount () {
     ipcRenderer.removeListener('showAboutDialog', this.onShowAbout)
-    ipcRenderer.removeListener('showAboutDialog', this.onError)
+    ipcRenderer.removeListener('error', this.onError)
```
karissa: snuck a tiny bug fix in theree

src/renderer/components/Menu.js
```diff
@@ -44,7 +44,7 @@ class Controller {
     if (selectedChat && selectedChat.contacts.length) {
       var contact = selectedChat.contacts[0]
       var message = tx('dialogs.blockContact', contact.displayName)
-      dialogs.confirmation(message, yes => {
+      confirmation(message, yes => {
```
karissa: another cool UI improvement step would to be to make a nicer confirmation dialog that follows our styling, too

src/renderer/components/dialogs/ForwardMessage.js
```diff
@@ -7,7 +7,7 @@ const {
 
 const ContactList = require('../ContactList')
 
-class ForwardMessageDialog extends React.Component {
+class ForwardMessage extends React.Component {
```
karissa: Now all Dialogs need to have their filename match the component they're exporting so that the controller doesn't get confused

src/renderer/components/dialogs/index.js
```diff
@@ -23,5 +23,62 @@ module.exports = {
...
+  }
+
+  render () {
+    const { saved, userFeedback } = this.props
+    const { dialogs } = this.state
+
+    return (
+      <div>
+        {Object.values(dialogs).map((dialog) => {
+          var name = dialog.Component.name
```
karissa: perhaps we shouldn't be rendering all dialogs all the time, or we could find some other way to only render when absolutely necessary (when state.dialogs has changed?)

PR #453, opened 2018-12-17
select chat after contact request has been granted

### LINKED ISSUE #452: Clicking on contact requests doesn't open a new chat

karissa:
fixes #452


karissa:
should be deadDrop.id in the case it's a message, and in the case it comes up in the chatview, it should be also chat.deaddrop.id <- where deaddrop is the message

PR #461, opened 2018-12-19
remove single folder compat message.

karissa:
depends upon https://github.com/deltachat/deltachat-node/pull/210

### DIFF
--- a/src/renderer/ScreenController.js
+++ b/src/renderer/ScreenController.js
@@ -190,9 +190,6 @@ function LoginScreen (props) {
         </Navbar>
       </NavbarWrapper>
       <div className='window'>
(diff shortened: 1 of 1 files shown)


ralphtheninja:
Can still merge this without #210 being done, just not release it.

karissa:
I was hoping we would be able to make a release before merging this because there are some more critical bug fixes and features we could get in, and it seems like 0.30 is stalled today

karissa:
unless you think it's fine and we can just merge it?

hpk42:
On Wed, Dec 19, 2018 at 09:11 -0800, Karissa McKelvey wrote:
> unless you think it's fine and we can just merge it?

technically it can go after the new core and the second imap thread
is started properly in desktop.  practically it can be merged now ;)

PR #464, opened 2018-12-19
autocrypt dialog improvements, fixes #282

### LINKED ISSUE #282: Improvements to entering transfer autocrypt key dialog ui/ux
It would be nice that
1. we don't loose the entered transfer verification key if it's incorrect (currently it gets cleared after submit)
2. as soon as we typed in 4 figures into one input, automatically focus the next one so that users don't have to press tab after every 4th figure

karissa:
* We don't lose the entered transfer verification key if it's incorrect (currently it gets cleared after submit) 
* The autocrypt key is discarded once dialog is exited manually using the X
* Typing 4 figures into one input, automatically focuses the next one so that users don't have to press tab 
* Update and improve visual status messages 
* More accurate base English strings
* Various CSS improvements

### DIFF
--- a/_locales/en.json
+++ b/_locales/en.json
@@ -886,6 +886,15 @@
   },
   "loginKnownAccountsTitle": {
     "message": "Known Accounts"
+  },
+  "autocryptIncorrect": {
+    "message": "Incorrect setup code, please try again."
+  },
+  "autocryptCorrect": {
+    "message": "Successfully transferred Autocrypt setup!"
+  },
+  "autocryptEnterMessage": {
+    "message": "Please enter the setup code that is displayed on the other device."
   }
 }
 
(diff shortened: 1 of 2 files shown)


Jikstra:
Looks good to me & seems to work. @ralphtheninja lets merge? 
I would like to do another minor visual change, remove those increase/decrease things. https://stackoverflow.com/questions/3790935/can-i-hide-the-html5-number-input-s-spin-box

karissa:
Sorry can you be a bit more specific? what do you mean the first two digits weren't shown

Jikstra:
Sorry, normally if you open a autocrypt setup message the first two digits are prefilled. But this maybe doesn't have anything to do with this pr. I will try it out. Maybe a unrelated bug, don't worry :D

PR #466, opened 2018-12-19
Fix #465 not rendering account elements after the one that got deleted

### LINKED ISSUE #465: Forget account button deletes more than one account for render if it's not the last account in list
Basically this is because of this line:
`app.logins.splice(app.logins.indexOf(addr))`
it should be `app.logins.splice(app.logins.indexOf(addr), 1)` otherwise we're slicing out more than one element. 

How to reproduce:
- add multiple accounts
- delete the first from the multiple accounts
- all accounts after the first one dont  get shown too

Jikstra:


karissa:
nice catch!

PR #480, opened 2018-12-26
Temporary partial solution to the group info/edit screen mess

Simon-Laux:
contacts_workaround
+ More visual difference for members and contacts
+ Group members are now at the top

### DIFF
--- a/src/renderer/components/ContactList.js
+++ b/src/renderer/components/ContactList.js
@@ -51,10 +51,10 @@ class ContactList extends React.Component {
   }
 
   render () {
-    const { childProps, onContactClick, sortFunc } = this.props
+    const { childProps, onContactClick, filter } = this.props
     let contacts = this.props.contacts
     if (!contacts) contacts = this._getContacts()
-    if (sortFunc) contacts = contacts.sort(sortFunc)
+    if (filter) contacts = contacts.filter(filter)
     return <div>
       <SearchInput
         onChange={this.handleSearch}
--- a/src/renderer/components/ContactList.js
+++ b/src/renderer/components/ContactList.js
@@ -51,9 +51,10 @@ class ContactList extends React.Component {
   }
 
   render () {
-    const { childProps, onContactClick } = this.props
+    const { childProps, onContactClick, sortFunc } = this.props
     let contacts = this.props.contacts
     if (!contacts) contacts = this._getContacts()
+    if (sortFunc) contacts = contacts.sort(sortFunc)
     return <div>
       <SearchInput
         onChange={this.handleSearch}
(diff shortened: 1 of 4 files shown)


Simon-Laux:
The only disadvantage would be that the method I used doesn't respect the alphabetical order.
But we will replace this screen anyway and re-factor some of the components.

ralphtheninja:
@Simon-Laux Mind renaming the PR title to something better? ".. the group info/edit screen mess" is quite subjective and doesn't really say what the problem is.

ralphtheninja:
I like the differences with an outer circle, but I don't like that the list is being re-sorted each time I click on something. Maybe the list can be sorted _once_ before load and not while actually clicking around? I.e. if I click on a user on accident, it might jump to the top and I have to scroll up and find the user that I accidentally clicked to unselect it.

Jikstra:
> I like the differences with an outer circle, but I don't like that the list is being re-sorted each time I click on something. Maybe the list can be sorted once before load and not while actually clicking around? I.e. if I click on a user on accident, it might jump to the top and I have to scroll up and find the user that I accidentally clicked to unselect it.

Yes we should sort it in the backend.

src/renderer/components/ContactList.js
```diff
@@ -51,9 +51,10 @@ class ContactList extends React.Component {
   }
 
   render () {
-    const { childProps, onContactClick } = this.props
+    const { childProps, onContactClick, sortFunc } = this.props
     let contacts = this.props.contacts
     if (!contacts) contacts = this._getContacts()
+    if (sortFunc) contacts = contacts.sort(sortFunc)
```
Jikstra: Nice idea, but we don't really want to sort the contacts again. They are already sorted and there's one rule, never ever sort again things we get from the core, as they are already sorted. Sorting is slow, the fastest it can get is somewhere around `O n*log(n)`. We don't need to sort, we just need to pick out the contacts if they are in the group or not and keep the order of it. This would result in a time complexity of `O n` plus it would keep the actual alphabetical ordering.

Simon-Laux:
How about this one? it doesn't re-sort them (just filters). And it also has color indication on which contacts were added and removed

Simon-Laux:
If you agree with that method, we should style it a bit and add those strings in the translation system`<thead><tr><th>Groupmembers</th><th>Not in this group</th></tr></thead>`  before merging

src/renderer/components/ContactList.js
Simon-Laux: I replaced the sorting with a filter

Simon-Laux: But we should probably use the right core methods in the future to get the members

ralphtheninja:
@Simon-Laux Please don't merge master into your branch like this. It was bad enough with the translations branch. It's much better if you _rebase_ your commits on top of latest master.

screenshot from 2019-01-02 01-29-17

ralphtheninja:
@Simon-Laux I like the idea though! :)

Simon-Laux:
@ralphtheninja rebase like that?

ralphtheninja:
> @ralphtheninja rebase like that?

Almost! :wink: 

```
$ git checkout master
$ git reset --hard upstream/master
$ git checkout -
$ git rebase master
```

PR #485, opened 2018-12-27
Handle self not in group error

ralphtheninja:
Fixes:

screenshot from 2018-12-27 14-23-25

### DIFF
--- a/src/main/deltachat.js
+++ b/src/main/deltachat.js
@@ -117,19 +117,21 @@ class DeltaChatController extends events.EventEmitter {
       log.warning(warning)
     })
 
-    dc.on('DC_EVENT_ERROR', (error) => {
-      this.emit('DC_EVENT_ERROR', error)
+    const onError = error => {
+      this.emit('error', error)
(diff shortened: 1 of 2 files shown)


src/main/deltachat.js
```diff
@@ -117,19 +117,21 @@ class DeltaChatController extends events.EventEmitter {
...
     dc.on('DC_EVENT_ERROR_NETWORK', (first, error) => {
-      this.emit('DC_EVENT_ERROR_NETWORK', first, error)
-      log.error(error)
+      onError(error)
     })
 
     dc.on('DC_EVENT_ERROR_SELF_NOT_IN_GROUP', (error) => {
-      this.emit('DC_EVENT_SELF_NOT_IN_GROUP', error)
-      log.error(error)
+      onError(error)
```
ralphtheninja: Time to refactor this now since we had three events doing the same thing. We emit `'error'` from the delta chat controller, and the ui should present all this to the user regardless of what type of error it was.

Issue #482, opened 2018-12-26
No chats, no Divider


Simon-Laux:
![no_chats_no_divider](https://user-images.githubusercontent.com/18725968/50459903-36aff880-0970-11e9-929c-900167654c40.png)

Simon-Laux:
Ah just noticed that you don't even have a divider with chats... although the scrollbar acts like one...
Nevertheless we need a divider IMHO.

Simon-Laux:
@Jikstra

ralphtheninja:
Agreed. We need a divider regardless if there are chats or not.

PR #486, opened 2018-12-27
add a divider between chatlist and chatview

Simon-Laux:
#482 
@Jikstra approved it IRL


ralphtheninja:
@Simon-Laux Also nice that you are using upstream directly :)

PR #497, opened 2018-12-29
WIP Logger

Simon-Laux:
Progress:

- [x] Logger class
- [x] main process- update code to use new logger
- [x] log to to file
- [x] stacktrace on warnings and errors
- [x] renderer - update code to use new logger
- switch to rc for app config  -> moved to #550
- logger - options -> moved to #550
- Dedicated Log View Window -> moved to #539


src/logger.js
```diff
@@ -1,24 +1,52 @@
+const esp = require('error-stack-parser')
+const LoggerVariants = [console.debug, console.info, console.warn, console.error, console.error]
 var handler
+
+var OPTIONS = {
+  logDebug: true,
+  alsoLogInLocalConsole: true
```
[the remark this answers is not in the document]
Simon-Laux: If we do that is the fact clear, that it enables both stdout on main and output in the chome console

ralphtheninja:
Ignore my comments for now. Let's just get this working and merged. I can make a PR with refactoring later on.

src/logger.js
```diff
@@ -1,24 +1,52 @@
...
+}
+
 /** specify function that passes the message to the logger in the main process */
 function setLogHandler (LogHandler) {
   handler = LogHandler
 }
 
+function log (channel, lvl, ...args) {
+  if (handler) {
```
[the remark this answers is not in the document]
ralphtheninja: Yeah, but this is a result of the design. If we make sure to pass the handler at log creation we should be good. This is really an event emitter pattern, so the logger class should instead be an event emitter and the handler just a listener to a log event, then we don't need the extra checks.

Simon-Laux:
We only have left to implement these Flags before starting with the gui part:
```
Flags -> (log Options)
> don't log debug messages
> log also to respective consoles
> copy core logs to devconsole?```
The question now is how to implement them? I have 3 ideas:
- `ENV Variables` - Probably not the best for releases
- starting arguments like `--debug` 
- config file in the user directory - that can maybe also be modified via the settings

Maybe also a combination of the 2 latter ones like use args, if they don't exists then config file and if there isn't even a config file - then use the default ones.

Am I overengineering with my thoughts and should just use args or is this a valid point - whats your opinion?

**Comment by ralphtheninja** [comment]
Check out the rc module by dominictarr. It let's you pick any of the methods you listed or any combination there of.

We should use rc for app config as a whole

**Comment by Jikstra** [comment]
> Check out the rc module by dominictarr. It let's you pick any of the methods you listed or any combination there of.
>
> We should use rc for app config as a whole

`rc` looks good. The whole app config stuff is a mess currently, i think with rc and some refactoring we could make it a lot nicer. We also need this for features like auto login or an intro on first login.

**Comment by ralphtheninja** [comment]
I think we should try to close this asap, this PR is getting bigger and bigger. I propose we make a separate issue and PR for `rc`. Once that's in place, we can start configuring the logger. Lets just give it some sane defaults for now.

Also see https://github.com/deltachat/deltachat-desktop/pull/540#issuecomment-451713444

**Comment by Simon-Laux** [comment]
So.. @ralphtheninja after you confirmed that it works as expected we can merge

**Comment by Simon-Laux** [comment]
and also before I forget - we should also push the language strings when merging

File: src/main/deltachat.js [hunk 3]
```diff
@@ -253,14 +257,14 @@ class DeltaChatController extends events.EventEmitter {
   createChatByContactId (contactId) {
     const contact = this._dc.getContact(contactId)
     if (!contact) {
-      log.warning('no contact could be found with id', contactId)
+      log.warn('no contact could be found with id', contactId)
       return 0
     }
     const chatId = this._dc.createChatByContactId(contactId)
-    log('created chat', chatId, 'with contact', contactId)
+    log.debug(`created chat ${chatId} with contact' ${contactId}`)
```
ralphtheninja: There's a left over `'` right after `contact`.

src/main/deltachat.js
```diff
@@ -278,7 +282,7 @@ class DeltaChatController extends events.EventEmitter {
    * Dispatched from EditGroup
    */
   modifyGroup (chatId, name, image, remove, add) {
-    log('modify group', chatId, name, image, remove, add)
+    log.debug('action - modify group', { chatId, name, image, remove, add })
```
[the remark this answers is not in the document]
Simon-Laux: yeah we should..., but also have the option to set an logging code/label -> a machine readable thing

src/main/deltachat.js
```diff
@@ -46,11 +50,11 @@ class DeltaChatController extends events.EventEmitter {
     dc.open(cwd, err => {
       if (err) throw err
       const onReady = () => {
-        log('Ready')
+        log.info('Ready')
         this.ready = true
         this.configuring = false
         this.emit('ready')
-        log('dc_get_info', dc.getInfo())
+        log.info('dc_get_info', dc.getInfo(), 'dc_get_info')
```
ralphtheninja: Does the third parameter really add any value here? It's identical to the first.

src/main/deltachat.js
```diff
@@ -29,13 +29,17 @@ class DeltaChatController extends events.EventEmitter {
...
+
   /**
    * Dispatched when logging in from Login
    */
   login (credentials, render, coreStrings) {
     // Creates a separate DB file for each login
     const cwd = this.getPath(credentials.addr)
-    log('Using deltachat instance', cwd)
+    log.info('Using deltachat instance', cwd, 'dc_instance')
     this._dc = new DeltaChat()
```
ralphtheninja: Why the third parameter?

ralphtheninja: Nevermind.

src/main/deltachat.js
ralphtheninja: Nevermind.

src/main/deltachat.js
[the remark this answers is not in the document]
ralphtheninja: I'm just thinking out loud. Lets use it and see how it feels first.

src/main/deltachat.js
Simon-Laux: it does because its the 'error' code

src/main/deltachat.js
Simon-Laux: 'error' code

src/main/deltachat.js
Simon-Laux: oh right that seems to be useless

PR #501, opened 2018-12-29
improve build / run instructions

obestwalter:
Trying it out on Arch Linux I noticed that the install step is missing, before building.

I also fixed a minor formatting problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # deltachat-desktop
 
-> Desktop Application for [delta.chat](https://delta.chat)
+**Desktop Application for [delta.chat](https://delta.chat)**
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,6 @@
 # deltachat-desktop
 
-> Desktop Application for [delta.chat](https://delta.chat)
+**Desktop Application for [delta.chat](https://delta.chat)**
```
ralphtheninja: Why this change?

obestwalter: I assumed that this was an oversight (and to me it looks wrong). 

`>` is meant for quotes but that sentence isn't, so I replaced it with another form of emphasis that doesn't look like somebody accidentally formatted it as a quote.

ralphtheninja: It's quite common to use `>` for short descriptions in README.

ralphtheninja: But I'll let your change pass to keep _everyone_ happy but me.

Issue #502, opened 2018-12-30
Translation fallback for language variants


ralphtheninja:
See https://github.com/deltachat/deltachat-desktop/pull/422#issuecomment-450535198

`electron` has support for the following locales https://electronjs.org/docs/api/locales (`app.getLocale()`)

PR #518, opened 2019-01-02
Use locale fallback, e.g. en-US falls back to en if it does not exist

ralphtheninja:
Closes https://github.com/deltachat/deltachat-desktop/issues/502

### DIFF
--- a/src/localize.js
+++ b/src/localize.js
@@ -1,9 +1,3 @@
-module.exports = {
-  setup,
-  translate,
-  getLocaleMessages
-}
-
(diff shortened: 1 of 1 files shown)


Jikstra:
Is there a easy way to test this by setting some environment variables?

ralphtheninja:
> Is there a easy way to test this by setting some environment variables?

I guess you could edit your `~/.config/DeltaChat/config.json` file:

screenshot from 2019-01-02 15-12-26

Just change locale to e.g. `en-US`

Issue #493, opened 2018-12-28
autocrypt setup message dialogue stays open


muelli:
- Platform (linux/mac/windows):
linux

- Delta Chat Desktop Version:
0.97.0

- Expected behavior:

I expected the autocrypt setup dialogue to go away once successful.

- Actual behavior:

it stays open

- Steps to reproduce the problem:

successfully transfer your key using autocrypt setup message

- Screenshots:

![screenshot from 2018-12-28 19-02-37](https://user-images.githubusercontent.com/322807/50524046-c4f0bf80-0ad3-11e9-979c-165321973c1c.png)

- Logs:

PR #524, opened 2019-01-04
Use provided onClose() method (this.onClose() does not exist)

ralphtheninja:
Closes https://github.com/deltachat/deltachat-desktop/issues/493


src/renderer/components/dialogs/SetupMessage.js
```diff
@@ -131,7 +131,7 @@ class SetupMessage extends React.Component {
     } else {
       this.setState({ loading: false })
       this.props.userFeedback({ type: 'success', text: tx('autocrypt_correct_desktop') })
-      this.onClose()
+      this.props.onClose()
```
ralphtheninja: This is yet another bug that happened during rewrite/refactoring.

PR #525, opened 2019-01-04
Update readme on transifex

ralphtheninja:
Note that I have run `tx push --source` and it seems to work fine, the desktop strings were added to `strings.xml` on Transifex.

@r10s You should probably do `tx pull` in android repo to sync with desktop updates.

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -4,7 +4,7 @@ host = https://www.transifex.com
 [delta-chat-app.stringsxml]
 file_filter = _locales/<lang>.xml
 minimum_perc = 0
-source_file = _locales/xx.xml
-source_lang = xx
+source_file = _locales/en.xml
+source_lang = en
 type = ANDROID
 
--- a/README.md
+++ b/README.md
@@ -133,14 +133,23 @@ or keyboard while the tests are running.
 
 Install the [transifex-client](https://docs.transifex.com/client) and get added to the `Delta Chat App` project.
 
-And periodically we can run the following command to get the new translation
-strings from translators:
+And periodically we can run the following command to get the new translation strings from translators:
 
 ```
-npm run translations
+tx pull --all
 ```
 
-This converts the translations from xml to json. Commit these new translations to the repository.
+Note that this command updated `_locales/*.xml`. In order to convert from xml to json you can do:
+
+```
+npm run build-translations
+```
+
+When you need to modify language strings, this should be done in `_locales/en.xml`. To sync with Transifex you do:
+
+```
+tx push --source
+```
 
 ### Deploy workflow
 


r10s:
yip, i run `tx pull` regularly also to get the new translations as such.

ralphtheninja:
@r10s I'm a bit worried that we will overwrite new strings, say some new strings are added to `android-ii` and you push them, I add some strings to `desktop` and push them, thus overwriting your changes.

r10s:
great work :)

@ralphtheninja overwriting strings may happen. however, when i do a `git diff`after pulling the source from transifex (just [wrote a little script for this](https://github.com/deltachat/deltachat-android-ii/blob/master/tools/tx-pull-source.sh)), this should not be a big problem.

in fact, i just did this and re-added a string that was added by android-dev in between, plus some comments i added for the translators.

wrt `_desktop` suffixed-strings: as a rule of thumb, i would try to add the suffix only for strings that are _different_ from existing ones, eg. the strings save_desktop ("Save") or select_group_image_desktop ("Select group image") may also become needed on android/ios/whatever. same for "undo" "redo" and so on.

however, i think for now it's fine :)

Jikstra:
> in fact, i just did this and re-added a string that was added by android-dev in between, plus some comments i added for the translators.

We should be really careful about this.

ralphtheninja:
> @ralphtheninja overwriting strings may happen. however, when i do a `git diff`after pulling the source from transifex (just [wrote a little script for this](https://github.com/deltachat/deltachat-android-ii/blob/master/tools/tx-pull-source.sh)), this should not be a big problem.

We just need to communicate more around when making these changes. @Jikstra 's idea on experimental changes will help with this.

ralphtheninja:
> wrt `_desktop` suffixed-strings: as a rule of thumb, i would try to add the suffix only for strings that are _different_ from _existing_ ones, eg. the strings save_desktop ("Save") or select_group_image_desktop ("Select group image") may also become needed on android/ios/whatever. same for "undo" "redo" and so on.
> (of course i can use select_group_image_desktop also on android, however, this looks weird :)

We have tried to do this, but maybe some strings slipped through the cracks. We can solve this over time "refactoring" the resources. Maybe even start with a script that looks for duplicates.

r10s:
an alternative might be to do pr to transifex.
btw. i just corrected a type in the sources ("frennch") and did a `tx push -s` without doing a `tx pull -l en` before - i hope you're not pushed sth in between. i have also to be more careful about this ...

r10s:
cool :)
one thing to mention: the `tx pull -l en` is also important as otherwise the `tx push -s` will not only overwrite the existing source file (we can handle that, see above) but also translations already done.

PR #532, opened 2019-01-05
Give conversion script a better name

ralphtheninja:


Jikstra:
how about `bin/translations-convert-xml-to-json` ? Otherwise it's unclear that it's translations related.

ralphtheninja:
What about `script-for-converting-transifex-xml-files-to-json-files-because-that-is-what-desktop-wants`? :laughing:

Jikstra:
There's a thin line between being too verbose and too less verbose :D But naming is hard ^^

PR #537, opened 2019-01-05
add google noto + google noto emoji and remove image emojis; fixes #483

### LINKED ISSUE #483: Emojis in input field are displayed as tofu

Simon-Laux:
closes #483 
It also adds Google Noto Mono as fallback to make tofu chars also in other languages much more unlikely.


Simon-Laux:
We also need someone that can test it on a mac just to be sure that it works there too

src/renderer/components/conversations/Emojify.js
```diff
@@ -1,33 +1,19 @@
 const React = require('react')
 const classNames = require('classnames')
-const is = require('@sindresorhus/is')
```
ralphtheninja: We still use this elsewhere? Or can it be removed now?

ralphtheninja: screenshot from 2019-01-05 20-46-28

Doesn't seem like it.

Simon-Laux: I'll remove it

PR #540, opened 2019-01-06
Refactor and simplify state load

ralphtheninja:
Many small commits, but simplifies quite a bit. Squash at merge.

### DIFF
--- a/src/renderer/lib/state.js
+++ b/src/renderer/lib/state.js
@@ -59,39 +59,21 @@ function getDefaultState () {
     saved: {
       markRead: true,
       notifications: true,
-      showNotificationContent: true
+      showNotificationContent: true,
+      locale: 'en'
     }
   }
-}
-
-/* If the saved state file doesn't exist yet, here's what we use instead */
-function setupStateSaved (cb) {
-  const saved = {
-    locale: null,
-    version: config.APP_VERSION,
-    markRead: true
-  }
-  cb(null, saved)
 }
 
 function load (cb) {
   appConfig.read(function (err, saved) {
-    if (err || !saved.version) {
-      console.log('Missing config file: Creating new one')
-      setupStateSaved(onSavedState)
-    } else {
-      onSavedState(null, saved)
+    if (err) {
+      console.log('Missing configuration file. Using default values.')
     }
-  })
-
-  function onSavedState (err, saved) {
-    if (err) return cb(err)
     const state = getDefaultState()
-    // override default savable settings with saved ones
-    state.saved = Object.assign(state.saved, saved)
-
+    state.saved = Object.assign(state.saved, err ? {} : saved)
     cb(null, state)
-  }
+  })
 }
 
 // Write state.saved to the JSON state file


src/renderer/lib/state.js
```diff
@@ -64,21 +64,15 @@ function getDefaultState () {
...
-    markRead: true
-  }
-  cb(null, saved)
-}
-
 function load (cb) {
   appConfig.read(function (err, saved) {
     if (err || !saved.version) {
       console.log('Missing config file: Creating new one')
-      setupStateSaved(onSavedState)
```
ralphtheninja: Backwards way of simply calling a function.

src/renderer/lib/state.js
```diff
@@ -66,11 +66,10 @@ function getDefaultState () {
 
 function load (cb) {
   appConfig.read(function (err, saved) {
-    if (err || !saved.version) {
+    if (err) {
       console.log('Missing config file: Creating new one')
       onSavedState(null, {
         locale: null,
-        version: config.APP_VERSION,
```
ralphtheninja: This information is completely irrelevant. We always know the version from `package.json`.

src/renderer/lib/state.js
```diff
@@ -69,7 +70,6 @@ function load (cb) {
     if (err) {
       console.log('Missing config file: Creating new one')
       onSavedState(null, {
-        locale: null,
```
ralphtheninja: Lets use sane defaults, `null` isn't one of them.

ralphtheninja: Note that this value can be moved up to the default values above, since we're merely doing `Object.assign()` on top of the default values, this value might as well be part of the default values.

src/renderer/lib/state.js
```diff
@@ -69,9 +69,7 @@ function load (cb) {
   appConfig.read(function (err, saved) {
     if (err) {
       console.log('Missing config file: Creating new one')
-      onSavedState(null, {
-        markRead: true
```
ralphtheninja: `markRead` is already `true` in default state above.

src/renderer/lib/state.js
```diff
@@ -69,10 +69,8 @@ function load (cb) {
   appConfig.read(function (err, saved) {
     if (err) {
       console.log('Missing config file: Creating new one')
-      onSavedState(null, {})
-    } else {
-      onSavedState(null, saved)
     }
+    onSavedState(null, err ? {} : saved)
```
ralphtheninja: Now it's enough to call `onSavedState()` _once_, with a value depending on the value of `err`.

src/renderer/lib/state.js
```diff
@@ -70,15 +70,12 @@ function load (cb) {
     if (err) {
       console.log('Missing config file: Creating new one')
     }
-    onSavedState(null, err ? {} : saved)
+    onSavedState(err ? {} : saved)
   })
 
-  function onSavedState (err, saved) {
```
ralphtheninja: `err` is always null, so no need for this parameter or doing `cb(err)` since it's dead code.

src/renderer/lib/state.js
```diff
@@ -70,14 +70,10 @@ function load (cb) {
     if (err) {
       console.log('Missing config file: Creating new one')
     }
-    onSavedState(err ? {} : saved)
-  })
-
-  function onSavedState (saved) {
     const state = getDefaultState()
     state.saved = Object.assign(state.saved, saved)
     cb(null, state)
```
ralphtheninja: Just move this code into the function above, no need for `onSavedState()`.

src/renderer/lib/state.js
```diff
@@ -68,7 +68,7 @@ function getDefaultState () {
 function load (cb) {
   appConfig.read(function (err, saved) {
     if (err) {
-      console.log('Missing config file: Creating new one')
```
ralphtheninja: This message is incorrect, we're not creating the configuration file here, it's missing so we're using default values.

ralphtheninja:
Note that `~/.config/DeltaChat/config.json` (other path for windows) is actually used for saving the `state` which doesn't have _anything_ to do with `src/config.js`, which is a bit confusing.

Simon-Laux:
Is this really a state or is this config or settings?

ralphtheninja:
> Is this really a state or is this config or settings?

It's a state I guess or a mix thereof. We persist e.g. window size etc.

Issue #528, opened 2019-01-04
Missing translations


ralphtheninja:
Select `More info...` on a chat message

![screenshot from 2019-01-04 20-13-43](https://user-images.githubusercontent.com/308049/50706198-6ee4c480-105d-11e9-85a2-77326f2cd30e.png)

ralphtheninja:
And 

![screenshot from 2019-01-05 22-57-49](https://user-images.githubusercontent.com/308049/50729625-6ebbf600-113d-11e9-8fd1-686b25e84578.png)

In the media page

PR #546, opened 2019-01-06
Missing translations

ralphtheninja:
Closes https://github.com/deltachat/deltachat-desktop/issues/528

### DIFF
--- a/src/renderer/components/Media.js
+++ b/src/renderer/components/Media.js
@@ -53,28 +53,25 @@ const MediaGalleryItem = styled.div`
 
 const GROUPS = {
   images: {
-    key: 'Images',
     values: [C.DC_MSG_GIF, C.DC_MSG_IMAGE]
   },
-  videos: {
-    key: 'Videos',
+  video: {
     values: [C.DC_MSG_VIDEO]
   },
   audio: {
-    key: 'Audio',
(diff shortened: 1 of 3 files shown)


src/renderer/components/Media.js
```diff
@@ -53,28 +53,25 @@ const MediaGalleryItem = styled.div`
...
   },
   audio: {
-    key: 'Audio',
     values: [C.DC_MSG_AUDIO, C.DC_MSG_VOICE]
   },
   documents: {
-    key: 'Documents',
     values: [C.DC_MSG_FILE]
   }
 }
```
ralphtheninja: Lets just use the keys from `GROUPS` object. No need to add extra keys which are essentially identical.

src/renderer/components/Media.js
```diff
@@ -53,28 +53,25 @@ const MediaGalleryItem = styled.div`
 
 const GROUPS = {
   images: {
-    key: 'Images',
     values: [C.DC_MSG_GIF, C.DC_MSG_IMAGE]
   },
-  videos: {
-    key: 'Videos',
+  video: {
```
ralphtheninja: Should be singular

ralphtheninja:
pushed `en.xml`

Issue #544, opened 2019-01-06
Some characters are escaped (chat list)


ralphtheninja:
![screenshot from 2019-01-06 13-39-08](https://user-images.githubusercontent.com/308049/50736080-ab790300-11b8-11e9-93f2-63d8022e09fe.png)

This _might_ be related to conversion from xml to json and not just this message.

ralphtheninja:
Maybe related (if not, it should be a separate issue). I changed group name in desktop and the result is

![screenshot from 2019-01-06 14-00-56](https://user-images.githubusercontent.com/308049/50736272-a8334680-11bb-11e9-9df2-03031916fd33.png)

PR #554, opened 2019-01-07
Make unit tests less spammy

ralphtheninja:
Makes sure each `.xml` file has a corresponding `.json` file and instead of making a single test for each file, we test that all files are valid .json (less spammy).


ralphtheninja:
I'd like to pop this PR in before solving https://github.com/deltachat/deltachat-desktop/issues/544 (want to add tests for this).

PR #560, opened 2019-01-08
appstream: updated appstream file with release information

muelli:


muelli:
it'd be nice if this file could be kept in sync, then the flatpak version would not need to maintain it in a separate repository.

static/chat.delta.desktop.appdata.xml
```diff
@@ -19,15 +20,93 @@ Free, independent, email compatible.
...
-  </kudos>
 
  <url type="homepage">https://delta.chat/</url>
 
   <releases>
-    <release version="v1.7.4" date="2018-11-07"/>
-    <release version="v1.7.3" date="2018-11-02"/>
-​  </releases>
+    <release version="v0.98.1" date="2019-01-06">
+        <description>
```
ralphtheninja: @muelli Did you write this manually? If not, maybe we should have a script to update this when changelog has changed?

muelli: yeah, it's a manual effort.
It would be good to generate this file.
Ideally, electron-builder does this, but so far it doesn't generate any appstream files.
If these release notes were used in git tags, then it would be possible to extract it from git. But it seems no annotated tags are being used.

PR #563, opened 2019-01-08
Restyle create chat buttons

Jikstra:
screenshot_20190108_175351

### DIFF
--- a/src/renderer/components/CreateChat.js
+++ b/src/renderer/components/CreateChat.js
@@ -91,7 +91,7 @@ class CreateChat extends React.Component {
         <div className='window'>
           <div className='CreateChat'>
             <OvalDeltaButton onClick={this.onCreateGroup}>{tx('menu_new_group')}</OvalDeltaButton>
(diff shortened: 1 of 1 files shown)


src/renderer/components/CreateChat.js
```diff
@@ -10,9 +10,29 @@ const {
...
 } = require('@blueprintjs/core')
 
+const styled = require('styled-components').default
+const StyleVariables = require('./style-variables')
+
 const ContactList = require('./ContactList')
 const NavbarWrapper = require('./NavbarWrapper')
 
+const OvalDeltaButton = styled.button`
+  background-color: ${StyleVariables.colors.deltaPrimaryBg}
```
ralphtheninja: Nice!

Jikstra: We could move that into an own file as soon as we want to reuse that button. It's better than the normal button, but still not very nice. It's time for some designer :)

Jikstra:
This should be ready to merge @ralphtheninja

PR #564, opened 2019-01-08
When clicking on advanced, the top of the login form was hidden below

Jikstra:
the navbar

### DIFF
--- a/src/renderer/ScreenController.js
+++ b/src/renderer/ScreenController.js
@@ -177,13 +177,14 @@ const LoginWrapper = styled.div`
   height: calc(100vh);
 
   .bp3-card {
+    width: 400px;
     margin-top: 20px;
   }
 
-  .window {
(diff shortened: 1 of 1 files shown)


src/renderer/ScreenController.js
```diff
@@ -177,13 +177,14 @@ const LoginWrapper = styled.div`
...
   .bp3-card {
+    width: 400px;
     margin-top: 20px;
   }
 
-  .window {
-    max-width: 400px;
-    height: auto;
+  .window { 
+    padding-left: calc((100vw - 400px) / 2)
```
ralphtheninja: What's 100vw? Volks Wagen? :)

Jikstra: viewport width :) quite a useful unit for css hacking ^^
https://www.w3schools.com/CSSref/css_units.asp

PR #568, opened 2019-01-08
Restyle buttons in create chat

Jikstra:
for search input. This currently removes the possibility to clear the search input which is a regression, but maybe only a small one. Putting in a js version of this should be doable.

### DIFF
--- a/src/renderer/components/SearchInput.js
+++ b/src/renderer/components/SearchInput.js
@@ -1,13 +1,37 @@
 const React = require('react')
 const { InputGroup } = require('@blueprintjs/core')
+const styled = require('styled-components').default
 
-module.exports = function (props) {
-  const tx = window.translate
-  return <InputGroup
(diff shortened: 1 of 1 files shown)


hpk42:
nice ;)

Jikstra:
@ralphtheninja @Simon-Laux maybe lets add in the old cancel button even if it's ugly? Better than to not have one.

ralphtheninja:
> @ralphtheninja @Simon-Laux maybe lets add in the old cancel button even if it's ugly? Better than to not have one.

I have no strong opinion. Do what you feel like is best :)

Simon-Laux:
I don't care about that button, I rarely use such buttons anyway... Rather remove it before the release and add a better version later I would say

Jikstra:
I'm going to keep it in and replace it at some other point.

Issue #515, opened 2019-01-01
xml to json convert script bugs when _text node is an array


ralphtheninja:
Consider the following resource (from `_locales/sq.xml`):

```xml
<string name="qrscan_no_addr_found">Ky kod QR përmban shenja gishtash, por jo 
adresë email.\n\nPër një verifikim <em>out-of-band</em>, ju lutemi, së pari vendosni 
me marrësin një lidhje të fshehtëzuar.</string>
```

Which results in the following json resource (e.g. `_locales/sq.json`), where the `message` is actually an array rather than a string.

```json
"qrscan_no_addr_found": {
    "message": [
      "Ky kod QR përmban shenja gishtash, por jo adresë email.\\n\\nPër një verifikim ",
      ", ju lutemi, së pari vendosni me marrësin një lidhje të fshehtëzuar."
    ]
  }
```

ralphtheninja:
We should probably just do a join of those strings. But more investigation is needed.

PR #569, opened 2019-01-09
Fix escaped characters in translations

ralphtheninja:
Closes https://github.com/deltachat/deltachat-desktop/issues/544, closes https://github.com/deltachat/deltachat-desktop/issues/515

screenshot from 2019-01-09 15-28-28
screenshot from 2019-01-09 15-28-53

### DIFF
--- a/_locales/tr.xml
+++ b/_locales/tr.xml
@@ -236,13 +236,13 @@
     <string name="welcome_intro2_headline">Bağımsız</string>
     <string name="welcome_intro2_message">Hiç Delta Chat sunucusu yoktur. Uygulama, yalnızca posta sunucunuzu kullanır.</string>
     <string name="welcome_intro3_headline">Hızlı</string>
-    <string name="welcome_intro3_message">İletileri saniyeler içinde gönderin.<br/>Hızlı arabirim.</string>
+    <string name="welcome_intro3_message">İletileri saniyeler içinde gönderin.\nHızlı arabirim.</string>
     <string name="welcome_intro4_headline">Güçlü</string>
     <string name="welcome_intro4_message">Sınırsız sohbetler, görüntüler, videolar, sesli iletiler ve daha fazlası. Çoklu istemci yetenekli.</string>
     <!-- Translators: "free" as in freedom here, not "free" as in "free beer" -->
     <string name="welcome_intro5_headline">Özgür</string>
     <!-- Translators: "free" as in freedom here, not "free" as in "free beer" -->
-    <string name="welcome_intro5_message">Delta Chat sonsuza dek özgürdür.<br/>Açık kaynaklı. Reklâm yok. Abonelik yok. Satıcı kilidi yok.</string>
+    <string name="welcome_intro5_message">Delta Chat sonsuza dek özgürdür.\nAçık kaynaklı. Reklâm yok. Abonelik yok. Satıcı kilidi yok.</string>
     <string name="welcome_intro6_headline">Şifreli</string>
     <string name="welcome_intro6_message">Tüm yaygın algoritmalarla şifrelidir. İletiler sunucularınızda kalır.</string>
     <string name="welcome_intro7_headline">Güvenilir</string>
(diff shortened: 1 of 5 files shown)


ralphtheninja:
A tonne of erroneous translations. For now I have adjusted the xml resources as well, but we should probably tweak the conversion script later to handle all the edge cases (some translators add xml tags _inside_ the resources, such as `<br/>` and `<em>foobar</em>`).

_locales/tr.xml
```diff
@@ -236,13 +236,13 @@
...
     <string name="welcome_intro2_message">Hiç Delta Chat sunucusu yoktur. Uygulama, yalnızca posta sunucunuzu kullanır.</string>
     <string name="welcome_intro3_headline">Hızlı</string>
-    <string name="welcome_intro3_message">İletileri saniyeler içinde gönderin.<br/>Hızlı arabirim.</string>
+    <string name="welcome_intro3_message">İletileri saniyeler içinde gönderin.\nHızlı arabirim.</string>
     <string name="welcome_intro4_headline">Güçlü</string>
     <string name="welcome_intro4_message">Sınırsız sohbetler, görüntüler, videolar, sesli iletiler ve daha fazlası. Çoklu istemci yetenekli.</string>
     <!-- Translators: "free" as in freedom here, not "free" as in "free beer" -->
     <string name="welcome_intro5_headline">Özgür</string>
     <!-- Translators: "free" as in freedom here, not "free" as in "free beer" -->
-    <string name="welcome_intro5_message">Delta Chat sonsuza dek özgürdür.<br/>Açık kaynaklı. Reklâm yok. Abonelik yok. Satıcı kilidi yok.</string>
```
ralphtheninja: This doesn't have a new line in the original

test/unit/index.js
```diff
@@ -14,20 +14,48 @@ test('that translation files are valid json', t => {
       jsonFiles.map(name),
       'each xml file has a corresponding json file'
     )
+    const testFile = file => {
+      let json = null
+      try {
+        json = require(file)
+      } catch (e) {
+        console.error(e.message)
+        return false
```
ralphtheninja: First tests makes sure the file is valid json.

test/unit/index.js
```diff
@@ -14,20 +14,48 @@ test('that translation files are valid json', t => {
...
+      }
+      return Object.keys(json).every(k1 => {
```
ralphtheninja: For every resource ..

test/unit/index.js
```diff
@@ -14,20 +14,48 @@ test('that translation files are valid json', t => {
...
+        const v1 = json[k1]
+        return Object.keys(v1).every(k2 => {
```
ralphtheninja: .. every key of that resource (e.g. `"message"` etc) ..

test/unit/index.js
```diff
@@ -14,20 +14,48 @@ test('that translation files are valid json', t => {
...
+          const v2 = v1[k2]
+          if (typeof v2 !== 'string') {
+            console.error(
+              `> ${JSON.stringify(v2)} not a string (${file} -> ${k1} -> ${k2})`
+            )
+            return false
+          }
+
```
ralphtheninja: It should be a string ..

test/unit/index.js
```diff
@@ -14,20 +14,48 @@ test('that translation files are valid json', t => {
...
+              console.error(`> ${JSON.stringify(v2)} contains ${str} (${file} -> ${k1} -> ${k2}) (index: ${match.index})`)
+              return false
+            }
+            return true
+          }
+
+          if (!testString('\\\\n')) return false
+          if (!testString("\\\\'")) return false
+          if (!testString('\\\\\\"')) return false
+          if (!testString('\\\\')) return false
```
ralphtheninja: And it shouldn't contain these.

ralphtheninja: Could probably refactor this into one `return` statement. Up to you guys.

r10s:
> It would be nice to merge this today. It kills two bugs.

i just pulled the translations from transifex - are these already the changes from this pr?

just asking because it is a bit dangerous to push sources to transifex as this will always overwrite all work of the translators (transifex is not line bases afaik) since the last pull (16 hours ago in this case).
but probably you know this best :)

for android, therefore, i never edit the strings-LANG.xml files locally but do this always in the web-frontend.

ralphtheninja:
@r10s I didn't push these to Transifex. I edited them on Transifex.

test/unit/index.js
Jikstra: It's fine with multiple returns to me

PR #571, opened 2019-01-09
Update install instructions in README to cover Arch & Mac.

### LINKED ISSUE #504: Improve install instructions in README
- [ ] Mac
- [ ] Linux
  - [ ] Arch (with AUR)
  - [x] Generic linux (`.AppImage` and `.deb`)
- [x] ~~Windows?!~~
- [x] How to install from source

Jikstra:
- [x] Link to arch linux in table of contents isn't working
- [x] Maybe we should do a general troubleshooting for missing dependencies? This is important during building from source but also on appimages/dmg... as sasl or some other can be missing.

Feel free to push against this branch to make this readme nice and shiny :)

 Fixes #504

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,12 @@
 <details><summary>Click to expand</summary>
 
 - [Install](#install)
+  - [Linux](#linux)
+    - [AppImage](#appimage)
+    - [Debian/Ubuntu](#debian/ubuntu)
+    - [Arch Linux](#arch linux)
+  - [macOS](#macOS)
+  - [From Source](#from)
 - [Configuration and Databases](#configuration-and-databases)
 - [Troubleshooting](#troubleshooting)
 - [How to Contribute](#how-to-contribute)
@@ -27,7 +33,10 @@ The application can be downloaded from the [`Releases`](https://github.com/delta
 
 ### Linux
 
-For common linux based systems we offer two variants. An `.AppImage` and a `.deb` file.
+#### AppImage
+
+AppImages are a generic way to install software accross most linux distributions
+and systems.
 
 To install a `.AppImage` based release:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,6 +14,12 @@
 <details><summary>Click to expand</summary>
 
 - [Install](#install)
+  - [Linux](#linux)
+    - [AppImage](#appimage)
+    - [Debian/Ubuntu](#debian/ubuntu)
+    - [Arch Linux](#arch linux)
+  - [macOS](#macOS)
+  - [From Source](#from)
```
ralphtheninja: This is automatically generated by `hallmark` and will be overwritten, currently only supports `##` level

cc @vweevers

README.md
```diff
@@ -27,20 +33,50 @@ The application can be downloaded from the [`Releases`](https://github.com/delta
 
 ### Linux
 
-For common linux based systems we offer two variants. An `.AppImage` and a `.deb` file.
+#### AppImage
+
+AppImages are a generic way to install software accross most linux distributions
+and systems.
```
ralphtheninja: :+1:

README.md
vweevers: If we were to add a `maxDepth` option to hallmark (which is a slippery slope), and you set it to 4 here, then note that headings like [`Code Structure`](https://github.com/deltachat/deltachat-desktop#code-structure) would also be included in the TOC.

What's the reason for wanting these subsections of `Install` in the TOC? Perhaps we can somehow translate that reasoning to code.

vweevers: Alternatively, we could make `hallmark` scan + respect the existing TOC and its depth.

ralphtheninja: > What's the reason for wanting these subsections of `Install` in the TOC?

Maybe it's enough to have `##` sections only.

Issue #550, opened 2019-01-06
switch to rc for app config


Simon-Laux:
- [x] switch to rc for app config
- [x] seperate app constants and config/settings
- [x] implement that logger config is also affected by config and by --debug argument

ralphtheninja:
I suspect this issue will be around for a while since I see this as a gradual change.

PR #574, opened 2019-01-10
Add rc

ralphtheninja:
Ticks off first part of https://github.com/deltachat/deltachat-desktop/issues/550

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -27,8 +27,6 @@ process.on('exit', function () {
 
 // Ensure CONFIG_PATH exists.
 mkdirp.sync(config.CONFIG_PATH)
-/* *CONFIG* */
-let argv = sliceArgv(process.argv)
 
 if (config.IS_PRODUCTION) {
   // When Electron is running in production mode (packaged app), then run React
@@ -73,11 +71,8 @@ function onReady (err, results) {
 }
 
 app.once('ipcReady', () => {
-  /* *CONFIG* */
-  log.info(`Command line args: ${argv}`, argv, 'cmd_args')
   console.timeEnd('init')
-
-  var win = windows.main.win
+  const win = windows.main.win
   win.on('close', e => {
     if (!app.isQuitting) {
       e.preventDefault()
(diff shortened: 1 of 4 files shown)


ralphtheninja:
@Simon-Laux Once this has been merged, we can add default configuration for the logger to `src/rc.js`.

src/main/index.js
```diff
@@ -26,18 +27,12 @@ process.on('exit', function () {
 
 // Ensure CONFIG_PATH exists.
 mkdirp.sync(config.CONFIG_PATH)
-/* *CONFIG* */
-let argv = sliceArgv(process.argv)
 
 if (config.IS_PRODUCTION) {
   // When Electron is running in production mode (packaged app), then run React
   // in production mode too.
   process.env.NODE_ENV = 'production'
```
ralphtheninja: @Jikstra I wonder if this is needed or even matters. Do you know if `NODE_ENV` means anything to react when in production?

ralphtheninja: The more we can get rid of small complexities/oddities like this, the better. This code doesn't say anything about _why_ this is important.

src/rc.js
```diff
@@ -0,0 +1 @@
+module.exports = require('rc')('DeltaChat', {})
```
ralphtheninja: @Simon-Laux Just add some properties here with sane defaults for the logger.

src/main/index.js
Jikstra: > @Jikstra I wonder if this is needed or even matters. Do you know if `NODE_ENV` means anything to react when in production?

Would be suprised about that, as far as i know the process global isn't defined in browser js. Let's remove and see, can add it back if we need to. We aren't using NODE_ENV anyway so who cares.

ralphtheninja: What I'm actually getting at is if we can remove `config.IS_PRODUCTION` :)

src/rc.js
Simon-Laux: Ok, I will do that If you didn't already let's see

Simon-Laux: Update: looks like we have to first fix the issue I mentioned here: https://github.com/deltachat/deltachat-desktop/pull/577#issuecomment-453229778

PR #577, opened 2019-01-10
Tweak logger

ralphtheninja:
Ticks off third checkbox in https://github.com/deltachat/deltachat-desktop/issues/550

### DIFF
--- a/src/logger.js
+++ b/src/logger.js
@@ -51,6 +51,7 @@ class Logger {
     if (!OPTIONS.logDebug) return
     log(this.channel, 0, message, errorCode, payload)
   }
+
   /**
      * Log a message on **info** level
      * @param {string} message the message (human readable)
@@ -60,6 +61,7 @@ class Logger {
   info (message, payload, errorCode) {
     log(this.channel, 1, message, errorCode, payload)
   }
+
   /**
      * Log a message on **warning** level
      * @param {string} message the message (human readable)
@@ -67,8 +69,9 @@ class Logger {
      * @param {string} payload (optional) JSON payload
      */
   warn (message, payload, errorCode) {
-    log(this.channel, 1, message, errorCode, payload, getStackTrace())
+    log(this.channel, 2, message, errorCode, payload, getStackTrace())
   }
+
   /**
      * Log a message on **error** level
      * @param {string} message the message (human readable)
@@ -79,6 +82,7 @@ class Logger {
     // TODO add stacktrace to payload
     log(this.channel, 3, message, errorCode, payload, getStackTrace())
   }
+
   /**
      * Log a message on critical level
      * @param {string} message the message (human readable)
--- a/src/logger.js
+++ b/src/logger.js
@@ -1,5 +1,7 @@
 const esp = require('error-stack-parser')
-const rc = require('./rc')
+const { app, remote } = require('electron')
+
+const rc = remote ? remote.app.rc : app.rc
 
 const LoggerVariants = [
   { log: console.debug, prefix: 'DEBUG' },
--- a/src/logger.js
+++ b/src/logger.js
@@ -47,7 +47,7 @@ class Logger {
      * @param {string} errorCode (optional) machine readable error code (string in snake_case)
      * @param {string} payload (optional) JSON payload
      */
-  debug (message, payload = undefined, errorCode = undefined) {
+  debug (message, payload, errorCode) {
     if (!OPTIONS.logDebug) return
     log(this.channel, 0, message, errorCode, payload)
   }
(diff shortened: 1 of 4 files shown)


src/logger.js
```diff
@@ -47,7 +47,7 @@ class Logger {
      * @param {string} errorCode (optional) machine readable error code (string in snake_case)
      * @param {string} payload (optional) JSON payload
      */
-  debug (message, payload = undefined, errorCode = undefined) {
+  debug (message, payload, errorCode) {
```
ralphtheninja: These are not needed since all parameters are `undefined` in js by default

src/logger.js
```diff
@@ -60,15 +61,17 @@ class Logger {
...
+
   /**
      * Log a message on **warning** level
      * @param {string} message the message (human readable)
      * @param {string} errorCode (optional) machine readable error code (string in snake_case)
      * @param {string} payload (optional) JSON payload
      */
   warn (message, payload, errorCode) {
-    log(this.channel, 1, message, errorCode, payload, getStackTrace())
+    log(this.channel, 2, message, errorCode, payload, getStackTrace())
```
ralphtheninja: :bug:

src/logger.js
```diff
@@ -24,7 +32,7 @@ function log (channel, lvl, ...args) {
   handler(timestamp, channel, lvl, ...args)
   if (OPTIONS.alsoLogInLocalConsole) {
     const variant = LoggerVariants[lvl]
-    variant(channel, ...args)
+    variant.log(variant.prefix, channel, ...args)
```
ralphtheninja: This makes it clearer what type of message is shown in the console

Simon-Laux: Yeah that makes sense, because electron's chrome doesn't show the icons that as example Firefox does

src/logger.js
Simon-Laux: glad you found it

src/logger.js
```diff
@@ -30,7 +24,7 @@ function log (channel, lvl, ...args) {
     throw Error('Failed to log message - Handler not initilized yet')
   }
   handler(timestamp, channel, lvl, ...args)
-  if (OPTIONS.alsoLogInLocalConsole) {
+  if (rc['log-to-console']) {
```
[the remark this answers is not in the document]
Simon-Laux: not quite sure on how that is linked to this line of code, do you mean we should add another level check here?

src/logger.js
Simon-Laux: OK

Simon-Laux:
Oh before I forget it: How does the render process get the rc arguments?

Simon-Laux:
```json
{
"_":[],
"sandbox":false,
"service-pipe-token":"[CENSORED]",
"lang":"en-US",
"app-path":"[CENSORED]",
"node-integration":"true",
"webview-tag":"true",
"background-color":"#282828",
"num-raster-threads":4,
"enable-main-frame-before-activation":true,
"enable-compositor-image-animations":true,
"service-request-channel-token":"[CENSORED]",
"renderer-client-id":3,
"shared-files":"v8_context_snapshot_data:100,v8_natives_data:101"
}
```
Is what the renderer gets, none of my original arguments are in there.

Simon-Laux:
I think we'll have to provide a different way for the renderer to access the rc config

src/logger.js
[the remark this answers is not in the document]
ralphtheninja: I commented in wrong location, meant the `debug` check below.

ralphtheninja:
> Is what the renderer gets, none of my original arguments are in there.

I don't know what you mean. I haven't only moved out local arguments `OPTIONS.logDebug` -> `rc.debug` and `OPTIONS.alsoLoginConsole` -> `rc[log-to-console]`.

ralphtheninja:
> I think we'll have to provide a different way for the renderer to access the rc config

Aha I think I know what you're getting at. The render process requires `rc.js` separately and doesn't see any of that state. We should pass config to render process instead from main process.

src/logger.js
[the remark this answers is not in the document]
Simon-Laux: K, makes sense now

Simon-Laux:
> The render process requires `rc.js` separately and doesn't see any of the state from the main process? (E.g. `--log-debug` etc)

exactly.

ralphtheninja:
@Simon-Laux I'll think of something.

ralphtheninja:
And the tests are crying of course. So close.

src/rc.js
```diff
@@ -1 +1,4 @@
-module.exports = require('rc')('DeltaChat', {})
+module.exports = require('rc')('DeltaChat', {
+  'log-debug': false,
+  'log-to-console': false
+})
```
ralphtheninja: @Simon-Laux This is technically not needed since they are falsey it will not matter, but it shows which properties you _can_ configure.

ralphtheninja:
@Simon-Laux Good to merge?

src/rc.js
```diff
@@ -1,4 +1,4 @@
-module.exports = require('rc')('DeltaChat', {
+module.exports = Object.freeze(require('rc')('DeltaChat', {
   'log-debug': false,
```
ralphtheninja: cc @Jikstra

ralphtheninja:
Will rebase on latest master.

PR #578, opened 2019-01-10
Split config

ralphtheninja:
Ticks off final checkbox in https://github.com/deltachat/deltachat-desktop/issues/550


Simon-Laux:
When we're already at it, should we add height and width arguments to rc to add like a "--window-imensions" or "--win-dim [widht]x[height]" argument? It might be useful for some users

src/main/windows/main.js
```diff
@@ -176,12 +180,6 @@ function toggleDevTools () {
   }
 }
 
-function getIconPath () {
-  return process.platform === 'win32'
-    ? config.APP_ICON + '.ico'
```
ralphtheninja: Broken. Lets fix this when we need .ico file for windows.

src/renderer/lib/state.js
```diff
@@ -1,14 +1,10 @@
 const appConfig = require('application-config')('DeltaChat')
 const path = require('path')
 const { EventEmitter } = require('events')
-/* *CONFIG* */
-const config = require('../../config')
 const log = require('../../logger').getLogger('renderer/state')
 
 const SAVE_DEBOUNCE_INTERVAL = 1000
 
-appConfig.filePath = path.join(config.CONFIG_PATH, 'config.json')
```
ralphtheninja: Not needed, `appConfig.filePath` defaults to `/path/to/config.json`

ralphtheninja:
> When we're already at it, should we add height and width arguments to rc to add like a `--window-imensions` or `--win-dim [widht]x[height]` argument? It might be useful for some users

Basic rule of thumb: Never add something that might be useful. We should add default values to `rc` though for the log stuff. Default is `undefined` at the moment and it's good practice to set e.g. `false` instead.

src/renderer/components/dialogs/About.js
```diff
@@ -32,10 +35,10 @@ class About extends React.Component {
         canOutsideClickClose={false}>
         <div className={Classes.DIALOG_BODY}>
           <Unselectable>
-            <p style={{ color: 'grey' }}>{`Version ${APP_VERSION}`}</p>
+            <p style={{ color: 'grey' }}>{`Version ${appVersion()}`}</p>
             <p>Official Delta Chat Desktop app.</p>
-            <p>This software is licensed under <ClickableLink href='https://github.com/deltachat/deltachat-desktop/blob/master/LICENSE' text='GNU GPL version 3' />.</p>
-            <p>Source code is available on <ClickableLink href='https://github.com/deltachat/deltachat-desktop' text='GitHub' />.</p>
+            <p>This software is licensed under <ClickableLink href={gitHubLicenseUrl()} text='GNU GPL version 3' />.</p>
+            <p>Source code is available on <ClickableLink href={gitHubUrl()} text='GitHub' />.</p>
```
ralphtheninja: Got rid of some hardcoded strings yay! :tada:

test/integration/setup.js
```diff
@@ -7,8 +7,8 @@ const PNG = require('pngjs').PNG
 const tempy = require('tempy')
 const mkdirp = require('mkdirp')
 
-const config = require('../../src/config')
-mkdirp.sync(config.CONFIG_PATH)
+const { getConfigPath } = require('../../src/application-constants')
+mkdirp.sync(getConfigPath())
```
ralphtheninja: Not sure this code is needed either.

src/main/index.js
```diff
@@ -45,7 +43,7 @@ function onReady (err, results) {
   const state = results.state
   app.logins = results.logins
 
-  var cwd = process.env.TEST_DIR || config.CONFIG_PATH
+  var cwd = process.env.TEST_DIR || getConfigPath()
   log.info('cwd', cwd, 'cwd')
```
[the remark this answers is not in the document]
ralphtheninja: Done.

test/integration/setup.js
ralphtheninja: Done.

test/integration/setup.js
```diff
@@ -4,11 +4,6 @@ const electronPath = require('electron')
 const fs = require('fs')
 const path = require('path')
 const PNG = require('pngjs').PNG
-const tempy = require('tempy')
-const mkdirp = require('mkdirp')
-
-const { getConfigPath } = require('../../src/application-constants')
-mkdirp.sync(getConfigPath())
```
ralphtheninja: `tempy` takes care of this, not needed

Issue #576, opened 2019-01-10
Cleanup src/renderer/lib/state.js


ralphtheninja:
I noticed something when making a console.log in `src/config.js` and `src/renderer/lib/state.js`. Both were logged in both the main and the render processes. I expected only `src/config.js` to be logged by both processes.

Turns out that `src/renderer/lib/state.js` is required by both processes, which is slightly odd.

* main process does `State.load()`
* render process does `State.load()`, `State.save()` and `State.saveImmediate()`

Also, `State` is an `EventEmitter` that we have two instances of, which both can do `.load()` etc. There's also some ipc going on between the processes (which is fine) but it's unclear who owns this state. I think we should move it to either process, most likely the main process, and whenever the render process needs to save state, it should do ipc to the main process.

ralphtheninja:
I've also noticed some inconsistencies when saving state. It might be related to this. Especially when you have no config set and the default is used, we should save the default config immediately.

PR #583, opened 2019-01-13
State refactor

ralphtheninja:
Closes https://github.com/deltachat/deltachat-desktop/issues/576


src/main/ipc.js
```diff
@@ -167,9 +155,14 @@ function init (cwd, state) {
...
-    const json = dc.render()
-    windows.main.setTitle(json.credentials.addr)
-    windows.main.send('render', json)
+    const deltachat = dc.render()
+    windows.main.setTitle(deltachat.credentials.addr)
+    sendState(deltachat)
+  }
+
+  function sendState (deltachat) {
+    Object.assign(state, { deltachat })
```
ralphtheninja: Here we pass _all_ state to render process and not only the `deltachat` data.

PR #588, opened 2019-01-15
Misc cleanup

ralphtheninja:


src/main/ipc.js
```diff
@@ -16,122 +16,110 @@ const setupNotifications = require('./notifications')
...
     app.emit('ipcReady')
   })
 
-  ipc.on('setAspectRatio', (e, ...args) => main.setAspectRatio(...args))
-  ipc.on('setBounds', (e, ...args) => main.setBounds(...args))
-  ipc.on('setProgress', (e, ...args) => main.setProgress(...args))
-  ipc.on('show', () => main.show())
-  ipc.on('setAllowNav', (e, ...args) => menu.setAllowNav(...args))
-  ipc.on('chooseLanguage', (e, locale) => {
+  ipcMain.on('setAspectRatio', (e, ...args) => main.setAspectRatio(...args))
```
ralphtheninja: It's easier to grep for `ipcMain` to see which ipc event listeners we have in the main process

Issue #258, opened 2018-11-01
Auto login on startup for last logged in account


Jikstra:
Auto login into the account which was selected the last time the app was used.

ralphtheninja:
We should merge https://github.com/deltachat/deltachat-desktop/pull/583 before doing this

PR #589, opened 2019-01-16
Login automatically if last credentials were saved in state

ralphtheninja:
Closes https://github.com/deltachat/deltachat-desktop/issues/258

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -62,7 +62,6 @@ function onReady (err, results) {
   menu.init()
 
   if (rc.debug) windows.main.toggleDevTools()
-
 }
 
 app.once('ipcReady', () => {
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -27,10 +27,10 @@ process.on('exit', function () {
 
 // Report uncaught exceptions
 process.on('uncaughtException', (err) => {
-  console.error(err)
   const error = { message: err.message, stack: err.stack }
   windows.main.send('uncaughtError', 'main', error)
   log.error('uncaughtError', error, 'uncaught_error')
+  throw err
 })
 
 mkdirp.sync(getConfigPath())
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -44,11 +44,12 @@ function onReady (err, results) {
 
   app.saveState = () => State.save({ saved: state.saved })
 
+  localize.setup(app, state.saved.locale || app.getLocale())
+
   const cwd = getConfigPath()
   log.info('cwd', cwd, 'cwd')
   ipc.init(cwd, state)
 
-  localize.setup(app, state.saved.locale || app.getLocale())
   windows.main.init(app, { hidden: false })
   menu.init()
 
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -25,6 +25,14 @@ process.on('exit', function () {
   logHandler.closeWriteStream()
 })
 
+// Report uncaught exceptions
+process.on('uncaughtException', (err) => {
+  console.error(err)
+  const error = { message: err.message, stack: err.stack }
+  windows.main.send('uncaughtError', 'main', error)
+  log.error('uncaughtError', error, 'uncaught_error')
+})
+
 mkdirp.sync(getConfigPath())
 
 app.ipcReady = false
@@ -55,13 +63,6 @@ function onReady (err, results) {
 
   if (rc.debug) windows.main.toggleDevTools()
 
-  // Report uncaught exceptions
-  process.on('uncaughtException', (err) => {
-    console.error(err)
-    const error = { message: err.message, stack: err.stack }
-    windows.main.send('uncaughtError', 'main', error)
-    log.error('uncaughtError', error, 'uncaught_error')
-  })
 }
 
 app.once('ipcReady', () => {
--- a/src/main/ipc.js
+++ b/src/main/ipc.js
@@ -19,10 +19,12 @@ function init (cwd, state) {
   const main = windows.main
   const dc = new DeltaChat(cwd, state.saved)
 
-  dc.on('ready', () => {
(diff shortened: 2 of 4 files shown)


src/main/index.js
```diff
@@ -44,11 +44,12 @@ function onReady (err, results) {
 
   app.saveState = () => State.save({ saved: state.saved })
 
+  localize.setup(app, state.saved.locale || app.getLocale())
```
ralphtheninja: This needs to be done before calling `txCoreStrings()` passed to `dc.login()`

Jikstra:
I'm getting this error:
I had to move the `process.on('uncaughtExecption'...` directly after the setup of the logger otherwise it wouldn't get caught and only a weird popup from electron was there with the error message but no way to actually get it. Will push this change to this branch. 
```
TypeError [ERR_INVALID_ARG_TYPE]: The first argument must be one of type string, Buffer, ArrayBuffer, Array, or Array-like Object. Received type undefined                                                                                                                       
    at Function.from (buffer.js:207:11)
    at DeltaChatController.getPath (/home/kerle/Workspace/deltachat-desktop/src/main/deltachat.js:29:39)
    at DeltaChatController.login (/home/kerle/Workspace/deltachat-desktop/src/main/deltachat.js:41:22)
    at Object.init (/home/kerle/Workspace/deltachat-desktop/src/main/ipc.js:165:8)
    at onReady (/home/kerle/Workspace/deltachat-desktop/src/main/index.js:52:9)
    at end (/home/kerle/Workspace/deltachat-desktop/node_modules/run-parallel/index.js:18:15)
    at done (/home/kerle/Workspace/deltachat-desktop/node_modules/run-parallel/index.js:22:10)
    at each (/home/kerle/Workspace/deltachat-desktop/node_modules/run-parallel/index.js:28:7)
    at /home/kerle/Workspace/deltachat-desktop/node_modules/run-parallel/index.js:38:43
    at series (/home/kerle/Workspace/deltachat-desktop/src/main/logins.js:21:7)
ERROR main/index uncaughtError uncaught_error { message:
   'The first argument must be one of type string, Buffer, ArrayBuffer, Array, or Array-like Object. Received type undefined',
  stack:
   'TypeError [ERR_INVALID_ARG_TYPE]: The first argument must be one of type string, Buffer, ArrayBuffer, Array, or Array-like Object. Received type undefined\n    at Function.from (buffer.js:207:11)\n    at DeltaChatController.getPath (/home/kerle/Workspace/deltachat-desktop/src/main/deltachat.js:29:39)\n    at DeltaChatController.login (/home/kerle/Workspace/deltachat-desktop/src/main/deltachat.js:41:22)\n    at Object.init (/home/kerle/Workspace/deltachat-desktop/src/main/ipc.js:165:8)\n    at onReady (/home/kerle/Workspace/deltachat-desktop/src/main/index.js:52:9)\n    at end (/home/kerle/Workspace/deltachat-desktop/node_modules/run-parallel/index.js:18:15)\n    at done (/home/kerle/Workspace/deltachat-desktop/node_modules/run-parallel/index.js:22:10)\n    at each (/home/kerle/Workspace/deltachat-desktop/node_modules/run-parallel/index.js:28:7)\n    at /home/kerle/Workspace/deltachat-desktop/node_modules/run-parallel/index.js:38:43\n    at series (/home/kerle/Workspace/deltachat-desktop/src/main/logins.js:21:7)' } [ { columnNumber: 9,                                         
    lineNumber: 129,
    fileName: '/home/kerle/Workspace/deltachat-desktop/src/main/index.js',
    functionName: 'process.on',
    source:
     '    at process.on (/home/kerle/Workspace/deltachat-desktop/src/main/index.js:129:9)' },
  { columnNumber: 15,
    lineNumber: 187,
    fileName: 'events.js',
    functionName: 'process.emit',
    source: '    at process.emit (events.js:187:15)' },
  { columnNumber: 27,
    lineNumber: 451,
    fileName: 'internal/bootstrap/node.js',
    functionName: 'process._fatalException',
    source:
     '    at process._fatalException (internal/bootstrap/node.js:451:27)' } ]
```

ralphtheninja:
@Jikstra Maybe this is the error we get on jenkins. I don't get this on my machine. I wonder why.

Jikstra:
@ralphtheninja not sure, just pushed the change to this branch so we now should get it. I think it's because of the getPath method. The path argument is null/undefined but we try to do some Buffer.onSomething method on it. Can investigate it further.

ralphtheninja:
I think we should throw an error. If we have an uncaught error we don't want the process to keep hang.

src/main/index.js
```diff
@@ -27,10 +27,10 @@ process.on('exit', function () {
 
 // Report uncaught exceptions
 process.on('uncaughtException', (err) => {
-  console.error(err)
   const error = { message: err.message, stack: err.stack }
   windows.main.send('uncaughtError', 'main', error)
```
ralphtheninja: Actually, I'm not sure we should even do this. Sending events to the window can cause other uncaught events to happen. I think we should just remove the uncaught handler completely.

src/main/ipc.js
```diff
@@ -161,8 +161,13 @@ function init (cwd, state) {
     main.send('render', state)
   }
 
-  if (state.saved.credentials) {
```
ralphtheninja: But how the ... fucking javascript

src/main/index.js
Jikstra: I'm fine with removing the `windows.main.send()` stuff. But not sure what it does.

src/main/ipc.js
Jikstra: Problem is that state.saved.credentials is sometimes `null`, sometimes `{}`. We need to make sure that it's not null, it's a object and it's not empty. Maybe this can be done easier by preventing it from getting an empty object so that we only have to check if it's not null.

PR #592, opened 2019-01-16
[wip] Enable linux on Travis again

ralphtheninja:


ralphtheninja:
Just playing around a bit. I wonder why it wants to try building the node bindings. It should work :/

ralphtheninja:
Aaah it maybe fails because we're running on an ancient Ubuntu (14.04).

Issue #597, opened 2019-01-18
Integration tests try to auto login


ralphtheninja:
When running integration tests, they will try to auto login using data from e.g. `~/.config/DeltaChat/config.json` even though we're using a temporary app folder. We never see this on CI since nothing has made a real login before. Only affects local integration tests.

Issue #596, opened 2019-01-18
getConfigPath() bugs during test


ralphtheninja:
Since it's using `tempy` different folder will be returned each time. Only affects integration tests.

PR #598, opened 2019-01-18
Wrap application-config module for proper appConfig during production and testing

ralphtheninja:
Closes https://github.com/deltachat/deltachat-desktop/issues/597, closes https://github.com/deltachat/deltachat-desktop/issues/596

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -18,6 +18,8 @@ const windows = require('./windows')
 const logHandler = require('./developerTools/logHandler')
 const log = require('../logger').getLogger('main/index')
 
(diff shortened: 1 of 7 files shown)


src/main/index.js
```diff
@@ -18,6 +18,8 @@ const windows = require('./windows')
 const logHandler = require('./developerTools/logHandler')
 const log = require('../logger').getLogger('main/index')
 
+mkdirp.sync(getConfigPath())
```
ralphtheninja: This should be done as early as possible

PR #601, opened 2019-01-18
Improve app size by bundle backend dependencies with webpack and better fronend bundling

### LINKED ISSUE #594: Use webpack for electron-main too
Currently we only bundle/minify the code for the electron renderer process (read react) but not for the backend. This should be possible and reduce build size a lot. 
@Frando knows stuff about this, but can't find the repo anymore with the webpack.electron.config.js .

Jikstra:
Current problems:
- [x] the dynamic expression for language imports is making trouble
```
Error: JSON parse error in language file '/home/kerle/Workspace/_locales/en.json'
    at getLocaleMessages (webpack-internal:///./src/localize.js:90:11)
    at Object.setup (webpack-internal:///./src/localize.js:51:19)
    at onReady (webpack-internal:///./src/main/index.js:54:12)
    at end (/home/kerle/Workspace/deltachat-desktop/node_modules/run-parallel/index.js:18:15)
    at done (/home/kerle/Workspace/deltachat-desktop/node_modules/run-parallel/index.js:22:10)
    at each (/home/kerle/Workspace/deltachat-desktop/node_modules/run-parallel/index.js:28:7)
    at /home/kerle/Workspace/deltachat-desktop/node_modules/run-parallel/index.js:38:43
    at series (webpack-internal:///./src/main/logins.js:21:7)
    at end (/home/kerle/Workspace/deltachat-desktop/node_modules/run-series/index.js:10:15)
    at done (/home/kerle/Workspace/deltachat-desktop/node_modules/run-series/index.js:13:10)
```
- [x] we don't get the right path for the `static/main.html`. This is because `__dirname` isn't a absolute path currently.
```
Not allowed to load local resource: file://static/main.html
```

- [x] We should put all build files into a `/dist` folder or something, including copying the static/_locales..., stuff there.

- [x] `npm start` should acutally use the built `index.bundled.js`

- [x] we need to somehow only put deltachat-node to the node_modules folder in a electron pack and ignore everything else

- [x] the deltachat-node dependency is quite big and has useless files in it, we should remove them before packaging.

- [x] Adjust watch scripts, maybe even use webpack-watch

How to test:
1. `npm run build-backend # there are also build-frontend and build-*end:prod scripts `
2. `npm start`

Will fix #594 at some day

#### This reduces the app.asar size from `55.8mb` to `25mb`. Suprisingly this has no real impact on the download size (only `2mb` less) which still is around 60mb. 

@ralphtheninja do you think it's worth it? Having a 20mb less app.asar file? This could greatly improve the install size on system where we don't have to bundle electron, but currently it doesn't make a big difference.


Jikstra:
So what i did:
- We now output all files created/needed for running the app to `./build`
- electron builder now consumes everything only from `./build`
- added a script which generates a second `package.json` which is only for electron-builder so that it knows to online install the `deltachat-node` dependency, all other dependencies get bundled with webpack.
- added more/new build steps which are now seperated into frontend/backend
- for frontend & backend all dependencies are getting bundled, this makes the initial app load slower, but this can get improved by splitting this again (that's the main idea of webpack)
- for backend we even do tree shaking

ralphtheninja:
Where to start?

> I'm stripping out the prebuilds as they aren't really platform specific (at least i found prebuilds for all platforms) and i think it makes more sense on packaging/distributing to rebuild deltachat-node for the electron build.

Why do you think this makes more sense? This means we need to add more complexity to `desktop` so Travis and Jenkins can build the binaries, which the node bindings already know how to do. Essentially duplicating that logic and adding more maintenance work.

It would be better to strip out the binaries not needed.

> It's a bit more build complexity

Four webpack configuration files and nine build scripts in package.json is a lot more build complexity. Also, `npm run dist` is broken, which is the important part here, `npm run pack` is never used.

> We have different setups for development/production which could introduce bugs that we only can reproduce in production

This is a big no no for me and adds unnecessary risk. Running in "development" should be identical to running the installed application.

I think you have added more complexity which has resulted in more hackish solutions, such as rewriting `package.json` for really no big improvements. Come to think of it, I don't even know why we should/need to package the backend code in the first place. Can't you tell webpack is pushing back at you? You're also possibly trying to solve too many things at the same time.

Jikstra:
@ralphtheninja i totally see your points, that also the reason why i have so many Cons written down. Goal of this pr is more or less understanding how webpack works (I think I got it by now :D) and what possibilities we have to reduce the app size. We don't have to take all of those paths, we should cherry pick the ones that actually make sense for us. Soo the current setup can and should get simplified/adjusted.

> Why do you think this makes more sense? This means we need to add more complexity to desktop so Travis and Jenkins can build the binaries, which the node bindings already know how to do. Essentially duplicating that logic and adding more maintenance work.

I do this for 2 reasons:
- `deltachat-node` install always falls back to recompiling `deltachat-core` on my arch machine ^^ (other issue)
- electron is another runtime than node. It uses another node version and because of this problems can arise. Recompiling native modules for electron can make sense
- I just didn't take care of this too much, i only wanted to explore webpack and see how far we can go with reducing the build size. And currently we ship some megabytes of useless binaries to the different platforms because we include build & prebuilds for different platforms.

> It would be better to strip out the binaries not needed.

Stripping out the not needed binaries makes sense, but that's not too easy to figure out which binaries are actually used if we want to take prebuilds into account. Always recompiling them for the actual used electron version makes this build step easier.

> Four webpack configuration files and nine build scripts in package.json is a lot more build complexity. 

Yup it is. But the current approach in this PR is to get the most out of all we have, it makes sense to take some steps back and see again. 

> Also, npm run dist is broken, which is the important part here, npm run pack is never used.

Shouldn't be broken, at least not for me. You first need to run `npm run pack` and then `npm run dist`, that should work. But i didn't introduce this behaviour, it was like that before.

> This is a big no no for me and adds unnecessary risk. Running in "development" should be identical to running the installed application.

Yup i think the same, especially in our current way of releasing/testing/finding bugs during development.

> I think you have added more complexity which has resulted in more hackish solutions, such as rewriting package.json for really no big improvements.

Sadly electron-builder really likes the idea of having 2 package-json files where they have to be quite similiar but the second one isn't allowed to have the build section (....). I guess this step can get removed/simplified.

> Can't you tell webpack is pushing back at you?

Sure it does :D Same as building shit on mac and trying to fix windows ^^ But I think it was worth it as webpack no longer is that big black box for me as it was before. Same for building on mac. Hopefully same for building on windows.

> You're also possibly trying to solve too many things at the same time.

Yup very true, but it was just possible without too much hassle when the initial setup was ready.

> Lets take a step back and try to see what's the really big problem(s) here and lets break that down and solve them one by one.

Very good idea. I think we can also postpone this. Definetly not something for any january releases as this might break stuff.

ralphtheninja:
> Shouldn't be broken, at least not for me. You first need to run `npm run pack` and then `npm run dist`, that should work. But i didn't introduce this behaviour, it was like that before.

No, previously it was enough to do `npm run build` and then `npm run dist`, so this behavior has changed.

In general, I'm fine with adding complexity if it's really worth it. I think we should try to _reduce_ complexity as much as possible and I think we're more or less on the same page there.

ralphtheninja:
>     * electron is another runtime than node. It uses another node version and because of this problems can arise. Recompiling native modules for electron can make sense

It can, I'm just arguing it's not needed :)

ralphtheninja:
Made an issue here https://github.com/electron-userland/electron-builder/issues/3621

Jikstra:
> Correct me if I'm wrong, but it seems to me the biggest culprit here is electron@4. Just by switching electron version the AppImage grew from 68Mb to 90Mb.

Yes it is. But we can't do much on this. Besides that electron is more like java/openjdk, if the size grows, you can't do a lot. But in a better world, electron is getting provided seperately by the distrubutions repositories and we only have to pack our dependency and require electron4. This will be possible on some distributions (eg. arch as soon as electron4 is in the main repos). And then reducing the app size by 50% is a good improvement.

I will come back to this pr at some day, not sure how to proceed exactly.

PR #605, opened 2019-01-20
change language dropdown to use the local names

### LINKED ISSUE #516: Use only the languages own language translation
Currently we translate every language name to every other language which results in having german translated to "Deutsch, German, Allemand...". This doesn't help anyone to find the right language, as every german will recognize the "Deutsch" language string. This is also the behaviour the android version has.

Simon-Laux:
closes #516

### DIFF
--- a/bin/extract_local_language_names.js
+++ b/bin/extract_local_language_names.js
@@ -1,10 +1,10 @@
-const { readdirSync, readFileSync, writeFileSync } = require('fs')
+const { readdirSync, writeFileSync } = require('fs')
 const { join } = require('path')
 
 const localeDir = join(__dirname, '..', '_locales')
 
 // English Lang as fallback
-const englishLang = JSON.parse(readFileSync(join(localeDir, 'en.json')))
+const englishLang = require(join(localeDir, 'en.json'))
 
 function getAvailableLanguages () {
   return readdirSync(localeDir)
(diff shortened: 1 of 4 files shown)


bin/extract_local_language_names.js
```diff
@@ -0,0 +1,25 @@
+const { readdirSync, readFileSync, writeFileSync } = require('fs')
+const { join } = require('path')
+
+const localeDir = join(__dirname, '..', '_locales')
+
+// English Lang as fallback
+const englishLang = JSON.parse(readFileSync(join(localeDir, 'en.json')))
```
ralphtheninja: Note that you can just as well do

```js
const englishLang = require(join(localeDir, 'en.json'))
```

bin/extract_local_language_names.js
```diff
@@ -0,0 +1,25 @@
...
+
+function getAvailableLanguages () {
+  return readdirSync(localeDir)
+    .filter(l => {
+      return !l.startsWith('_') && l.endsWith('.json')
+    })
+    .map(l => {
+      const locale = l.split('.json')[0]
+      const content = JSON.parse(readFileSync(join(localeDir, l)))
```
ralphtheninja: And here.

bin/extract_local_language_names.js
Jikstra: We should prefer to use require as webpack has good support for it. Sure on a require you could somehow inline js code which gets executed but this shouldn't be a problem.

Simon-Laux: I didn't thought of that, thanks for pointing it out to me.

ralphtheninja:
@Simon-Laux I can solve the conflict if you like.

Simon-Laux:
> @Simon-Laux I can solve the conflict if you like.

Go ahead, thanks.

Simon-Laux:
I confirm that it works for me

PR #608, opened 2019-01-20
Implement emoji picker 

Jikstra:
Still needs:
- [x] Some css/js to show/hide the picker above the other elements
- [x] Actually input the emoji into the chat input
- [x] inclue the emoji packs somehow/use the right ones, currently we load them from unpkg (not good)
- [x] use onMouseOver above the emoji icon to show the picker, when it's outside of picker/button hide it again
- [x] Use better emoji icon/readjust icons in composer

Fixes https://github.com/deltachat/deltachat-desktop/issues/475

screenshot_20190123_004501

PR #614, opened 2019-01-24
Emoji picker

ralphtheninja:
Supersedes https://github.com/deltachat/deltachat-desktop/pull/608


src/logger.js
```diff
@@ -35,32 +36,72 @@ function getStackTrace () {
...
+    // TODO add stacktrace to payload
+    log(this.channel, 4, message, errorCode, payload, getStackTrace())
   }
 }
 
+/**
+ * Creates a new Logger
+ * @param {string} channel The part/module where the message was logged. Like 'Tanslations'
+ * @returns {Logger}
+*/
```
ralphtheninja: Why are these changes part of the emoji picker?

PR #617, opened 2019-01-24
fix emailadress parsing bug in message

### LINKED ISSUE #613: mailaddresses in messages cause doubling of message text
current master: if you send a message with an email address, say: "email@email.org blaaaaa" then the rendering of the message on the sending side will be "email@email.org blaaaaa blaaaa".   Android shows the message correctly so this is quite definitively a desktop processing/rendering problem.

Simon-Laux:
fixes #613

### DIFF
--- a/src/renderer/components/conversations/Linkify.js
+++ b/src/renderer/components/conversations/Linkify.js
@@ -37,7 +37,7 @@ class Linkify extends React.Component {
           </a>
         )
       } else {
-        results.push(renderNonLink({ text, key: count++ }))
+        results.push(renderNonLink({ text: originalText, key: count++ }))
       }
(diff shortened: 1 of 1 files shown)


src/renderer/components/conversations/Linkify.js
```diff
@@ -37,7 +37,7 @@ class Linkify extends React.Component {
           </a>
         )
       } else {
-        results.push(renderNonLink({ text, key: count++ }))
+        results.push(renderNonLink({ text: originalText, key: count++ }))
```
ralphtheninja: :+1: 

@Simon-Laux It's fine to not review everything. Good call.

PR #619, opened 2019-01-25
Selected chat now uses the delta (light) color

Jikstra:
Before:
screenshot_20190125_020543

After:
screenshot_20190125_020401

After with light color:
screenshot_20190125_151235

Opinions? @r10s @hpk42 @Simon-Laux @ralphtheninja

### DIFF
--- a/src/renderer/components/ChatList.js
+++ b/src/renderer/components/ChatList.js
@@ -26,6 +26,19 @@ const ChatListWrapper = styled.div`
 
   .module-conversation-list-item--is-selected {
     background-color: ${StyleVariables.colors.deltaSelected};
+    color: ${StyleVariables.colors.deltaPrimaryFg};
+    
+    span.module-contact-name {
+      color: ${StyleVariables.colors.deltaPrimaryFg};
+    }
+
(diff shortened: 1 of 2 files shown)


violoncelloCH:
I would go for a slightly lighter version of the delta chat colour...
I think header and selected item shouldn't have the exact same colour, tough using something that fits into the delta colour scheme is a great idea!
Do we have a lighter "version" of the delta colour? Otherwise it would probably be good to define such a colour globally...

ralphtheninja:
> I would go for a slightly lighter version of the delta chat colour...

:+1:

r10s:
in general, this looks good to me. much better than before - in fact, i was sometimes a bit confused if the mouse rests about an item so that the hovered item looks similar to the selected one.

for a _slightly_ lighter color we could go for #4c6e7d and use delta-standard-green-#415e6b

Jikstra:
> I think header and selected item shouldn't have the exact same colour

@violoncelloCH very good point, thanks! In general i want to get rid of the colored navbar/header anyway, but going for a slightly lighter delta color is a good idea :)

Implemented the light color and added a new screenshot with the light color.

Issue #523, opened 2019-01-03
Mark verified groups as experimental


Jikstra:
Verified groups is currently a experimental feature which might change in the future. We should indicate to the user that this can change in the current form.

Jikstra:
This should be handled in the shared translations, not something for us to care about.

ralphtheninja:
Cool. Lets leave it open until we have pulled in the updated translations.

ralphtheninja:
So, who is making sure this happens? :)

Jikstra:
Good question, i don't know how we can roll out a "replacement" of a already translated string. Maybe we need to introduce a new string and push it to tx? If so, we can just do that.

r10s:
i think if we pick up the strings from the current strings.xml, it should be fine.

for desktop, i've removed the "verified" wording also from `show_qr_verify_code_desktop` and from otherwise deprecated strings.

ralphtheninja:
When I try pulling the strings I don't get an update of `en.xml`.

hpk42:
It's also a layout thing -- the "verified group" should be right-most and renamed to "Verified group (experimental)" or so, maybe with a little extra spacing? 
![vgroup](https://user-images.githubusercontent.com/73579/51503269-64a32400-1dda-11e9-8e84-ffe8bbde6626.png)

PR #620, opened 2019-01-25
Move create verified group button to the right and add some margin

Jikstra:
this still misses:
- [x] pulling in new translations which add the (experimental) thing

I don't have the transiflex client installed and don't know shit about this, @ralphtheninja can you help me out with pulling in the translations?

Will fix https://github.com/deltachat/deltachat-desktop/issues/523

### DIFF
--- a/_locales/en.xml
+++ b/_locales/en.xml
@@ -12,6 +12,7 @@
     <string name="custom">Custom</string>
     <string name="none">None</string>
     <string name="open">Open</string>
+    <string name="delete">Delete</string>
     <string name="not_now">Not now</string>
     <string name="never">Never</string>
     <string name="one_moment">One moment…</string>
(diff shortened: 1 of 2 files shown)


ralphtheninja:
As I said here https://github.com/deltachat/deltachat-desktop/issues/523#issuecomment-455183323 I don't get an update of `en.xml` when doing `tx pull --all`.

Jikstra:
@r10s any ideas why we don't get the new translations?

r10s:
not sure what exactly `--all` does, on android, we're using `tx pull -l en`, see script at https://github.com/deltachat/deltachat-android/blob/master/tools/tx-pull-source.sh

ralphtheninja:
> not sure what exactly `--all` does, on android, we're using `tx pull -l en`, see script at https://github.com/deltachat/deltachat-android/blob/master/tools/tx-pull-source.sh

That did the trick!

PR #623, opened 2019-01-26
Appdata updates and cleanup

har9862:
+ Changed the metadata_license identifier from CC0 > CC0-1.0 to comply with [SPDX](https://spdx.org/licenses/).
+ Added url for bugtracker, faq, donation and translate.
+ Updated with regenerated [OARS](https://hughsie.github.io/oars/generate.html)-data. Content attributes with the value of 'none' are no longer necessary to be explicitly specified.
+ General code cleanup. Release description should not look messy on Flathub etc anymore.

### DIFF
--- a/static/chat.delta.desktop.appdata.xml
+++ b/static/chat.delta.desktop.appdata.xml
@@ -1,18 +1,15 @@
 <!--<?xml version="1.0" encoding="UTF-8"?>-->
 <component type="desktop">
   <id>chat.delta.desktop</id>
-  <metadata_license>CC0</metadata_license>
+  <launchable type="desktop-id">chat.delta.desktop</launchable>
+  <metadata_license>CC0-1.0</metadata_license>
   <project_license>GPL-3.0-or-later</project_license>
   <name>Delta Chat</name>
   <summary>Delta Chat email-based messenger</summary>
-  
+
   <description>
-    <p>The messenger with the broadest audience in the world.
-Free, independent, email compatible.
-    </p>
+    <p>The messenger with the broadest audience in the world. Free, independent, email compatible.</p>
   </description>
-  
-​  <launchable type="desktop-id">chat.delta.desktop</launchable>
 
   <screenshots>
     <screenshot type="default">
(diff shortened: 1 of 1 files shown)


static/chat.delta.desktop.appdata.xml
```diff
@@ -1,173 +1,160 @@
...
-            Clean up unused configuration code (@ralphtheninja)
-
-        Fixed
-
-            Fix non verified contacts in verified groups (@ralphtheninja)
-            Fix escaped characters in translations (@ralphtheninja)
-            Adjust login form so it's not hidden below navigation bar (@jikstra)
-            Fix broken rimraf dependency (@jikstra)
-        </p>
-        </description>
```
Jikstra: I'm unsure if it makes sense to have the changelog in appdata.xml as long as we don't display it somewhere. And i think i never saw it in the desktop yet, maybe it's somewhere hidden ^^ @har9862 do you know more about this and what it's goo for?

Jikstra:
Thanks for your PR :) Would like to have a short review by @ralphtheninja

ralphtheninja:
We should consider automating this. Would it be ok to just link to the changelog entry on github?

har9862:
A release version and date is probably wanted or else it will be empty on places making use of appdata (such as on Flathub, where it's a [requirement](https://github.com/flathub/flathub/wiki/App-Requirements#appstream)), but using a url instead of a long description should be fine.

For example:
```
<releases>
  <release version="3.2" date="2019-02-08">
    <url>https://www.example.com/changelog-3.2</url>
  </release>
  <release version="3.1" date="2018-09-05">
    <url>https://www.example.com/changelog-3.1</url>
  </release>
</releases>
```
The url can also be used together with a summarized release description if wanted, this is what for example [Gimp](https://github.com/GNOME/gimp/blob/master/desktop/org.gimp.GIMP.appdata.xml.in.in) do who usually has a very detailed and long description for every release.

Please see the specs for more info:
https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-releases

muelli:
> Updated with regenerated [OARS](https://hughsie.github.io/oars/generate.html)-data. Content attributes with the value of 'none' are no longer necessary to be explicitly specified.

but then semantic is lost, no? Or are missing tags interpreted as "none" rather than "unknown" or equivalent?

Issue #495, opened 2018-12-28
Make it impossible to start deltachat-desktop multiple times


Jikstra:
Currently its possible to have more than one instance of deltachat running, we should block that to make multi writes to db impossible. Also normally you dont want that.

PR #649, opened 2019-01-30
Prevent running multiple instances

ralphtheninja:
Closes https://github.com/deltachat/deltachat-desktop/issues/495

* Tries to get a lock and quits if it fails
* The process with the lock focuses the main window


Jikstra:
Very nice!

Issue #543, opened 2019-01-06
let input field grow up to 5 lines or more


webratte:
The input field should grow up if I enter a new line.
Maybe up to 5 lines or more if you think there is space enough (in android client it's currently 3 lines).

ralphtheninja:
The current design is quite a mess to be honest. Before growing this thing we should consider rewriting the current page layout.

At the moment if you increase the height of the text area the whole page scrolls:

![screenshot from 2019-01-09 17-59-16](https://user-images.githubusercontent.com/308049/50915159-7975d400-1438-11e9-9105-6e7c17db2dfb.png)

I think the chat view to the right should have its own scrollbar. The entire area should never scroll. The textarea at the bottom of the chat view should just increase in size and the chat area adapts itself to whatever height the text area has.

I think `Keybase` solves this pretty nicely:

![screenshot from 2019-01-09 18-03-17](https://user-images.githubusercontent.com/308049/50915379-f86b0c80-1438-11e9-9756-c940eefe18b1.png)

![screenshot from 2019-01-09 18-03-42](https://user-images.githubusercontent.com/308049/50915384-fd2fc080-1438-11e9-975b-4d484bcbe401.png)

Jikstra:
Currently we're extactly defining the height of the ChatView and the Composer (where you input the message). We need to change both heights to get this, which is messy. We should switch to a grid layout or something, everything else is quite hacky.

ralphtheninja:
> We should switch to a grid layout or something, everything else is quite hacky.

Yes!

PR #654, opened 2019-02-01
Resize the composer based on the newlines in the text

Jikstra:
screenshot_20190131_162940

Still missing:
- [x] emoji picker isn't at the right position currently
- [x] resizing isn't on point yet
- [x] we sometimes have scrollbars even if we don't need them
- [x] composer buttons should stick to the bottom (position them absolutely)
- [x] some weird css bugs if we're havent selected a chat yet
- [x] resizing composer is getting called too often

Fixes https://github.com/deltachat/deltachat-desktop/issues/543

### DIFF
--- a/src/renderer/components/ChatView.js
+++ b/src/renderer/components/ChatView.js
@@ -16,9 +16,11 @@ const ChatViewWrapper = styled.div`
   width: 70%;
   background-color: #eeefef;
   float: right;
+  display: grid;
+  grid-template-columns: auto;
+  height: 100vh;
 
   #the-conversation {
-    height: calc(100vh - 50px - 40px);
     overflow: scroll;
     background-image: url("../images/background_hd.jpg");
     background-size: cover;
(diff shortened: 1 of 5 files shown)


src/renderer/components/ChatView.js
```diff
@@ -87,7 +91,7 @@ class ChatView extends React.Component {
     this.lastId = this.props.chat.id
     this.previousScrollHeightMinusTop = null
 
-    this.composerRef = React.createRef()
+    this.conversationRef = React.createRef()
```
ralphtheninja: This breaks setting focus on the composer, used in `SplittedChatListAndView.js`.

ralphtheninja: cc @Simon-Laux

Simon-Laux: If that works, do it.

Jikstra:
Just found a bug with the three dots menu, the top part is hidden. I'm trying to fix it.

PR #662, opened 2019-02-04
Multiline keybinding setting

### LINKED ISSUE #542: Add option "Enter" send or line break
Currently (v.0.98.1-preview), by pressing the enter key, the message will be sent.

I would like the behavior:
Enter key makes a line break.

A option for the enter key behavior would be great.

Simon-Laux:
closes #542 

What's missing/not 100% working:
- [x] remove console logs
- [x] root state isn't refreshed when updating rc settings
- [x] we should maybe use React.createContext() instead of passing saved over to all components until we reach Composer component

### DIFF
--- a/src/logger.js
+++ b/src/logger.js
@@ -21,6 +21,7 @@ function log (channel, level, ...args) {
   const variant = LoggerVariants[level]
   if (!handler) {
     console.log('Failed to log message - Handler not initilized yet')
+    console.log(`Log Message: ${channel} ${level} ${args.join(' ')}`)
     throw Error('Failed to log message - Handler not initilized yet')
   }
   handler(channel, variant.level, ...args)
--- a/src/main/ipc.js
+++ b/src/main/ipc.js
@@ -159,6 +159,7 @@ function init (cwd, state, logHandler) {
   ipcMain.on('setConfig', (e, key, value) => {
     e.returnValue = dc.setConfig(key, value)
   })
+
   ipcMain.on('getConfigFor', (e, keys) => {
     e.returnValue = dc.getConfigFor(keys)
   })
@@ -201,6 +202,7 @@ function init (cwd, state, logHandler) {
   ipcMain.on('updateSettings', (e, saved) => {
     dc.updateSettings(saved)
     app.saveState()
+    render()
   })
 
   ipcMain.on('updateCredentials', (e, credentials) => {
--- a/src/main/state.js
+++ b/src/main/state.js
@@ -34,6 +34,7 @@ function getDefaultState () {
      */
     saved: {
       markRead: true,
+      enterKeySends: false,
       notifications: true,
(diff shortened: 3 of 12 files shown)


Simon-Laux:
@ralphtheninja how should I access the state to save such settings like keybindings?

Jikstra:
@Simon-Laux I think rc does that too https://github.com/dominictarr/rc#readme

Jikstra:
Sooo what I did until now:
- Implemented the React.createContext() stuff. This is quite nice because it allows us to have "global variables" which we can consume in every child component without passing the same prop through many levels. This technique we could also use for deltachat or all other "global" states.
- I put ComposerMessageInput in it's own file. The Composer file was quite big with more than 400 lines.
- logger.js didn't log the message at all if handler wasn't defined, that's bad. I added a log message. Faced this issue and a very unhelpful error message during refactoring the composer.

src/main/ipc.js
```diff
@@ -201,6 +202,7 @@ function init (cwd, state, logHandler) {
   ipcMain.on('updateSettings', (e, saved) => {
     dc.updateSettings(saved)
     app.saveState()
+    render()
```
Jikstra: Make sure that we rerender so that the new state is in the frontend

src/logger.js
```diff
@@ -21,6 +21,7 @@ function log (channel, level, ...args) {
   const variant = LoggerVariants[level]
   if (!handler) {
     console.log('Failed to log message - Handler not initilized yet')
+    console.log(`Log Message: ${channel} ${level} ${args.join(' ')}`)
```
Jikstra: We should log the message even if the log handler is not initialized yet

src/renderer/App.js
```diff
@@ -7,15 +8,22 @@ const enLocaleData = require('react-intl/locale-data/en')
...
-    const { state } = this.props.STATE_WRAPPER
+    const { state } = this
     return (
-      <IntlProvider locale={window.localeData.locale}>
-        <ScreenController
-          logins={state.logins}
-          saved={state.saved}
-          deltachat={state.deltachat} />
-      </IntlProvider>
+      <SettingsContext.Provider value={state.saved}>
```
Jikstra: This stuff sets up the new context

src/main/state.js
```diff
@@ -48,6 +49,7 @@ function load (cb) {
       log.info('Missing configuration file. Using default values.')
     }
     const state = getDefaultState()
+    console.log('s1', state)
```
ralphtheninja: Remove

src/main/windows/main.js
```diff
@@ -28,6 +28,7 @@ function init (app, options) {
   if (main.win) {
     return main.win.show()
   }
+  console.log('s', app.rc)
```
ralphtheninja: Remove

PR #666, opened 2019-02-05
Improvement: change handle position for context menu

nicodh:
Since deltachat supports custom backgrounds, the handle for context menus (only visible on hover) should not be near the bubble but in the upper right corner. Otherwise it might not have enough contrast.

I added 10px padding-top to the text to enable a proper distance, may be there is a better solution...

context-menu-handle

### DIFF
--- a/css/conversations/_modules.scss
+++ b/css/conversations/_modules.scss
@@ -51,11 +51,12 @@
 .module-message__buttons {
   position: absolute;
   top: 0;
-  bottom: 0;
+  right: 0;
   display: inline-flex;
   flex-direction: row;
   align-items: center;
   opacity: 0;
+  z-index: 10;
 }
 
 .module-message:hover .module-message__buttons {
@@ -63,11 +64,9 @@
 }
 
 .module-message__buttons--incoming {
-  left: 100%;
+  right: 10px;
 }
-.module-message__buttons--outgoing {
-  right: 100%;
-}
+
 
 .module-message__buttons__download {
   height: 24px;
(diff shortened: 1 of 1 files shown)


Jikstra:
Thanks for the pr, that's already a very good improvement to the current situation, but i think ui/ux wise it's even better to have the three dots rotated by 90° and positioned on the left or right side of the message bubble.

In case you want to apply that already to this PR this would be very nice, otherwise i would merge this pr and apply the other changes.

Thanks for contributing :)

Jikstra:
@nicodh I would keep the three dots vertical, as this is how we use them in other places too. For the download button it should be fine to move that option into the three dots menu in my opinion.

nicodh:
"keep them vertical" ? 
they are horizontal now so do you mean: "keep them horizontal" or "change to vertical"?

nicodh:
Should we continue this discussion in the forum or in the bugtracker?
Changing it to vertical is just one line css, but should we put it for incoming to the left and for outgoing to the right? Then we will get into trouble in group discussions, since there is the avatar on the left and the name of the message author is on the upper left corner (where the handle would be then...)

Jikstra:
I'm currently not looking into the forum (which is bad on my side, but there are many channels to watch, i currently concentrate on mailinglist, irc & github).

I also have the idea of getting rid of the three dots menu at all and open the menu by right clicking on the message which is maybe also a good solution. But maybe it makes sense to talk about this in the forum? I'm unsure.

violoncelloCH:
I would not completely remove the three dots menu but rather making a right click (anywhere on the bubble) toggling the same context menu as the three dots. Because there might be users not expecting things to be hidden in right click (especially in a web/electron application). Also people coming from mobile platforms all know the tree dots, but maybe not the right click... The three dots menu is a good UI element to show, that there is a context menu. 

(sorry for continuing the discussion here)

PR #670, opened 2019-02-06
Do not send empty messages (i.e. only spaces)

nicodh:
I agree, empty messages should not be sent. We still should discuss, if we trim spaces from messages...


Jikstra:
Cool!

PR #673, opened 2019-02-07
Add some integration tests

nicodh:
Use ethereal account for tests
Add test config
Change promises to async await syntax
Add id selector for AdvancedButton

The ethereal account is only a temporary solution. Would be better to have real mailboxes but each test should create (and destroy) them on-the-fly...

TODO:

- [x] Figure out why prebuilts doesn't work on Travis/linux (most likely related to invalid ssl)
- [x] Change email addresses for integration tests (Maybe use `DC_ADDR` for address)
- [x] Update Travis with new secure environment variables (`DC_ADDR` and `DC_MAIL_PW`)
- [x] Change `fixtures/config.json` to just `fixtures/config.js` (then we can put environment variables directly inside configuration file)
- [x] Wait for https://github.com/prebuild/node-gyp-build/pull/18 to be merged and released (will help with debugging why the prebuild node bindings fail to load)


nicodh:
It seems the ethereal server are not accessible in Jenkins...

nicodh:
Logging doesn't help really. Just saying: 
SMTP-login failed for user uh3jbynm5f75jieu@ethereal.email (Login failed)

PR #685, opened 2019-02-17
Drop File to send

Simon-Laux:
What works currently: 

- [x] drag on input field/composer to send file

What should be in the pull request: 

- [x] drag anywhere into chatview
- [x] ~Visual Feedback on dragging~ Postponed

related to #415 

EDIT: add that link: https://electronjs.org/docs/api/file-object

### DIFF
--- a/src/renderer/components/ChatView.js
+++ b/src/renderer/components/ChatView.js
@@ -183,7 +183,7 @@ class ChatView extends React.Component {
     const forbiddenPathRegEx = /DeltaChat\/[\d\w]*\/db\.sqlite-blobs\//gi
     for (let i = 0; i < files.length; i++) {
       const { path } = files[i]
-      if (!forbiddenPathRegEx.test(path)) {
+      if (!forbiddenPathRegEx.test(path.replace('\\', '/'))) {
         ipcRenderer.send('sendMessage', chat.id, 'Droped File', path)
       } else {
         log.warn('Prevented a file from being send again while dragging it out')
--- a/src/renderer/components/ChatView.js
+++ b/src/renderer/components/ChatView.js
@@ -179,9 +179,15 @@ class ChatView extends React.Component {
     const { chat } = this.props
     e.preventDefault()
     e.stopPropagation()
+    // TODO maybe add a clause here for windows because that uses backslash instead of slash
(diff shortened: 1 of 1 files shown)


Simon-Laux:
Visual feedback the way I wanted proves to be much more difficult for me to get working in react, than expected. I'll probably postpone it...

src/renderer/components/ChatView.js
```diff
@@ -174,13 +174,35 @@ class ChatView extends React.Component {
     this.setState({ composerSize: size })
   }
 
+  onDrop (e) {
+    const files = e.target.files || e.dataTransfer.files
+    const { chat } = this.props
+    e.preventDefault()
+    e.stopPropagation()
+    // TODO maybe add a clause here for windows because that uses backslash instead of slash
```
Simon-Laux: I should modify the RegEx that it also works with windows paths.

src/renderer/components/ChatView.js
```diff
@@ -174,13 +174,35 @@ class ChatView extends React.Component {
...
+    const forbiddenPathRegEx = /DeltaChat\/[\d\w]*\/db\.sqlite-blobs\//gi
+    for (let i = 0; i < files.length; i++) {
+      const { path } = files[i]
+      if (!forbiddenPathRegEx.test(path)) {
```
Simon-Laux: The whole RegEx thing is a workaround because I found no better solution.

src/renderer/components/ChatView.js
Simon-Laux: Before merging

src/renderer/components/ChatView.js
Jikstra: Not sure if this is needed. I get the point with not sneaking out the deltachat passwords and stuff, but if we are this far, that someone can control this part of the code, the evil person could also sneak out your `/etc/passwd` or your `~/porn collection`. Not sure if this really helps.

Simon-Laux: Not thats not the point, this part prevents draging out a file from dc and drop it into dc and so sending it again by accident

Jikstra: Aaah i get it :)

src/renderer/components/ChatView.js
Simon-Laux: Did it

Simon-Laux:
Should we merge that?

PR #690, opened 2019-02-20
Open context menu on right click

### LINKED ISSUE #672: Open context menu on right click on message instead of showing the three dot menu
Additional thing would be to move the download attachment button also into the context menu.

@nicodh even if your changes already highly improve the situation, this might make it even better and introduces a general ui/ux element which we can also use for the chat list items and don't need to show/hide something on hover.

nicodh:
Resolves #672

### DIFF
--- a/src/renderer/components/conversations/Message.js
+++ b/src/renderer/components/conversations/Message.js
@@ -543,6 +543,7 @@ class Message extends React.Component {
 
     return (
       <div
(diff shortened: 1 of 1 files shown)


violoncelloCH:
@Simon-Laux I would not remove the 3dot-menu, see https://github.com/deltachat/deltachat-desktop/pull/666#issuecomment-461214159

PR #695, opened 2019-02-23
Settings login

nicodh:
As long as wrong credentials result in a crash this PR should not be merged


nicodh:
See https://github.com/deltachat/deltachat-node/pull/250

nicodh:
Found the right place now to handle the error. Password handling is now similar to Android app

Jikstra:
Good coe, good fix. Thanks :+1:

PR #698, opened 2019-02-24
context menu for chatListItems

### LINKED ISSUE #374: Rightclick on chat in chatlist
I think it would be a good idea to enable a context menu when right-clicking on a chat in the chat list that has the basic chat options like delete and archive.

Simon-Laux:
closes #374
## TODO

- [x] Find out why onContextMenu isn't triggered and fix it

Get the information for those actions:
- [x] show Encryption Info
- [x] EditGroup screen
- [x] BlockContact

What works:
- [X] Leave Group
- [X] Archive Chat
- [X] DeleteChat

TODO on chat objects

- [X] deduplicate creation
- [x] rename semi chat object (that is used in the chat list) to chatListItem to clear up things
- [x] optimize chatlistitem object (remove everything unused)


Simon-Laux:
@Jikstra @ralphtheninja @karissa The 3 things in the middle require some architecture improvements to work. Except for `chatId to contactId for BlockContact` that could be solved with a new ipc call.
What do you think about this matter?

Jikstra:
Can you explain what architecture needs improvements and why?

Simon-Laux:
It's hard to fit into words, but first of all the code use a good refactoring on some parts and I need some passed down information (through countless this.props requirements) to perform actions that shouldn't be depended on it like switching to the settings page on non current chat... But better discuss that in person @Jikstra maybe you find better words for it.

karissa:
You can open a dialog using the openDialog function which gets passed to children

https://github.com/deltachat/deltachat-desktop/blob/d8f9fa9ff6093ed581a4073d35d120f0dbbf1033/src/renderer/components/Menu.js#L76

Right now this requires you to have the chat object and pass that through. Perhaps we need the ability to get a chat by id in the frontend, which could then be passed to the dialog (or the dialog could take an id and do this lookup)

Simon-Laux:
Yeah it needs the full chat object, an I only have the fraction of it that is there to generate the chat list itself at that point.
https://github.com/deltachat/deltachat-desktop/pull/698/files#diff-ce7102dfccf06ac82fe8a3ede9bdbe4bR170

I can get the data somewhere, but I rather have a somewhat clean way of doing it. not an ugly workaround and three extra ipc calls that are then only used here.

static/main.html
```diff
@@ -1,6 +1,9 @@
 <!doctype html>
 <html lang="en">
   <head>
+    <!-- uncomment this to add the react debugger:
```
Simon-Laux: we maybe want to add the react-devtools rather in our webpack configuration than here.
https://github.com/facebook/react-devtools/tree/master/packages/react-devtools

Simon-Laux:
Done, finally.

src/main/deltachat.js
```diff
@@ -445,6 +445,41 @@ class DeltaChatController extends EventEmitter {
...
+    if (!chat) return null
+
+    if (chatId === C.DC_CHAT_ID_DEADDROP) {
+      const chat = this._dc.getChat(C.DC_CHAT_ID_DEADDROP)
+      return (chat && Object.assign(chat.toJson(), { isDeaddrop: true })) || null
+    }
+
+    var messageIds = this._dc.getChatMessages(chat.id, C.DC_GCM_ADDDAYMARKER, 0)
+    // This object is NOT created with object assign to promote consistency and to be easier to understand
+    return {
```
karissa: There is a repetition of code here that you might want to abstract out. Check out the function where the chat list is fetched, you'll see chats getting created (also with the summary). Can we reuse that code here so the front end always can guarantee that a chat object has the same properties, generated by the same code?

Simon-Laux: we can, but I'm not sure if we should at that point... If I were too choose I would completely rewrite deltachat-node to be more object oriented and also typed. I personally would wait with refactoring that code until we make a big refactoring/code cleanup...

Simon-Laux: Also I'm not sure about this whole dead drop topic.

karissa: I'm suggesting a small improvement on this pr to reuse some existing code, that shouldnt require refactoring much else. Could be a source of bugs in the future to have two different ways to generate chat objects.

Simon-Laux: we now have 3 different sources (selectedChat, chatList, getChatById) of those objects, but if its no memory issue I can optimize it as you suggest.

Simon-Laux: I'm concerned about the messages property

Simon-Laux: like if there are many messages the chat list object will get very very big and because it gets send over ipc it might slow down the application

Simon-Laux: So I'll improve this thing I set the state of this pr back to `wip` and added some points to the todo list.

Simon-Laux: Do you agree to these todo points? @karissa @Jikstra

Simon-Laux: I optimized it and renamed the symbol/var name in the ChatList to make clear that the object there isn't the complete chat object. I'll mark this as resolved.

Jikstra:
Applied some changes but couldn't push them to your fork, so they live on this repo in a seperate branch: https://github.com/deltachat/deltachat-desktop/commit/de46aa43610a99cf4172696a045dbf98d61af437

PR #699, opened 2019-02-25
WIP: Keep message drafts (unsent messages) per chat

### LINKED ISSUE #582: change a (not sendet message) to draft like in android client
- Platform (linux/mac/windows): Linux/Kubuntu
- Delta Chat Desktop Version: 0.98.2

- Expected behavior: 

If I switch to another chat,messages which not yet send, should stored as draft and the input field now should be empty, how it works in android client.

- Actual behavior: 

If I write a message to Alice and click on Bobs chat, the text will stay and can be send by accident to Bob,

-Open a chat to Alice. 
-Write a message e.g.: Alice, last night I never will forget. Bob should never know what happens last night.
-Don't send this message. 
-Click on Bobs chat, e.g to check his last message.
-Now (you are in hurry) click on the send button.

**Probably now you lost a friend ;-)**

nicodh:
resolves #582

### DIFF
--- a/src/main/deltachat.js
+++ b/src/main/deltachat.js
@@ -332,6 +332,11 @@ class DeltaChatController extends EventEmitter {
     return config
   }
 
+  setDraft (chatId, msg) {
+    log.debug(`setDraft: ${msg}, ${chatId}`)
+    this._dc.setDraft(chatId, msg)
+  }
+
   getAdvancedSettings () {
     return {
       addr: this._dc.getConfig('addr'),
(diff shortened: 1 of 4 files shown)


Jikstra:
DeltaChat Core has some api methods to save/get drafts from the database. This would be the better approach. I would prefer to have it that way, this would make sure that we don't loose drafts by closing the program. Are you interested in diving into that topic? It would require some modifications to the deltachat backend (`src/main/deltachat.js`) and calling some method over ipc.

nicodh:
Since I can't debug the bindings I have no clue whats going wrong here.

Jikstra:
@nicodh  looked into the -node bindings code and it looks fine. How do you determine that it has no effect?

nicodh:
I already checked the sqlite db. Nothing is changing if I call dc.setDraft. And if I add an entry for a chat (draft_timestamp and draft_txt) nothing is returned when calling dc.getDraft. It would be nice to get these values together with the chat object as properties. Not sure if the core supports this. I didn't try to recompile deltachat-node so far...

Jikstra:
> It would be nice to get these values together with the chat object as properties. Not sure if the core supports this. I didn't try to recompile deltachat-node so far...

Shouldn't be a problem as we're constructing our own chat object anyways.

PR #703, opened 2019-02-28
Change background color for success feedback

nicodh:
At least on my screens the is a very bad contrast with the dark green background
New color has same value as $color-core-green variable

### DIFF
--- a/static/main.css
+++ b/static/main.css
@@ -50,7 +50,7 @@ input:focus {
   background-color: red;
 }
 .user-feedback.success {
-  background-color: green;
+  background-color: #4caf50;
 }
 img {
   max-width: 500px;


nicodh:
btw: why is main.css static and not rendered with sass? Historical reasons I guess?

Simon-Laux:
> btw: why is main.css static and not rendered with sass? Historical reasons I guess?

We still need to cleanup the css/scss anyway -> still many unused parts that can be removed.
But I would start that after we have rewritten the chat view to handle jump to message and infinity scrolling. 

Also see this milestone under which we can put issues to the css/scss restructure things: https://github.com/deltachat/deltachat-desktop/milestone/4

Issue #700, opened 2019-02-25
Composer input handler is slow


Jikstra:
Not sure why, but something is really slow there. I think we shouldn't need to update the state every time the text changes.

We should debounce the resize methods

nicodh:
Can't reproduce this. Performance on Mac is fine...

Jikstra:
The problem is that the resize...() method in the Composer takes up to 30ms on my machine and because it's getting called on each change of the textarea, they can queue up for quite some time. Maybe not easily reproducable on a strong machine.

PR #704, opened 2019-02-28
Fix multline performance

Jikstra:
Fixes https://github.com/deltachat/deltachat-desktop/issues/700


Simon-Laux:
I think we'll get a conflict when we try to merge with #662 @Jikstra

PR #706, opened 2019-03-01
Adapt width & height of renderMedia overlay

### LINKED ISSUE #470: X for closing video element not clickable
screenshot from 2018-12-21 00-02-02

You can get rid of the video view by clicking outside of it though.

nicodh:
resolves #470

- add z-index to closing element
- overlay width relative to current window (70%)
- change controls attribute value to 1 to avoid console warning

In my test videos were displayed far too big, so the
cross was not even visible. Now that only happens,
if you have a width/height ratio > 3:2

Not sure if the img {max-width: 500px} is needed
somewhere else...

### DIFF
--- a/static/main.css
+++ b/static/main.css
@@ -52,9 +52,6 @@ input:focus {
 .user-feedback.success {
   background-color: #4caf50;
 }
-img {
-  max-width: 500px;
-}
 
 ul {
   list-style: none;
@@ -84,6 +81,11 @@ ul li button {
   display: flex;
   align-items: center;
   justify-content: center;
+}
+
+.bp3-overlay-content img,
+.bp3-overlay-content video {
+  width: 100%;
 }
 
 .module-conversation-list-item__is-verified {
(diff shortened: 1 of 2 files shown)


Jikstra:
I have some pictures which are shown way too big, which wasn't the case before. Scaling dynamic sized images is a hard task... atleast with css. Maybe we want some js to do the job?

static/main.css
```diff
@@ -52,9 +52,6 @@ input:focus {
 .user-feedback.success {
   background-color: #4caf50;
 }
-img {
-  max-width: 500px;
```
Jikstra: I think this made sure that the images are somehow more or less shown in their full size. But not 100% sure.

nicodh:
Hmm, I tested with a JPG image wich was  3.456 x 4.608 Pixel
What type of image did you test? div.bp3-overlay-content is always 70% width on my screen. 
Was your picture in a portrait ratio or in landscape?

Jikstra:
I gave it a try: https://github.com/deltachat/deltachat-desktop/pull/710

Simon-Laux:
How is this issue related to #710 , can both be merged or is #710 a newer version of this one?

Jikstra:
@Simon-Laux  They both try to fix the same problem, so we should only pick one of them

PR #474, opened 2018-12-21
import/export keys fixes #251

karissa:
this still needs to ask the user for a password and then check it in the backend

PR #707, opened 2019-03-01
Keys export import

nicodh:
continues #474

### DIFF
--- a/_locales/_experimental_en.json
+++ b/_locales/_experimental_en.json
@@ -1,5 +1,17 @@
 {
   "login_instruction_desktop": {
     "message": "Login with an existing e-mail account (use your e-mail credentials here)"
+  },
+  "keysImport": {
+    "message": "Import keys"
+  },
+  "keysExport": {
+    "message": "Export keys"
+  },
+  "keysExportConfirmationMessage": {
+    "message": "Exporting your end-to-end setup! Keep these keys in a safe place and delete them as soon as they are not needed anymore."
+  },
+  "settingsManageKeys": {
+    "message": "Manage Keys"
   }
 }
--- a/_locales/_experimental_en.json
+++ b/_locales/_experimental_en.json
@@ -2,21 +2,6 @@
   "login_instruction_desktop": {
     "message": "Login with an existing e-mail account (use your e-mail credentials here)"
   },
-  "keys_import": {
-    "message": "Import keys"
-  },
-  "keys_export": {
-    "message": "Export keys"
-  },
-  "keys_export_confirmation_message": {
-    "message": "Exporting your end-to-end setup! Keep these keys in a safe place and delete them as soon as they are not needed anymore."
-  },
-  "keys_export_success_message": {
-    "message": "Keys successfully exported to %1$s"
-  },
-  "settings_manage_keys": {
-    "message": "Manage Keys"
-  },
   "update": {
     "message": "Update"
   },
--- a/_locales/_experimental_en.json
+++ b/_locales/_experimental_en.json
@@ -2,16 +2,19 @@
   "login_instruction_desktop": {
     "message": "Login with an existing e-mail account (use your e-mail credentials here)"
   },
-  "keysImport": {
+  "keys_import": {
     "message": "Import keys"
   },
-  "keysExport": {
+  "keys_export": {
     "message": "Export keys"
   },
-  "keysExportConfirmationMessage": {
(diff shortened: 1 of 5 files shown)


nicodh:
@karissa mentioned it should ask for the password. But which one? The password for password protected keys?

karissa:
The password to the account which is checked by logging in or checking the config var in the backend. Use an IPC call for this.

nicodh:
Hmm, the account password is easily to get from the settings since 4925f93c674607cd2d9cee77b933bdf3cf0dcb1f so this would not really improve security. The android app asks for the PIN (of the phone) when you export the secret keys. But afaik it is not possible to ask for the users OS account password with electron. Might be useful to introduce some PIN/password protection for deltachat desktop but that's another topic...

Jikstra:
Is this really needed to ask for the email password? @r10s @hpk42

_locales/_experimental_en.json
```diff
@@ -2,6 +2,21 @@
...
+  },
+  "keys_export_confirmation_message": {
+    "message": "Exporting your end-to-end setup! Keep these keys in a safe place and delete them as soon as they are not needed anymore."
+  },
+  "keys_export_success_message": {
+    "message": "Keys successfully exported to %1$s"
+  },
+  "settings_manage_keys": {
+    "message": "Manage Keys"
+  },
```
Jikstra: Can't we use the already translated translation keys? I think they are adequate to the ones here. Maybe some leftovers of the old pr from @karissa? We now have access to all the translations which we also use on deltachat-android.
```
  "pref_managekeys_menu_title": {
    "message": "Manage keys"
  },
  "pref_managekeys_ask_for_storage_permission": {
    "message": "Delta Chat needs access to your files to import/export keys."
  },
  "pref_managekeys_export_secret_keys": {
    "message": "Export secret keys"
  },
  "pref_managekeys_export_explain": {
    "message": "Export secret keys to \"%1$s\"?"
  },
  "pref_managekeys_import_secret_keys": {
    "message": "Import secret keys"
  },
  "pref_managekeys_import_explain": {
    "message": "Import secret keys from \"%1$s\"?\n\n• Existing secret keys will not be deleted\n\n• The last imported key will be used as the new default key unless it has the word \"legacy\" in its filename"
  },
  "pref_managekeys_secret_keys_exported_to_x": {
    "message": "Secret keys written successfully to \"%1$s\"."
  },
  "pref_managekeys_secret_keys_imported_from_x": {
    "message": "Secret keys imported from \"%1$s\"."
  },
```

nicodh:
DC_EVENT_IMEX_FILE_WRITTEN should be triggered after keys are exported, but is not...?

Jikstra:
> DC_EVENT_IMEX_FILE_WRITTEN should be triggered after keys are exported, but is not...?

Good question. Maybe @r10s can help us on this question.
Exporting seems to work. Importing too, but didn't properly test this. Also not sure how to test that exactly.

r10s:
> DC_EVENT_IMEX_FILE_WRITTEN should be triggered after keys are exported, but is not...?

it's a bug in core, i'll fix it.

nicodh:
So this works fine now with the fix in core. Except the fact that we can't show a success message after having successfully imported keys since there is no event for that. Even if a DC_EVENT_IMEX_FILE_WRITTEN event would be triggered  in https://github.com/deltachat/deltachat-core/blob/master/src/dc_imex.c#L706 we couldn't distinguish in the event handler which action triggered it. 
At the moment I distinguish export-backup from export-keys just by checking the "bak" ending exists in filename...
https://github.com/deltachat/deltachat-desktop/blob/keys-export-import/src/renderer/ScreenController.js#L70
Which is kind of a hack... ;-)

r10s:
it's all about export?

there are the events DC_EVENT_IMEX_PROGRESS - on success 1000 permille will be reported, on failure, it's 0.

in fact, these events do not regard single export failed. i'll change this in the branch.

@nicodh thanks for having a closer look at this :) it is obviously one of the parts that lacks some details :)

r10s:
i've updated the branch, [DC_EVENT_IMEX_PROGRESS](https://c.delta.chat/group__DC__EVENT.html#ga006ea41d9c1a76ffc672752484c61e6c) reports success/failure on exporting keys now.

wrt to decide whether the event belongs to a backup or to a key-export (or whatever will be added in the future - location-export?):

there is always only one export/import running at the same time. so, you can just remember the last action started.

PR #710, opened 2019-03-04
Fix fullscreen image view, we now use the full height if possible, an…

Jikstra:
…d 90% of the screen width. Changed ui a bit.

screenshot_20190304_222949

What i would like to have:
- [x] bigger icons
- [x] another icon for the download button

### DIFF
--- a/src/renderer/components/dialogs/RenderMedia.js
+++ b/src/renderer/components/dialogs/RenderMedia.js
@@ -1,7 +1,7 @@
 const React = require('react')
 const styled = require('styled-components').default
 const createGlobalStyle = require('styled-components').createGlobalStyle
-const { Button, Overlay } = require('@blueprintjs/core')
+const { Button, Icon, Overlay } = require('@blueprintjs/core')
 
 const OverlayGlobal = createGlobalStyle`
   .bp3-overlay-backdrop {
@@ -31,7 +31,8 @@ const RenderMediaWrapper = styled.div`
 const CloseButtonWrapper = styled.div`
   float: right;
   position: absolute;
-  right: 0;
+  right: 5px;
+  cursor: pointer;
 `
 
 const DownloadButtonWrapper = styled.div`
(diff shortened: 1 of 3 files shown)


nicodh:
Video was missing (thats how I found this issue since video was far too big)
But now it's fine and we can use this solution.

src/renderer/components/dialogs/RenderMedia.js
```diff
@@ -1,17 +1,44 @@
 const React = require('react')
 const styled = require('styled-components').default
+const createGlobalStyle = require('styled-components').createGlobalStyle
```
nicodh: why global style if it's only used here?

Jikstra: Because i can't overwrite the bp3-overlay thing without this. I tried to add a wrapper around the <Overlay> component but it didn't show up in the dom. I guess blueprint is doing some stuff there to only use the overlay component.

Jikstra:
> Video was missing (thats how I found this issue since video was far too big)
> But now it's fine and we can use this solution.

Good catch! Never tried videos in fullscreen, but i should :D

Simon-Laux:
Should we add the image caption to the full screen view?

Simon-Laux:
We could maybe add a hint, that the escape key can also be used to close it, like discord does:

Simon-Laux:
When you open it from the gallery it should present buttons to go to the previous and the next thing in line, but that could be a bit more complicated: When the user has selected to filter the gallery to only videos, only videos should appear in next and previous.
(This comment is probably worth its own issue)

Simon-Laux:
There is a bug with portrait video in full screen
fullscreen_video
And after minimizing again:
fullscreen_video2

nicodh:
I tested images in landscape/portrait ratio, videos in landscape/portrait ration & audio. Looked all fine.
Maybe you can send me the video which didn't work via deltachat? ;-)

nicodh:
This is how portrait videos look like:

image

Jikstra:
This looks like some glitch and not something which we can fix with some js/css/html. Maybe an electron bug or your graphics/desktop environment whatever.

Simon-Laux:
full screen the video then minimize it again, with the full screen button on the video

nicodh:
you mean the white rectangle which appears some times? Probably a rendering problem, not css related imho...

Jikstra:
Can't reproduce on my machine. Seems like a gpu/graphics/linux/electron problem to me.

Simon-Laux:
LGTM, besides the two open questions I still have:
- Include Image Caption?
- Hint to the posibility to use the escape button to quit this view?

Jikstra:
> LGTM, besides the two open questions I still have:
> 
>     * Include Image Caption?
> 
>     * Hint to the posibility to use the escape button to quit this view?

I prefer to tackle those issues in a seperate pr.

nicodh:
+1

Jikstra:
@nicodh just saw the changes you made with bigger icon + other download icon. Thanks!

Issue #713, opened 2019-03-05
Uncaught TypeError: Cannot read property 'slice' of undefined


Jikstra:
```
Uncaught TypeError: Cannot read property 'slice' of undefined
    at ComposerMessageInput.insertStringAtCursorPosition (bundle.js:1972)
    at ComposerMessageInput.onKeyDown (bundle.js:1928)
    at HTMLUnknownElement.callCallback (/opt/DeltaChat/electron_app/node_modules/react-dom/cjs/react-dom.development.js:149)
    at Object.invokeGuardedCallbackDev (/opt/DeltaChat/electron_app/node_modules/react-dom/cjs/react-dom.development.js:199)
    at invokeGuardedCallback (/opt/DeltaChat/electron_app/node_modules/react-dom/cjs/react-dom.development.js:256)
    at invokeGuardedCallbackAndCatchFirstError (/opt/DeltaChat/electron_app/node_modules/react-dom/cjs/react-dom.development.js:270)
    at executeDispatch (/opt/DeltaChat/electron_app/node_modules/react-dom/cjs/react-dom.development.js:561)
    at executeDispatchesInOrder (/opt/DeltaChat/electron_app/node_modules/react-dom/cjs/react-dom.development.js:583)
    at executeDispatchesAndRelease (/opt/DeltaChat/electron_app/node_modules/react-dom/cjs/react-dom.development.js:680)
    at executeDispatchesAndReleaseTopLevel (/opt/DeltaChat/electron_app/node_modules/react-dom/cjs/react-dom.development.js:688)
/opt/DeltaChat/electron_app/node_modules/react-dom/cjs/react-dom.development.js:289 Uncaught TypeError: Cannot read property 'slice' of undefined
    at ComposerMessageInput.insertStringAtCursorPosition (bundle.js:1972)
    at ComposerMessageInput.onKeyDown (bundle.js:1928)
    at HTMLUnknownElement.callCallback (/opt/DeltaChat/electron_app/node_modules/react-dom/cjs/react-dom.development.js:149)
    at Object.invokeGuardedCallbackDev (/opt/DeltaChat/electron_app/node_modules/react-dom/cjs/react-dom.development.js:199)
    at invokeGuardedCallback (/opt/DeltaChat/electron_app/node_modules/react-dom/cjs/react-dom.development.js:256)
[...]

PR #714, opened 2019-03-05
fallback to set textValue to empty string

nicodh:
resolves #713 

(at least I hope so)

PR #718, opened 2019-03-06
We don't get the text value from the parent component and don't need …

Jikstra:
…it. Properly initialize the text state variable. Also fixes inserting emojis.

Fixes https://github.com/deltachat/deltachat-desktop/issues/713

Only this or https://github.com/deltachat/deltachat-desktop/pull/714 should get merged

### DIFF
--- a/src/renderer/components/Composer.js
+++ b/src/renderer/components/Composer.js
@@ -188,7 +188,7 @@ class Composer extends React.Component {
 
   onEmojiSelect (emoji) {
     log.debug(`EmojiPicker: Selected ${emoji.id}`)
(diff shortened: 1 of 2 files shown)


nicodh:
you are right - I shouldn't have committed my late night hot fix without proper investigating...

Simon-Laux:
now I have `Uncaught TypeError: this.insertStringAtCursorPosition is not a function` errors when inserting a smiley

Jikstra:
> now I have Uncaught TypeError: this.insertStringAtCursorPosition is not a function errors when inserting a smiley

Oh! Good bug hunting, fixed it :)

PR #730, opened 2019-03-17
Update and integrate appstream file

flub:
The appstream/appdata file is required for linux destop appliciations
and is not packaging-format specific so makes sense to include
upstream.

- Update existing appdata file.

- Add .desktop file since this is referred to by the appdata.

- Update the travis config to test this appdata and .desktop files
  making sure they remain correct.  We have to add linux back to the
  job matrix for this, but we only want to publish a binary release
  using electron-builder on OSX.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,64 @@
+sudo: required
 language: node_js
+dist: xenial
+
+addons:
+  apt:
+    packages:
+      - appstream-util
+      - desktop-file-utils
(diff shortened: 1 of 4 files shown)


flub:
I experimented with multiple jobs instead of the default matrix as well, but the main downside is there is no data shared between jobs so i stuck with the not overly nice main matrix thing instead.  The ugly part of that is how i select things which are only for one platform, but it works.

PR #741, opened 2019-03-24
[needs testing] move the screenfunctions to a context

### LINKED ISSUE #727: use a react Context instead of passing down functions as props
Especially for functions like:
- changeScreen
- openDialog

See https://github.com/deltachat/deltachat-desktop/blob/master/src/renderer/ScreenController.js

@Jikstra fee free to improve the wording of this issues

Simon-Laux:
resolves #727 for the time being. We should come up with an even better solution in the future.

Only merge after enough testing and/or review.

### DIFF
--- a/src/renderer/ScreenController.js
+++ b/src/renderer/ScreenController.js
@@ -14,6 +14,7 @@ const {
   Button
 } = require('@blueprintjs/core')
 
+const ScreenContext = require('./contexts/ScreenContext')
 const NavbarWrapper = require('./components/NavbarWrapper')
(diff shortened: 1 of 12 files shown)


Jikstra:
@Simon-Laux is it possible to pass down the ScreenContext also as a prop? Because using contexts in non hook react components is quite frustrating sometimes, switching to functional components is maybe the way to go but a huge refactoring. In general i would like to rebase this and get it in.

Jikstra:
@nicodh any issues with this pr? Otherwise lets merge it.

PR #752, opened 2019-04-02
[wip] simplify deb-building with circle-ci

hpk42:
Misses:
- [ ] only build on master and new tags
- [ ] upload builds to github


ralphtheninja:
@hpk42 So this will replace linux on Travis, right? In that case I think we should remove `- linux` from `.travis.yml`.

hpk42:
On Wed, Apr 03, 2019 at 01:29 -0700, Lars-Magnus Skog wrote:
> @hpk42 So this will replace linux on Travis, right? In that case I think we should remove `- linux` from `.travis.yml`.

i guess so.  still need to sit with others to figure things out 
as i am new to what happens with building with desktop.
just wanted to help along with circle-ci config and make
sure it's similar to how things are structured in 
https://github.com/deltachat/deltachat-core/tree/master/ci_scripts

Jikstra:
It won't replace the -linux on travis because this only builds .debs and doesn't run any tests or anything. This should also only get triggered on new releases/tags.

ralphtheninja:
+845 rows. Seriously? :)

ci_scripts/ubuntu-18_10-rpgp/binding.gyp
```diff
@@ -0,0 +1,44 @@
...
+        }]
+      ],
+      "sources": [
+        "./src/module.c",
+        "./src/eventqueue.c",
+        "./src/strtable.c"
+      ],
+      "include_dirs": [
+        "deltachat-core/src",
+        "<!(node -e \"require('napi-macros')\")"
```
ralphtheninja: What's the purpose of this file? And why do we need three versions of it? It's a nightmare to maintain.

Jikstra: Just some temporary hack until deltachat-node uses rpgp by default.

Jikstra:
> +845 rows. Seriously? :)

Currently all the build stuff is redundant for every ubuntu version. Some packagenames change or some commands or environment variables. I don't want to add an extra layer just to get down the lines of code. Also the binding.gyp files are redundant. I think even if this is a lot of "lines of code", this is all straight forward building shit. I prefer having many straight forward lines over having an extra layer which generates docker files and build contexts and shit.

Jikstra:
Pushing all this up to deltachat-node isn't that easy.
The biggest problem is that we can't bundle all the dependencies needed the same way as the python wheels do it or as it's possible on mac with install_name_tool. The problem is that patchelf isn't correctly patching the the deltachat.node elf file https://github.com/NixOS/patchelf/issues/158. So I would still go with this solution until we can provide good and self contained prebuilts in deltachat-node.

PR #753, opened 2019-04-03
[wip] Debug build fails

Jikstra:


Jikstra:
Seems like currently the only problem for the ci is that mac doesn't find meson which is maybe because it's not in path.

Jikstra:
Current status:
- linux builds build successful
- mac builds fail on compiling deltachat-node because of some hardcoded paths in binding.gyp. Fixing this wont be really easy or hacky and definetly will be obsolet as soon as @flub fixes mac on deltachat-node.

Jikstra:
Superseeded by https://github.com/deltachat/deltachat-desktop/pull/796

PR #761, opened 2019-04-20
[wip] [PoC] Improve performance by not recalculating state for chatList and

Jikstra:
selectedChat on every render.

### DIFF
--- a/src/main/deltachat.js
+++ b/src/main/deltachat.js
@@ -189,6 +189,7 @@ class DeltaChatController extends EventEmitter {
    */
   searchChats (query) {
     this._query = query
+    this.updateStateChatList()
     this._render()
   }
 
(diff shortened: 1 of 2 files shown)


Jikstra:
> I would like to discuss state handling in the next days. 

Yes! I think we can come up with some good ideas :)

> Although I see the improvement here it is hard (as you commented) to manually update certain parts of state in different places...

Yes this should only be seen as a poc why rendering/the app is so slow at some points. Biggest problem is the chatlist which takes forever to calculate it's state.

> After all that's one reason why redux was invented...

But redux does exactly that as far as i know, it allows to update parts of the state. But don't know much about it besides some short introduction @frando gave me.

hpk42:
this is superseded by https://github.com/deltachat/deltachat-desktop/pull/801 can be closed, no?

nicodh:
Yes can be closed

PR #773, opened 2019-04-26
remeber map controll settings

### LINKED ISSUE #768: Save last map control setting per chat
The last setting for the map controls should be persisted

Simon-Laux:
closes #768

### DIFF
--- a/src/renderer/components/map/MapComponent.js
+++ b/src/renderer/components/map/MapComponent.js
@@ -35,6 +35,11 @@ class MapComponent extends React.Component {
 
   componentDidMount () {
     this.componentDidMount = Date.now()
+    const { selectedChat, userid } = this.props
+    const savedState = window.localStorage.getItem(`${userid}_${selectedChat.id}_map`)
+    if (typeof savedState === 'string') {
+      this.setState(JSON.parse(savedState))
+    }
     mapboxgl.accessToken = MapLayerFactory.getAccessToken()
     this.map = new mapboxgl.Map(
       {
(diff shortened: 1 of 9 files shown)


src/renderer/components/map/MapComponent.js
```diff
@@ -35,6 +35,11 @@ class MapComponent extends React.Component {
 
   componentDidMount () {
     this.componentDidMount = Date.now()
+    const { selectedChat, userid } = this.props
+    const savedState = window.localStorage.getItem(`${userid}_${selectedChat.id}_map`)
+    if (typeof savedState === 'string') {
+      this.setState(JSON.parse(savedState))
```
ralphtheninja: Why `JSON.parse()`?

src/renderer/components/map/MapComponent.js
```diff
@@ -50,6 +55,14 @@ class MapComponent extends React.Component {
     this.map.addControl(new mapboxgl.NavigationControl({ showCompass: false }))
   }
 
+  componentWillUnmount () {
+    // save parts of the state we wanna keep
+    const { selectedChat, userid } = this.props
+    let saveState = JSON.parse(JSON.stringify(this.state))
```
ralphtheninja: Why the JSON dance? If you want a copy of the state you can instead do `const saveState = Object.assign({}, this.state)`

src/renderer/components/map/MapComponent.js
Simon-Laux: because localstorage stores only strings

src/renderer/components/map/MapComponent.js
Simon-Laux: I wanted a deep copy just to be sure, because I'm deleting things from the state. We probably don't need it at all as the state is destroyed later anyway when I understood react correctly.

Simon-Laux: I removed the copy altogether now

PR #795, opened 2019-05-18
change emoji picker behavior

### LINKED ISSUE #793: Stop showing the emojipicker on hover
PLEASE make it on click only, instead of hover, it continues to pop in my face, even if I am just clicking on the send button

Simon-Laux:
closes #793


ralphtheninja:
Nice!

PR #787, opened 2019-05-07
Split deltachat backend into multiple files

Jikstra:

PR #800, opened 2019-05-20
Reapply lost changes

nicodh:
It seems with #787 several changes were lost. Since it was hard for git to track the effective changes, it was hard for me too...

@Jikstra can you please check again, if all parts you moved to splitted files had really the latest code from master. I seems that at least the changes from 257bdf625 got lost...

I also added a second commit which just moves the event handler to index, but that's not related to the topic of this PR. So if you want me to split the PR in 2 I can do so...


Jikstra:
@nicodh actually i can't remember that there were already any changes from you in master. Possible that we forgot this.

nicodh:
@Jikstra you can see it in git history: you merged the changes on 6th of may...
But nevertheless, I only changed deltachat.js and I think we have these changes back now...

PR #801, opened 2019-05-22
WIP: Refactoring chatview

### LINKED ISSUE #785: Avoid superfluous renderings and refactor state handling
Right now we have these event processings

1. DC_CORE_EVENT => handled by node DeltaChatController resulting in 2 possible actions:
	- emit an event which is handled by ipc main
	- call render() to recalculate a complete state wich is send to renderer
2. UI_EVENTs which are send to ipc main and passed to node DeltaChatController (like selectChat, searchChat etc.)
3. ELECTRON events (like setAspectRatio etc.) which are handled by ipc main

At the moment there are many events which trigger a complete rerendering (like each incoming message, DC_EVENT_LOCATION_CHANGED etc.)

A possible solution could be:
1. state is kept in renderer context since we often need the current state to decide how to react on events
2. handle most of the DC_CORE_EVENTs in renderer, node DeltaChatController emits directly to renderer ( new method sendToRenderer(evt))
3. introduce new event: EVENT_DC_FUNCTION_CALL which has a method name in payload to be called on node DeltaChatController (to avoid boilerplate in ipc main like on(searchChat) => dc.searchChat) (TODO: provide a list of "allowed" methods for security)
4. introduce new event type: DD_CHAT_LOADED, DD_LOCATIONS_UPDATED etc. which has the result data as payload
5. during migration we can move events step by step to be handled by renderer

In renderer we could use stores with to kinds of processing: "reducer" (which update the state) and "effects" which trigger new events and send them to ipc main

I suggest to have different stores which dispatch different events and hold different parts of the state. If we split the handling in different stores we are more flexible.
A store would listen only to events which are related to it's state.
The location store for example holds a state which is almost independent of the react component state, since its state is only related to map content

### LINKED ISSUE #788: Delete Message -> Chat doesn't get updated
## Expected:
Message should disappear when I delete it.
## Actual Happening:
Message only disappears after changing chat and coming back.

This UI update problem has similarities to #757

nicodh:
resolves #785, resolves #788 

This refactoring avoids superfluous database calls and renderings.

It's still WIP since the same has to be done for chat list, as for now chat list is not updated,
when a message is read or new messages were received 

Comparison: number of messages loaded i.e. calls to [dc.messageIdToJson(id)](https://github.com/deltachat/deltachat-desktop/blob/master/src/main/deltachat/messagelist.js#L54):

action | master | refactored
------------|------------ | -------------
initial start | 1,148 | 0
select a chat | 369 | 21
scrolling up 10 pages |12,137 | 231
click edit group and back | 2,118 | 0
sending a message | 748 | 2

### DIFF
--- a/src/main/deltachat/chatlist.js
+++ b/src/main/deltachat/chatlist.js
@@ -16,7 +16,9 @@ function selectChat (chatId) {
   log.debug(`action - selecting chat ${chatId}`)
   this._pages = 1
   this._selectedChatId = chatId
-  this._render()
+  const chat = this._getChatById(chatId)
+  this.sendToRenderer('DD_EVENT_CHAT_SELECTED', { chat })
+  // this._render()
 }
 
 function _chatList (showArchivedChats) {
--- a/src/main/deltachat/chatlist.js
+++ b/src/main/deltachat/chatlist.js
@@ -32,7 +32,7 @@ function _chatList (showArchivedChats) {
 
   for (let i = 0; i < maxChats; i++) {
     const chatId = list.getChatId(i)
-    const chat = this._getChatById(chatId)
+    const chat = this._getChatById(chatId, true)
 
     if (!chat) continue
 
@@ -57,7 +57,7 @@ function _chatList (showArchivedChats) {
   return { chatList, listCount }
 }
 
-function _getChatById (chatId) {
+function _getChatById (chatId, skipMessages) {
   if (!chatId) return null
   const rawChat = this._dc.getChat(chatId)
   if (!rawChat) return null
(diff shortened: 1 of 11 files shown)


ralphtheninja:
The new message notification doesn't go away, not even if I click another chat and then back again.

nicodh:
Yes - that's what I mentioned in the comment. Since we don't have a "rerender chatlist" for now, but only a "rerender all". We still have to migrate the chatlist to the new state concept - that's why this PR is still WIP...

ralphtheninja:
> Yes - that's what I mentioned in the comment. Since we don't have a "rerender chatlist" for now, but only a "rerender all". We still have to migrate the chatlist to the new state concept - that's why this PR is still WIP...

Sorry. I remember reading that but apparently already forgot about it.

karissa:
Nice one! Pragmatic and simple.

ralphtheninja:
Noticed this one in the console:

Screenshot from 2019-05-29 13-37-49

PR #805, opened 2019-05-22
Fixed & refactored ubuntu .deb build scripts, implementing latest ubuntu versions

Jikstra:
Fixed working with new deltachat-node, reduced redundancy, improved build-package.sh script and implemented build dirs for ubuntu 19.04 & 19.10.


Jikstra:
@hpk can you review/merge? I think you're the only one who ever looked into this code.

ci_scripts/shared-build-context/build.sh
```diff
@@ -0,0 +1,64 @@
...
+npm run rebuild-bindings
+
+
+# actually build desktop files
+cd /build
+#npm run build
+
+
+# build .deb with electron builder (rebuilds deltachat-core for electron)
+PKG_CONFIG_PATH=/opt/DeltaChat/rpgp/lib/pkgconfig/ npx electron-builder -c /shared-build-context/electron-builder-ubuntu.json
```
hpk42: you could use ``export PKG_CONFIG_PATH=...`` so that the below recurrence doesn't need to specify the same path again.

ci_scripts/ubuntu-19_04-rpgp/build.sh
```diff
@@ -0,0 +1,63 @@
...
+mv extract/opt/DeltaChat/deltachat-desktop extract/opt/DeltaChat/deltachat-desktop-electron
+cp /shared-build-context/deltachat-desktop extract/opt/DeltaChat
+
+# Repack into .deb
+rm $NAME_DEB
+dpkg-deb -b extract $NAME_DEB
+
+# Clean up extract folder
+rm -rf extract
+
```
hpk42: is this build.sh different from the shared one?

hpk42:
my comments are minor, so merging already. Great work!  Looking forward to .deb ubuntu releases!

Jikstra:
Just to note it down, making one (shared) build.sh which just selects another electron-build-ubuntu.json is definetly the way to go if we keep those build files and keep building like this for ubuntu. But as the rust core is maybe quite close to us using it, I don't want to put too much effort into this.

hpk:
Hi @Jikstra , could you please be aware of which Github handle you use when pinging @hpk42 ? I get notified when you use my handle (@hpk) instead of theirs.

Thanks!

PR #811, opened 2019-05-27
Progress indicator on account setup

Simon-Laux:


src/renderer/components/Login.js
```diff
@@ -19,14 +20,25 @@ class Login extends React.Component {
         showAdvanced: false,
         showPasswordMail: false,
         showPasswordSend: false
-      }
+      },
+      progress: 0
     }
+    this._updateProgress = this._updateProgress.bind(this)
+    ipcRenderer.on('DC_EVENT_CONFIGURE_PROGRESS', this._updateProgress)
```
Jikstra: To have a consistency, add the listener in the componentDidMount method, Besides that I think it's fine to use an inline function for a one liner.

src/renderer/components/Login.js
```diff
@@ -19,14 +20,25 @@ class Login extends React.Component {
...
     this.handleSubmit = this.handleSubmit.bind(this)
     this.renderPasswordInput = this.renderPasswordInput.bind(this)
   }
 
+  componentWillUnmount () {
+    ipcRenderer.removeListener('DC_EVENT_CONFIGURE_PROGRESS', this._updateProgress)
+  }
+
+  _updateProgress (ev, progress) {
+    this.setState({ progress })
```
Jikstra: Does this actually update it in the correct object path? The progress variable is in `state.ui.progress`, this seems like it puts it in `state.progress`. I assume it's wrong because for me the progress bar never updates/shows any green progress (only the grey loading one).

Simon-Laux: I didn't set the color, thats why it is grey

Simon-Laux: changed it to blue

src/renderer/components/Login.js
```diff
@@ -239,6 +258,15 @@ class Login extends React.Component {
               </div>
             </FormGroup>
           </Collapse>
+          {
+            loading &&
+            <ProgressBarWrapper>
```
Jikstra: :+1:

Issue #724, opened 2019-03-14
Disable message box when you left the group


Simon-Laux:
And have this property in the chat-state/-object to do other stuff like hiding the leave button and maybe replace it with a button to join the group again, though joining again a group is a feature- request matter on its own and as a feature proposal it should be discussed on the forum.

Edit: with `message box`  I mean the composer

Jikstra:
How is this currently done on android?

Simon-Laux:
Its not done at all on android

Simon-Laux:
Just the error message

PR #813, opened 2019-05-27
solve #724 (disable composer on left groups) for now

Simon-Laux:
#801 could probably break it again, because it relies on re-rendering

### DIFF
--- a/src/renderer/components/ChatView.js
+++ b/src/renderer/components/ChatView.js
@@ -1,5 +1,6 @@
 const React = require('react')
 const { ipcRenderer } = require('electron')
+const C = require('deltachat-node/constants')
 const styled = require('styled-components').default
 
 const Composer = require('./Composer')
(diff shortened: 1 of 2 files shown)


Simon-Laux:
I set it to wip because I get a strange react error that needs some investigation

Simon-Laux:
```
react-dom.development.js:506 Warning: Encountered two children with the same key, `97`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
    in div (created by ChatList)
    in div (created by Context.Consumer)
    in StyledComponent (created by styled.div)
    in styled.div (created by ChatList)
    in div (created by ChatList)
    in ChatList (created by SplittedChatListAndView)
    in div (created by SplittedChatListAndView)
    in div (created by SplittedChatListAndView)
    in SplittedChatListAndView (created by ScreenController)
    in div (created by ScreenController)
    in ScreenController (created by App)
    in IntlProvider (created by App)
    in App
```
seems to be something in the chatlist and is also there on current master so this pr dowsn't affect it...

Jikstra:
Let's wait for nicos pr and reimplement this based on his refactoring/rewriting.

PR #824, opened 2019-05-29
Changed default english signature text

KylePiira:
Changed "Sent with my Delta Chat Messenger" to "Sent via Delta Chat Messenger" which sounds more natural.

### DIFF
--- a/_locales/en.xml
+++ b/_locales/en.xml
@@ -349,7 +349,7 @@
     <!-- Translators: The value entered here is visible only to recipients who DO NOT use Deltachat, so its not a "Status" but the last line in the E-Mail. -->
     <string name="pref_default_status_label">Signature text</string>
     <!-- Translators: The URL should not be localized, it is not clear which language the receiver prefers and the language will be detected on the server -->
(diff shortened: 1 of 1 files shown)


Simon-Laux:
Did you change it in transifex too?

KylePiira:
@Simon-Laux Nope. I tried signing up for the Delta Chat Transifex group but it doesn't seem that I have permission to edit the source language strings.

Simon-Laux:
I think you can request access somehow and I can approve you join request (I'm not super familiar with transifex)

KylePiira:
I'm not all that familiar with Transifex either. When I joined the group it required me to pick another language to translate (I picked Spanish because English was not an option), however, after approval it seems I can only modify the Spanish strings.

Simon-Laux:
Ok changing the english strings seems to be impossible from transifex because its the source language.
I recommend filing an issue on the https://github.com/deltachat/deltachat-android repo for this string to be changed in this case.

KylePiira:
Alright will do.

PR #834, opened 2019-06-05
hot/live reload for experimental language strings

Simon-Laux:
/me wonders if this proof of concept is good enough for a merge yet

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,6 +247,9 @@ The syntax is the exact same as for all other `_locales/*.json` files.
 Example:
 `{"foobar_desktop": {"message": "This is a test"}}`
 
+> Tipp: run with `--translation-watch` (included in `npm start`) to start in translation
+ watch mode - which watches the experimental language strings and hot reloads them into dc-desktop on save 
+
 ### Tests
 
 Running `npm test` does the following:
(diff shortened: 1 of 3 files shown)


Jikstra:
I don't really see the big improvement for users/translators when using this :/

Simon-Laux:
It's rather for developers like us - we can build out it in the future to use any file and then we could build a translation editor as external program to edit those files

Jikstra:
Not sure about this, @ralphtheninja what's your opinion?

ralphtheninja:
I'm +-0

OzancanKaratas:
I have reviewed this codes. No problem in terms of translators. I think, you should merge it. We can revert this if needed.

Simon-Laux:
This only affects developers that cloned the repo and work from there, because in the release versions the experimental language file is inaccesible because its inside of an asar archive. So for making this usable for everyone we would have to add the possebility to select the file to be watched or to also watch a file in the ~/.config/DeltaChat folder

Jikstra:
> This only affects developers that cloned the repo and work from there, because in the release versions the experimental language file is inaccesible because its inside of an asar archive. So for making this usable for everyone we would have to add the possebility to select the file to be watched or to also watch a file in the ~/.config/DeltaChat folder

Can you add one simple/pragmatic way to do this? I don't know, for example a command line argument with which you can point to one language file which is getting watched? Than i would see an improvement and a useful feature.

PR #836, opened 2019-06-05
try using prebuilt node bindings

ralphtheninja:
I have tested this on windows and linux and it works well. Making a temporary PR in case someone would like to try it out on mac.


ralphtheninja:
We should remove the `SYS_DC_CORE="true"` builds.

ralphtheninja:
This clearly only works with already configured accounts due to `mail_pw` instead of `mailPw` etc.

Issue #757, opened 2019-04-10
New message counter on selected chat doesn't update immediately


Simon-Laux:
- Expected behavior: remove/update instantly when I read messages
- Actual behavior: counter keeps being there until I select an other chat or do an other refresh

could be related to #616 
version: master on 453252c9890f7082e505e813e00695be95f6e083

Issue #831, opened 2019-06-04
New group not showing up directy. It's there after restart.


ralphtheninja:
- Platform (linux/mac/windows): linux
- Delta Chat Desktop Version: chatlist-refactoring branch
- Expected behavior: group is added
- Actual behavior: group is not shown until application has been restarted
- Steps to reproduce the problem: invite someone to a group and send a message to it
- Screenshots:
- Logs:

Jikstra:
Yup this is one of the things why i haven't merged the refactoring stuff of nico yet, there's still some stuff which is not ready yet. Not sure how to proceed, as it implements quite some new stuff. We could merge it and see if we can fix it ourselves without nico?

Jikstra:
This also happens when i'm accepting a contact request. Seems like we need to redraw the chatlist or do something there.

ralphtheninja:
> We could merge it and see if we can fix it ourselves without nico?

Yes.

nicodh:
I think for now it should be sufficient to call dc.updateChatList on all events that need a complete rerendering of the chat list, in this case DC_EVENT_CHAT_MODIFIED.
Later we might improve that by just adding a new created chat.

Issue #802, opened 2019-05-22
Typing a message is slow


ralphtheninja:
- Platform (linux/mac/windows): linux
- Delta Chat Desktop Version: latest master
- Expected behavior: typing/editing a message should be smooth
- Actual behavior: it's slow
- Steps to reproduce the problem: type quickly or hold down key
- Screenshots:
- Logs:

Typing a message quickly or holding down a key (maybe erasing something using backspace) is slow. The log spits out `DC_EVENT_MSGS_CHANGED` for each key pressed:

```
main/deltachat DEBUG [] Core Event DC_EVENT_MSGS_CHANGED
main/deltachat DEBUG [] Core Event DC_EVENT_MSGS_CHANGED
main/deltachat DEBUG [] Core Event DC_EVENT_MSGS_CHANGED
main/deltachat DEBUG [] Core Event DC_EVENT_MSGS_CHANGED
main/deltachat DEBUG [] Core Event DC_EVENT_MSGS_CHANGED
main/deltachat DEBUG [] Core Event DC_EVENT_MSGS_CHANGED
main/deltachat DEBUG [] Core Event DC_EVENT_MSGS_CHANGED
main/deltachat DEBUG [] Core Event DC_EVENT_MSGS_CHANGED
main/deltachat DEBUG [] Core Event DC_EVENT_MSGS_CHANGED
main/deltachat DEBUG [] Core Event DC_EVENT_MSGS_CHANGED
main/deltachat DEBUG [] Core Event DC_EVENT_MSGS_CHANGED
main/deltachat DEBUG [] Core Event DC_EVENT_MSGS_CHANGED
main/deltachat DEBUG [] Core Event DC_EVENT_MSGS_CHANGED
main/deltachat DEBUG [] Core Event DC_EVENT_MSGS_CHANGED
main/deltachat DEBUG [] Core Event DC_EVENT_MSGS_CHANGED
main/deltachat DEBUG [] Core Event DC_EVENT_MSGS_CHANGED
main/deltachat DEBUG [] Core Event DC_EVENT_MSGS_CHANGED
main/deltachat DEBUG [] Core Event DC_EVENT_MSGS_CHANGED
main/deltachat DEBUG [] Core Event DC_EVENT_MSGS_CHANGED
main/deltachat DEBUG [] Core Event DC_EVENT_MSGS_CHANGED
main/deltachat DEBUG [] Core Event DC_EVENT_MSGS_CHANGED
[...]

PR #837, opened 2019-06-05
Chatlist refactoring

ralphtheninja:
@nicodh did some really awesome work and this contains a lot of improvements. We should merge this as is and later on @nicodh can continue on sorting.

Closes https://github.com/deltachat/deltachat-desktop/issues/788
Closes https://github.com/deltachat/deltachat-desktop/issues/757

### DIFF
--- a/src/main/deltachat/chatlist.js
+++ b/src/main/deltachat/chatlist.js
@@ -16,7 +16,9 @@ function selectChat (chatId) {
   log.debug(`action - selecting chat ${chatId}`)
   this._pages = 1
   this._selectedChatId = chatId
-  this._render()
+  const chat = this._getChatById(chatId)
+  this.sendToRenderer('DD_EVENT_CHAT_SELECTED', { chat })
+  // this._render()
 }
 
 function _chatList (showArchivedChats) {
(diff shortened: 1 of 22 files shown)


Jikstra:
What's the difference between this and https://github.com/deltachat/deltachat-desktop/pull/801?
Also there's some standard fix missing.

ralphtheninja:
The difference is that this has been rebased.

ralphtheninja:
@Jikstra Standard fixed. I went through the backlog and this closes _at least_ two issues :wink:

Jikstra:
@ralphtheninja is this pr more or less usable? Because nicos introduced some bugs/things that weren't implemented. I would love to merge this pr, but don't want to introduce half working thins :)

Simon-Laux:
We should check for those issues when merging:
Issues that might disappear with #801: (but test with c and rust core)

--Switch account while new messages are fetched results in a small freeze #702
--Chat updates (such as delete or new incomming message) result in jumping to the end #712
--Leaving an freshly created group #697
--Typing a message is slow #802 
--New message counter on selected chat doesn't update immediately #757
--If a lot of messages are incoming, don't re-render all the time #781

ralphtheninja:
@Jikstra It's definitely usable.

ralphtheninja:
Please have a look and test as well. I don't know what you mean with half working things. It puts the application in a better state than before.

Jikstra:
@ralphtheninja https://github.com/deltachat/deltachat-desktop/issues/831 this one for example. I think it's fine to merge this and just fix whatever is not working.

PR #840, opened 2019-06-05
Sort chat list on update

### LINKED ISSUE #831: New group not showing up directy. It's there after restart.

nicodh:
Update chat list when chat changes
Resolves #831

The chat list gets updated now on every DC_EVENT_CHAT_MODIFIED event.
Maybe we should test and check for more events which might not be handled...

### DIFF
--- a/src/renderer/stores/chatList.js
+++ b/src/renderer/stores/chatList.js
@@ -11,6 +11,10 @@ const chatListStore = new Store(defaultState)
 // value for text1Meaning
 const DC_TEXT1_DRAFT = 1
 
+function sortChatList (first, second) {
+  return first.summary.timestamp > second.summary.timestamp ? -1 : 1
+}
+
 // complete update of chat list
 ipcRenderer.on('DD_EVENT_CHATLIST_UPDATED', (evt, payload) => {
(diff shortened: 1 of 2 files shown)


src/renderer/stores/chatList.js
```diff
@@ -11,15 +11,19 @@ const chatListStore = new Store(defaultState)
 // value for text1Meaning
 const DC_TEXT1_DRAFT = 1
 
+function sortChatList (first, second) {
+  return first.summary.timestamp > second.summary.timestamp ? -1 : 1
```
Jikstra: I think that sorting the chatlist isn't needed, instead we should have another array which just holds the chatIds in the correct order. This we can just refetch with one call. But I will try to implement that on top of this, first lets get this bug fixed.

ralphtheninja:
A general nit pick: Please try to remember to clean up branches after merge.

PR #842, opened 2019-06-06
Forward message to any chat (instead of any contact)

### LINKED ISSUE #720: Forward message to group chat
- Expected behavior: Forward messages to group
- Actual behavior: Only contacts are shown in the forward dialog

Simon-Laux:
Fixes #720


src/renderer/components/ForwardToList.js
```diff
@@ -0,0 +1,47 @@
...
+class ForwardToList extends SearchableList {
+  constructor (props) {
+    super(props)
+    this.state.showVerifiedContacts = false
+    this.handleSearch = this.handleSearch.bind(this)
+    this.search = this.search.bind(this)
+  }
+
+  _getData () {
+    return ipcRenderer.sendSync(
```
Jikstra: sendSync is bad, but maybe okay for the moment. I would like to see the Store pattern here that @nicodh introduced.

ralphtheninja: Agreed. Lets try to use this pattern whenever we implement new things.

Simon-Laux: I use the same thing as in the contacts list... Also this isn't using the same data as the chatlist, because it's a different list.
so I see no real advantage in using a store for this. Can you explain why you wanna use it here?

nicodh: ralphtheninja already said it. We should move to the store pattern step by step if we refactor or implement new stuff. One advantage is that it's easier to maintain a software which has a homogeneous architecture than one using different pattern at different parts...

Jikstra: @Simon-Laux if you want we can look into it together tomorrow in brombi if you have time :)

nicodh: I will provide a suggestion how to implement contactList and forwardChatList with a store. Then you can review it together, maybe...

nicodh:
See https://github.com/deltachat/deltachat-desktop/tree/fix_720-with-stores

Jikstra:
Closed because of superseeding https://github.com/deltachat/deltachat-desktop/pull/849

PR #843, opened 2019-06-07
replace each occurence of i18n with tx

Simon-Laux:
with `tx` or `const tx = window.translate`
Also remove the extractor script because it's deprecated.


Jikstra:
Sehr schön!

Issue #841, opened 2019-06-05
Login not working with newest dc-node


Simon-Laux:
"Make options in bindings 1-1 with options in core " breaks compatibility with desktop. I propose a major increase of the version number, cause it breaks api.
```
main/index ERROR [ { columnNumber: 7,
    lineNumber: 27,
    fileName:
     '/home/simon/Documents/Projects/deltachat-desktop/deltachat-desktop/src/main/index.js',
    functionName: 'process.on',
    source:
     '    at process.on (/home/simon/Documents/Projects/deltachat-desktop/deltachat-desktop/src/main/index.js:27:7)' },
  { columnNumber: 15,
    lineNumber: 187,
    fileName: 'events.js',
    functionName: 'process.emit',
    source: '    at process.emit (events.js:187:15)' },
  { columnNumber: 27,
    lineNumber: 488,
    fileName: 'internal/bootstrap/node.js',
    functionName: 'process._fatalException',
    source:
     '    at process._fatalException (internal/bootstrap/node.js:488:27)' } ] uncaughtError { message: 'Missing .mail_pw',
  stack:
   'Error: Missing .mail_pw\n    at DeltaChat.configure (/home/simon/Documents/Projects/deltachat-node/lib/deltachat.js:107:13)\n    at dc.open.err (/home/simon/Documents/Projects/deltachat-desktop/deltachat-desktop/src/main/deltachat/login.js:45:10)\n    at binding.dcn_open.err (/home/simon/Documents/Projects/deltachat-node/lib/deltachat.js:579:9)' }
/home/simon/Documents/Projects/deltachat-desktop/deltachat-desktop/src/main/index.js:28
  throw err
  ^

Error: Missing .mail_pw
    at DeltaChat.configure (/home/simon/Documents/Projects/deltachat-node/lib/deltachat.js:107:13)
    at dc.open.err (/home/simon/Documents/Projects/deltachat-desktop/deltachat-desktop/src/main/deltachat/login.js:45:10)
    at binding.dcn_open.err (/home/simon/Documents/Projects/deltachat-node/lib/deltachat.js:579:9)
```

Jikstra:
Newest dc-node is which version?

ralphtheninja:
There isn't any newest version. I just made a test version.

@Simon-Laux We will bump the version to a new major as you suggested. It's a good idea!

Issue #692, opened 2019-02-22
Password is stored in cleartext


shark0der:
The password is stored in cleartext in `~/.config/DeltaChat/config.json` file.

- Platform (linux/mac/windows): linux
- Delta Chat Desktop Version: 0.101.0-x86_64
- Expected behavior: The password must be stored in system keyring or encrypted
- Actual behavior: The password is stored in cleartext (unencrypted)
- Steps to reproduce the problem: cat ~/.config/DeltaChat/config.json
- Screenshots: -
- Logs: -

Simon-Laux:
Storing in the Mac keyring seems to be straight forward, but how should encryption be used to protect the credentials on the other systems? 
I mean if you encrypt it, the application must also store the key and that one can be used by attackers to decrypt the credentials which results in no more protection what so ever.

I personally see the mac key-chain as reasonable secure, but I wouldn't trust its windows equivalent.

Also currently the messages and logs are also still unencrypted. Our thought for the beta was that the user is responsible for securing its local data. But you're right - at least the email credentials should be stored more securely than they are stored now. Let's discuss here which methods we should use.

tie:
Plaintext password is stored in `db.sqlite` too.

```sh
cd ~/Library/"Application Support"/DeltaChat
sqlite3 */db.sqlite "select * from config where keyname like '%_pw'"
```

r10s:
there are rough ideas to use sqlcipher to encrypt the whole database, see https://github.com/deltachat/deltachat-core/issues/194

for android the app-states are completely separated; for linux/mac, however, at least the account is separated from the access from others users. not sure what the current state on windows is.

r10s:
just checked ~/.config/DeltaChat/config.json on my machine: the password is _not_ stored there, just a true/false flag.

shark0der:
Encrypting the whole database would be a nice idea. The password for the account as well as the encryption key should be stored safely though.

I've found this npm package that seems to integrate well with keychains on all 3 platforms: https://atom.github.io/node-keytar/

r10s:
thanks for the link. i also think protecting the imap-password and the private-key by the system-login-key makes some sense and does not bother the user again with a master-password or so.

[...]

Issue #687, opened 2019-02-18
deb paket prüft abhängigkeit nach  libssl.so.1.1 nicht


dxu2019:
- Platform (linux/mac/windows): 4.15.0-45-generic #48~16.04.1-Ubuntu SMP Tue Jan 29 18:03:48 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux Mint 18.2 Sonya
- Delta Chat Desktop Version: 0.101.0 Oreview
- Expected behavior: start
- Actual behavior: not start
- Steps to reproduce the problem: 
A JavaScript error occurred in the main process
Uncaught Exception:
Error: libssl.so.1.1: Kann die Shared-Object-Datei nicht Ã¶ffnen: Datei oder Verzeichnis nicht gefunden
    at process.module.(anonymous function) [as dlopen] (ELECTRON_ASAR.js:160:31)
    at Object.Module._extensions..node (internal/modules/cjs/loader.js:722:18)
    at Object.module.(anonymous function) [as .node] (ELECTRON_ASAR.js:169:18)
    at Module.load (internal/modules/cjs/loader.js:602:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:541:12)
    at Function.Module._load (internal/modules/cjs/loader.js:533:3)
    at Module.require (internal/modules/cjs/loader.js:640:17)
    at require (internal/modules/cjs/helpers.js:20:18)
    at load (/opt/DeltaChat/resources/app.asar/node_modules/node-gyp-build/index.js:18:10)
    at Object.<anonymous> (/opt/DeltaChat/resources/app.asar/node_modules/deltachat-node/binding.js:1:198)
- Screenshots:
- Logs:

- Solution:
openssl-1.1.1a # ./config --prefix=/opt/ssl111 --openssldir=/etc/ssl --libdir=lib shared zlib-dynamic && make && sed -i 's# libcrypto.a##;s# libssl.a##;/INSTALL_LIBS/s#libcrypto.a##' Makefile && make MANSUFFIX=ssl install && ln -sf /opt/ssl111/lib/libcrypto.so.1.1 /lib/ && ln -sf /opt/ssl111/lib/libssl.so.1.1 /lib/^C

- Problem: Deb-Paket prüft Abhängigkeit nicht

Issue #606, opened 2019-01-20
Build & Start on Windows


Jikstra:
Blockers:
- [ ] deltachat-core doesn't build/work on windows yet https://github.com/deltachat/deltachat-core/issues/306

homdx:
Builded binary deltachat-core need testing:
https://github.com/deltachat/deltachat-core/issues/306#issuecomment-461026117
Possible problems: Windows does not support utf-8 coding, windows uses ansi encoding. (need fix).

homdx:
I did more more then 300+ commits for testing with travis-ci build deltachat-node under Windows
Now you can reproduce and build with some manual hands command core module under Windows
My testing repository for build under Win is:
https://github.com/homdx/test-win-build

I would continue to work build under Windows, then you can help repair the assembly with version 0.40 (lates build under win only 0.39).
Please test under Windows. Thanks)

Jikstra:
Wow! Very awesome :+1: I will test it out tomorrow, need to get my hands on that one windows laptop :)

kpleines:
I've tried to install the build environment using the cygwin, pyton 3, ninja, manson etc.
The build failed with:

./step1-bzip.sh: Zeile 74: patch: Kommando nicht gefunden.
./step1-bzip.sh: Zeile 79: make: Kommando nicht gefunden.

can you help to fix this problem with my development environment? Which required package is missing?

homdx:
Hello.
All files and requirements are writed in file

https://github.com/homdx/deltachat-windows/blob/master/setup-cygwin-packages.cmd

You need install: openssl-devel, patch, dos2unix, make, cmake, libtool, automake, unzip, wget

Maybe some package, I will missing to write, but you can sea and found it at cyg-get command in this file.

testbird:
With all the news above, how large are the remaining steps to set up an experimental deltachat-desktop windows build system and releases?

kpleines:
I don't know how long it takes, but for me it is acual to complex. 
I would test the windows build if it where more easy to install. Sorry, it seems I missed some skills for that stage of developing.

Jikstra:
> With all the news above, how large are the remaining steps to set up an experimental deltachat-desktop windows build system and releases?

Still hard and unknown how it works in the end. Currently the deltachat-node bindings don't work with the compiled cygwin windows library binaries. The more exact problem is that pthread on windows might work a bit different than the one for windows and is not 100% compatible. Needs more investigation.

[...]

PR #845, opened 2019-06-07
chore: update node bindings to 1.0.0-alpha.0

ralphtheninja:
Maybe leave this hanging until we have gotten `v0.104.0` out.

Closes https://github.com/deltachat/deltachat-desktop/issues/692
Closes https://github.com/deltachat/deltachat-desktop/issues/841
Closes https://github.com/deltachat/deltachat-desktop/issues/687
Closes https://github.com/deltachat/deltachat-desktop/issues/606

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ env:
 
   matrix:
     - DC_CORE_VERSION="master" SYS_DC_CORE="false"
-    - DC_CORE_VERSION="master" SYS_DC_CORE="true"
+#    - DC_CORE_VERSION="master" SYS_DC_CORE="true"
 
 cache:
   directories:
--- a/src/renderer/components/Login.js
+++ b/src/renderer/components/Login.js
@@ -1,3 +1,5 @@
+/* eslint-disable camelcase */
+
 import React from 'react'
 import { ipcRenderer } from 'electron'
 import * as update from 'immutability-helper'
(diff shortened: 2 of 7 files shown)


src/renderer/components/Login.js
```diff
@@ -48,19 +50,18 @@ export default class Login extends React.Component {
...
+      mail_user: this.props.mail_user || '',
+      mail_pw: process.env.DC_MAIL_PW || this.props.mail_pw || '',
+      mail_server: this.props.mail_server || '',
+      mail_port: this.props.mail_port || '',
+      mail_security: this.props.mail_security || '',
+      send_user: this.props.send_user || '',
+      send_pw: this.props.send_pw || '',
+      send_server: this.props.send_server || '',
+      send_port: this.props.send_port || '',
+      send_security: this.props.send_security || ''
```
ralphtheninja: These options matches exactly 1-1 to the ones in core.

ralphtheninja:
Once this has been merged, we should go over the ci again. In theory we should be able to ship windows after this. However, needs testing as well.

ralphtheninja:
Hmm. Maybe we should just merge this into master right now and instead branch off when making the next release (the last using c-core and updated rpgp).

Jikstra:
@ralphtheninja we should first merge https://github.com/deltachat/deltachat-desktop/pull/844 otherwise we need to cherry pick on that other branch. But yes, lets get it merged asap.

.travis.yml
```diff
@@ -25,7 +25,7 @@ env:
 
   matrix:
     - DC_CORE_VERSION="master" SYS_DC_CORE="false"
-    - DC_CORE_VERSION="master" SYS_DC_CORE="true"
+#    - DC_CORE_VERSION="master" SYS_DC_CORE="true"
```
Jikstra: Why did you disable this?

ralphtheninja: Because it's not working.

Jikstra: This explains things ^^ It's not very important, we only need it if we want to use a system wide installed libdeltachat. As there's currently no such a package using this, i'm fine with breaking it.

ralphtheninja: We first need to make it happen in rust core, i.e. there's no `make install` step there yet to install the library system wide. So I figured we scrap it in desktop until that is working properly.

Jikstra: Yup i'm fine with this :) Let's merge!

PR #850, opened 2019-06-09
Fix 720 with stores

nicodh:


nicodh:
Ok I see your point, but that does not mean we can not use stores but that we might have to adapt our state.

For example in GroupBase we can not just listen to the DD_EVENT_CONTACTS_UPDATED when filtering  by query since we have two different lists of contacts. (contacts in group and contacts not in group)

But that are questions of the implementation not of the concept or pattern. I would suggest to hold a list of all contacts in the contact store singleton and do the filtering in renderer without calling dc.getContacts on any change of the query. (at least as long as filtering is so simple as it is now)

In the end the state in our stores should hold the state of the whole application. So if we have different lists of data we need different state properties for them. We still have to decide what should be part of the state and what not. For example: if I type query1 to filter the list of contacts in the group and query2 for contacts not in group, should we have these 2 queries in state or do we just consider them as a part of the UI interaction that is not needed in state?

nicodh:
I pushed a change which takes into account that there are many instances of ContactList which I didn't before. The filtering is now done in the ContactComponent depending on contacts, LIST_FLAG & query string.

If we decide to us a server side filtering we have to extend the state of the contactsStore accordingly, but at the moment I see no need for that...

I also understand better now how SearchableList was intended. We could discuss, if it is better to apply the filter in _getData or if we apply the props.filterFunction explicitly as it was before...

Simon-Laux:
It's also the chatlist for forwarding it has the same bug:
1. Put in searchterm in chatList.
2. Open forward menu

Also do we want to do everything in the renderer? searching in js is slower compared to letting the core do it.(displaying the result might be faster though...) But I still don't like that we're bypassing the core in this matter.

Also why do we need so much own reducer, again? I thought react has this shadow-dom thing?

nicodh:
We don't have to bypass the core. But if we don't we have to extend the state of the chatList store (which is fine of course).
If we do all stuff async (which is the best way to go) we need different events for different actions...

Jikstra:
I think we shouldn't do filtering in the js renderer, if it's not really needed for performance reasons. I didn't get why a store is not the proper thing? Because we only need it temporary or what is the problem?

Simon-Laux:
It's possible with stores, but we need different stores for different things. We can't use the same store for two seperate searchable lists and so on.

nicodh:
So I think this is fully functional now. I also rebased the branch.

We still have one issue with forwarding to chat when in archived chats view mode: the list of chats to forward to is also in archive mode, but I think we do not need to fix that...

Jikstra:
> So I think this is fully functional now. I also rebased the branch.
> 
> We still have one issue with forwarding to chat when in archived chats view mode: the list of chats to forward to is also in archive mode, but I think we do not need to fix that...

So if we're in archived chats view mode we can only forward to archived chats? I think that's not the behaviour I would expect, but maybe fine for the moment but definetly something to get fixed.

nicodh:
Well I would say we can skip forwarding in archived chats. You can always unarchive a chat if you want, and it's an archive so no need for forwarding I would say...

nicodh:
Otherwise we have to keep two chat lists in store: chatList, archivedChatlist...

Jikstra:
> Well I would say we can skip forwarding in archived chats. You can always unarchive a chat if you want, and it's an archive so no need for forwarding I would say...

Seems like a weird workaround for me :D

> Otherwise we have to keep two chat lists in store: chatList, archivedChatlist...

I don't know why we even want to store the chats for forwarding anyways. In my opinion we should asynchronously get them when we want to forward a message,  and only keep them as long as we're in that dialog. As soon as we exit/forward the message, forget all that stuff again.

nicodh:
Well fetching the chat list might be expensive and we have it in the store up-to date (i.e. if something changes it gets updated)
To me the feature "forward messages in archive chat" seems to be an edge use case, if it's a use case at all, that's why I considered not to support it. But 2 lists in store is not an issue and can be implemented easily, so fi we decide to support the forwarding I can do it...

src/renderer/components/ForwardToList.js
```diff
@@ -0,0 +1,52 @@
...
+      return []
+    }
+    let data = chatList
+    if (filterFunction) {
+      data = chatList.filter(filterFunction)
+    }
+    data = data.filter(chat =>
+      `${chat.name}`.indexOf(this.state.queryStr) !== -1
+    )
+    data = data.filter(chat => (!chat.deaddrop && !chat.isArchiveLink))
```
Jikstra: Is it possible to combine those 3 filters into one?

Jikstra:
> But 2 lists in store is not an issue and can be implemented easily, so fi we decide to support the forwarding I can do it...

I would prefer if we can support it. Going through your archived chats to find that one interesting Message and forwarding it to someone looks like a valid usecase to me.

PR #854, opened 2019-06-11
Remove unneeded dependencies from docker images, remove -netpgp build

Jikstra:
instructions

This speeds up ci quite a lot and makes it easier to maintain.
- Uses shared scripts for ci and deb building
- Removes docker from ci which makes linux builds faster
- Removes brew from ci which makes mac builds faster
- Refactors ci/pkg building scripts


Jikstra:
CI running on mac is now from 15minutes down to <4minutes. Thanks rust!

ci_scripts/ubuntu-19_04/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+# Install dependencies for deltachat-node
+RUN apt -y install python2
+
+# Install rust
+COPY ./install-rust.sh /tmp
+RUN chmod +x ./tmp/install-rust.sh && /tmp/install-rust.sh
+ENV PATH=/root/.cargo/bin:$PATH
+
+# Install dependencies for electron builder
+RUN apt -y install binutils
```
ralphtheninja: Do we really need any of these docker files? It should be enough to `npm install` and run `electron-builder`.

Jikstra: It's possible that we don't need them anymore, i would maybe keep one to have a more or less stable environment to build the .deb . Not sure if i want to build an ubuntu package on a arch distro.

ralphtheninja: Go for it!

ci_scripts/ci-prelude.sh
```diff
@@ -30,49 +22,15 @@ if [ -z "$TRAVIS_OS_NAME" ]; then
...
-        ./ci_scripts/shared-image-context/build-install-sasl.sh --with-openssl=/usr/local/opt/openssl
-
-        export PKG_CONFIG_PATH=/usr/local/opt/openssl/lib/pkgconfig
-        
-        # Install rust
-        curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs \
-            | sh -s -- --default-toolchain nightly -y
-        . ~/.cargo/env
+# Install rust
+chmod +x ./ci_scripts/shared-prelude-context/install-rust.sh
```
flub: git is supposed to keep permissions so this shouldn't be needed normally?  not that it does much harm.

ci_scripts/ci-prelude.sh
```diff
@@ -30,49 +22,15 @@ if [ -z "$TRAVIS_OS_NAME" ]; then
...
-            sudo ninja -v -C deltachat-core-build install
-        fi
-        rm -rf deltachat-core-build deltachat-core-src cyrus-sasl-*
-        ;;
-    *)
-        echo "Unknown OS: $TRAVIS_OS_NAME" >&2
-        exit 1
-esac
+# Include rust tools in PATH
+if [ "$TRAVIS_OS_NAME" == "linux" ]; then
```
flub: out of curiosity, why does the normal sourcing of .cargo/env not work on linux?

ci_scripts/shared-prelude-context/install-rust.sh
```diff
@@ -1,6 +1,5 @@
 #/bin/bash
 set -xe
 
-apt -y install curl
 curl https://sh.rustup.rs -sSf | sh -s -- --default-toolchain nightly-2019-03-23 -y
```
flub: the nightly version seems like something that may want to be controlled.  maybe put it in a variable, maybe even overwritable by the caller?  might also be premature optimisation

ci_scripts/shared-prelude-context/nvm-use.sh
```diff
@@ -0,0 +1,4 @@
+#/bin/bash
+set -xe
+
+nvm install v10
```
flub: The v10 part seems like to easy to get out of sync with the travis.yml file.  Can this get the node major version from a travis environment var?  would that be a good idea?  (same in install-nvm-node.sh but it wouldn't let me comment on that line)

ci_scripts/ubuntu-19_10/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+ENV BUILDER_NAME="ubuntu-19_10"
+
+# Update system
+RUN true \
+    && apt update \
+    && apt -y upgrade
+
+# Install generic dependencies
+RUN apt -y install curl wget git
```
flub: The electron-builder-*.json files indicate the packages depend on libssl and libsqlite, yet no such packages or -dev packages are installed during the building in the docker file.  That seems a little suspicious to me, but I suppose it works so I'm probably missing something.

ci_scripts/ci-prelude.sh
ralphtheninja: Afaik, it works well to source.

ci_scripts/ubuntu-19_10/Dockerfile
Jikstra: the electron-builder*.json files are wrong than ^^ Thought i modified them, maybe something went wrong. Thanks!

ci_scripts/shared-prelude-context/nvm-use.sh
Jikstra: We're not only using this file from the travis environment, so I would prefer to have the travis environment depend on this, but this would slow down the ci. So I decided to let travis to this thing, we're not changing this very often so it's fine in my opinion. But sure, it's redundant.

ci_scripts/shared-prelude-context/install-rust.sh
Jikstra: I would say premature optimisation, sure we can have a variable for it but i think it's clear enough that one liner.

ci_scripts/ci-prelude.sh
Jikstra: Good question, i just didn't try it :D

ci_scripts/ci-prelude.sh
Jikstra: Good idea!

ci_scripts/ci-build.sh
```diff
@@ -28,15 +24,6 @@ fi
 SYS_DC_CORE=${SYS_DC_CORE:-true}
```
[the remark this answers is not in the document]
Jikstra: But i think it will come again?

PR #855, opened 2019-06-12
Cleanup styles

Simon-Laux:


ralphtheninja:
Niiice!

Simon-Laux:
I'm planning to discuss this this evening with @Jikstra as he had the complete opposite idea: Moving completely to styled components. 

If we shouldn't come to a conclusion we can take the first commits of this branch and only push them.

Simon-Laux:
I'm hoping to reduce react compile time and keep the styles separate from the js code to make it easier to understand and optimize.

Jikstra:
> I'm hoping to reduce react compile time and keep the styles separate from the js code to make it easier to understand and optimize(the scss also the js files get easier to read).

What's wrong with react compile times? I think they are fast enough. For me it takes, 10 seconds, disabling source maps 7 seconds. But for developing this is not a big problem, the newly introduced watch mode only rebuilds files that have changed which is quite fast (besides sourcemaps, but they are often slow and can get disabled)

Simon-Laux:
Ok, for now we only remove the unused styles in this pr.

Issue #457, opened 2018-12-18
Dark Theme


Simon-Laux:
For the future this is a useful feature, that will requested from the users anyways sooner or later.

Simon-Laux:
#442 is also related to this

Issue #409, opened 2018-12-10
Properly move out all color/styling variables


Jikstra:
Currently we still have some variables in the components, we should maybe also use the styled-components theme thing. https://www.styled-components.com/docs/advanced#theming

Simon-Laux:
We could use the new css variables for theming and completely move away from any pre-processors:
https://css-tricks.com/css-custom-properties-theming/

Though that would mean we can't use the scss/sass functions anymore (such as lighten, darken and custom ones) and we only have the vanilla css functions:
https://www.w3schools.com/csSref/css_functions.asp

Jikstra:
Wouldn't use the css variables, they are quite shitty from what i know. Weird syntax, not all the features we need. I think styled-components are fine for what we need.

Simon-Laux:
https://ramenhog.com/blog/2017/06/07/theming-with-css-custom-properties

Simon-Laux:
> I think styled-components are fine for what we need.

I bet that they increase the compile time and we don't get the benefit of scss/sass

Simon-Laux:
https://github.com/danbahrami/react-custom-properties

Jikstra:
> I bet that they increase the compile time and we don't get the benefit of scss/sass

As all js code increases the compile time, this does too. But sass/scss will increase it even further :)
The things we're using styled components for is more than just variables, it's small styled react components which are quite helpful in my opinion. 

Theming with styled components: https://www.styled-components.com/docs/advanced#theming

Simon-Laux:
solved with #858

PR #858, opened 2019-06-13
foundation work on Theming + style cleanup

Simon-Laux:
-> anything that is already in styled components can use themed color vars
now
+some scss cleanup

Just laying some foundation for theming,
Momentary you can set your theme over in the dev console:
```
Overwrite your current theme:
window.ThemeManager.setTheme(theme_object)
Reset your active Theme:
window.ThemeManager.setTheme({})
Get current Theme:
window.ThemeManager.getCurrentTheme()
```
It's automatically saved in local storage on change.


src/renderer/Theming.js
```diff
@@ -0,0 +1,77 @@
...
+  constructor () {
+    super()
+    const themeJSON = window.localStorage.getItem('theme')
+    this.currentTheme = themeJSON !== null ? JSON.parse(themeJSON) : {}
+
+    window.ThemeManager = this // only for using fron the dev console
+  }
+
+  setTheme (theme) {
+    window.localStorage.setItem('theme', JSON.stringify(theme))
```
Jikstra: Can we use rc for this? This way you can set the theme via the config.ini file

src/renderer/Theming.js
```diff
@@ -0,0 +1,77 @@
...
+    this.state = {
+      theme: manager.getCurrentTheme()
+    }
+    this.update = this.update.bind(this)
+  }
+
+  componentDidMount () {
+    manager.addListener('update', this.update)
+  }
+
```
Jikstra: Does it really need this much? Isn't something like (just an idea, no actual code) enough:
```
const lightTheme = {...}
const darkTheme = {...}

...
function ThemeProvider {
   const [selectedTheme, setSelectedTheme] = useState('light')
   dc.on('DC_SELECT_THEME', setSelectedTheme)
   return (<StyledThemeProvider theme={rc.theme === 'light' ? lightTheme : darkTheme}>
      {this.props.children}
    </StyledThemeProvider>)
}
```

src/renderer/Theming.js
```diff
@@ -0,0 +1,77 @@
...
+  deltaChatMessageBubbleSelf: '#efffde',
+  deltaChatMessageBubbleOther: '#ffffff',
+  deltaInfoMessageBubbleBg: '#000000',
+  deltaInfoMessageBubbleColor: 'white',
+  deltaChatMessageBubbleSelfStatusColor: '#4caf50',
+  deltaFocusBlue: '#42A5F5',
+  deltaSelected: '#4c6e7d' // same as deltaPrimaryBgLight
+})
+
+class ThemeManager extends EventEmitter {
```
Jikstra: At least I don't really understand why we need a ThemeManager/EventEmitter

Jikstra:
Very cool that you're pushing Theming!

src/renderer/Theming.js
Simon-Laux: Thats my quick 'n (maybe) dirty approach to make it changeable at runtime

src/renderer/Theming.js
Simon-Laux: I'd like to have more than one theme / maybe custom themes. But doing it from the backend would be good for working with files.

src/renderer/Theming.js
Simon-Laux: I like the idea to use a config file and/or theme file. What about a steam like aproach? having a theme folder where users can add theme files. that then get displayed in an theme dropdown in the settings. But on the other hand I like the option to create your theme from the the dev console so maybe a combination of both methods... Or just a load from file option in the front end like upload a `theme.yaml` file to import it. And download to export it...

Jikstra: I would prefer having it in the config folder and not in some browser/electron storage (as an upload/download would suggest). But for the moment it's not about making customized themes but integrating a dark theme.

Simon-Laux: > But for the moment it's not about making customized themes but integrating a dark theme.

Why not set the foundation now? we don't have to show the option to the user. I'll see in how far it makes sense. I mean we don't need a theme editor now for example. But a folder with theme files from that you get a drop down which the user can select from in the settings and maybe the option to change it from the dev console seems doable and not too much out of scope...

Simon-Laux:
I have a few Ideas on how the theme can be loaded, we can hold an InRealLife/IRC meeting or a/v call about this topic. @Jikstra @MIntrovrt
My current Opinion:
- I like to have the possibility to change the current theme on the fly, like over the js api I implemented now or over watching and hot reloading the theme file (when watching make it an command line option.)
- I also like having the themes and which theme to use in files rather than local storage.

Jikstra:
> I also like having the themes and which theme to use in files rather than local storage.

+1 we should put as less stuff in local storage as possible. The gold standard would be to move a `.config/DeltaChat` at a completely different system and being able to continue as nothing would have happend.

> I like to have the possibility to change the current theme on the fly, like over the js api I implemented now or over watching and hot reloading the theme file (when watching make it an command line option.)

Makes sense. Also i think it would be good to be able to add some global css to overwrite more stuff than just changing some color variables. But, maybe that's hard because of the random names of StyledComponents (that's a minus point for them :P)

Simon-Laux:
`.config/DeltaChat` - the local storage is also saved there as far as I know

> to overwrite more stuff than just changing some color variables

thats the point where you probably also want to change some structure and some html so forking will be fine.
Something that we could provide after color theming and changing chat view background.. Is changing emoji-&normal fonts but to cram them into theme files we would need folders or zip archives so I would rather provide them all as separate options.

Simon-Laux:
We could have 3 standard options in the menu:
`Use System style (dark or light)`
`Dark Theme`
`Light Theme`
And then all themes from the theme folders (one user folder in config and one in the deltachat program directory for more included ones)

Simon-Laux:
related: #409, #457

Simon-Laux:
As this cleans up more css and doesn't do anything by default, I suggest merging it noalso @Jikstra's Theming efforts can use the new variables so its easier to add theming later.
#889

PR #865, opened 2019-06-15
Reset state when closing key transfer dialog

### LINKED ISSUE #636: Trying to send a autocrypt message a second time opens up old dialog
This isn't necessarily bad, but it's not possible to trigger a second sending of an autocrypt message.

How to reproduce: Send an autocrypt message, try to send a new one again :)

nicodh:
resolves #636

### DIFF
--- a/src/renderer/components/dialogs/KeyTransfer.js
+++ b/src/renderer/components/dialogs/KeyTransfer.js
@@ -72,6 +72,7 @@ class KeyTransfer extends React.Component {
       loading: false,
       key: false
     }
+    this.onClose = this.onClose.bind(this)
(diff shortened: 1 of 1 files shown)


src/renderer/components/dialogs/KeyTransfer.js
```diff
@@ -88,26 +89,32 @@ class KeyTransfer extends React.Component {
     ipcRenderer.removeListener('initiateKeyTransferResp', this.initiateKeyTransferResp)
   }
 
+  onClose () {
+    const { onClose } = this.props
+    this.setState({ key: null })
+    onClose()
```
ralphtheninja: `this.props.onClose()`?

PR #876, opened 2019-06-28
Electron upgrade

nicodh:
Besides upgrading to electron 5 and babel 7 I also did some cleanup, removing unnecessary polyfills.

We still have 2 vulnerabilities coming from hallmark dependencies:
remark-git-contributors > git-contributors
the latter being very old (> 4 years) and using a vulnerable lodash version.
Not a real security issue at all  - but would be nice to get rid of that...

There seems no option to supress the warning just for on warning yet: https://github.com/npm/npm/issues/20565

Please test the upgrade on linux...


.babelrc
```diff
@@ -1,11 +1,13 @@
...
+    "@babel/preset-react"
+  ],
   "env": {
     "test": {
       "plugins": [
-        "transform-es2015-modules-commonjs",
-        "transform-es2015-classes",
-        "transform-object-rest-spread"
+        "transform-modules-commonjs",
+        "transform-classes"
```
nicodh: Not sure if we need that anymore. Maybe remove it and readd it only when writing more tests and it turns out we need it?

package.json
```diff
@@ -44,22 +44,23 @@
...
     "class-autobind": "^0.1.4",
     "debounce": "^1.2.0",
     "deltachat-node": "1.0.0-alpha.0",
     "emoji-js-clean": "^4.0.0",
     "emoji-mart": "^2.11.1",
     "error-stack-parser": "^2.0.2",
-    "filesize": "^3.6.1",
-    "immutability-helper": "^3.0.0",
+    "filesize": "^4.1.2",
+    "hallmark": "^0.1.0",
```
Simon-Laux: why did you move hallmark to the normal dependencies?

nicodh: oops - that was not on purpose. I just checked if the vulnerability warnings would vanish without hallmark and readded it wrong

vweevers:
> We still have 2 vulnerabilities coming from hallmark dependencies:
> remark-git-contributors > git-contributors

I'll fork `git-contributors` and get that into `remark-git-contributors`; I want to tweak some things anyway.

.babelrc
ralphtheninja: Yep

ralphtheninja:
@nicodh I'll try this PR on windows as well

PR #884, opened 2019-07-04
Restyle dialogs

Jikstra:
Before:
dialogs-before

After:
dialogs-after

### DIFF
--- a/scss/modules/_bp3-dialogs.scss
+++ b/scss/modules/_bp3-dialogs.scss
@@ -1,0 +1,25 @@
+.bp3-dialog {
+  background: #ececec;
+  border-radius: 2px;
+}
+
+.bp3-dialog-body {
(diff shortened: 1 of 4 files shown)


nicodh:
Hmm, in this design I would prefer smaller margins - like 4px instead of 20...

Why did you redesign? Not enough space somewhere?

Jikstra:
> Why did you redesign? Not enough space somewhere?

No just ugly and not matching colors/wasting space because of too much padding.

scss/modules/_bp3-dialogs.scss
```diff
@@ -0,0 +1,31 @@
+.bp3-dialog {
+  background: #ececec;
```
Simon-Laux: please use scss var here theming :wink:

PR #892, opened 2019-07-12
Make search case insensitive

### LINKED ISSUE #891: Case insensitive search in chat-list
I suggest to make the search case-insensitive with the chat-list.  In Android the search seems to be case-insensitive.

nicodh:
resolves #891


hpk42:
thanks!

PR #898, opened 2019-07-15
fix read reciept

Simon-Laux:
We still need to fix the defaults in the node bindings that it is enabled by default, but otherwise it works now. The fix was easier than expected.

### DIFF
--- a/src/main/deltachat/chatlist.js
+++ b/src/main/deltachat/chatlist.js
@@ -21,11 +21,9 @@ function selectChat (chatId) {
     if (chat.freshMessageCounter > 0) {
       this._dc.markNoticedChat(chat.id)
       chat.freshMessageCounter = 0
-      if (this._saved.markRead) {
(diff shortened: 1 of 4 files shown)


src/main/deltachat/chatlist.js
```diff
@@ -21,11 +21,9 @@ function selectChat (chatId) {
     if (chat.freshMessageCounter > 0) {
       this._dc.markNoticedChat(chat.id)
       chat.freshMessageCounter = 0
-      if (this._saved.markRead) {
-        log.debug('markSeenMessages', chat.messages.map((msg) => msg.id))
-        this._dc.markSeenMessages(chat.messages.map((msg) => msg.id))
-        app.setBadgeCount(this._getGeneralFreshMessageCounter())
-      }
+      log.debug('markSeenMessages', chat.messages.map((msg) => msg.id))
+      this._dc.markSeenMessages(chat.messages.map((msg) => msg.id))
```
Jikstra: can you save chat.messages.map... to a variable?

PR #904, opened 2019-07-24
Chat List Item: last Message status

Simon-Laux:
- [X] add real message status to chatlist items
- [X] Add ~an icon~ a style for draft status
- [x] Fix Bug: Chat list message status icon doesn't update right away (draft and also on sending done) @nicodh can maybe help with this one because this needs to fit in with the optimizations he made.


nicodh:
I will have a look on the icon issue - but not before friday...

nicodh:
I did a force push to rebase the PR with all the standard js fixes...

PR #863, opened 2019-06-14
Move LoginScreen in own file, fix importing backups

Jikstra:
Missing to complete:
- Updates deltachat-node to v1.0.0-alpha.5
- Moves LoginScreen in own file
- Fixes mounting all dialogs all the time
- Implements Importing of Backups on login screen

closes #815 
fixes #506
fixes #763

PR #911, opened 2019-07-25
Theming efforts

Simon-Laux:
The goal of this pr/sprint is to 

- [x] 1. Centralize all component color vars (no general non-telling var anymore)
- [x] 2. Create an abstraction layer with vars that have meaningfull names and convert from them to the component colors.
- [x] 3. Create some sample themes to test if we did it right

- [x] If #863 is getting merged we need to change the deltaFocusBlue variable name


Simon-Laux:
at this commit rate we should squash merge them in the end..

Simon-Laux:
@MIntrovrt 
we still need vars for
- vars for Bp3 Buttons (text and bg + hover )
- .bp3 input fields
- .bp3 toggle button

MIntrovrt:
I found a bunch more places where there are missing vars, namely: the text input field and names in the contact list

scss/_variables.scss
```diff
@@ -38,36 +38,151 @@ $roboto-light: Roboto-Light, $emojifonts, 'Helvetica Neue', Arial, Helvetica, No
 // New colors
 
 html {
```
Simon-Laux: These here act as fallback, but I think about removing them to save a few lines and reduce duplication..

src/renderer/ThemeSystem.md
```diff
@@ -0,0 +1,49 @@
...
+// get the theme data as you set it/ like it is stored in localstorage
+getCurrentTheme ()
+
+// get the default theme/-data for reference
+getDefaultThemeData ()
+getDefaultTheme ()
+```
+
+
+### TODO
```
Simon-Laux: the todo is here now

src/renderer/ThemeBackend.js
```diff
@@ -0,0 +1,225 @@
+var Color = require('color')
+
+function changeContrast (colorString, factor) {
+  // TODO make the black check code work
+  const color = Color(colorString).hex() === '#000000' ? Color('#010101') : Color(colorString)
+  if (color.isDark()) {
+    console.log('dark')
+    return color.lighten(factor).rgb().string()
+  } else if (color.isLight()) {
+    console.log('light')
```
Jikstra: console.log, maybe a log.debug if it's not too spammy?

src/renderer/ThemeBackend.js
```diff
@@ -0,0 +1,225 @@
...
+  } else if (factor > 0) {
+    return color.saturate(factor).rgb().string()
+  } else {
+    return color.rgb().string()
+  }
+}
+
+function undefinedGuard (value, func) {
+  if (typeof value === 'undefined') return undefined
+  else return func(value)
```
Jikstra: no need for the else, looks weird
Maybe even do `return typeof value === 'undefined' ? undefined : func(value)`

src/renderer/ThemeBackend.js
```diff
@@ -0,0 +1,225 @@
...
+}
+
+export function ThemeDataBuilder (theme) {
+  // todo resolve all strings to be dependent on vars of the highLevelObject
+  // Its ok when highLevelObject is missing some properties
+  // and the returned object misses some values as result,
+  // because it gets merged with the default theme later anyway
+  /// Don't use color directly here
+  console.log(theme.textNavBar, changeContrast(theme.textNavBar, 0.3))
```
Jikstra: log.debug

scss/_variables.scss
```diff
@@ -38,36 +38,151 @@ $roboto-light: Roboto-Light, $emojifonts, 'Helvetica Neue', Arial, Helvetica, No
...
+$color-conversation-list-unread-count-label: var(--clr-conversation-list-unread-count-label);
+$color-conversation-list-message-error: var(--clr-error);
+$color-conversation-list-avatar-text: var(--clr-avatar-label);
+$chatlist-isgroup-icon: var(--clr-global-text);
+
+/* Message Bubble */
+$color-message-author: var(--clr-message-author);
+$color-message-attachment-container-bg: var(clr-message-attachment-container-bg);
+$color-message-attachment-filename-incomming: var(--clr-message-attachment-filename-incomming);
+$color-message-attachment-filesize-incomming: var(--clr-message-attachment-filesize-incomming);
```
Jikstra: incoming only needs one `m`

scss/_variables.scss
```diff
@@ -38,36 +38,151 @@ $roboto-light: Roboto-Light, $emojifonts, 'Helvetica Neue', Arial, Helvetica, No
...
+  --clr-message-padlock-incomming: #a4a6a9;
+  --clr-message-metadata-image-no-caption: #fff;
+  --clr-message-text: #070c14;
+  --clr-message-text-underline: #070c14;
+  --clr-message-generic-attachment-icon-extension: #070c14;
+  --clr-message-generic-attachment-filename: #070c14;
+  --clr-message-generic-attachment-filesize: #070c14;
+  --clr-message-attachment-container-bg: #ffffff;
+  --clr-message-attachment-filename-incomming: #ffffff;
+  --clr-message-attachment-filesize-incomming: #ffffff;
```
Jikstra: For me it looks like --clr variables are the real base and they just get pulled in into the --color variables which allow a more specific adjustment. If this is the case, i would prefer to have only one --clr variable for every value and give it a proper name. For example there should be no need to have message-text and message-text-underline variable, that's way to verbose for base colors.

src/renderer/ThemeBackend.js
```diff
@@ -0,0 +1,225 @@
...
+    globalText: theme.textPrimary,
+    mapOverlayBg: theme.bgPrimary,
+    videoPlayBtnIcon: theme.accentColor,
+    videoPlayBtnBg: '#ffffff', // Only changable with theme.raw
+    scrollBarThumb: undefinedGuard(
+      theme.scrollbarTransparency, c => Color('black').alpha(c).rgb().string()
+    ),
+    scrollBarThumbHover: undefinedGuard(
+      theme.scrollbarTransparency, c => Color('black').alpha(c + 0.14).rgb().string()
+    )
```
Jikstra: I don't know how much work this adds that if we want to define a new variable, we need to hardcode it at several places. Also it seems to me that we can't override those variables here from a theme?

src/renderer/ThemeBackend.js
```diff
@@ -0,0 +1,225 @@
...
+export const ThemeVarOverwrite = (theme) => `
+--clr-avatar-label: ${theme.avatarLabelColor};
+--clr-bp3-dialog-bg: ${theme.bp3DialogBg};
+--clr-bp3-dialog-card-bg: ${theme.bp3DialogCardBg};
+--clr-bp3-heading: ${theme.bp3Heading}
+--clr-bp3-button-text: ${theme.bp3ButtonText};
+--clr-bp3-button-bg: ${theme.bp3ButtonBg}
+--clr-bp3-dialog-header-bg: ${theme.bp3DialogHeaderBg}
+--clr-broken-media-bg: ${theme.brokenMediaText};
+--clr-broken-media-bg-text: ${theme.brokenMediaBg};
```
Jikstra: Even more hardcodes with different naming conventions and names, that's confusing. Can we streamline this? For example start using camelCase everywhere?

Simon-Laux: This is setting the css variables and as far as i know the naming for those is --[name in lowercase + '-'] 
https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties

src/renderer/ThemeBackend.js
Simon-Laux: we need to hardcode it once for styled components and 3-4 times for scss

scss/_variables.scss
Simon-Laux: $colors you mean? we used this to resolve and group them centraly, we might resolve them again (put the `var(--clr-component-name)` directly in the components).
> For example there should be no need to have message-text and message-text-underline variable
I think one might want to change the link color independently from the message text color (telegram has a different color for links)

src/renderer/ThemeSystem.md
Simon-Laux: @Jikstra have you read this file?

src/renderer/ThemeBackend.js
Simon-Laux: the plan is to remove it after we fixed a problem with this function

src/renderer/ThemeBackend.js
Simon-Laux: same here this is a debug statement that is deprecated now

src/renderer/ThemeBackend.js
Simon-Laux: Also for scss: we could just transfer the ("highlevel") theme vars and resolve them in scss, but some of them need additional transformation that must be done in js, because at runtime we don't have the fancy scss functions.

scss/_variables.scss
Simon-Laux: Maybe you want to have different color shades for filename and filesize... but for this one we might get away with setting opacity for this effect, then we only need one var and if message-text-collor == filename-color == filesize-color we don't need any extra var for it.

Simon-Laux:
We merge this now, but keep in mind that there are still somethings left to do for good theming.

src/renderer/ThemeBackend.js
Jikstra: @Simon-Laux what kind of additional transformation do they need?

src/renderer/ThemeBackend.js
Jikstra: I would propose following ideas:
1. use snake case name everywhere
2. use snake case for css custom properties, automatically (with help of a method) translate it into a camelCase name
3. use camelCase everywhere as it seems to work: https://codepen.io/anon/pen/xvXPWN?&editable=true

src/renderer/ThemeSystem.md
```diff
@@ -0,0 +1,65 @@
...
+```
+
+## Hot reload:
+> Warning: This will also (probably) change in the future.
+```
+npx electron . --theme-watch "./themes/dark.json"
+```
+Quirks:
+- You're theme gets only updated and loaded on change of the specified theme file
+- Doesn't work with `npm run dev` because this argument gets lost
```
Jikstra: We should fix this to make npm run dev allow arguments.

src/renderer/ThemeBackend.js
Simon-Laux: The actual theme gets generated here from the "highlevel" theme the user specified, after that when the "raw" object is set, its values replace the the generated values, so you still can have whole controll of it.

src/renderer/ThemeBackend.js
Simon-Laux: you only need to "hardcode"/define it in one place and thats `ThemeBackend.js`, I don't see it as big deal. Also then you have the advantage that you basicaly have a directory of every themeable var which you can use as reference for creating a theme.
If you have an scss var you define it twice sometime additionally in `_variables.scss` we could probably resolve the ones in  `_variables.scss` so we only have the ones left in `ThemeBackend.js`.
You're right we could generate the var overwrite to save even more space. I'll try some things.

PR #915, opened 2019-07-28
Filter accessible functions

nicodh:
Just to make sure it's not possible to call "any" method on deltachat via EVENT_DC_FUNCTION_CALL event from frontend context

### DIFF
--- a/src/main/deltachat/index.js
+++ b/src/main/deltachat/index.js
@@ -5,6 +5,27 @@ const log = require('../../logger').getLogger('main/deltachat')
 const windows = require('../windows')
 const { app } = require('electron')
 
+const publicAccessibleMethods = [
+  'archiveChat',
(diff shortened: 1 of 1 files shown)


Jikstra:
Very important! Thanks :)

src/main/deltachat/index.js
```diff
@@ -43,13 +64,17 @@ class DeltaChatController extends EventEmitter {
...
   *
   * @param evt
   * @param methodName
   * @param args
   */
   handleRendererEvent (evt, methodName, args) {
+    if (publicAccessibleMethods.indexOf(methodName) < 0) {
+      const message = 'Method not accessible: ' + methodName
+      log.error(message)
+      throw (message)
```
ralphtheninja: Maybe do `throw new Error(message)` instead?

PR #918, opened 2019-07-29
Restyle deaddrop dialog & confirmationDialogs

Jikstra:
- Restyle deaddrop dialog:
   - before:
deaddrop dialog before
   - after:
deaddrop dialog

- This also added a new ConfirmationDialog which is now rendered inside the main window instead of opening a new window. For now i only replaced the ones in the ChatListContextMenu, so leaving/deleting/blocking now uses the new ConfirmationDialog.
  - before:
confirmation dialog before
  - after
Screenshot_20190731_153231

- Refactor ContextMenu and Menu

@r10s i'm currently not using the color that the android app uses, can you put that green color in the interface repo?


Simon-Laux:
Can you add the chats name to the dialog?

Jikstra:
> Can you add the chats name to the delete dialog?

Not really as the translations don't have a placeholder for this

Simon-Laux:
> > Can you add the chats name to the delete dialog?
> 
> Not really as the translations don't have a placeholder for this

I think it's important... Can't you make one as dialog header? (https://material.angular.io/components/dialog/overview)
```
Delete [Chat name]
```

Jikstra:
> I think it's important... Can't you make one as dialog header? (https://material.angular.io/components/dialog/overview)

Sure but than the delete needs to be translated an this pr is more about restyling. The old dialog had the same problem that the dialog didn't say which chat/group you're currently deleting. But I agree that this would be a useful improvement, but i would prefer to have it in the text instead of using an extra headline (it looks shitty).

Simon-Laux:
Can you make the buttons bold as well? I like the color idea, we can add a danger color in the theming for error messages and delete button color.
Also logicaly it makes more sense to me to align the middle button(not now) to the left becauae its more similar to never than to yes

Jikstra:
> Can you make the buttons bold as well? I like the color idea, we can add a danger color in the theming for error messages and delete button color.
Also logicaly it makes more sense to me to align the middle button(not now) to the left becauae its more similar to never than to yes

I like it, can we do all those things in a second pr? And streamline it with your theming efforts?

src/renderer/components/Menu.js
```diff
@@ -1,160 +1,103 @@
-const React = require('react')
-const { ipcRenderer } = require('electron')
-const autobind = require('class-autobind').default
-const confirmation = require('./dialogs/confirmationDialog')
+import React, { useContext } from 'react'
```
Simon-Laux: Does import for react work now?

Jikstra: Yup, already since some months ^^

PR #922, opened 2019-08-03
CI test for misplaced console.log statements

Simon-Laux:


bin/log-conventions.js
```diff
@@ -0,0 +1,57 @@
+require('colors')
```
Jikstra: Nice :) Output looks really good in the ci! Failing tests are now visually appealing ^^

bin/log-conventions.js
```diff
@@ -0,0 +1,57 @@
...
+
+walker.on('file', function (root, fileStats, next) {
+  if (!fileStats.name.includes('.js')) next()
+  const filename = path.join(root, fileStats.name)
+  fs.readFile(filename, 'utf-8', function (err, data) {
+    if (err) throw err
+    const lines = data.split('\n')
+    for (let i = 0; i < lines.length; i++) {
+      const line = lines[i]
+      if (line.indexOf('console.') !== -1) {
```
Jikstra: I don't really like the right drift here, more a coding style thing. Will try to reimplement it a bit nicer.

package.json
```diff
@@ -7,7 +7,7 @@
     "watch": "npm run watch-other && npm run watch-js",
     "watch-other": "node bin/watch &",
     "watch-js": "webpack --mode=development --progress --profile --colors --watch",
-    "test": "standard && node test/unit",
+    "test": "standard && node ./bin/log-conventions && node test/unit",
```
Jikstra: Let's split this into three subcommands. test-standard, test-unit, test-log-convenctions.

Simon-Laux: feel free to continue this... I'm done for today (still awake from yesterday)

Jikstra: I will apply some changes, get some rest :) Will very likely be at brombi tuesday/wednesday

Jikstra:
I think this could be fine for now, but if we ever have problems with this implementation, we should use https://eslint.org/docs/rules/no-console for this. Maybe we can adjust standard to do this too, otherwise i would prefer to switch to eslint and use the standard rules combined with no-console.

PR #931, opened 2019-08-05
copy selected message text to clipboard

### LINKED ISSUE #930: Right click - Copy selected text
like in telegram if you have selected text you get this option in the context menu

Simon-Laux:
closes #930

### DIFF
--- a/_locales/_experimental_en.json
+++ b/_locales/_experimental_en.json
@@ -16,5 +16,8 @@
   },
   "pref_on_demand_location_streaming": {
     "message": "On-demand location streaming"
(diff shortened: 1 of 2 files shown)


_locales/_experimental_en.json
```diff
@@ -16,5 +16,8 @@
   },
   "pref_on_demand_location_streaming": {
     "message": "On-demand location streaming"
+  },
+  "copy_selected_text": {
+    "message": "Copy selected Text"
```
Jikstra: text should be lower cased

PR #933, opened 2019-08-11
WIP: Replace camelCase with snake_case

### LINKED ISSUE #925: Settings are not saved or restored correctly
- Delta Chat Desktop Version: commit f00ec6532c6be58fd093984c57bf66cc1e624349
- Expected behavior: when settings are changed and settings dialog is opened, the dialog should use those settings
- Actual behavior: settings are saved but not restored to the settings dialog when it's opened
- Steps to reproduce the problem: change for example "Profile Info" and close the dialog and re-open it
- Screenshots:
- Logs:

nicodh:
resolve #925

### DIFF
--- a/src/renderer/components/Login.js
+++ b/src/renderer/components/Login.js
@@ -73,7 +73,7 @@ export default class Login extends React.Component {
       }
     })
     const dirty = Object.keys(updatedState.credentials).find(
-      key => this.props[key] != updatedState.credentials[key]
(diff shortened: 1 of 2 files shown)


nicodh:
Now settings are loaded again in the form, but not updated. No idea why setState was set in componentDidUpdate which is an anti pattern anyway and with the filter if (this.props.isOpen && !prevProps.isOpen) seems never to be called...

Issue #924, opened 2019-08-04
Theming prs broke some styling


Jikstra:
The interface got a bit uglier with the theming pr, guess there are some bugs.

Simon-Laux:
you mean the lightgreen background?

nicodh:
the background of the chat view (messages) - it's not light green but "poison green" ;-)
Why did that change? The background looked not bad and was/should be the same as in the app (as default)

Simon-Laux:
for testing and I forgot to change it back to `#e6dcd4` from `lime`

Jikstra:
There's more broken:
- scrollbarcolor/hover in chat list
- color of checkboxes in settings
- the menu on the three dots
- color of callouts
- probably more

Simon-Laux:
I mention the todo section in the `THEMES.md` doc file which contains more things that need to be included into theming.
Also can you be more specific than saying broken here? what do you mean by broken? I would rather use words like missing in the theme system and looking bad / needs adjustment.

Jikstra:
Looks bad because of wrong colors ^^

PR #934, opened 2019-08-12
[wip] Theming fixes

Simon-Laux:
maybe that fixes some of #924 
let's fix even more things on this branch before merging @MIntrovrt @Jikstra 

EDIT by @Jikstra:
ToDo:
- [x] color of scrollbar
- [x] color of menu
- [ ] something is weird with the toggles in settings (i think you improved them, but the blue is a bit aggressive, you know that i'm quite opinionated and picky on those things ^^)
- [x] the background of callouts
- [x] the section after the callout is white

from the todo in `docs/THEMES.md`:
- [x] map controlls
- [x] emoji picker border can be a bit more pronounced
- [x] main menu rightside triangle/arrow
- [ ] toggle hover
- [ ] gradients? namely:
    - [ ] bp3 input focus
    - [ ] gradient between chatlist and chatview
- [ ] see todo comments in _bp3_overwrites.scss
     - [ ] dialog shadows in darkmode look weird
     - [ ]  main menu shadow thingie looks weird in dark modes
     - [ ] map controlls hover
     - [ ] main menu seperation bars


scss/modules/_bp3-overwrites.scss
```diff
@@ -88,4 +95,21 @@
...
   color: var(--bp3MenuText);
+}
+
+.bp3-popover .bp3-popover-content {
+  background: unset;
+}
+
+.bp3-popover {
+  -webkit-box-shadow: none;
+  box-shadow: 0 0 0 1px rgba(16, 22, 26, 0.1), 0 2px 4px rgba(16, 22, 26, 0.2), 0 8px 24px rgba(16, 22, 26, 0.2); // TODO include color into theming
```
Jikstra: We maybe need to simplify this and make it themeable at some point. But not really sure how (besides having multiple variables for this)

Simon-Laux: maybe we can set some of these problematic styles in an scss override in theme data like defining a value in the raw section:
(semi pseudo-code)
`defaultTheme.shadowColor = "rgb(16, 22, 26)"`
```
bp3PopoverBoxShadow: undefinedGuard (
theme.shadowColor, c => `0 0 0 1px ${Color(c).alpha(0.1).rgb}, 0 2px 4px ${Color(c).alpha(0.2).rgb}, 0 8px 24px ${Color(c).alpha(0.2).rgb}`
),
```
and then while setting the css for the variables:
```
.bp3-popover {
  -webkit-box-shadow: none;
  box-shadow: ${themeData.bp3PopoverBoxShadow};
}
```

Jikstra:
What's still wrong:
- [ ] color of scrollbar
- [ ] color of menu
- [ ] something is weird with the toggles in settings (i think you improved them, but the blue is a bit aggressive, you know that i'm quite opinionated and picky on those things ^^)
- [ ] the background of callouts

Basically they all have the wrong color which is a weird grey instead of a white.

Jikstra:
@Simon-Laux during fixing the scrollbar i encountered some weird things:
- i had to overwrite the scrollbarThumb and scrollbarThumbHover in the raw section, the undefined guard only calculated it from scrollbarTransparency. Why?
- i had to manually run `npm start -- --theme="./themes/light.json"` to actually see changes, even when restarting. Does the theming manager cache the themes somehow?

Simon-Laux:
> i had to overwrite the scrollbarThumb and scrollbarThumbHover in the raw section, the undefined guard only calculated it from scrollbarTransparency. Why?

I think if we can do that we would have less variables to care about in the highlevel theme object, but as the calculation is off we would need to **change it to match** or **throw it out and make those two vars accessible without using `raw`**.

One note on what the undefined Guard is for:
When you don't have a value in your highlevel object its undefined and so is the resulting low level themedata property that take this value.
Before applying the `undefined` fields of your lowlevel theme data are filled with the values of the default lowlevel theme data.
But when using functions like color undefined gets translated into `black`, so we have the undefined guard which just returns `undefined` if one of the input/parameters is `undefined`.

>i had to manually run npm start -- --theme="./themes/light.json" to actually see changes, even when restarting. Does the theming manager cache the themes somehow?

It still stores the current theme in localstorage, we haven't changed that yet. Also we don't have the logic to load the theme from a file yet, the `--theme` only injects the themedata (basically calls `setTheme`)

Jikstra:
Thanks for your explanation, very helpful!
As the todo section got quite long in this pr, i would only fix the light theme and do the dark theme in another pr. Would like to fix the toggles but that's it. It has not to be perfect, at least not yet ;)

PR #939, opened 2019-08-17
Refactor desktop settings

### LINKED ISSUE #925: Settings are not saved or restored correctly

nicodh:
resolves #925

### DIFF
--- a/src/renderer/components/SplittedChatListAndView.js
+++ b/src/renderer/components/SplittedChatListAndView.js
@@ -185,9 +185,9 @@ class SplittedChatListAndView extends React.Component {
                   <Button minimal icon='map' onClick={this.onMapIconClick} />
                 )}
               </SettingsContext.Consumer>
-              }
+              }""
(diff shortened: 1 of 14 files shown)


src/renderer/components/SplittedChatListAndView.js
```diff
@@ -185,9 +185,9 @@ class SplittedChatListAndView extends React.Component {
                   <Button minimal icon='map' onClick={this.onMapIconClick} />
                 )}
               </SettingsContext.Consumer>
-              }
+              }""
```
Simon-Laux: remove this

PR #946, opened 2019-08-20
Open file on click

Simon-Laux:
..When clicking on the file icon (the same one you drag to drag out the file.


Jikstra:
Opening random attachments from e-mail always is a problem. But I think it's a feature too.

PR #948, opened 2019-08-21
Update electron to v6.0.3 (fixes #895)

### LINKED ISSUE #895: Upgrade to electron6 when electron/#17972 is fixed
https://github.com/electron/electron/issues/17972

And maybe also others...

Jikstra:
fixes #895

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1325,9 +1325,9 @@
       }
     },
     "@types/node": {
-      "version": "10.14.10",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-10.14.10.tgz",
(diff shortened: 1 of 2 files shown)


Jikstra:
I'm currently using electron 6.0.3 for developing but i just had random crashed without any log/trace. So i think it makes to wait until electron 6 is getting more stable.

Simon-Laux:
It isn't fixed, still same error:
```
[16993:0901/205912.105102:FATAL:setuid_sandbox_host.cc(157)] The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that /home/simon/Documents/Projects/dcd-debug-env/deltachat-desktop/node_modules/electron/dist/chrome-sandbox is owned by root and has mode 4755.
```

Jikstra:
> It isn't fixed, still same error:

Pff...

PR #952, opened 2019-08-23
Rework encryption info dialog and add method to async call dc methods over ipc

Jikstra:
- Implement callDCMethod to asynchronously (without blocking renderer) call a dc method and send back the return value over ipc. The syntax for doing so is quite easy:
```
callDcMethod('functionName', [arg1, arg2...], (returnValue) => {...})
```
- Rewrite and restyle EncryptionInfo dialog
Screenshot_20190823_180638


Jikstra:
> We should consider not having different approaches in parallel...

Agree. I'm going to change the names a bit to fit your changes but as discussed we need to get over this again once you're back.

PR #957, opened 2019-08-27
Rework create/edit group/chat

Jikstra:
Missing:
- [x] implement search
- [x] implement create group
  - [x] adding/removing group members over dialog search
  - [x] setting group name
  - [x] setting group picture
  - [x] unset group picture
  - [x] showing qr code
  - [x] adding group members over own dialog
  - [x] add members to group
  - [x] properly set groupImage
  - [x] delete group on aborting/closing dialog
- [x] implement create verified group
- [x] implement add contact
- [x] fix slow rerendering if only one chat changes (implements partial update on proper DC_EVENT_CHAT_CHANGED event). This greatly improves the performance
- [x] fix multiple chatlistitems with same key
- [x] refactor EditGroup & GroupBase
- [x] separate group related stuff from create chat related stuff in CreateChat styles
- [x] fix QR code view
- [ ] update chat view header after editing a group (to show updated group properties)
- [ ] update deltachat-node

ToDo (probably not in this pr):
- [ ] when people join over qr in group creation dialog, update member list
- [ ] show toast/error when trying to create group or qr code without a groupname
- [ ] fix grey border bug when scrolling all the way down


src/renderer/stores/contacts.js
```diff
@@ -16,11 +17,11 @@ ipcRenderer.on('DD_EVENT_BLOCKED_CONTACTS_UPDATED', (evt, payload) => {
   contactsStore.setState({ ...state, blockedContacts })
 })
 
-ipcRenderer.on('DD_EVENT_CONTACTS_UPDATED', (evt, payload) => {
+ipcRenderer.on('DD_EVENT_CONTACTS_UPDATED', debounce((evt, payload) => {
```
Jikstra: Todo: I should remove this again.

src/renderer/components/ChatList.js
```diff
@@ -110,56 +110,37 @@ class ChatList extends React.Component {
...
-                      name={chatListItem.name}
+                      chatListItem={chatListItem}
                     />
                   </ArchivedChats>
                 )
               } else {
                 return (
                   <ChatListItem
-                    key={chatListItem.id}
+                    chatListItem={chatListItem}
```
nicodh: why did you remove key attribute here? It might be an performance issue when reordering the chat list or when inserting new items at the beginning if react has no key to easy "identify" unchanged items. https://reactjs.org/docs/reconciliation.html#recursing-on-children
I know we have an issue with duplicate keys sometimes but that should not be solved by removing the key. Consider that the chat list might have hundreds of items...

Jikstra: That wasn't on purpose. I actually added the correct ones to the map values (ContactRequestItemWrapper & ArchivedChats) which fixes the multiple error problem (we had the index of the element in the array + the chatId which in some cases where the same). So this problem is fixed with the latest commit.

Jikstra:
@nicodh @Simon-Laux this is a huge pr and probably hard to review. Therefore i think it's maybe easier to do it together at bb?

nicodh:
I will do the refactoring for edit group & base group...

nicodh:
@Jikstra I would prefer to have 2 columns for editing groups, was it an explicit UI decisison to have only one column in CreateChat view?
This is roughly how edit group looks like in the dialog implementation with 2 columns now:
 
Editgroup-Screenshot

Jikstra:
> @Jikstra I would prefer to have 2 columns for editing groups, was it an explicit UI decisison to have only one column in CreateChat view?
This is roughly how edit group looks like in the dialog implementation with 2 columns now:|

Puuh. Sorry but I don't like it. Before implementing this i would prefer only adding them over the "Add member" button. As usual, i'm very opinionated and picky about ui/ux decisions.
I don't like having two columns as it spreads too much stuff over the screen and the whole width is only used in the group dialog. For selecting a contact for a chat you can't use the whole width as you do in Add Group dialogs. Switching from New Chat to Create Group will result in doubling the width of the dialog which I want to prevent.

nicodh:
Well I didn't mean to change it in the CreateChat view but only in the edit group view. So no switching of width etc. 
Usability wise I consider it a huge benefit to see members and non members side by side, that's why I would like to keep it in the edit group dialog...

Jikstra:
Okay some points why I think it's not a good idea:
1. I prefer in a ui/ux to reuse ideas/components as often as possible. In my opinion it's confusing if adding/removing members in the Create Group dialog is done differently than in the edit group dialog.
2. We have no other place in the whole ui where we show Contacts or Chats (which are visually extremely similiar) in 2 columns/left&right to each other. Doing this in one place is confusing. 
3. Opening the Edit Group dialog and having members displayed that aren't in this group is confusing. This is already problematic in the one column approach of me, but a bit better as it happens on a search input and can easily get undone, plus the checkboxes (which all are visually aligned at the same position, as it's in the same colum) makes it more clear that there's something different between members which i can add and members that i can remove. This benefit you don't have if they are spread accross two columns.

In case we can figure out a good way of the 2 column approach, i'm fine with using this approach as the general "Add member" view.

nicodh:
Ok - although I still don't get why we should reduce the usable space to 25% of the available width and add extra clicks and forth and back switches and check boxes  to the add/remove user process.
But I'm NOT very opinionated about UI stuff ;-)
So let's go for the one column. Maybe we can get other opinions in the UI sessions next days...

Jikstra:
> add extra clicks and forth and back switches and check boxes to the add/remove user process

Hmm good point! The most forth/back that we currently have is because we can't create a new (verified) group from the main menu. If we implement this, zack, one click less for creating a (verified) group. Adding/removing a member is currently possible with some keystrokes and one click which is quite less. Showing the qr code is also one click. Proof me wrong, but for me it doesn't feel like Creating a Chat/Group is a Point and Click Adventure (yet) :D 

> why we should reduce the usable space to 25% of the available width 

We don't have more to show. We have 4 things: Group image, group name, qr code and group members. We could show the qr code, but the qr code changes whenever we change the groupname. Updating the qr code whenever we change the groupname makes the ui busy and distracts the user. Splitting the create group flow into two parts, first where we enter the groupname and the second where we add groupmembers/show the qr code would solve this, but comes with the cost that we need to throw away that group to change the name and we have one click more again. Yes we can show the member lists in two columns, but we can't use the other column in the upper part for anything else. Letting the groupname take the whole width looks weird and besides that we have nothing, as the group image is tightly coupled with the groupname.

> Maybe we can get other opinions in the UI sessions next days...

:+1: 

Another point which is important in my opinion is that we don't have a ui/ux guideline for the desktop client as there is for android (material) ios (that apple thing). Both platforms have a strict styling guide and many examples how to do things. We don't have that. It's not like there's one way on how to do it right with electron. That's the reason why im copying so many things from android and the material design, because i think it works well on desktop too and because I think people are used to the android/smartphone flow. If we completely change how we interact with the desktop ui, we also need to come up with new components, buttons, maybe even colors. I don't know that shit, so i go the more safe path and try to stay as close with the android ui as possible, modifying things where it makes sense and using the telegram desktop ui as a good inspiration for a bridge between a desktop app and a mobile app.

Jikstra:
The reason why we currently cant display the qr code is because rust master returns null for the qr code.

hpk42:
in which situation does which function precisely erturn null?

PR #963, opened 2019-09-01
Fix missing/unused dependencies. Fixes #962

### LINKED ISSUE #962: Fix dependencies
Just ran depcheck and output is following:
```
$ depcheck 
Unused dependencies
* @blueprintjs/select
* class-autobind
* lodash
* moize
Unused devDependencies
* @types/mapbox-gl
* babel-loader
* chai
* depcheck
* electron-react-devtools
* sinon
* webpack-cli
* webpack-node-externals
Missing dependencies
* classnames
```

I think it's fine to ignore webpack-cli & react-devtools, babel-loader but besides that...

Jikstra:


package.json
```diff
@@ -87,28 +84,23 @@
     "@babel/core": "^7.5.5",
     "@babel/preset-env": "^7.5.5",
     "@babel/preset-react": "^7.0.0",
-    "@types/mapbox-gl": "^0.54.2",
```
Simon-Laux: This is needed for auto-completion in my IDE, so I would keep it / upgrade it to 54.3 when we upgraded mapbox-gl to 1.3.0

Jikstra: Let's put it in `optionalDependencies`?

PR #964, opened 2019-09-01
Add missing labels to experimental.json

nicodh:
We should cleanup this file. Not sure if all label are needed.
At least the one I added should go into the _locales files...


Simon-Laux:
Do you mean translation keys by labels?

Issue #870, opened 2019-06-24
Add labeled items for accessibility


OzancanKaratas:
No Windows builds. So I cannot give any information.

NVDA is a free and open source screen reader for Windows. We are using it to use the computers. Visually impaired users can access to the Javascript applications such as Skype.

NVDA APIs is used for manage and read messages from external applications. If Delta Chat's continuous integrations does a build for Windows, please interest this issue. I'm ready to help and send feedback for blinds.

[NVDA repository](https://github.com/nvaccess/nvda)

Note: NVDA APIs written with C++. We should work for Javascripts codes.

Simon-Laux:
Great Idea with including screen reading...
But I see a few points against integrating NVDA api in deltachat desktop electron:
1. We already had/have difficulties with deltachat-node building, adding another native dependency won't help.
2. we should add #488 first in my opinion (Using a mouse when blind sounds no fun, but on the other hand I don't know how you navigate a computer when blind at all)
3. Libraries for deltachat-desktop (electron edition) should be cross platform (linux,macOS,windows).

I would love a good feature proposal on the forum (https://support.delta.chat/) as accessibility is a topic that can be interesting for other platforms too.

PS: We could maybe start a new speech-based deltachat client with the core that has a comandline interface for input and speaks with the user. Let's talk about such ideas on the forum https://support.delta.chat/

OzancanKaratas:
I saw the button labeling process is better than adding NVDA API. Unlabeled buttons bigger problem. I think we should label the DeltaChat buttons. My screen reader does not read the several buttons, just announces button. I must select the button to see what happens.

Simon-Laux:
I guess its the buttons that have an icon instead of text, I'll try to label them with the area-label attribute.

Simon-Laux:
I think this is done now, please reopen or create a new issue when new labeling or accessibility issues arise.
The accessibility strings will get into transifex when we do the next "putting all new language strings into transifex" task.

PR #966, opened 2019-09-02
Label Icon Buttons for accessibility

Simon-Laux:
#870
@OzancanKaratas please test if that works the way you want.
Also the text is in `_locales/_experimental_en.json` if you want to change it to sound better.

### DIFF
--- a/_locales/_experimental_en.json
+++ b/_locales/_experimental_en.json
@@ -27,13 +27,13 @@
     "message": "Sending message failed"
   },
   "a11y_send_btn_label": {
-    "message": "Send Button"
+    "message": "Send"
   },
   "a11y_emoji_btn_label": {
-    "message": "Emoji Button"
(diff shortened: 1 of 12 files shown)


Simon-Laux:
I would wait a bit with moving "a11y_goto_media_label" and "a11y_goto_chat_label" because they might become a dialog soon @jikstra

OzancanKaratas:
@Simon-Laux I will test it but you should release a Windows version.

OzancanKaratas:
Screen readers does not love prefixes like button, link. Because it already speaks the prefixes. Please remove it. It should be:

Send button = Send

Simon-Laux:
Ok, what about the delivery status labels? are they ok or should they just say sent/delivered/read and so on?

OzancanKaratas:
I don't see a problem. However, I should try a compiled version.

PR #968, opened 2019-09-07
Rework create chat avoid react errors

nicodh:
@Simon-Laux I used the same hardcoded fallback color #505050 for pseudo group/contact that was used in AvatarBubble. Not sure if we need to adapt that for theming...

### DIFF
--- a/src/renderer/components/dialogs/CreateChat-Styles.js
+++ b/src/renderer/components/dialogs/CreateChat-Styles.js
@@ -105,11 +105,11 @@ export const GroupImageWrapper = styled.div`
 `
 
 export const GroupImage = (props) => {
-  const { groupImage, onSetGroupImage, onUnsetGroupImage } = props
+  const { groupImage, onSetGroupImage, onUnsetGroupImage, ...otherProps } = props
   return (
     <GroupImageWrapper>
       { groupImage && <AvatarImage src={groupImage} onClick={onSetGroupImage} {...props} /> }
-      { !groupImage && <AvatarBubble onClick={onSetGroupImage} {...props}>G</AvatarBubble> }
+      { !groupImage && <AvatarBubble onClick={onSetGroupImage} {...otherProps}>G</AvatarBubble> }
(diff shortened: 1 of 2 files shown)


Simon-Laux:
Ah right that would be a usecase of a default color... anyways avatar color is nothing that will be changable with a theme, because the text is always white and the bg color comes from the core so I think we'll be fine for this one.

src/renderer/components/dialogs/CreateChat-Styles.js
```diff
@@ -105,11 +105,11 @@ export const GroupImageWrapper = styled.div`
 `
 
 export const GroupImage = (props) => {
-  const { groupImage, onSetGroupImage, onUnsetGroupImage } = props
+  const { groupImage, onSetGroupImage, onUnsetGroupImage, ...otherProps } = props
   return (
     <GroupImageWrapper>
       { groupImage && <AvatarImage src={groupImage} onClick={onSetGroupImage} {...props} /> }
-      { !groupImage && <AvatarBubble onClick={onSetGroupImage} {...props}>G</AvatarBubble> }
+      { !groupImage && <AvatarBubble onClick={onSetGroupImage} {...otherProps}>G</AvatarBubble> }
```
Simon-Laux: not sure if we should use such a wildcard

nicodh: Well before it passed all properties - now all except onSetGroupImage, onUnSetGroupImage, GroupImage where onSetGroupImage, onUnSetGroupImage are not known properties for AvatarBubble...

Simon-Laux: ok

src/renderer/components/dialogs/CreateChat-Styles.js
```diff
@@ -105,11 +105,11 @@ export const GroupImageWrapper = styled.div`
 `
 
 export const GroupImage = (props) => {
-  const { groupImage, onSetGroupImage, onUnsetGroupImage } = props
+  const { groupImage, onSetGroupImage, onUnsetGroupImage, ...otherProps } = props
```
Jikstra: From where are you getting otherProps?

PR #969, opened 2019-09-08
Renaming proposals

nicodh:
This PR is mainly about renaming but also a bit refactoring (last commit)
I consider 'viewMode' more intuitive than 'show'

I also would like to get rid of the name "ScreenController" and the "screen" variables. To me "screen" is more a hardware term (although I'm not native speaker...)
"ScreenController" => "FrontendController" ?
"changeScreen" => "changeView"
etc.

But let's start with this PR first...

### DIFF
--- a/src/renderer/stores/locations.js
+++ b/src/renderer/stores/locations.js
@@ -16,10 +16,7 @@ const getLocations = (chatId, mapSettings) => {
   const { timestampFrom, timestampTo } = mapSettings
   callDcMethod(
     'getLocations',
-    chatId,
-    0,
-    timestampFrom,
-    timestampTo
+    [chatId, 0, timestampFrom, timestampTo]
   )
 }
 
--- a/src/renderer/stores/locations.js
+++ b/src/renderer/stores/locations.js
@@ -16,7 +16,8 @@ const getLocations = (chatId, mapSettings) => {
   const { timestampFrom, timestampTo } = mapSettings
   callDcMethod(
     'getLocations',
-    [chatId, 0, timestampFrom, timestampTo]
+    [chatId, 0, timestampFrom, timestampTo],
+    (locations) => locationStore.setState({ ...locationStore.getState(), locations })
   )
 }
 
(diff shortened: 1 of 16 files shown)


src/renderer/stores/locations.js
```diff
@@ -13,13 +14,9 @@ const locationStore = new Store(defaultState)
 
 const getLocations = (chatId, mapSettings) => {
   const { timestampFrom, timestampTo } = mapSettings
-  ipcRenderer.send(
-    'EVENT_DC_FUNCTION_CALL',
+  callDcMethod(
```
Jikstra: Can you refactor those methods to also return the locations? Otherwise we still need to listen for another event to get back the results. (If that's possible without too much refactoring)

Jikstra:
I don't really like viewMode, it seems to generic to me. How about just naming it view? Or innerView? Puuh naming is hard.

Simon-Laux:
please rename the whole conversations thing to chatlist/view

Jikstra:
> conversations thing
?

Simon-Laux:
.module-conversation-list-item for example

Simon-Laux:
just do a grep conversation to find where it hides.
I find it confusing to have 2 namings.
Either conversation or chat.
I vote for chat.

nicodh:
I agree but that should be done if the rework create chat PR is merged...

Jikstra:
> I agree but that should be done if the rework create chat PR is merged...

Yup. Plus I want to throw out all the conversations stuff anyway and replace it with our own implemenations. The Create Chat does this already for Contacts and Chats, only messages are missing (which is currently around 600 lines of sass code, no idea what the hell they are doing to style some message/speech bubbles). So i'm fine with keeping it as it is, will remove conversations in another pr somewhere in the future.

nicodh:
> I don't really like viewMode, it seems to generic to me. Puuh naming is hard. Something like `dialogSection`, `dialogView`, `dialogViewChoice`? It's specific to dialogs because only there we want to have different "innerViews". 

I would say generic is fine. We have a dialog with different viewModes. No need to include "dialog" in the name imho...
To me it seems very readable and intuitive. 
For example
` viewMode.startsWith('createGroup') or setViewMode('createGroup-main')`

Jikstra:
Yup you're right, should be fine :)

Issue #982, opened 2019-09-11
What does "Enter key sents" switch does?


KAction:
Here are two screenshots of settings with "Enter key sends" toggle. Whether it is on or off, text is same, so I am confused. Probably, it is about Enter vs Ctrl-Enter, like in Telegram. Please, make text describe what is currently configured state.

git:  [c568ff]

![foo1](https://user-images.githubusercontent.com/44864956/64742284-37cb8a80-d4eb-11e9-8ab6-5f58c1b80ac1.png)
![foo2](https://user-images.githubusercontent.com/44864956/64742288-3b5f1180-d4eb-11e9-90df-73eb79614673.png)

nicodh:
I would say that's common behavior. The text remains the same but the button can be on or off. That's how all the settings buttons in Android or Iphone look like...

pabzm:
All our boolean settings have a fixed text-label that doesn't change when toggled. We can't change that just for one setting.
If we'd have many issues in which people report problems with the toggle buttons we could discuss how to improve the UI, but currently that apparently isn't the case.

PR #983, opened 2019-09-12
[wip] Change setting for enter key sends or ctrl+enter sends to make it better

Jikstra:
understandable. Refactor handleXSettingsChange methods

Screenshot_20190912_130522

- [ ] fix warning of uncontrolled state
- [ ] for whatever reason both radiobutton groups don't rerender on setState

Fixes https://github.com/deltachat/deltachat-desktop/issues/982

BLOCKED on me fixing https://github.com/deltachat/deltachat-core-rust/pull/503

### DIFF
--- a/_locales/_experimental_en.json
+++ b/_locales/_experimental_en.json
@@ -82,5 +82,8 @@
   },
   "a11y_message_context_menu_btn_label": {
     "message": "message actions"
+  },
+  "pref_ctrl_enter_sends": {
+    "message": "Ctrl+enter sends"
   }
 }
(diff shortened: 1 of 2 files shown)


Simon-Laux:
@OzancanKaratas Even if its only a small change we normally do pull requests. For editing the language files I have mention that only `_languages.json` and `_experimental_en.json` can be edited by hand. The other files in this directory are pulled from transifex and so changes to those files will be overwritten. https://www.transifex.com/delta-chat/delta-chat-app/dashboard/

PR #990, opened 2019-09-14
Fix forward message dialog & refactor SearchableList

nicodh:


nicodh:
Forwarding messages works now but the search input in nav header should look a bit different than in the dialog. I already changed the background color to transparent, but I would prefer not to have all these styled divs spreaded over all components.  It makes the components harder to reuse and mixes styles, code and markup. 
We should organize all styles in sass imho...

PR #994, opened 2019-09-16
Try if sourcing cargo env file is enough for travis to find cargo again

Jikstra:


nicodh:
Cool- now it works even with dependency to git repository. lgtm - or do you want to remove the logging again?

Issue #396, opened 2018-12-07
Search should only search non archived


ralphtheninja:
Currently the search functionality also searches archived chats, so it's possible to search and then select an archived chat, but when you delete the search query (backspace) the chat goes away. I think we should limit the search to the active chat list, e.g. search only non archived or search archived.

r10s:
i think the global search should search unarchived chats as well as archived chats (in fact, in the mid-term, it would be great if also contacts and messages are searched).

i would also not make a difference if the global search is started from the archived list or from the normal list.

the archived state is a rather weak state, eg. chats are automatically unarchived when new messages come in. so i think it would be confusing if chats/messages are found or not depending on this state. 

however, i understand the ui problem.

but maybe we can just ignore it and mitigate the issue by clearly showing the that chat _is an archived chat_ - in android, we show the text [Archived] below the date for archived chats (instead of the delivery indicators).

this way, the user at least knows where he can find the chat.

karissa:
makes sense. I am going to label this as a feature not a bug then.

ralphtheninja:
I don't mind that the search is also quering archived chats, I'm more against the ui quirk, where when you delete your search (after have selected an archived chat) makes the chat disappear. I'm thinking we just need to tweak the ux a bit.

r10s:
maybe it is totally sufficient to not change the selected search when the chat list changes.
i would recommend to make the chatlist from the selected chat anyway. i think there may be situations where a strict requirement that the selected chat has to be in the chatlist may cause problems esp. on multi-device.

together with an "archived" label in the chatlist everything should be fine imo, this is how the label looks like on android (btw. the screenshot was transferred using deltachat-android and deltachat-desktop :).
![image-5](https://user-images.githubusercontent.com/9800740/49679135-6b274780-fa89-11e8-8deb-4530b38e591f.png)

Simon-Laux:
I think we need a better way of displaying search results, maybe using the chatview and having the search results temporarily taking up a slot in the chatsList.

Also advanced searches with filters like `from:[name]` would be awesome, but that's an idea for the future. (Discord has a good UI for those searches - it makes visible what filters belong together and gives you a drop-down menu with suggestion)

r10s:
hm, i think showing the search results on the left instead of the chatlist is okay, at least for now.
i would think of it as a filter of the chatlist.
however, of course this can be improved, but i think for now it is fine :)

karissa:
[...]

PR #997, opened 2019-09-19
Rework Forward Message Dialog, Chatlist and implement async lazy loading chatlist...

### LINKED ISSUE #917: chatList.js:57 Uncaught TypeError: Cannot read property 'state' of null     at EventEmitter.module.exports../src/renderer/stores/chatList.js.ipcRenderer.on (chatList.js:57)     at EventEmitter.emit (events.js:187)
Just imported an older backup and this error showed up multiple times.

### LINKED ISSUE #951: Displayed wrong message in Import secret keys implementation
Displayed %1$s instead of folder or file name.

Jikstra:
- Implement add contact over chat search (Fixes #917)
- Implement lazy loading chatlist which only fetches what's in view
- Removes chatlist store (Fixes https://github.com/deltachat/deltachat-desktop/issues/396)
- Improves chatlist styling
- Use lazy loading chatlist in forward message dialog and tweak styling of that dialog
- Start moving from styled components to sass (thanks @Simon-Laux & @nicodh for pushing this, was a long trip for me to change my mind :D)
- Make settings dialog the same width as CreateChat/ForwardMessage dialog
- Refactored everything that was in my way 
- Restyle & refactor MessageDetail dialog
- Fix #951, changes import secret keys file import title


Simon-Laux:
Forward message - chat search was Case Sensitive(different results when searching lower or uppercase) and displayed new message count badges.
Does this pr already fix those two points?

Jikstra:
> Forward message - chat search was Case Sensitive

yes

> displayed new message count badges.

I'm still displaying this, but can include fixing this into this pr

src/main/deltachat/index.js
```diff
@@ -104,84 +65,54 @@ class DeltaChatController extends EventEmitter {
...
     dc.on('DC_EVENT_MSGS_CHANGED', (chatId, msgId) => {
-      if (chatId === 0) {
-        this.updateChatList()
-        return
-      }
-      // Don't update if a draft changes
-      if (msgId === 0) return
+      this.onChatListChanged()
+      this.onChatListItemChanged(chatId)
       this.onMessageUpdate(chatId, msgId, 'DC_EVENT_MSGS_CHANGED')
```
Jikstra: @nicodh the reason why we call onChatListChanged and onChatListItemChanged is because onChatListChanged triggers a refetch of the chatlist, which would result in a resorting of the displayed chatlists. The useLazyChatListItems hook only tries to fetch chats in view after chatListIds changed, but doesn't force refetch them (this is needed for search, where only the sort/which ids are shown changes but we don't want to refetch all chats). And this is where onChatListItemChanged comes into play, because if called with chatId=0, it force refetches all chats in view. It probably makes sense to split this into two events, onChatListItemChanged onChatListForceUpdateInView or something like this.

Simon-Laux:
_Compliance department says_ YOLO

PR #1001, opened 2019-09-23
Select background

Simon-Laux:
- [x] Logic
- [x] Styling
- [x] Translation Strings

### DIFF
--- a/scss/_global.scss
+++ b/scss/_global.scss
@@ -137,20 +137,3 @@ code {
   image-rendering: pixelated;
 }
 
-
-// ChatView Background selector
-
-.background-preview {
-  width: 100pt;
-  height: 100pt;
-  border-radius: 5pt;
-  background-image: url(../images/background_light.svg);
-  background-color: var(--chatViewBg);
-}
-
-.background-options {
-  display: flex;
-  flex-direction: column;
-  padding-left: 8pt;
-  justify-content: space-evenly;
-}
--- a/scss/_global.scss
+++ b/scss/_global.scss
@@ -145,7 +145,6 @@ code {
   height: 100pt;
   border-radius: 5pt;
   background-image: url(../images/background_light.svg);
-  background-size: cover;
   background-color: var(--chatViewBg);
 }
 
--- a/scss/_global.scss
+++ b/scss/_global.scss
@@ -136,3 +136,22 @@ code {
   image-rendering: -o-crisp-edges;
   image-rendering: pixelated;
 }
+
+
+// ChatView Background selector
+
+.background-preview {
+  width: 100pt;
+  height: 100pt;
+  border-radius: 5pt;
+  background-image: url(../images/background_light.svg);
+  background-size: cover;
+  background-color: var(--chatViewBg);
+}
+
+.background-options {
+  display: flex;
+  flex-direction: column;
+  padding-left: 8pt;
+  justify-content: space-evenly;
+}
--- a/src/main/ipc.js
+++ b/src/main/ipc.js
@@ -1,6 +1,6 @@
 module.exports = { init }
 
-const { app, ipcMain } = require('electron')
+const { app, ipcMain, dialog } = require('electron')
 const rimraf = require('rimraf')
 const path = require('path')
 const fs = require('fs')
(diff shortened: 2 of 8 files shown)


Simon-Laux:
Ah we still need to decide on the default background on all platforms for next release @r10s @Jikstra
If we put in the new one("potato-stamps") on android too or if we keep the old one and decide at the next release.
It would bring some visual change for users to see.

r10s:
tbh, i would prefer a decision on that not now. as said, however, the potato-stamp is not my favorite, but when everyone likes it, we can go for it. but it could be technically better.

but maybe we can also use different backgrounds in the upcoming release and sync them later. wfiw, ios currently has _no_ background at all.

src/main/ipc.js
```diff
@@ -173,11 +173,34 @@ function init (cwd, state, logHandler) {
...
   })
 
-  ipcMain.on('updateDesktopSetting', (e, key, value) => {
+  const updateDesktopSetting = (e, key, value) => {
     const { saved } = app.state
     saved[key] = value
     app.saveState({ saved })
     render()
+  }
+  ipcMain.on('updateDesktopSetting', updateDesktopSetting)
```
ralphtheninja: Maybe do

```js
 ipcMain.on('updateDesktopSetting', (e, k, v) => updateDesktopSetting(k, v))
```

src/main/ipc.js
```diff
@@ -173,11 +173,34 @@ function init (cwd, state, logHandler) {
     e.returnValue = app.localeData
   })
 
-  ipcMain.on('updateDesktopSetting', (e, key, value) => {
+  const updateDesktopSetting = (e, key, value) => {
```
ralphtheninja: Maybe remove `e` since we don't use it?

src/main/ipc.js
```diff
@@ -173,11 +173,34 @@ function init (cwd, state, logHandler) {
...
+    }, (filenames) => {
+      if (!filenames) { return }
+      log.info('BG-IMG Selected File:', filenames[0])
+      const newPath = path.join(getConfigPath(), 'background' + path.extname(filenames[0]))
+      fs.copyFile(filenames[0], newPath, (err) => {
+        if (err) {
+          log.error('BG-IMG Copy Failed', err)
+          return
+        }
+        updateDesktopSetting(null, 'chatViewBgImg', `url("${newPath}")`)
```
ralphtheninja: .. and if so, remove `null` here :)

OzancanKaratas:
I saw some typos. However, I cannot create pull request because commits are conflicting. Please manually merge [this commit](https://github.com/OzancanKaratas/deltachat-desktop/commit/846b561d1aae9b791ef0d0216d0e7c0e2e92d206.patch).

ralphtheninja:
> I saw some typos. However, I cannot create pull request because commits are conflicting. Please manually merge [this commit](https://github.com/OzancanKaratas/deltachat-desktop/commit/846b561d1aae9b791ef0d0216d0e7c0e2e92d206.patch).

Why are commits conflicting? It should work fine if you rebase your work on latest master or on this branch?

Simon-Laux:
@OzancanKaratas please make a new pr for this, It's not changing many typos that were introduced in this pr from what I can see with my sleepy eyes. Also the translations get also fixed on transifex sooner or later so I don't even know if it's worth the effort to do it right now.

src/main/ipc.js
Simon-Laux: a new function... maybe more clearer, but us the definition is above it should be fine

scss/_global.scss
```diff
@@ -136,3 +136,21 @@ code {
   image-rendering: -o-crisp-edges;
   image-rendering: pixelated;
 }
+
+
+// ChatView Background selector
+
+.background-preview {
```
Jikstra: Please move into scss/modules/_settings.scss

src/renderer/components/ChatView.js
```diff
@@ -242,36 +243,44 @@ class ChatView extends React.Component {
...
-        <Composer
-          ref={this.refComposer}
-          chatId={chat.id}
-          draft={chat.draft}
-          onSubmit={this.writeMessage}
-          setComposerSize={this.setComposerSize.bind(this)}
-        />
-      </ChatViewWrapper>
+      <SettingsContext.Consumer>
+        {(settings) => (
```
Jikstra: Is it possible to use hooks here?

PR #1008, opened 2019-09-26
Prepare for v0.200.0

Jikstra:
I currently don't have access to npmjs.org and therefore can't publish a new npm package. @ralphtheninja can you pull latest master on deltachat-node and run `npm run download-prebuilds` and `npm publish`? Then we can update deltachat-node to latest npm version and release desktop v0.200.


ralphtheninja:
Yep. On it. Managed to mess up alpha.7 so publishing alpha.8 instead. Updating CHANGELOG in this PR after that.

ralphtheninja:
Timing issues on the integration tests?

PR #1017, opened 2019-09-28
Chatview refactoring

nicodh:


src/renderer/components/ChatListContextMenu.js
```diff
@@ -76,7 +76,7 @@ const ChatListContextMenu = React.memo((props) => {
           <Icon icon='lock' /> {tx('encryption_info_desktop')}
         </MenuItem>
       }
-      {chat.isGroup
+      {chat.isGroup && chat.contacts.find(contact => contact.id === 1)
```
Jikstra: Why?

nicodh: Its more explicit with the next commit

src/main/deltachat/chatlist.js
```diff
@@ -96,6 +96,10 @@ function getSmallChatById (chatId, list, i) {
   const summary = list.getSummary(i).toJson()
   const lastUpdated = summary.timestamp ? summary.timestamp * 1000 : null
 
+  if (summary.text2 === '[The message was sent with non-verified encryption.. See "Info" for details.]') {
```
Simon-Laux: Isn't there a better way to do this? some msg error enum or the like?

Jikstra: The core doesn't give us anything better

Simon-Laux: did we already made an feature request there?

Jikstra: Nope but no other client currently translates this string into another string. This is a error and normally shouldn't happen.

Simon-Laux:
DM chats are broken

src/main/deltachat/chatlist.js
Jikstra: @nicodh i think translating this error message doesn't really make sense as it shouldn't happen in the first case. @hpk42 @r10s any opinions on this?

src/main/deltachat/chatlist.js
```diff
@@ -115,7 +119,8 @@ function getSmallChatById (chatId, list, i) {
     isVerified: chat.isVerified,
     isGroup: chat.isGroup,
     freshMessageCounter: chat.freshMessageCounter,
-    isArchiveLink: chat.isArchiveLink
+    isArchiveLink: chat.isArchiveLink,
+    selfInGroup: chat.selfInGroup
```
Jikstra: I propose isSelfInGroup, just to be consistent with the two variables above. Naming is hard and i was just reminded about our talk about naming some days ago ^^

src/renderer/components/ChatView.js
```diff
@@ -106,118 +106,123 @@ const ConversationWrapper = styled.div`
...
     callDcMethod(
       'sendMessage',
       [chat.id, opts.text, opts.filename]
     )
   }
 
-  fetchNextMessages () {
-    const { chat } = this.props
+  const fetchNextMessages = () => {
+    const { chat } = props
```
Jikstra: define this `const` at the top of `ChatView()`

src/main/deltachat/chatlist.js
nicodh: Yes sure - just a temporary workaround

nicodh:
> DM chats are broken; please fix

@s

src/main/deltachat/chatlist.js
nicodh: hmm - I don't agree. isArchiveLink means this chat IS an archive link. selfInGroup means "you are in the group of this chat" not the chat IS selfInGroup...
naming issues... ;-)

src/main/deltachat/chatlist.js
nicodh: And we have another not translated error message: [foo@bar is not a member of this verified group. See "Info" for details.]
So maybe better wait still the core translates these messages? Or gives a flag or something so we can adapt the UI

nicodh:
DM chats should work again...

src/main/deltachat/chatlist.js
Jikstra: I don't think we need to translate those errors as they shouldn't happen. It's more to trace bugs down as far as i understand this.

src/main/deltachat/chatlist.js
Jikstra: Fair :)

Jikstra:
Not sure what went wrong on resolving the merge conflict, can look into it later. But looks fine to me so far, would like to merge it and test it out further. I'm currently working on the Composer which is quite nearby and would like to have the outer components as hooks (they are passing around some refs which is a bit annoying). So would be happy to have this on master asap.

PR #1020, opened 2019-09-29
Some UX improvements

Jikstra:
- Rename the `Create group` button to a more generic `ok` button
- Don't delete the group which got created during showing qr code on
close button again
- Make all dialogs closable with escape key but don't close them when
clicking outside

@r10s

### DIFF
--- a/src/renderer/components/Composer.js
+++ b/src/renderer/components/Composer.js
@@ -255,7 +255,7 @@ class Composer extends React.Component {
           </EmojiPickerWrapper>
         }
         <SendButtonCircleWrapper onClick={this.sendMessage}>
-          <SendButton aria-label={tx('a11y_send_btn_label')}/>
+          <SendButton aria-label={tx('a11y_send_btn_label')} />
         </SendButtonCircleWrapper>
       </ComposerWrapper>
     )
--- a/src/renderer/components/Composer.js
+++ b/src/renderer/components/Composer.js
@@ -254,8 +254,8 @@ class Composer extends React.Component {
             />
           </EmojiPickerWrapper>
         }
-        <SendButtonCircleWrapper onClick={this.sendMessage} aria-label={tx('a11y_send_btn_label')}>
-          <SendButton />
+        <SendButtonCircleWrapper onClick={this.sendMessage}>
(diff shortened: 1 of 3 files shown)


OzancanKaratas:
I noticed the send button appears blank and “Send” text appears separated from the button. Can you merge the button with “Send” text?

Jikstra:
@OzancanKaratas is it about the aria label on the send button?
The html markup looks like this:
```
<div aria-label="Send" class="sc-frDJqD cTeyJv">
  <button class="sc-kvZOFW XIDDm"></button>
</div>
```

Probably the aria-label should be at the button element and not the wrapper element. I can do a pr for you on which you could give it a try.

Jikstra:
@OzancanKaratas I just pushed it to this branch to make things easier. How do the aria labels work for you? During the release cycle i got a bit confused with them as they broke tests and removed them accidentially...

OzancanKaratas:
Everything looks good for now. Thank you for your interest. By the way, I had separated “Delta” and “Chat” words such as in iOS and Android clients. Why did you deleted my changes?

Jikstra:
> Everything looks good for now. Thank you for your interest.

Sure, thanks for reporting those issues and being interested in delta chat :)

> By the way, I had separated “Delta” and “Chat” words such as in iOS and Android clients. 

Very helpful!

> Why did you deleted my changes?

I had to delete the strings that introduced to the _experimental_en.json which haven't been in that file before because they overwrite that string with your english version and ignoring what language is selected. So if someone uses deltachat in german language, your change would overwrite the german string with your english one. For translations we use transifex and a central "db" which we share across all platforms (ios/android/desktop). And every now and then we move all experimental strings to the central db where they get properly translated into all languages (otherwise there's always only the english experimental translation shown).
I hope this clarifies things a bit.

OzancanKaratas:
I think I didn't understand. Does experimental strings not overwritten existing ones if newer? So Transifex marks them as untranslated if we changed.

Jikstra:
It does overwrite, but you can only do one language with it, experimental doesnt support the different languages, only english. It's a bit of a hack for us developing on desktop so that we don't need to upload every language string first to transifex before we can use it.

PR #1026, opened 2019-09-30
Improve logger a bit

Simon-Laux:
- omit stacktrace for debug and info
- move core event logs to their own log channel (makes event output smaller and less annoying)
- change logile extention to .log.tsv
- add colors, time elapsed and generally rework main process logging a bit:

### DIFF
--- a/src/logger.js
+++ b/src/logger.js
@@ -17,7 +17,7 @@ function setLogHandler (LogHandler) {
   handler = LogHandler
 }
 
-function log (channel, level, ...args) {
+function log (channel, level, stacktrace, ...args) {
   const variant = LoggerVariants[level]
   if (!handler) {
     /* ignore-console-log */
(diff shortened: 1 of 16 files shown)


src/logger.js
```diff
@@ -26,9 +28,28 @@ function log (channel, level, ...args) {
...
+    if (isMainProcess) {
+      const begining = `${Math.round((Date.now() - startTime) / 100) / 10}s ${[
+        '[D]',
+        colors.blue('[i]'),
+        colors.yellow('[w]'),
+        colors.red('[E]'),
+        colors.red('[C]')
+      ][level]}${colors.grey(channel)}:`
+      if (!stacktrace) {
+        console.log(begining, ...args)
```
Jikstra: `/* ignore-console-log */` is missing

src/logger.js
```diff
@@ -26,9 +28,28 @@ function log (channel, level, ...args) {
...
+      } else {
+        console.log(begining, ...args, colors.red(stacktrace))
```
Jikstra: same

Simon-Laux:
> Very helpful. You said that core events are now in their own channel, how can i still see them?

With channel I mean channel of the logger. At the moment there is no filtering going on at all so all channels are displayed. We might develop a log reader that can filter the channels of the logfiles to make reading them easier.
At the moment the change is:
before:
```
main/deltachat DEBUG [] Core Event DC_EVENT_INFO  IMAP-IDLE SPAWNING
```
after:
```
core/event DEBUG [] DC_EVENT_INFO  IMAP-IDLE SPAWNING
```
after all changes:
```
3.2s [D]core/event: DC_EVENT_INFO  IMAP-IDLE SPAWNING
```

PR #1047, opened 2019-10-12
Bring back userinfo

### LINKED ISSUE #1046: .user-feedback class went missing
`.user-feedback` class went missing

Error messages are now displayed under the navbar, where the user can't see them.

(please restore or/and improve design/display of these (error)messages)

Simon-Laux:
fixes #1046
and make it look different, its still nowhere near pretty but now its a bit better imho


r10s:
to make think look nicer without much effort: i suggest to use the #ffffba instead of #red. this also makes the whole thing better readable and is "alarming enough" i'd say.

PR #1054, opened 2019-10-19
mark seen quick fix

### LINKED ISSUE #1043: Unread bagde doesn't update
When you are in an chat it just counts up and is only updated after you switch to another chat and re-enter the previous chat again.

@nicodh and/or @Jikstra might know what is causing it.

Simon-Laux:
closes #1043 for now.

ping @hpk42


hpk42:
So this marks incoming messages as seen if the chat is selected, right?

Simon-Laux:
exactly, It also takes into account whether the window is selected

Issue #681, opened 2019-02-15
Language choice duplicates?


ghoffart:
Are these duplicates?
- Dutch, Nederlands
- Russian (Russia), Russian (cyrillic version)
- If Russian is displayed with cyrillic letters, then Japanese, Korean, Chinese, Ukrainian should be displayed with native glyphs, too? And which 'Chinese'?

- Platform (linux/mac/windows): macOS 10.14.x
- Delta Chat Desktop Version: 0.101.0
- Expected behavior: de-duped/unique languages
- Actual behavior: Dutch & Nederlands, Russian (Russia) & Russian entry duplicates?
- Steps to reproduce the problem: open the language dropdown
- Screenshots:
- Logs:

link2xt:
ru_RU and ru are duplicates indeed. The "Russian (Russia)" version `ru_RU.xml` is incomplete and can be removed.

I am not familiar with your Transifex translation update process, but this entry and the corresponding file can be safely removed:
https://github.com/deltachat/deltachat-desktop/blob/cd2217d385d7f902deb90a9009a9aa45efa47024/_locales/_languages.json#L83

r10s:
just tried on mac, lgtm now.

link2xt:
@r10s Dutch and Nederlands are still both present, is it ok?

r10s:
@link2xt thanks, filed an extra issue for that https://github.com/deltachat/deltachat-desktop/issues/1308

PR #1055, opened 2019-10-20
Remove ru_RU locale

link2xt:
ru locale is what is translated on Transifex now.

Address #681


OzancanKaratas:
You did the right thing. It can be merged and close #681.

PR #1056, opened 2019-10-21
Improve delta chat controller structure

Simon-Laux:
this brings back intellisense and typeinformations.
progress:
- [X] './autocrypt'
- [X] './backup'
- [X] './chatlist'
- [x] './chatmethods'
- [x] './locations'
- [x] './login'
- [X] './messagelist'
- [X] './settings'
- [X] './stickers'

my steps are: 
1. transform to class
2. check whether to expose them
3. change call signature where they are used
4. make sure that they don't access stuff on this that is under this._controller


src/main/deltachat/index.js
```diff
@@ -62,16 +66,45 @@ class DeltaChatController extends EventEmitter {
     logCoreEv.debug(event, data1, data2)
   }
 
+  __resolveNestedMethod (self, methodName) {
+    let previous = self
+    let scope
+    for (const key of methodName.split('.')) {
```
Jikstra: How about making this simpler and only allowing a 2 level nesting? We should never have more than namespace.method

src/main/deltachat/index.js
```diff
@@ -42,46 +42,59 @@ class DeltaChatController extends EventEmitter {
...
+      autocrypt: new DCAutocrypt(this),
+      backup: new DCBackup(this),
+      chatList: new DCChatList(this),
+      contacts: new DCContacts(this),
+      chat: new DCChat(this),
+      locations: new DCLocations(this),
+      loginController: new DCLoginController(this),
+      messageList: new DCMessageList(this),
+      settings: new DCSettings(this),
+      stickers: new DCStickers(this)
```
Jikstra: Ah here we go.

src/main/deltachat/index.js
```diff
@@ -22,10 +27,17 @@ class DeltaChatController extends EventEmitter {
...
     if (!saved) throw new Error('Saved settings are a required argument to DeltaChatController')
     this.loadSplitOuts()
+    /**
+     * @type {DeltaChat}
+     */
+    this._dc = undefined
+  }
+
+  get autocrypt () {
+    return new DCAutocrypt(this._dc)
```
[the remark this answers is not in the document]
Simon-Laux: shouldn't have been a big issue anyway (just instancing a class and assigning 2 vars

src/main/deltachat/index.js
Simon-Laux: I also thought about this, I can do that.

Simon-Laux:
```
__resolveNestedMethod (self, methodName) {
    const parts = methodName.split('.')
    if (parts.length > 2) {
      const message = 'Resolving of nested method name failed: Too many parts, only two allowed: ' + methodName
      log.error(message)
      throw new Error(message)
    }
    const scope = self[parts[0]]
    if (typeof scope === 'undefined') {
      const message = 'Resolving of nested method name failed: ' + methodName
      log.error(message)
      throw new Error(message)
    }
    const method = scope[parts[1]]
    if (typeof method !== 'function') {
      const message = '(nested) Method is not of type function: ' + methodName
      log.error(message)
      throw new Error(message)
    }
    return method.bind(scope)
  }
```
Doesn't look so much simpler, does it?

PR #1060, opened 2019-10-22
fix #1045

### LINKED ISSUE #1045: stop using dc_chat_get_subtitle()
the core-function `dc_chat_get_subtitle()` will be deleted.

the subtitle, however,  shall stay and can be created eg. as follows:

```
int chat_contacts = dc_context_get_chat_contacts(context, chat_id); 
int type = dc_chat_get_type(chat);
if( type==DC_CHAT_TYPE_GROUP || type==DC_CHAT_TYPE_VERIFIED_GROUP ) {
  subtitle = get_fine_plural_form("n_members", chat_contacts.length);
}
else if( members.length>=1) {
  contact = dc_get_contact(context, chat_contacts[0]);
  subtitle = dc_contact_get_addr(contact);
}
else {
  subtitle = "ErrTitle";
}
```

the example does not use unrefs - but it won't run anyway ;)

the reason for the function to go away is 

- due to the "members" plurals form this would require a lot of things if we want to do that properly.
- the alternative are hacks - that are [more](https://github.com/deltachat/deltachat-android/pull/1053/commits/bdccf84bf572895042eb0e782f3a4545657afde5) [complex](https://github.com/deltachat/deltachat-ios/commit/926d0a91eb96003ecbb5b3b4be83cc37265e7a15) than the few lines above.
- desktop can now show "1 member" correctly - and not "1 members" :)
- give more power to the ui

Simon-Laux:
please review and approve #1056 before this.


Simon-Laux:
this also makes the translation function a tiny tiny bit more stable

src/main/deltachat/chatlist.js
```diff
@@ -164,6 +163,28 @@ module.exports = class DCChatList extends SplitOut {
...
+          return tx('chat_self_talk_subtitle')
+        }
+        return contacts[0].address
+      }
+    } else {
+      switch (chat.id) {
+        case C.DC_CHAT_ID_DEADDROP:
+          return tx('menu_deaddrop_subtitle')
+        // case C.DC_CHAT_ID_STARRED:
+        //   return 'stared'
```
Jikstra: `starred`

PR #1062, opened 2019-10-22
Show line breaks in message body

Jikstra:


Simon-Laux:
why is this better than regular `<br>`?

Jikstra:
> why is this better than regular <br>?

Regular br was for whatever reason not shown.

PR #1071, opened 2019-10-24
Fix broken system notifications

### LINKED ISSUE #1068: Notifications not shown any more
System notifications are not shown any more, since the registered event listener for Incoming message is not triggered

nicodh:
resolves #1068

- adds 2 new controls in settings component (notifications & showContentInNotification)

what I don't like is that we need stuff like dc._dc.on(evt, handler) etc. but that should be refactored in another PR


Simon-Laux:
> what I don't like is that we need stuff like dc._dc.on(evt, handler) etc. but that should be refactored in another PR

I don't like that either, I would say dc controller should emit those events too

Issue #1053, opened 2019-10-19
file sending is broken


Simon-Laux:
it doesn't sent them, but all text messages after send the image/file.

MIntrovrt:
When trying to reproduce, it did send the file and message (unencrypted) and it seems to have gotten to the target mail server, since I got a mail from the target server telling me the file-type is forbidden (which was on purpose).

Simon-Laux:
strange stuff, anyways this needs investigation

Simon-Laux:
seems to be fixed already in current master

PR #1075, opened 2019-10-24
fix draft bug

Simon-Laux:
Co-author: @Jikstra


nicodh:
Which bug does it fix?

Simon-Laux:
#1053

Issue #1036, opened 2019-10-03
Add advanced options to configure certificate checks


link2xt:
Since options `imap_certificate_checks` and `smtp_certificate_checks` were added to the Rust core in https://github.com/deltachat/deltachat-core-rust/pull/647 they need to be added to the Desktop client.

Going to try to do it myself.

Jikstra:
You probably need to first add the constants to deltachat-node and update the rust core there. Can help you with this, if you want we can also do an audio/video hack session on this.

link2xt:
I did not any constants to FFI, going to use `0`, `1`, `2` and `3` for now :)

r10s:
cmp question at https://github.com/deltachat/deltachat-ios/issues/322

i think it is important to keep the non-power-users in mind. there are tons of self-signed click-hosters here in germany. so we have to choose the defaults very carefully. 

i think the idea @hpk42 pointed out somewhere else is reasonable: be lazy on default - but if we know due to the provider-database or by other sources (maybe autoconfigure) that the certificatiate should be good, we should check that.

i do not think it is of much worth to check _always_ and scare the users even before they're really onboarded. i think we do not need this additional ux barrier currently.

link2xt:
This setting goes into advanced settings, so users who want certificate checking can select "strict" mode. Default option is `0` which means Automatic. Now it means no certificate checks, but in the future it is going to use provider database and can be improved with TOFU.

pabzm:
This appears to have been implemented, closing it.

PR #1076, opened 2019-10-25
Add UI for {imap,smtp}_certificate_checks options

link2xt:
This PR is to solve #1036 issue


link2xt:
Waiting for #1082 to be merged first.

Also, `configure` function in `deltachat-node/lib/deltachat.js` needs to be extended to handle these options.

link2xt:
Blocked by https://github.com/deltachat/deltachat-node/pull/388 now

src/renderer/components/Login.js
```diff
@@ -206,6 +211,17 @@ export default class Login extends React.Component {
               <option value='starttls'>STARTTLS</option>
               <option value='plain'>{tx('off')}</option>
             </DeltaSelect>
+            <DeltaSelect
+              id='imap_certificate_checks'
+              label={tx('login_imap_certificate_checks')}
```
cyBerta: I think the entry for login_imap_certificate_checks is missing in the sources, or?

Jikstra:
@link2xt can you update the package.json to use the current deltachat-node git master? This way we can test master a couple of days without having a new -node release.

link2xt:
@Jikstra Done. Doing `npm install` and testing now, will try to rebase on top of master then.

src/renderer/components/Login.js
Jikstra: Yup seems like the translations (at least experimental/untranslated) are missing.

r10s:
i think there should be only _one_ option for the certificate checks, there is no good reason to be strict on the one but not on the other; i think this was the consens on irc together with @hpk42 @adbenitez and others.

on android, @cyBerta  just did a thing at https://github.com/deltachat/deltachat-android/pull/1096

Screen Shot 2019-11-01 at 12 09 35

technically, the option sets both core options, the current state is driven by `imap_certificate_checks`

link2xt:
Ok, I'll change the interface, add translations and mark it as ready for review once it works.

link2xt:
I have merged two options into one similarly to Android and added strings to `_untranslated_en.json`. By the way translated strings are already on Transifex.

Simon-Laux:
> By the way translated strings are already on Transifex.

looks like we could do a transifex pull again.

src/renderer/components/Login.js
```diff
@@ -57,21 +58,35 @@ export default class Login extends React.Component {
...
     const { mode } = this.props
     const { id, value } = event.target
-    const updatedState = update(this.state, {
-      credentials: {
+    let updatedCredentials;
+    if (id == 'certificate_checks') {
+      // Change to certificate_checks updates certificate checks configuration
+      // for all protocols.
+      updatedCredentials = {
+        ['imap_certificate_checks']: { $set: value },
```
Jikstra: The [] aren't needed in this case i think.

PR #1078, opened 2019-10-25
Preload some more chat list items as buffer

nicodh:
Without this I always saw "empty" items when scrolling. Now you have at least to scroll fast to see empty items...


src/renderer/components/chat/ChatListHelpers.js
```diff
@@ -73,6 +73,8 @@ export const useLazyChatListItems = chatListIds => {
 
   const chatIdsInView = (offset) => {
     let [indexStart, indexEnd] = getIndexStartEndInView()
+    // load 10 more than in view to have a buffer when scrolling
+    indexEnd += 10
```
Jikstra: The offset option is actually for exactly this case. So you can just increase offset when the chatIdsInView method is getting called.

Issue #469, opened 2018-12-20
Media view: Some video files shown in Documents


ralphtheninja:
`.webm` movie is shown in `Documents` and not in `Video`

![screenshot from 2018-12-20 23-59-13](https://user-images.githubusercontent.com/308049/50315546-80519c80-04b3-11e9-927e-c7d5027adac0.png)

karissa:
Do you know if the video was sent with desktop or android-dev? The way the file is classified is based on how it is sent, the `type` is sent from the sender. Core tries to make a good guess about it, though

ralphtheninja:
It was sent from desktop.

nicodh:
still exists in current master...

karissa:
Probably I would guess because desktop doesn't do a good enough job at sending the right type to core

link2xt:
I have tried sending a [sample WebM from Wikipedia](https://upload.wikimedia.org/wikipedia/commons/8/87/Schlossbergbahn.webm) to myself using the latest DeltaChat Desktop (commit cd2217d385d7f902deb90a9009a9aa45efa47024). Then I saved email to `.eml` file and decrypted it with `gpg -d email.eml`. The attachement is `Content-Type: application/octet-stream; name="Schlossbergbahn.webm"`. So the bug is indeed in the Desktop, it does not set the MIME type correctly.

link2xt:
DeltaChat Desktop does not set the optional mime argument of [`setFile` method](https://github.com/deltachat/deltachat-node#msgsetfilefile-mime):
https://github.com/deltachat/deltachat-desktop/blob/b9a5cf2d26200fc7195a204c2ec2f299a0c73c12/src/main/deltachat/messagelist.js#L9

`deltachat-core-rust` then tries to guess MIME type based on extension, but it only recognizes `.png`, `.jpg`, `.jpeg`, `.jpe` and `.gif`:
https://github.com/deltachat/deltachat-core-rust/blob/8a0fc609e6c1aa61947ee9fa965d87b104b6a716/src/dc_mimefactory.rs#L1209-L1226

The solution is probably to add proper MIME type support to DeltaChat Desktop and manually set the second argument of `setFile`, because there are cases when file extension may be incorrect or unavailable, for example if the file is pasted from the clipboard.

pabzm:
Could someone check if this still is broken?

link2xt:
@pabzm Just tried, sent the same Webm from wikipedia to myself, it goes into documents.

[...]

PR #1080, opened 2019-10-25
WIP: Detect mime type for files

nicodh:
related #469

I added the mime type as second parameter, but it does not help for webm files. I see the correct mime type in the msgs database table: 
c=1
f=$BLOBDIR/BBH_gravitational_lensing.webm
m=video/webm
but the file is not listed in the video group.
So maybe it's a core bug?

### DIFF
--- a/src/main/deltachat/messagelist.js
+++ b/src/main/deltachat/messagelist.js
@@ -2,13 +2,17 @@ const C = require('deltachat-node/constants')
 const log = require('../../logger').getLogger('main/deltachat/messagelist')
 const { integerToHexColor } = require('./util')
 const CHATVIEW_PAGE_SIZE = 20
+const mime = require('mime-types')
 
 const SplitOut = require('./splitout')
 module.exports = class DCMessageList extends SplitOut {
   sendMessage (chatId, text, filename, location) {
     const viewType = filename ? C.DC_MSG_FILE : C.DC_MSG_TEXT
     const msg = this._dc.messageNew(viewType)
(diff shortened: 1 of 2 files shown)


nicodh:
I also enabled sending a file without a message which was not possible before

Simon-Laux:
You're currently only setting the file mime type, where is the detection?

> I also enabled sending a file without a message which was not possible before

Do you mean the bug by that caused the file to be attached at every message that was send?
#1053
@Jikstra fixed it in #1075

nicodh:
@Simon-Laux detecting the mime type happens in the line const mimeType = mime.lookup(filename)
 and that mime type is passed as second argument to  msg.setFile(filename, mimeType)

nicodh:
Messages with empty text were not send because of the check if (message.match(/^\s*$/)) { ... send message
so a message with only a filename could not be sent. That's why I changed it to
 if (message.match(/^\s*$/) && !filename) {
https://github.com/deltachat/deltachat-desktop/pull/1080/files#diff-6839c55827d070d419c836f21ad49137L32

Simon-Laux:
Ok, but lets rather ask the core dev team if they could implement it in the core in the future, because it seems like its needed on all platforms.
Also mimetype detection over file name isn't the most reliable methode. the Magic number in the file could be taken into account too.

nicodh:
Should be implemented in core

hpk42:
it's better to keep UI issues open, and link to a core-issue rather
than closing.

PR #1081, opened 2019-10-25
Consistently capitalize "STARTTLS"

link2xt:
It is "STARTTLS" in all places but this one.


link2xt:
Pushed it as https://github.com/deltachat/deltachat-desktop/commit/e4dfa4152fce9bcb65092a8e775e19da474bf478

PR #1082, opened 2019-10-26
Fix server flags configuration

### LINKED ISSUE #1025: Advanced Settings Ignored
- Platform (linux/mac/windows): windows
- Delta Chat Desktop Version: DeltaChat-0.201.0-win64
- Expected behavior: Login to personal mail server sans SSL
- Actual behavior: SSL Failure
- Screenshots:
- Logs: https://pastebin.com/XFNVtCRA

Good day. 

I am trying to use my local mail server with Delta-Chat, my family uses it for internal communications and a few IoT uses...

It does not have a proper SSL chain, nor does it need one as it's most often accessed only on LAN. I figured like most other apps I could tell this program to use no security, and I saw that option here but when using it I fail to login with this error:

Could not connect to IMAP-server 192.168.2.21:143. (A certificate chain processed, but terminated in a root certificate which is not trusted by the trust provider. (os error -2146762487))

Other applications connect just fine without SSL, so I'm wondering if this is an issue with Deltachat trying for SSL even though I set security off?

link2xt:
This is an attempt to fix #1025

### DIFF
--- a/src/main/deltachat/login.js
+++ b/src/main/deltachat/login.js
@@ -110,8 +110,6 @@ function serverFlags ({ mailSecurity, sendSecurity }) {
     flags.push(C.DC_LP_SMTP_SOCKET_PLAIN)
   }
 
-  if (!flags.length) return null
-
   return flags.reduce((flag, acc) => {
     return acc | flag
   }, 0)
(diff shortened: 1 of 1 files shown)


link2xt:
I have tested that at least on the initial configuration, server flags are set successfully in the database:
```
17|server_flags|66048
```

They are not `undefined` or `null` anymore.

nicodh:
In fact the intial configuration is now set correct, but updating the configuration still doesn't work. Should be fixed in another PR

PR #1086, opened 2019-10-27
Advanced settings update

nicodh:
This is an attempt to fix the update advanced settings implementation

### DIFF
--- a/scss/_global.scss
+++ b/scss/_global.scss
@@ -181,6 +181,3 @@ code {
   }
 }
 
-.login-form button[type='submit'] {
-  margin-right: 1rem;
-}
--- a/scss/_global.scss
+++ b/scss/_global.scss
@@ -180,3 +180,7 @@ code {
     background-color: unset;
   }
 }
+
+.login-form button[type='submit'] {
+  margin-right: 1rem;
+}
--- a/src/main/deltachat/login.js
+++ b/src/main/deltachat/login.js
@@ -23,7 +23,7 @@ module.exports = class DCLoginController extends SplitOut {
     this._controller._sendStateToRenderer()
   }
 
-  login (credentials, sendStateToRenderer, coreStrings) {
+  login (credentials, sendStateToRenderer, coreStrings, updateConfiguration) {
     // Creates a separate DB file for each login
     this._controller.fullCwd = this.getPath(credentials.addr)
     log.info(`Using deltachat instance ${this._controller.fullCwd}`)
@@ -49,7 +49,7 @@ module.exports = class DCLoginController extends SplitOut {
         log.info('dc_get_info', dc.getInfo())
         sendStateToRenderer()
       }
-      if (!this._dc.isConfigured()) {
+      if (!this._dc.isConfigured() || updateConfiguration) {
         this._dc.once('ready', onReady)
         this._controller.configuring = true
         this._dc.configure(addServerFlags(credentials))
--- a/src/main/deltachat/login.js
+++ b/src/main/deltachat/login.js
@@ -20,17 +20,17 @@ module.exports = class DCLoginController extends SplitOut {
       this._dc.setStockTranslation(Number(key), strings[key])
     })
 
-    this._controller._render()
+    this._controller._sendStateToRenderer()
   }
 
-  login (credentials, render, coreStrings) {
+  login (credentials, sendStateToRenderer, coreStrings) {
     // Creates a separate DB file for each login
     this._controller.fullCwd = this.getPath(credentials.addr)
     log.info(`Using deltachat instance ${this._controller.fullCwd}`)
     const dc = new DeltaChat()
     this._controller._dc = dc
     this._controller.credentials = credentials
-    this._controller._render = render
+    this._controller._sendStateToRenderer = sendStateToRenderer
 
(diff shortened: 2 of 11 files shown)


scss/_global.scss
```diff
@@ -180,3 +180,7 @@ code {
     background-color: unset;
   }
 }
+
+.login-form button[type='submit'] {
```
nicodh: @Simon-Laux maybe move this to a better place - I am not so involved in css structure

src/main/deltachat/login.js
```diff
@@ -48,13 +48,13 @@ module.exports = class DCLoginController extends SplitOut {
...
         this._controller.emit('ready', this._controller.credentials)
         log.info('dc_get_info', dc.getInfo())
-        render()
+        sendStateToRenderer()
       }
-      if (!this._dc.isConfigured()) {
+      if (!this._dc.isConfigured() || updateConfiguration) {
         this._dc.once('ready', onReady)
         this._controller.configuring = true
         this._dc.configure(addServerFlags(credentials))
```
nicodh: Here I would expect, that the credentials are updated in the database but they are not...??

src/main/deltachat/settings.js
```diff
@@ -27,4 +27,16 @@ module.exports = class DCSettings extends SplitOut {
   keysExport (directory) {
     this._dc.importExport(C.DC_IMEX_EXPORT_SELF_KEYS, directory)
   }
+
```
nicodh: with this method we could update the settings in the database, but it won't update the prefixed keys (configured_mail_user etc)

src/main/deltachat/login.js
link2xt: configure() function three lines below the change calls set_config.

src/main/deltachat/settings.js
link2xt: Prefixed keys are updated by the core, there is no need to ever update them manually. If they are not updated, that means configuration didn't succeed

src/main/deltachat/login.js
nicodh: Yes but if I test it I don't get the updated values in the database...??

scss/_global.scss
Simon-Laux: currently login styles seem to be still in scss

Simon-Laux: src/renderer/components/Login-Styles.js

nicodh:
So what I found so far: if I uncomment this line https://github.com/deltachat/deltachat-node/blob/master/lib/deltachat.js#L99 advanced settings are updated except the server flags. For the server flags I see the value for "server_flags" updated as expected, but not for "configured_server_flags". No matter what the server_flags value is, the configured_server_flags value always is 66052 (DC_LP_IMAP_SOCKET_SSL & DC_LP_SMTP_SOCKET_STARTTLS).
Maybe a core bug?

src/main/ipc.js
```diff
@@ -12,7 +12,7 @@ const localize = require('../localize')
 const menu = require('./menu')
 const windows = require('./windows')
 const log = require('../logger').getLogger('main/ipc', true)
-const DeltaChat = (() => {
+const DeltaChatController = (() => {
```
Simon-Laux: +1

scss/_global.scss
Simon-Laux: or you can already create a new scss file like scss/login/login-styles.scss
and don't forget to add the file to manifest.scss when you do that.
(you don't need to move everything there, only creating it for this class if fine for now)

nicodh:
I also wonder why we need this temporary DC instance to test the login.
https://github.com/deltachat/deltachat-desktop/blob/master/src/main/ipc.js#L221
That's something DC core does anyway. If the new configured values are not working, the "configured_[keys]" in the DB  are not updated. So it's no problem to try to update the configuration immediately for the current account...

src/renderer/components/Login.js
```diff
@@ -139,7 +147,7 @@ export default class Login extends React.Component {
     return (
       <React.Fragment>
         {this.renderLoginHeader(mode)}
-        <form onSubmit={this.handleSubmit}>
+        <form onSubmit={this.handleSubmit} className={'login-form'}>
```
Jikstra: className='login-form' is enough

src/main/deltachat/settings.js
```diff
@@ -2,6 +2,16 @@ const C = require('deltachat-node/constants')
 const log = require('../../logger').getLogger('main/deltachat/settings', true)
 
 const SplitOut = require('./splitout')
+
+const serverFlagMap = {
```
link2xt: I don't get what this map is for. Nobody uses strings like `mail_security_ssl` anywhere, they are never passed to `getConfigFor`.

What is the reason for this server flags refactoring?

nicodh: It's just for displaying the values in the Settings dialog. DC core uses the flags but humans need them to be translated into the according string values. So we get stings (like "ssl"or "plain") from the frontend and have to pass a bitmask to the dc core and vice versa

link2xt:
Just tested, server flags configuration is fixed.

During login, "context not open" errors are displayed.

PR #1096, opened 2019-11-01
Fix key warnings in message list related components

Jikstra:
There is still one more bug, that we have some messageids duplicated. I'm tracking it down, but didn't find it yet.


Simon-Laux:
whats the relation to #1099

Jikstra:
> whats the relation to #1099 ?

None really. Besides this all happend on the same branch but besides that nada.

PR #1100, opened 2019-11-03
Restyle user-feedback

Jikstra:
Error:
Screenshot_20191103_140240

Success:
Screenshot_20191103_140358

Default:
Screenshot_20191103_140426


Simon-Laux:
text could be a bit bigger, also the opacity of `default` can be a bit more opaque.

PR #1105, opened 2019-11-05
Do not call messageToJson for DC_MSG_ID_DAYMARKER

### LINKED ISSUE #1106: get_chat_msgs results contain ID 9 sometimes
DC desktop throws this warning when rendering any chat for me:
```
12.9s [D]core/event: DC_EVENT_WARNING  dc_get_msg called with special msg_id=9, returning empty msg
12.9s [D]main/deltachat: sendToRenderer: DC_EVENT_WARNING
12.9s [w]main/deltachat: dc_get_msg called with special msg_id=9, returning empty msg
```

That happens because `DCMessageList._messagesToRender` is called with a list that contains ID 9. These IDs come from `this._dc.getChatMessages`, which calls `get_chat_msgs` via node bindings.

For some reason, my database contains special IDs in `msgs` table:
```
$ sqlite3 ~/.config/DeltaChat/*/db.sqlite 'select * from msgs where id<=9'
1|||0|0|0|0|0|0|0|0|0||||0|0|0|0||||1|0
2|||0|0|0|0|0|0|0|0|0||||0|0|0|0||||1|0
3|||0|0|0|0|0|0|0|0|0||||0|0|0|0||||1|0
4|||0|0|0|0|0|0|0|0|0||||0|0|0|0||||1|0
5|||0|0|0|0|0|0|0|0|0||||0|0|0|0||||1|0
6|||0|0|0|0|0|0|0|0|0||||0|0|0|0||||1|0
7|||0|0|0|0|0|0|0|0|0||||0|0|0|0||||1|0
8|||0|0|0|0|0|0|0|0|0||||0|0|0|0||||1|0
9|||0|0|0|0|0|0|0|0|0||||0|0|0|0||||1|0
```

I don't know if that is because of the bug in some previous version of DC because I restore from backup, but still, my backup and backups/databases of others may contain these IDs.

Not sure how to solve this. The simplest solution seems to be to filter out these IDs in SQL queries. Maybe we can delete them from the database?

link2xt:
Fixes #1106

### DIFF
--- a/src/main/deltachat/messagelist.js
+++ b/src/main/deltachat/messagelist.js
@@ -56,16 +56,19 @@ module.exports = class DCMessageList extends SplitOut {
     const messages = []
     for (let i = 0; i < messageIdsToRender.length; i++) {
       const id = messageIdsToRender[i]
-      const json = this.messageIdToJson(id)
-      if (id === C.DC_MSG_ID_DAYMARKER && i > 0) {
-        if (messages[i - 1]) {
-          json.daymarker = {
-            timestamp: messages[i - 1].msg.timestamp,
-            id: 'd' + i
-          }
+      if (id !== C.DC_MSG_ID_DAYMARKER) {
+        messages.push(this.messageIdToJson(id))
+      } else if (i > 0 && messages[i - 1]) {
+        const json = {
+          id,
+          msg: { text: '' }
         }
+        json.daymarker = {
+          timestamp: messages[i - 1].msg.timestamp,
+          id: 'd' + i
+        }
+        messages.push(json)
       }
-      messages[i] = json
     }
     // remove the additional message again
     messages.shift()


hpk42:
Thanks for fixing that, was annoying me as well -- it actually goes back to a core refactoring -- and the id9 thing would crash desktop, so when fixing it i inserted this warning.

src/main/deltachat/messagelist.js
```diff
@@ -56,16 +56,19 @@ module.exports = class DCMessageList extends SplitOut {
...
     for (let i = 0; i < messageIdsToRender.length; i++) {
       const id = messageIdsToRender[i]
-      const json = this.messageIdToJson(id)
-      if (id === C.DC_MSG_ID_DAYMARKER && i > 0) {
-        if (messages[i - 1]) {
-          json.daymarker = {
-            timestamp: messages[i - 1].msg.timestamp,
-            id: 'd' + i
-          }
+      if (id !== C.DC_MSG_ID_DAYMARKER) {
```
flub: I don't know how generic this code is, but there are other special message IDs that could be returned, `DC_MSG_ID_MARKER1` being the obvious one, but perhaps you should check `id <= DC_MSG_ID_LAST_SPECIAL` instead?

link2xt: I am preparing a better fix: #1114

PR #1115, opened 2019-11-06
rename "logout" to "switch account"

Simon-Laux:

### DIFF
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -31,5 +31,8 @@
   },
   "about_licensed_under_desktop": {
     "message": "This software is licensed under GNU GPL version 3 and the source code is available on GitHub."
+  },
+  "switch_account_desktop":{
+    "message": "Switch Account"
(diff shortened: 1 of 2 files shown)


lefherz:
Great idea, will test it out directly.

Maybe an issue at https://github.com/deltachat/interface/issues would be a good idea to discuss streamlining this behavior or so? Or will the other apps just copy desktop behavior because it is the first implementation of multi-account?

lefherz:
Tested it, looks great.

_locales/_untranslated_en.json
```diff
@@ -31,5 +31,8 @@
   },
   "about_licensed_under_desktop": {
     "message": "This software is licensed under GNU GPL version 3 and the source code is available on GitHub."
+  },
+  "switch_account_desktop":{
+    "message": "Switch Account"
```
OzancanKaratas: Account's A should be lowercase. Because all strings so.

PR #1120, opened 2019-11-06
fix message rendering new line behaviour

### LINKED ISSUE #1119: empty lines are not preserved
separating paragraphs using an empty line does not work, multiple linefeeds seems to be converted to a single one.

```
this is a paragraph.

and this another.
```

is converted to
```
this is a paragraph.
and this another.
```

i think it is okay to convert >2 linefeeds to 2 linefeeds, but 1 or two linefeeds should be allowed.

Simon-Laux:
fixes #1119

### DIFF
--- a/scss/message/_message.scss
+++ b/scss/message/_message.scss
@@ -633,9 +633,16 @@ $color-message-icon-error: var(--errorColor);
     background-color: transparent;
 }
 
-.module-message .line-break {
-  height: 14px;
-  &:last-child {
+.module-message {
+  .double-line-break {
+    height: 28px;
+  }
+  .line-break {
+    height: 14px;
+  }
+
+  .line-break:last-child {
     height: 0px;
   }
 }
+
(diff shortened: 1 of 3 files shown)


Simon-Laux:
I would like to move the message parser (markdown and links) to the core. But I'll need to learn how to code parsers in rust first.

hpk42:
On Wed, Nov 06, 2019 at 14:53 -0800, Simon Laux wrote:
> I would like to move the message parser (markdown and links) to the core. But I'll need to learn how to code parsers in rust first.

pretty sure there is a crate for parsing markdown ;)

Simon-Laux:
We'll have our own flavor of markdown, so I believe it makes sense to do it from scratch. Also for the learning experience.

Jikstra:
Please give it a try for learning experience, but our markdown flavour isn't that much different, and if so, i would prefer to drop that and stay with standard markdown.

Jikstra:
But links are supported in markdown? If only mentions is the problem, we can hack around that without writing our own markdown parser (which is probably not as easy to get it properly working, especially if you never did that. Still a very good thing to try and a good side project).

Simon-Laux:
links with `[description](link)` we won't and shouldn't support (security for average users).
In essence its a custom flavor of markdown. Also as I wrote, we need a special AST format, html won't do it, because we want to use it in iOS and android too. Anyways that markdown parser isn't too important right now. I'll work on #1123 first, before starting the message parser.

PR #1124, opened 2019-11-07
WIP: Refactoring preview builds GitHub action

lefherz:
Thanks @arnested for this suggestion!

reference: https://github.community/t5/GitHub-Actions/Publishing-download-links-in-the-check-details-instead-of/m-p/37273/highlight/false#M2880

### DIFF
--- a/.github/workflows/build-preview.yml
+++ b/.github/workflows/build-preview.yml
@@ -24,7 +24,6 @@ jobs:
         export PULLREQUEST_ID=$(echo "${{ github.ref }}" | cut -d "/" -f3)
         echo ::set-output name=prid::$PULLREQUEST_ID
         if [ $(expr length "${{ secrets.USERNAME }}") -gt "1" ]; then echo ::set-output name=uploadtoserver::true ;fi
-        md5sum package-lock.json > pkgsum
     - name: Use Node.js 10.x
       uses: actions/setup-node@v1
       with:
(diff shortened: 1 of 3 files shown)


lefherz:
@Simon-Laux refactored it a bit, as you complained about code prettiness ;) do we actually have coding guidelines somewhere? e.g. I didn't respect the 80 chars max screen width I know from python.

Jikstra:
> @Simon-Laux refactored it a bit, as you complained about code prettiness ;) do we actually have coding guidelines somewhere? e.g. I didn't respect the 80 chars max screen width I know from python.

For js we use the standard guidelines. You can start the linter by `npx standard`

.github/workflows/build-preview.yml
```diff
@@ -61,32 +58,39 @@ jobs:
...
         curl -f -s -X POST --header "authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" --url "$GITHUB_API_URL" --header "content-type: application/json" --data "$STATUS_DATA"
-        export STATUS_DATA="{\"state\": \"success\", \"description\": \"Download it here:\", \"context\": \"Windows Preview Build\", \"target_url\": \"${EXE_URL}\"}"
+        # Post Windows download link to check details
+        export STATUS_DATA="{\"state\": \"success\", \
+                             \"description\": \"Download it here:\", \
+                             \"context\": \"Windows Preview Build\", \
+                             \"target_url\": \"${EXE_URL}\"}"
         curl -f -s -X POST --header "authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" --url "$GITHUB_API_URL" --header "content-type: application/json" --data "$STATUS_DATA"
+        if md5sum -c pkgsum --status; then 
+            export STATUS_DATA="{\"state\": \"failed\", \  # this value is probably not valid
```
Simon-Laux: `# this value is probably not valid`
resolve this comment before merge?

lefherz: I don't have time before next week, but yes, I will resolve this.

lefherz:
Isn't the point that the package-lock.json is changed through npm install? and that it's concerning if it doesn't change?

In that case we could just save the first sum after the `change build name and Product Name to development id` step instead of beforehand. But maybe I got it totally wrong, I have no experience with npm at all, and pretty much guessed what kind of action you need^^

Simon-Laux:
> Isn't the point that the package-lock.json is changed through npm install? and that it's concerning if it doesn't change?

its actually the other way around - it must not change.

Simon-Laux:
@missytake whats the progress of this pr? Is it done and we just need to resolve the conflicts?

missytake:
quite far iirc. I can look into it over the weekend

lefherz:
Hm, for some reasons the package-lock file I build locally and the package-lock file the check-lock action builds differ, even though we use the same node version o.0 any explanations?

```
v10.17.0
Checksum before npm i:
85466864afe574c48a03c5986020eab6 package-lock.json
Checksum after npm i:
be99f5b99d46749eba83e079109dea30  package-lock.json
```

PR #1125, opened 2019-11-07
Make appdata file compliant with the validator

muelli:
These changes are necessary to make the validator happy.

It'd be nice if the release would be filled out, though.

cf. https://github.com/hughsie/appstream-glib/issues/333 and https://github.com/flathub/chat.delta.desktop/pull/23

### DIFF
--- a/static/chat.delta.desktop.appdata.xml
+++ b/static/chat.delta.desktop.appdata.xml
@@ -25,7 +25,7 @@
   <launchable type="desktop-id">chat.delta.desktop.desktop</launchable>
   <screenshots>
     <screenshot type="default">
-      <image>https://github.com/deltachat/deltachat-desktop/raw/master/screenshot.png</image>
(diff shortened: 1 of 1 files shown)


static/chat.delta.desktop.appdata.xml
```diff
@@ -25,7 +25,7 @@
   <launchable type="desktop-id">chat.delta.desktop.desktop</launchable>
   <screenshots>
     <screenshot type="default">
-      <image>https://github.com/deltachat/deltachat-desktop/raw/master/screenshot.png</image>
+      <image>https://github.com/deltachat/deltachat-desktop/raw/master/README_ASSETS/screenshot.png</image>
```
Simon-Laux: how about changing this to an absolute commit to prevent a missing image when its path should change again?

Jikstra: But then we need to update the path on every new screenshot push

PR #1131, opened 2019-11-11
Fix #1130

### LINKED ISSUE #1130: subtitle disappears in chatview
The first picture is when i select the dev chat. 
one

the second picture is after i typed a few characters into the message input box. 

two

the "18 members" is gone.

Simon-Laux:
fix #1130
the issue was that the chatlistitem was used to describe a new chat state.

### DIFF
--- a/src/main/deltachat/chatlist.js
+++ b/src/main/deltachat/chatlist.js
@@ -27,19 +27,8 @@ module.exports = class DCChatList extends SplitOut {
   }
 
   onChatModified (chatId) {
-    const listFlags = 0
-    const list = this._dc.getChatList(listFlags, '')
-    const chatList = []
-    let i = -1
-    for (let counter = 0; counter < list.getCount(); counter++) {
-      const id = list.getChatId(counter)
-      chatList.push(id)
-      if (id === chatId) i = counter
-    }
-    if (i === -1) return undefined
-    // const chat = this.getChatListItemById(chatId, list, i) <- this should never be called
-    // this._controller.sendToRenderer('DD_EVENT_CHAT_MODIFIED', { chatId, chat })
-    // did we mean to send the full chat?
+    const chat = this.getFullChatById(chatId)
+    this._controller.sendToRenderer('DD_EVENT_CHAT_MODIFIED', { chatId, chat })
   }
 
   getChatListIds (listFlags, queryStr, queryContactId) {
--- a/src/main/deltachat/chatlist.js
+++ b/src/main/deltachat/chatlist.js
@@ -37,8 +37,9 @@ module.exports = class DCChatList extends SplitOut {
       if (id === chatId) i = counter
     }
     if (i === -1) return
-    const chat = this.getSmallChatById(chatId, list, i)
-    this._controller.sendToRenderer('DD_EVENT_CHAT_MODIFIED', { chatId, chat })
+    // const chat = this.getChatListItemById(chatId, list, i) <- this should never be called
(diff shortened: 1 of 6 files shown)


src/main/deltachat/chatlist.js
```diff
@@ -81,62 +83,64 @@ module.exports = class DCChatList extends SplitOut {
...
 
-    if (draft) {
-      chat.draft = draft.getText()
-    } else {
-      chat.draft = ''
-    }
+  getFullChatById (chatId, loadMessages) {
+    const chat = this.__getChatById(chatId)
+    if (chat === null) return null
+    this._controller._pages = 0
```
Simon-Laux: why is a global var change in a getter function? I know its a parameter for `this._controller.messageList._messagesToRender(messageIds)` but saving it globally seems not so good practice in my eyes.

Jikstra: Which variable is global? I know that we somewhere have a hack to save the chatlist object somewhere as a global variable because we otherwise need to refetch it again and again.

src/main/deltachat/chatlist.js
```diff
@@ -36,9 +36,10 @@ module.exports = class DCChatList extends SplitOut {
       chatList.push(id)
       if (id === chatId) i = counter
     }
-    if (i === -1) return
-    const chat = this.getSmallChatById(chatId, list, i)
-    this._controller.sendToRenderer('DD_EVENT_CHAT_MODIFIED', { chatId, chat })
+    if (i === -1) return undefined
+    // const chat = this.getChatListItemById(chatId, list, i) <- this should never be called
+    // this._controller.sendToRenderer('DD_EVENT_CHAT_MODIFIED', { chatId, chat })
+    // did we mean to send the full chat?
```
Jikstra: Where do we return the chatList? Is this pr done?

src/main/deltachat/chatlist.js
Simon-Laux: `this._controller._pages` global relativ to the dc contoller

src/main/deltachat/chatlist.js
Simon-Laux: returns undefined, because otherwise standard complains.
I left the invalid code there (commented out), because maybe you still need it and want to fix it using the full chat object.

src/main/deltachat/chatlist.js
Jikstra: yup this is ugly code. This whole stuff is near a refactoring, i want to have a method getChatListItem and getMessages. Currently it's spread over various places and reuses code/fucntions because of the old state architecture.

src/main/deltachat/chatlist.js
Jikstra: But we don't send anything to the frontend anymore.

Simon-Laux: no, we don't I assume. I could change that before merge this (using the complete chat object.) should I solve it that way?

Jikstra: You removed the line sending the object to the frontend, so i guess you should add it back :D Sorry i commented on the wrong line, you removed it on line 41 https://github.com/deltachat/deltachat-desktop/pull/1131/files/b22982600f8acd41fe52e0273e0dab9dc49a455b#diff-a34cb1175ba7e32523eea0092de1dfc6L41

Jikstra:
As my pr evolved to use new core api which isn't there yet, lets merge this first :)

Issue #1136, opened 2019-11-13
Automate packaging Process for releases


Simon-Laux:
- [x] Mac OS with signing
 -> https://kilianvalkhof.com/2019/electron/notarizing-your-electron-application/ looks like good guide
- [x] windows setup
- [x] windows portable (see https://support.delta.chat/t/please-make-deltachat-for-windows-a-truly-portable-standalone-program/675/6)
- [x] Linux AppImage
- [ ] Linux AppImage deltas to last versions for updates?
- [x] Linux .Deb
- [ ] windows .appx for windows store

Jikstra:
@missytake

missytake:
> Linux AppImage deltas to last versions for updates

@Simon-Laux does the build process produce this already?

missytake:
>  Mac OS with signing -> https://kilianvalkhof.com/2019/electron/notarizing-your-electron-application/ looks like good guide

Can we add such a certificate to our GitHub secrets? Or do we want to do this on a separate machine?

Simon-Laux:
Seperate machine of course

Am 8. Dezember 2019 14:55:26 MEZ schrieb missytake <notifications@github.com>:
>>  Mac OS with signing ->
>https://kilianvalkhof.com/2019/electron/notarizing-your-electron-application/
>looks like good guide
>
>Can we add such a certificate to our GitHub secrets? Or do we want to
>do this on a separate machine?
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/deltachat/deltachat-desktop/issues/1136#issuecomment-562951016

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

missytake:
Okay, maybe we should merge #1186. There are still 3 checkboxes not included in that PR, but they may not be that important for now:

- macOS build with signing and notarizing can not be done with a github action alone
- AppimageUpgrade would be nice, but no merge-blocker for now
- Windows Appstore build: what's the status here?

Simon-Laux:
you must build the appx on a windows runner.
please apply this diff for the appx to work [diff-get-on-msstore.txt](https://github.com/deltachat/deltachat-desktop/files/3940533/diff-get-on-msstore.txt)

missytake:
hm, can you please reupload the file without colors? I constantly get this error:

```
$ git apply diff-get-on-msstore.txt
error: unrecognized input
```

[...]

PR #1138, opened 2019-11-13
add command to make portable builds

Simon-Laux:
related to https://support.delta.chat/t/please-make-deltachat-for-windows-a-truly-portable-standalone-program/675/8

### DIFF
--- a/bin/prepare-build-portable.js
+++ b/bin/prepare-build-portable.js
@@ -1,0 +1,13 @@
+const { join } = require('path')
+const fs = require('fs')
+const appConfig = join(__dirname, '../src/application-config.js')
+
+const action = process.argv[2] === 'undo' ? 'undo' : 'do'
+
+const fileContent = fs.readFileSync(appConfig, 'utf-8')
+  .replace(
+    /const portable = (?:false|true)/,
+    `const portable = ${JSON.stringify(action === 'do')}`
+  )
+
+fs.writeFileSync(appConfig, fileContent)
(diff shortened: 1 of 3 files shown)


Jikstra:
I would suggest having one script which runs all commands to build all install files (skipping .dmg if we're not on a mac). I think this makes more sense than having x npm run scripts.

Simon-Laux:
> I would suggest having one script which runs all commands to build all install files (skipping .dmg if we're not on a mac). I think this makes more sense than having x npm run scripts.

I totally agree, but I don't want to fiddle around with spawn process now. Let's keep this in mind for #1136

ralphtheninja:
So this is done because a single user wanted a portable build?

Simon-Laux:
I though it would be usefull even before they asked for it. Its useful in many cases and it gives freedom. Also for one user that speaks, most of the time many want it too.
I worked on the dl site and I can remove an ugly disclaimer there when providing the portable build.

Am 14. November 2019 00:21:09 MEZ schrieb Lars-Magnus Skog <notifications@github.com>:
>So this is done because a single user wanted a portable build?
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/deltachat/deltachat-desktop/pull/1138#issuecomment-553650448

-- 

STPKITT:
@ralphtheninja: Better then not doing it just because you don't like it. I mean wth? Btw I came up with this topic in the German DeltaChat XMPP group and multiple people wanted it. Only then I requested this in the Delta.Chat forum.

Simon-Laux:
Don't panic. Lets keep this conversation civil.

> Better then not doing it just because you don't like it

Don't over-interpret @ralphtheninja's question @STPKITT. Lets assume its just a polite question.

ralphtheninja:
@Simon-Laux @STPKITT is right. I could have used a better tone when asking this. Sorry about this.

STPKITT:
All good :-)

PR #1141, opened 2019-11-16
Use callDcMethod for map click callback

### LINKED ISSUE #1127: Map: freeze when clicking on POI
mouse has hand icon, but I can't click anything anymore

nicodh:
resolves #1127

### DIFF
--- a/src/renderer/components/map/MapComponent.js
+++ b/src/renderer/components/map/MapComponent.js
@@ -2,7 +2,6 @@ const { callDcMethod } = require('../../ipc')
 const React = require('react')
 const ReactDOMServer = require('react-dom/server')
 const ReactDOM = require('react-dom')
-const { ipcRenderer } = require('electron')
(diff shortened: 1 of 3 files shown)


src/renderer/components/map/MapComponent.js
```diff
@@ -257,22 +256,24 @@ class MapComponent extends React.Component {
     })
     if (contactFeature) {
       if (contactFeature.properties.msgId) {
-        const messageObj = await ipcRenderer.sendSync(
```
Jikstra: I guess switching to callDcMethod allows us to delete the old ipc event?

nicodh: which "old ipc event" do you mean? This one:
https://github.com/deltachat/deltachat-desktop/blob/master/src/main/ipc.js#L106
that wasn't called here before this change...?
But seems to be obsolete anyway...

PR #1144, opened 2019-11-17
Fix selecting a file to import a secret key.

### LINKED ISSUE #1057: Import secret key?
On rust-master and released .deb one apparently can't select a secret key for import in the settings. cc @toberndo  what is the exact pattern there that filters the selection -- is there a bug?

pabzm:
Closes #1057


Jikstra:
Thanks :)

PR #1145, opened 2019-11-18
Save log to file containing correct day and month.

pabzm:
`getMonth()` starts at 0, and `getDay()` returns the day of the week.

### DIFF
--- a/src/main/log-handler.js
+++ b/src/main/log-handler.js
@@ -10,8 +10,8 @@ function logName () {
   }
   const fileName = [
     `${d.getFullYear()}-`,
-    `${pad(d.getMonth())}-`,
-    `${pad(d.getDay())}-`,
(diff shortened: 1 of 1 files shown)


src/main/log-handler.js
```diff
@@ -10,8 +10,8 @@ function logName () {
   }
   const fileName = [
     `${d.getFullYear()}-`,
-    `${pad(d.getMonth())}-`,
-    `${pad(d.getDay())}-`,
+    `${pad(d.getMonth() + 1)}-`,
+    `${pad(d.getDate())}-`,
```
hpk42: is getDate() correct here? Looks like it should be getDay()

pabzm: It is correct, Javascript it just stupid: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getDate

PR #1146, opened 2019-11-18
Map icons

### LINKED ISSUE #1126: Add icon on message that has location
like on android

nicodh:
- Single letters are rendered as markers
- POI have an image icon (you are welcome to upload a better one ;-)
- messages with a location have a marker icon in message list (closes #1126)
- terrain layer is removed now

map-marker


Simon-Laux:
> messages with a location have a marker icon in message list

cool, though I'd like if its only shown when location streaming is enabled in the settings.

nicodh:
I just pass the settings flag down to the MessageWrapper for now, since that part will be refactored anyway...

PR #1147, opened 2019-11-20
Refactor message2

Simon-Laux:
see commit messages for details.
In essence I moved more stuff between files in hope to make the code more readable and prepare merging of message-wrapper and message.

### DIFF
--- a/src/renderer/components/message/Message.js
+++ b/src/renderer/components/message/Message.js
@@ -164,6 +164,7 @@ const Message = (props) => {
     status,
     attachment
   } = props
+  const tx = window.translate
 
   const authorAddress = message.contact.address
 
@@ -213,7 +214,7 @@ const Message = (props) => {
           message }} />
 
         <div dir='auto' className='text' >
-          <MessageBody text={text || ''} />
+          {message.msg.isSetupmessage ? tx('autocrypt_asm_click_body') : <MessageBody text={text || ''} />}
         </div>
         {longMessage && <button onClick={onShowDetail}>...</button>}
         <MessageMetaData {...props} />
--- a/src/renderer/components/message/Message.js
+++ b/src/renderer/components/message/Message.js
@@ -8,7 +8,7 @@ const MessageMetaData = require('./MessageMetaData')
 
 const ContactName = require('../conversations/ContactName')
 const { ContextMenu, ContextMenuTrigger, MenuItem } = require('react-contextmenu')
-const Attachment = require('./Attachment')
+const Attachment = require('./Attachment').default
 
 const Avatar = ({ contact }) => {
   const {
@@ -160,7 +160,8 @@ const Message = (props) => {
     message,
     text,
     disableMenu,
-    status
+    status,
+    attachment
   } = props
 
   // This id is what connects our triple-dot click with our associated pop-up menu.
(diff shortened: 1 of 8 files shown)


src/renderer/components/message/Message.js
```diff
@@ -123,13 +124,15 @@ const contextMenu = (props, textSelected, triggerId) => {
...
-        <MenuItem onClick={onDownload}>
+        <MenuItem onClick={onDownload.bind(null, message.msg)}>
           {tx('download_attachment_desktop')}
         </MenuItem>
       ) : null}
+      {/*
       <MenuItem onClick={onReply}>
         {tx('reply_to_message_desktop')}
       </MenuItem>
+       */}
```
Simon-Laux: this should be in the next release

src/renderer/components/message/MessageList.js
```diff
@@ -3,9 +3,9 @@ import MessageWrapper from './MessageWrapper'
 import ScreenContext from '../../contexts/ScreenContext'
 import { callDcMethod } from '../../ipc'
 import chatStore from '../../stores/chat'
-import { isDisplayableByRenderMedia } from './Attachment'
-import { shell } from 'electron'
-import logger from '../../../logger'
+
+import { getLogger } from '../../../logger'
+const log = getLogger('render/msgList')
```
Simon-Laux: I fixed the logger here.

src/renderer/components/message/MessageList.js
```diff
@@ -75,17 +75,6 @@ export default function MessageList ({ chat, refComposer, locationStreamingEnabl
...
-      openDialog('RenderMedia', { message })
-    } else {
-      if (!shell.openItem(message.msg.attachment.url)) {
-        logger.info("file couldn't be opened, try saving it in a different place and try to open it from there")
-      }
-    }
-  }
-
   const onClickSetupMessage = setupMessage => openDialog('EnterAutocryptSetupMessage', { setupMessage })
   const onShowDetail = message => openDialog('MessageDetail', { message, chat })
```
Simon-Laux: why do we have to reuse the chat object here? I would appreciate it when that dialog could fetch it from the core again, that would allow us to move this function around and inside of the message.

Simon-Laux: same on delete message, the dependency to the chatStore is annoying

Simon-Laux: Also I would appreciate if we could call openDialog without a ScreenContext.

Jikstra: > 

why do we have to reuse the chat object here? I would appreciate it when that dialog could fetch it from the core again, that would allow us to move this function around and inside of the message.


The message object should be enough for both? I think we shouldn't refetch anything that we already have, also because this could introduce inconsistencies if something changed in between. ChatStore is annoying anyways... openDialog without a screenContext sounds not so easy for me, how do you want to achieve this? Global variable?

src/renderer/components/message/messageFunctions.js
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * @param {MsgObject} msg
+ */
+export function onDownload (msg) {
+  const defaultPath = path.join(remote.app.getPath('downloads'), path.basename(msg.file))
+  remote.dialog.showSaveDialog({
+    defaultPath
+  }, (filename) => {
+    if (filename) ipcRenderer.send('saveFile', msg.file, filename)
```
Jikstra: Can you use the ipcBackend variable from the rendered/ipc file? This makes it easier to switch to websockets/whatever at some point.

Simon-Laux: doesn't need to be done now IMO, I just copied it over and we need to revisit this soon anyway, to remove the `remote` thing

PR #1151, opened 2019-11-22
Set profile image

Simon-Laux:
also fix the instant closing bug of the color picker.

The styling could be improved before releasing this in the wild. The plan is to use symbols instead of the text and style the buttons similar to the other buttons.

### DIFF
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -34,5 +34,8 @@
   },
   "switch_account_desktop":{
     "message": "Switch account"
+  },
+  "select_profile_image_desktop": {
+    "message": "Select your new Profile image"
   }
 }
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -36,7 +36,7 @@
     "message": "Switch account"
   },
   "select_profile_image_desktop": {
-    "message": "Select your new Profile image"
+    "message": "Select your new profile image"
   },
   "a11y_profile_image_label":{
     "message": "Your profile image"
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -37,5 +37,14 @@
   },
   "select_profile_image_desktop": {
     "message": "Select your new Profile image"
+  },
+  "a11y_profile_image_label":{
+    "message": "Your profile image"
+  },
+  "a11y_profile_image_select":{
+    "message": "set a new profile image"
+  },
+  "a11y_profile_image_remove":{
+    "message": "remove your profile image"
   }
 }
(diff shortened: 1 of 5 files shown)


_locales/_untranslated_en.json
```diff
@@ -34,5 +34,17 @@
   },
   "switch_account_desktop":{
     "message": "Switch account"
+  },
+  "select_profile_image_desktop": {
```
OzancanKaratas: Please remove “desktop” prefix. Other clients will use this.

_locales/_untranslated_en.json
```diff
@@ -34,5 +34,17 @@
...
+    "message": "Select your new Profile image"
```
OzancanKaratas: Fix typo. P must be small.

_locales/_untranslated_en.json
Simon-Laux: this is desktop specific, its the title of the file picker window.

_locales/_untranslated_en.json
```diff
@@ -34,5 +34,17 @@
   },
   "switch_account_desktop":{
     "message": "Switch account"
+  },
+  "select_profile_image_desktop": {
+    "message": "Select your new profile image"
+  },
+  "a11y_profile_image_label":{
```
r10s: do we really need a a11y string for every button and control? maybe, we could just use the normal "Your profile image" or just "Profile image" string here.

_locales/_untranslated_en.json
r10s: hm, in general i agree with @OzancanKaratas , we should create os-specific strings only if it happens that we really need them. for most other hints etc. we do not have specific stuff.
This "Select your new profile image" might be useful on other os as well.
however, we can streamline this when adding it to androids strings.xml, this should not block merging.

_locales/_untranslated_en.json
Simon-Laux: every button that will have no text, plan is to use a symbol instead of the text.

Jikstra:
Styling is a bit ugly but first make it work, then make it look good. BTW, where did the border of the dialog header go? Did you & evelyn remove it at some point?

_locales/_untranslated_en.json
r10s: k, but the having a simple `your_profile_image` would do the job - if can be used as a input field label as well as for a11y. i would use explicit a11y strings only if they are really different from some labels, same as for _desktop, _android etc. suffixes.

r10s:
> Approving besides the a11y thingies, not sure what to do about them.

as the strings are currently only in untranslated.* it should be no problem, we can figure things out these days when merging the strings.

PR #1153, opened 2019-11-24
add "Copy link" into context menu

Simon-Laux:
when right clicking on a link

### DIFF
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -1,4 +1,7 @@
 {
+  "menu_copy_link_to_clipboard": {
+    "message": "Copy link"
+  },
   "messaging_disabled_not_in_group": {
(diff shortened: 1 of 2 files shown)


_locales/_untranslated_en.json
```diff
@@ -1,4 +1,7 @@
 {
+  "menu_copy_link_to_clipboard": {
+    "message": "Copy Link"
```
Jikstra: link should be small

OzancanKaratas: Yes @Jikstra. This should be "Copy link". You're welcome, but it would be quicker to fix it than to say it. But I don't know what I have to do to be with you. Unfortunately, I can't convince you.

Simon-Laux: telegram has it uppercase but whatever I make it small/

PR #1154, opened 2019-11-26
Add missing groupName argument to ShowQrCodeInnerDialog

link2xt:


link2xt:
Without this, group name is not displayed in the QR code dialog for verified groups, it's just `""`.

PR #1159, opened 2019-11-30
Feat dialogify blocked contacts

Jikstra:
- [ ] Glitch when scrolling down, probably some more. Not sure if i want to fix this now, as this pr is still a good improvement


scss/dialogs/_delta-dialog.scss
```diff
@@ -0,0 +1,26 @@
...
+.delta-dialog-card {
+  background-color: var(--bp3DialogHeaderBg); 
+  padding: 20px;
+  height: 100%;
+  &:last-child {
+    height: -webkit-fill-available;
+  }
+}
+
+.bp3-small-dialog {
```
Simon-Laux: this looks like bp3 overwrite and it isn't scoped so I suggest moving this class to the bp3 overwrites file.

src/renderer/components/message/MessageList.js
```diff
@@ -87,6 +87,7 @@ export default function MessageList ({ chat, refComposer, locationStreamingEnabl
   return (
     <div id='message-list' ref={messageListWrap}>
       <ul>
+        {console.log(chat.messages, chat.id)}
```
Simon-Laux: remove this before merging

scss/dialogs/_delta-dialog.scss
Jikstra: It's not an overwrite, just bad naming.

Issue #1158, opened 2019-11-29
Messages appear in wrong chat (on scrolling up)


Simon-Laux:
On repeated scrolling up and switching the chat the messages get mixed up so that I see messages from different chats in the chatview

adbenitez:
same issue for me, beside the messages appearing in wrong chats, the scrolling simply doesn't work, I get the same messages when scrolling up in an endless loop, also then trying to delete a message or see the message info doesn't work

Issue #1122, opened 2019-11-07
First mesage to new contact is not shown properly and vanishes


duckdalbe:
- Platform (linux/mac/windows): Linux
- Delta Chat Desktop Version: 0.840.0-PREVIEW
- Expected behavior: Show initial message to a new contact in message window
- Actual behavior: An initial message sent to a new contact from a third device is only shown in the list of contacts, but not in the message window. Following messages make the initial message disappear from the list of contacts, while it still doesn't show up in the message window — it is lost from DCD.
- Steps to reproduce the problem: 1. Send an initial message to a new contact with another client (I used deltachat-android 0.950.0, then look at deltachat-desktop (doesn't matter if deltachat-desktop was open already or not): message is not shown in message window. 2. Send a reply or another message from another client: first message vanishes from DCD.
- Screenshots: 
![deltachat-desktop-missing-message](https://user-images.githubusercontent.com/557240/68374069-6a3ce100-0144-11ea-9d5a-11f6832b1346.png)

- log: [2019-10-04-10-04-51.log.txt](https://github.com/deltachat/deltachat-desktop/files/3818730/2019-10-04-10-04-51.log.txt) (stripped by private messages)

nicodh:
Just tried to reproduce this with master and everything worked as expected. Can you please test with next release?

pabzm:
I will as soon as my DC-desktop stops crashing on startup, please wait a little.

Issue #477, opened 2018-12-23
Chat scrolls not to completely to bottom if pictures are present


Simon-Laux:
Probably its when pictures get loaded and grow in height after scrolling to the bottom.

A possible solution could be to use a bigger "skeleton" image placeholder that shrinks to the right size after the image is loaded.

ralphtheninja:
I've also noticed this.

nicodh:
Can't reproduce this. Maybe it's obsolete?

nicodh:
@ralphtheninja @Simon-Laux 
if you also can't reproduce this anymore I will close the issue

ralphtheninja:
I can still reproduce.

PR #1161, opened 2019-11-30
Rewrite MessageList store logic and refactor chat store

### LINKED ISSUE #712: Chat updates (such as delete or new incomming message) result in jumping to the end
- Expected behavior: Scroll position is kept after new messages / deleting an old message
- Actual behavior: chat jumps to the end in these cases
- Steps to reproduce the problem: scroll a bit up and delete one of the earlier messages. 

This problem could be solved in the planned refactoring of the chatview #536

Jikstra:
- Moves fetch/paging logic into frontend and store
- Moves message convertion into backend
- fixes scroll jumping by scrolling to a specific message

Missing:
- [x] day markers
- [x] sending messages
- [x] deleting messages
- [x] update on incoming msgs
- [x] some message methods aren't working properly
- [x] iron all those bugs out
  - [x] `Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at module.exports../src/renderer/stores/MessageList.js.MessageListStore.effects.push (MessageList.js:123)`
  - [x] UI hangs on showing more message details
  - [x] restore default state when deleting currently selected chat
  - [x]  selecting a chat with unread messages needs two clicks
  - [x] sending a message doesn't scroll down to the message
- [x] make sure images have a fixed height

Fixes https://github.com/deltachat/deltachat-desktop/issues/1166
Fixes https://github.com/deltachat/deltachat-desktop/issues/1158
Fixes https://github.com/deltachat/deltachat-desktop/issues/477
Fixes https://github.com/deltachat/deltachat-desktop/issues/1180
Fixes https://github.com/deltachat/deltachat-desktop/issues/1122#issuecomment-561795867
Fixes #712


scss/message/_message.scss
```diff
@@ -142,10 +142,10 @@
     }
 
     .module-message__img-attachment {
```
Simon-Laux: some wide pictures get cut off, maybe make the height smaller in this case if possible.

src/renderer/stores/chat.js
```diff
@@ -15,85 +16,215 @@ const defaultState = {
...
-      return msg.id === messageObj.id ? messageObj : msg
+chatStore.effects.push(async ({ type, payload }, state) => {
+  if (type === 'SELECT_CHAT') {
+    const chatId = payload
+    const chat = await callDcMethodAsync('chatList.selectChat', [chatId])
+    const messageIds = await callDcMethodAsync('messageList.getMessageIds', [chatId])
+    const oldestFetchedMessageIndex = messageIds.length - PAGE_SIZE
+    const newestFetchedMessageIndex = messageIds.length
+    const messageIdsToFetch = messageIds.slice(oldestFetchedMessageIndex, newestFetchedMessageIndex)
+    const messages = await callDcMethodAsync('messageList.getMessages', [messageIdsToFetch])
```
nicodh: Why do we need 3 times to call callDcMethodAsync?
Why is messageIDs not fetched together with chatList.selectChat
Why do we need to calculate messageIdsToFetch here? PAGE_SIZE could be a shared variable, so the calculation could be done also together with chatList.selectChat

scss/message/_message.scss
Jikstra: We can't dynamically set the height of images because this would make the scrollbar jump.

Simon-Laux: I thought the saving image dimensions in the db api was already done, but now I see that its only partly implemented

PR #1174, opened 2019-12-04
add file drag in confirmation dialog 

### LINKED ISSUE #1169: ask the user before a dropped file is actually sent
i just discussed with @Simon-Laux that an advanced attachment-staging-area as on android is planned.

however, as a quick fix for _now_ it would be great if a simple `confirm()` or sth. like that would be added.

this way the user has a chance to avoid an accidentally picked up file being sent by an thoughtless mouse movement :)

Simon-Laux:
Peek 2019-12-04 16-29
fixes #1169

### DIFF
--- a/src/renderer/components/message/MessageListAndComposer.js
+++ b/src/renderer/components/message/MessageListAndComposer.js
@@ -7,6 +7,8 @@ import MessageList from './MessageList'
 
 import SettingsContext from '../../contexts/SettingsContext'
 
+import ScreenContext from '../../contexts/ScreenContext'
+
 import { DC_CHAT_ID_DEADDROP, DC_CHAT_ID_STARRED } from 'deltachat-node/constants'
 
 export default function MessageListAndComposer (props) {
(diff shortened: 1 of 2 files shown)


r10s:
great!

src/renderer/components/message/MessageListAndComposer.js
```diff
@@ -17,22 +19,31 @@ export default function MessageListAndComposer (props) {
...
+      const { path, name } = files[i]
       if (!forbiddenPathRegEx.test(path.replace('\\', '/'))) {
-        callDcMethod(
-          'messageList.sendMessage',
-          [chat.id, null, path]
-        )
+        openDialog('ConfirmationDialog', {
+          message: tx('ask_send_file_desktop', [name, chat.name]),
+          cb: yes => {
+            if (!yes) { return }
```
Jikstra: `{ }` are not needed here
for the sending part we should use the chat store as soon as it's merged. The call would be `chatListDispatch({type:'SEND_MESSAGE', [chat.id, null, path])`

src/renderer/components/message/MessageListAndComposer.js
```diff
@@ -17,22 +19,31 @@ export default function MessageListAndComposer (props) {
...
+            callDcMethod(
+              'messageList.sendMessage',
+              [chat.id, null, path]
+            )
```
Jikstra: for the sending part we should use the chat store as soon as it's merged. The call would be `chatListDispatch({type:'SEND_MESSAGE', [chat.id, null, path])`

PR #1210, opened 2019-12-10
take message-image height from core into account

Simon-Laux:
It still doesn't completely beautifies the image dimensions, but it doesn't hurt and it revives dead code...
relies on https://github.com/deltachat/deltachat-node/pull/397


scss/message/_message.scss
```diff
@@ -145,7 +145,6 @@
         object-fit: cover;
         width: auto;
         max-width: 100%;
-        height: 400px;
```
Simon-Laux: that was always overwritten by the max height `.module-message__img-attachment` so I removed it.

src/renderer/components/message/Attachment.js
```diff
@@ -168,7 +165,7 @@ export default function Attachment (props) {
       >
         <video
           className='module-message__img-attachment'
-          height={Math.min(MAXIMUM_IMG_HEIGHT, height)}
+          style={{ height: `${Math.min(MAXIMUM_IMG_HEIGHT, height)}px` }}
```
Jikstra: `Math.min(MAXIMUM_IMG_HEIGHT, height) + 'px'`

PR #1216, opened 2019-12-10
Account folders

### LINKED ISSUE #1209: Replace strange id by email addr (account folder)
because the id is annoying for finding the right folder when debugging the db.

Simon-Laux:
This is made to be backwards compatible (it creates symlinks, not 100% whether this part works on windows so lets better test it)
fixes #1209

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,6 @@
     "react-qr-svg": "^2.1.0",
     "react-string-replace": "^0.4.4",
     "rimraf": "^2.7.1",
-    "run-parallel": "^1.1.9",
     "simple-markdown": "^0.7.1",
     "styled-components": "^4.4.1",
     "tempy": "^0.3.0",
--- a/package.json
+++ b/package.json
@@ -83,7 +83,6 @@
     "react-string-replace": "^0.4.4",
     "rimraf": "^2.7.1",
     "run-parallel": "^1.1.9",
-    "run-series": "^1.1.8",
     "simple-markdown": "^0.7.1",
     "styled-components": "^4.4.1",
     "tempy": "^0.3.0",
--- a/package.json
+++ b/package.json
@@ -71,7 +71,6 @@
     "lodash.merge": "^4.6.1",
     "mapbox-gl": "^1.6.1",
     "mime-types": "^2.1.25",
-    "mkdirp": "^0.5.1",
     "moize": "^5.4.4",
     "moment": "^2.22.2",
     "rc": "^1.2.8",
--- a/package-lock.json
+++ b/package-lock.json
@@ -10599,11 +10599,6 @@
         "aproba": "^1.1.1"
       }
     },
-    "run-series": {
-      "version": "1.1.8",
-      "resolved": "https://registry.npmjs.org/run-series/-/run-series-1.1.8.tgz",
-      "integrity": "sha512-+GztYEPRpIsQoCSraWHDBs9WVy4eVME16zhOtDB4H9J4xN0XRhknnmLOl+4gRgZtu8dpp9N/utSPjKH/xmDzXg=="
-    },
     "rw": {
       "version": "0.1.4",
       "resolved": "https://registry.npmjs.org/rw/-/rw-0.1.4.tgz",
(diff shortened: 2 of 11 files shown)


Jikstra:
> it creates symlinks, not 100% whether this part works on windows so lets better test it

I would prefer to move/rename the folder. This probably breaks backwards compatibility but I think this is fine in this case. But we should be tactical about when to merge this. What currently comes to my mind is not merging features/things that break the current release and focus on bug fixing bug fixing bug fixing and having this pr as the only breaking thing.

r10s:
first of all, i agree, i would be careful with merging this now.

also - does anyone knows why the choice was made to use a hash over the email address as the foldername? i am not really sure if really all valid email addresses are valid filenames on all supported systems. so, from this point of view, the hash was better.

using symlinks also sounds sensible to me. there are huge differences between windows and unix systems on this kind of stuff, i would avoid that totally if possible. so - why not scan the basedir and check _and_ subdir if this this is a valid account? this way we would be free to name things as we want and can change this in future.

but i think this can be done at some point if we have stable releases out. the user won't mind, it does not really matter how the database and folder is named.

Simon-Laux:
The symbolic link is ONLY for backwards compatibility

r10s:
> Also email addresses are valid filenames because they don't contain the forbidden chars https://stackoverflow.com/questions/1976007/what-characters-are-forbidden-in-windows-and-linux-directory-names#31976060

i would not bet on that: https://en.wikipedia.org/wiki/Email_address#Local-part shows several characters allowed in email addresses that are not allowed in filenames. sure, this may be rare, but esp. in different contexts they may be used.

hpk42:
how about mangled_email (leave out any characters or "-" them that are not allowed on all filesystems) 
such that the dir name is mangled_email--BASE64 where BASE64 is like now? 
it's easy to list all files and determine by BASE64, but we also have a human-readable part. 

In any case, let's postpone this into January or so -- it's now indeed bug fixing and getting
usable releases out the door ... Maybe close this PR with a label "later" and open an issue
where all "later" things are listed/referenced.

adbenitez:
what about just url-encode the email address, or something like that?
example code in python:

```
In [6]: urllib.parse.quote('''!#$%&'*+-/=?^_`{|}~;adb@nauta.cu''', safe='@')
Out[6]: '%21%23%24%25%26%27%2A%2B-%2F%3D%3F%5E_%60%7B%7C%7D~%3Badb@nauta.cu'
```

Simon-Laux:
I like the uri encode idea

Simon-Laux:
I made it more stable and tested it also with IMEX-Backup and new empty directory.
It is backwards compatible on Linux. I didn't test the other platforms yet.
My quick and dirty research says you need to run DeltaChat as Admin on windows, because you need admin rights to create symlinks on windows... maybe we should add an os check that disables the links, that way the backwards compatibility is not given.

Backwards compatibility in the sense that you can still open the accounts that were migrated with an old dc version.

Simon-Laux:
I made this mergable, Now the new behaviour isn't active but it is compatible to the new behaviour, se we can remove some code in a future version to activate the new behaviour and migration.

package.json
```diff
@@ -81,7 +82,9 @@
     "react-qr-svg": "^2.1.0",
     "react-string-replace": "^0.4.4",
     "rimraf": "^2.7.1",
-    "simple-markdown": "^0.7.0",
+    "run-parallel": "^1.1.9",
+    "run-series": "^1.1.8",
```
Simon-Laux: @Jikstra didn't I removed them on purpose?

package.json
```diff
@@ -69,8 +69,9 @@
     "immutability-helper": "^3.0.1",
     "lodash": "^4.17.15",
     "lodash.merge": "^4.6.1",
-    "mapbox-gl": "^1.5.0",
-    "mime-types": "^2.1.21",
+    "mapbox-gl": "^1.6.1",
+    "mime-types": "^2.1.25",
+    "mkdirp": "^0.5.1",
```
Simon-Laux: also `mkdirp` is not used anymore

package.json
Jikstra: Probably i resolved the merge conflicts wrongly

PR #1229, opened 2019-12-12
Fix MessageList not updating/scrolling down on incoming message

### LINKED ISSUE #1222: Receiving message in chat but message not added to the chat area
I opened a chat with another test email, send 3 consecutive messages from the other side, in my delta chat the messages are received and shown in the chat list in the last message preview of the chat, but in the opened chat the message doesn't appear until I start typing something (and the scroll isn't moved to the end) or when changing to other chat and going back to the chat, probably related to my previous issue

Jikstra:
Fixes #1222

### DIFF
--- a/src/renderer/components/message/MessageList.js
+++ b/src/renderer/components/message/MessageList.js
@@ -41,8 +41,10 @@ export default function MessageList ({ chat, refComposer, locationStreamingEnabl
 
   useEffect(() => {
     if (scrollToBottomIfClose === false) return
-    if (messageListRef.current.scrollHeight - messageListRef.current.scrollTop > 400) {
-      // we are close to bottom so scroll down to show the new message
+    // For whatever reason the scrollTop is ~1000 off...
+    const distanceToBottom = messageListRef.current.scrollHeight - (messageListRef.current.scrollTop + 1000)
+    if (distanceToBottom > (window.screen.height / 2)) {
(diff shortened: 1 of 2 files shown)


src/renderer/components/message/MessageList.js
```diff
@@ -41,8 +41,10 @@ export default function MessageList ({ chat, refComposer, locationStreamingEnabl
 
   useEffect(() => {
     if (scrollToBottomIfClose === false) return
-    if (messageListRef.current.scrollHeight - messageListRef.current.scrollTop > 400) {
-      // we are close to bottom so scroll down to show the new message
+    // For whatever reason the scrollTop is ~1000 off...
```
Simon-Laux: _For the protocol: I do not like this "we don't know what its doing lets fix it anyway somehow" approach_

Simon-Laux:
The scrolling is not working (I only tested it with the window in the background) but the update bug is fixed.

src/renderer/stores/chat.js
```diff
@@ -66,8 +66,7 @@ chatStore.reducers.push(({ type, payload, id }, state) => {
       messages: {
         ...state.messages,
         ...payload.messagesIncoming
-      },
-      scrollToBottomIfClose: true
```
nicodh: this line is needed: in my tests it only scrolls down if this is true. Why did you remove it?

Issue #1225, opened 2019-12-11
subtitle in device-chat


r10s:
currently, the subtitle in the device-chat shows the fake-email-address device@localhost

i suggest to replace this, by the string `device_talk_subtitle` (reads sth. as "locally generated messages")

also the text "Messaging disabled (this is the device chat)" that shows instead of the input field look a bit weird to me, at least the explanation in brackets. maybe just hide the input bar and show nothing (this is what was done on android+ios)

Simon-Laux:
> also the text "Messaging disabled (this is the device chat)" that shows instead of the input field look a bit weird to me, at least the explanation in brackets. maybe just hide the input bar and show nothing (this is what was done on android+ios)

This is standard for desktop, as its the same disabled composer when leaving a group, I believe hiding it is more confusing.

r10s:
okay, then maybe reword the text, i did a suggestion at https://github.com/deltachat/deltachat-desktop/pull/1231 - i think for now this would do the job.

PR #1231, opened 2019-12-12
reword the hints when the user cannot write to a chat

r10s:
tackles https://github.com/deltachat/deltachat-desktop/issues/1225

just a suggestion. feel free to change the text or to close this or :)

### DIFF
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -6,7 +6,7 @@
     "message": "You are not in this group and cannot write therefore. To join, ask another member."
   },
   "messaging_disabled_device_chat": {
-    "message": "This chat contains locally generated messages. Therefore you can not write any messages."
+    "message": "This chat contains locally generated messages. Therefore you cannot write any messages."
   },
   "message_not_verified": {
     "message": "The message was sent with non-verified encryption. See \"Info\" for details."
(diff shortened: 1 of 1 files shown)


_locales/_untranslated_en.json
```diff
@@ -3,10 +3,10 @@
     "message": "Copy link"
   },
   "messaging_disabled_not_in_group": {
-    "message": "Messaging disabled (you are not in this group)"
+    "message": "You are not in this group and cannot write therefore. To join, ask another member."
   },
   "messaging_disabled_device_chat": {
-    "message": "Messaging disabled (this is the device chat)"
+    "message": "This chat contains locally generated messages and can be read only."
```
hpk42:
```suggestion
    "message": "This chat contains notification messages generated by the Delta Chat application running on your device."
```

maybe?

r10s: i think the point is to clarify that the user cannot _type_ - the text appears where normally the input field takes place. otoh, it should not look like an error or as if the user has made sth. wrong.

also, there is already the subtitle that says "Locally generated messages" and the device message that explains further.

hpk42: k -- in any case i think the phrase "can be read only" is not proper english i think :)

hpk42:
```suggestion
    "message": "This chat contains locally generated messages. Therefore you can not write any messages."
```

PR #1237, opened 2019-12-13
strip markdown from chatlistitem (fix #1227)

### LINKED ISSUE #1227: remove markdown from chatlist
the current desktop-only device-message uses markdown (tbh, i do not like that as we cannot really use markdown in the device-messages coming from core currently; so we have two types of styles. of messages).

however, this issue is not about markdown as such, but that the markdown is also displayed in the chatlist. i think the chatlist should be clean in this regard, so no bold, italic or whatever. bold eg. is also used for showing an unread chat. having some markup here would make the whole list too noisy. also telegram and others do not show markdown in the chatlist btw.

Simon-Laux:
strip some markdown and display everything in clear text that doesn't match (chatlistitem message preview)
fix #1227

### DIFF
--- a/src/renderer/components/message/MessageBody.js
+++ b/src/renderer/components/message/MessageBody.js
@@ -4,8 +4,12 @@ import { getSizeClass, getRegex, replaceColons } from '../conversations/emoji'
 import { previewRules, rules } from './MessageMarkdown'
 import SimpleMarkdown from 'simple-markdown'
 
+import logger from '../../../logger'
+const log = logger.getLogger('renderer/messageMarkdown')
+
 const emojiRegex = getRegex()
 const previewParser = SimpleMarkdown.parserFor(previewRules, { inline: true })
+const previewAst2react = SimpleMarkdown.outputFor(previewRules, 'react')
 const parser = SimpleMarkdown.parserFor(rules)
 const ast2react = SimpleMarkdown.outputFor(rules, 'react')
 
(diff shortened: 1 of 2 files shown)


Simon-Laux:
> so, for the issue #1227, the simple fix is to remove the bold headline from the message text. done. and safe.

sounds reasonable

> things get difficult when we start figuring things out with strings, converting plain text to html and back. esp. on electron, this was the cause of all kinds of exploits in the past.

react uses text nodes here so its unlikely that something slips through. (its not direct html)

r10s:
> react uses text nodes here so its unlikely that something slips through.

unlikeky. this is what i mean :)

r10s:
so, as #1240 is merged, i suggest to close this pr and postpone all markdown efforts. maybe also disable existing markdown->html conversions which already introduce problems.

Simon-Laux:
> maybe also disable existing markdown->html conversions which already introduce problems.

what are the other problems? besides the _italics_ one that @adbenitez reported?

r10s:
well, i was referring to @adbenitez report and to the chatlist issue. but there will probably more issues popping up if we follow this way. so i would not start fixing these things but just postpone markdown. this gives us much room to did other things - which are also easier then. eg. notifications are missing, so maybe we have to route things to some system ui, also these kind of things are easier to bootstrap if one knows, everything is plain text.

adbenitez:
I am a fan of markdown support, but agree with @r10s to postpone this, at least until Android adds support, and also the issue I described is fixed

Simon-Laux:
So this pr strips the markdown in the chat list items, should we merge or close it? It fixes one of the two issues.

PR #1238, opened 2019-12-13
subtitle in device-chat (fixes #1225)

### LINKED ISSUE #1225: subtitle in device-chat

Simon-Laux:


r10s:
great!

PR #1239, opened 2019-12-13
fix Menu Items (closes #1232)

### LINKED ISSUE #1232: Chat options in menu are enabled when no group is selected
when no group is selected, the chat options like 'delete chat', 'block contact', etc are enabled, causing errors if user clicks them, this option should be disabled, or better: hidden/removed from the menu if there are no chat selected

Simon-Laux:
refactor menu and chatListItem contextmenu for ease of development
hide some options that don't make any sense on selftalk and deviceTalk,
in mainMenu and chatListItem contextmenu

### DIFF
--- a/src/main/deltachat/chatlist.js
+++ b/src/main/deltachat/chatlist.js
@@ -99,8 +99,8 @@ module.exports = class DCChatList extends SplitOut {
       freshMessageCounter: this._dc.getFreshMessageCount(chatId),
       isArchiveLink: chat.id === C.DC_CHAT_ID_ARCHIVED_LINK,
       contactIds,
-      isSelfTalk: chat.isSelfTalk || undefined, // only transfer if needed (json strips undefined)
-      isDeviceTalk: chat.isDeviceTalk || undefined,
+      isSelfTalk: chat.isSelfTalk,
+      isDeviceTalk: chat.isDeviceTalk,
       selfInGroup: isGroup && contactIds.indexOf(C.DC_CONTACT_ID_SELF) !== -1
     }
   }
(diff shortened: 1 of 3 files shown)


src/main/deltachat/chatlist.js
```diff
@@ -99,6 +99,8 @@ module.exports = class DCChatList extends SplitOut {
       freshMessageCounter: this._dc.getFreshMessageCount(chatId),
       isArchiveLink: chat.id === C.DC_CHAT_ID_ARCHIVED_LINK,
       contactIds,
+      isSelfTalk: chat.isSelfTalk || undefined, // only transfer if needed (json strips undefined)
+      isDeviceTalk: chat.isDeviceTalk || undefined,
```
Jikstra: Why strip it out? I don't think that we need to make the json strings smaller for any performance improvements.

nicodh: I would also prefer to always have it false or true and not undefined....

Simon-Laux: wasn't space saving the whole idea of chatlist item instead of fullchat?

nicodh: No the idea was not to fetch to many unused data (causing more database queries). Memory wise we have no issue with json fields I would assume...

Simon-Laux: then we can do that, but I'm still skeptical how fast the electron ipc can handle much json data fast.

Jikstra: Don't be skeptical, measure it :) As long as there's no measurement, I wouldn't strip stuff out.

PR #1245, opened 2019-12-13
Use default image for profile preview

nicodh:


Simon-Laux:
I would prefer the Letter avatar here instead of the random camera symbol.

nicodh:
Go ahead - that would be fine for me too...

Simon-Laux:
done

PR #1249, opened 2019-12-14
Improve confirm dialog

### LINKED ISSUE #1223: order of yes/no buttons
the current order of yes/no buttons eg. in the "delete" dialog is a bit unexpected. showing "cancel" on the right is contrary to what the user has learnt from most uis.

Screen Shot 2019-12-11 at 23 08 49

as a simple fix, i would swap the buttons. it would also be nice to use "cancel" and "delete" as labels (we already have strings for that) instead of "yes" and "no". finally, it is great that the destructive option is shown in red :)

moreover, when touching this, is it probably a good idea to have a look at similar dialogs. EDIT: just noticed that for the "delete account" and "leave group" dialogs, but there are probably more.

Simon-Laux:
closes #1223

### DIFF
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -76,5 +76,5 @@
   },
   "confirm_delete_message": {
     "message": "Delete"
-  },
+  }
 }
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -56,25 +56,7 @@
   "ask_send_file_desktop":{
     "message": "Send %s to %s?"
   },
-  "confirm_leave_group": {
-    "message": "Leave"
-  },
-  "confirm_delete_chat": {
-    "message": "Delete"
-  },
-  "confirm_block_contact": {
-    "message": "Block"
-  },
-  "confirm_unblock_contact": {
-    "message": "Unblock"
-  },
-  "confirm_forget_login": {
-    "message": "Forget"
-  },
-  "confirm_send_file": {
-    "message": "send"
-  },
-  "confirm_delete_message": {
-    "message": "Delete"
+  "remove_account": {
+    "message": "Remove account"
   }
 }
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -55,5 +55,26 @@
   },
   "ask_send_file_desktop":{
     "message": "Send %s to %s?"
-  }
+  },
+  "confirm_leave_group": {
+    "message": "Leave"
+  },
+  "confirm_delete_chat": {
+    "message": "Delete"
+  },
+  "confirm_block_contact": {
+    "message": "Block"
+  },
+  "confirm_unblock_contact": {
+    "message": "Unblock"
+  },
+  "confirm_forget_login": {
+    "message": "Forget"
+  },
+  "confirm_send_file": {
+    "message": "send"
+  },
+  "confirm_delete_message": {
+    "message": "Delete"
+  },
 }
--- a/src/renderer/components/LoginScreen.js
+++ b/src/renderer/components/LoginScreen.js
@@ -193,7 +193,7 @@ export default function LoginScreen (props) {
     const message = tx('forget_login_confirmation_desktop')
     openDialog('ConfirmationDialog', {
       message,
-      confirmLabel: tx('confirm_forget_login'),
+      confirmLabel: tx('remove_account'),
       cb: yes => { if (yes) sendToBackend('forgetLogin', login) }
     })
   }
--- a/src/renderer/components/LoginScreen.js
+++ b/src/renderer/components/LoginScreen.js
@@ -193,6 +193,7 @@ export default function LoginScreen (props) {
     const message = tx('forget_login_confirmation_desktop')
     openDialog('ConfirmationDialog', {
       message,
+      confirmLabel: tx('confirm_forget_login'),
       cb: yes => { if (yes) sendToBackend('forgetLogin', login) }
     })
   }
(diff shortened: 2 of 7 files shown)


_locales/_untranslated_en.json
```diff
@@ -55,5 +55,26 @@
...
+    "message": "Unblock"
+  },
+  "confirm_forget_login": {
+    "message": "Forget"
+  },
+  "confirm_send_file": {
+    "message": "send"
+  },
+  "confirm_delete_message": {
+    "message": "Delete"
```
r10s: please just use the string "delete" here.

_locales/_untranslated_en.json
```diff
@@ -55,5 +55,26 @@
...
+    "message": "Block"
+  },
+  "confirm_unblock_contact": {
+    "message": "Unblock"
+  },
+  "confirm_forget_login": {
+    "message": "Forget"
+  },
+  "confirm_send_file": {
+    "message": "send"
```
r10s: use "menu_send" here

_locales/_untranslated_en.json
```diff
@@ -55,5 +55,26 @@
...
+  "confirm_leave_group": {
+    "message": "Leave"
+  },
+  "confirm_delete_chat": {
+    "message": "Delete"
+  },
+  "confirm_block_contact": {
+    "message": "Block"
+  },
+  "confirm_unblock_contact": {
```
r10s: please just name the string "unblock" so it can be used more widely. we want to have special sorted translations only if differencing is really required - and this is usually not forseeable

_locales/_untranslated_en.json
```diff
@@ -55,5 +55,26 @@
...
   "ask_send_file_desktop":{
     "message": "Send %s to %s?"
+  },
+  "confirm_leave_group": {
+    "message": "Leave"
+  },
+  "confirm_delete_chat": {
+    "message": "Delete"
+  },
+  "confirm_block_contact": {
```
r10s: please use menu_block_contact

_locales/_untranslated_en.json
```diff
@@ -55,5 +55,26 @@
   },
   "ask_send_file_desktop":{
     "message": "Send %s to %s?"
+  },
+  "confirm_leave_group": {
+    "message": "Leave"
+  },
+  "confirm_delete_chat": {
+    "message": "Delete"
```
r10s: just "delete"

_locales/_untranslated_en.json
```diff
@@ -55,5 +55,26 @@
   },
   "ask_send_file_desktop":{
     "message": "Send %s to %s?"
+  },
+  "confirm_leave_group": {
+    "message": "Leave"
```
r10s: use menu_leave_group here

_locales/_untranslated_en.json
```diff
@@ -55,5 +55,26 @@
...
+  "confirm_delete_chat": {
+    "message": "Delete"
+  },
+  "confirm_block_contact": {
+    "message": "Block"
+  },
+  "confirm_unblock_contact": {
+    "message": "Unblock"
+  },
+  "confirm_forget_login": {
```
r10s: please just use "forget" to create this new string. this way, we can use strings more widely; in general, we're create very specific labels only if we really realize there is a need. alternatively, we have delete_account, delete_account_ask which are used on ios. however, we can streamline things later, this should be no blocker.

_locales/_untranslated_en.json
Simon-Laux: the is way a longer string and it will be split in two lines, you still want that?

Simon-Laux:
> please have a look at the existing strings

I did that first, but most of them take two lines so one button has the double height and I assumed that you wouldn't like that.

> i'd remove the gap between the two buttons

Out of scope. Also I don't want to take part in this button discussion anymore so ask @Jikstra or @nicodh about the confirm button style.

Jikstra:
> I did that first, but most of them take two lines so one button has the double height and I assumed that you wouldn't like that.

For which strings is this the case?

> i'd remove the gap between the two buttons

Didn't try it but probably looks weird. We can give it a try, shouldn't be hard to do.

_locales/_untranslated_en.json
r10s: i think, in general, a button with 2 words should work on desktop. i would leave it, we can restyle the things later on. this functionality is not used so frequently, so it is fine this way for now.

r10s:
i would still go for the "Block contact" button, if restyling is out of scope for now, it is not that bad for now, the dialog is not used that often, so we can postpone restyling. but, in general, a button with two words or so should work.

_locales/_untranslated_en.json
Simon-Laux: I can Imagine that the meaning can get lost when just using an English string. Languages are more complicated that key-value, words can have different meanings in different contexts. So I would be careful with universal translating single English word and using it at multiple places.

Simon-Laux:
I actually like more words on the button the thing is rather that the label in the cancel button should be centered then.

Also as alternative to a gap we can do the buttons in two lines, that's not uncommon in material design.

r10s:
> Also as alternative to a gap we could do the buttons in two lines, that's not uncommon in material design.

yip, but on smartphones there is much less space as on desktop - and the default for material design is still to put it in one row. on desktop, having alerts with buttons below each other is uncommon and unneccessary. i would not do that.

Simon-Laux:
so fixed the styling, now it fits in a row

Simon-Laux:
>  i would not do that.

me neither, just wanted to point out the possibility.:smirk:

PR #1262, opened 2019-12-17
Don't fetch chat twice

### LINKED ISSUE #1260: Bug: chatstore is not correctly cleared when account is switched?
somehow it tries to open a chat with the id of the chat that was open in the last account.

nicodh:
resolves #1260

### DIFF
--- a/src/renderer/components/message/MessageList.js
+++ b/src/renderer/components/message/MessageList.js
@@ -47,7 +47,6 @@ export default function MessageList ({ chat, refComposer, locationStreamingEnabl
   }, [scrollToLastPage, scrollHeight])
 
   useEffect(() => {
-    chatStoreDispatch({ type: 'SELECT_CHAT', payload: chat.id })
     isFetching.current = false
   }, [chat.id])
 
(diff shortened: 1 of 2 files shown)


nicodh:
Interesting - so we have completely different behaviour.
I will investigate further...

src/renderer/components/message/MessageList.js
```diff
@@ -47,7 +47,6 @@ export default function MessageList ({ chat, refComposer, locationStreamingEnabl
   }, [scrollToLastPage, scrollHeight])
 
   useEffect(() => {
-    chatStoreDispatch({ type: 'SELECT_CHAT', payload: chat.id })
     isFetching.current = false
```
Jikstra: Do we need the isFetching.current code?

Jikstra:
@Simon-Laux can't reproduce your issue, works as expected for me. Besides this, selecting a chat is now also fast again :)

nicodh:
I think my second issue fixed the bug mentioned in this ticket...

Simon-Laux:
:+1: can confirm `master` works now nicely again

PR #1267, opened 2019-12-21
Use empty placeholder to avoid rendering overhead

nicodh:
I tested with an acoount with ~600 chats.
Since the chatlist renders a skeleton for each item upfront there is quite some overhead with 600 chat items.. 

I tested using just an empty div with fix height as skeleton and while one might say that it not looks as nice as before it really improves performance.
See 3 profiles attached (can be loaded in Chrome dev tools)
The difference is probably only noticeable with many chats but there it is evident...

The breath animation seems also expensive if you look at the profiles...

[Profiles.zip](https://github.com/deltachat/deltachat-desktop/files/3991583/Profiles.zip)

### DIFF
--- a/scss/chat/_chat-list-item.scss
+++ b/scss/chat/_chat-list-item.scss
@@ -17,6 +17,13 @@ $color-chat-list-item-avatar-text: var(--avatarLabelColor);
   &:hover {
     background-color:#bbb;
   }
+  &.skeleton {
+    background-repeat: no-repeat;
+    background-image: radial-gradient(circle 24px, grey 99%, transparent 0),linear-gradient(grey 40px, transparent 0), linear-gradient(grey 40px, transparent 0), linear-gradient(grey 40px, transparent 0);
+    background-size: 56px 56px, 200px 18px, 300px 16px, 40px 16px;
+    background-position: 6px, 67px 12px, 67px 35px, 96% 12px;
+    opacity: 0.1;
+  }
 }
 
 .chat-list-item--has-unread {
@@ -206,3 +213,4 @@ $color-chat-list-item-avatar-text: var(--avatarLabelColor);
     background-color: #bbb;
   } 
 }
+
--- a/scss/chat/_chat-list.scss
+++ b/scss/chat/_chat-list.scss
@@ -10,37 +10,5 @@
 }
 
 .chatlist-item-skeleton {
-  opacity: 0.1;
-
-  & .bone {
-    background: grey;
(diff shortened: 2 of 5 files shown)


Simon-Laux:
I don't believe that that is where the performance loss is.
I'll try to reproduce and look at the profiles soon.
maybe there is some css that allows us to only render empty placeholders when they are outside the view?

Simon-Laux:
Sure we can make the placeholders simpler to save on render time, maybe even use an image?

Simon-Laux:
> The breath animation seems also expensive if you look at the profiles...

I didn't see that, only that the animations are active for a long time. But I probably just don't know how to interpret the profiler.

Jikstra:
The animations shouldn't happen if the chat is not in view. This should be easy to implement as the useLazyChatList stuff implements this already,

nicodh:
> But a real virtual list should be the way to go no matter what
I don't get what you mean by that? The chat list is already a virtual list...
Do you want to get rid of the placeholder? The animations were only applied to the placeholder which should not be visible in most cases.

What is your opinion about the empty skeletons? Do you agree to leave them empty? Would you prefer some background image to make them look like before?
I think the better performance justifies to have empty items instead of more realistic looking placeholders

Jikstra:
> I don't get what you mean by that? The chat list is already a virtual list...

It's not. A virtual list would only keep the items rendered. Currently we have a lazy loading list. Once you scrolled down all chat items are in the dom. I think a hotfix to render them empty is fine, but looks shit. I would try to figure out a way to only render/animate the placeholder ones that are in view.

nicodh:
ok. So maybe merge this PR in yours and then add some stuff to make the skeletons look nicer? I will test performance then with my 600 chat account..

Simon-Laux:
> I think a hotfix to render them empty is fine, but looks shit

exactly it looks bad. You can throw out the animation if that eats too much performance, it not important at all.
Using an image might help because it would only need to be rendered once and just repeated, I'll guess.
At the moment the placeholder has like 4-8 sub elements, we can simplify the placeholder by using less elements, maybe web components could help us here? @nicodh

Simon-Laux:
we could use https://css-tricks.com/building-skeleton-screens-css-custom-properties/ to build a skeleton that doesn't need extra elements.

nicodh:
For me the empty item would be totally fine - it's only visible for fractions of a second anyway so I wouldn't put too much effort in it. But these css skeletons seem a good solution.

PR #1271, opened 2019-12-22
Typescript support

nicodh:
Kickoff to migrate to typescript step by step
TODO (incomplete):
- [x] create useful type definitions
- [x] create useful tslint config
- [ ] check if deployment routine needs changes

fell free to add more points here...


Simon-Laux:
when you're already tinkering with webpack, you could try if https://parceljs.org/ can give us a performance boost or not.

nicodh:
I would like to merge this soon since it would result in merge conflicts with most of future changes.
It has a dependency to the master of deltachat-nodes with index.d.ts declaration

Simon-Laux:
Is it done in you opinion? (this still has the WIP Tag)

src/types.d.ts
```diff
@@ -0,0 +1,148 @@
...
+  ipcReady: boolean;
+  localeData?: any
+  state?: any;
+  saveState?: (arg: AppState) => void;
+  translate?: (key: string, substitutions: Array<string>, opts: any) => string;
+}
+
+export interface Credentials {
+  addr: string;
+  passwd?: string;
```
Simon-Laux: this should be here should it?

nicodh: Yes it should, since sometimes the credentials are passed without password like in deltachat state, so the password is optional.

Simon-Laux: and where is it used?

Simon-Laux:
Nonetheless I would put this in project resurrection as its to big a change when we have like 5 open prs next to it open

pabzm:
To be honest, I would also prefer to postpone this change. I see that a lot of work has been put into it, but without meaningful tests I'm afraid that it will break things, which would be unfortunate in the current situation where we head towards v1.0.

Simon-Laux:
Nah it will not break much thing, but it is a big task to convert it all and we can only do it when no other pr are currently open. We should meet a week and convert it all to ts together.

nicodh:
I don't think it will break much but if there are no resources now to review I'm fine with postponing. But I would really avoid to "convert it all to ts" in one big change. That would only result in another huge PR no one can really review. 
I added the first step as single PR which has very few changes: https://github.com/deltachat/deltachat-desktop/pull/1333/files
So it is very unlikely that it will "break things"...

Simon-Laux:
I'll close this for now

PR #1274, opened 2019-12-24
Pimp about delta chat dialog, small steps towards moving away from direct ipc

Simon-Laux:
See commit messages for more info, best reviewed commit by commit.
Also styling the about dialog a bit:

### DIFF
--- a/scss/dialogs/_about.scss
+++ b/scss/dialogs/_about.scss
@@ -1,7 +1,23 @@
-.dialog-about__dc-info {
-    width: 100%;
-    background-color: #ececec;
-    border-color: white;
-    border-color: #ececec;
+.dialog-about__dc-details {
+    width: 460px;
+    height: 460px;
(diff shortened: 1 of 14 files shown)


Simon-Laux:
I also tried to convert the settings:
``` 
 /** Saves settings to deltachat core */
  handleDeltaSettingsChange (key, value) {
    callDcMethodAsync('settings.setConfig', [key, value])
      .then(success => {
        if (!success) { return }
        const settings = this.state.settings
        settings[key] = String(value)
        this.setState({ settings })
      })
  }
```
But unfortunately that results in lag and an unresponsive experience (especially in the text fields)

Jikstra:
> But unfortunately that results in lag and an unresponsive experience (especially in the text fields)

This would write the new value into the db on each key press? That's of course laggy, especially if you don't asyncify the backend (the frontend process is blocked until the event handler in the backend returns)

Issue #1121, opened 2019-11-06
Prefill advanced settings with unprefixed, not "configured_" values


link2xt:
See https://github.com/deltachat/deltachat-android/issues/1106#issuecomment-550467772

Maybe also show `configured_` value as a placeholder, because it may be useful to know which port is currently used.

r10s:
+1 for placeholder.

if i find some time i'd like to refactor android and ios to using placeholders "Automatic" and "See above" (for the SMTP password).

once configured, i would show "Automatic (imap.server.bla)" and so on.

PR #1281, opened 2019-12-29
Only show advanced settings the user entered

nicodh:
relates #1121

First step of reworking the advanced settings. Taking https://github.com/deltachat/deltachat-android/issues/1106#issuecomment-550467772 into account.

In current master the configured: prefixed values are displayed instead of the manually entered values.


nicodh:
Although there is no distinction between placeholder and label in material design I made a difference to show the user what kind of default values are used.

nicodh:
new-advanced-settings

nicodh:
Very strange: CI fails because of a console.log statement in a non existing file: src/renderer/components/dialog/ViewProfile.js ???

link2xt:
> Very strange: CI fails because of a console.log statement in a non existing file: src/renderer/components/dialog/ViewProfile.js ???

But the file does exist: https://github.com/deltachat/deltachat-desktop/blob/master/src/renderer/components/dialogs/ViewProfile.js

Maybe you have deleted it locally?

nicodh:
Yes it exists in master, but it didn't  in this branch before the last merge with master

Issue #1282, opened 2019-12-29
Clicking 1:1(private) chat title does nothing


adbenitez:
now that a nice profile view was added, when user clicks the chat title of a private chat or the avatar of a contact in a group, the contact's profile view should be shown.

- Delta Chat Desktop Version: master

PR #1285, opened 2019-12-29
Fix view profile

Jikstra:
Fixes https://github.com/deltachat/deltachat-desktop/issues/1282

### DIFF
--- a/src/renderer/components/helpers/ChatMethods.js
+++ b/src/renderer/components/helpers/ChatMethods.js
@@ -50,8 +50,9 @@ export function openMapDialog (screenContext, selectedChat) {
   screenContext.openDialog('MapDialog', { selectedChat })
 }
 
-export async function openViewProfileDialog (chatid, screenContext) {
-  const fullChat = await callDcMethodAsync('chatList.getFullChatById', chatid)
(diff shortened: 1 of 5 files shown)


src/renderer/components/helpers/ChatMethods.js
```diff
@@ -50,8 +50,9 @@ export function openMapDialog (screenContext, selectedChat) {
   screenContext.openDialog('MapDialog', { selectedChat })
 }
 
-export async function openViewProfileDialog (chatid, screenContext) {
-  const fullChat = await callDcMethodAsync('chatList.getFullChatById', chatid)
-  const contact = fullChat.contacts[0]
+export async function openViewProfileDialog (screenContext, contact) {
+  if(Number.isInteger(contact)) {
+    contact = await callDcMethodAsync('chatList.getContact', contact)
+  }
```
Simon-Laux: I would just always re-fetch the contact in this case.

Jikstra: I dont think we need to do this. We need to update the selectedChat anyways if a contact changes, and if we already have it why refetch it?

PR #1292, opened 2020-01-01
Use core configure routine instead of tmp instance

### LINKED ISSUE #1270: Changing Password does not work
currently trying to get a gmx account going, and after first setup i had to change the password. I already had setup desktop and it basically worked.  Then i went to "password and account" and changed my password and hit "update".  the login failed ... and looking at the password (hitting the eye) sign i saw that the password had revert4ed to the old value.  Something is borked -- password changes do not seem to actually happen.

nicodh:
resolves #1270

Not yet ready: I think we should give better feedback what is going wrong in case login fails. If imap was not successful, or smtp and not just a generic fail or success


nicodh:
Core configure() saves only working values for credentials - but if IMAP credentials are correct and SMTP is not it won't save the correct IMAP configuration.
So let's say I have a new imap password. When I open my account I get IMAP connect errors. Now if I open my settings and update configuration with a correct IMAP password and a wrong SMTP password you will get IMAP connect errors since the core tries to connect with the old IMAP password (configured_raw_mail_pw).
I want to show the user if the IMAP connect and/or the SMTP succeeded, so he won't think that the IMAP credentials are wrong if they in fact aren't.
@r10s is that  intended behaviour? Has this been discussed before?

nicodh:
So with current master this works fine as expected. We should improve how to show the success/failure messages:
IMAP success, SMTP failed looks like this:
SMTP-failed-IMAP-successfull

Both successfull:
IMAP-Success-SMTP-Success

nicodh:
What we can't show: if SMTP is fine but IMAP isn't, we don't get the SMTP connect success event since core does not send it after IMAP failure.

Simon-Laux:
Don't forget to file the issue on core.

nicodh:
So now we have this status. Assuming the use case that the mail password changed:
1. when opening DC desktop you will see error messages (can not connect etc.)
2. if you change the password to correct settings and click update you see the update status bar running till you get a connection success message (but before the success you still get error messages which is a bit confusing)

So far so good. The problem is: if you enter a wrong password you also get an error message but the configure process "hangs". Last progress update event has 650.
I have no clue how to stop the configuring process with Core API. If I just try to call configure again I get an error "there is already an ongoing process..."

Simon-Laux:
related core buǵ - for reference: https://github.com/deltachat/deltachat-core-rust/issues/1159

nicodh:
For testing: Create an auto configured account - advanced settings should be empty which is not the case in current master. Updating password should work for both: auto configured an manually created accounts

nicodh:
Related: https://github.com/deltachat/deltachat-core-rust/issues/1290

PR #1300, opened 2020-01-05
🚧 Fix integration test for loading config

### LINKED ISSUE #1299: Integration test is broken
It seems since 51112a7 the integration test loading settings from existing config file is broken
@Jikstra @Simon-Laux can you please have a look?

Simon-Laux:
closes #1299

 - [X]  don't crash when last account isn't found
 - [ ] add debug log of found accounts to logins.js
 - [ ] logic to create account folder and account database (aka. fix the test)

Still needs logic to create account folder and account database, because the logic changed.
@nicodh do you wanna complete this integration test fix? I guess we could use dc node to configure the account folder.

### DIFF
--- a/test/integration/setup.js
+++ b/test/integration/setup.js
@@ -37,6 +37,7 @@ function createApp () {
 function createAppWithConfig (overrideConfig) {
   const TEST_DIR = tempy.directory()
   const defaultConfig = getConfig()
+  // TODO create account folder and acoount database
   if (overrideConfig) {
     const extendedConfig = Object.assign({}, defaultConfig, overrideConfig)
     fs.writeFileSync(path.join(TEST_DIR, 'config.json'), JSON.stringify(extendedConfig))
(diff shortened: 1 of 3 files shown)


test/integration/setup.js
```diff
@@ -37,6 +37,7 @@ function createApp () {
 function createAppWithConfig (overrideConfig) {
   const TEST_DIR = tempy.directory()
   const defaultConfig = getConfig()
+  // TODO create account folder and acoount database
```
nicodh: Does this means the behaviour changed: account dir and database were created if they did not exist and now they aren't any more?

Simon-Laux: yes

nicodh:
This is obsolete if we merge #1336

Simon-Laux:
> This is obsolete if we merge #1336

Then we still need need to merge the changes from comits:
- remove mail pw from config 
- don't crash/panic on account not found
- add the list of found accounts to the error

so its not completely obsolete.

PR #1313, opened 2020-01-07
Reintroduce msg-body div and fix message overflow

Jikstra:
This fixes attachments not being clickable


Simon-Laux:
new clickable area

PR #1322, opened 2020-01-08
Include FAQ page as in-app help

pabzm:
Uses the faq-page of the website as content. The HTML has been changed
to only contain the div#content without the h1, and the path of
delta-what.png has been fixed to "./"

- ~~[ ] preprocess the HTML when building the app, not when serving the page.~~ (later)
- [x] write tests for the preprocessing
- [x] figure out how to serve localized content.

### DIFF
--- a/src/renderer/components/dialogs/HelpPage.js
+++ b/src/renderer/components/dialogs/HelpPage.js
@@ -1,0 +1,70 @@
+import React from 'react'
+import reactStringReplace from 'react-string-replace'
+import parse from 'html-react-parser'
+import domToReact from 'html-react-parser/lib/dom-to-react'
+import DeltaDialog, { DeltaDialogBody, DeltaDialogFooter } from './DeltaDialog'
+import { Card } from '@blueprintjs/core'
+import ClickableLink from '../helpers/ClickableLink'
+import { join } from 'path'
+import fs from 'fs'
+
+const contentFilePath = join(__dirname, `../../../../static/help/__locale__/help.html`)
+
+function helpPageReadContentFile () {
+  const locale = window.localeData.locale
+  let contentFile = contentFilePath.replace('__locale__', locale)
+  if (!fs.existsSync(contentFile)) {
+    if (locale !== 'en') {
+      contentFile = contentFilePath.replace('__locale__', 'en')
+    } else {
+      throw new Error(`Cannot read HTML content, file not found: '${contentFile}'`)
+    }
+  }
+  return fs.readFileSync(contentFile, 'utf-8')
+}
+
+export function HelpPageContent ({ props }) {
+  // Make links open externaly (without this, nothing happens if one clicks on a link).
+  const parserOptions = { replace: ({ attribs, children }) => {
+    if (attribs && attribs.href && attribs.href.startsWith('http')) {
+      return (
+        <ClickableLink href={attribs.href}>{domToReact(children, parserOptions)}</ClickableLink>
+      )
+    }
+  } }
+  return parse(helpPageReadContentFile(), parserOptions)
+}
+
+function HelpPageLocalCopyHint ({ props }) {
+  const locale = window.localeData.locale
+  const localizedHref = `https://delta.chat/${locale}/help`
+  const text = window.translate('help_page_local_copy_hint')
+  const linkedText = reactStringReplace(text, /\[LINK:([^\]]+)\]/, (match, index) => (
+    <ClickableLink href={localizedHref} key={index}>{match}</ClickableLink>
+  ))
+
+  return (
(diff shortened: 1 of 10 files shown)


pabzm:
Could someone give me a hint why the links in the dialog don't trigger opening the url in a browser? I'm a little clueless about that.

pabzm:
Related to https://github.com/deltachat/deltachat-android/pull/1207

Simon-Laux:
security, you need too loop through them and convert them like here:
https://github.com/deltachat/deltachat-desktop/blob/master/src/renderer/components/message/MessageMarkdown.js#L55-L59

Jikstra:
Can we have a script which converts the faq from the deltachat-pages repo?

r10s:
> Can we have a script which converts the faq from the deltachat-pages repo?

yip, the plan is to put that to deltachat-pages/tools/ - the script writes to ../deltachat-android/ ../deltachat-desktop and ../deltachat-ios then.

Simon-Laux:
_when we already do a script the proper faq format in own repo is not far..._

hpk42:
On Wed, Jan 08, 2020 at 21:23 -0800, Simon Laux wrote:
> _when we already do a script the proper faq format in own repo is not far..._

but this requires new workflows for translations -- IMO for january it's good 
to just keep pages working including established translations flows,
and extract the info into the other repos for inclusion in-app.

r10s:
i've create a script that strips unneeded things from the faq at https://github.com/deltachat/deltachat-pages/pull/245 - the script is called from deltachat-pages and assumes deltachat-android and deltachat-desktop located beside deltachat-pages  - i would rename static/faq/faq.html to static/faq/html.html and then we can just add it to the python script.

pabzm:
I'm reassigning myself until my part is finished

pabzm:
Done, please review.

src/renderer/components/dialogs/HelpPage.js
```diff
@@ -0,0 +1,70 @@
...
+    <p style={{ textAlign: 'center', fontSize: 'smaller', borderTop: 'solid thin lightgrey', paddingTop: '0.3em' }}>
+      {linkedText}
+    </p>
+  )
+}
+
+export default function HelpPage (props) {
+  return (
+    <DeltaDialog
+      title={window.translate('help_page_title')}
```
Jikstra: The menu item in german says 'Hilfe' but the dialog title is 'Frequently asked Questions'. I would like to have it consistent, but probably this is a translation fault?

Jikstra:
What is this `.keep-dir` file?

src/renderer/components/dialogs/HelpPage.js
r10s: yes, i would also call it "Help" in the app everywhere.

pabzm:
`.keep-dir` is a "hack" to make git not ignore empty directories. git tracks content, thus files, not file system entries like directories. In order to be able to run the script from deltachat-pages that produces the HTML-files we must have the directories in the repo. As long as we have a HTML-file in the repo, this "hack" isn't actually necessary, but I figured that it doesn't hurt either.

Simon-Laux:
I saw `.gitkeep` already I few times, maybe that's a more common/standard name for it?

pabzm:
To git the name of the file doesn't matter. If you prefer your suggestion feel free to rename it.

Simon-Laux:
So just reviewed it with holger:
- Performance wise we don"t need to do anything its fine as is.
- The indentation of the links at the beginning is missing (probably some css).
- the FAQ should also be reachable over the window menubar "Help" menupoint (there where also the about button is).

Something worth considering but not urgent enough to do now:
- Open it in a separate electron window so you can read the FAQ while using the app - that would be super useful when following instructions on settings from the FAQ.
- the html tag causes an react error in the dev console (idea - maybe we could just an iframe that has a small script tag that converts the links for us?, that way the help.css would make more sense again)

pabzm:
Parsing the HTML when building the app, instead of when the "help"-Link was clicked, would win two things:
1. Performance
2. The guarantee that a packages app contains parseable HTML.

Even if 1. is neglectable, 2. is still a valid point I think. But we can also add this check as a test.

pabzm:
A separate window would be great! I just didn't know how to do it and wanted a relevantly useful feature first.

The react error is a warning, IIRC. I can live with that, if the alternative is ugly or expensive:
I tried shortly to use an iFrame, but then dropped it to find a easier solution. If the iFrame can't access it's resource via an URI we would need to write content into it using ugly hacks (stringify content, assign that as html of the iframe's document; I did it in a different project and still am ashamed ;).

And using a separate window would maybe solve the second issue, too, anyway?

PR #1324, opened 2020-01-08
Fix qr images overflowing on small screens/windows sizes

Jikstra:


r10s:
great! i also noticed that at some point, forgot to file an issue :)

PR #1333, opened 2020-01-10
Typescript step1

nicodh:
First split of the typescript PR for more transparency


src/renderer/types.d.ts
```diff
@@ -0,0 +1,38 @@
+export interface Credentials {
+  addr: string;
+  passwd?: string;
```
Simon-Laux: nope: please remove this line to avoid confusion

PR #1336, opened 2020-01-11
Switch to mocha testrunner

nicodh:
I tried mocha as testrunner but it's not really improving the test behaviour. MacOS tests fail while Linux doesn't and I have no clue why. So I removed OSX test until we have a better test environment.

At least these test run shorter now and still have value since the test real scenarios.
I added 2 more accounts in the CI environment, so we can start testing communication also.

For now I just added 2 own mail accounts, but we should switch to testrun.org accounts...


nicodh:
Still the same random success/failure behaviour. :-(
So the problem is obviously Spectron/Selenium in CI not producing reliable results...

Jikstra:
A lot better output, nice!

test/integration/fixtures/config.js
```diff
@@ -1,30 +1,48 @@
...
+  if (typeof process.env.DC_ADDR !== 'string') {
+    console.error('Missing DC_ADDR environment variable!')
+    process.exit(1)
+  }
+  if (typeof process.env.DC_MAIL_PW !== 'string') {
+    console.error('Missing DC_MAIL_PW environment variable!')
+    process.exit(1)
+  }
+  if (typeof process.env.DC_MAIL_PW1 !== 'string') {
+    console.error('Missing DC_MAIL_PW1 environment variable!')
```
Jikstra: Why do we need PW, PW1 and PW2 variables?

src/application-constants.js
```diff
@@ -40,8 +40,8 @@ function windowDefaults () {
   const messageHeight = 100
   return {
     bounds: {
-      width: 500,
-      height: headerHeight + messageHeight * 6
+      width: 1100,
+      height: headerHeight + messageHeight * 8
```
Jikstra: What's about this change?

test/integration/fixtures/config.js
nicodh: I would like to have tests including starting a one-to-one chat and a group chat, so we would need 3 accounts.
I will provide a sample how to configure that for local tests with a .DCrc file including the needed configuration as json.

src/application-constants.js
nicodh: If we set the default width to 500px the titles of chats might get truncated which is bad for testing. I also see no reason why we set the default width to 500, that's even less than a mobile device would have...

test/integration/fixtures/config.js
Jikstra: But you didn't introduce this test yet in this pr? I would prefer to have this pr only focus on switching to mocha if possible, even if i like testing group chats quite a lot! But for this I think it would be a better approach to implement the usage of the testrun.org burner account creation script. This way you only need a api token and everything else is getting created on the fly.

test/integration/fixtures/config.js
```diff
@@ -1,30 +1,48 @@
...
+    process.exit(1)
+  }
+  if (typeof process.env.DC_MAIL_PW2 !== 'string') {
+    console.error('Missing DC_MAIL_PW2 environment variable!')
+    process.exit(1)
+  }
+  conf.account1 = {
+    credentials: {
```
Simon-Laux: Can you move these values into an extra test scope, because they aren't being used by the app (except for addr - that var is used to determine the last active account)
(aka please don't use config.credentials scope to avoid confusion)

test/integration/fixtures/config.js
```diff
@@ -1,30 +1,48 @@
...
+    process.exit(1)
+  }
+  if (typeof process.env.DC_MAIL_PW1 !== 'string') {
+    console.error('Missing DC_MAIL_PW1 environment variable!')
+    process.exit(1)
+  }
+  if (typeof process.env.DC_MAIL_PW2 !== 'string') {
+    console.error('Missing DC_MAIL_PW2 environment variable!')
+    process.exit(1)
+  }
```
Simon-Laux: This section should test the config argument from rc instead

test/integration/fixtures/config.js
```diff
@@ -1,30 +1,48 @@
...
```
Simon-Laux: This should check the config object from rc instead

test/integration/fixtures/config.js
Simon-Laux: oh outdated... I commented the new section bellow

pabzm:
(Note: Closed in favour of #1355)

PR #1337, opened 2020-01-13
indroduce Typescript

Simon-Laux:
Improved version of #1333, even if it is a bit bigger it works better now.
It doesn't include #1271 though so we need to look at resurrecting that pr after merging this one.

Also please review commit by commit - its easier

_Note: Credit for pushing this feature forward and doing the start goes to @nicodh_


src/renderer/components/map/MapComponent.js
```diff
@@ -9,10 +9,9 @@ const geojsonExtent = require('@mapbox/geojson-extent')
 const moment = require('moment/moment')
 const formatRelativeTime = require('../conversations/formatRelativeTime')
 const MapLayerFactory = require('./MapLayerFactory')
-const { Slider, Button, Collapse } = require('@blueprintjs/core/lib/esm/index')
```
nicodh: Why did you remove this line?

nicodh: I readded it assuming it was by error removed

nicodh:
I reviewed this commit by commit. Looks good to me and can be merged if it is rebased!

src/renderer/components/map/MapComponent.js
Simon-Laux: I believe I changed it to the commonjs counterpart in the parcel pr and it slipped in that way whiles moving stuff back into the ts pr.
Good catch. That"s what reviewing is for :+1:

PR #1338, opened 2020-01-13
replace webpack with Parcel

Simon-Laux:
Because webpack is annoying IMO
We can re-setup babel to work with parcel, but ATM we don't need it anyway, because our only target is electron-chromium (everything besides the basic babel stuff that parcel does anyway)


Simon-Laux:
I'll add this to project resurrection as we won't merge it before the next release.

r10s:
and meanwhile parcel is replaced by esbuild ;) https://github.com/deltachat/deltachat-desktop/pull/2032

PR #1345, opened 2020-01-13
[Performance] Store log_debug in constant variable instead of getting it on every l…

Jikstra:
…og.debug call over ipc


Simon-Laux:
Nice catch

PR #1348, opened 2020-01-13
Refactor Timestamp

Jikstra:


pabzm:
Please describe the background of your refactoring, and why you think it matters.

Jikstra:
It's less lines, uses hooks instead of components, is easier to read and I now know how Timestamps work.

Jikstra:
It happend during an investigation what and how timestamps work. Basically every timestamp Component is running code every minute to recalculate the 'relative time' and updates the state if it changes.

src/renderer/components/conversations/Timestamp.js
```diff
@@ -1,57 +1,40 @@
...
+  const moduleName = module || ''
+
+  if (timestamp === null || timestamp === undefined) return null
+
+  const calculateRelativeTime = () => formatRelativeTime(timestamp, { extended })
+  const [relativeTime, setRelativeTime] = useState(() => calculateRelativeTime())
+
+  // Update relative time every UPDATE_FREQUENCY ms
+  const recalculateRelativeTime = () => setRelativeTime(calculateRelativeTime())
+  useInterval(recalculateRelativeTime, UPDATE_FREQUENCY)
```
Simon-Laux: where do you clear the interval again?

Jikstra: The useInterval hook is taking care of this. See line 13.

pabzm:
So with your changes less code is running every minute?

src/renderer/components/conversations/Timestamp.js
Simon-Laux: ah now I understand how it works, please add a comment that it is cleared automatically by the useHook to prevent confusion.

Issue #1187, opened 2019-12-06
Add "send copy to self" option


adbenitez:
desktop is missing the new option "send copy to self" switch in the "IMAP folder handing" category in settings, should be easy to add, config key is "bcc_self"

PR #1350, opened 2020-01-14
Implement bcc self setting

Jikstra:
Anmerkung 2020-01-14 045357

Fixes https://github.com/deltachat/deltachat-desktop/issues/1187

### DIFF
--- a/src/renderer/components/dialogs/Settings.js
+++ b/src/renderer/components/dialogs/Settings.js
@@ -72,7 +72,8 @@ export default class Settings extends React.Component {
       'displayname',
       'selfstatus',
       'mdns_enabled',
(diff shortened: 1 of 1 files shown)


pabzm:
Nice, thanks!

I would suggest to move the option somewhere else, though. All other switches above and below deal with IMAP-folders, this one doesn't.

r10s:
> I would suggest to move the option somewhere else, though. All other switches above and below deal with IMAP-folders, this one doesn't.

while one can discuss if "Send copy to self" belongs to "IMAP folder handling" or not, desktop does just the same as android and ios here, so it is at least internally consistent.

so, i would not move the option, i think it is okay for now.

PR #1354, opened 2020-01-15
Fix some integration tests after removing welcome text.

pabzm:
7b115c609b removed text that these tests tried to find.


pabzm:
The failing `test-standard` has been fixed in #1322, thus I'm not fixing it again here.

PR #1359, opened 2020-01-16
In app faq refinements

Simon-Laux:
- add androids help page styling to inApp FAQ
- add FAQ to help menu

### DIFF
--- a/scss/help/_help.scss
+++ b/scss/help/_help.scss
@@ -1,0 +1,40 @@
+.help-content {
+    a {
+        text-decoration: none;
+    }
+
+    h2,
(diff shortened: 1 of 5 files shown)


r10s:
isn't "frequently asked questions" more common? but anyway, i would call the thing just "help". even if the menu title is "help" and the first item is "help" as well. this is a common pattern and known to the user. also we can adapt the content more freely then.

pabzm:
But the actual point are the answers to the questions, not that they have been asked a lot!?

But, ok, if you prefer the more common wording, feel free to change it.

PR #1363, opened 2020-01-19
🚧 Testcafe

nicodh:
At the moment testcafe hangs if the bundle is included in test page, although the appended client script logs a message if page is ready and has access to the DOM as expected...
You can run the tests with:
npm i
npm run build
npm run test-cafe


nicodh:
Posted an issue here: https://github.com/DevExpress/testcafe-browser-provider-electron/issues/58

static/test.html
```diff
@@ -2,17 +2,6 @@
...
-    content="default-src 'unsafe-inline';
-             style-src 'self' 'unsafe-inline';
-             font-src 'unsafe-inline';
-             script-src 'self' 'unsafe-inline';
-             worker-src blob: ;
-             child-src blob: ;
-             img-src 'unsafe-inline' data: blob: ;
-             media-src 'unsafe-inline';
-             connect-src https://*.tiles.mapbox.com https://api.mapbox.com https://events.mapbox.com 'unsafe-inline'"
-    >
```
Simon-Laux: Is that security meta tag really unnecessary?

pabzm: No, it's not, let's not loose that. I tried using an iframe to integrate the FAQ at first and this tag blocked me from doing it. At first I was annoyed, but then I agreed.

nicodh: That page is only used for integration tests. Testcafe injects Javascripts which couldn't be loaded in first tries because of Content Security Policy. So I tried to set all values to "unsaf-inline" but realized later that it's sufficient to set the Content Security Policy in the headers: https://github.com/deltachat/deltachat-desktop/pull/1363/files#diff-7b1ed8e005428717d211bea7f7e5c1feR166
The content security for usage in production mode is not affected at all...

pabzm: Sorry, I didn't look at the filename properly... You're right.

PR #1367, opened 2020-01-21
add windows icons

### LINKED ISSUE #1212: Create and add icons for windows
This is a requirement for uploading to the Microsoft store, MS Visual studio (not code, the normal one) has a generator for it, but you'll need to have a windows os + login with a ms account after 30days.
- https://www.electron.build/configuration/appx#appx-assets
- https://docs.microsoft.com/en-us/windows/uwp/design/style/app-icons-and-logos

the icon can be found at https://github.com/deltachat/interface/blob/master/icons/delta-v7-pathed.svg

NeoRon82:
closes #1212


Simon-Laux:
I believe the icons need to be directly in the `appx` folder:
> AppX assets need to be placed in the appx folder in the build directory.
~ https://www.electron.build/configuration/appx#appx-assets

Simon-Laux:
nice!

PR #1369, opened 2020-01-22
add verified checkmark into chat title in navbar

### LINKED ISSUE #1368: Add verified-checkmark right of Group-title (like Android)
Currently only the chatlist shows a verified-checkmark. 
The group-title above the message-list should also show a verified-checkmark like on Android. 
This makes it easier to quickly check if i am writing in a "secure" channel.

Simon-Laux:
closes #1368

### DIFF
--- a/scss/main_screen/_main_screen.scss
+++ b/scss/main_screen/_main_screen.scss
@@ -4,6 +4,12 @@
         font-size: medium;
         font-weight: bold;
         cursor: pointer;
+
+        img.verified-icon {
+            width: 0.8em;
+            height: 0.8em;
+            margin-right: 3px;
+        }
     }
 
     .navbar-chat-subtile {
--- a/scss/main_screen/_main_screen.scss
+++ b/scss/main_screen/_main_screen.scss
@@ -8,7 +8,7 @@
         img.verified-icon {
             width: 0.8em;
             height: 0.8em;
-            margin-right: 3px;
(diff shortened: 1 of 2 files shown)


r10s:
can we have the icon _right_ of the name please?
preferably also in the chatlist?

however, this should not block merging, it's already a great improvement :)

Simon-Laux:
sure we can have it on the right side, in the Chat I wouldn't move it to the right, I would prefer to move it to overlay the corner of the avatar like iOS does it.

r10s:
>  I would prefer to move it to overlay the corner of the avatar like iOS does it

yip, that's also good, also security-wise better as users cannot be tricked by adding some emoticons to the title. ios does the same for the title bar btw.

Simon-Laux:
well we don't have the avatar in the title bar.

r10s:
> well we don't have the avatar in the title bar.

i think it should be added anyway.

Simon-Laux:
sure? we removed it "recently" because it wasn't round and we decided that it doesn't need to be in there anyway.

Simon-Laux:
We could try how it looks I'm not against it.

r10s:
> sure? we removed it "recently" because it wasn't round and we decided that it doesn't need to be in there anyway.

yip, sure :)
and yes, it should be round then.
but it is not urgent in any way and not needed for now.

PR #1370, opened 2020-01-22
New version of Help page, with image.

pabzm:
The image's path requires a little extra handling. Maybe that path manipulation should go into the script in deltachat-pages?


Simon-Laux:
or you could parse and modify it like the links, but changing the pages script is probably easier..

Simon-Laux:
as this still needs some image path attention before merging, I added WIP to the title.

r10s:
> or you could parse and modify it like the links, but changing the pages script is probably easier..
How does android handle this image and its path? @r10s 

it's done by the [create-local-help](https://github.com/deltachat/deltachat-pages/blob/master/tools/create-local-help.py)-script in deltachat-pages.

the script changes the paths in the html-files to match the following directory structure:
```
DEST_DIR/help/LANG/help.html
DEST_DIR/help/help.css
DEST_DIR/help/image.jpg
```
also, the script copies the images. in the example, `help.html` would refer the image using `` - so, might be, this already works on desktop

pabzm:
This is not a WIP, it does contain the required path-adaptions. It is ready for reviewing and merging.

`../` does not work because all URLs are relative to `./static`. Therefore we need `./help/` as path. We would need to change the python-script to use a dynamic path, instead of the currently hard-coded `../`

r10s:
> Therefore we need ./help/ as path. We would need to change the python-script to use a dynamic path, instead of the currently hard-coded ../

yip, i think the script should be adapted then, maybe by passing a prefix or so that is different for android+desktop then.

Simon-Laux:
> This is not a WIP, it does contain the required path-adaptions. It is ready for reviewing and merging.

sorry for the misunderstanding.

PR #1372, opened 2020-01-22
remove duplicated netherlands language

### LINKED ISSUE #1308: in language list, combine "Dutch" and "Nederlands (Nederland)" to just Nederlands
we do not have two different languages for that and the local name is "Nederlands", the iso-code  in transifex is nl_NL.

Simon-Laux:
fixes #1308

### DIFF
--- a/_locales/_languages.json
+++ b/_locales/_languages.json
@@ -76,10 +76,6 @@
     "name": "Italian"
   },
   {
-    "locale": "nl",
-    "name": "Dutch"
-  },
-  {
     "locale": "nl_NL",
     "name": "Nederlands (Nederland)"
   },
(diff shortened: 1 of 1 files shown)


_locales/_languages.json
```diff
@@ -75,10 +75,6 @@
     "locale": "it",
     "name": "Italian"
   },
-  {
-    "locale": "nl",
-    "name": "Dutch"
-  },
   {
     "locale": "nl_NL",
     "name": "Nederlands (Nederland)"
```
r10s: although, for whatever reason, "nl_NL" is used as the language code, in the ui, we should not show countries or regions unless we have different language packets for them (true or portuguese). so, the string should just read "Nederlands" (which is an official language eg. also in  Belgium).

r10s:
perfect 👌

PR #1375, opened 2020-01-23
Show title, not search field when showing archived chats.

### LINKED ISSUE #1366: searchbar in archive view is not working
when the archive is opened, there is still the search input field atop of the list, however, it is not working and also the core does not provide sth. as "search over archive" - the reason is mainly that the "archived" state is nothing stable, things become unarchived when messages arrive, so that an explicit search is more confusing than helping.

the simplest fix would be to remove the searchbar (a search over the archive is already there, however, due to missing "archive" tags (#1365), not that visible) - the user can browse the archived chats and go back. 

alternatively, maybe use just the same search for archive view and non-archived view, however, i would quit the archive implicitly on any search then, to make clear it is a global search.

pabzm:
The search wasn't functional. This change is preliminary until someone
builds a change that fixes the search and jumps out of the archive-view
(the search is global, not restricted to the archives).

Screenshot:
2020-01-23-141354_408x165_scrot

Closes #1366

### DIFF
--- a/scss/main_screen/_navbar_wrapper.scss
+++ b/scss/main_screen/_navbar_wrapper.scss
@@ -70,4 +70,12 @@
     .bp3-popover-target {
         margin-right: 6px;
     }
+
+    .archived-chats-title {
+      flex: 1 1 auto;
+      overflow-x: hidden;
+      white-space: nowrap;
+      text-overflow: ellipsis;
+      font-size: 18px;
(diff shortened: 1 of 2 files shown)


Simon-Laux:
Is good in my opinion, but before merging I'd like to hear @r10s' opinion as he opened the issue.

PR #1378, opened 2020-01-23
various Gallery improvements

### LINKED ISSUE #1234: Clicking the play button of a video doesn't actually play it
when you click the play button of a video in the chat area, it is expanded to the full window, but isn't actually played, user has to click the play button again 

- Platform (linux/mac/windows): GNU/Linux
- Delta Chat Desktop Version: master

### LINKED ISSUE #1364: video ui drawing errors
tested with 0.901.0 on mac:

- the "play button" seems to be draw twice, but not at the same position which looks weird, see first image

- when a video in portrait-format is started, there is a reproducible hole in the overlay, see second image. maybe  the whole layout is different from the landscape-mode, eg. the huge shadow at the bottom is not present at all in landscape mode. wondering if this is desired.

Screen Shot 2020-01-20 at 12 59 12

Screen Shot 2020-01-20 at 13 00 23

Simon-Laux:
- fix #1364
- fix #1234
- [x] gallery should have own attachment types
- [x] move gallery/media to ts 

tasks moved to a new pr:
- cleanup used styles-scss
- improve how files look in the media - document view

_Note: 5950 of the added lines are from package-lock.json_


pabzm:
This is huge already, even if I don't count the changes in package-lock.json. Please don't grow it even bigger, rather finish this and open a second PR for the other open tasks.

Simon-Laux:
fair point.

Jikstra:
Let us merge this in the next release?

Simon-Laux:
It fixes 2 issues for our roadmap, review it commit by commit if it seems to be to much change.

src/renderer/components/attachment/Attachment.tsx
```diff
@@ -0,0 +1,83 @@
...
+const SUPPORTED_VIDEO_MIME_TYPES = Object.freeze([
+  'video/mp4',
+  'video/ogg',
+  'video/webm'
+])
+/* EndSection - Data Copied in part from Signal */
+
+// TODO define this correctly 
+// (maybe inside shared module??, but that depends on wether its also used in the backend or just exists in the frontend)
+export type attachment = { [key: string]: any, contentType: string, fileName: string, url: any }
```
pabzm: Most other type definitions are split over multiple lines (one per attribute). I find that easier to read, could you adapt this line, too?

Also what about the TODO?

src/renderer/components/attachment/messageAttachment.tsx
```diff
@@ -0,0 +1,106 @@
...
+import classNames from 'classnames';
+import { openAttachmentInShell } from '../message/messageFunctions';
+import { C } from 'deltachat-node/constants.enum';
+import { ScreenContext } from '../../contexts';
+import { isDisplayableByFullscreenMedia, isImage, isVideo, isAudio, getExtension, dragAttachmentOut, attachment } from './Attachment';
+
+const MINIMUM_IMG_HEIGHT = 150
+const MAXIMUM_IMG_HEIGHT = 300
+
+type AttachmentProps = { // TODO: replace "any" by the right type here
```
pabzm: Do you want to fix the TODO in the PR?

src/renderer/components/attachment/Attachment.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+/* Section - Data Copied in part from Signal */
+// Supported media types in google chrome
+// See: https://en.wikipedia.org/wiki/Comparison_of_web_browsers#Image_format_support
+const SUPPORTED_IMAGE_MIME_TYPES = Object.freeze([
+  'image/bmp',
+  'image/gif',
+  'image/jpeg',
+  'image/jpg',
+  // No need to support SVG
```
pabzm: Just curious: do you know why there is no need to support SVG?

src/renderer/components/attachment/messageAttachment.tsx
Simon-Laux: No I believe typing the message type is out of scope for this pr, as its already large and the message type is probably our biggest type.

src/renderer/components/attachment/Attachment.tsx
Simon-Laux: I'm kinda curious too, this code was copied from signal so I don't know why. It was previously in the now deleted GoogleChrome.js

src/renderer/components/attachment/Attachment.tsx
Simon-Laux: that's a stub, and also a incomplete file. That type will later get defined closer to where its created and then imported here so I didn't do it pretty here to save lines. But If you absolutely need that for it to be merged I can split it into multiple line, of course.

src/renderer/components/attachment/Attachment.tsx
pabzm: Ok, let's leave it at that.

src/renderer/components/attachment/messageAttachment.tsx
pabzm: Ok

src/renderer/components/attachment/Attachment.tsx
pabzm: I think we should write code under the assumption that it will live longer than expected. Quick&dirty leads to badly maintainable code, like DC Desktop already has a lot.
But I won't insist on this, no.

Simon-Laux:
svg in image tag doesn't seem to be vulnerable to xss: https://security.stackexchange.com/questions/148507/how-to-prevent-xss-in-svg-file-upload/192736#192736

src/renderer/components/attachment/Attachment.tsx
Simon-Laux: nah its already better than just writing `any` everywhere

PR #1379, opened 2020-01-23
🚧 WIP: Show all chat members, not only those present in searchContacts.

### LINKED ISSUE #1244: [BUG] Group has 7 members but ony 5 are shown when editing group
Analyzing the DB:
The contacts that are not shown have `64` in the `origin` field (contacts that I haven't a 1:1 chat with)

**NOTES:**
- The `chats_contacts` table lists them correctly in the group.
- Calling `chat.get_contacts()` with the python bindings return the 7 members so core is fine.

- Platform (linux/mac/windows): GNU/Linux
- Delta Chat Desktop Version: master
- Screenshots:
group

pabzm:
searchContacts are those contacts that the core returns as "known
contacts". This excludes e.g. contacts which haven't ever been included
in a message from a known sender. Those contacts can still be a group
member already.

Closes #1244

### DIFF
--- a/src/renderer/components/dialogs/EditGroup.js
+++ b/src/renderer/components/dialogs/EditGroup.js
@@ -123,7 +123,7 @@ export function EditGroupInner (props) {
               <GroupMemberSearchInput onChange={onSearchChange} value={queryStr} placeholder={tx('search')} />
               {renderAddMemberIfNeeded()}
               <ContactList2
-                contacts={searchContacts.filter(({ id }) => groupMembers.indexOf(id) !== -1)}
+                contacts={chat.contacts}
                 onClick={removeGroupMember}
                 showCheckbox
                 isChecked={() => true}


pabzm:
@nicodh, @Simon-Laux, @Jikstra: Do you see any reason not to merge this? I think the previous approach was more complicated, which sometimes has a reason.

Simon-Laux:
I don't know that part of the code well enough to know whats going on. @nicodh and @Jikstra know more than me because they recently worked on this component.

Jikstra:
I think the reason for this is, that we show two Contact lists on a search. In the upper one we have all Contacts which are in the group and which are in the search results, in the second we show all members which are not in the group but in the search results.

pabzm:
@Jikstra To rephrase my first comment:

`dc_get_contacts` (which is used to fill `searchContacts` in the code that we discus) does *not* return *all* contacts, but all *known* contacts. The latter is defined like this:

> Contacts that are known, i. e. they came in via accepted contacts or themselves an accepted contact. Known contacts are shown in the contact list when one creates a chat and wants to add members etc.

A group-member can be absent from that list.

My question is: why does the current code select the list of members from the list of known contacts, while `chat.contacts` is available, too?

Or: Was the list of *known* contacts just mistaken to be a list of *all* contacts, or is there some hidden logic to it?

Jikstra:
> Was the list of known contacts just mistaken to be a list of all contacts, or is there some hidden logic to it?

Ooooh now I get it. Yes this is what I thought I will get. Everything is a lie! :D

The problem I see with your pr, is, that the "search through members of this group" functionality should be broken (didn't test it, but the filter function is for exactly this).

r10s:
> The problem I see with your pr, is, that the "search through members of this group" functionality should be broken (didn't test it, but the filter function is for exactly this).

if that gets too complicated to fix, i'd say we can disable the "search through members of this group" functionality for now. compared to a correct memberlist, this search-function it is not super-important and there is the "add members" button.

Jikstra:
It's anyways confusing I have to say, even if i put it in there :D

Simon-Laux:
Is this still a discussion or is this pr merge-able?

pabzm:
> the "search through members of this group" functionality should be broken

It indeed is. I'll have a look to find a fix that doesn't break that feature. Thank you for the hint!

pabzm:
> Is this still a discussion or is this pr merge-able?

I'd like to do one attempt to fix the issue without breaking search in group members. If that fails (in reasonable time), I'd also go with the currently solution.

pabzm:
Unfortunately I found that with my change, newly added group members are not visible either. Displaying them also bases on the imagined "list of all contacts". So I'll have to find another way to fix the original bug.

pabzm:
I'm handing over to @Jikstra, he might have a better idea how to fix this.

PR #1385, opened 2020-01-24
add button for opening DM to profile view

### LINKED ISSUE #1357: Profile view: button to start DM missing
I only can open mutual chats, but not the direct chat with that person

Simon-Laux:
fixes #1357

### DIFF
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -70,5 +70,11 @@
   },
   "help_page_local_copy_hint": {
     "message": "This content is a local copy of [LINK:the FAQ page] on our website."
+  },
+  "profile_actions": {
+    "message": "Actions"
+  },
+  "profile_action_direct_message": {
+    "message": "Open Direct Message"
   }
 }
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -71,10 +71,7 @@
   "help_page_local_copy_hint": {
     "message": "This content is a local copy of [LINK:the FAQ page] on our website."
   },
-  "profile_actions": {
-    "message": "Actions"
-  },
   "profile_action_direct_message": {
-    "message": "Open Direct Message"
+    "message": "Open Chat"
   }
 }
--- a/src/main/deltachat/contacts.js
+++ b/src/main/deltachat/contacts.js
@@ -62,4 +62,18 @@ module.exports = class DCContacts extends SplitOut {
   markNoticedContact (contactId) {
     return this._dc.markNoticedContact(contactId)
   }
+
+  /** Gets the direct message chat id with this contact and creates it if it doesn't exist yet */
+  getDMChatId (contactId) {
+    const existingChatId = this._dc.getChatIdByContactId(contactId)
+    if (existingChatId !== 0) {
+      return existingChatId
+    }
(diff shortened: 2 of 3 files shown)


Simon-Laux:
I'm not quite sure about the language here, using the existing "new Chat" string that is shown in android doesn't look good.
Bot on the other hand I don't know if a user knows what "Direct Message" means, it could workout well if translators translate it well like for german "Einzelchat" would be easy to understand IMO.

pabzm:
Thanks for building this!

I agree that "Direct Message" is not a good phrase here, and neither is "new chat". I see nothing fitting in the locale-files, we appear to not have a good phrase for 1-on-1-chats yet. so let's invent something new?

How about "Chat with %1" (similar to `ask_start_chat_with`, but without the question mark). Or "Personal chat with %1".

Simon-Laux:
> "Personal chat with %1"

I like that

Jikstra:
Why not have the plus avatar as the first item in the `Shared Chats` section and name it `New Chat`? I think this should be fine, better then introducing a new section which only has one item/action.

r10s:
i would also go for the android way. at least it is there since some time this way and there have never been complains about it afaik. seeing this "new chat" in the context of existing chat is maybe pretty obvious.

Simon-Laux:
I would complain about the "new chat" wording on android, but It wasn't important enough for me yet
We could move it that button to the other section, yes.
I believe we should ask some native speakers about the phrasing of direct message, but on the otherhand, do we need perfect wording to merge this in the first iteration?

A new idea: what about having an (message bubble icon) icon only button next to the contact in the top (right side, there is still space left)?
-- 

r10s:
let's go just for the same term we're using on android and ios currently, so "New chat".

even if not everyone agrees, this ensures internal consistence, makes help and faqs easier etc.  i think it is bad to introduce a new term here that has to be explained and that does not even exist on android/ios.  
we should try to use the same wording for basic functionality - and the shared translations are a great way for that.

at some point, we can discuss the wording in the interface repo, and, if needed, change it on all platforms then.

Simon-Laux:
I first tried that. I don't like it.

> but please use the existing term

Sure that you don't just want to keep the new translation strings low count? :wink:

Simon-Laux:
What about a button next to the Profile in the top just that says `[Open Chat]`?
I think could solve it.
2020-01-24_06-21-idea

r10s:
not sure if this is better - "opening" something implies it is already there - and this is not always true. i would just stuck with the existing one for now - at least, this saves lots of energy that can be used for different things :)

Simon-Laux:
> "opening" something implies it is already there - and this is not always true

Same argument works even better for "new chat"

I think I'll just implement  the thing above ("Open Chat" next to profile) instead of us discussing endlessly about whether we should or shouldn't one or two new language string/s. (the discussion here right now is the biggest waste of time IMHO)

r10s:
it's not about adding a string or not. but i agree to close the discussion :)

pabzm:
Would it be very complicated to check if there's a chat with that contact already, and select the wording accordingly? "Open chat" if there's a chat, "New chat" if there isn't one yet?

(I find both terms not fitting for both cases, that's why I suggested "Personal chat", which doesn't include the notion of existance (or lack thereof).)

Issue #1134, opened 2019-11-13
Electron apps must be updated to get into macOS App Store


duckdalbe:
Currently no electron app can be submitted to the macOS App Store, because Apple tightend the checks: <https://github.com/electron/electron/issues/20027>.

There's a development version of electron available that disables certain features, which passes Apples test, but works slower: <https://github.com/electron/electron/pull/20965>.

AFAIK DC desktop should be available via the macOS App Store, once it's stable. So that issue should be watched.

Simon-Laux:
We're currently on electron 4 because there was an SUID sandbox issue on debian when going above version 4 (https://github.com/deltachat/deltachat-desktop/issues/895) not sure if its already fixed because the upstream electron issue was closed before it was really fixed.

Jikstra:
The suid thing wont get fixed as far as i know. This SUID thing seems to be the default way to do proper sandboxing on unix.

Simon-Laux:
so we need another binary for debian?

pabzm:
Electron versions https://github.com/electron/electron/releases/tag/v5.0.13 and https://github.com/electron/electron/releases/tag/v6.1.7 fix this by including a workaround.

These workarounds may introduce performance regressions according to the electron developers, but many people reported that their apps still run fine with it.

For electron >= 7 there's apparently no fix yet, because the developers [want a "real" solution](https://github.com/electron/electron/issues/20027#issuecomment-565271491) [to the problem](https://github.com/electron/electron/issues/20027#issuecomment-565865582) in those releases.

Simon-Laux:
we upgraded electron to `6.1.7` in #1400

PR #1400, opened 2020-01-29
Update to electron-builder 22 (fixes SUID sandbox issues)

pabzm:
electron-builder >= 22.0.0 fixes the setuid problem with the sandboxing helper: https://github.com/electron-userland/electron-builder/pull/4163

This helper is required if `kernel.unprivileged_userns_clone` is not allowed (`0`), e.g. on Debian and derivates. (On other linux systems user name-spaces are used instead of the sandboxing helper.)

Fixing the sandbox-helper problem would enable us to update electron (see #895 and #948), which is apparently required to get into the MacOS app store (#1134, https://github.com/electron/electron/issues/20027).

The update is jumping to a new major version, but [the Changelog](https://github.com/electron-userland/electron-builder/releases) doesn't list any breaking changes, so I suppose this shouldn't make problems.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6032,9 +6032,9 @@
       "dev": true
     },
     "electron": {
-      "version": "4.2.12",
-      "resolved": "https://registry.npmjs.org/electron/-/electron-4.2.12.tgz",
-      "integrity": "sha512-EES8eMztoW8gEP5E4GQLP8slrfS2jqTYtHbu36mlu3k1xYAaNPyQQr6mCILkYxqj4l3la4CT2Vcs89CUG62vcQ==",
+      "version": "6.1.7",
+      "resolved": "https://registry.npmjs.org/electron/-/electron-6.1.7.tgz",
+      "integrity": "sha512-QhBA/fcYJit2XJGkD2xEfxlWTtTaWYu7qkKVohtVWXpELFqkpel2DCDxet5BTo0qs8ukuZHxlQPnIonODnl2bw==",
       "dev": true,
       "requires": {
         "@types/node": "^10.12.18",
@@ -6043,9 +6043,9 @@
       },
       "dependencies": {
         "@types/node": {
-          "version": "10.17.13",
-          "resolved": "https://registry.npmjs.org/@types/node/-/node-10.17.13.tgz",
-          "integrity": "sha512-pMCcqU2zT4TjqYFrWtYHKal7Sl30Ims6ulZ4UFXxI4xbtQqK/qqKwkDoBFCfooRqqmRu9vY3xaJRwxSh673aYg==",
+          "version": "10.17.14",
+          "resolved": "https://registry.npmjs.org/@types/node/-/node-10.17.14.tgz",
+          "integrity": "sha512-G0UmX5uKEmW+ZAhmZ6PLTQ5eu/VPaT+d/tdLd5IFsKRPcbe6lPxocBtcYBFSaLaCW8O60AX90e91Nsp8lVHCNw==",
           "dev": true
         }
       }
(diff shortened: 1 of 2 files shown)


Simon-Laux:
can some folks test the appimage of this PR? I forgot the password of my debian VM

Simon-Laux:
Would be really awesome if this works.

pabzm:
Doesn't work... :( Oh my, it sounded like a good plan. I'll try to find out, what's wrong.

You don't need debian to test, just do `sudo sysctl kernel.unprivileged_userns_clone=0`.

pabzm:
Hmm, maybe this runs only after installing a packages somewhere, but not during packaging? https://github.com/electron-userland/electron-builder/pull/4163/files

I'll investigate tomorrow.

pabzm:
Well, duh... The commit message of https://github.com/electron-userland/electron-builder/pull/4163 says that this is a fix for the debian package. So maybe this does help debian people by enabling them to use the debian-package at least? Could someone test that?

pabzm:
Works! 🎉

I installed a locally built .deb in a Debian Buster (v10) VM and could start it, I'd say that's a sufficient check?

From my point of view we could go down this route. Debian users will still not be able to use the AppImage out of the box, but I think that's a sufferable loss.

pabzm:
I also checked that running the flatpak-installed thing (from flathub.org, not locally modifed) works, so Debian users had one option left anyway. Now they'd have two.

Simon-Laux:
linking some related links:
https://github.com/electron-userland/electron-builder/issues/4495
https://github.com/electron-userland/electron-builder/pull/4496

Simon-Laux:
I agree to merge this, but I'm still a bit unsure about the AppImage...
Let's do it and see what happens - YOLO :grin: .

Simon-Laux:
we should scroll up from this https://www.electronjs.org/docs/breaking-changes#planned-breaking-api-changes-40 and check for apis we use.

Jikstra:
Seems like we can upgrade electron & builder without bigger issues. For the AppImages we could still stay on electron 4 if that helps.

pabzm:
@Simon-Laux: Why did you merge this despite open questions? @Jikstra had suggested to stay with electron 4 for the AppImage — I don't agree to that, but we should not ignore each another like this.

@Jikstra Wouldn't that mean to maintain different dependencies for different build targets? I don't know how this could work, and I am afraid of the complexity it brings, and I think it's not worth the effort: People shouldn't use AppImage anyway, if any other alternative works (AppImage has no validated download, no update-channel, etc.).
I'd rather suggest to put some effort into setting up a Debian repository in order to make it easy for Debian users to install Delta Chat.

Simon-Laux:
I thought everything was solved? I also discussed this with jikstra IRL.
Debian might have an error then on appimage start I don't see that as a big deal for now, we can upgrade electron builder again when  https://github.com/electron-userland/electron-builder/pull/4496 is merged to fix the appimage again.

> (AppImage has no validated download, no update-channel, etc.).

We could fix that if we want, but I don't value it important enough right now.

About the incompatibilities I checked again before merging and the changed default does not affect us, because we don't use a preload script.
-- 

pabzm:
Ok, thank you for the explanation. I didn't know you had been talking with jikstra AFK. Could you next time drop a hint, so I won't bother you again in such a case?

Also thanks for checking out electron's breaking changes!

Simon-Laux:
I'll make sure to not forget to mention IRL/AFK decisions here, in the future.
PS: Electron 7 is more breaking than 6, so good choice.

Am 4. Februar 2020 14:48:43 MEZ schrieb Pablo <notifications@github.com>:
>Ok, thank you for the explanation. I didn't know you had been talking
>with jikstra AFK. Could you next time drop a hint, so I won't bother
>you again in such a case?
>
>Also thanks for checking out electron's breaking changes!
>
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/deltachat/deltachat-desktop/pull/1400#issuecomment-581917486

-- 

PR #1401, opened 2020-01-29
Update node-sass to fix a vulnerability.

pabzm:
https://www.npmjs.com/advisories/961


Simon-Laux:
Is not critical as it is run at build time, but can't hurt I guess.

PR #1404, opened 2020-01-30
Replace standardjs with prettier for code formatting

### LINKED ISSUE #1040: Linter / Codefromatter for SCSS?
would be useful IMO, I currently use the vscode plugin `michelemelluso.code-beautifier` for it, but that isn't a solution for @Jikstra because he uses a different editor.
I would like something like standardjs in terms of auto fixing capabilities and vscode integration availability (but the latter one isn't a huge deal breaker because I can add npm tasks to vscode.)

Simon-Laux:
Only the first commit has stuff to review, the second commit is the result from running prettier.

For linting I planned using TSLints and https://www.npmjs.com/package/tslint-config-prettier for compatibility with the prettier formatter. But unfortunately tslint gets deprecated and merged into eslint right now, I don't know the exact state, but it doesn't seem to be ready so I didn't touched that in this pr.

closes #1040


Simon-Laux:
But maybe we don't even need tslint that much because some standard checks (as example unused imports)  are also checked by the typescript compiler.

nicodh:
what is the exact reason to change to prettier?

Simon-Laux:
standard doesn't work on scss and ts. I searched for scss linters/formatters but found none that has a`--fix` command. I tried `prettier` and it worked while also having a check command.

Simon-Laux:
At best rebase and merge this when there are no other code prs open

PR #1415, opened 2020-02-03
Performance improvements for messagelist

Jikstra:
On each update a functional component will "rerender", or at least build the virtual dom stuff and compare it with what is there. To avoid this the 2nd memo parameter allows us to check if we actually need to do anything. I added this to the new MessageListInner component which checks if the messages to show changed (if we need to show new messages) and we also check this for each Message component if the message object changed it's identity and only if so we rerender. Besides this i moved the useStore call out of MessageWrapper as a useState change always triggers a rerender, causing us to rerender all Messages no matter what changed in the chat store.

### DIFF
--- a/src/main/ipc.js
+++ b/src/main/ipc.js
@@ -186,7 +186,7 @@ function init (cwd, state, logHandler) {
   ipcMain.on('saveLastChatId', (e, chatId) => {
     const { lastChats } = app.state.saved
     lastChats[dcController.credentials.addr] = chatId
-    sendStateToRenderer()
+    //sendStateToRenderer()
     // don't save to disk, because this is already done on close and it might block
     // we can ignore the crash case, because a crash isn't supposed to happen
     // and it's not important data
--- a/src/main/ipc.js
+++ b/src/main/ipc.js
@@ -186,7 +186,6 @@ function init (cwd, state, logHandler) {
   ipcMain.on('saveLastChatId', (e, chatId) => {
     const { lastChats } = app.state.saved
     lastChats[dcController.credentials.addr] = chatId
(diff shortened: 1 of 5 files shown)


src/main/ipc.js
```diff
@@ -186,7 +186,7 @@ function init (cwd, state, logHandler) {
   ipcMain.on('saveLastChatId', (e, chatId) => {
     const { lastChats } = app.state.saved
     lastChats[dcController.credentials.addr] = chatId
-    sendStateToRenderer()
+    //sendStateToRenderer()
```
pabzm: Shouldn't this line be deleted if it shall not be executed? Commenting out looks like you would like to re-use it soon.

Jikstra: Sure :) On each update a functional component will "rerender", or at least build the virtual dom stuff and compare it with what is there. To avoid this the 2nd memo parameter allows us to check if we actually need to do anything. I added this to the new MessageListInner component which checks if the messages to show changed (if we need to show new messages) and we also check this for each Message component if the message object changed it's identity and only if so we rerender. Besides this i moved the useStore call out of MessageWrapper as a useState change always triggers a rerender, causing us to rerender all Messages no matter what changed in the chat store.

Jikstra: Yes that line is unneeded. I only commented it out as this is still WIP with a lot of debug stuff. Aggressive performance improvements sometimes also cause bugs, especially when caching/memoizing things. I'm currently testing this branch to check if things are broken or if we can merge it.

pabzm: Ok, thanks!

Sorry for interrupting your Work In Progress, I didn't see the marking before!

PR #1416, opened 2020-02-03
Provider db

### LINKED ISSUE #1408: Make use of the new provider-db
Desktop should use the provider-db [like Android already did](https://github.com/deltachat/deltachat-android/pull/1219).

Simon-Laux:
- [x] update deltachat-node

closes #1408 
Adds provider db and converts 2 files to typescript.
[video.zip](https://github.com/deltachat/deltachat-desktop/files/4146289/video.zip)
github doesn't let me upload the mp4 as video, so zipped it.

Screenshot_20200207_212501
Screenshot_20200207_212532

### DIFF
--- a/src/renderer/components/Login.js
+++ b/src/renderer/components/Login.js
@@ -8,9 +8,7 @@ import {
   DeltaInput,
   DeltaPasswordInput,
   DeltaSelect,
-  DeltaProgressBar,
-  AdvancedButtonIconClosed,
-  AdvancedButtonIconOpen
+  DeltaProgressBar
 } from './Login-Styles'
 import {
   Collapse,
@@ -230,7 +228,7 @@ export default class Login extends React.Component {
 
           <p className='text'>{tx('login_no_servers_hint')}</p>
           <div className='advanced' onClick={this.handleUISwitchStateProperty.bind(this, 'showAdvanced')} id={'show-advanced-button'}>
-            {(showAdvanced ? <AdvancedButtonIconClosed /> : <AdvancedButtonIconOpen />)}
+            <div className={`advanced-icon ${showAdvanced && 'opened'}`} />
             <p>{tx('menu_advanced') }</p>
           </div>
           <Collapse isOpen={showAdvanced}>
--- a/src/renderer/components/Login.js
+++ b/src/renderer/components/Login.js
@@ -13,12 +13,14 @@ import {
   DeltaProgressBar,
   AdvancedButton,
   AdvancedButtonIconClosed,
-  AdvancedButtonIconOpen
+  AdvancedButtonIconOpen,
+  BeforeLoginHint
 } from './Login-Styles'
 import {
   Collapse,
   Intent
 } from '@blueprintjs/core'
+import { callDcMethodAsync } from '../ipc'
 
 export default class Login extends React.Component {
   constructor (props) {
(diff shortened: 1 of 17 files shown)


src/renderer/components/Login.tsx
```diff
@@ -1,6 +1,7 @@
 /* eslint-disable camelcase */
 
-import C from 'deltachat-node/constants'
+import { C } from 'deltachat-node/constants.enum';
+import { DeltaChat } from 'deltachat-node';
```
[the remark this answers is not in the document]
Simon-Laux: necessary for importing types

src/renderer/components/Login.js
```diff
@@ -215,6 +224,14 @@ export default class Login extends React.Component {
             password={mail_pw}
             onChange={this.handleCredentialsChange}
           />
+
+          { provider_info.before_login_hint &&
+          provider_info.status !== 0 &&
+          <BeforeLoginHint className={provider_info.status === C.DC_PROVIDER_STATUS_BROKEN && 'broken'}>
+            {provider_info.before_login_hint}
```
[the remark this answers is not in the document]
Simon-Laux: good catch, I forgot that

src/renderer/components/Login.tsx
```diff
@@ -215,14 +256,24 @@ export default class Login extends React.Component {
...
-          <DeltaText>{tx('login_no_servers_hint')}</DeltaText>
-          <AdvancedButton onClick={this.handleUISwitchStateProperty.bind(this, 'showAdvanced')} id={'show-advanced-button'}>
-            {(showAdvanced ? <AdvancedButtonIconClosed /> : <AdvancedButtonIconOpen />)}
+
+          { provider_info?.before_login_hint &&
+          <div className={`before-login-hint ${(provider_info.status === C.DC_PROVIDER_STATUS_BROKEN && 'broken')}`}>
+            <p>
+              {provider_info.before_login_hint}
+            </p>
+            <ClickableLink href={provider_info.overview_page}>More Information at {provider_info.overview_page}</ClickableLink>
```
pabzm: I would prefer to use just "More info" as link text, like the android app does. Showing URLs is ugly, in my eyes. And we would have a consistent behaviour with the android app.

If you reject my suggestion: "information" must not be capitalized.

r10s: if the decision goes towards "More info", android uses the string originally introduced by desktop, `more_info_desktop`

Simon-Laux: IDK my hope with this was to don't force opening the link in the system browser. (privacy stuff) but we can also do the `More info` text, but maybe add an `(online)` to make clear that it goes to the browser?
Either way is fine for me, just wanted to throw that thought out here.

Simon-Laux:
Before merging we should also reference a new dc-node version here that has the type, otherwise it won't compile nor run.

src/renderer/components/Login.tsx
pabzm: What does "privacy stuff" mean concretely?

People who care for privacy probably have a system browser with decent privacy settings, I would assume. And the link doesn't reveal a lot: only that this person might be using Delta Chat, and has an email address with provider X. Both of which is probably already known to the system browser, or the connectivity provider.

pabzm: Any more opinions or reactions?

Simon-Laux: We should merge it now(after requiring a new dc-node version) and do a separate pr to change it later to prevent this being stale.

pabzm: Why not discuss this as long as this is waiting anyway?

You wrote:
> we can also do the More info text

If you are OK with it, and no one objected my suggestion, and you don't fancy further discussion — why not just do it?

PR #1417, opened 2020-02-03
Improve styling of help pages

Jikstra:
Before:
grafik

After:
grafik


r10s:
is the hint to the online FAQ page really that important that it requires a permanent hint?

tbh, i would just remove it. there is the link "learn more about Delta Chat" that opens the homepage in the main menu. and, if we really think, we need some kind of "this is a copy from ...", we can just add it to the page, this way, we would have it also on all systems. but i do not think it is needed at all.

pabzm:
> is the hint to the online FAQ page really that important that it requires a permanent hint?

It's not, this is just what we discussed (as of my holey mind). I like the hint, but I don't find it important, so go ahead.

Jikstra:
@Simon-Laux can't reproduce:
grafik

PR #1428, opened 2020-02-07
Open help in browser

Jikstra:
Screenshot_20200207_181840

Open the help page in the local browser instead of the dialog.
You can also press `F1` now to bring up the help.


pabzm:
Why?

Simon-Laux:
> Why?

description will follow later, when the pr is done

Point is that the code for the inline help page is too complicated in the sense that it requires nodejs native modules in the renderer process and that is bad.
Also we'd need to adjust the generator script because getting an misplaces html element error is not too nice.

Also with the browser window you have a seperate window that you can even search like any other website.

Simon-Laux:
Maybe we should change the font, the default system font doesn't look that good (atleast on my system)

Simon-Laux:
Also to copy it every time on open (even if it wasn't changed (aka. same version) is be a bit inefficient here.
But on the other hand this solution is still fast enough IMHO.

pabzm:
I think that opening a file in a browser from a GUI is horrible UX. If — at all — something is opened in the browser the relevant link must explicitly say so. But having "Open help in browser" in the menu is really ugly, too. Please don't do this.

Jikstra:
I think it's perfect UX
- people can browse the help side by side with using the app, the dialog is really unusable for this
- it uses the full screen, it's good to read

And also from development side:
- we only need half the lines of code
- don't need to include another page in our frontend code
- don't need to require fs in the frontend (dafuq?)
- we do the same as android or ios
- it will work in the browser as soon as we leave electron behind

pabzm:
The footer had a link to the online copy of the file, which served the same purpose (external window, more width) and more (maybe more up-to-date content).

Opening links in unsolicited new windows is one of the biggest, oldest DON'Ts of UX/webdesign, it made the list of "Top 10 Mistakes in Webdesign" by UX-guru Jacob Nielson [in 1999](https://www.nngroup.com/articles/the-top-ten-web-design-mistakes-of-1999/) *and* [in 2011](https://www.nngroup.com/articles/top-10-mistakes-web-design/). 

And opening local files in a browser is not a thing normal people do. We build Delta Chat mostly for normal people, not for nerds. (Of course I do like if nerds use Delta Chat, but nerds always will figure out how to do stuff — when designing the app and its UX we must think of "normal" people.)

But I learned through another channel that you actually discussed this with different people afk, and don't want to interfere with your discourse. (Please do start to explain your issues and PRs to everyone who isn't aware of your afk talk, so we can avoid these kind of discussions.)

PR #1435, opened 2020-02-07
Fix disabled composer hint overflowing on small screens

Jikstra:


src/renderer/components/composer/Composer.js
```diff
@@ -70,12 +70,11 @@ const Composer = React.forwardRef((props, ref) => {
...
   if (isDisabled) {
     if (disabledReason) {
-      return <div
-        ref={ref}
-        className={'composer composer--disabled-message-input'}
-        style={{ textAlign: 'center', padding: '0.5rem', color: '#999' }}>
-        {tx(disabledReason)}
-      </div>
+      return (
+        <div ref={ref} className='composer composer--disabled-message-input'>
```
Simon-Laux: you could rename this class(`composer--disabled-message-input`) to `disabled`.

PR #1441, opened 2020-02-11
Fix fullscreen

Jikstra:


Simon-Laux:
please rebase:  "Remove preview and version from window app title" does NOT belong here

PR #1444, opened 2020-02-11
Show empty chat and gallery texts

### LINKED ISSUE #1224: show some text when a chat is empty
when a chat is empty, currently the whole screen is empty as well (at least in the cases i've checked).

instead showing an empty screen, we should show some helpful text.

the text snippets are already there:

- for an **empty 1:1 chat:** `chat_no_messages_hint`
- for an **empty group:** if unpromoted: `chat_new_group_hint` - otherwise: `chat_no_messages`
- for an **empty contact request view:** `chat_no_contact_requests` if show_emails!=DC_SHOW_EMAILS_ALL - otherwise just use `chat_no_messages `
- for an **empty saved-messages chat:** `saved_messages_explain` 
- for an **empty device-chat:** `device_talk_explain` - (may happen when the user deletes all messages)

i suggest to show these text just in the same style as the system-info-messages (member-added etc.), no need to go for more styles or so.

this is the android code for figuring out the different states: https://github.com/deltachat/deltachat-android/blob/master/src/org/thoughtcrime/securesms/ConversationFragment.java#L161
corresponding ios issue: https://github.com/deltachat/deltachat-ios/issues/428

nicodh:
resolves #1224

### DIFF
--- a/scss/gallery/_media.scss
+++ b/scss/gallery/_media.scss
@@ -14,4 +14,9 @@
             float: left;
         }
     }
+    .no-media-message {
+      font-size: 1.1rem;
+      text-align: center;
+      margin-top: 5rem;
+    }
 }
--- a/scss/message/_message-list.scss
+++ b/scss/message/_message-list.scss
@@ -42,6 +42,14 @@
         clear: both;
         height: 0;
       }
+      .info-message {
+        max-width: 550px;
+        font-size: 1rem;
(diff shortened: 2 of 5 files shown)


Jikstra:
Screenshot_20200212_000852

It's fine but doesn't really fit into the other system messages in my opinion. Is this on purpose or do we want to introduce this new type of system message?

PR #1446, opened 2020-02-12
Remove checkbox in quick member add on new group, apply same ux on edit group

Jikstra:
Screenshot_20200212_171526

Screenshot_20200212_171742


nicodh:
Looks good to me. The only point which is counter intuitive for me is that it is required to click on the arrow to add the selected new members. Would be better if clicking on the "Add member" would work also...
But that can be fixed in another PR

PR #1454, opened 2020-02-13
Upgrade deltachat-node to `v1.0.0-beta.26`

Simon-Laux:


nicodh:
There are a lot of StandardJS errors

Jikstra:
@Simon-Laux not sure what's wrong but seems like all tests fail

PR #1459, opened 2020-02-16
Add --multiple-instances flag to allow multiple deltachat instances r…

Jikstra:
…unning at the same time


Simon-Laux:
whats the reason for this pr?

Jikstra:
To be able to run multiple deltachat instances to make debugging easier. Needed this to check if mark messages read works (2accounts, 2deltachat instances, 1 computer)

Issue #1461, opened 2020-02-16
QR Image too big and scrolled


Jikstra:
Fixed a similiar bug already some weeks ago, not sure what's wrong, but for sure CSS

PR #1462, opened 2020-02-16
Fix qr images getting scrollable

Jikstra:
Fixes https://github.com/deltachat/deltachat-desktop/issues/1461


nicodh:
Can't reproduce the issue and thus can't confirm that it's solved...

PR #1464, opened 2020-02-18
add a Build script

Simon-Laux:
run `npm run builder -- --help` for usage of the new build script.
goal is to make the build process fast(er) and to fix the source map mappings.
An Idea was also to have less of the logic inside of package.json#scripts


nicodh:
I still see all variables minified in source maps...
Is this PR intended to solve that?

Simon-Laux:
Now it doesn't minify anymore (unless --prod is specified)

PR #1468, opened 2020-02-19
Fix import of formatRelativeTime

nicodh:
Now locations are shown on map again


Simon-Laux:
we should really convert more file to ts soon, to prevent such error from appearing...

Issue #1469, opened 2020-02-20
Creating new chat from search bar is broken


pabzm:
If I enter an email address that I didn't chat with before into the search bar, I get the option to create a new chat with that address ("+ New contact").

Clicking on that button doesn't work (anymore). In the app, nothing happens, the log output shows this:

```
3499.9s [D]main/ipc: EVENT_DC_DISPATCH_CB (214) : contacts.createContact false,deltabot@example.net
3499.9s [E]main/deltachat: Error calling contacts.createContact(false, deltabot@example.net):
 TypeError: DeltaChat.maybeValidAddr is not a function
    at DCContacts.createContact (/home/user/dev/dc/deltachat-desktop/tsc-dist/main/deltachat/contacts.js:29:24)
    at DeltaChatController.<anonymous> (/home/user/dev/dc/deltachat-desktop/tsc-dist/main/deltachat/controller.js:156:37)
    at Generator.next (<anonymous>)
    at /home/user/dev/dc/deltachat-desktop/tsc-dist/main/deltachat/controller.js:7:71
    at new Promise (<anonymous>)
    at __awaiter (/home/user/dev/dc/deltachat-desktop/tsc-dist/main/deltachat/controller.js:3:12)
    at DeltaChatController.callMethod (/home/user/dev/dc/deltachat-desktop/tsc-dist/main/deltachat/controller.js:143:16)
    at Object.<anonymous> (/home/user/dev/dc/deltachat-desktop/tsc-dist/main/ipc.js:87:48)
    at Generator.next (<anonymous>)
    at /home/user/dev/dc/deltachat-desktop/tsc-dist/main/ipc.js:7:71 
DeltaChatController.<anonymous> (/home/user/dev/dc/deltachat-desktop/tsc-dist/main/deltachat/controller.js:159:21),
/home/user/dev/dc/deltachat-desktop/tsc-dist/main/deltachat/controller.js:7:71,
__awaiter (/home/user/dev/dc/deltachat-desktop/tsc-dist/main/deltachat/controller.js:3:12),
DeltaChatController.callMethod (/home/user/dev/dc/deltachat-desktop/tsc-dist/main/deltachat/controller.js:143:16),
Object.<anonymous> (/home/user/dev/dc/deltachat-desktop/tsc-dist/main/ipc.js:87:48)
3499.9s [D]main/ipc: EVENT_DC_DISPATCH_CB (215) : contacts.createChatByContactId 
[...]

PR #1470, opened 2020-02-20
Fix creating new chat by fixing the wrong import of deltachat-node. U…

Jikstra:
…se es6 imports for deltachat/contacts

Fixes https://github.com/deltachat/deltachat-desktop/issues/1469


Simon-Laux:
could you convert contact.js to typescript, too?
and if possible don't use 'default' in the splitouts (IMHO)

Jikstra:
> could you convert contact.js to typescript, too?

Would like to do this at a later point. I'm not really used to typescript yet and would like to not do unforseeable things currently.

> and if possible don't use 'default' in the splitouts (IMHO)

Why?

PR #1472, opened 2020-02-20
Move/Refactor/abstract all electron imports to src/renderer/ipc

Jikstra:
This allows us to move all the electron depending methods in src/renderer/ipc to gradually move them to a generic browser solution. This also converts src/renderer/ipc to typescript.

Won't merge this before the 1.0 release

### DIFF
--- a/src/renderer/ipc.ts
+++ b/src/renderer/ipc.ts
@@ -3,10 +3,29 @@
  * to be able to switch this layer later on...
  */
 
-import { ipcRenderer } from 'electron'
-const log = require('../shared/logger').getLogger('renderer/ipc')
+import { ipcRenderer, remote, clipboard as _clipboard } from 'electron'
+import { ExtendedApp, AppState } from '../shared/shared-types'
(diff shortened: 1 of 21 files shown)


Simon-Laux:
this is in direct conflict with https://github.com/deltachat/deltachat-desktop/pull/1465

src/renderer/ipc.ts
```diff
@@ -3,11 +3,30 @@
...
+
+export const openItem = remote.shell.openItem
+
+export const showSaveDialog = remote.dialog.showSaveDialog
+
+export const showMessageBox = remote.dialog.showMessageBox
+
+export const getPath = remote.app.getPath 
+
+export const clipboard = _clipboard
```
Simon-Laux: use navigator.clipboard instead, we don't need the electron variant, anyway would be better If you had opened this on top of my preload pr.....

Simon-Laux:
Let's re-implement it later

PR #1494, opened 2020-03-03
Focus composer textarea when chat id changes

Jikstra:
Fixes https://github.com/deltachat/deltachat-desktop/issues/1492

### DIFF
--- a/src/renderer/components/message/MessageList.js
+++ b/src/renderer/components/message/MessageList.js
@@ -65,7 +65,7 @@ export default function MessageList({
 
   useEffect(() => {
     isFetching.current = false
-    
+
     const composerTextarea = refComposer.current.childNodes[1]
     composerTextarea && composerTextarea.focus()
   }, [chat.id])
--- a/src/renderer/components/message/MessageList.js
+++ b/src/renderer/components/message/MessageList.js
@@ -65,6 +65,9 @@ export default function MessageList({
 
   useEffect(() => {
     isFetching.current = false
+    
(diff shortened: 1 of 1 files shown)


src/renderer/components/message/MessageList.js
```diff
@@ -65,6 +65,9 @@ export default function MessageList({
 
   useEffect(() => {
     isFetching.current = false
+    
+    const composerTextarea = refComposer.current.childNodes[1]
```
Simon-Laux: this looks like a hack to me... how about making a focus function inside of the composer component and calling it here?

Jikstra: You can only bind functions to class components.

Simon-Laux: I vote for making the compose a class, it makes sense, also when we have more advanced compositing going on it might be useful to have more control over that part.

Simon-Laux: like `:emoji:`- and user-mention-(dropdown|prediction) 
or fancy richtext/markdown previews like the ones discord introduced (though that will be far in the future I believe)

Jikstra: I'm fine with making it a class, whatever floats the boat, but this pr is to fix a bug and not to rewrite something.

Simon-Laux: okey.. but let's don't forget it :wink:

Simon-Laux:
please `npm run fix-formatting` that the test pass, then we can merge it.

PR #1495, opened 2020-03-03
Typescriptify backend

Jikstra:
Lets not merge this before releasing 1.1.0 (the ecc fix release)


src/main/deltachat/chat.ts
```diff
@@ -16,24 +19,24 @@ module.exports = class DCChat extends SplitOut {
...
   }
 
-  leaveGroup(chatId) {
+  leaveGroup(chatId: number) {
     log.debug(`action - leaving chat ${chatId}`)
     this._dc.removeContactFromChat(chatId, C.DC_CONTACT_ID_SELF)
   }
 
-  setName(chatId, name) {
+  setName(chatId: number, name: any) {
```
Simon-Laux: we know that the name is a string don't we?
If it also can be null, which it  probably can **not** in this case, you could do `string|null` in order to be supper correct and to comply with the compiler if we would ever activate "no Implicit null"

Jikstra: We should first specify this in deltachat-node and then bubble it up.

Simon-Laux: feel free to do so

Simon-Laux:
I'm taking over this pr (continue it)

Simon-Laux:
So the backend is basicaly converted (only the backend in meaning the deltachat folder).
The only thing left to do is review it and resolve some `any`s in login.ts:

> login.ts has still some `any`s that can be resolved such as credentials

Simon-Laux:
Also I have the feeling that I smashed some smaller bugs in the process or at least potential one... :wink:

PR #1497, opened 2020-03-07
Fix that 'copy selection' is only visible on

Simon-Laux:
selected text. the problem was that the option was always displayed.


Jikstra:
Don't really get what this is fixing. But what I notice is that the "Copy Link" menu entry when right clicking on a message is shown no matter if there's a link or not in that message.

Screenshot_20200310_143755

Issue #890, opened 2019-07-11
Adjust global font size


nicodh:
It would be nice (and concerning accessibility more than nice) to be able to adjust the font size.
Maybe this can be done together with the #889 Theming?

Simon-Laux:
We should do like 3-5 presets that we know don't break the layout

Simon-Laux:
maybe `.bp3-large`can help a bit?

Simon-Laux:
we could use the css unit `rem` for this (which is relative to font-size of the root element)

Vistaus:
It would still be great to have this. Even on the biggest zoom factor, the text is almost unreadable for me.

strk:
Using CTRL +/- to make fonts bigger/smaller would be great (it's what I use in Firefox, for all the web-based IM clients)

nicodh:
> It would still be great to have this. Even on the biggest zoom factor, the text is almost unreadable for me.

Can you describe your environment? Show a screenshot and let us know what resolution your device has?

We could add higher zoom level, but only if there is a real need for it...

Vistaus:
Hi Nico,

Environment: KDE Plasma 6 (currently 6.3.5) on CachyOS (Arch Linux), on a 4K screen (3840x2160)
I attached two screenshots, one of Delta Chat and one of Dolphin. This is with Delta Chat on the highest zoom level, but as you can see, text is still very small. System font size is the same everywhere (19/20) and in all other apps it's much bigger, and thus much more readable. Font family also doesn't matter: if I choose e.g. Verdana or Ubuntu, it doesn't make any difference whatsoever with regards to the font size that Delta Chat uses.

![Image](https://github.com/user-attachments/assets/30a1542f-d865-4ea0-8d9e-2314f679da5d)
![Image](https://github.com/user-attachments/assets/4b2ec1fd-ccf3-4b85-a610-5d06be6369a0)

Vistaus:
Yes, in Chromium text size is fine. I can also adjust text size there if I want to, unlike in Delta.

nicodh:
Maybe related to this:
https://github.com/electron/electron/issues/10101

a comment says: "I'm having the same issue with KDE5, so it might be plasma specific. "

nicodh:
And maybe this one, which is newer https://discussion.fedoraproject.org/t/fix-ui-scale-bug-electron-apps-too-small/142940

Sadly I can't reproduce it, so it's hard to find the reason...

[...]

PR #1501, opened 2020-03-09
Add zoom factor option in order to adjust font size.

Simon-Laux:
(It's found under `View`->`Zoom Factor`)
Screenshot from 2020-03-09 14-42-59
related to #890

keep in mind that this method is electron specific and not the same as the css zoom property.


Jikstra:
How hard would it be to have this as a css var which we can use in various places to make this non electron specific?

PR #1530, opened 2020-04-03
open help in seperate electron window

### LINKED ISSUE #1477: Help doesnt open on Mac
```
36.8s [E]main/ipc: Can not load help file [Error: EACCES: permission denied, stat '/Users/bb/Library/Application Support/DeltaChat/help-cache/help/en/help.html'] {
  errno: -13,
  code: 'EACCES',
  syscall: 'stat',
  path: '/Users/bb/Library/Application ' +
    'Support/DeltaChat/help-cache/help/en/help.html'
}
```

Simon-Laux:
Edit: Updated screenshot to include back to top button and other recent changes.

This should fix #1477

~~we could/should add a menu with options to zoom to this window in the future.~~ added a default view menu which contains zooming options


r10s:
great, that you push this forward!

- just use "Help" as the title, this is already localized with the string `menu_help`. if needed, `Delta Chat -` can be prefixed
- remove the green-blue bar in the help-view
- use plain white as the background, this new color does not really fit to the rest of the app.

src/main/windows/help.ts
```diff
@@ -0,0 +1,94 @@
...
+  }
+  win = new BrowserWindow({
+    backgroundColor: '#282828',
+    darkTheme: true, // Forces dark theme (GTK+3)
+
+    icon: appIcon(),
+    minHeight: defaults.minHeight,
+    minWidth: defaults.minWidth,
+    show: false,
+    title: appWindowTitle + '-' + app.translate('menu_help'),
```
r10s, line 52: i think, this is mean as sort of hierarchy, as in the main window. in this case, we should use spaces around the minus character, so `DeltaChat - Help` (as in the main window that reads `DeltaChat - me@example.org`)

Simon-Laux:
android has an search https://github.com/deltachat/deltachat-android/pull/1281 we should implement that too.

PR #1532, opened 2020-04-03
add primitive & experimental contact import

Simon-Laux:
(create contacts from json array)
mainly to provide **a** way to import multiple contacts at once into the desktop app. (related to https://support.delta.chat/t/using-importing-contacts-on-desktop-program/368/9)

Better to have the feature this experimental way than not having that possibility at all for the moment.

### DIFF
--- a/src/renderer/experimental.ts
+++ b/src/renderer/experimental.ts
@@ -1,0 +1,35 @@
+import { callDcMethodAsync } from './ipc'
+import { getLogger } from '../shared/logger'
+
+const log = getLogger('renderer/experiments')
+
+class Experimental {
+  help = `
+These functions are highly experimental, use at your own risk.
+- importContacts (contacts:[email,name][]) // for mass importing contacts
+    example: type "exp.importContacts([['email1@example.com', 'Heinz Herlich'],['bea@example.com','Berta Bissig']])"
+    `
+  constructor() {
+    window.exp = this
+  }
+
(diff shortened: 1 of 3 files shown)


Simon-Laux:
this should probably use `dc_add_address_book` instead

src/renderer/experimental.ts
```diff
@@ -0,0 +1,35 @@
+import { callDcMethodAsync } from './ipc'
+import { getLogger } from '../shared/logger'
+
+const log = getLogger('renderer/experiments')
+
+class Experimental {
+  help = `
+These functions are highly experimental, use at your own risk.
+- importContacts (contacts:[email,name][]) // for mass importing contacts
+    example: type "exp.importContacts([['email1@example.com', 'Heinz Herlich'],['bea@example.com','Berta Bissig']])"
```
Jikstra, line 10: Why is this string in the class and not locales/_experimental?

Simon-Laux: because it describes something experimental and needs to be 100% up to date with the code it is describing.

Jikstra: That's fine but still it seems to be weird to me to have a translation in the code itself. I mean we can have them in locales/experimental_en.json and just never put that one into the official language repo?

Simon-Laux: we could also add a markdown file to the docs folder...

PR #1545, opened 2020-04-08
add thumbnails for chat backgrounds in settings screen

### LINKED ISSUE #1467: Finish builder script
Todo points that did not make it into #1464:
- [x] option to enable development build (to allow for react profiling)
- [x] implement watch option for javascript
- [x] script to generate thumbnails for background images (needs to store all file in the same format? [should be no problem as we'll save space when excluding some junk from our package])

Simon-Laux:
This should remove the small lag when opening the settings.
closes #1467


Jikstra:
:*

PR #1550, opened 2020-04-11
Import qr code

nicodh:
Create chat with contact and group are implemented.

Now I need some input:
1. how to distinguish verified from non verified groups or contacts?
2. do we need these types on desktop:
 - domain
 - text
 - address
 - URL
 3. what are the use cases for the finger print scan responses?


r10s:
> 1. how to distinguish verified from non verified groups or contacts?

- if [`dc_check_qr()`](https://c.delta.chat/classdc__context__t.html#a34a865a52127ed2cc8c2f016f085086c) returns DC_QR_ASK_VERIFYCONTACT you already have an unverified contact, contact-id is returned in `dc_lot_t::id`.  
you can then ask the user if they wants to verify the contact,  just use the string `ask_start_chat_with` with the name set to dc_contact_get_name_n_addr(contact-id) -
to start the verification pass 
if the user wants to verify, call [`dc_join_securejoin()`](https://c.delta.chat/classdc__context__t.html#ae49176cbc26d4d40d52de4f5301d1fa7) with the scanned qr-code - on success, you have a verified contact, however, you also get the even DC_EVENT_CONTACT_CHANGED, so there is no need to handle this explicitly

- if `dc_check_qr()` returns DC_QR_ASK_VERIFYGROUP, you should ask the user if they want to join the group, use the string `qrscan_ask_join_group` with the name set to `dc_lot_t::text1` for this purpose.  
the the user wants to, again call `dc_join_securejoin()` - you get all results as an event. you cannot know if the finally created group will be a verified group or not or if the thing succeeds at all. but you will get events as usual for group creation, so, again, nothing special to do here.

> 2. do we need these types on desktop: domain text address URL

i think they can be ignored until there is some option to really scan a qr-code. 

> 3. what are the use cases for the finger print scan responses?

same, you can ignore them for now.

however, DC_QR_ACCOUNT is important to handle, maybe even more important than the verification currently, if this type is returned from `dc_check_qr()` ask the user with the string `qraccount_ask_create_and_login` (domain in dc_lot_t::text1) if they want to create a new account.  
if so, pass the qr-code to [`dc_set_config_from_qr()`](https://c.delta.chat/classdc__context__t.html#ac95d7275f01862fba70169b8cbb937c0)

of course, this DC_QR_ACCOUNT is only useful from within the account creation dialog. however, i would not like to have a "input qr-code" too big and too much in the view of the user for now, it is still a bit hackish, therefore i would suggest a main-menu entry as "Enter qr-code" - this can be available in account-creation as well as later in the normal chat-view.

i hope, i could bring some light to this section :)

if you want to have a look at the flow on other os: for android: https://github.com/deltachat/deltachat-android/blob/master/src/org/thoughtcrime/securesms/qr/QrCodeHandler.java#L44 and for ios: https://github.com/deltachat/deltachat-ios/blob/master/deltachat-ios/Controller/QrViewController.swift#L142

nicodh:
I implemented the adding via secure_join now, but there are several issues:

1. the process is blocking and especially the verified group process needs some time
2. the CHAT_MODIFIED event is triggered several times in case a new group is added and the inviting person is a new contact. It might also happen that another chat is modified because a message arrived, but there is no chance to know if the modified chat is the new chat or am I missing something
3. If I get invited by a new contact to a verified group the new contact is correct added as verified contact, but if I just add a new contact the "verified" status is not added...

Jikstra:
> the process is blocking and especially the verified group process needs some time

Seems like a problem but not something we can fix. Is the core not spawning a thread for this? @r10s 

> the CHAT_MODIFIED event is triggered several times in case a new group is added and the inviting person is a new contact. It might also happen that another chat is modified because a message arrived, but there is no chance to know if the modified chat is the new chat or am I missing something

In general the CHAT_MODIFIED event should come with a chatId so we should know which chat changes.

Jikstra:
Just tried it out, this is how it looks without a camera:
Screenshot_20200424_041036

Besides this also the import qr image button is not shown (i need to scroll down but somehow the dialog is not scrollable so some css is broken)

nicodh:
Scroll etc.is fixed.But I have no idea how to check if the camera is available. 
This https://www.electronjs.org/docs/api/system-preferences#systempreferencesgetmediaaccessstatusmediatype-macos seems only available on MacOS...

Jikstra:
From looking into react-qr-reader (which is quite straight forward, maybe it's nice to reimplement it directly in deltachat-desktop to have full control on the component, include screen capturing, selecting the video device...), i think it's done this way: https://github.com/JodusNodus/react-qr-reader/blob/master/src/getDeviceId.js#L22

hpk42:
On Fri, Apr 24, 2020 at 05:52 -0700, Jikstra wrote:
> From looking into react-qr-reader (which is quite straight forward, maybe it's nice to reimplement it directly in deltachat-desktop to have full control on the component, include screen capturing, selecting the video device...), i think it's done this way: https://github.com/JodusNodus/react-qr-reader/blob/master/src/getDeviceId.js#L22

the qr-processing code is contained with the desktop distribution in any case, right? 
(if not then i think it should be to guard against untrusted/hacked websites).

Jikstra:
> the qr-processing code is contained with the desktop distribution in any case, right?
(if not then i think it should be to guard against untrusted/hacked websites).

Sure. It's not getting downloaded from a CDN or something like this.

PR #1551, opened 2020-04-13
[wip] Overhaul building & use snowpack 

Jikstra:
# What this pr does

Since we introduced typescript, compilation times got a lot worse. @Simon-Laux tried to improve it several iterations and did a good job, but still it took more then 2 minutes on my x230 with an i5 to compile deltachat desktop. For some reason also incremental builds didn't work. Typescript itself was fine most of the time, but the parcel minification & bundling part took a long time, even longer then typescript from my observartions.

I recently stumpled over snowpack, it's a build tool which does nothing else then bundling (sometimes even just copying) the needed modules to `somefolder/web_components`. The module format (think AMD/systemjs/commonjs) it uses is the new ESM, a module format that is supported from nodejs and most browsers. It's nothing special, it's just the import/export syntax we know and use. The huge benefit of snowpack is, you only have to run it when dependencies change, as it only affects them. As this doesn't happen most of the time, but webpack/parcel... still need to bundle it into the "bundle.js", we don't need to care about that. Not only does this make the building a lot faster, it also makes the complete building logic a lot easier and we need to transpile the code less, the whole import/export lines in our .ts code can stay almost the same. 

# Why is building less complex? 

Because we now need nothing more then typescript (and snowpack for dependencies). The typescript compiler is smart enough to be able to output the code that works fine in the browser, no need to babel/transpile anything, typescript already does this for us. The outputted code is a lot easier to read, we don't apply any minification/obfuscation. We have source maps to get stacktraces back to the .ts code, but even reading the "compiled" code is quite straight forward, as it mainly removes the types and that's it.

Also instead of depending on the monolith build script we had before in bin/builder, we can now just go back to have the scripts in package.json. 
- `npm run build` builds everything
- `npm run build:backend` just builds the backend
- `npm run build:frontend` just builds the frontend, including frontend ts, scss & copying static files
- `npm run build:frontend:scss` just builds the scss
- `npm run build:frontend:ts` builds ts
- `npm run build:frontend:static` copies static files
- ... i think you got the idea. 

# What else happend?

I also took the oppertunity to refactor the folder structure. Currently the root folder is quite messy, we have a lot of folders with "source" code and when building we currently output tsc-dist, static and html-dist. We don't know what belongs to the backend, what belongs to the frontend and what is shared without reading the development notes. Therefore i refactored the folder structure:
```
├── bin                       # various helper scripts
├── dist                      # this folder contains all files built during `npm run build` 
│   ├── shared         
│   ├── frontend 
│   └── backend 
├── sources                   # all source files
│   |── test                  # all tests for backend/shared/frontend. Probably we could move them to the various folders.
│   |── backend               # all source files of the backend (just *.ts)
│   ├── shared                # all source files that are shared between backend & frontend (just *.ts)
│   └── frontend              # all source files of the frontend. Stylesheets, images, html, themes, typescript (.ts)...
│     |── themes
│     |── ts      
│     |── scss     
│     └── static  
│         ├── *.html
│         ├── images  
│         └── icons  
```

# Todo:
- [ ] fix electron bundling/dist
  - [ ] test all platforms to work with new code format
- [ ] fix tests
- [ ] i hardcoded the version as requiring the package.json isn't possible. Can get fixed with a script or something.
- [ ] we don't include the themes currently, as json import doesn't work. We can probably fix this.
- [ ] colors for logger got thrown out because it threw weird import errors, should be easy to implement on our own.
- [ ] the popup for maps used react-dom/server which isn't available in the browser (was only possible because we can use require in electron renderer process for whatever reason)
- [ ] the build folder slipped away, bring it back.

# Benchmarks

### snowpack
```
$ time npm run build
...
real    0m40,158s
user    1m19,165s
sys     0m1,915s

```

### old
```
$ time npm run build
...
real    1m43,223s
user    2m20,790s
sys     0m5,597s
```

This means building with snowpack is 63seconds faster for a clean build. Incremental builds are also a lot faster.


nicodh:
Just gave this a try and get the same error as the build script in CI:
Error: Cannot find module 'es-abstract/package.json'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
    at Function.resolve (internal/modules/cjs/helpers.js:33:19)
    at resolveDependencyManifest (/~/deltachat/desktop/node_modules/snowpack/dist-node/index.js:434:33)
    at Object.resolveId (/~/deltachat/desktop/node_modules/snowpack/dist-node/index.js:758:28)
    at Promise.resolve.then (/~/deltachat/desktop/node_modules/rollup/dist/shared/rollup.js:17846:25)

Jikstra:
@nicodh should be fixed :)

nicodh:
Still get an error when running npm i:
0.2s [D]main/dev: Added Extension:  React Developer Tools
(node:8077) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'parse' of undefined
    at getStackTrace (~/deltachat/dev-desktop/deltachat-desktop/dist/shared/logger.js:76:51)
    at Logger.critical (/~/deltachat/dev-desktop/deltachat-desktop/dist/shared/logger.js:103:22)
    at /~/deltachat/dev-desktop/deltachat-desktop/dist/backend/index.ts:63:9
(node:8077) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'parse' of undefined
    at getStackTrace (/~/deltachat/dev-desktop/deltachat-desktop/dist/shared/logger.js:76:51)
    at Logger.critical (/home/ndh/projekte/deltachat/dev-desktop/deltachat-desktop/dist/shared/logger.js:103:22)
    at /~/deltachat/dev-desktop/deltachat-desktop/dist/backend/index.ts:63:9
(node:8077) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:8077) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:8077) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
(node:8077) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.

Jikstra:
> Still get an error when running npm i:
> 0.2s [D]main/dev: Added Extension: React Developer Tools

Do you mean when starting the application? Because this for sure doesn't happen when running npm i.

PR #1556, opened 2020-04-15
improve logging of core info events

### LINKED ISSUE #1555: Provide at least "INFO" logging by default
very often desktop users get back with useless logs because the default is to only provide WARN/ERROR level logs when the ".tsv" is created.  Please provide at least INFO-level messages or maybe even DEBUG simply.  The main reason is helping devs to help users.

Simon-Laux:
This should close #1555

logs always info,warn&error core-events
and logs 'ALL' events only when 'log-debug' is enabled
also --debug enables now --log-debug & --log-to-console

### DIFF
--- a/src/main/deltachat/controller.ts
+++ b/src/main/deltachat/controller.ts
@@ -72,11 +72,7 @@ export default class DeltaChatController extends EventEmitter {
 
     if (data1 === 0) data1 = ''
 
-    if (event === 'DC_EVENT_INFO') {
-      logCoreEv.info(event, data1, data2)
-    } else {
-      logCoreEv.debug(event, data1, data2)
-    }
+    logCoreEv.debug(event, data1, data2)
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/main/deltachat/controller.ts
```diff
@@ -72,7 +72,11 @@ export default class DeltaChatController extends EventEmitter {
 
     if (data1 === 0) data1 = ''
 
-    logCoreEv.debug(event, data1, data2)
+    if (event === 'DC_EVENT_INFO') {
+      logCoreEv.info(event, data1, data2)
+    } else {
+      logCoreEv.debug(event, data1, data2)
+    }
```
hpk42, line 79: not seeing it here but i guess Errors are also logged, right?

Jikstra:
I would prefer if this is added as a dc.on('DC_EVENT_INFO'.... event listener as it's done for all the other events too (including error events).

PR #1563, opened 2020-04-18
Remove pinned chatlist item background and make the pinned icon color…

Jikstra:
… more grey

Screenshot_20200418_161900

Screenshot_20200418_161922


r10s:
what is the reason to remove the background? i find it very nice and useful. only an icon is easy to overlook.

Jikstra:
> what is the reason to remove the background? i find it very nice and useful. only an icon is easy to overlook.

It made problems and i didn't like it too much. But probably it makes sense to have it, no real strong opinion on it.

PR #1570, opened 2020-04-22
"Delta remote" Part 1

Simon-Laux:
Adds typing to callDCAsync calls to improve type-checking.
and make it more practicable by using the Promise variant everywhere.
Also `callDcMethodAsync` was moved to `DeltaBackend.call`.
This is part of the effort to create an documented/typed api where we can replace the backend more easily by a rust-websocket-api to increase speed, get less dependend on electron and allow for other applications such as DeltaChatWeb (use you phone as backend and chat on a bigger screen.), Speedup development for new Clients/ Platforms such as deltachat-kaiOS.

### DIFF
--- a/src/main/deltachat/chat.ts
+++ b/src/main/deltachat/chat.ts
@@ -1,10 +1,11 @@
 import { C } from 'deltachat-node'
 import logger from '../../shared/logger'
 import SplitOut from './splitout'
+import { MessageType } from '../../shared/shared-types'
 
 const log = logger.getLogger('main/deltachat/chat')
 export default class DCChat extends SplitOut {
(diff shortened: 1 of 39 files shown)


Jikstra:
I like the call syntax and having typings for those things is neat! What I don't like is that the function overloading is just incredibly long and not very readable. Is the idea to have this as a first step and move from there to independent methods? I think having a call syntax like `DeltaBackend.chatlist.getChatListItem(id:number)` would be neat too.

Simon-Laux:
>  Is the idea to have this as a first step and move from there to independent methods

yes it's only the first step.

PR #1571, opened 2020-04-23
Make hovering over a selected chatlist item almost unnoticeable

Jikstra:
# After
deltachat-hover-selected-chat-list-item-after

# Before

deltachat-hover-selected-chat-list-item-before

In my opinion it should be almost unnoticable and stay in the back. The element is already selected and flashing background colors is just to busy.


Simon-Laux:
Perfect

PR #1574, opened 2020-04-24
Style cleanup adventures

Simon-Laux:
another tale of exiling BOM and styled-components...


src/renderer/components/message/Message.js
```diff
@@ -36,8 +36,8 @@ const Avatar = (contact, onContactClick) => {
 }
 
 const ContactName = props => {
-  const { email, name, profileName, module, color, onClick } = props
-  const prefix = module || 'module-contact-name'
+  const { email, name, profileName, color, onClick } = props
+  const prefix = 'author'
```
Jikstra, line 40: I think this variable is not needed. Just hardcode them in the classNames.

PR #1576, opened 2020-04-24
Ensure that the QR-data string prefix is upper-case.

pabzm:
This works around the behaviour of some browsers, that lower-case the
prefix when using "Copy Link Location". The core only accepts the string
if the prefix is upper-case, though.

This occurs when trying to use the discourse login bot and copying the "Manual link" via "Copy Link Location". We could work around this problem in the login bot, too. But this might not be the only situation in which the prefix get's lower-cased, I'd suppose. And it doesn't hurt much to work around it in DC Desktop.

### DIFF
--- a/src/renderer/components/dialogs/ImportQrCode.tsx
+++ b/src/renderer/components/dialogs/ImportQrCode.tsx
@@ -44,6 +44,10 @@ export function DeltaDialogImportQrInner({
   const [secureJoinOngoing, setSecureJoinOngoing] = useState(false)
 
   const handleResponse = async (scannedQrCode: string) => {
+    // Work around: Some browsers lower-case the prefix when using "Copy Link
+    // Location", but the core only accepts the string if the prefix is
+    // upper-case.
+    scannedQrCode = scannedQrCode.replace('openpgp4fpr', 'OPENPGP4FPR')
     setQrCode(scannedQrCode)
     const tx = window.translate
     let error = false


r10s:
while as a quick hack, that might be okay, it is maybe better to modify the core so that it accepts both variants. this is probably also an issue for other prefixes as `DCACCOUNT:`

Simon-Laux:
> maybe better to modify the core so that it accepts both variants

I agree that's the way to go in the future.

pabzm:
For the time being I changed the discourse-login-bot to let the user copy the qr-data to their clipboard via Javascript. The "Manual link" isn't useful yet, anyway.

Simon-Laux:
> For the time being I changed the discourse-login-bot to let the user copy the qr-data to their clipboard via Javascript. The "Manual link" isn't useful yet, anyway.

I disagree, It is useful for iOS and for android. also jikstra currently works on implementing it into desktop. #1577

pabzm:
DC Android registers the scheme already? I looked at the Changelog, which doesn't mention this. Neither does the Changelog of DC iOS.

I assumed that DC Desktop would be the first implementation, which isn't released, thus I said "yet".

But of course if the functionality is live already, I'll re-introduce the link.

Jikstra:
The way that the uri scheme is registered is fully lowercased. So `OPENPGP4FPR:` is invalid.
https://metacode.biz/openpgp/openpgp4fpr

pabzm:
I changed the login-bot to show both options (https://github.com/deltachat-bot/deltachat-node-webbot-base/pull/1, the changes are already live).

Please decide on your own if the workaround proposed in this PR should be merged or closed.

Jikstra:
@pabzm had a talk with @r10s and we came to the conclusion that the core should support both. So mainly a thing to fix in the core. As it's not a super urgent thing i would close this workaround pr in favour of a core implementation.

Jikstra:
The fix in core: https://github.com/deltachat/deltachat-core-rust/pull/1434

PR #1577, opened 2020-04-24
Implement deep linking for openpgp4fpr: uri

Jikstra:
This should open the deltachat app when clicking on a `openpgp4fpr:foobar` URI and open the ImportQrCode dialog.

# NOTICE
- Clicking on links only works if you have an installed deltachat version of this pr. There's no other way of registering the uri scheme on the system. I don't know if this will work for appimages or portable builds. @Simon-Laux  any ideas?
- You can also pass openpgp4fpr uris as a commandline argument. For example: "npm start -- "OPENPGP4FPR:....."
- This pr supports also lower cased openpgp4fpr uris, but the core not. I won't do workaround here as this should get fixed in the core (and should be quite easy to do)


Simon-Laux:
for this to work in the appimage we will need to patch a bug in electron builder / use a patched version with this change:
https://github.com/Simon-Laux/electron-builder/commit/b958ebbf489704091a06ad9c781fe048e661e7b9

Jikstra:
For whatever reason i can't reference your patched electron-builder. Let's open a pr and see if they merge it upstream,until then this won't work on appimages. Let's merge and thanks for your work!

PR #1580, opened 2020-04-28
Remove padding around context menu items and change background color on hover to the same as from chat list items

Jikstra:
## before 
context-menu-before

## after
Screenshot_20200429_013253


Simon-Laux:
needs more padding left

r10s:
nice 👍

PR #1582, opened 2020-04-30
[wip] Replace all require with import

Jikstra:
- replace colors dependency with own implementation (it makes problems with import syntax, not sure why)
- the themes aren't getting required anymore but retrieved over ipc
- currently we hardcode the version number in src/shared/application-constants.js

Missing:
- [ ] fix tests


Simon-Laux:
we should hardcode the default (light) theme.

src/shared/constants.ts
```diff
@@ -7,8 +7,6 @@ export const gitHubLicenseUrl = gitHubUrl + '/blob/master/LICENSE'
 // we can only convert this to an import when we find out how we can ignore
 // "'rootDir' is expected to contain all source files."
 // for this line
-const { version } = require('../../package.json')
-
-export const appVersion = version
+export const appVersion = '1.3.0'
```
Simon-Laux, line 10: we should probably instruct our build script to update this value.

Jikstra: Yup

Jikstra:
@Simon-Laux how?

> we should hardcode the default (light) theme.

How without having a script doing it? I think it's fine to get it over ipc. If it should be too slow we can still hardcode it.

Issue #1586, opened 2020-04-30
scan qr-code does not work on mac


r10s:
scanning qr-code does not work on mac, also, there is no error displayed or so.

Simon-Laux:
no error? something about it in the devconsole or in the logfile?
The issue could be a permission error? is the camera part not working or is the scanning not working?

r10s:
the camera is not shown, no image is displayed. i do not get an error or so.

this is how it looks like:

<img width="866" alt="Screen Shot 2020-05-01 at 11 59 21" src="https://user-images.githubusercontent.com/9800740/80797947-647c0180-8ba3-11ea-859f-788a1253371b.png">

wrt log - this entry is either gone or moved so that i cannot find it, i file a separate issue for that.

r10s:
this seems to be fixed by #1589

PR #1589, opened 2020-05-01
Update electron to v6.1.9

Jikstra:


Jikstra:
@r10s can you try out the mac os preview build and tell us if the menu is working again? In electron v6.1.8 some mac/menu specific things got fixed

r10s:
@Jikstra thanks for pushing that forward, however, it does not make a difference on mac yet :/

r10s:
@Jikstra but ....................... 

the qr-code-scanning is working on mac with this version \o/ just did a verification between mac-desktop and ios :) so, this fixes #1586 :)

Jikstra:
@Simon-Laux didn't you fix something for the mac keybindings or something? Could this be a source of the problem?

Simon-Laux:
That code could be the source of the problem if it fails on mac, but shouldn't this throw some-kind of error?
Also I can't explain what the menu has to do with the setting to have no system title bar.

PR #1595, opened 2020-05-03
Replace requires with imports, get rid of colors dependency and retrieve themes over ipc

Jikstra:
Another step into the direction of using snowpack and making the frontend work in normal browsers too


Simon-Laux:
lets redo, rebase this...

src/shared/tsconfig.json
```diff
@@ -8,6 +8,7 @@
         "module": "commonjs",
         "target": "es6",
         "allowJs": true,
+        "esModuleInterop": true
```
Simon-Laux, line 11: what was the reason to enable this?

Simon-Laux:
close in favor of #1666

PR #1596, opened 2020-05-03
Build previews on the same platform and ditch builder script

Jikstra:


Jikstra:
@Simon-Laux seems like the builder scripts fails on windows :/

package.json
```diff
@@ -16,14 +15,28 @@
...
+    "watch": "npm run watch:shared & npm run watch:frontend & npm run watch:backend",
+    "watch:shared": "npm run build:shared -- -w",
+    "watch:backend": "npm run build:backend -- -w",
+    "watch:frontend": "npm run watch:frontend:ts & npm run watch:frontend:scss && npm run watch:frontend:static",
+    "watch:frontend:ts": "npm run watch:frontend:ts:tsc & npm run watch:frontend:ts:parcel",
+    "watch:frontend:ts:tsc": "npm run build:frontend:ts:tsc -- -w --preserveWatchOutput",
+    "watch:frontend:ts:parcel": "npx parcel watch tsc-dist/renderer/main.js --out-dir html-dist --out-file bundle.js --public-url ./ --target browser",
+    "watch:frontend:scss": "npm run build:frontend:scss -- -w",
+    "watch:frontend:static": "npm run build:frontend:static -- -w",
+    "build4production": "NODE_ENV=production npm run build",
```
Simon-Laux, line 37: this should also enable code minification

Jikstra: But isn't this done with the NODE_ENV variable? Don't see any other step in your script.

Simon-Laux: https://github.com/deltachat/deltachat-desktop/blob/master/bin/build/build-js.js#L28-L31

PR #1611, opened 2020-05-06
add a thin seperator line between chatlist items

### LINKED ISSUE #1599: thin separators lines in chatlist
i suggest to add  a thin line below each chat cell

- this improves readability, chats are better separated
- if you have more than one pinned chat, this "grey area" does not look good
- there is a subtile difference between the chatlist and other lists as contacts (these should not get the lines)
- looking around, these lines are used on android/ios - as well as eg. by telegram, also on desktop

btw. we removed the lines also when switching the android-ui that time - and got several report of users, that this minor element is pretty important :)

i just played around, a `.chat-list-item { border-bottom: 1px solid #ddd; }` already improves things significantly imho.

Screen Shot 2020-05-04 at 11 58 17

Simon-Laux:
closes #1599

### DIFF
--- a/scss/chat/_chat-list-item.scss
+++ b/scss/chat/_chat-list-item.scss
@@ -8,6 +8,8 @@
   align-items: center;
   cursor: pointer;
 
+  border-bottom: 1px solid var(--chat-list-item-seperator);
+
   & > .content {
     flex-grow: 1;
     margin-left: 10px;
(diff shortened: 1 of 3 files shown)


Jikstra:
I don't know, but I don't like it too much. @r10s can you pull the preview and maybe tweak the colors? I don't really feel this..

PR #1613, opened 2020-05-08
Mirror video

### LINKED ISSUE #1591: mirror qr-scanner horizontally
it is a bit hard to place the qr-code before the camera as the view is not mirrored - ppl know how a mirror works and can easily correct left and right with that. but with the "real" camera image, one is always first trying to move the smartphone to the wrong direction.

so, it would be great if the video shown during qr-scanning is mirrored horizontally (btw. this is also nearly always the case when using the front-selfie-camera of a smartphone)

nicodh:
resolves #1591


r10s:
great! this will make qr-code scanning 5 times faster :)

PR #1623, opened 2020-05-12
Typescriptify Frontend

Simon-Laux:
- [X] LoginScreen.js
- [X] contact list
- [X] message
- [X] Menu.js

note: rebase mute notification branch after merging this pr (because of Menu.js)

### DIFF
--- a/src/renderer/components/helpers/ChatMethods.ts
+++ b/src/renderer/components/helpers/ChatMethods.ts
@@ -5,6 +5,7 @@ import {
   ChatListItemType,
   JsonContact,
   FullChat,
+  DCContact,
 } from '../../../shared/shared-types'
 import { C } from 'deltachat-node/dist/constants'
 
@@ -94,10 +95,9 @@ export function openMapDialog(
 
 export async function openViewProfileDialog(
   screenContext: unwrapContext<typeof ScreenContext>,
-  contact: number | JsonContact
+  contact_id: number
 ) {
-  if (typeof contact === 'number' && Number.isInteger(contact)) {
-    contact = await DeltaBackend.call('contacts.getContact', contact)
-  }
-  screenContext.openDialog('ViewProfile', { contact })
+  screenContext.openDialog('ViewProfile', {
+    contact: await DeltaBackend.call('contacts.getContact', contact_id),
+  })
 }
--- a/src/renderer/components/message/Message.tsx
+++ b/src/renderer/components/message/Message.tsx
@@ -7,8 +7,15 @@ import MessageMetaData from './MessageMetaData'
 
 import { ContextMenu, ContextMenuTrigger, MenuItem } from 'react-contextmenu'
 import Attachment from '../attachment/messageAttachment'
+import { MessageType, DCContact } from '../../../shared/shared-types'
+import { attachment } from '../attachment/Attachment'
 
-const Avatar = (contact, onContactClick) => {
+type msgStatus = 'error' | 'sending' | 'draft' | 'delivered' | 'read' | ''
+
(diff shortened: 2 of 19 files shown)


src/renderer/components/helpers/ChatMethods.ts
```diff
@@ -94,10 +95,9 @@ export function openMapDialog(
 
 export async function openViewProfileDialog(
   screenContext: unwrapContext<typeof ScreenContext>,
-  contact: number | JsonContact
+  contact_id: number
```
Jikstra, line 98: Are you sure that we never call it with just a contact?

src/renderer/components/message/Message.tsx
```diff
@@ -26,7 +33,7 @@ const Avatar = (contact, onContactClick) => {
       ? String.fromCodePoint(codepoint).toUpperCase()
       : '#'
     return (
-      <div className='author-avatar default' alt={alt} onClick={onClick}>
+      <div className='author-avatar default' aria-label={alt} onClick={onClick}>
```
Jikstra, line 36: Shouldn't we have alt + aria-label? I guess the idea of the alt war that you can hover over it and see the name. Maybe not needed here ux wise?

src/renderer/components/message/MessageList.tsx
```diff
@@ -33,7 +45,7 @@ export default function MessageList({
     chatStoreDispatch,
   ] = useChatStore()
   const messageListRef = useRef(null)
-  const lastKnownScrollHeight = useRef([null, null])
+  const lastKnownScrollHeight = useRef<number>(null)
```
Jikstra, line 48: Isn't this changing it from an array with two null elements to a single null ?

src/renderer/components/message/MessageWrapper.tsx
```diff
@@ -18,30 +28,28 @@ export const render = props => {
...
     const tx = window.translate
     const screenContext = useContext(ScreenContext)
     const { openDialog } = screenContext
 
-    message.onForward = () => openDialog('ForwardMessage', { message })
-
-    const conversationType = GROUP_TYPES.includes(chat.type)
+    const conversationType: 'group' | 'direct' = GROUP_TYPES.includes(chat.type)
       ? 'group'
       : 'direct'
```
Jikstra, line 41: This is one of the things i dont like about typescript, you have the types there twice and it's getting really verbose. I sometimes have the feeling that I need to duplicate type annotations quite a lot.

src/renderer/components/helpers/ChatMethods.ts
Simon-Laux: yes I changed it everywhere

src/renderer/components/message/Message.tsx
Simon-Laux: alt is not a valid property on div (in react), so we should probably remove the thing here.

src/renderer/components/message/MessageList.tsx
Simon-Laux: I'm not sure what the array should represent here other than an typescript error.

src/renderer/components/message/MessageWrapper.tsx
Simon-Laux: the typescript compile is not the smartest... But if it were, build times would probably be longer.

src/renderer/components/message/MessageList.tsx
Jikstra: Yup we initialized it as an array but set it to a single number ^^ Typical js bug that's not possible with typescript.

PR #1624, opened 2020-05-12
Login migration upgrade guides and error handling

Simon-Laux:
- Add information about changed behavior. Also it shouldn't crash anymore if the migration failed, instead it opens an error dialog
- show error dialogs on uncaught errors, in hope to get more less technical people to report crashes

### DIFF
--- a/src/main/logins.ts
+++ b/src/main/logins.ts
@@ -38,13 +38,13 @@ async function migrate(dir: string) {
         await fs.move(join(dir, basename(account.path)), join(dir, newFolder))
       } catch (error) {
         log.error(
-          "Moving failed, make sure you don't have multiple account folders for the same account!",
+          "Moving failed, make sure you don't have multiple account folders for the same e-mail address!",
           error
         )
         dialog.showErrorBox(
           'Account migration',
           `Migration of ${account.addr} failed,\n` +
-            "make sure you don't have multiple account folders for the same account!\n" +
+            "Please make sure that you don't have multiple account folders for the same e-mail address!\n" +
             'See the logfile for details'
         )
         continue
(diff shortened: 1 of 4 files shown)


src/main/logins.ts
```diff
@@ -33,8 +34,23 @@ async function migrate(dir: string) {
...
+        await fs.move(join(dir, basename(account.path)), join(dir, newFolder))
+      } catch (error) {
+        log.error(
+          "Moving failed, make sure you don't have multiple account folders for the same account!",
+          error
+        )
+        dialog.showErrorBox(
+          'Account migration',
+          `Migration of ${account.addr} failed,\n` +
+            "make sure you don't have multiple account folders for the same account!\n" +
```
Jikstra, line 47: make sure you don't have the same e-mail adress multiple times

Simon-Laux: I see your point but your proposed solution misses the point with the account folders

Jikstra: `Please make sure that you don't have multiple account folders for the same e-mail address`

Simon-Laux, line 47:
```suggestion
            "Please make sure that you don't have multiple account folders for the same e-mail address!\n" +
```

src/main/logins.ts
```diff
@@ -33,8 +34,23 @@ async function migrate(dir: string) {
         'accounts',
         escapeEmailForAccountFolder(account.addr)
       )
-      await fs.move(join(dir, basename(account.path)), join(dir, newFolder))
-      // Backwards compatibility
+      try {
+        await fs.move(join(dir, basename(account.path)), join(dir, newFolder))
+      } catch (error) {
+        log.error(
+          "Moving failed, make sure you don't have multiple account folders for the same account!",
```
Simon-Laux, line 41:
```suggestion
          "Moving failed, make sure you don't have multiple account folders for the same e-mail address!",
```

PR #1631, opened 2020-05-12
Merge 1.3.3

Jikstra:
I don't really know if this makes sense, maybe we should just merge the bugfix?


Simon-Laux:
I merged the bugfix separately. I'd say we remove tje bugfix from this pr and jsut merge the changelog/version number increase.

PR #1635, opened 2020-05-13
Refactor Autocrypt and it's settings

Jikstra:
- Factor out encryption settings into dialogs/Settings-Encryption.tsx
- Move SendAutocryptSetupMessage dialog into Settings-Encryption because
it's only opened from there
- Refactor SendAutocryptSetupMessage dialog to use react hooks
- Refactor EnterAutocryptSetupMessage dialog to use react hooks
- Get rid of old style rpc and use DeltaBackend.call instead for
autocrypt calls


Simon-Laux:
> Get rid of old style rpc and use DeltaBackend.call instead for
autocrypt calls

great, I had this on my todo list, too.

Jikstra:
Dark mode is not on my priority list and won't work on it for the moment. But feel free to fix it, but please not in this pr :)

Jikstra:
This would need a new -node version that will break the 1.3 release series. Or shall i deprecate the old api and add another method with a different name? @Simon-Laux what do you think?

Simon-Laux:
@Jikstra Lets deprecate the old method and add a new one with another name.

PR #1639, opened 2020-05-14
🚧 Offline toast

Simon-Laux:
Move the network error into a persistent toast message:

TODO:
- [ ] use language strings instead of hard-coded text
- [ ] move to seperate component  `<OfflineToast />`

### DIFF
--- a/src/main/deltachat/controller.ts
+++ b/src/main/deltachat/controller.ts
@@ -164,9 +164,6 @@ export default class DeltaChatController extends EventEmitter {
         logCoreEvent.warn(event, data1, data2)
       } else if (event === 'DC_EVENT_INFO') {
         logCoreEvent.info(event, data1, data2)
-      } else if (_event === C.DC_EVENT_ERROR_NETWORK) {
-        this.networkError = true
-        this.emit('update-network-status')
       } else if (event.startsWith('DC_EVENT_ERROR')) {
         this.emit('error', event, data1, data2)
         logCoreEvent.error(event, data1, data2)
@@ -174,22 +171,6 @@ export default class DeltaChatController extends EventEmitter {
         // in debug mode log all core events
         logCoreEvent.debug(event, data1, data2)
       }
-
-      // ugly hack to find out when the user goes online again:
-      if (
-        this.networkError &&
-        [
-          C.DC_EVENT_SMTP_CONNECTED,
-          C.DC_EVENT_IMAP_CONNECTED,
-          C.DC_EVENT_INCOMING_MSG,
-          C.DC_EVENT_MSG_DELIVERED,
-          C.DC_EVENT_IMAP_MESSAGE_MOVED,
-        ].includes(_event)
-      ) {
-        this.networkError = false
-        this.emit('update-network-status')
-      }
-
       this.sendToRenderer(event, [data1, data2])
     })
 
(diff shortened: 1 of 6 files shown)


Jikstra:
Mamma, someone was really motivated tonight. Have this thing on my list since months ^^

src/main/deltachat/controller.ts
```diff
@@ -159,6 +164,9 @@ export default class DeltaChatController extends EventEmitter {
         logCoreEvent.warn(event, data1, data2)
       } else if (event === 'DC_EVENT_INFO') {
         logCoreEvent.info(event, data1, data2)
+      } else if (_event === C.DC_EVENT_ERROR_NETWORK) {
+        this.networkError = true
+        this.emit('update-network-status')
```
Jikstra, line 169: why not do the main.send already here?

src/renderer/ScreenController.tsx
```diff
@@ -60,6 +69,7 @@ export default class ScreenController extends Component {
 
   componentDidMount() {
     ipcRenderer.on('error', this.onError)
+    ipcRenderer.on('update-network-status', this.onNetworkChange)
```
Jikstra, line 72: You could just listen for `DC_EVENT_ERROR_NETWORK` here

src/renderer/ScreenController.tsx
```diff
@@ -114,6 +129,24 @@ export default class ScreenController extends Component {
...
+                  15000
+                )
+                setTimeout(() => DeltaBackend.call('context.maybeNetwork'), 100)
+              }}
+              className={this.state.tryConnectCooldown ? '' : 'disabled'}
+            >
+              Try to connect now
+            </div>
+          </div>
+        )}
```
Jikstra, line 149: Also I would like to have this code in a seperate component.
`<OfflineToast />`

This way the component could keep all the state about being online or offline and just listen to the DC_NETWORK_ERROR and if it's offline, also to the SMTP_CONNECTED... things. Otherwise the ScreenController will grow to a beast the same way Settings did.

Simon-Laux:
This pr is not about bringing in the perfect event architecture, It just uses what is currently there and this code will certainly change with our work on the backend.
I agree with the separate component, though.

Jikstra:
> This pr is not about bringing in the perfect event architecture, It just uses what is currently there and this code will certainly change with our work on the backend.
> I agree with the separate component, though.

You don't need to change anything at the event architecture, all core event should get passed to the frontend anyways. Correct me if i'm wrong.

Simon-Laux:
Ok you convinced me let's listen for the events in the `OfflineToast`component

src/renderer/ScreenController.tsx
Simon-Laux: how would I know when it should be hidden again, then?

Simon-Laux:
Though it would be better if the core had an api to get its connection error. So we wouldn't need to listen for events and guess the state.

Simon-Laux:
Ok, now we listen in the front-end for events. But this also means we don't catch an initial offline state when starting the application into an existing account. So this method is less reliable.

Jikstra:
> Ok, now we listen in the front-end for events. But this also means we don't catch an initial offline state when starting the application into an existing account. So this method is less reliable. @Jikstra

Yup that's bad. I just reworked this pr and added the backend handling again, implemented listening to browser events and now it works quite well.

Simon-Laux:
superseded by  #1658

PR #1640, opened 2020-05-14
Drop-send multiple files➬ask for all files at once

### LINKED ISSUE #1622: Drop-send multiple files - ask for all files at once
A workaround could be to show only one dialog that lists the files and asks if you are sure to send them all.

_Originally posted by @Simon-Laux in https://github.com/deltachat/deltachat-desktop/issues/1552#issuecomment-614791607_

I believe this behavior would be better than the current showing of multiple dialogs.
Alone the reason that the backdrop adds up to much in the current version with multiple dialogs adds up waaaay too much.

Simon-Laux:
fixes #1622
new-file-drop-dialog

### DIFF
--- a/src/renderer/components/message/MessageListAndComposer.tsx
+++ b/src/renderer/components/message/MessageListAndComposer.tsx
@@ -31,30 +31,54 @@ export default function MessageListAndComposer({
   )
 
   const onDrop = (e: React.DragEvent<any>) => {
-    const files = (e.target as any).files || e.dataTransfer.files
     e.preventDefault()
     e.stopPropagation()
-    const tx = window.translate
-
-    // TODO maybe add a clause here for windows because that uses backslash instead of slash
-    const forbiddenPathRegEx = /DeltaChat\/.+?\.sqlite-blobs\//gi
-    for (let i = 0; i < files.length; i++) {
-      const { path, name } = files[i]
-      if (!forbiddenPathRegEx.test(path.replace('\\', '/'))) {
-        openDialog('ConfirmationDialog', {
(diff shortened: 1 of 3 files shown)


src/renderer/components/message/MessageListAndComposer.tsx
```diff
@@ -31,30 +31,54 @@ export default function MessageListAndComposer({
...
+        // if that is possible without a backend call to check wheter the file exists,
+        // maybe some browser api like FileReader could help
+        if (!forbiddenPathRegEx.test(path.replace('\\', '/'))) {
+          sanitizedFileList.push({ path, name })
+        } else {
+          log.warn(
+            'Prevented a file from being send again while dragging it out',
+            name
+          )
+        }
```
Jikstra, line 53: I think for something like this functional patterns come in very handy.

Just some pseudo code

```
let fileList: FileList = (e.target as any).files || e.dataTransfer.files
const forbiddenPathRegEx = /DeltaChat\/.+?\.sqlite-blobs\//gi
const isForbidden = ({path}) => !forbiddenPathRegEx.test(path.replace('\\', '/'))
const sanitizedFileList = fileList.filter(isForbidden)

if (sanitizedFileList.length < fileList.length) {
  log.warn('Prevented some files from being sent')
}
```

Jikstra:
I think functional patterns would be nicer but besides this it looks good.

src/renderer/components/message/MessageListAndComposer.tsx
Simon-Laux: nah, we can still change this later.. 
PS: I used a feature of `let` here, scoped usage. might have interesting usages.

PR #1642, opened 2020-05-14
Refactor DialogController a bit and add better types

Jikstra:


src/renderer/components/dialogs/DialogController.tsx
```diff
@@ -66,27 +65,36 @@ export type DialogProps = {
...
 
     this.state = { dialogs: {} }
-    this.close = this.close.bind(this)
+    this.openDialog = this.openDialog.bind(this)
+    this.closeDialog = this.closeDialog.bind(this)
   }
 
-  open(fnc: any, props?: todo /* infer from component */) {
+  openDialog(
+    fnc: DialogId | ((props: DialogProps) => React.ReactElement),
```
Simon-Laux, line 88:
```suggestion
    fnc: DialogId | ((props: DialogProps & DialogAdditionProps) => React.ReactElement),
```
probably

Simon-Laux: I think the infer thing makes more sense to do when we converted all dialogs to functional components.

PR #1650, opened 2020-05-16
improve logfiles

### LINKED ISSUE #1645: Just log to a file, not this weird csv
I repeatedly run into trouble looking at desktop logs -- the menu entry "open current logfile" brings me into libreoffice which complains alfready that the log file is too big, too many cells or something. 
Even if i manage to open it, it's hard to look at it, as there are so many columns. 

Please let's just write out a simple text logfile like android/ios do as well.  Then Notepad does the job, or vim, or emacs etc. There also is an idea to write a tool that analyzes logfiles (number of disconnects/reconnects/sent-attempts etc.) -- so having a similar output as android or the python bindings would be useful (maybe we can aim to really use a common log format but that can come stepwise as well).

Simon-Laux:
- save log-files as .log so that they aren't opened as tables by default
- padding for channels so that they are mostly aligned now
- remove `"` for timestamp, channel and level

closes #1645

### DIFF
--- a/src/main/log-handler.ts
+++ b/src/main/log-handler.ts
@@ -15,7 +15,7 @@ function logName() {
     `${pad(d.getHours())}-`,
     `${pad(d.getMinutes())}-`,
     `${pad(d.getSeconds())}`,
-    '.log.tsv',
+    '.log',
   ].join('')
   return join(dir, fileName)
 }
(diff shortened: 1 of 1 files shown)


src/main/log-handler.ts
```diff
@@ -40,10 +40,10 @@ export function createLogHandler() {
       ...args: any[]
     ) => {
       const timestamp = new Date().toISOString()
-      let line = [timestamp, channel, level]
-      line = line
-        .concat([stacktrace, ...args])
-        .map(value => JSON.stringify(value))
+      let line = [timestamp, fillString(channel, 22), level]
```
Jikstra, line 43: Why do we need to fill this? Doesn't it look weird?

Simon-Laux: because it makes it more readable in an text editor.

PR #1655, opened 2020-05-17
[1.3] Backport fix error handling

Jikstra:
Partly backports https://github.com/deltachat/deltachat-desktop/pull/1643


Jikstra:
@Simon-Laux this would have been more then enough... Not refactoring login methods and adding a CatchError2Event method (which is nice in a way, but who knows what problems it adds on top). Because of those commits I also couldn't fully backport the pr to the 1.3 branch. I already reverted the merge of the pr because i have overseen those changes and thought you just added this one line of code in the screencontroller. 

BUT I like the idea of getting rid of the error events and improving the async/await stuff of the DeltaChat.call thingy. That was the reason i finally reverted my revert on development branch. So no fault or anything, was just suprised of finding the Error2Event method.

PR #1658, opened 2020-05-17
Offline Toast

Jikstra:


src/renderer/ScreenController.tsx
```diff
@@ -76,7 +76,6 @@ export default class ScreenController extends Component {
     ipcRenderer.removeListener('error', this.onError)
     ipcRenderer.removeListener('DC_EVENT_ERROR', this.onError)
     ipcRenderer.removeListener('DC_EVENT_LOGIN_FAILED', this.onError)
-    ipcRenderer.removeListener('DC_EVENT_ERROR_NETWORK', this.onError)
```
Jikstra, line 79: This is in general fine when we're logged in but a problem if we're not :/

src/main/deltachat/controller.ts
```diff
@@ -166,6 +172,29 @@ export default class DeltaChatController extends EventEmitter {
         logCoreEvent.debug(event, data1, data2)
       }
 
+      // Network Status
+      if (event === 'DC_EVENT_ERROR_NETWORK') {
+        this.networkStatus = false
+        this.networkStatusMessage = data1 + data2
```
Simon-Laux, line 178: https://github.com/deltachat/deltachat-core-rust/blob/master/src/events.rs#L81-L95

either data 1 or data 2 seems unused, but I didn't look fully into how the core works here.

src/renderer/components/OfflineToast.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const onTryReconnectClick = () => {
+    setTryConnectCooldown(false)
+    setTimeout(() => setTryConnectCooldown(true), 15000)
+    setTimeout(() => DeltaBackend.call('context.maybeNetwork'), 100)
+  }
+
+  return (
+    networkStatus === false && (
+      <div className='OfflineToast'>
+        <a title={networkStatusMessage}>Offline</a>
```
Simon-Laux, line 61: Language strings are still missing here (translations)

src/renderer/components/OfflineToast.tsx
```diff
@@ -0,0 +1,71 @@
...
+        <div
+          className={tryConnectCooldown ? '' : 'disabled'}
+          onClick={onTryReconnectClick}
+        >
+          Try to connect now
```
Simon-Laux, line 66: todo translations

PR #1659, opened 2020-05-17
Implement searching

Jikstra:
# ToDo

- [x] fix archive link
- [x] sometimes something isn't loaded (investigate more?)
- [x] highlight query in messagetext


Jikstra:
@Simon-Laux what's missing? I would be fine with only having contacts show in search for now. That would be already super helpful.

Simon-Laux:
1. fix the lazy loading on scroll for the chats which is broken now.
2. for the message search its only missing to fetch and show the result objects, but that is only doable if we implement some lazy loading like we did for the chats. (otherwise the app will hang, an workaround solution would be to limit shown results)
3. rebase as github lists a conflicting file

Jikstra:
Let's not implement message search in this pr. It's not very helpful without the jump to message thingy and just blocks this pr.

src/renderer/components/chat/ChatList.tsx
```diff
@@ -68,18 +74,32 @@ export default function ChatList(props: {
     selectChat(chatId)
   }
 
+  const isSearchActive = queryStr !== ''
+
   const renderAddContactIfNeeded = () => {
-    if (queryStr === '' || chatListIds.length > 0) return null
+    if (chatListIds.length > 0) return null
```
Jikstra, line 80: @Simon-Laux  did you miss a isSearchActive here?

Jikstra:
Uff implementing lazy loading wont be that easy with the current way of doing it...

src/renderer/components/chat/ChatList.tsx
Simon-Laux: nope it is checked bellow (L116), no reason to check it twice.

src/renderer/components/chat/ChatList.tsx
```diff
@@ -67,34 +136,164 @@ export default function ChatList(props: {
...
+                    width={width}
+                    height={
+                      // take remaining space
+                      messagesHeight(height)
+                    }
+                  >
+                    {({ index, key, style }) => {
+                      const msrId = messageResultIds[index]
+                      return (
+                        <div style={style} key={key}>
```
Jikstra, line 264: Can we factor this out in an own function component? The function is already huuuge

Jikstra:
Awesome! Seems like you got how this virtualized stuff works. Never got it, maybe you can explain it to me one day or i keep looking at this code ^^

Jikstra:
Bugs I found so far:
- [ ] clicking on a contact with whom we don't have a chat yet doesn't create a new one but causes some bugs in the navbar:
Screenshot_20200521_133647
- [ ] Clicking on a contact or chat doesn't clear the search

Not really a bug but having three scrollable divs in the search feels weird to me. Can we change it or is it for a technical reason?

Simon-Laux:
> don't show chats or contacts seperator if it didn't find any chats or contacts

That's personal preference, I wouldn't call it a bug.

> Not really a bug but having three scrollable divs in the search feels weird to me. Can we change it or is it for a technical reason?

Both technical and functional reason. I didn't got the thing to lazy-load in one big scrollbar. And if we had one big scrollbar we'd need some things to give the user a sense of navigation where the user is currently and buttons to jump to a different result type. So its might be actually more practical UX wise. (even though I would prefer something tabbed or a whole screen search to use the space we have on desktop)

src/renderer/components/chat/ChatList.tsx
Simon-Laux: I don't think that makes much sense here.

Simon-Laux:
> Clicking on a contact or chat doesn't clear the search

Should it clear the search? @r10s what do you think?

Simon-Laux:
> I think the useLazyChatListItems function is not needed anymore?

ForwardMessage and ViewProfile Still use it.

Jikstra:
> Both technical and functional reason. I didn't got the thing to lazy-load in one big scrollbar. And if we had one big scrollbar we'd need some things to give the user a sense of navigation where the user is currently and buttons to jump to a different result type. So its might be actually more practical UX wise. (even though I would prefer something tabbed or a whole screen search to use the space we have on desktop)

Hmm my idea would be to have them fixed on the lower end of the chatlist. If you click on it, you jump to it. Probably not too easy to implement, but that would be quite nice and better then what telegram has (they just have one big search bar). Still I would prefer to not show the contacts/messages divider if there are no results for this.

src/main/deltachat/messagelist.ts
```diff
@@ -138,6 +141,34 @@ export default class DCMessageList extends SplitOut {
...
+      authorProfileImage: author.getProfileImage(),
+      author_name: author.getDisplayName(),
+      author_color: integerToHexColor(author.getColor()),
+      chat_name: chat.isSingle() ? null : chat.getName(),
+      message: message.getText(),
+      timestamp: message.getTimestamp(),
+    }
+  }
+
+  msgIds2SearchResultItems(ids: number[]) {
```
Jikstra, line 165: I propose `getMessageSearchResultItems(ids: number[])`

Jikstra:
Puuh a lot of code but lets merge.

PR #1666, opened 2020-05-18
Require to Imports

Simon-Laux:
this is the redo of #1595

This pr features additionally;
- convert logger to ts
- build script to generate build info (version and git commit/ref)
- logger:  disable js stack-trace for core events (wasn't useful and only spamming the log output)


src/renderer/components/dialogs/About.tsx
```diff
@@ -86,7 +83,7 @@ export default function About(props: { isOpen: boolean; onClose: () => void }) {
         <Card>
           <p
             style={{ color: 'grey', userSelect: 'all' }}
-          >{`Version ${appVersion}`}</p>
+          >{`Version ${VERSION} (git: ${GIT_REF})`}</p>
```
Jikstra, line 86: <3 Awesome!

PR #1667, opened 2020-05-19
Improve usability of dark theme

Simon-Laux:
tweaks theming with an strong focus on making the dark theme usable (avoid black text on dark background for example)

also adds information about css var naming to the docs (prefer snake-case from now on, because electron will give auto completion for css-vars with snake-case which makes it more fun to play with in the dev-console)

### DIFF
--- a/docs/STYLES.md
+++ b/docs/STYLES.md
@@ -28,5 +28,11 @@ When using generic classnames like `date`, `avatar`, `username`, `error` make su
 
 ```
 
-This document is not final, feel free to ask questions and discuss this with us.
+## How do we do colors and theming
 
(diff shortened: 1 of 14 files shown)


Simon-Laux:
maybe we should re-add the box shadow inset thingie from `.delta-dialog-content .delta-dialog-content-text-seperator` ?

Jikstra:
> maybe we should re-add the box shadow inset thingie from .delta-dialog-content .delta-dialog-content-text-seperator ?

Hmm yeah maybe. Not really sure what it changes, can you explain further?

Simon-Laux:
> Hmm yeah maybe. Not really sure what it changes, can you explain further?

look at the pictures in the last post, the one bellow is before the one above is after.

PR #1672, opened 2020-05-21
send a suggestion to update after 90 days

Simon-Laux:
This reminds people to update without requiring any server contact, it assumes that our release rate is smaller than 90days.

This feature is still being discussed in the forum and the general plan is to eventually implement it in the core instead, because this is useful for all platforms.

Forum Topic: https://support.delta.chat/t/update-availible-message-after-x-day/909

### DIFF
--- a/src/main/deltachat/controller.ts
+++ b/src/main/deltachat/controller.ts
@@ -21,6 +21,9 @@ import { ExtendedAppMainProcess } from '../types'
 import Extras from './extras'
 import { EventId2EventName as eventStrings } from 'deltachat-node/dist/constants'
 
+import { VERSION, BUILD_TIMESTAMP } from '../../shared/build-info'
+import { Timespans, DAYS_UNTIL_UPDATE_SUGGESTION } from '../../shared/constants'
+
 const app = rawApp as ExtendedAppMainProcess
 const log = getLogger('main/deltachat')
 const logCoreEvent = getLogger('core/event')
@@ -50,6 +53,11 @@ export default class DeltaChatController extends EventEmitter {
   constructor(public cwd: string) {
     super()
     this._resetState()
+    setInterval(
+      // If the dc is allways on
+      this.hintUpdateIfNessesary.bind(this),
+      Timespans.ONE_DAY_IN_SECONDS * 1000
+    )
   }
 
   readonly autocrypt = new DCAutocrypt(this)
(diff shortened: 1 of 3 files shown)


r10s:
if we want to go that way, adding that to the core sounds useful. if the text goes through the stock-translation dance as usual, it is also customizable by the implementations. sounds pretty simple. but, as mentioned, i would prefer a larger timespan :)

having that in that core, we can also refine the thing later, eg. by checking X-Mailer-Headers of incoming message

src/main/deltachat/controller.ts
```diff
@@ -50,6 +53,11 @@ export default class DeltaChatController extends EventEmitter {
   constructor(public cwd: string) {
     super()
     this._resetState()
+    setInterval(
+      // If the dc is allways on
+      this.hintUpdateIfNessesary.bind(this),
+      Timespans.ONE_DAY_IN_SECONDS * 1000
```
Jikstra, line 59: So this triggers after 1000 days? Not sure if this is helpful ^^

src/main/deltachat/controller.ts
```diff
@@ -339,6 +347,20 @@ export default class DeltaChatController extends EventEmitter {
     return DeltaChatNode.getProviderFromEmail(email)
   }
 
+  hintUpdateIfNessesary() {
+    if (
+      this._dc &&
+      Date.now() >
+        Timespans.ONE_DAY_IN_SECONDS * DAYS_UNTIL_UPDATE_SUGGESTION * 1000 +
+          BUILD_TIMESTAMP
```
Jikstra, line 355: Is the BUILD_TIMESTAMP in ms?

r10s:
sorry for the general statement above, i thought i was commenting an issue not a pr :)

src/main/deltachat/controller.ts
```diff
@@ -339,6 +347,20 @@ export default class DeltaChatController extends EventEmitter {
...
+    ) {
+      this._dc.addDeviceMessage(
+        `update-suggestion-${VERSION}`,
```
Jikstra, line 358: The general code but also this allows us to only send one notification per version that the client might be outdated. Not sure if we want to send another notification after 180 days? ^^

Simon-Laux: not now, if you want you can mention it in the forum post though.

src/main/deltachat/controller.ts
Simon-Laux: yes its produced by Date.now()

PR #1675, opened 2020-05-21
Overhaul build system

Jikstra:
- Split builder monolith in smaller scripts
- Use package.json as glue running the different build steps

Missing:
- [x] fix building translations
- [x] fix watch
- [x] remove builder


bin/build-frontend-static.js
```diff
@@ -0,0 +1,19 @@
...
+
+async function main(watch=false) {
+  const copy = async () => {
+    await fs.copy('./static/', './html-dist/')
+    console.log('- copied files from static folder')
+  }
+  await copy()
+
+  if (watch) {
+    // (whats the usecase of this? live update after pulling in translations??)
```
Simon-Laux, line 12: there aren't any translations in the static folder so I guess this comment doesn't belong here.

package.json
```diff
@@ -16,11 +16,15 @@
...
-    "build": "node ./bin/build/build-info.js && npm run build-themes && npm run build-main && npm run builder",
-    "build-main": "tsc --build src/main",
-    "builder": "node ./bin/build/builder.js",
-    "builder-help": "npm run builder -- --help",
-    "build-scss": "npm run builder -- --styles",
+    "clean": "rm -rf ./tsc-dist ./html-dist ./dist",
+    "build": "npm run build:shared && npm run build:backend && npm run build:frontend",
+    "build:backend": "npm run build:backend:ts",
+    "build:backend:ts": "tsc --build src/main",
+    "build:shared": "node ./bin/build-shared-version-info.js",
```
Simon-Laux, line 23: `build:shared` naming is a bit off here, don't ya think? it's more about stamping the build with build-info before building.

Jikstra: I think it's fine. It's about building something for the shared source folder and it's used by backend & frontend. I think that's something related to shared.

bin/build-frontend-static.js
Jikstra: Don't think that I wrote that comment. But can remove it for sure.

package.json
```diff
@@ -16,13 +16,26 @@
...
+    "build:backend": "tsc --build src/main",
+    "build:frontend": "npm run build:frontend:ts && npm run build:frontend:themes && npm run build:frontend:static && npm run build:frontend:scss",
+    "build:frontend:ts": "node ./bin/build-frontend-ts.js",
+    "build:frontend:themes": "node ./bin/build-frontend-themes.js",
+    "build:frontend:static": "node ./bin/copy.js ./static ./html-dist",
+    "build:frontend:scss": "npx node-sass ./scss/manifest.scss ./html-dist/main.css --output-style compressed",
+    "watch": "npm run watch:shared & npm run watch:backend & npm run watch:frontend",
+    "watch:shared": "npm run watch:shared:translations",
+    "watch:shared:translations": "npm run build:shared:translations -- -w",
+    "watch:backend": "npm run build:backend -- -w --preserveWatchOutput",
```
Simon-Laux, line 33: please remove the watching of the backend

Simon-Laux:
the fallback theme is missing, we can fix it in two ways: either build the fallback theme or just reference the light theme instead in the index html.

PR #1677, opened 2020-05-24
Do not replace data1 with "" when it is 0

link2xt:
There are events that store numbers in data1


link2xt:
See `DC_EVENT_MSG_DELETE_TIMEOUT_CHANGED` in https://github.com/deltachat/deltachat-core-rust/pull/1442, it sets `data1=0` to indicate that timeout should be cancelled.

PR #1682, opened 2020-05-26
WIP: Burner account

nicodh:


src/renderer/components/helpers/OpenPGP4FPRUrl.ts
```diff
@@ -0,0 +1,126 @@
+import { DeltaBackend } from '../../delta-remote'
```
Simon-Laux, line 1: Naming of this file is a bit strange: `OpenPGP4FPRUrl.ts` as the name only refers to a part of this file

nicodh:
Ok, invalid burner qr code should be handled. I will do that...

PR #1685, opened 2020-05-27
🚧 Virtualize chatview

Simon-Laux:
todo:
- [ ] implement to listen for remaining change messagelist events (see comments in https://github.com/deltachat/deltachat-desktop/blob/virtualize-chatview/src/renderer/components/message/messageListHook.ts#L63-L69)
- [X] jump to messages (still a bit buggy but that's probably due to the problems below)

problems to solve:
- [ ] huge messages result in jumping up (not always though)
	probably https://github.com/bvaughn/react-virtualized/issues/610
- [ ] content doesn't start at bottom
- [X] images
- [x] context menu missing
- [ ] scrolling is lagging (throws you forward and backwards a bit while scrolling)
- [ ] doesn't always start completely on bottom

### DIFF
--- a/docs/KEYBINDINGS.md
+++ b/docs/KEYBINDINGS.md
@@ -9,3 +9,5 @@ Search contact list: `ctrl + k`
 Focus message composer: `ctrl + n`
 
 Open settings: `cmd + ,` or `ctrl + ,`
+
+Scroll to end of chatView `alt + end`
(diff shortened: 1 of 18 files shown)


Simon-Laux:
we should backport the extraction/deduplication of the messagecontext menu to master and then rebase this branch.

Simon-Laux:
lets close this for now and put it into project resurrection / use it as reference later.

PR #1686, opened 2020-05-28
Update translations, use translated switch account & device_talk & saved_messages

Jikstra:


src/main/ipc.ts
```diff
@@ -371,6 +371,8 @@ function txCoreStrings() {
   strings[C.DC_STR_CANTDECRYPT_MSG_BODY] = tx('systemmsg_cannot_decrypt')
   strings[C.DC_STR_CANNOT_LOGIN] = tx('login_error_cannot_login')
   strings[C.DC_STR_SERVER_RESPONSE] = tx('login_error_server_response')
+  strings[68] = tx('device_talk')
+  strings[69] = tx('saved_messages')
```
Simon-Laux, line 375: why numbers instead of constants?

Jikstra: Because the constants aren't in the core currently/need an update to deltachat-node & -core. See https://github.com/deltachat/deltachat-core-rust/pull/1537

PR #1701, opened 2020-06-04
Less cursor switching

### LINKED ISSUE #1602: mouse pointer glitches
in general, i dislike the flicking between the arrow-pointer and the _pointing-hand_, it is always a bit unclear, when which one is shown.

eg. in the settings-dialog, there is lots of switching every few pixels, EDIT: but also if a message spans more than one line

1. i suggest to use the arrow-pointer for _everything but links in messages_ - we are an application, not a website :)

2. in the title above the chat (the one with "groupname" and "3 members"), the lower few pixels of the subtitle show the arrow-pointer, if you go up a few pixels, the pointing-hand is shown.  
however, when targeting 1. , this should already be fixed.

3. the right area in the title above the chat is sometimes not clickable, a second click usually works

4. when opening the group-profile and closing the dialog with escape - the pointer is gone

Simon-Laux:
- Default cursor for
   - chatlist items 
   - message text 
   - contactlist item 
   - context menu
- shrink clickable area of chattitle in navbar

fixes #1602

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@
 - open an electron error dialog on uncaught Excerptions
 - Tweak Login styling
 - move qr code description up
+- Tweak: Less cursor type switching
 
 ### Technical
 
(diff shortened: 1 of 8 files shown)


r10s:
nb: there are still some text-cursors, eg. on the date-field in the message bubble, but that can be fixed as one come over it ...

PR #1707, opened 2020-06-05
add chat avatar to navbar

Simon-Laux:
- add chat avatar to navbar
- move verified icon overlay over avatar
- refactor avatar class


Jikstra:
Nice to have the verified logo on top of the group logo, like it. But i'm not sure if having the avatar in the navbar really makes sense?

r10s:
> But i'm not sure if having the avatar in the navbar really makes sense?

it does. 
- if the chatlist scrolls away - which is much more probably with the new search, you have a much better idea where you are.
- in a group, you have all the user avatars where you can click on to go to the profile - but no group-avatar
- also, this way, we have a a recognizable thing between android/ios/desktop.

PR #1708, opened 2020-06-08
Change accounts folder structure again

Simon-Laux:
- You can have multiple accounts with the same email address (for example different backups you can distinguish them via the new size property on the known accounts section)
- Account folder name isn't tied to email address anymore, you can rename account folders as you like. New account folders have the format `acc[number]`. This mimics the behavior of dc iOS and dc android which could help with burner qr code reading (allows having the context before knowing the email address).

### DIFF
--- a/src/main/logins.ts
+++ b/src/main/logins.ts
@@ -5,7 +5,7 @@ import * as binding from 'deltachat-node/binding'
 import { getLogger } from '../shared/logger'
 const log = getLogger('main/find_logins')
 import { getAccountsPath, getConfigPath } from './application-constants'
-import { PromiseType } from '../shared/shared-types'
+import { DeltaChatAccount } from '../shared/shared-types'
 
 export async function getLogins() {
   // search for old accounts and convert them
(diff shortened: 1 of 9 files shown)


hpk42:
so this is completely backward-compatible with existing account dirs, right?

Simon-Laux:
Yes it is still compatible, but downgrading to under 0.999 is not possible anymore.

src/main/logins.ts
```diff
@@ -103,18 +74,46 @@ async function readDeltaAccounts(accountFolderPath: string) {
...
+      if (!binding.dcn_is_configured(dcn_context)) {
+        result = null
       } else {
-        resolve(result)
+        keys.forEach((key: string) => {
+          result[key] = binding.dcn_get_config(dcn_context, key)
+        })
       }
+      binding.dcn_close(dcn_context, () => {})
+      resolve(result)
```
Jikstra, line 100: This will very likely conflict with the async pr, but that's fine

PR #1714, opened 2020-06-12
Fix "dc_get_msg called with special msg_id=0" warnings

link2xt:
When startIndex == stopIndex == 0, ids was equal to [null] and caused
message with msgId = 0 to be loaded. This commit simplifies the logic
and fixes the warning.

<del>stopIndex + 1 is replaced with stopIndex, because stopIndex is limited
by Math.min(messageResultIds.length, 10). When length is less than 10,
it caused slice to try to load more elements that were in the array,
and 11 elements maximum were loaded, which seems to be off-by-one error.</del>


Simon-Laux:
strange.. I added the +1 because it forgot to load some items... I need to investigate this further.

PR #1742, opened 2020-06-22
Fix scrolling down on incoming messages and give images a fixed height

### LINKED ISSUE #1513: scroll down on each incoming message
when scrolling up to read the backlog of a chat, each incoming message scrolls down. this is annoying in a frequently used chat.

i think it should only scroll down when the last message is in view or so.

tbh, no idea how the situation is on android/ios :)

Jikstra:
Fixes https://github.com/deltachat/deltachat-desktop/issues/1513


r10s:
great, you push that forward!

not sure what the removed `scrollToBottom: true` exactly does, however, removing it _always_ seem to be not-so-good.

i'd say, if the last message is in view (or if scrolling is already at bottom), we should continued scrolling down - this way, you can sit back and watch a chat - or wait for a reply etc, without additional scrolling.

only if the user scrolled up manually, to read the backlog or so, we should not scroll back.

Jikstra:
@r10s we now scroll down if we're as close as 7px to the bottom of the Message List.

PR #1771, opened 2020-06-28
add markdown labeled links

Simon-Laux:
parse labeled link `[label](url)`

### TODO

- [X] Basic parsing and dialog
- [X] basic Punycode detection
- [x] limit space that the url can take in the dialog (dialog should not grow over the window size)
- [x] should we limit how big the labeled url can be?
- [x] Move the labeled link logic to its own file
- [x] Add a copy link to clipboard button to the dialog
- [x] Add a button for marking the domain as safe "Always open links from <this domain>" (~~where should we save the resulting list?~~ for now we save it under localstorage)
- [ ] rebase

### Notes
In the end we will handle the message parsing in the core and probably return somekind of AST that can be safely rendered on all platforms.
for more markdown ideas see https://support.delta.chat/t/markdown-support-in-chat/159/2 and https://github.com/deltachat/interface/pull/20


Simon-Laux:
should we limit how big the labeled url can be? and if so what should the limit be? 500, 1000 chars?

src/renderer/components/message/LabeledLink.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const onClick = (ev: React.MouseEvent<HTMLAnchorElement, MouseEvent>) => {
+    ev.preventDefault()
+    ev.stopPropagation()
+    //check if domain is trusted
+    if (isDomainTrusted(domain)) {
+      openExternal(target)
+      return
+    }
+    // not trusted - ask for confimation from user
+    ;(openDialog as OpenDialogFunctionType)(({ isOpen, onClose }) => {
```
Jikstra, line 53: Can you factor this out into an own LinkDialogFunction? I'm fine with this living in this file.

PR #1784, opened 2020-07-09
focus emoji seachbar on opening emoji picker

Simon-Laux:
on emoji click focus its searchbar, when closing refocus composer

### DIFF
--- a/src/renderer/components/composer/EmojiAndStickerPicker.tsx
+++ b/src/renderer/components/composer/EmojiAndStickerPicker.tsx
@@ -3,7 +3,7 @@ import React, {
   useEffect,
   forwardRef,
   PropsWithChildren,
-  useRef,
+  useLayoutEffect,
 } from 'react'
 import { Picker, EmojiData } from 'emoji-mart'
 import classNames from 'classnames'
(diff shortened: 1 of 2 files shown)


src/renderer/components/composer/EmojiAndStickerPicker.tsx
```diff
@@ -121,6 +123,18 @@ export const EmojiAndStickerPicker = forwardRef<
     setStickers(stickers)
   }, [])
 
+  const emojipicker = useRef(null)
+
+  useEffect(() => {
+    emojipicker.current
+      ?.querySelector('.emoji-mart-search')
+      ?.querySelector('input')
+      ?.focus()
```
Jikstra, line 132: `document.querySelector('.emoji-sticker-picker__emoji-picker > .emoji-mart-search > input')?.focus()`

src/renderer/components/composer/EmojiAndStickerPicker.tsx
```diff
@@ -121,6 +123,15 @@ export const EmojiAndStickerPicker = forwardRef<
     setStickers(stickers)
   }, [])
 
+  useLayoutEffect(() => {
+    document.querySelector('.emoji-sticker-picker__emoji-picker > .emoji-mart-search')
+      ?.querySelector('input')
```
Simon-Laux, line 128: this is separate because this way we don't need a type conversion

PR #1788, opened 2020-07-12
Fix black color in search input

Jikstra:


scss/misc/_search-input.scss
```diff
@@ -29,3 +29,7 @@ input.search-input {
     width: calc(100% - 80px);
   }
 }
+
+input#chat-list-search {
+  color: #fff;
```
Simon-Laux, line 34: hardcoded white? can you use a theme variable?

Simon-Laux:
can you rebase this?

scss/misc/_search-input.scss
Jikstra: Didn't use a themevariable as currently also the X icon is a hardcoded white. But for sure, a theming variable makes sense here. But I have to admit that I hate adding theming variables :D

Simon-Laux: yeah, but I believe the var you are searching already exists: `navBarText`

Simon-Laux:
after you re-based this we can merge this.

PR #1795, opened 2020-07-14
add alreaddy logged in check to login function

### LINKED ISSUE #1794: On Crash, better exit if context can not be stopped
i had duplicate message issues (https://github.com/deltachat/deltachat-core-rust/issues/1701 ) which i thought were core-related bu i now think it was just desktop. I had a Desktop crash multiple times and "reloaded" (a new option i didn't see before) but i now think that this does not "stop_io()" the context.  So this is how i ended up with at some point 4 contexts running, and saw each incoming messages four times. 

So if stop_io() can not be called anymore, it's better to only offer the option to exit, and not reload.

Simon-Laux:
fixes #1794

### DIFF
--- a/src/main/deltachat/login.ts
+++ b/src/main/deltachat/login.ts
@@ -40,6 +40,10 @@ export default class DCLoginController extends SplitOut {
     credentials: Credentials,
     updateConfiguration = false
   ) {
+    if(this._controller._dc) {
+      log.info("allready logged in, logging out first")
+      await this.logout()
+    }
     log.info(`Using deltachat instance ${this._controller.accountDir}`)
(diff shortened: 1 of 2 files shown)


src/main/deltachat/login.ts
```diff
@@ -40,6 +40,10 @@ export default class DCLoginController extends SplitOut {
     credentials: Credentials,
     updateConfiguration = false
   ) {
+    if(this._controller._dc) {
+      log.info("allready logged in, logging out first")
```
Jikstra, line 44:
```suggestion
      log.info("already logged in, logging out first")
```

PR #1814, opened 2020-07-26
[ux] Improve footer action buttons

Jikstra:


src/renderer/contexts.ts
```diff
@@ -29,7 +29,7 @@ export const i18nContext = React.createContext<getMessageFunction>(
  * Otherwise use the `<i18nContext.Consumer>` or when you don't need
  * the dynamic updating functionality use `window.static_translate` directly.
  */
-export const useTranslationFunction = () => useContext(i18nContext)
+export const useTranslationFunction = () => window.static_translate
```
Simon-Laux, line 32: nope don't change this

Simon-Laux:
some buttons are still grey (encryption info, auto-crypt setup confirm), can we make the default primary?

PR #1820, opened 2020-07-28
Summary without chatlist c-object

### LINKED ISSUE #1762: search shows placeholders
on some searches, maybe when typing a bit faster, not sure, only the first few chats are shown, for he others the placehoders stay.

the placeholders are not clickable.

the number of chats in the title is correct.

Simon-Laux:
todo:

- [X] new backend functions 
- [x] modify the virtual list to use the new functions
- [x] use the virtual chatlist also in forward message and profile view (get rid of unneeded code)

related:
requires https://github.com/deltachat/deltachat-node/pull/474
fixes #1762 

note:
doesn't update contact name immediately in chat-list as long as core doesn't throw an chat changed event (see https://github.com/deltachat/deltachat-core-rust/issues/1930 for progress on that bug)


Simon-Laux:
- [ ] some events don't return a message id, such as chat changed so I need to get the previous message id in this case.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 - Fix automatically login to last remembered account
 - Fix scanning qr code and general improvement of the whole process
 - Fix avatar initials of the text avatars, of people with no displayname, inside of the messagelist
+- Fix search shows placeholders
```
Jikstra, line 22: What does this mean?

src/main/deltachat/chatlist.ts
```diff
@@ -96,7 +111,67 @@ export default class DCChatList extends SplitOut {
     return chats
   }
 
-  async getChatListSummary(list: ChatList, i: number) {
+  async getChatListItemsByEntries(entries: [number, number][]) {
+    const label = '[BENCH] getChatListItemsByEntries'
```
Jikstra, line 115: I think we can uncomment the benchmarking?

Jikstra:
I have the feeling that this could introduce bugs, can we postpone this to the next version/merge window?

src/main/deltachat/chatlist.ts
Simon-Laux: yes probably

Simon-Laux:
It's not like I didn't test it.
I know two small bugs, but that is dependent on a new core and then will need like 10 lines of code to fix it:
- Contactname triggered chatname update (only update core, no changes necessary)
- Contactname updates in chat summaries (requires the said lines of code, see https://github.com/deltachat/deltachat-desktop/issues/1889)

CHANGELOG.md
Simon-Laux: #1762
can you think of a better sentence?

src/renderer/components/chat/ChatList.tsx
```diff
@@ -371,59 +382,129 @@ function useLogic(queryStr: string, showArchivedChats: boolean) {
...
         [chatId]: LoadStatus.LOADED,
       }))
     }
+    if (chatId === 0) {
+      // setChatLoading({})
+      // setChatCache({})
+    } else {
+      if (messageId === 0) {
+        // if no message id is provided it tries to take the old one
+        // workaround to get msgId
```
Simon-Laux, line 433: I think I need to redo this workaround as it uses chatListIds which can and probably will be a deprecated snapshot of the value

Issue #445, opened 2018-12-15
Put in system tray on close (Background Mode)?


Simon-Laux:
An option to put it in the system tray instead of completely closing it would be cool.
Telegram and Discord (also skype, but that application is a pain to quit) do that.

- [ ] Tray icon with menu

- [ ] Don't quit on close, close only the window

- [ ] Make a setting to disable the tray icon / background mode

Simon-Laux:
https://electronjs.org/docs/api/tray

Simon-Laux:
Yeah lets wait on opinions on that one first.
Also Thunderbird doesn't do a tray icon.

karissa:
+1 on the "wishlist" feature list

Jikstra:
@Simon-Laux in case you're still bored, this could be a nice feature?

Jikstra:
Closing as no one works on this currently

radasbona:
The Tray Feature is needed very much. (UbuntuMate)

r10s:
@radasbona i totally agree and i think everyone here would like to have this implemented.

however, this issue tracker is about bugs and to track things someone is currently working on. for feature requests and discussions, please use our forum at https://support.delta.chat

PR #1840, opened 2020-08-04
[wip] Add Tray icon to reduce app easily without quitting application definitively

### LINKED ISSUE #1548: Close desktop version to system tray
As I understand - delta chat can not be "closed" to system tray? Or I missing this setting?

### LINKED ISSUE #1692: There is no app icon for i3wm/i3-gaps (i3blocks)
- Operating System (Linux/Mac/Windows/iOS/Android): Arch Linux
- Delta Chat Version: Latest. Installed from aur git.
- Expected behavior: Showing icon on the pannel i3blocks, as doing other apps, for example Telegram on the screenshot below.
- Actual behavior: No any icon
- Screenshots:
18h_15m_42s_30_May_2020

- Logs:

pepea28:
I wanted the tray icon feature (#445).

I developped myself a basic version. I'm not a front-end developer so my javascript skills are limited.
I didn't implement the translations part, nor the part that could make this feature customizable.

This has only been tested on windows for now.

closes #1692
closes #1548

### DIFF
--- a/src/main/application-constants.ts
+++ b/src/main/application-constants.ts
@@ -1,11 +1,10 @@
 import appConfig from './application-config'
-import path, { dirname, join } from 'path'
+import { dirname, join } from 'path'
 import { screen } from 'electron'
 
 export function appIcon() {
   const iconFormat = process.platform === 'win32' ? '.ico' : '.png'
-  return `${path.join(__dirname, '..', '..', 'images', 'deltachat'+iconFormat)}`
-
+  return `${join(__dirname, '..', '..', 'images', 'deltachat' + iconFormat)}`
 }
 
 export function windowDefaults() {
--- a/src/main/application-constants.ts
+++ b/src/main/application-constants.ts
@@ -1,10 +1,11 @@
 import appConfig from './application-config'
-import { dirname, join } from 'path'
+import path, { dirname, join } from 'path'
 import { screen } from 'electron'
 
 export function appIcon() {
-  // TODO Add .ico file for windows
-  return join(__dirname, '..', 'images', 'deltachat.png')
+  const iconFormat = process.platform === 'win32' ? '.ico' : '.png'
+  return `${path.join(__dirname, '..', '..', 'images', 'deltachat'+iconFormat)}`
+
 }
 
 export function windowDefaults() {
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -3,12 +3,6 @@ console.time('init')
 import { ensureDirSync, watchFile } from 'fs-extra'
 import { app as rawApp, session, EventEmitter, dialog } from 'electron'
 import rc from './rc'
-import path from 'path'
-
-// Tray
-import { Menu, Tray } from 'electron'
-import { globalShortcut } from 'electron'
-let tray = null
 
 const app = rawApp as ExtendedAppMainProcess
 app.rc = rc
@@ -70,6 +64,7 @@ import * as devTools from './devtools'
 import { AppState, DeltaChatAccount } from '../shared/shared-types'
 import { ExtendedAppMainProcess } from './types'
 import { resolveThemeAddress, acceptThemeCLI } from './themes'
+import { updateTrayIcon } from './tray'
 
 app.ipcReady = false
 app.isQuitting = false
@@ -121,58 +116,33 @@ function onReady([logins, _appReady, loadedState]: [
   cleanupLogFolder().catch(err =>
     log.error('Cleanup of old logfiles failed: ', err)
   )
-
-  // Tray
-  tray = new Tray(
-    `${path.join(__dirname, '..', '..', 'images', 'deltachat.ico')}`
-  )
-  const contextMenu = Menu.buildFromTemplate([
-    {
-      label: 'Réduire',
-      type: 'normal',
-      accelerator: 'Escape',
-      click() {
-        const win = mainWindow.window
-        win.close()
-      },
-    },
-    {
-      label: 'Quitter',
-      type: 'normal',
-      click() {
-        globalShortcut.unregisterAll()
-        if (process.platform !== 'darwin') {
-          app.quit()
-        }
-      },
-    },
-  ])
-  tray.setToolTip('Deltachat')
-  tray.setContextMenu(contextMenu)
-  tray.on('double-click', (event, bounds) => {
-    mainWindow.show()
-  })
 }
 
 ;(app as EventEmitter).once('ipcReady', () => {
+  const IsMac = process.platform === 'darwin'
   console.timeEnd('init')
   if (process.env.NODE_ENV === 'test') {
     mainWindow.window.maximize()
   }
+
+  updateTrayIcon()
+
   mainWindow.window.on('close', e => {
     if (!app.isQuitting) {
       e.preventDefault()
       mainWindow.hide()
-      //quit(e)
+      if (!IsMac && !app.state.saved.minimizeToTray) {
+        quit(e)
+      }
     }
   })
 })
 
-function quit(e: Electron.Event) {
+export function quit(e?: Electron.Event) {
   if (app.isQuitting) return
 
   app.isQuitting = true
-  e.preventDefault()
+  e?.preventDefault()
 
   function doQuit() {
     log.info('Quitting now. Bye.')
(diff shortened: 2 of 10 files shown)


Simon-Laux:
First of all thanks for contributing!
I also tried doing this, problem was that the electron tray menu was buggy on my machine.

Also don't edit the language xml files, they are managed by the android repo and transifex the desktop repo only pulls them but does not push any language strings. 
we use an extra file for adding new strings in desktop,
see https://github.com/deltachat/deltachat-desktop/blob/master/docs/DEVELOPMENT.md#add-experimental-language-strings-

pepea28:
Thanks for your feedback.

I added the setting you did.
For now, it is necessary to restart the application for this to be taken into account. 
I'll maybe check if it's possible for me to update in real time the tray icon according to the setting.

Simon-Laux:
My problems on gnome were:
- In the previous version of electron that we used the click event wasn't triggered. (but this was solved with upgrading to electron 6)
-  realtime updating that setting (TrayIcon.destroy to remove it) failed on all electron versions I tried ( 4 an 6) instead I got more and more tray icons when switching the setting on and off.

So we need to test this pr on multiple platforms to see if our current electron version makes problems, because when it does we probably need to update electron.

Jikstra:
@pepea28 i changed some bits to make it work on linux, probably on windows it needs the .ico, while on linux it only works with the .png.

src/main/tray.ts
```diff
@@ -0,0 +1,67 @@
...
+      // User doesn't want tray icon => destroy it
+      if (tray != null) {
+        log.info('destroy icon tray')
+        tray.destroy()
+      }
+    } else {
+      // Add tray icon
+      log.info('add icon tray')
+      tray = new Tray(
+        `${path.join(__dirname, '..', '..', 'images', 'deltachat.ico')}`
```
Simon-Laux, line 30: rather use `appIcon()` here
to import it:
```js
import { appIcon } from './application-constants'
```
and extend the function in the application constants to return the path to the ico file instead of the png on windows. (https://nodejs.org/api/process.html#process_process_platform)
```

**Comment by Simon-Laux** [comment]
I can't see the tray icon for some reason, maybe it has something to to with the uncaught conversion error when starting up the app. I changed the code in my local working copy a bit, not the error is not there anymore, but the tray icon is not visible.

```js
const icon = nativeImage.createFromPath(appIcon())
tray = new Tray(icon)
```
(using path to ico and png directly both resulted in somekind of conversion error, maybe it is a bug in electron)

Jikstra:
@Simon-Laux did you check out what I changed? I used the .png and now it's shown. I think it's more of a windows/linux thing then a bug in electron.

Jikstra:
@pepea28 is it sitll working on windows with the .png?

pepea28:
@Jikstra with the png, icon worked but quality wasn't ok on Windows : 
image     
I updated `appIcon` to use ico or png depending on OS.

@Simon-Laux I will check on linux :

- that notification work on linux (@Jikstra [reported](https://github.com/pepea28/deltachat-desktop/pull/1#issuecomment-670950053) issue since this PR)
- if we can improve the behaviour with virtual desktops

src/main/application-constants.ts
```diff
@@ -1,10 +1,11 @@
 import appConfig from './application-config'
-import { dirname, join } from 'path'
+import path, { dirname, join } from 'path'
```
Simon-Laux, line 2: you can remove the `path` here and use only `join` instead of `path.join` bellow.
```suggestion
import { dirname, join } from 'path'
```

src/main/application-constants.ts
```diff
@@ -1,10 +1,11 @@
...
 import { screen } from 'electron'
 
 export function appIcon() {
-  // TODO Add .ico file for windows
-  return join(__dirname, '..', 'images', 'deltachat.png')
+  const iconFormat = process.platform === 'win32' ? '.ico' : '.png'
+  return `${path.join(__dirname, '..', '..', 'images', 'deltachat'+iconFormat)}`
```
Simon-Laux, line 7:
```suggestion
  return `${join(__dirname, '..', '..', 'images', 'deltachat'+iconFormat)}`
```

Simon-Laux:
The results of my new test:

- I now get the icon displayed - nice!
- Click and double click don't work for me on gnome, but that is most likely a gnome thing (telegram doesn't have it, too).
- When toggling the setting for the tray icon the old ones are not removed so I end up with multiple tray icons in my tray: 

We should now to try different newer versions of electron to find one where it works, and hopefully find one where it does work.

What I didn't try yet is whether the window gets displayed again when I click on a notification and it is minimized to tray, but I think this shouldn't be a problem for electron.

Maybe we should change the wording in the menu to contain the word DeltaChat like telegram does:

Also disabling the minimize to tray option when it is already active seems like another good idea we can do too.
(but let's ignore the mute notification button for now, if we want it we can implement it in another pr :wink: )

pepea28:
Too bad the click / double click doesn't work with gnome. I think @Jikstra mentioned that it works with KDE.
The electron documentation [says ](https://www.electronjs.org/docs/api/tray):
`When app indicator is used on Linux, the click event is ignored.`

I don't know why the icon destruction is not working on Linux when toggling the setting. Is each "ghost icon" clickable and opens context menu ?

On Windows, clicking on the notification when the application is minimized opens Delta Chat. I will test on Linux (as soon as I reinstall it on my computer).

src/main/application-constants.ts
pepea28: You're right, thank you

Simon-Laux:
ghost icon with context menu, but the context menu items don't do anything.

pepea28:
I've tested on gnome and I do have the behavior you described.

I did a test with electron 9.0.2 ([tray-electron-9 branch](https://github.com/pepea28/deltachat-desktop/tree/tray-electron-9)) and it works: 
Peek 16-08-2020 21-43

However, there seems to be a bug with the latest version of electron again (9.2.0) : https://github.com/electron/electron/issues/24976



I don't know if there are any plans to upgrade the electron version for deltachat.

Jikstra:
We're normally upgrading the electron version whenever we need it. The problem with upgrading is, that very often we had to downgrade again because some things broke or new bugs got introduced in the newer electron version. But I don't see a problem with upgrading electron for this feature.

src/main/tray.ts
```diff
@@ -0,0 +1,86 @@
...
+const app = rawApp as ExtendedAppMainProcess
+const log = getLogger('main/tray')
+
+ipcMain.on('updateTrayIcon', updateTrayIcon)
+
+export function updateTrayIcon() {
+  const tx = app.translate
+
+  const IsMac = process.platform === 'darwin'
+  if (!IsMac) {
```
Jikstra, line 21: Can't we do an early return here?

`if (process.platform === 'darwin') return`

src/renderer/components/dialogs/Settings.tsx
```diff
@@ -115,6 +116,9 @@ export default function Settings(props: DialogProps) {
       (await DeltaBackend.call('settings.setDesktopSetting', key, value)) ===
       true
     ) {
+      if (key === 'minimizeToTray') {
+        ipcBackend.send('updateTrayIcon')
```
Jikstra, line 120: Can we refactor this into a `DeltaBackend.call('tray.updateTrayIcon')` ? @Simon-Laux what's your opinion on this?

src/main/index.ts
```diff
@@ -118,24 +119,35 @@ function onReady([logins, _appReady, loadedState]: [
...
+  })
 })
 
-function quit(e: Electron.Event) {
+export function quit(e?: Electron.Event) {
   if (app.isQuitting) return
 
   app.isQuitting = true
-  e.preventDefault()
+  e?.preventDefault()
```
Jikstra, line 150: Nice :)

Jikstra (commented):
Sorry for postponing this for so long, want to have this in the next release and merge asap :)

src/main/tray.ts
Jikstra: Would love to have less indentation levels here if possible

pepea28:
I tried to update to the last version of electron (10.1.3). Same issue on gnome. ([tray-electron-10 branch](https://github.com/pepea28/deltachat-desktop/tree/tray-electron-10))

src/renderer/components/dialogs/Settings.tsx
Simon-Laux: remember that we want to weaken the connection between electron and core?
Please create a function under the Runtime interface and let it do the ipc call.

Simon-Laux:
Closing in favour of https://github.com/deltachat/deltachat-desktop/pull/1906 . @pepea28 thanks for your work :)

PR #1848, opened 2020-08-08
UI/UX overhaul login, implement welcome screen, overhaul account overview

Jikstra:
ezgif-7-1fe940af76df

new-account-overview
account-overview-flow

# ToDo

- [x] processing qr code dialog looks weird on burner account setup
- [x] qr dialog doesn't close after successfull burner account setup
- [x] remove account button is missing
- [x] fix testcafe


Jikstra:
@nicodh got rid of the extra test_ci.js file for the testcafe things. I for now disabled the contact request test (which for whatever reason fails on the ci) by checking if the CI environment variable is set to true. Now we don't need to maintain two testcafe files. anymore.

PR #1852, opened 2020-08-10
fix labeled links dialog issues

### LINKED ISSUE #1839: Improve labeled links
## Problems with the current implementation

- [x] Trust domain contains junk
- [x] Punycode encoding should only affect the domain, not the path or the anchor

### Trust domain contains junk

the domain that is asked here should be only the domain, not domain + anchor

Expected: Don't ask again for example.com

### Punycode encoding should only affect the domain, not the path or the anchor

Expected: https://example.org#Straße

### Text shouldn't leave the dialog

Though I don't know if there are domains long enough to trigger this, this behavior is visible because of "Trust domain contains junk"

Simon-Laux:
add a real url parser to handle the domains, this removes the 3 bugs of #1839

**Trust domain contains junk** -> now its only the hostname as it should be
**Punycode encoding should only affect the domain, not the path or the anchor** -> encoding to punycode treatment is only applied to the hostname now.
**Text shouldn't leave the dialog** -> as the remember domain text contains only the hostname it shouldn't be a problem right now (domains aren't too long most of the time, but we could also make the test breaking in the future to fix it the right way)
fix #1839

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### Fixed
 
 - Fix sluggish ui while typing in email address in login form
+- Fix url parding in labeled link confirmation dialog
 
 
 ## [1.12.0] - 2020-07-31
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 ### Fixed
 
 - Fix sluggish ui while typing in email address in login form
-- Fix url parding in labeled link confirmation dialog
+- Fix url parsing in labeled link confirmation dialog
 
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 ### Fixed
 
 - Fix sluggish ui while typing in email address in login form
+- Fix url parding in labeled link confirmation dialog
```
Jikstra, line 13:
```suggestion
- Fix url parsing in labeled link confirmation dialog
```

Issue #1876, opened 2020-09-09
Double check that qr scanning is working & make sure error reporting behaves the same as on android


Jikstra:
EDIT: error i get
```
2.8s [i]main/state: Saving state to /home/kerle/.config/DeltaChat/config.json
7s [D]main/ipc: EVENT_DC_DISPATCH_CB (6) : checkQrCode dcaccount:https:/xxxxx
7s [E]main/deltachat: Error calling checkQrCode(dcaccount:xxxx):
 TypeError: Cannot read property 'checkQrCode' of undefined
    at DeltaChatController.checkQrCode (/home/kerle/Coding/deltachat-desktop/tsc-dist/main/deltachat/controller.js:280:25)
    at DeltaChatController.<anonymous> (/home/kerle/Coding/deltachat-desktop/tsc-dist/main/deltachat/controller.js:147:37)
    at Generator.next (<anonymous>)
    at /home/kerle/Coding/deltachat-desktop/tsc-dist/main/deltachat/controller.js:8:71
    at new Promise (<anonymous>)
    at __awaiter (/home/kerle/Coding/deltachat-desktop/tsc-dist/main/deltachat/controller.js:4:12)
    at DeltaChatController.callMethod (/home/kerle/Coding/deltachat-desktop/tsc-dist/main/deltachat/controller.js:133:16)
    at Object.<anonymous> (/home/kerle/Coding/deltachat-desktop/tsc-dist/main/ipc.js:63:52)
    at Generator.next (<anonymous>)
    at /home/kerle/Coding/deltachat-desktop/tsc-dist/main/ipc.js:8:71

Simon-Laux:
afaik we don't use the core function for it because it needs a context and we needed the email address to create a context. as this is not longer the case we could just create a folder/context with the normal random name and use the core implementation instead. If it fails, just delete the incomplete account again.

Jikstra:
Yes I have a local pr doing exactly this. Will push it but still needs some love.

PR #1879, opened 2020-09-10
Fix qr scanning issues & code

Jikstra:
- Use checkQrCode core method also on login qr code scanning
- Refactor processOpenQrUrl() method

# ToDo

- [x] the alert dialogs are weirdly on top of the process qr dialog, refactor processOpenQrUrl() method to also open the process dialog
- [x] move the QrState and CheckQrResponse types into shared types and use them in backend too

Fixes https://github.com/deltachat/deltachat-desktop/issues/1876


Simon-Laux (approved):
code LGTM

PR #1882, opened 2020-09-13
use displayname for avatar initials

### LINKED ISSUE #1881: fallback avatar does not always show letters
in groups, the avatars of the other recipients are sometimes shown as `#` instead of showing a letter, maybe because of a missing name.

in profile-file, contact-list, one-to-one-chat, however, the avatars are shown correctly.

Screen Shot 2020-09-13 at 14 55 33

Simon-Laux:
in message list - fixes #1881


r10s:
great!

Issue #1877, opened 2020-09-09
Error message is weird on wrong autocrypt setup message key


Jikstra:

Jikstra:
@r10s how is it done in the android/ios code? I'm getting back a false from the `continueKeyTransfer` method, but i'm also getting a DC_EVENT_ERROR event, with `[0, "dc_continue_key_transfer: "invalid symmetric key algorithm""]`. Can I just ignore this error event and listen to the false?

r10s:
to know if [dc_continue_key_transfer()](https://c.delta.chat/classdc__context__t.html#a5af2cdd80c7286b2a495d56fa6c0832f) succeeded or not, you have to check the return value.

the DC_EVENT_ERROR-event can be used as pop-up or might be captured or whatever. but the events do not indicate the success or failure of any function.

Jikstra:
@r10s 

> the DC_EVENT_ERROR-event can be used as pop-up or might be captured or whatever. but the events do not indicate the success or failure of any function.

The problem I have is that they have a weird and not-for-normal-users-helpful error message ("invalid symmetric key algorithm")

r10s:
> The problem I have is that they have a weird and not-for-normal-users-helpful error message ("invalid symmetric key algorithm")

ii think, that should be fixed in core then. but as mentioned on irc just now, maybe we should go for some better getting-error-strings.

however, i just had a closer look on the concrete handling in ios/android:

we're displaying `autocrypt_continue_transfer_succeeded` on success and `autocrypt_bad_setup_code` on failure. maybe you should just go for that, at least for now.

PR #1883, opened 2020-09-13
Fix autocrypt setup message

Jikstra:
- [x] Change color of asm input borders to colorNone
- [x] fix human readable error on wrong key

Fixes https://github.com/deltachat/deltachat-desktop/issues/1877


src/renderer/ScreenController.tsx
```diff
@@ -107,6 +107,10 @@ export default class ScreenController extends Component {
   onError(_event: any, [data1, data2]: [string | number, string]) {
     if (this.state.screen === Screens.Login) return
     if (data1 === 0) data1 = ''
+    console.log(data1, data2)
+    // Can get removed as soon as we use a rust core including this fix: https://github.com/deltachat/deltachat-core-rust/pull/1911
```
Simon-Laux, line 111: maybe we should create an issue for it, only a comment in the source code gets forgotten quickly.

Jikstra: It's not jigga important to remove it, but yeah if you feel like an issue, feel free to open one :)

PR #1896, opened 2020-09-30
Update deltachat-node to 1.47.0

### LINKED ISSUE #1857: adapt to new configure api
once a master with https://github.com/deltachat/deltachat-core-rust/pull/1835 is used (probably core46), some configure handling has to be adapted:

- split server_flags into server_flags/mail_security/send_security,   
otherwise, the ui-security-flags might not work.  
take care: depending on how the ffi is used, this might not be noticed by compile failures!

- dc_configure() error strings are placed in data2 when DC_EVENT_CONFIGURE_PROGRESS reports failure.  
capturing the error strings from other events is no longer needed and is not guaranteed to work any longer.

for details, see the closed issues for android/ios/core in https://github.com/orgs/deltachat/projects/26

### LINKED ISSUE #1886: Update to deltachat-node v1.46.0 and fix breaking changes

Jikstra:
- Changes login to fit new api

# ToDo
- [x] triple check logging in with different mail/send_securitz
- [x] default port placefolder is wrong currently
- [x] release new node bindings
- [x] don't listen for `DC_EVENT_ERROR_NETWORK` while configuring, listen for errors in data2 in ConfigureProgressEvent instead https://github.com/deltachat/deltachat-core-rust/pull/1905
     - ~~blocked by https://github.com/deltachat/deltachat-core-rust/issues/1956~~ we can't change it now, just gotta live with it until new core version comes out
- [x] rebase
- [x] update to 1.47.0 (make sure it compiles/builds, we can do the other adjustments in their own prs)

closes #1857
closes #1886


Simon-Laux:
we should rebase this

Simon-Laux:
So let's try out the login:

1. login to non email provider example.com => Stuck at some smtp step, no progress after that
2. Login to an non existing domain provider: => Error: Configuration failed. Error: “IMAP autoconfig did not succeed”
3. Login to a provider with wrong password: => Error: Configuration failed. Error: “IMAP autoconfig did not succeed”

To avoid problem 1 we could add a note for the user to check the credentials after a timeout (like 60 seconds?)

Simon-Laux:
Added some generic text to the error so the user will not be completely left in the dark.

Simon-Laux:
current blocker: `npm i https://github.com/deltachat/deltachat-node#core_1_47` does not work:
core compile doesn't work, because the byte_include macro for the images fails saying it cannot find them.

Simon-Laux:
I still get useless configure errors, but I can login to an new account so lets merge this for now.

r10s:
> Added some generic text to the error so the user will not be completely left in the dark.

probably useful for now. in the long run, i think, that should be removed again. the idea is that the core tells a meaningful error messages that is just displayed by the ui.

PR #1906, opened 2020-10-04
Implement tray icon

### LINKED ISSUE #1548: Close desktop version to system tray

### LINKED ISSUE #1692: There is no app icon for i3wm/i3-gaps (i3blocks)

Jikstra:
- [x] Refactor code and remove if/else indentations if possible
- [x] Hide window when clicking on tray icon if it's visible
- [x] refactor code to not depend on ipcBackend
- [x] minimize window instead of hiding it, to make it faster to show it again
- [x] focus window if it's on another virtual desktop instead of hiding it
- [x] package-lock.json should not change?
- [x] in the menu it should be named `minimize to tray` instead of `reduce`
- [x] check behavior on mac (not the trayicon, just that it isn't closed fully and can be restored from the docs)
   - I just checked mac behavior and it does not look good. Clicking on the dock icon does not restore the window.
- [x] investigate gnome bug? (toggle tray):

- [x] move setting to experimental

about the gnome bug we could also create an issue when merging and postpone the investigation.

successor of 1840
closes #1692
closes #1548


Jikstra:
Actually I would like to merge this into @pepea28/tray but somehow I can't make this work. @Simon-Laux @pepea28 any ideas?

Simon-Laux:
you'll need access to @pepea28's repo, @Jikstra

Simon-Laux:
otherwise we could make a collective tray icon branch on the deltachat/deltachat-desktop repo in which we merge the prs regarding it.

Jikstra:
@Simon-Laux hmm but also for creating a pull request? Other thing in my mind, probably it's fine to not draw the badge icon for now and do it in a seperate pr. We need to do some canvas magic, all the libraries i saw are either not on npm or do weird things. element-web is also doing it by themselves.

pepea28:
The settings part is not present in this PR. So it is difficult to test. Is it normal ?
It's a good idea to focus the window if it's on another virtual desktop.

Does minimizing the window instead of hiding it leave Deltachat in the taskbar ?

Jikstra:
@pepea28 
> The settings part is not present in this PR. So it is difficult to test. Is it normal ?

Should work? Should work, at least I didn't remove it :D

> Does minimizing the window instead of hiding it leave Deltachat in the taskbar ?

Yes, this is not optimal but hiding and showing it feels to slow on my device, that's why I chose this path.

Simon-Laux:
- [ ] in the menu it should be named `minimize to tray` instead of `reduce`
- [ ] check behavior on mac (not the trayicon, just that it isn't closed fully and can be restored from the docs)
- [ ] investigate gnome bug? (toggle tray):

src/main/menu.ts
```diff
@@ -169,6 +169,13 @@ function getMenuTemplate(logHandler: LogHandler): rawMenuItem[] {
                 translate: 'global_menu_file_quit_desktop',
                 role: 'quit',
               },
+              {
+                translate: 'global_menu_file_reduce_desktop',
+                accelerator: 'Escape',
```
Simon-Laux, line 174: I wouldn't introduce the escape shortcut here, escape also does other things and I think the normal user gets confused when escape results in the window being invisible.

src/main/menu.ts
```diff
@@ -169,6 +169,13 @@ function getMenuTemplate(logHandler: LogHandler): rawMenuItem[] {
                 translate: 'global_menu_file_quit_desktop',
                 role: 'quit',
               },
+              {
```
Simon-Laux, line 172: This option should only be shown when the tray icon setting is active

src/main/tray.ts
```diff
@@ -0,0 +1,102 @@
...
+      type: 'normal',
+      click() {
+        win.show()
+      },
+    },
+    {
+      id: 'reduce_window',
+      label: tx('global_menu_file_reduce_desktop'),
+      type: 'normal',
+      accelerator: 'Escape',
```
Simon-Laux, line 58: I don't know what the `accelerator` should do here, I suggest to remove this line

src/main/menu.ts
Simon-Laux: or if we are on mac

Jikstra:
Just for completeness, the electron bug for the multiple tray icons:
https://github.com/electron/electron/issues/17622

Jikstra:
@Simon-Laux what do you mean with "I just checked mac behavior and it does not look good. Clicking on the dock icon does not restore the window." ? Just tried it out on the one around me, activating the tray icon has no effect (and could get hidden) but besides this everything works as expected.

PR #1908, opened 2020-10-06
refresh "empty chat" info meassage on chat changes

### LINKED ISSUE #1850: Changing contact name does not refresh views
Current master-desktop/node/core: when changing the profile name of a contact & clicking  ok, the chatview does not refresh with the new name.  Selecting a different chat fixes the problem.  So the name is updated but a redraw event is missing, it seems.

Simon-Laux:
should work, haven't tested it yet because it needs a new core(1.47) (https://github.com/deltachat/deltachat-core-rust/pull/1936) which isn't out yet

fixes #1850

### DIFF
--- a/src/renderer/components/message/MessageList.tsx
+++ b/src/renderer/components/message/MessageList.tsx
@@ -1,4 +1,4 @@
-import React, { useRef, useEffect, useLayoutEffect } from 'react'
+import React, { useRef, useEffect, useState } from 'react'
 import { MessageWrapper } from './MessageWrapper'
 import {
   useChatStore,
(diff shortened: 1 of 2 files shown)


hpk42:
makes sense to merge to #1896 maybe?

hpk42:
also are the CI failures expected?

Simon-Laux:
> makes sense to merge to #1896 maybe?

Maybe, as it allows to be run with the newest core... But I think it would make more sense to make an 1.13 branch and progress the other version on master, but for the time being this pr can still lay around.
Also keep in mind that this change requires core 1.47 and not 1.46 so merging it in the 1.46 branch wouldn't make too much sense in my eyes.

> also are the CI failures expected?

that were just formatting mistake and missing changelog change this time, but normally they would be expected on a pr that wants a different core version.

src/renderer/components/message/MessageList.tsx
```diff
@@ -250,6 +226,52 @@ export const MessageListInner = React.memo(
...
+  const tx = useTranslationFunction()
+
+  const [chat, setChat] = useState<FullChat>(chatStoreState)
+  useEffect(() => {
+    const refresh = (_ev: any, chatId: number) => {
+      if (chatId == chat.id) {
+        DeltaBackend.call('chatList.getFullChatById', chat.id).then(
+          fullChat => {
+            setChat(fullChat)
+          }
```
Jikstra, line 239: Shouldn't this get handled by the chat store?

Simon-Laux: probably, but I didn't want to deal with the chatstore. If you want to move it there feel free to do so.

Simon-Laux: looks like you're right this gets already handled by the chat store, I'll change it to use the chat-store instead.

PR #1911, opened 2020-10-06
Account selection improvements

### LINKED ISSUE #1085: Remove account confirm should show which account gets removed
`this Login` should be replaces by the accounts email address.
This requires to change the translation string to make use of substitutions.

I did a similar change recently on delete chat: https://github.com/deltachat/deltachat-desktop/commit/54ba2c75913c3f0bf8bb28e0a4a0359146a90133 (for reference on such a change)

Simon-Laux:
- Add Keyboard navigation between accounts in account selection screen (arrow keys or tab)
Peek 2020-10-06 23-50

- Add the account name to the account deletion-confirmation dialog (closes #1085)

- Add a hover tool-tip with account path and size when hovering over the email address of an account
  (this is useful when importing a backup of an account you already have and can't distinguish which is the old account and which is the imported one)
Peek 2020-10-06 23-18
- add hover effect to remove button (see previous gif)

### DIFF
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -199,6 +199,6 @@
     "message": "DEPRECATED - PLEASE DELETE THIS STRING"
   },
   "account_info_hover_tooltip_desktop":{
-    "message": "email: %s\nsize: %s\npath: %s"
+    "message": "E-Mail: %s\nSize: %s\nPath: %s"
   }
 }
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -197,5 +197,8 @@
   },
   "forget_login_confirmation_desktop": {
     "message": "DEPRECATED - PLEASE DELETE THIS STRING"
+  },
+  "account_info_hover_tooltip_desktop":{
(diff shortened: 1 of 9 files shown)


_locales/_untranslated_en.json
```diff
@@ -149,5 +149,17 @@
...
+    "message": "Delete %s?"
+  },
+  "delete_account_confirmation_desktop":{
+    "message": "All account data of \"%s\" on this device will be deleted, including your end-to-end setup, contacts, chats, messages and media. This action cannot be undone."
+  },
+  "forget_login_confirmation_desktop": {
+    "message": "DEPRECATED - PLEASE DELETE THIS STRING"
+  },
+  "account_info_hover_tooltip_desktop":{
+    "message": "email: %s\nsize: %s\npath: %s"
```
Jikstra, line 163:
```suggestion
    "message": "E-Mmail: %s\nSize: %s\nPath: %s"
```

Jikstra (commented):
I don't like how the hover on the remove icon looks, but makes sense to have it, also for removing group members.

Simon-Laux:
> also for removing group members.

at the moment its only for the login screen.

PR #1912, opened 2020-10-07
fix linux/debian appIcon

### LINKED ISSUE #1772: Installed Delta Chat the icon is not visible
- Delta Chat Version: 1.4.2
- Expected behavior: When I install Delta Chat, the icon is visible
- Actual behavior: The default icon of an executable appears, instead of the Delta Chat icon.
- Steps to reproduce the problem: Install the .deb app overwriting the old version
- Screenshot:
delta_chat

Simon-Laux:
fixes #1772

It turned out the icon property was missing for linux. The one for mac is probably still wrong but I'll rather look at it on a actual mac to make sure it wasn't an intentional workaround or sth. (the icon points to a non existent `resource` directory, but the file is in the build directory)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased][unreleased]
 
+## Fixed
+
+- Fixed missing application icon for linux
+
 ## [1.13.1] - 2020-10-06
 
 ### Changed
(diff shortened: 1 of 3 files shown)


Jikstra:
<3

PR #1917, opened 2020-10-09
change used translation string for message details

### LINKED ISSUE #1916: use more precise label for "Message details" in context menu
the menu entry that will open the "Message details" is called "More info" in the context menu, however, i see no reason to use a different naming here (in contrast to eg. ios, where _really_ short strings are needed).

i suggest to just use `menu_message_details` also in the context menu.

Simon-Laux:
fixes #1916

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased][unreleased]
 
+### Changed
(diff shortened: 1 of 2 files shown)


r10s (approved):
great! i think i can approve that without knowing much about details :)

PR #1918, opened 2020-10-09
Own context menu implementation

### LINKED ISSUE #1790: Own Context menu implementation
### goals/requirements
- with backdrop that closes the context menu on click
- with simpler api and only be in DOM when its visible otherwise it might interfere with accessibility tools like the current context menu does.
- use only one context menu for message list (don't create a context menu for every message as its done right now, instead reuse only one like the chatlist context menu)
- keybindings, at-least escape

### why now?
- to improve the api, because its currently hard/hacky to work with
- fix one of the problems that make our voice over experience bad
- make it easier to switch to a virtual list for the messages (because current one is not playing well with those)
- reduce current amount elements in DOM

### LINKED ISSUE #1924: Enlargen Group Image does not work
Nothing happens on click.

Simon-Laux:
## TODO:
- [X] find out in which direction it should open
- [X] open a context menu
- [X] Block other keybindings when ContextMenu is active
- [X] controllable via keybindings (up, down, escape, enter)
- [x] provide a screen context function to open it
- [x] replace all ocurrences of the other context menu (current task)
- [x] remove react-context-menu
- [x] remove the remaining debug dummy code
- [x] what do we do about too long text? maybe do a line break?
     then the calculations will be off... maybe still better expected buggy behaviour that unexpected glitchy behaviour -> settled for estimation, works relatively good from my testing
- [ ] Optional: look a bit on the accessibility side of things?

closes #1790
also fixes #1924


Simon-Laux:
I'm currently asking myself whether the backdrop was such a good idea, looks like noboy else is doing it. though we might technically need it to detect closing, but for right click it should be click through.

Jikstra:
backdrop?

src/renderer/ScreenController.tsx
```diff
@@ -135,6 +138,13 @@ export default class ScreenController extends Component {
     this.dialogController.current.closeDialog(...args)
   }
 
+  openContextMenu(...args: Parameters<showFnType>) {
+    if (!this.contextMenuShowFn) {
+      throw new Error('Context Menu Controller not availible')
```
Jikstra, line 143:
```suggestion
      throw new Error('Context Menu Controller not available')
```

src/renderer/components/ContextMenu.tsx
```diff
@@ -0,0 +1,319 @@
...
+
+    const getValue = (key: string) =>
+      Number(style.getPropertyValue(key).replace('px', ''))
+
+    const ContextMenuWidth = getValue('--local-menu-width')
+    const ContextMenuItemLineHeight = getValue('--local-menu-item-Line-height')
+    const ContextMenuItemVerticalMargin = getValue(
+      '--local-menu-item-vertical-padding'
+    )
+    /** the margin that the context menu should have to the windo border */
```
Jikstra, line 78:
```suggestion
    // the margin that the context menu should have to the window border
```

Jikstra:
Nice! Seems to work already really well and I have the feeling that the ChatList renders faster. 

# Bugs i found

- [ ] resizing the windows fucks up the positioning, not sure how to exactly fix this
- [ ] scrolling should be disabled. This already was a problem with the old context menu. For example, open context menu for a chat list item, and then scroll the chatlist.

# Would be nice
- [ ] replace the three dot menu with this context menu so that we don't need the bp3 menu

src/renderer/components/ContextMenu.tsx
```diff
@@ -0,0 +1,319 @@
...
+import { useKeyBindingAction, KeybindAction } from '../keybindings'
+import { useRef } from 'react'
+import { ScreenContext, unwrapContext } from '../contexts'
+
+export type ContextMenuItem = { label: string; action: () => void }
+
+type showFnArguments = {
+  cursorX: number
+  cursorY: number
+  items: ContextMenuItem[]
```
Jikstra, line 11: I have the feeling that this could be a problem at some point to not be able to render own html in the ContextMenu. I guess it's easy to change but we should at least introduce a tr/divider. Besides that really like it to not need to define the items with jsx :)

Jikstra (approved):

Jikstra:
Awesome!

src/renderer/components/ContextMenu.tsx
Simon-Laux: ContextMenuItem type could be a type that can be different types, like
```ts
type ContextMenuItem = {label:string, action:Function} | { divider:true }
```
But as this needs also be taken into account in the estimation how tall the dialog will probably be and the menu has other requirements (needs to open in a static place) I'm not sure if its worth to reuse the context menu instead of doing something extra for the menu.

Simon-Laux:
> replace the three dot menu with this context menu so that we don't need the bp3 menu

I'd like to not blow up this pr that much... otherwise I can think of also solving the issues that can be solved with this pr already in this pr.
But nice idea to get rid of bp3menu as-well. even though it is a bit of another use-case as it is always in the same place.

> resizing the windows fucks up the positioning, not sure how to exactly fix this

We could close it as soon as the windows looses focus and/or gets resized.

PR #1932, opened 2020-10-13
🚧 Add preference to trust all labeled links 

### LINKED ISSUE #1894: improve button order in markdown-link confirmation
the current dialog looks like:

Screen Shot 2020-09-26 at 23 53 59

i suggest to move the "copy" or "copy link" button to the left side and add the normal "cancel" and "ok" or maybe "open" to the right:

Screen Shot 2020-09-26 at 23 53 59-2

ftr: in general, i am not a fan of this dialogs at all. showing the clickable link directly (as in the past) seems to be better. but this is another issue/discussion :)

Simon-Laux:
- Adjust order of buttons at labeled link dialog

- Add preference to trust all labeled links 

Fixes #1894

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Add the account name to the account deletion-confirmation dialog
 - Add a hover tool-tip with account path and size when hovering over the email address of an account
   (this is useful when importing a backup of an account you already have and can't distinguish which is the old account and which is the imported one)
+- add preference to trust all labeled links (don't show the confirmation dialog, just open them)
 
 ### Changed
 - Change "More info" translation to "Message Details"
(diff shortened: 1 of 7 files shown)


Simon-Laux:
So @r10s said I should remove the "Trust all Labeled Links"-option.
So here the new TODO list for this issue:
- [ ] use just the word "copy" for the button
- [ ] remove the "Trust all Labeled Links"-option
- [ ] somekind of whitelist to trust labeled links from the device chat (or atleast the dc changelogs)

r10s:
>  somekind of whitelist to trust labeled links from the device chat (or atleast the dc changelogs)

i suggest to just not use this kind of links. 

and not to go ahead in using them.

tbh, i would do the tiny layout fixes, and leave things as is for now. and go for sth. else :)

PR #1933, opened 2020-10-14
Gallery context menu

### LINKED ISSUE #1543: gallery context menu
right clicking on gallery items should open a context menu:
- View in full window
- Download
- message info
- (maybe jump to message, but that requires a message-list in which we can jump/scroll to messages.)

@r10s what else do you expect in that menu?

### LINKED ISSUE #1874: No way to download audio? (in gallery)
- Operating System (Linux/Mac/Windows/iOS/Android): Ubuntu
- Delta Chat Version: 1.12.0
- Expected behavior: Should have some way to download audio attachments
- Actual behavior: Can't find a way. Images have an obvious download button; and I can download audio from the Android app.

Simon-Laux:
fixes #1543
fixes #1874

additional Ideas:
We could also add delete message if we want.
I'm unsure of forward message, because you might accidentally forward some text that you didn't want to forward that way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Add the account name to the account deletion-confirmation dialog
 - Add a hover tool-tip with account path and size when hovering over the email address of an account
   (this is useful when importing a backup of an account you already have and can't distinguish which is the old account and which is the imported one)
+- Add context menu to gallery
 
 ### Changed
 - Change "More info" translation to "Message Details"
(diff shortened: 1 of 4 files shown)


Jikstra:
On android it's possible to forward the message. @r10s how is it done there? Only the media or the whole message?

Jikstra (approved):

Jikstra:
Maybe it makes sense to have a file like https://github.com/deltachat/deltachat-desktop/blob/master/src/renderer/components/helpers/ChatMethods.ts to have for MessageMethods?

Simon-Laux:
I don't think so as there is not enough logic in there right now.

PR #1951, opened 2020-11-06
Implement draft message with attachment & reply to message

### LINKED ISSUE #1680: Combined picture+annotation type message - missing feature? unclear workflow?
- Platform (linux/mac/windows): windows 10
- Delta Chat Desktop Version: 1.2.0 & 1.3.4
- Expected behavior: Being able to create similar picture-with-note-below messages like the android DC app.
- Actual behavior: I fail to achieve this - however I add the attachment/picture, the text and pic are always in two separate boxes.
- Steps to reproduce the problem: try to add a picture with (embedded) annotation below.
- Screenshots: -
- Logs: - 

Having messages composed of a picture and an annotation right below, in one surounding context frame, it is for my perception much better than having them separated - thus I favor this way. 
I simply fail to get that done in the desktop app. :-(

Maybe my intuition is just twisted, and I'd just need a hint in the help or the UI... or is this actually a missing feature?

Simon-Laux:
Add a draft area:
- Add possibility to add an description to an attachment
- Add quoting / reply to message

- Also removed the inline message buttons - the context menu is enough, we don't need them and they overlap sometimes with the text or the author name.

Peek 2020-11-06 21-14

## Todo

- [x] Make it prettier -> especially attachments in the staging area like images, videos and audio files don't look good.
- [x] Private reply, not too useful without the functionality of jumping to message but still we could already implement it. (make sure that it only sets the quoted message of the draft and don't change anything else if there is already a draft in that chat) 

closes  #1680

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased][unreleased]
 
-# Added
+### Added
 - Upgrade Emoji Picker(Emoji 12.1) and emojifont(Unicode 13.1) for new emoji support 🦾
 - Add Keyboard navigation between accounts in account selection screen
 - Add the account name to the account deletion-confirmation dialog
@@ -14,6 +14,9 @@
 - Add simple support for displaying qoutes (no attachment preview nor jump to message yet)
 - Show sending indicator for outgoing info messages #1867
 - Add info log message that lists all unconfigured accounts, so you don't need to find them yourself to delete them.  (see #1952)
+- Add draft/staging area
+  - add a description to the files you send
+  - reply to other messages
 
 ### Changed
 - Change "More info" translation to "Message Details"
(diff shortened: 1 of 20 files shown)


Jikstra:
@Simon-Laux I updated the look of the clear quote icon, it now looks like this:
screenshot design clear quote

Jikstra:
Other stylistic things:
I would say, let's get rid of the reply icon. The context menu got already removed i think (i like it), i think it's nicer if we don't have icons that flash up when we move around with the mouse. Just do the right click on the message and that's it. Telegram on mac also has the possibility that you can do a double click on a message to reply to it. Maybe that's neat too.

Jikstra:
Overhauled attachment drafting too

screenshot design quote and file attachment
screenshot design quote and pic attachment

PR #1957, opened 2020-11-15
Remove failed-to-configure accounts

### LINKED ISSUE #1952: Failed Login attempts aren't cleared up properly
Expected: When canceling the setup or when it fails the account folder of the newly created account is deleted again.
Actual: The account folder stays and clutters the accounts-folder

Solution:
solve it and maybe add a function that removes such empty-incomplete accounts (but that function may never delete actual accounts!!!)

Simon-Laux:
- fixes deletion of failed-to-configure accounts
- shows a log message with all found previous orphaned/invalid accounts 

closes #1952

such an log message:
```
0.6s [i]main/find_logins: unconfigured, likely orphaned accounts, you may delete them [
  '/home/username/.config/DeltaChat/accounts/ac12',
  '/home/username/.config/DeltaChat/accounts/ac19',
  '/home/username/.config/DeltaChat/accounts/ac20',
  '/home/username/.config/DeltaChat/accounts/ac21',
  '/home/username/.config/DeltaChat/accounts/ac23',
  '/home/username/.config/DeltaChat/accounts/ac24',
  '/home/username/.config/DeltaChat/accounts/ac30',
  '/home/username/.config/DeltaChat/accounts/ac31',
  '/home/username/.config/DeltaChat/accounts/ac32',
  '/home/username/.config/DeltaChat/accounts/ac33',
  '/home/username/.config/DeltaChat/accounts/ac34',
  '/home/username/.config/DeltaChat/accounts/ac36',
  '/home/username/.config/DeltaChat/accounts/ac38',
  '/home/username/.config/DeltaChat/accounts/ac39',
  '/home/username/.config/DeltaChat/accounts/ac40',
  '/home/username/.config/DeltaChat/accounts/ac41',
  '/home/username/.config/DeltaChat/accounts/ac42'
]
```
This log message can be easily copied and either transformed a shell script that removes all them or the json array could be  consumed by an deletion script.

### DIFF
--- a/src/main/deltachat/login.ts
+++ b/src/main/deltachat/login.ts
@@ -8,6 +8,7 @@ import SplitOut from './splitout'
 import { Credentials, DeltaChatAccount } from '../../shared/shared-types'
 import { getNewAccountPath, getLogins, removeAccount } from '../logins'
 import { ExtendedAppMainProcess } from '../types'
+import { remove } from 'fs-extra'
 const log = getLogger('main/deltachat/login')
 
 const app = rawApp as ExtendedAppMainProcess
@@ -115,7 +116,13 @@ export default class DCLoginController extends SplitOut {
 
   async newLogin(credentials: Credentials): Promise<DeltaChatAccount> {
     const newAccountPath = getNewAccountPath()
-    await this.login(newAccountPath, credentials)
+    try {
+      await this.login(newAccountPath, credentials)
+    } catch (error) {
+      log.debug("Detected account creation error, deleting unfinished account", newAccountPath)
+      await remove(newAccountPath)
+      throw error
+    }
     const logins = await this.getLogins()
 
     return logins.find(account => account.path === newAccountPath)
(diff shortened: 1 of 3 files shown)


src/main/deltachat/login.ts
```diff
@@ -115,7 +116,13 @@ export default class DCLoginController extends SplitOut {
 
   async newLogin(credentials: Credentials): Promise<DeltaChatAccount> {
     const newAccountPath = getNewAccountPath()
-    await this.login(newAccountPath, credentials)
+    try {
+      await this.login(newAccountPath, credentials)
+    } catch (error) {
+      log.debug("Detected account creation error, deleting unfinished account", newAccountPath)
```
Jikstra, line 122: Is this safe that we delete the account on any error? I would prefer to have some string matching here

Simon-Laux: hmm? what kind of string matching?
 > Would be bad if we accidentally delete account folders...

of other accounts you mean?

Jikstra: Ahh it only happens if we are doing a new login?

Simon-Laux: yes, see the upper part of the function:
```js
    const newAccountPath = getNewAccountPath()
```

PR #1959, opened 2020-11-19
🚧 Build the preview builds on the right platforms

### LINKED ISSUE #1559: Build the preview builds on the right platforms
#### Problems:
- At the moment all preview builds fail because wine fails to install - this would be solved by building on windows.
- We can not build/pack DMG files on the current linux ci - using a mac runner would solve that.

Simon-Laux:
remaining Problems:
- [ ] Mac signing problem - try to ignore signing? - will it run without being signed?
  - [ ] what about generating a dmg instead of mas -> probably some signing thing that we should look into?
- [X] windows alternative strategy -> upload zip artifact -> "zip not found" error (removed zip step)
- [x] windows upload via scp -> scp action does not support windows?
- [x] bin/github-actions/postLinksToDetails.js seems to not work (maybe the required sha hash is a different one than the one I used)

- [x] windows and mac builds are uploaded into the wrong folder: preview/preview/[file]

closes #1559

Please SQUASH merge when its done


Jikstra:
I think the only really important todo is:
windows and mac builds are uploaded into the wrong folder: preview/preview/[file]

Let's get this fixed and merge after 1.14.1

.github/workflows/build-preview.yml
```diff
@@ -0,0 +1,100 @@
...
+        dir preview
+        cd ..
+    # Upload Step
+    - name: upload folder
+      id: upload
+      shell: bash
+      run: |
+        echo -e "${{ secrets.KEY }}" >__TEMP_INPUT_KEY_FILE
+        chmod 600 __TEMP_INPUT_KEY_FILE
+        scp -o StrictHostKeyChecking=no -v -i __TEMP_INPUT_KEY_FILE -P "22" -r dist/preview/ "${{ secrets.USERNAME }}"@"download.delta.chat":"/var/www/html/download/desktop/preview/"
```
Jikstra, line 87:
```suggestion
        scp -o StrictHostKeyChecking=no -v -i __TEMP_INPUT_KEY_FILE -P "22" -r dist/preview/* "${{ secrets.USERNAME }}"@"download.delta.chat":"/var/www/html/download/desktop/preview/"
```

Probably this does the trick?

Issue #1939, opened 2020-10-21
Add Group Member dialog is unintutive


Simon-Laux:
When creating a group and using "Add member" while searching in between to select multiple contacts only one contact/only the last selected contacts is/are added
-> for that we wouldn't need a checkbox.

Expected: remember all selected ones and add them all, not just the ones from the last search+selection. Maybe do it like telegram and show the selected contacts in the searchfield as chips.

Whats a chip? you might ask: A chip is an ui component that looks like:
![2020-10-21_19-10](https://user-images.githubusercontent.com/18725968/96754076-1832b100-13d1-11eb-956d-b9a0121aff57.png)
This image is taken from https://ionicframework.com/docs/api/chip

r10s:
> When creating a group and using "Add member" while searching in between to select multiple contacts only one contact/only the last selected contacts is/are added
-> for that we wouldn't need a checkbox.

yip, this is a simple bug. on group-creation multiple-selection is possible (these are two dialogs, or? maybe it makes sense to merge the code at some point, iirc, we had it multiple time that things were fixed for the one but not for the other)

wrt chip: yip, this is also nice. but i would not go for for that now. this will introduce new issues, lots of work etc. the dialog is not _that_ bad. 

offtopic: if i should suggest a tiny change: i would remove the "search" input line below the "N members" headline. i find that always confusing and adding unneeded noise, making things harder to understand while adding only few benefit.

Simon-Laux:
The dialog should be readonly and have an edit mode/menu also see https://github.com/deltachat/deltachat-desktop/issues/1905.

> offtopic: if i should suggest a tiny change: i would remove the "search" input line below the "N members" headline. i find that always confusing and adding unneeded noise, making things harder to understand while adding only few benefit.

I think this method is actually more usable than the broken add member dialog.

> i find that always confusing and adding unneeded noise

In my opinion the most noise comes from the group profile view and the group edit view being the same thing. But we can and should discuss this screen to make it right, it could also profit from keybindings, same goes for the new chat/group dialog (see https://github.com/deltachat/deltachat-desktop/issues/1450)

Simon-Laux:
> on group-creation multiple-selection is possible

Nope that's where I cam across this bug.
If you click done and have a search active only selected contacts that are part of this search are added.

r10s:
> Nope that's where I cam across this bug.

i see, that is even worse. but i am pretty sure, it worked at some point.

[...]

PR #1966, opened 2020-11-21
Solve the bug part of #1939

### LINKED ISSUE #1964: "Add members" annoyingly forgets actually added members
Using pretty current master but this issue has been here for a long time: 

When you create a new group chat, then "add members" and use the searchbar to find individuals, then "check" the box for adding them, then continue to search, add another member .... and maybe add 7 members this way, then only the very last one will actually be added.  It doesn't accumulate my "clicking the checkboxes" of indivudal members but seems to forget it as soon as a new search or a new click or something comes.  I've ran into this at least 10 times and now decided to raise the issue, finally :)

Simon-Laux:
The problem that it is un-intuitive remains

Solve the bug part of #1939
closes #1964

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@
 - Fix two issues with the labeled link (see  #1893)
 - Fix refresh of "empty chat" info meassage on chat changes
 - Fix removing incompleted account (see #1952)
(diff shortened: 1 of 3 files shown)


Jikstra:
Can you explain what is un-intuitive?

PR #1969, opened 2020-11-22
Overhaul UI of about dialog

Jikstra:
# Before
Screenshot_20201122_170105

# After
Peek 2020-11-22 16-52

### DIFF
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -22,5 +22,8 @@
   },
   "pref_remove_profile_photo": {
     "message": "Remove profile photo"
+  },
+  "copy_json": {
+    "message": "Copy JSON"
   }
 }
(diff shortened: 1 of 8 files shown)


scss/dialogs/_settings.scss
```diff
@@ -68,31 +68,6 @@
       color: white;
       text-align: center;
     }
-
-    button {
```
Simon-Laux, line 72: how is this related?

Simon-Laux: ah so you moved the class to a dedicated file.

Simon-Laux (approved):
LGTM

_locales/_untranslated_en.json
```diff
@@ -22,5 +22,8 @@
   },
   "pref_remove_profile_photo": {
     "message": "Remove profile photo"
+  },
+  "copy_json": {
```
Simon-Laux, line 26: maybe there could be more context to help the translators, isn't absolutely necessary.

PR #1987, opened 2020-11-27
Add typescript eslint

### LINKED ISSUE #1965: Can we decrease unused var check to warning?
at the moment it makes a error that means the typescript build fails, this breaks the fast-iteration workflow.

So maybe there is a way to reduce it to a warning so its shown but not mandatory to fix immediately.
Or create another script that just checks it and call it lint or something like that.

Simon-Laux:
also includes a small refactoring of the set-background functionality

for more info see the documentation I added: https://github.com/deltachat/deltachat-desktop/blob/0eaa3e0090969219e23930b72c997861ec7e9433/docs/DEVELOPMENT.md#linting-

fixes #1965


docs/DEVELOPMENT.md
```diff
@@ -52,10 +55,50 @@ After making your changes go in the deltachat/electron Dev-console and press `F5
...
+
+We use [eslint](https://eslint.org) with typescript rules to catch common bad-practices.
+
+Commands:
+
+```sh
+# Run to test
+npm run lint
+# fix autofixable problems and fix formatting
+npm run lint:fix-format
```
Jikstra, line 70: how about `npm run lint:fix` ? I think keeping those things short is good.

Simon-Laux: also formats the document... I kinda wanted to include what it does

Issue #1982, opened 2020-11-26
reorder message context menu items


r10s:
currently, "Copy to clipboard" is the topmost entry, however, i think new "Reply" is more interesting and important and will probably be used more often (context menus usually have some order like that).

therefore, i suggest to move "Copy to clipboard" below "Forward".

order would be:

```
Reply
Reply privately
Forward
---------------------
[Download attachment]
Copy
[Copy link]
Message details
---------------------
Delete
```

this is also nice as the first entry is the one that is executed by a double click  - also a thing, you will see in several context menu.

i've also added some horizontal lines - the main menu have them, so, it seems possible somehow. but also without the lines, i suggest to adapt the order.

Simon-Laux:
> the main menu have them, so, it seems possible somehow.

the main menu uses a different implementation. But as we control. the code for the context menu we can really do anything we want in theory.

when we start with lines we could also think about making delete red IMO... (that you can find it faster)

where would you put `Copy Link` in your order? Top most? because it only appears if you right click on a link.

r10s:
> where would you put Copy Link in your order? Top most? because it only appears if you right click on a link.

i would put it above/below "Copy to clipboard". it is not a command that is used that frequently and i would vote for keeping a recognizable order and internal consistence in this case.

Simon-Laux:
>  it is not a command that is used that frequently

It it is the most used command from the menu for me

r10s:
i refined the suggestion and extended it to the chat-menu at https://github.com/deltachat/deltachat-desktop/pull/1993#pullrequestreview-540369603

PR #1993, opened 2020-11-27
Overhaul ux of context menus

### LINKED ISSUE #1978: do not word-wrap context menu items
if the context menu entries are a bit longer, they are word-breaked. this is quite unusual for menus in general as it is harder for the eyes to get the next entry.

example: german entry "In Zwischenablage kopieren"

Jikstra:
- Capitalize item texts
- Change some translations for items of message context menu
- Reorder actions for message context menu

Screenshot_20201127_205425
Screenshot_20201127_205919-2

I like the capitalized context menu in english quite a bit. Not so much in german. Not sure.

Fixes https://github.com/deltachat/deltachat-desktop/issues/1982
Fixes #1978

### DIFF
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -43,5 +43,8 @@
   },
   "copy_json": {
     "message": "Copy JSON"
+  },
+  "copy_text": {
+    "message": "Copy text"
   }
 }
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -43,8 +43,5 @@
   },
   "copy_json": {
     "message": "Copy JSON"
-  },
-  "copy_text": {
-    "message": "Copy text"
   }
 }
--- a/scss/misc/_context_menu.scss
+++ b/scss/misc/_context_menu.scss
@@ -35,7 +35,6 @@
   .item {
     line-height: var(--local-menu-item-Line-height);
     padding: var(--local-menu-item-vertical-padding) 12px;
-    text-transform: capitalize;
 
     &:hover {
       background-color: var(--chatListItemBgHover);
--- a/scss/misc/_context_menu.scss
+++ b/scss/misc/_context_menu.scss
@@ -35,6 +35,7 @@
   .item {
     line-height: var(--local-menu-item-Line-height);
     padding: var(--local-menu-item-vertical-padding) 12px;
+    text-transform: capitalize;
 
     &:hover {
       background-color: var(--chatListItemBgHover);
--- a/scss/overwrites/_bp3-overwrites.scss
+++ b/scss/overwrites/_bp3-overwrites.scss
@@ -184,6 +184,9 @@ $bp3-dialog-radius: 2px;
 
 .bp3-menu-item {
   padding: 5px 12px;
+  & > .bp3-fill {
+    text-transform: capitalize;
+  }
 }
 
 .bp3-menu-item:hover,
--- a/scss/overwrites/_bp3-overwrites.scss
+++ b/scss/overwrites/_bp3-overwrites.scss
@@ -184,9 +184,6 @@ $bp3-dialog-radius: 2px;
 
 .bp3-menu-item {
   padding: 5px 12px;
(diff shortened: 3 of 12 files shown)


scss/misc/_context_menu.scss
```diff
@@ -35,6 +35,7 @@
   .item {
     line-height: var(--local-menu-item-Line-height);
     padding: var(--local-menu-item-vertical-padding) 12px;
+    text-transform: capitalize;
```
r10s, line 38: i think, this can go wrong in many languages, not only german. even in english, words as "and" are not capitalized in menus.  i would remove auto-capitalize.

scss/overwrites/_bp3-overwrites.scss
```diff
@@ -184,6 +184,9 @@ $bp3-dialog-radius: 2px;
 
 .bp3-menu-item {
   padding: 5px 12px;
+  & > .bp3-fill {
+    text-transform: capitalize;
```
r10s, line 188: same here.

_locales/_untranslated_en.json
```diff
@@ -43,5 +43,8 @@
   },
   "copy_json": {
     "message": "Copy JSON"
+  },
+  "copy_text": {
```
r10s, line 47: maybe just use the word "Copy" for the default (obvious) action or if there is only one copy action.

"Copy link" etc. can be used for the more specialized ones and if there is more than one action.

r10s (changes requested):
to the ordering, i think, 

- the last item should be "Delete" for both menus.

- i also think, "Reply" and "Forward" belong together.

- for the message-context-menu-ordering, refining the suggestion from https://github.com/deltachat/deltachat-desktop/issues/1982, i would suggest
`Reply`  
`Reply privately`  
`Forward`  
`----------`  
`Copy`  
`[Copy link]`  
`[Download attachment]`  
`Message details`  
`----------`  
`Delete` (red)

- for the chat-context-menu, a reasonable order might be:
`Pin`  
`Archive`  
`Mute`  
`----------`  
`Edit group/View profile`   
`Show encryption info`   
`----------`  
`Leave Group/Block contact` (red)   
`Delete chat` (red)

if red items and horizontal lines are too complicated, this can also be postponed.

i request changes, mainly because of the `capitalize` thing :)

Simon-Laux:
There are two copy options depending on whether text is selected.
Either `copy message` or `copy selection`.

r10s:
> There are two copy options depending on whether text is selected.
> Either copy message or copy selection.

i know - but is that really important? if some text is selected, it is usual that the selection is copied.

but i would also be fine with differing between the options, np.

Jikstra:
@r10s applied your requested changes (besides the red and lines). Not sure if i like it more like this. But what annoys me most is the the repetetive usage of Message/Group/Chat in the translations. I think it's not needed. Right clicking on message doesn't need to tell me that the "Delete" will "Delete message", the "Forward" means "Forward message" and "Details" means "Message details". Same for the chat list things. Also we sometimes use "group", sometimes "chat".

r10s:
> But what annoys me most is the the repetetive usage of Message/Group/Chat in the translations

yip, i just had a look - we have short variants for most actions:

- for message menu: `notify_reply_button`, `reply_privately`, `forward` --- `global_menu_edit_copy_desktop`,  `menu_copy_link_to_clipboard`, `download_attachment_desktop`,  `message_details` --- `delete`

- for chat menu: `pin`, `archive`, `mute`, `copy` --- `menu_edit_group`/`profile`, `encryption_info_title_desktop` -- [`menu_leave_group`], `delete`

"Message details" would stay as is, however, maybe not that bad, to have some context here. we can rename that as well, maybe "Info" is better here, however, i would do/discuss that cross-platform and not in this pr. the whole dialog will be improved at some point anyway.

also, for "Delete", having the longer option is maybe better - it is the last item, does not add much noise and being explicit on an action that cannot be undone should at least be considered.

but even when leaving "Message details" and "Delete chat/message" as is, this would be a big overall improvement i'd say.

PR #1999, opened 2020-12-01
select message input field on chat selection

### LINKED ISSUE #1986: Selecting a chat doesn't focus composer input anymore

Simon-Laux:
the timeout seems to be needed
fixes #1986


Jikstra:
The setTimeout is a "hack" in js land to execute the callback on the next js event loop cycle.

PR #2001, opened 2020-12-03
call `maybeNetwork` when waking up from standby

Simon-Laux:
call `maybeNetwork` when waking up from standby (macOS and Windows only)

- [ ] Test if this improves anything

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Show linebreaks in quotes
 - trust all labeled links from device chat
 - Adjust order of buttons at labeled link dialog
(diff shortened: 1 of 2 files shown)


Jikstra:
I think @hpk42 and @link2xt reported this problem, and i doubt that either of them uses windows or mac, sooooo i also doubt that this pr fixes the bug for them. I would suggest adding logging for when we get a wakeup event and when we call maybeNetwork, so that we can debug it a bit better with the logs of them.

Simon-Laux:
> every time we focus the app.

probably a good idea

Simon-Laux:
Discussed this with Jikstra and he convinced me that #2002 is enough

Issue #1949, opened 2020-11-04
Search - select archived chat doesn't switch view


Simon-Laux:
When I select an archived chat and exit the search I'm not switched into the archived view so the chat does not appear in the chatlist.
Also occurs when opening an archived chat from the mutual-chats list in the profile view. 

A solution could be to switch the show archived stated var inside of the chatlist on chat selection.

Jikstra:
Hmm I don't really now what a good ux would be for this issue. Switching to archived chats when selecting an archived chat feels weird, unarchiving it when we select it in the search seems weird to me too.

Simon-Laux:
> unarchiving it when we select it in the search seems weird to me too

Who would propose something like that? :confused:

Simon-Laux:
Don't close issues that are not completely fixed yet

PR #2005, opened 2020-12-04
Switch to showing archived chats when selecting an archived chat in t…

Jikstra:
…he chat search and emptying the search

Fixes https://github.com/deltachat/deltachat-desktop/issues/1949

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Adjust order of buttons at labeled link dialog
 - Change order and ux of context menus, especially for messages
 - Increase padding of draft/reply area
+- Switch to showing archived chats when selecting an archived chat in the chat search and emptying the search
 
 ## Fixed
 - Fix "copy link" context menu option for labeled links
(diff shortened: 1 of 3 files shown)


Simon-Laux (changes requested):
There are also other methods of selecting an archived chat, mainly selection it from mutual chats.
So the switch to the archived view should happen on every chat selection if its archived.
Maybe listening for an event from the chat-store or triggering an chatlist event from the chatstore.

Jikstra:
> There are also other methods of selecting an archived chat, mainly selection it from mutual chats.
So the switch to the archived view should happen on every chat selection if its archived.
Maybe listening for an event from the chat-store or triggering an chatlist event from the chatstore.

Looked into it but it's not easy to fix without bigger refactoring. The showArchivedChats state is hold by MainScreen, and we need to change it from various places. Would postpone this for now, is also not a huuge problem.

Simon-Laux:
We could abuse the keybinding system (new `KeybindAction`)

PR #2007, opened 2020-12-04
Try to always have composer selected if possibl

Jikstra:
e, without preventing clicking on other inputs/selecting text

I would wait for merging and include this only in 1.15 as this probably causes new bugs?


Jikstra:
Lets get it merged, I don't like prs which are getting moldy. We can test this the next days from the master branch.

Simon-Laux:
I would wait until 1.14 is out...

src/renderer/components/message/MessageList.tsx
```diff
@@ -107,6 +107,11 @@ export default function MessageList({
     composerTextarea && composerTextarea.focus()
   }, [chat.id])
 
+  useEffect(() => {
+    const composerTextarea = refComposer.current.childNodes[1]
```
Simon-Laux, line 111: I don't quite like the usage of `childNodes[1]`, but well maybe there is no obvoius more elegant solution...

Simon-Laux:
double click for quoting doesn't refocus the composer

PR #2010, opened 2020-12-07
Allow importing of .tar backup

Jikstra:
Fixes https://github.com/deltachat/deltachat-desktop/issues/2006

See https://github.com/deltachat/deltachat-core-rust/pull/2023

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Adjust order of buttons at labeled link dialog
 - Change order and ux of context menus, especially for messages
 - Increase padding of draft/reply area
+- All import of .tar backup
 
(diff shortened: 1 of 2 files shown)


Jikstra:
@Hocuri do we need anything else for tar backups? I guess it will be part of core 1.51?

Hocuri:
If tar import works now, then that's the most important thing to avoid not being able to import a backup on Desktop that was exported on Android. It would be good to make a bugfix release after this PR is merged as the Android nightlies already export tar. 

And you should probably test whether tar export works when you build Desktop with the current core master (there were no breaking changes since the release, so that should be no problem). This way, we can avoid surprises when updating to the next core release.

PR #2011, opened 2020-12-07
Switch to archived/normal view when selecting chat

### LINKED ISSUE #1949: Search - select archived chat doesn't switch view

Simon-Laux:
depending whether its archived
fix #1949


Jikstra:
Can you add a comment that this is a nice hack but still a hack and we should probably have a chatlist store or a more general store?

Simon-Laux:
Why not rename the action emitter so its not only for keybindings?
I'm not sure about the whole store thingie, though I get that it would be nice to have such data in a central place.
But probably not in this pr.

src/renderer/components/MainScreen.tsx
```diff
@@ -31,15 +31,24 @@ import { getLastSelectedChatId } from '../ipc'
 import { useKeyBindingAction, KeybindAction } from '../keybindings'
 import { Avatar } from './Avatar'
 import OfflineToast from './OfflineToast'
+import { C } from 'deltachat-node/dist/constants'
 
 export default function MainScreen() {
   const [queryStr, setQueryStr] = useState('')
   const [media, setMedia] = useState(false)
   const [showArchivedChats, setShowArchivedChats] = useState(null)
+  useKeyBindingAction(KeybindAction.ChatList_SwitchToArchiveView, () =>
```
Jikstra, line 40:
```suggestion
    // Small hack/misuse of keyBindingAction to setShowArchivedChats from other components (especially
    // ViewProfile when selecting a shared chat/group)
    useKeyBindingAction(KeybindAction.ChatList_SwitchToArchiveView, () =>
```

Issue #1996, opened 2020-11-29
Draft area sometimes not cleared after sending message


Jikstra:
So the reply/attachment stays there for the chat.

Simon-Laux:
any more details on this? I thought I had this fixed.

Jikstra:
No :/ But I will investigate it and look if it happens again

Jikstra:
Didn't happen to me in a long time, probably fixed

PR #2014, opened 2020-12-08
[needs testing] try fixing the draft bug

Simon-Laux:
rearrange draft code a bit in hope that it deletes old draft on sending more reliable now
related to the bug #1996

with this I have not yet seen that bug again, but its hard to trigger anyway so it might still be in there.


Jikstra:
Let's test on master, a rare bug on a branch wont get tested ;)

PR #2032, opened 2021-01-03
replace parcel with esbuild

Simon-Laux:
remaining todo:
- [ ] Test on windows

```sh
# previous
$ time npm run build
real    0m53.432s
user    2m17.756s
sys     0m7.784s
$ time NODE_ENV=production npm run build
real    1m6.012s
user    3m21.269s
sys     0m7.955s
# now
$ time npm run build
real    0m22.582s
user    0m33.789s
sys     0m1.833s
$ time NODE_ENV=production npm run build
real    0m45.523s
user    1m9.489s
sys     0m2.584s
```


Jikstra:
I'm currently getting this error locally:
```
- Executing "npx esbuild tsc-dist/renderer/main.js --bundle --sourcemap --sources-content=false --outfile=html-dist/bundle.js --define:process.env.NODE_ENV=\"development\""
 > node_modules/react-virtualized/dist/es/WindowScroller/utils/onScroll.js: error: No matching export for import "bpfrpt_proptype_WindowScroller"
    75 │ import { bpfrpt_proptype_WindowScroller } from '../WindowScroller.js';
       ╵          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

1 error
npxexited with ERR CODE 1
(node:31554) UnhandledPromiseRejectionWarning: ERR CODE 1
(node:31554) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:31554) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.

```

Not sure why as it's not happening on the ci.

Jikstra:
What is the exact problem why we cannot fully use esbuild and depend on first using typescript and then esbuild?

Simon-Laux:
> What is the exact problem why we cannot fully use esbuild and depend on first using typescript and then esbuild?

1. esbuild does not check types, but we can use tsc --check for it.
2. esbuild does not fully support typescript.
3. esbuild does not compile `object?.property` yet so the browser can not use the result.

```
> - Executing "npx esbuild tsc-dist/renderer/main.js --bundle --sourcemap --sources-content=false --outfile=html-dist/bundle.js --define:process.env.NODE_ENV=\"development\""
 > node_modules/react-virtualized/dist/es/WindowScroller/utils/onScroll.js: error: No matching export for import "bpfrpt_proptype_WindowScroller"
    75 │ import { bpfrpt_proptype_WindowScroller } from '../WindowScroller.js';
       ╵          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

esbuild doesn't detect that this import is unused and tries to resolve it. Basically its an issue in the library. (locally for testing you can just delete that line, but I'll fork the repo and fix it there.

Simon-Laux:
Somehow this has a problem on windows, we should investigate this on a windows device before merging.

Jikstra:
> esbuild does not check types, but we can use tsc --check for it.
> esbuild does not fully support typescript.
> esbuild does not compile object?.property yet so the browser can not use the result.

Can we adjust the codebase to work with esbuild? I prefer to have super fast builds then using every feature of typescript

Simon-Laux:
> Can we adjust the codebase to work with esbuild? I prefer to have super fast builds then using every feature of typescript

First: Let's already merge this, we can think about saving even more time later.

Well I prefer the real typescript instead of not using important typescript features (and who knows what other quirks will need to be followed) to just to turn like 20 seconds into 2 seconds.
```ts
if (msgRef?.current?.property){
// stuff    
}
```
turns into:
```js
if ((_a = msgRef === null || msgRef === void 0 ? void 0 : msgRef.current) === null || _a === void 0 ? void 0 : _a.property) {
    // stuff    
}
```
you don't want that in the code you write, do you?

If you really want something good and fast you could dig into how to use https://swc.rs/. The documentation of their bundler is not the best yet, but If we wanna patch something its easier for us if its in rust than in go, right?
Otherwise we could create an issue on the esbuild github and an wait until someone that knows how esbuild works fixes it.

PR #2043, opened 2021-01-12
Include git branch into development version string

Jikstra:
Screenshot_20210211_124138

### DIFF
--- a/bin/build-shared-version-info.js
+++ b/bin/build-shared-version-info.js
@@ -2,14 +2,25 @@ const { spawnSync } = require('child_process')
 const fs = require('fs-extra')
 const { join } = require('path')
 
+function gatherProcessStdout(cmd, args) {
+  const { status, stdout, stderr } = spawnSync(cmd, args)
+  if (status !== 0) throw new Err(stderr)
+  return stdout.toString().replace(/\n/g, '')
+}
+
 async function gatherBuildInfo() {
-  const { status, stdout, stderr } = spawnSync('git', ['describe'])
-  let git_ref = "None"
-  if (status !== 0) {
-    console.log(stderr)
-  } else {
-    git_ref = stdout.toString().replace(/\n/g, '')
+  let git_describe, git_branch;
+  try {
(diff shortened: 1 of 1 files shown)


Simon-Laux (approved):

bin/build-shared-version-info.js
```diff
@@ -2,14 +2,25 @@ const { spawnSync } = require('child_process')
 const fs = require('fs-extra')
 const { join } = require('path')
 
+function gatherProcessStdout(cmd, args) {
+  const { status, stdout, stderr } = spawnSync(cmd, args)
+  if (status !== 0) throw new Err(stderr)
```
Simon-Laux, line 7:
```suggestion
  if (status !== 0) throw new Error(stderr)
```

Simon-Laux (changes requested):
Looks like we need to investigate why the ci doesn't like it first:
```
Error: fatal: ref HEAD is not a symbolic ref

    at gatherProcessStdout (/home/runner/work/deltachat-desktop/deltachat-desktop/bin/build-shared-version-info.js:7:27)
    at gatherBuildInfo (/home/runner/work/deltachat-desktop/deltachat-desktop/bin/build-shared-version-info.js:15:18)
    at Object.<anonymous> (/home/runner/work/deltachat-desktop/deltachat-desktop/bin/build-shared-version-info.js:34:1)
    at Module._compile (internal/modules/cjs/loader.js:999:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
    at Module.load (internal/modules/cjs/loader.js:863:32)
    at Function.Module._load (internal/modules/cjs/loader.js:708:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:60:12)
    at internal/main/run_main_module.js:17:47
```
Maybe we can try a second strategy if getting the branch this way fails, like fetching the branch from github ci vars.

Jikstra:
Are we on the ci probably shallow cloning the branch for quicker checkouts? If I remember it correctly it fetches only the latest commit without the history. Probably this causes problems.

Issue #2040, opened 2021-01-10
.deb not running on latest debian anymore


Jikstra:
See https://support.delta.chat/t/debian-package-fails-to-install-on-testing-bullseye-nonexisting-dependency/1364 for details. I guess we have to update electron.

Simon-Laux:
I guess the problem requires an electron builder update.
Also I'm asking myself if changing this breaks backwards compatibility and whether we need to provide multiple `.deb` files for the different systems.

stmllr:
If you need testers, please ping me

Jikstra:
@stmllr okay so i looked into it. We would need to have two .deb files, one for bullseye and later, one for the previous ones. Or start finding people who want to maintain ppas. Trying to make the libappindicator libraries optional didn't help, at least i'm getting a glibc error (which is somehow weird).

vali-um:
> I guess the problem requires an electron builder update.
> Also I'm asking myself if changing this breaks backwards compatibility and whether we need to provide multiple `.deb` files for the different systems.

there is an upstream bug reported for this against electron in hope of a solution on their side: electron/electron#27527

eighthave:
I think a PPA would help here, this would also provide arm64 support for Debian/Ubuntu.  I would be happy to setup a Launchpad PPA for DeltaChat and serve as technical support for it.  I can't promise I would be able to upload the releases to it on an ongoing basis.

Jikstra:
> I think a PPA would help here, this would also provide arm64 support for Debian/Ubuntu. I would be happy to setup a Launchpad PPA for DeltaChat and serve as technical support for it. I can't promise I would be able to upload the releases to it on an ongoing basis. I do think the process can be scripted and integrated into an existing DeltaChat release process.

This would be awesome! For arm64 we would need some more ci on deltachat-node. I think too that we can script it. I sent you a mail on the address shown in your github profile :)

mimi89999:
> I think a PPA would help here, this would also provide arm64 support for Debian/Ubuntu. I would be happy to setup a Launchpad PPA for DeltaChat and serve as technical support for it. I can't promise I would be able to upload the releases to it on an ongoing basis. I do think the process can be scripted and integrated into an existing DeltaChat release process.

@eighthave I don't think that it's the right solution. If it was removed from Debian, there was probably a very good reason for that.

eighthave:
DeltaChat was removed from Debian?  If so, I wasn't aware of that.  As far as I 
[...]

PR #2053, opened 2021-01-13
[wip] Fix dependencies for debian build

Jikstra:
Fixes https://github.com/deltachat/deltachat-desktop/issues/2040

Cannot test it currently as the electron-builder process never finishes...


mimi89999:
I think that the correct fix is https://github.com/develar/app-builder/pull/53

PR #2055, opened 2021-01-13
fix hard to alter text in the middle of "Video chat instance" input field

### LINKED ISSUE #2016: hard to alter text in the middle of "Video chat instance" input field
assume, "Settings / Video chat instance" is not empty and you want to enter sth. in between, say change the domain or a parameter.

if you put the cursor somewhere in the middle of the "Settings / Video chat instance" input field and enter a character, the cursor moves to the end. this makes it very hard to alter the string.

Simon-Laux:
fix #2016

### DIFF
--- a/src/renderer/components/dialogs/Settings.tsx
+++ b/src/renderer/components/dialogs/Settings.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, useContext } from 'react'
+import React, { useState, useEffect, useContext, useRef } from 'react'
 import { DeltaBackend } from '../../delta-remote'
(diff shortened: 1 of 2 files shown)


src/renderer/components/dialogs/Settings.tsx
```diff
@@ -57,6 +57,50 @@ export function SettingsSelector(props: any) {
   )
 }
 
+function DeltaInput({
```
Jikstra, line 60: Hmm we already have a component called DeltaInput in LoginStyles. Can you either fuse them or give it nother name?

PR #2058, opened 2021-01-14
🚧 Upgrade electron to version 11

Simon-Laux:
- Upgrade electron to version 11 and refactor some stuff to make it work.
- update testcafe
- use only esbuild for frontend building
- add new ci check
### TODO

- [x] fix stickers are broken
- [x] fix background image is broken
- [x] fix Help is broken
- [X]  fix Calls are broken,  but its broken
(was already broken in current version, though)
- [x] preload function download -> move to runtime
- [x] sourcemaps are broken -> devtools isn't even asking the backend via the `dc://` scheme for the file. and it says unexpected end of json, but only in debug mode. -> solved with workaround (putting them inline for now)
- [X] fix copy text to clipboard -> move to runtime
- [ ] ~~optional: find solution to fix react devtools (it's broken by electron issues)~~
- [x] fix testcafe (can't connect)

### Testing Points
These areas were changed specifically so we should especially test them (after the todo points above are completed):
- [ ] key import (especially the confirmation dialog)
- [ ] key export (especially the confirmation dialog)
- [ ] select profile picture
- [ ] set group image
- [ ] import backup
- [ ] attach file to message
- [ ] testcafe
- [ ] calls
- [ ] help
- [x] downloadFile
- [ ] attachment and all places where blobs are displayed
- [ ] stickers
- [ ] background image
- [ ] copy to / read from clipboard via clicks (not the key-combinations)


Jikstra:
Bugs I found so far:
- [ ] Setting a new profile picture works (i think) but doesn't show the picture properly when selected
- [ ] opening map fails with `TypeError: Cannot read property 'getZoom' of undefined`

Simon-Laux:
> opening map fails with TypeError: Cannot read property 'getZoom' of undefined

can't reproduce and I think this wasn't introduced by this pr. But I can add some checks to that logic.

PR #2077, opened 2021-01-28
Remove call to removed core api isIORunning()

### LINKED ISSUE #2049: Stop using isIORunning()
If I/O should be restarted, stop and start without checking.

https://github.com/deltachat/deltachat-core-rust/issues/2052

Jikstra:
Somehow depends on https://github.com/deltachat/deltachat-node/pull/494

Fixes https://github.com/deltachat/deltachat-desktop/issues/2049


r10s (approved):
lgtm, i think, this can be merged when still using core50, isIORunning() is deprecated and not needed since 6 months or so.

PR #2080, opened 2021-01-29
Fix running on windows

Jikstra:


src/main/internal-app-schemes.ts
```diff
@@ -88,23 +87,13 @@ app.once('ready', () => {
...
     const otherFolder = ALLOWED_FOLDERS.findIndex(folder => {
-      console.log(
-        process.platform,
-        file,
-        folder + '\\',
-        file.startsWith(folder + '\\')
-      )
       return process.platform === 'win32'
         ? file.startsWith(folder + '\\')
         : file.startsWith(folder + '/')
```
Simon-Laux, line 92: why not use `path.sep` https://nodejs.org/api/all.html#path_path_sep?

Jikstra: Oooh good idea :)

static/fix-missing-vars.js
```diff
@@ -1,2 +1,3 @@
 const process = {env:{}}
-const global = {}
\ No newline at end of file
+const global = {}
+const development = 'production'
```
Jikstra, line 3: I'm concerned about this @Simon-Laux what do you think? In theory this should change depending on production or development build.

Simon-Laux: why did you add it here?

Jikstra: That was what i was looking for. Didn't find it when i was scrolling through the esbuild docs :)

static/fix-missing-vars.js
```diff
@@ -1,2 +1,3 @@
 const process = {env:{}}
-const global = {}
\ No newline at end of file
+const global = {}
+const development = process.env.NODE_ENV
```
Simon-Laux, line 3: who uses this `development` - variable?

Jikstra: React threw an error on windows without having this. I don't know why. Adding it and everything was good.

Simon-Laux: now it is a string, but shouldn't it be a boolean?

Jikstra: Nope should be a string

Issue #2045, opened 2021-01-12
Make sure that stacktraces in backend and frontend use source maps for line hints


Jikstra:
Before esbuild stacktraces used `build.js:39109283` which is obviously super unhelpful. I would love to have something like `src/frontend/.../foobar.tsx:129`

Issue #2082, opened 2021-01-29
Crash when opening settings


the-voidl:
- Operating System: Linux (Manjaro)
- Delta Chat Version: v.1.14.1 (git: v1.14.1-140-gb7b52dbc), installed via yay from the AUR
- Expected behavior: Open the settings
- Actual behavior: Crash (Actually DC survives and displays an error)
- Steps to reproduce the problem: Go to "... menu" -> "Settings"
- Logs:
```
TypeError: Cannot read property 'startsWith' of undefined
    at dc://deltachat/bundle.js:102317:56
    at updateContextConsumer (dc://deltachat/bundle.js:16089:27)
    at beginWork$1 (dc://deltachat/bundle.js:16330:22)
    at HTMLUnknownElement.callCallback2 (dc://deltachat/bundle.js:2766:22)
    at Object.invokeGuardedCallbackDev (dc://deltachat/bundle.js:2791:24)
    at invokeGuardedCallback (dc://deltachat/bundle.js:2822:39)
    at beginWork$$1 (dc://deltachat/bundle.js:19958:15)
    at performUnitOfWork (dc://deltachat/bundle.js:19262:20)
    at workLoopSync (dc://deltachat/bundle.js:19244:30)
    at performSyncWorkOnRoot (dc://deltachat/bundle.js:18957:19)
```
```
2021-01-29T21:40:51.596Z	renderer/react-crashhandler	ERROR	"
    CrashScreen.componentDidCatch (dc://deltachat/bundle.js:104418:13),
    CrashScreen.callback (dc://deltachat/bundle.js:18402:20),
    callCallback (dc://deltachat/bundle.js:11870:20),
    commitUpdateEffects (dc://deltachat/bundle.js:11896:15),
    commitUpdateQueue (dc://deltachat/bundle.js:11888:11),
    commitLifeCycles (dc://deltachat/bundle.js:17606:17),
    commitLayoutEffects (dc://deltachat/bundle.js:19669:15),
    HTMLUnknownElement.callCallback2 (dc://deltachat/bundle.js:2766:22)"	"The app encountered an react error"	{}
```

Issue #2068, opened 2021-01-20
React Devtools are broken


Simon-Laux:
with the electron 11 upgrade (#2058), the react devtools are no longer working.

https://github.com/MarshallOfSound/electron-devtools-installer/issues/149

I thought about switching to https://github.com/dimitarnestorov/react-devtools-electron, but they also have an similar issue: https://github.com/dimitarnestorov/react-devtools-electron/issues/162

I tracked it back to https://github.com/electron/electron/issues/24011 which seems to be the issue blocking it.

Update: looks like electron fixed this in newer versions (with an additional flag: `allowFileAccess` option in `loadExtension()`)

Simon-Laux:
no actionable item currently for us, so moving to resurrection

Issue #2067, opened 2021-01-20
Find a way to get Sourcemaps working without inlining them


Simon-Laux:
with the electron 11 upgrade (#2058), the electron DevTools don't load the source-maps anymore. I couldn't find a quick way to fix it so I decided to inline them, but that is not good for loading time for folks with slow disks.

I suspect the devtools aren't able to load the sourcemaps via the `dc://` scheme, because I logged each request to that scheme and the sourcmaps did not show up in them.
We could also rewrite to sourcemaps in the file during building, but AFAIK the devtools don't have permission to use file:// either.

related electron issue: https://github.com/electron/electron/issues/21699

PR #2083, opened 2021-01-29
Fix using source maps

Jikstra:
Maybe fixes _ https://github.com/deltachat/deltachat-desktop/issues/2067
Fixes https://github.com/deltachat/deltachat-desktop/issues/2045

NOTE: This embeds/inlines the source maps for frontend js and css. This means start up times are probably slower. But it seems like dev tools cannot use the dc-file protocol and i guess it's the same bug in electron as with https://github.com/deltachat/deltachat-desktop/issues/2068 . But I prefer having source maps in slow then none, as otherwise fixing bugs like https://github.com/deltachat/deltachat-desktop/issues/2082 is a pain.


Jikstra:
Maybe on top of https://github.com/deltachat/deltachat-desktop/issues/2074 this could work again?

Jikstra:
Need sourcemaps on master, we can still change this/improve this.

PR #2086, opened 2021-01-30
Fix set avatar

### LINKED ISSUE #2074: Setting Group or profile image doesn't show the new image
this is because of the electron upgrade pr, paths will need to be adjusted somehow.

Jikstra:
Fixes https://github.com/deltachat/deltachat-desktop/issues/2074

### DIFF
--- a/src/main/internal-app-schemes.ts
+++ b/src/main/internal-app-schemes.ts
@@ -28,6 +28,11 @@ const HTML_DIST_DIR = htmlDistDir()
 const ACCOUNTS_DIR = getAccountsPath()
 
 app.once('ready', () => {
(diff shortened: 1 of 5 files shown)


Jikstra:
I guess this would allow us to remove all those other custom url schemes? @Simon-Laux what do you think?

Simon-Laux:
I'm not really in favor of allowing the renderer to access any file of the system.

Jikstra:
> I'm not really in favor of allowing the renderer to access any file of the system.

That's fine, I also don't like it. So we read in the file through javascript, but how do we send it to the backend? If we use raw browser apis, we don't get the full path to the file. So we have to implement a way to the send the file over ipc/websocket.

PR #2089, opened 2021-02-01
Move mark messages read logic to getMessages

Jikstra:


Jikstra:
I removed the setBadgeCount logic as it was broken anyways. I want to properly implement it for all platforms, showing the notification count on top of the tray icon (on mac of course still on the dash icon).

PR #2098, opened 2021-02-01
remove integrated basicwebrtc

### LINKED ISSUE #2059: Basicwebrtc (integrated) calls are broken
basicwebrtc update is needed,
when updating make sure that the new version implements a proper version check.

screenshare is also broken.

Simon-Laux:
because we currently lack the resources to fix and maintain it. 
this closes #2059


cracker0dks (approved):
can't say much about the code changes but it is the right way to remove it for now

PR #2100, opened 2021-02-02
Add electron context menu

### LINKED ISSUE #2047: no context menu on input fields
normally, a right-click in an input field shows a system-context-menu with cut/copy/paste and maybe other system-special things (spell correction etc).

currently, this system-context-menu does not appear - but also, no replacement that may be done by chrome appears.

not sure, if the system-context-menu is doable, however, having a menu with at least "paste" is expected in many workflows. i just checked signal, who are also using electron afaik, and there is a context menu.

maybe there is just a return value from some event handler wrong? so that passing the click to the default handler is skipped?

(tested on macos with delta 1.14.1)

Simon-Laux:
This adds an electron context menu as **fallback if no other context menu is set**.
fixes #2047


Jikstra:
Why adding the electron context menu if we could also show our own? Not only does this tie us more to electron, but also the context menu looks different then everywhere else?

Simon-Laux:
Browsers have their own context menu, this is only a poly-fill so electron has one too.
So `Not only does this tie us more to electron,` is not true.
We can talk about consistent styling and somehow more or less complicated use our menu implementation for this fallback menu, but that won't work in the browser, because we don't have the api's there to call the browser copy/paste functions there directly and I wouldn't expect there to be an api to get spellchecking suggestions to put it into our own menu either.

Jikstra:
> So Not only does this tie us more to electron, is not true.

Okay, that's good.

> because we don't have the api's there to call the browser copy/paste functions there directly

Talking about truth, i think there's another truth too, at least here on stackoverflow: https://stackoverflow.com/questions/5579232/simple-copy-paste-function-in-javascript

Simon-Laux:
I know there are apis to modify the clipboard, but we need to build some selection logic around it, so means that we implement copy selection, paste in text at position and cut out of text ourselves, so technically its true that we can do those simpler things. though the spellchecker issue remains.

PR #2110, opened 2021-02-09
Update failed messages in chat store

### LINKED ISSUE #1361: Failed message gets wrongly displayed.
After leaving this chat and reentering it the message gets displayed as expected

And when clicking on the wrongly displayed one I get a white screen (react error):

### LINKED ISSUE #2101: Failed messages' state does not get updated in message view
if a message fails, the message state isn't updated immediately in the message list for that message (to trigger an error you can just send a too big attachment)
Though it already gets properly updated in the chat-list so this might be just an event-listener missing or sth. like that.

Jikstra:
Fixes https://github.com/deltachat/deltachat-desktop/issues/2101
Fixes https://github.com/deltachat/deltachat-desktop/issues/1361

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@
 - Fix copy of labeled links
 - Fix hard to alter text in the middle of "Video chat instance" input field (see #2016)
 - Add missing options (at once, one minute...) for autodelete from server
+- Update failed messages also in messagelist
 
 ## [1.14.1] - 2020-12-15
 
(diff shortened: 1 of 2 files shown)


Simon-Laux (dismissed):
There is still an issue, when sending a too big file. after reentering the chat its displayed correctly.

Jikstra:
But then this error doesn't emit a DC_EVENT_MSG_FAILED event.

Simon-Laux:
probably https://github.com/deltachat/deltachat-desktop/issues/1361

Issue #1977, opened 2020-11-25
forwarded-hint is misleading


r10s:
currently, the "forwarded message" title is shown above the name of the sender:

![Screen Shot 2020-11-25 at 00 59 19](https://user-images.githubusercontent.com/9800740/100165072-97784080-2eb9-11eb-9ba7-df51ac4ae4e5.png)

the example above looks as if holger wrote "WTF!!!!" and this message was forwarded by someone.

in fact, holger did not wrote "WTF!!!!" - holger forwarded a message with this text.

so, the two titles should at least be swapped. ~~to make it even clearer, i suggest to strike the name completely - just say "forwarded message" - this is what we are doing on ios/android and what eg. also whatsapp is doing~~ EDIT: we are about to fine-tune that, see below.

Simon-Laux:
> so, the two titles should at least be swapped. to make it even clearer, i suggest to strike the name completely - just say "forwarded message" - this is what we are doing on ios/android and what eg. also whatsapp is doing.

Then who forwarded the message in a group?

We could maybe do something "holger has forwarded this message" or a similar combined thing. or something different like really swapping it and maybe a some subtle ux indication.

r10s:
> Then who forwarded the message in a group?

there is still the avatar, this is enough.

> We could maybe do something "holger has forwarded this message" or a similar combined thing. or something different like really swapping it and maybe a some subtle ux indication.

i would just strike the name and replace it by "Forwarded message", not swap. we are doing this one android since "forever", there were no issues with that.

r10s:
i suggest to target https://github.com/deltachat/deltachat-android/issues/1743 (quoting forwarded message no longer look as beeing forwarded ) when on that as well, **all in all**, i suggest the following:

1. define some unknown_sender grey color, on android we use #99999 for that, that fits well in light mode as well as in dark mode and has a similar brightness as the contact/chat colors (that also do not differ between light and darkmode)

2. when displaying a forwarded message, display the string "Forwarded message"  in unknown_sender grey above the message. do not display the name of the forwarder above the message in groups (there is still the avatar and message-info to get these details)

3. when a quote is displayed, and the quote is a forwarded message, do the same: display "Forwarded message" instead of the forwarder name. bar and string should be in unknown_sender grey. it is a bit unfortunate, that "jump to original" is not yet implemented, however, as for 5. the sender sees exactly the same on composing and this is a cornercase anyway, it is fine. if would vote for simple and unique display across platforms here.

[...]

PR #2111, opened 2021-02-09
Indicate who forwarded a message in groups

Jikstra:
Screenshot_20210210_105056
Screenshot_20210210_105041

Fixes https://github.com/deltachat/deltachat-desktop/issues/1977


Simon-Laux (changes requested):
LGTM, but don't forget to update the part in the change-log to represent the current state of the translation update:
```
- update translations (mid dezember 2020)
```

r10s:
nice! please show the message also in a forwarding-grey color as suggested in #1977
- to make it more clear, that this is not a message that was "typed in" 
- to let groups and one-to-ones stay visually different (we do not have the user-colors in one-to-ones)
- to stay visually compatible with android/ios

Jikstra:
@r10s so you mean showing the "Forwarded by..." line in the grey instead of the user color? It hurts a little bit my eyes, but for sure it sticks out :P

r10s:
> so you mean showing the "Forwarded by..." line in the grey instead of the user color? It hurts a little bit my eyes, but for sure it sticks out :P

yip. tried to explain it in some details :)  sorry for your eyes :) but i think, a bit of grey make the remaining colors even more beautiful 🌹

and in addition, i would stay with the string "forwarded message" in one-to-one chats, also in grey then (not sure if that is changed by this pr as well)

PR #2126, opened 2021-02-15
Display contact status in the profile

### LINKED ISSUE #2113: Display contact status in the profile
See related core PR: https://github.com/deltachat/deltachat-core-rust/pull/2218

New API is `dc_contact_get_status`

Simon-Laux:
closes #2113

@Jikstra plz make it pretty 🥺

@Simon-Laux okay!
Screenshot_20210508_050408


r10s:
> @Jikstra plz make it pretty 🥺

@Jikstra my rough idea for android/ios is to show in in a "post it" background-color (very lite yellow) below name and address, however, i do not have a strong opinion on that, more important is that the system look similar :) if you like, we can have an a/v on that tomorrow  (now i am more or less afk)

r10s:
i created a mockup, that is similar to the one we [agreed on on android](https://github.com/deltachat/deltachat-android/pull/1805):

Screen Shot 2021-03-09 at 15 23 18

- the current pr shows the signature-text (aka status) between address and "Send message", however, that is not so nice as the "send message" button, which is a central element, would no longer be at the same position - and may even scroll out of screen. therefore, the signature-text should go below the send button - there is also more room in width.

- show a title just re-use the "Shared chats" header layout for that, text for the title is `pref_default_status_label` 

- the signature-text should be shown in foreground and background colors of an incoming message, so black on white

- it would be great, if links are clickable as usual and the text can be selected and copied to the clipboard

- wrt scrolling: the element can grow in height as the whole dialog is scrollable anyway (the "shared chats" list will move down, but that is okay)

we might change the wording at some point, however,  that should be done on all apps then, for now, it is good to use the existing wording and also to tune the whole stuff down a bit (as it is currently a mix of footer/signature/status anyway)

PR #2132, opened 2021-02-17
[Feature] mailing list support

### LINKED ISSUE #2128: add mailinglist and bot-impersonation support
bot-impersonation is part of mailinglist-support, therefore, there is only one issue for both.

things to be done for mailinglist and bot-impersonation:

- if `dc_msg_get_overriden_sender_name()` returns != NULL, that name shall be used instead of `dc_contact_get_display_name()`

- in the bubbles, the overridden names shall be prefixed by a `~`, also colored. in summaries, notifications, where the user cannot access information to the contact directly, this is not needed, see https://github.com/deltachat/deltachat-android/pull/1808 for an overview of past discussions, reasoning and also a screenshot

things to be done for mailinglists:

- there is a new chattype, DC_CHAT_TYPE_MAILINGLIST, that needs to be shown mostly similar to groups

- subtitle for DC_CHAT_TYPE_MAILINGLIST is string `mailing_list`

- for mailinglists, hide unused options from mailinglist profile (disappearing/memberlist/leave) - might be that the memberlist will reappear, however, that is not decided yet. currently, it cannot be accessed.  
we decided to allow editing names/avatars for now, even if they are not synced, it is useful as there is still some chance we do not detect the mailinglist name correctly.

adaption to new contact-request is also needed, however, that is already done in #2127

related issues https://github.com/deltachat/deltachat-ios/pull/1051 (there is a screenshot of the ios profile), https://github.com/deltachat/deltachat-core-rust/pull/1964, https://github.com/deltachat/deltachat-core-rust/pull/1964 and more linked from there.

### LINKED ISSUE #2215: add alias support
once https://github.com/deltachat/deltachat-core-rust/pull/2297 (EDIT: done in core52) is merged, we need to check for [`dc_msg_get_override_sender_name()`](https://c.delta.chat/classdc__msg__t.html#a47c8d53848c954277926f456919c7eab) even in one-to-one-chats.

if there is an override-sender-name, _and only then_, the name prefixed by a `~` and the avatar should be shown (as for groups and mailinglists).

if there is no override-sender-name, name and avatar should stay hidden for one-to-one-chats.

see also: https://github.com/deltachat/deltachat-android/pull/1840, https://github.com/deltachat/deltachat-ios/issues/1145

Simon-Laux:
- [X] subtitle
 - [x] mailing-list profile (currently it opens the profile view for the self contact)
   - [x] mailing list don't show members yet - so hide them
   - [x] basically like group profile view but with hidden members section. ("we decided to allow editing names/avatars for now, even if they are not synced, it is useful as there is still some chance we do not detect the mailing-list name correctly." ~ r10s on #2128)
 - [x] disable composer for mailing lists with message (with short explanation text)
 - [X] Impersonation (prefix with `~`)
 - [x] add line to changelog

closes #2128
closes #2215

### DIFF
--- a/src/renderer/components/message/MessageList.tsx
+++ b/src/renderer/components/message/MessageList.tsx
@@ -3,6 +3,7 @@ import { MessageWrapper } from './MessageWrapper'
 import { useChatStore, ChatStoreState } from '../../stores/chat'
 import { useDebouncedCallback } from 'use-debounce'
 import { C } from 'deltachat-node/dist/constants'
+import type { ChatTypes } from 'deltachat-node'
 import moment from 'moment'
 
 import { getLogger } from '../../../shared/logger'
@@ -149,6 +150,14 @@ export default function MessageList({
   )
 }
 
+/** Object holding type information about a chat for messages in that chat */
+export type ConversationType = {
+  /* whether this chat has multiple participants */
+  hasMultipleParticipants: boolean
+  isDeviceChat: boolean
+  chatType: ChatTypes
+}
+
 export const MessageListInner = React.memo(
   (props: {
     onScroll: (event: React.UIEvent<HTMLDivElement>) => void
@@ -175,8 +184,13 @@ export const MessageListInner = React.memo(
 
     let specialMessageIdCounter = 0
 
-    const conversationType: 'group' | 'direct' =
-      chat.type === C.DC_CHAT_TYPE_GROUP ? 'group' : 'direct'
+    const conversationType: ConversationType = {
+      hasMultipleParticipants:
+        chat.type === C.DC_CHAT_TYPE_GROUP ||
+        chat.type === C.DC_CHAT_TYPE_MAILINGLIST,
+      isDeviceChat: chat.isDeviceChat,
+      chatType: chat.type,
+    }
 
     return (
       <div id='message-list' ref={messageListRef} onScroll={onScroll}>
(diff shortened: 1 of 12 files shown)


r10s:
great that that gets into shape!

can we have the "Show full message in browser" button a bit nicer and also a bit more tuned down? 

see https://github.com/deltachat/deltachat-android/issues/1895 for an ios screenshot, the corresponding android-pr is at https://github.com/deltachat/deltachat-android/pull/1897

excerpt from the reasoning: as the button is shown for many messages, and does not always offer really additional content (and we cannot really say if that is the case or not), it is better to tune the button down a bit, let them appear as a usual text link. the button should not get too much attraction, depending on the chat, it may be shown inside nearly every bubble.

Simon-Laux:
> can we have the "Show full message in browser" button a bit nicer and also a bit more tuned down?

this is not part of the mailing-list pr, I moved it to #2250

src/renderer/components/message/MessageList.tsx
```diff
@@ -149,6 +150,14 @@ export default function MessageList({
   )
 }
 
+/** Object holding type information about a chat for messages in that chat */
+export type ConversationType = {
```
Jikstra, line 154: Should this type distinguish between device chat, one to one chat, group chat...?

Simon-Laux: The reason for it is rather to provide some information about the chat to the messages without passing all of the `FullChat` type down.
Essentially yes at the moment it only provides information about the chattype.

PR #2146, opened 2021-02-26
update translations and add new languages.

### LINKED ISSUE #2137: add Khmer, Kurdish and more languages
just a reminder, not urgent or so:

- Khmer and Kurdish are new and pretty well translated
- Sardu, Slovak, Norwegian are a bit older, however seems to be missing

a complete list, including english and local names is at  https://github.com/deltachat/deltachat-android/blob/master/res/values/arrays.xml#L8

Simon-Laux:
closes #2137


_locales/_languages.json
```diff
@@ -32,4 +37,4 @@
   "uk": "Українська",
   "zh_CN": "简体中文",
   "zh_TW": "繁體中文"
-}
+}
```
r10s, line 40: is that the order how it appears in the ui? if so, i would change to to alphabetic order inside the latin alphabet, followed by the other alphabets.

that is probably also not perfect, however, a bit less random (eg. in 1.15 the two cyrillic languages appear at quite different positions, without an obvious rule).

maybe just use the same order as on android.

r10s (commented):
apart from my comment wrt sorting, this lgtm. did not test it, though :)

_locales/_languages.json
Simon-Laux: no, the sorting is done at runtime in https://github.com/deltachat/deltachat-desktop/blob/master/src/main/menu.ts#L32

PR #2153, opened 2021-03-02
pause io while exporting an backup

### LINKED ISSUE #2148: Call stop_io()/start_io() before/after backup export
See https://github.com/deltachat/deltachat-core-rust/pull/2255

For Android, I already made a PR: https://github.com/deltachat/deltachat-android/pull/1822/files

For backup import it's not necessary as IO is not running before doing an import.

For key import/export, it doesn't hurt but isn't necessary.

Simon-Laux:
closes #2148

### DIFF
--- a/src/main/deltachat/backup.ts
+++ b/src/main/deltachat/backup.ts
@@ -15,7 +15,12 @@ import SplitOut from './splitout'
 import { DeltaChatAccount } from '../../shared/shared-types'
 export default class DCBackup extends SplitOut {
   export(dir: string) {
+    this._dc.stopIO()
     this._dc.importExport(C.DC_IMEX_EXPORT_BACKUP, dir, undefined)
+    // look to when it is done
+    this._dc.once('DC_EVENT_IMEX_FILE_WRITTEN', () => {
+      this._dc.startIO()
+    })
   }
 
   import(file: string): Promise<DeltaChatAccount> {
--- a/src/main/deltachat/backup.ts
+++ b/src/main/deltachat/backup.ts
@@ -14,12 +14,28 @@ const log = getLogger('main/deltachat/backup')
 import SplitOut from './splitout'
 import { DeltaChatAccount } from '../../shared/shared-types'
 export default class DCBackup extends SplitOut {
-  export(dir: string) {
+  async export(dir: string) {
     this._dc.stopIO()
-    this._dc.importExport(C.DC_IMEX_EXPORT_BACKUP, dir, undefined)
-    // look to when it is done
-    this._dc.once('DC_EVENT_IMEX_FILE_WRITTEN', () => {
+    try {
+      await this._internal_export(dir)
+    } catch (err) {
       this._dc.startIO()
+      throw err
+    } finally {
+      this._dc.startIO()
+    }
+  }
+
+  private async _internal_export(dir: string) {
+    return new Promise((resolve, reject) => {
+      this._dc.importExport(C.DC_IMEX_EXPORT_BACKUP, dir, undefined)
+      this._dc.on('DC_EVENT_IMEX_PROGRESS', (data1: number) => {
+        if (data1 === 0) {
+          reject('Backup export failed (progress==0)')
+        } else if (data1 === 1000) {
+          resolve()
+        }
+      })
     })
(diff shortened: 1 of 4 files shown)


src/main/deltachat/backup.ts
```diff
@@ -15,7 +15,12 @@ import SplitOut from './splitout'
 import { DeltaChatAccount } from '../../shared/shared-types'
 export default class DCBackup extends SplitOut {
   export(dir: string) {
+    this._dc.stopIO()
     this._dc.importExport(C.DC_IMEX_EXPORT_BACKUP, dir, undefined)
+    // look to when it is done
+    this._dc.once('DC_EVENT_IMEX_FILE_WRITTEN', () => {
```
r10s, line 21: by definition, [`DC_EVENT_IMEX_FILE_WRITTEN`](https://c.delta.chat/group__DC__EVENT.html#ga7ffbea55be6a5e6da7ac7e35ba6bf985) does not say that the backup is complete, it just just says that _one_ file is written - so using that as a signal that the backup is complete may cause problems in the future. also, it is not emitted in case of errors, in this case, the app would be continued without io started.

instead, better listen to  [`DC_EVENT_IMEX_PROGRESS(data1=1000)`](https://c.delta.chat/group__DC__EVENT.html#ga006ea41d9c1a76ffc672752484c61e6c) that indicated success by definition. moreover, startIo should also be called if imex fails (indicated by `DC_EVENT_IMEX_PROGRESS(data1=0)`)

r10s (commented):
i cannot say much to the javascript peculiarities, but logic-wise, wrt using IMEX_PROGRESS instead of FILE_WRITTEN, this looks good to me now.

Jikstra:
Should this be in 1.15.2 or do we need a new core?

Simon-Laux:
yes, see https://github.com/deltachat/deltachat-desktop/issues/2148#issuecomment-789093673

src/main/deltachat/backup.ts
```diff
@@ -14,8 +14,29 @@ const log = getLogger('main/deltachat/backup')
...
-    this._dc.importExport(C.DC_IMEX_EXPORT_BACKUP, dir, undefined)
+  async export(dir: string) {
+    this._dc.stopIO()
+    try {
+      await this._internal_export(dir)
+    } catch (err) {
+      this._dc.startIO()
+      throw err
+    } finally {
+      this._dc.startIO()
```
Jikstra, line 25: Why put it in a finally statement? Is somehow confusing to me. I guess it won't get executed if you throw an err in the catch branch?

src/main/deltachat/backup.ts
```diff
@@ -14,8 +14,29 @@ const log = getLogger('main/deltachat/backup')
...
+    }
+  }
+
+  private async _internal_export(dir: string) {
+    return new Promise((resolve, reject) => {
+      this._dc.importExport(C.DC_IMEX_EXPORT_BACKUP, dir, undefined)
+      this._dc.on('DC_EVENT_IMEX_PROGRESS', (data1: number) => {
```
Jikstra, line 32: We should clean up/remove this event listener again

src/main/deltachat/backup.ts
```diff
@@ -14,8 +14,29 @@ const log = getLogger('main/deltachat/backup')
...
+        if (data1 === 0) {
+          reject('Backup export failed (progress==0)')
+        } else if (data1 === 1000) {
+          resolve()
+        }
```
Jikstra, line 37: Please clean up the event listener inside this function

Jikstra:
PR for my requested changes https://github.com/deltachat/deltachat-desktop/pull/2158

src/main/deltachat/backup.ts
Simon-Laux: > I guess it won't get executed if you throw an err in the catch branch?

yes, exactly

src/main/deltachat/backup.ts
Simon-Laux: right that was forgotten here

Simon-Laux:
now this pr needs to have only formatting to be fixed before we merge

Issue #2123, opened 2021-02-14
Crash when user tries to open "Settings" after first (initial) profile setup


gerryfrancis:
- Operating System (Linux/Mac/Windows/iOS/Android):
Windows 8.1 Pro.

- Delta Chat Version:
1.15.1.

- Expected behavior:
The user can access the "Settings" window.

- Actual behavior:
The user cannot access the "Settings" window. (Delta Chat crashes when "Settings" is selected.) When the application is closed and rerun again, "Settings" can be access normally.

- Steps to reproduce the problem:
-- Install Delta Chat from scratch.
-- Run Delta Chat and restore a local backup.
-- Open "Settings" from the three-dot menu.

- Screenshots:
![grafik](https://user-images.githubusercontent.com/51095533/107890043-e52fa280-6f16-11eb-9ba5-062c7c01fc73.png)

- Logs:
`2021-02-14T15:11:49.569Z	renderer/react-crashhandler	ERROR	"\nKw.componentDidCatch (dc://deltachat/bundle.js:624:681696),\nKw.SL.s.componentDidCatch.r.callback (dc://deltachat/bundle.js:2:79748),\nHT (dc://deltachat/bundle.js:2:45914),\njT (dc://deltachat/bundle.js:2:45702),\nmre (dc://deltachat/bundle.js:4:11871),\nqa.unstable_runWithPriority (dc://deltachat/bundle.js:1:11874),\nI2 (dc://deltachat/bundle.js:2:41398),\nhf (dc://deltachat/bundle.js:4:8083)"	"The app encountered an react error"	{}`

Jikstra:
@gerryfrancis can you post a longer log?

Simon-Laux:
> "Cannot read property 'displayname' of null"
> ~error from the image

What do you need a log for @Jikstra

gerryfrancis:
> can you post a longer log?

@Jikstra I would like to, but I deleted my backup from Android already, and would have to use a USB data cable to copy a new backup to my Windows notebook. Unfortunately I do not have such a cable at hand right now. Please check again after importing a Delta Chat backup from Android, thanks!

Simon-Laux:
there are other methods of file transfer like ftp over wifi or https://syncthing.net/ or using the sdcard if your android device has one.
Or maybe Bluetooth file transfer.
or use https://github.com/svenstaro/miniserve  `miniserve --upload-files path/to/destination/folder`

gerryfrancis:
[...]

PR #2155, opened 2021-03-03
Fix crash in settings dialog when settings state is empty

Jikstra:
Fixes https://github.com/deltachat/deltachat-desktop/issues/2123


src/renderer/components/message/MessageList2.tsx
```diff
@@ -0,0 +1,270 @@
+import React, { useRef, useEffect } from 'react'
```
Simon-Laux, line 1: this file looks unrelated

PR #2161, opened 2021-03-03
Fix source map support in crash screen

Jikstra:


package.json
```diff
@@ -82,6 +83,7 @@
     "@types/debounce": "^1.2.0",
     "application-config": "^1.0.1",
     "array-differ": "^3.0.0",
+    "buffer": "^6.0.3",
```
Jikstra, line 86: This is not needed

Simon-Laux (approved):
LGTM, did not test though

Issue #2106, opened 2021-02-05
Corrupted letter


TheLudian:
The letter "i" becomes corrupted when I type "fi" (Ubuntu 20.04.2 LTS).

Simon-Laux:
what's your installed deltachat version?
can you post a screenshot of the corrupted input?

Ludan64:
> what's your installed deltachat version?
> can you post a screenshot of the corrupted input?

![Corrupted letter](https://user-images.githubusercontent.com/78592236/107112314-69bb5a80-685f-11eb-8fff-7afa0580e48b.png)

r10s:
i think, this is a feature of the used font. an `f` followed by an `i` is often shown differently to be not confused with `f` and `l` - fi vs. fl - see https://duckduckgo.com/?q=ligature+fi&iax=images&ia=images

i assume, the same happens elsewhere when the same font is used under similar conditions (kerning enabled etc.)

might be that these ligatures can be disabled system-wise somehow, not sure. but i do not think, there is much Delta Chat can/should do on that.

Simon-Laux:
Right that isn't corruption that is a feature of the used font.
When it is a real problem for you @TheLudian I could look into adding a setting to disable them.

Simon-Laux:
@TheLudian please open a new feature proposal (if possible on https://support.delta.chat (you can use your github to login)) if you want an setting/option to disable font ligatures.

PR #2166, opened 2021-03-06
add an experimental switch to disable font ligatures

Simon-Laux:
https://support.delta.chat/t/disabling-font-ligatures/1490

we could also think about changing the default, I'm not sure what option is better.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased][unreleased]
 
+### Added
+- add experimental setting to disable font ligatures, see https://support.delta.chat/t/disabling-font-ligatures/1490
+
 ## [1.15.2] - 2021-03-03
 
 ### Fixed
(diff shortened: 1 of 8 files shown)


Jikstra:
Is there any problem with just disabling font ligature? Don't like the idea of adding an option for something so subtle.

r10s:
there was some request in the issue tracker at https://github.com/deltachat/deltachat-desktop/issues/2106 .

however, tbh, i would also not add such an option, at least not to the official settings and not in the usual "settings" dialog, no even as "experimental". that will come at cost, i do not think, this is worth maintaining.

there are much "larger" apps not having something like that. if someone is uncomfortable with the system font and the ligatures used, we should point him to changing this in the system settings.

r10s:
i just checked signal and telegram desktop apps: i did not find such an option, but both seems to have ligatures disabled, at least if i type in `fi` that looks differently there. so, maybe, that is the more straight forward approach. however, compared to other things, i think, this is a really tiny detail :)

Simon-Laux:
so we are for disabling it in general?

Simon-Laux:
superseded by #2167

r10s:
> so we are for disabling it in general?

_if_ there are really issues with ligatures, i think, this is the better approach, yes.

Issue #2170, opened 2021-03-17
Restrict file protocol


Simon-Laux:
so that only certain locations can be accessed:
```js
// folders the renderer need to load resources from
const ALLOWED_RESOURCE_FOLDERS = ['images', 'node_modules', 'html-dist']
// folders the renderer wants to load source files from (when using the devtools)
const ALLOWED_SOURCE_FOLDERS = ['src', 'scss', 'node_modules']
const ALLOWED_BLOB_FOLDERS = ['db.sqlite-blobs', 'stickers']
```
If it's possible to restrict that easily - alternative in hardening security would be to load from a local web-server.

PR #2171, opened 2021-03-17
Go back to file scheme

Simon-Laux:
There were some problems loading video and audio attachments with the custom schemes so we decided to revert back to the file scheme.
fixes audio and video seeking issues.

- [X] revert to using file scheme
- [X] fix help
- [-] ~~restrict file scheme~~ postponed to #2170
- [X] external sourcemaps should work again

### Points that need to be tested again now:
- attachments
- stickers
- background image
- avatars
- help


link2xt:
This change broke stickers indeed: #2206

PR #2179, opened 2021-03-19
Document GitHub Actions

### LINKED ISSUE #2175: Document GitHub Actions
In https://github.com/deltachat/deltachat-desktop/blob/master/.github/workflows/test-windows.yml#L1, it says the test-windows.yml workflow is documented in the sysadmin repository - this is wrong, and probably a copy-paste error from the build-preview.yml github action.

It is not obvious at first glance what the following GitHub actions are supposed to do - so they should be documented, at best directly in the workflow yml file:

- [ ] test-windows.yml
- [ ] test-linux.yml

missytake:
closes #2175 

This makes it slightly easier to understand for the people who didn't write these actions ;)

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -1,4 +1,5 @@
-name: Run unit & integration tests on Linux
+# Run unit & integration tests on Linux
+name: Linux
 
 on: [push]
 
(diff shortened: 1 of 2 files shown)


Simon-Laux (approved):

Jikstra:
Can you put it into a comment at the top? The name is used in the check overview on github and quite long. Plus using the name as the os makes it easier to see on which os something fails.
Screenshot_20210319_161221

missytake:
Fine for me :) I changed it

PR #2186, opened 2021-03-19
Add current account to contacts in locations

nicodh:
The own location and paths were not displayed in the map since only selectedChat.contacts were shown and the current account is not included in the contacts

### DIFF
--- a/src/renderer/components/map/MapComponent.tsx
+++ b/src/renderer/components/map/MapComponent.tsx
@@ -1,5 +1,5 @@
 import { DeltaBackend, sendMessageParams } from '../../delta-remote'
-import { C } from 'deltachat-node'
+import { C } from 'deltachat-node/dist/constants'
 import React from 'react'
 import ReactDOMServer from 'react-dom/server'
 import ReactDOM from 'react-dom'
--- a/src/renderer/components/map/MapComponent.tsx
+++ b/src/renderer/components/map/MapComponent.tsx
@@ -172,21 +172,25 @@ export default class MapComponent extends React.Component<
     const currentContacts: JsonContact[] = []
     ;(this.mapDataStore as any).locationCount = locations.length
     if (locations.length > 0) {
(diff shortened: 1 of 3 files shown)


nicodh:
seems master is not the right branch (tests failed)

nicodh:
Maybe this PR should go to branch 1.15.x - in master all tests fail...?

src/renderer/components/map/MapLayerFactory.ts
```diff
@@ -37,7 +37,8 @@ export default class MapLayerFactory {
         'line-cap': 'round',
       },
       paint: {
-        'line-color': '#' + color.toString(16),
+        'line-color':
+          color.indexOf('#') > -1 ? color : '#' + color.toString(16),
```
Simon-Laux, line 41: colors returned by delta-chat node should always be returned in hex, so the conversion is not needed anymore.

Also `'#' + color.toString(16)` doesn't work by itself anymore since the color generation in the core changed (new master core, so I would just remove the conversion here and rely on deltachat-node doing the conversion right.

src/renderer/components/map/MapComponent.tsx
```diff
@@ -1,4 +1,5 @@
 import { DeltaBackend, sendMessageParams } from '../../delta-remote'
+import { C } from 'deltachat-node'
```
Simon-Laux, line 2: must be
```suggestion
import { C } from 'deltachat-node/dist/constants'
```
because otherwise frontend bundler thinks it needs to bundle nodejs stuff

Simon-Laux (approved):
LGTM, only a changelog entry is missing now

Simon-Laux:
> Maybe this PR should go to branch 1.15.x - in master all tests fail...?

the test failing is due to an pr of mine, I'll fix it soon
after this pr merged we could also think about backporting to 1.15.x branch and releasing versions with this fix.

PR #2190, opened 2021-03-22
tray icon unread-indication

Simon-Laux:
### To Do

- [x] design better indication icon
- [x] have an `.ico` version of the unread icon to work correctly under windows (currently its just the normal icon)
- [x] think about whether we need it on mac (I'd say no, because the dock can already have an unread badge)
- [x] add change-log entry


Jikstra:
I think we don't need it on mac, but we could just have it for simplicity

Simon-Laux:
> but we could just have it for simplicity

We already have special code for macOS, so this wouldn't make anything simpler...

Jikstra:
Actually not so easy to design an unread icon for this... The normal icon has too many details to be so small and even the template icon has too thin details/lines (the delta). I think to make this work we need to overhaul those icons.

Jikstra:
Screenshot_20211115_150442

Played around a bit with the icon and came up with this

r10s:
> think about whether we need it on mac (I'd say no, because the dock can already have an unread badge)

yip, agree. on mac, the unread-counter on the app-icon is already great :)

PR #2204, opened 2021-04-01
Update README.md for arch installation guidelines

eerielili:
Hello,
    I suggest to use wget with `--no-config` so if a config file for wget exist, it won't temper with the installation instructions and follow default wget behavior.

Kind regards,
ererielili.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ Otherwise you can still do it manually:
 
 ```sh
 # Download the latest snapshot of the PKGBUILD
-wget https://aur.archlinux.org/cgit/aur.git/snapshot/deltachat-desktop-git.tar.gz
+wget --no-config https://aur.archlinux.org/cgit/aur.git/snapshot/deltachat-desktop-git.tar.gz
 
 # extract the archive and rm the archive file afterwards
 tar xzfv deltachat-desktop-git.tar.gz && rm deltachat-desktop-git.tar.gz


Jikstra:
Thanks for your PR, is this a common case that people tweak their wget config that much that a download of a tar gz does weird things?

eerielili:
> Thanks for your PR, is this a common case that people tweak their wget config that much that a download of a tar gz does weird things?

Thanks for replying. I don't think it's that prevalent among normal desktop users on Arch to be frank, but for example here's a config that'll download the directory structure (not content unless it's an index):
```
convert_links = on
adjust_extension = on
span_hosts = on
page_requisites = on
wait = 2
```

With such a config, the `wget https://aur.archlinux.org/cgit/aur.git/snapshot/deltachat-desktop-git.tar.gz` will save the tarball under the directory `aur.archlinux.org/cgit/aur.git/snapshot/`.
I think it should be better to just _mention_ using `--no-config` if anyone got problem with wget or just turn off `page_requisites` for the download, so it doesn't mess with configured proxy settings if any.

Jikstra:
Hmm but would --no-config also ignore proxies? I don't really see the point of adding this to the README. If somebody has put effort into tweaking their wget config, they are power users that know how to handle this.

eerielili:
Yeah, powerusers will figure this out quickly or use the good old `curl -OJL`.

I'm closing the PR, thanks a lot for your attention !

PR #2211, opened 2021-04-07
adding french translation of inAppHelp

eerielili:
Hello, here's a french translation of the help page.

### DIFF
--- a/static/help/fr/help.html
+++ b/static/help/fr/help.html
@@ -1,0 +1,977 @@
+<!DOCTYPE html>
+<html><head><meta charset="UTF-8" /><meta name="viewport" content="initial-scale=1.0" /><link rel="stylesheet" href="../help.css" /></head><body><ul id="top">
+  <li><a href="#what-is-delta-chat">Qu'est ce que Delta Chat?</a>
+    <ul>
+      <li><a href="#how-can-i-find-people-to-chat-with">Comment trouver des gens avec qui discuter?</a></li>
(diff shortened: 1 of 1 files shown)


Simon-Laux:
thanks, but these pages are generated by https://github.com/deltachat/deltachat-pages/blob/master/tools/create-local-help.py, which uses the translations from transifex.
so doing it manually will be overwritten in the future.

eerielili:
Okay, I'm switching to transifex right now :)

Simon-Laux:
feel free to notify me when your done, then I can trigger the script.

r10s:
> Okay, I'm switching to transifex right now :)

yeah, having a french help would be great!
current help is https://delta.chat/fr/help which is created from transifex and is translated only partly yet.

once the translation is mostly complete it will  go to andoid/ios/desktop.

eerielili:
> feel free to notify me when your done, then I can trigger the script. ^^

I've send a "Join team" request in the [transifex](https://www.transifex.com/delta-chat/delta-chat-pages/), is that what you meant?

r10s:
> I've sent a "Join team" request in the transifex, is that what you meant?

yes, thanks a lot! i approved your request, you should now be able to translate to french.

eerielili:
I don't know if it's the right place to point this but,

[The RFC about the protected "Subjet" headers is _expired_](https://datatracker.ietf.org/doc/draft-autocrypt-lamps-protected-headers/). It does mean it won't get implemented or the
IETF has filled a better or more up-to-date RFC ?

The third bullet point in the _"How does Delta Chat protect my Metadata?"_ section mentions it.

Simon-Laux:
superseded by #2227

@eerielili thank you for your help in making DeltaChat better!

PR #2159, opened 2021-03-03
[wip] Overhaul MessageList

Jikstra:
# ToDo

- [x] Fix https://github.com/deltachat/deltachat-desktop/issues/2149
- [x] Fix https://github.com/deltachat/deltachat-desktop/issues/2069
- [x] Fix https://github.com/deltachat/deltachat-desktop/issues/1684
- [x] Fix https://github.com/deltachat/deltachat-desktop/issues/1330
- [x] Fix pages getting loaded twice
- [x] Fix jump to message
- [x] msg change
  - [x] scroll restore buggy
- [x] delete message
- [x] send message
  - [x] not scrolling to send message
- [x] incoming message
  - [x] bugs
- [x] contact requests
- [x] mark all messages read when click jump-to-bottom-button
- [x] weird bug rendering empty messages or wrong chat
- [ ] Fix https://github.com/deltachat/deltachat-desktop/issues/1540
- [ ] Fix https://github.com/deltachat/deltachat-desktop/issues/1407
- [ ] Fix https://github.com/deltachat/deltachat-desktop/issues/1293
- [ ] highlight message on jumpToMessage
- [x] add missing scrollToBottom icon
- [x] show scrollToBottom button also when one page away from newest message
- [x] dont pop/remove markerOne when the new messages will be in view
- [ ] Fix https://github.com/deltachat/deltachat-desktop/issues/1483
- [x] Fix contact request list
- [x] Fix context menu
- [x] Fix chat audit log
- [x] Fix author name on quote

PR #2220, opened 2021-04-12
fix: use contact.getDisplayName() for message author name

link2xt:
Core has all the logic necessary to display correct name or email address.


link2xt:
Maybe should be merged/cherry-picked into #2159 instead, but it does not conflict as far as I see, so would be nice to merge it before #2159 so everyone using master branch can have the fix already.

PR #2224, opened 2021-04-15
Add context menu for videochat messages

AUtech-dev:
The videochat messages currently do not have a context menu. As they are basically also normal messages things like forwarding, replying and message details are also present and interesting for the user (at least for me ;).

This pull request adds the already existing context menu to the text part of the videochat message.

Secondly it adds a context menu entry for this messages that allows to copy the URL for  the videochat into the clipboard. The same strings as copying a link from a normal message is used.


Simon-Laux (approved):
Great, thanks for your PR :+1:

PR #2237, opened 2021-05-06
update deltachat-node to v1.54.0

Simon-Laux:


Simon-Laux:
no api changes, just the core. we'll see if its more stable than the core version we currently use

PR #2248, opened 2021-05-16
status input to textarea

### LINKED ISSUE #2246: "Signature text" field in profile settings should be multiline
It is multiline on Android and latest core synchronizes the signature between devices, like the avatar. But then Desktop squashes it into a single line if you open the settings.

Simon-Laux:
- [ ] auto resizing to content would be nice (so that the field grows and shrinks with the content, not having it means that a scrollbar will be shown even though there is still much free space below)
- [ ] maybe the styling can be improved

addresses/closes #2246


Jikstra:
We can use this https://stackoverflow.com/questions/17731083/how-to-autogrow-a-textarea-with-css to have a autogroving "textarea"

PR #2251, opened 2021-05-16
adjust styling of show html button

### LINKED ISSUE #2250: adjust styling of "Show full message in browser" button
can we have the "Show full message in browser" button a bit nicer and also a bit more tuned down? 

see https://github.com/deltachat/deltachat-android/issues/1895 for an ios screenshot, the corresponding android-pr is at https://github.com/deltachat/deltachat-android/pull/1897

excerpt from the reasoning: as the button is shown for many messages, and does not always offer really additional content (and we cannot really say if that is the case or not), it is better to tune the button down a bit, let them appear as a usual text link. the button should not get too much attraction, depending on the chat, it may be shown inside nearly every bubble.

_Originally posted by @r10s in https://github.com/deltachat/deltachat-desktop/issues/2132#issuecomment-841730070_

Simon-Laux:
closes #2250

(don't wonder that no names are shown, the chat shown is a mailing-list and `[Feature] mailing list support #2132` is/was not merged yet)

Change log was not changed, because there is no change to the user compared to last released version.

### DIFF
--- a/scss/message/_message.scss
+++ b/scss/message/_message.scss
@@ -153,8 +153,8 @@
 
   button.show-html {
     border: none;
-    font-weight: bold;
     background-color: transparent;
+    padding-left: 0;
     color: var(--colorPrimary);
   }
 }
--- a/scss/message/_message.scss
+++ b/scss/message/_message.scss
@@ -150,6 +150,13 @@
     max-width: 100%;
     min-height: unset;
   }
+
+  button.show-html {
+    border: none;
+    font-weight: bold;
+    background-color: transparent;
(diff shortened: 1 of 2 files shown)


r10s:
yeah, much better! thanks for taking care!

some nitpicking:
- i would remove the left padding, so that the button begins at the same x-position as the text
- i think, there is no need to use bold text, just use the same font+size+style in that nice blue. by its position, the button is probably outstanding enough. that would also be consistent with android/ios.

r10s:
looks great!

PR #2255, opened 2021-05-19
add `--minimized` CLI option

Simon-Laux:
related forum topic: https://support.delta.chat/t/option-to-start-on-system-start/1663

when started minimized the tray-icon is enabled and its option is disabled for the session.


r10s:
not sure if i get the whole context, but if "starting minimized" really forces "tray icon always on" for whatever reason, i think, that should be reflected by the state of the switch as well: so, showing the switch "on and inactive", _not_ "off and inactive" as in the screenshot.

and maybe make that clear in the note: "Tray icon cannot be hidden as Delta Chat was started with the --minimized option".

Simon-Laux:
minimized means minimized to tray -> there is no windows shown in this case.

PR #2273, opened 2021-05-28
pull translations

Simon-Laux:


Jikstra:
Can we now remove things from untranslated?

Simon-Laux:
> Can we now remove things from untranslated?

Are there still things in untranslated? I removed them while pushing (pulled translations in that pr again after the English ones were upstream)

PR #2298, opened 2021-06-30
Prepare 1.20.3

Simon-Laux:
- update core to 1.56
- update translations


r10s (approved):
yeah! thanks a lot!

no line of js in this pr - i think, i can approve that even without knowing very few about desktop architecture things :)

PR #2302, opened 2021-07-08
Upgrade react to 17

Simon-Laux:
- [X] add eslint-plugin-react-hooks
- [X] fix the lint errors
- [x] fix the lint warnings
- [x] upgrade react and react-dom to `17.0.2`


Simon-Laux:
looks like fixing all the warning is not as easy as it might seem, removing the dependency array of the useEffect in `src/renderer/components/OfflineToast.tsx` produces somehow a frozen/slow app on switching between offline and online.

which is a bit strange because the only effect-full line of it should be
```
if (navigator.onLine === false) onBrowserOffline()
```
the rest is registering / un-registering event listeners.
somehow it called maybe network a few hundred times, I guess that function needs a debounce.

Simon-Laux:
- unblock contacts dialog (multiple contacts at once, does it update correctly?)
- system message on empty deaddrop (show_emails off and show_emails on)
- switch languages (ui should be updated)
- open settings and change settings (settings that affect the core)
- change background to a solid color and to the default color
- offlinetoast going offline/online -> thought the offline toast might be updated to connectivity view anyway
- Drafts (set draft, remove draft, send draft/message with attachments and quotes)
- messagelist
- look if timestamps still auto update (after each minute)
- encryption info
- disapearing messages dialog
- chat auditlog
- login, change credentials/imap setting -> focus on if the configure dialog -> does it work correctly?
- sticker picker
- edit group (adding, removing, searching contacts)
- chatlist (also chatlist in profile view and forward message)
- chatlist scroll selected chat into view
- search for chats, contacts and messages

Simon-Laux:
/me removed the types update because it broke the test

PR #2307, opened 2021-07-12
replace `react-virtualized` with `react-window`

Simon-Laux:
- [x] fix bug that each time the chat-list-item selection changes all items get re-rendered (if you disable caching in the network tab of the dev-tools you can see the avatar images flashing as they reload) 

even with that bug it already feels faster


CHANGELOG.md
```diff
@@ -11,12 +11,14 @@
...
 - Update testcafe to `1.15.0`
 - Upgrade `react` and `react-dom` to `17.0.2`
+- replace `react-virtualized` with `react-window`
 
 ## [1.20.3] - 2021-06-30
 
 ### Fixed
 - Chat Background fixes (fix black border on bee background, fix preview of background in settings and color picker now is set to the current color when its opened)
 - allow brackets in links (see #2238)
+- fix duplicated contacts in search
```
dotlambda, line 21: This shouldn't be here.

Issue #2306, opened 2021-07-11
wrong mail address


MrMini1025:
- Operating System (Linux/Mac/Windows/iOS/Android): Linux
- Delta Chat Version: Version 1.20.3 (git: v1.20.3)
- Expected behavior: Login and work **(the email addresses are valid as per RFC 5322)**
- Actual behavior: it doesn't work with tld domains
- Steps to reproduce the problem: use email with tld domain in login like **bob@mymail**
- Screenshots: 
![Screenshot_20210711_231932](https://user-images.githubusercontent.com/19595503/125209211-ff8c6300-e29f-11eb-862c-e8a7a73615b3.png)

link2xt:
Fixed by #2317?

Jikstra:
@link2xt yes should be

PR #2312, opened 2021-07-22
make tld email addresses possible

Simon-Laux:
desktop part of fixing #2306

using the core function is not easily possible, because we need the function in the renderer process as synchronous method.

### DIFF
--- a/src/shared/util.ts
+++ b/src/shared/util.ts
@@ -8,12 +8,7 @@ export function isValidEmail(email: string) {
   // empty string is not valid for local part
   if (local === '') return false
   // domain is too short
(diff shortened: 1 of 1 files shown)


cofob (approved):

Jikstra:
Why does it need to be synchronous?

Simon-Laux:
> Why does it need to be synchronous?

as you have seen in your "edit" the places we need it need it synchronous without modifying much / using hacks like useState, which could create nasty race condition bugs if not used carefully.

Simon-Laux:
doing it while searching did the trick, thanks for the idea @Jikstra

this pr is succeeded by #2317

PR #2313, opened 2021-07-23
get rid of dependency array-differ

dotlambda:
The existing implementation first turns the array on which `includes()` is called into a set: https://github.com/sindresorhus/array-differ/blob/master/index.js
If you think that's worth it despite the small group size in Delta Chat, I can reimplement it.
Also, feel free to say you'd rather keep the dependency.


Simon-Laux:
can you rebase this pr and also add a line to the changelog (or just to the list of removed unused dependencies)?

dotlambda:
> can you rebase this pr and also add a line to the changelog (or just to the list of removed unused dependencies)?

Done, sorry for the delay.

dotlambda:
@Simon-Laux Now I also resolved the merge conflict.

PR #2314, opened 2021-07-23
clean up dependencies

dotlambda:


dotlambda:
> Package "electron-builder" is only allowed in "devDependencies". Please remove it from the "dependencies" section in your package.json.

I guess `optionalDependencies` is really meant for optional run-time dependencies. I would however love to be able to distinguish between test and CI dependencies and those which are strictly necessary for building deltachat-desktop.

package.json
```diff
@@ -95,6 +95,7 @@
     "mime-types": "^2.1.25",
     "moment": "^2.27.0",
     "node-fetch": "^2.6.1",
+    "package-lock.json": "^1.0.0",
```
Simon-Laux, line 98: whats that? probably doesn't belong here

Simon-Laux:
I'd like to get https://github.com/deltachat/deltachat-desktop/pull/2303 merged before this one, also if we should go the optionalDependencies route we should mention what installation command you need in the readme/docs before the functions that need those "optional" dependencies.

Also it looks like electron builder does not like to be an optional dependency: 
```
Package "electron-builder" is only allowed in "devDependencies". Please remove it from the "dependencies" section in your package.json.
```

dotlambda:
Since `optionalDevDependencies` is not a thing, it might be a better option to have two `package.json` files: https://www.electron.build/tutorials/two-package-structure

Simon-Laux:
yeah also thought about that, that could be a solution. depending on how many files need to be moved... can we somehow do it without moving the whole source dir inside into an app folder, If not I'd rather do this when we don't have many PRs open.

CC @Jikstra what do you think?

PR #2317, opened 2021-07-24
Use core method to validate email addresses

Simon-Laux:
this pr is based on the react upgrade so that should be merged first


src/shared/util.ts
```diff
@@ -1,19 +0,0 @@
...
-  if (email === '') return false
-  const parts = email.split('@')
-  // missing @ character or more than one @ character
-  if (parts.length !== 2) return false
-  const [local, domain] = parts
-  // empty string is not valid for local part
-  if (local === '') return false
-  // domain is too short
-  if (domain.length <= 3) return false
-  const dot = domain.indexOf('.')
```
groovenectar, line 12: I noticed `indexOf()` rather than `lastIndexOf()` ... even though an email address can have a subdomain, is that not relevant in this context?

Simon-Laux: this code is removed by this pr anyway, so no need to discuss further why it is broken IMO

PR #2318, opened 2021-07-24
get rid of dependency fs-extra

dotlambda:
Feel free to close if you think this is not worth it.


Simon-Laux:
not possible with our current target of node 12, also I personally like using `fs-extra`, still might be something we could do after upgrading our node target version.

dotlambda:
> current target of node 12

Then you should not use `@types/node` v14.

But this was also done in https://github.com/deltachat/deltachat-node/pull/514, so that would have to be reverted if we stick with Node.js 12.

> I personally like using `fs-extra`

I don't know it very well, but felt like the additional functions it provides don't always have well-defined semantics.

bin/test_package_lock_version.js
```diff
@@ -1,4 +1,4 @@
-const { readJson } = require('fs-extra')
+const { readFile } = require('fs-extra')
```
Simon-Laux, line 1: still one reference here

Simon-Laux: I'll merge and fix on the other branch

PR #2322, opened 2021-07-30
Add Rocket.chat theme

BorisPAING:


Jikstra:
Thanks for contributing! Can you share a screenshot of how it looks? :)

BorisPAING:
Hi Jikstra, 

Sure! For now, it looks like this: 
http://uploads.borispaing.fr/delta.chat-theme-rocket.chat-2021-07-30.png

Cheers!

Simon-Laux:
Looks good already, do you plan to continue to work on it and send updates in separate prs?

Also if you need something feel free to ask, such as changing the height of chatlist items, maybe its possible for us to implement a way to specify chatlistitem height in themes if its needed.

BorisPAING:
I never plan anything ;-) 

Indeed, being able to change the height of chatlist items would be great.

I would also like to be able to reproduce Rocket.chat theme behavior of hiding the avatar on the following messages by the same author. I could if instead of classes being arranged this way:
`
<li>
   <div class="message incoming">
   </div>
</li>
`
they were instead arranged that way:
`<li class="message outgoing">
   <div>
   </div>
</li>`

because then I could use CSS selectors in the following manner:
`li.outgoing + li.outgoing .author {
   display: none;
}`

Simon-Laux:
Ok then I think the best way is to merge this now and then I can work on fixing the stuff you posted and after that or when you find other things to improve on it, you can do updates in new prs.

Simon-Laux:
> Styling of successive messages by the same author

I think that method would break in groups, maybe e could add a class on messages from the same author instead but I can't promise anything...

PR #2329, opened 2021-08-04
Cleanup backend

Simon-Laux:
- remove unused variables in backend
- make `chat.getChatMedia` independent of global selected chatid
- remove unused properties of `MessageType`
- flatten `MessageType` (no `.msg` property anymore`)
- rename `MessageType.contact` to  `MessageType.sender`
- remove unnessesary dependencies on FullChat.contacts
- remove duplicate of 'settings.setConfig'

Later, after we merge #2159 we can remove `this._controller._selectedChatId` completely and move `chatList.selectChat` to the frontend.

Testing notes:
- this changes stuff about creating DM chats, so creating / opening dm chats from various places should be tested.
- changing core settings (settings that are handled by the core)
- everything that is related to messages / drafts / autocrypt setup and so on...


src/main/deltachat/contacts.ts
```diff
@@ -41,34 +41,13 @@ export default class DCContacts extends SplitOut {
     return this._dc.createContact(name || '', email)
   }
 
+  /** Gets the direct message chat id with this contact and creates it if it doesn't exist yet */
   createChatByContactId(contactId: number) {
     const contact = this._dc.getContact(contactId)
     if (!contact) {
       log.warn(`no contact could be found with id ${contactId}`)
       return 0
```
Jikstra, line 49: With the try/catch you're not checking for this 0 anymore

Jikstra (changes requested):
Besides the unhandled zero return this looks good

PR #2355, opened 2021-09-10
Implement settings for socks5

Jikstra:
image

@r10s

### DIFF
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -9,21 +9,21 @@
     "message": "Open account to configure it"
   },
   "login_socks5": {
-    "message": "Socks5"
+    "message": "SOCKS5"
   },
   "login_socks5_enabled": {
-    "message": "Socks5 enabled" 
+    "message": "SOCKS5 enabled" 
   },
   "login_socks5_host": {
-    "message": "Socks5 host" 
+    "message": "SOCKS5 host" 
   },
   "login_socks5_port": {
-    "message": "Socks5 port" 
+    "message": "SOCKS5 port" 
   },
   "login_socks5_login": {
-    "message": "Socks5 user" 
+    "message": "SOCKS5 user" 
   },
   "login_socks5_password": {
-    "message": "Socks5 password" 
+    "message": "SOCKS5 password" 
   }
 }
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -25,5 +25,8 @@
   },
   "login_socks5_password": {
     "message": "SOCKS5 password" 
+  },
+  "login_socks5_experimental_warning": {
+    "message": "SOCKS5 support is currently experimental. Please use at your own risk. If you type in an adress in the e-mail field, there will be DNS lookup that won't get tunneld through SOCKS5."
   }
 }
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -11,8 +11,8 @@
   "login_socks5": {
     "message": "SOCKS5"
   },
-  "login_socks5_enabled": {
-    "message": "SOCKS5 enabled" 
+  "login_socks5_use_socks5": {
+    "message": "Use SOCKS5" 
   },
   "login_socks5_host": {
     "message": "SOCKS5 host" 
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -7,5 +7,23 @@
   },
   "unconfigured_account_subtitle": {
     "message": "Open account to configure it"
+  },
+  "login_socks5": {
+    "message": "Socks5"
+  },
+  "login_socks5_enabled": {
+    "message": "Socks5 enabled" 
+  },
+  "login_socks5_host": {
+    "message": "Socks5 host" 
+  },
+  "login_socks5_port": {
+    "message": "Socks5 port" 
+  },
+  "login_socks5_login": {
+    "message": "Socks5 user" 
+  },
+  "login_socks5_password": {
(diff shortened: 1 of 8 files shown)


_locales/_untranslated_en.json
```diff
@@ -7,5 +7,23 @@
...
+    "message": "Socks5 host" 
+  },
+  "login_socks5_port": {
+    "message": "Socks5 port" 
+  },
+  "login_socks5_login": {
+    "message": "Socks5 user" 
+  },
+  "login_socks5_password": {
+    "message": "Socks5 user"
```
dotlambda, line 27:
```suggestion
    "message": "Socks5 password" 
```

dotlambda (changes requested):

dotlambda:
[Tor](https://gitweb.torproject.org/torspec.git/tree/socks-extensions.txt) spells it `SOCKS5`. Shouldn't we do the same?

r10s:
are the options integrate in the settings (so just below IMAP and SMTP server settings) or do they have a separate dialog?

- i assume, the options below "SOCKS5 enabled" are "Yes" and "No"? in this case,  i would use a checkbox-like toggle and maybe better call it "Use SOCKS5"
- an extra headline is not needed in this case
- and if switched off, the other options can be hidden - that way, the dialog is not super-full by the new options
- when on that, double the space between the sections in general (so, between _Inbox-Outbox_ and _Outbox-CertificateChecks_ and _CertificateChecks-SOCKS5_ 

if the SOCKS5 options are in a separate dialog, the points are still valid, but less important ui-wise.

Simon-Laux:
also there a there is this tricky question:
socks5 setting should be activated before the provider info, because it could trigger a web request, but then most people would be confused by another setting outside of the advanced mode, so it would make beginner ux worse - ideas how to solve that ux problem?

Jikstra:
> i assume, the options below "SOCKS5 enabled" are "Yes" and "No"? in this case, i would use a checkbox-like toggle and maybe better call it "Use SOCKS5"

Used a switch and changed the language string

> an extra headline is superfluous in this case

Yup


> if switched off, the other options should be hidden - that way, the dialog is not super-full by the new options

Done
>  when on that, double the space between the sections in general (so, between Inbox-Outbox and Outbox-
CertificateChecks and CertificateChecks-SOCKS5

I made the headlines bold, i think that's good too. But we can overhaul that whole thing at some point, don't really like it.

r10s:
> also there a there is this tricky question:
socks5 setting should be activated before the provider info, because it could trigger a web request, but then most people would be confused by another setting outside of the advanced mode, so it would make beginner ux worse - do you have ideas how to solve that ux problem?

which web-request is triggered before `dc_configure()` is actually called?

i thought the only problems was the call of [`dc_get_oauth2_url()`](https://c.delta.chat/classdc__context__t.html#a86eb36a93aba65fc4ab7c675a4d76158) on android _before_ the login-button is hit - but oauth2 is not supported by desktop anyway.

Simon-Laux:
> which web-request is triggered before dc_configure() is actually called?

for some reason it is checked in the provider info 
https://github.com/deltachat/deltachat-core-rust/blob/master/deltachat-ffi/src/lib.rs#L3629-L3633

which is called upon changing the contents of the email input field.

r10s:
i see. maybe we should alter `dc_provider_new_from_email()` so that mx-checks are always disabled.

that would leave out "gmail" information for custom gmail domains.

oauth2 would continue working as expected.

maybe that is a good compromise it would also fix https://github.com/deltachat/deltachat-core-rust/issues/2351. asking users about SOCKS5 beforehand is probably even worse ux.

Jikstra:
> however, maybe this is all is for the next release, 1.21 is out, such changes in core would probably need a complete new testing round and would delay things further. if this is not acceptable, i would suggest to postpone SOCKS5 completely after 1.22.

I think even with this it's fine. We don't need to advertise socks5 much for now, maybe we can add a small warning hint to the socks5 settings.

src/main/deltachat/login.ts
```diff
@@ -56,6 +56,7 @@ export default class DCLoginController extends SplitOut {
   }
 
   async updateCredentials(credentials: Credentials): Promise<void> {
+    console.log('xxx', credentials)
```
Simon-Laux, line 59:
```suggestion
```

scss/login/_login.scss
```diff
@@ -182,9 +186,11 @@ div.delta-form-group {
...
     }
 
-    .bp3-button.bp3-minimal.bp3-intent-warning,
-    .bp3-button.bp3-minimal.bp3-intent-warning:hover {
+    .bp3-button.bp3-minimal.bp3-intent-warning {
       color: #62656a !important;
+      &:hover {
+        color: white !important;
+      }
     }
```
Simon-Laux, line 194: where is this used?

Jikstra: It fixes the show password icon being all grey on hovering. Fixed it on the go.

r10s:
maybe just flag the whole thing as being experimental, so "Use SOCKS5 (experimental)"

Jikstra:
> maybe just flag the whole thing as being experimental, so "Use SOCKS5 (experimental)"

Wont work as you cannot enable that on adding an account. I will leave it like this with the warning.

r10s:
> Wont work as you cannot enable that on adding an account. I will leave it like this with the warning.

ahh - i did not know that.

but why is a warning needed at all then?

it seems to be totally clear that things are not tunneled over _whatever_ if there is not even a setting for that.

and if it is later enabled in the settings, no additional/unexpected things will be leaked.

PR #2359, opened 2021-09-12
add switch to toggle "all account syncing"

Simon-Laux:
image_2021-09-12_00-54-09

- hides unread badges when syncing is off.

- also fix flashing up account list on startup

### DIFF
--- a/src/renderer/components/screens/AccountsScreen.tsx
+++ b/src/renderer/components/screens/AccountsScreen.tsx
@@ -150,14 +150,14 @@ export default function AccountsScreen({
   const tx = useTranslationFunction()
   const [logins, setLogins] = useState(null)
 
-  const [sync_all_accounts, set_sync_all_accounts] = useState(null)
+  const [syncAllAccounts, setSyncAllAccounts] = useState(null)
 
   useEffect(() => {
     ;(async () => {
       const desktopSettings = await DeltaBackend.call(
         'settings.getDesktopSettings'
       )
-      set_sync_all_accounts(desktopSettings.syncAllAccounts)
+      setSyncAllAccounts(desktopSettings.syncAllAccounts)
     })()
   }, [])
 
@@ -232,16 +232,16 @@ export default function AccountsScreen({
                     {tx('login_known_accounts_title_desktop')}
                   </h4>
                   <Switch
-                    checked={sync_all_accounts}
-                    label='Sync all'
+                    checked={syncAllAccounts}
+                    label={tx('sync_all')}
                     onChange={async () => {
-                      const new_state = !sync_all_accounts
+                      const new_state = !syncAllAccounts
                       await DeltaBackend.call(
                         'settings.setDesktopSetting',
                         'syncAllAccounts',
                         new_state
                       )
-                      set_sync_all_accounts(new_state)
+                      setSyncAllAccounts(new_state)
                     }}
                     alignIndicator={Alignment.RIGHT}
                   />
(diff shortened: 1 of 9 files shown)


src/renderer/components/screens/AccountsScreen.tsx
```diff
@@ -203,9 +222,30 @@ export default function AccountsScreen({
...
+                    Classes.DIALOG_HEADER,
+                    'bp3-dialog-header-border-bottom'
+                  )}
+                >
+                  <h4 className='bp3-heading'>
+                    {tx('login_known_accounts_title_desktop')}
+                  </h4>
+                  <Switch
+                    checked={sync_all_accounts}
+                    label='Sync all'
```
Jikstra, line 236: Use a translation string here

src/renderer/components/screens/AccountsScreen.tsx
```diff
@@ -142,6 +150,17 @@ export default function AccountsScreen({
   const tx = useTranslationFunction()
   const [logins, setLogins] = useState(null)
 
+  const [sync_all_accounts, set_sync_all_accounts] = useState(null)
```
Jikstra, line 153: When did we switch to snake_case for variables? I thought we enforced camelCase somewhere?

Jikstra (commented):
Nice! Some minor things and maybe we should tweak the translation string a bit? @r10s what do you think?

Jikstra:
Even with @r10s having some concerns, i'm fine to merge this into the next release. I will run for me only sync all accounts, @Simon-Laux only syncing one account so we should be fine.

r10s:
yip, i think i mentioned by concerns :)

i think, it is a bit unclear to the user, what will be synced if "Sync all" is off. but well, it is probably okay for now, esp. is "Sync all" is on by default, not so many questions will pop up ...

... what is the default? i suggest to keep that on by default, also as it is advertised by that and has the better overall ux as switching is much faster. and this "always on" is also also needed by AEAP to make qr-joins working better.

Jikstra:
> ... what is the default? i suggest to keep "Sync all" on by default, also as it is advertised by that and has the better overall ux as switching is much faster. and this "always on" is also also needed by AEAP to make qr-joins working better.

Yup has to be the default to sync all accounts

dotlambda:
Just out of curiosity: Why do you not want all accounts to be synced?

Simon-Laux:
it is the default to sync all accounts that is not even a question, look at the code.

I first thought about an icon instead of the text that would still be an option.

> Just out of curiosity: Why do you not want all accounts to be synced?

- In the case that one account has event log spam bugs or other bugs that might affect battery life / performance, there will be an workaround to still using dc until the next update fixes the bug in the "legacy" mode.
- If I'm in a public wifi, I might don't want to share/leak all providers I'm using.
- I might do a digital detox and only want to poll accounts manually and don't see the unread badges.

src/shared/state.ts
```diff
@@ -21,6 +21,7 @@ export function getDefaultState(): AppState {
       zoomFactor: 1,
       activeTheme: 'system',
       minimizeToTray: false,
+      syncAllAccounts: true,
```
Simon-Laux, line 24: that's the default

r10s:
> I might do a digital detox and only want to poll accounts manually and don't want to see the unread badges on all accounts. (as I'm addicted to clicking unread badges)

for that it would be better to just disable the badges, maybe just when notifications are disabled for an account.
(i am not suggesting to change that just now/seen, just an idea for some possible future :)

in general, i totally agree in that digital detox point; android/ios currently does not have these badges at all currently, btw (not because of digital detox but more because that is just not done yet, but still ;.

src/shared/state.ts
r10s: k, i did not dived into the code :)

PR #2367, opened 2021-09-19
move "Copy link to clipboard" to top

Simon-Laux:
in message context menu.

Why? because when I right click a link its the most relevant option for me.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased][unreleased]
 
+### Changed
+- move "Copy link to clipboard" to top in message context menu
+
 ## [1.21.1] - 2021-09-18
 
 ### Added
--- a/src/renderer/components/message/Message.tsx
+++ b/src/renderer/components/message/Message.tsx
@@ -138,6 +138,11 @@ function buildContextMenu(
   const selectedText = window.getSelection()?.toString()
 
   return [
+    // Copy link to clipboard
+    link !== '' && {
+      label: tx('menu_copy_link_to_clipboard'),
+      action: () => runtime.writeClipboardText(link),
+    },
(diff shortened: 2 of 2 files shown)


r10s:
> its the most relevant option for me.

well, that is probably correct :)

however, as we do not have another option or shortcut to _reply_, i'd say, in general, for most users, the _reply_ option is even more important. i would not change that. esp. i would not change that without having a siginigicant amount of user feedback that justifies that change and breaking existing flows.

Simon-Laux:
> it is not long ago, we had "double click to reply" - and that is a strong hint, that this is an important option.

I would not call jikstra's opinion a strong hint.

other option would be doing it like telegram: 
```
if selection {show "copy selected text" as top option }
copy option = if link { show "copy link" } else { "copy message text" }
```

r10s:
it was not only about @Jikstra , we were thinking about buttons and whatnot.

Simon-Laux:
what about doing it the telegram way? @r10s

r10s:
as said, i would not change things here. and not trying to make smart things. i mean, copy link to clipboard is not super-important, is it? telegram also does different things on mac, for example.

and, i would not try to change that now. let's maybe discuss that with an overall improvement of the menus, eg. also "Forward" is at the wrong position (should be after "Reply"), there are lots of things to improve ...

again, please just close that :)

Simon-Laux:
> also "Forward" is at the wrong position (should be after "Reply")

I wonder about the reason, but sure if you only want to talk later in time about it.

r10s:
"reply" and "forward" are functionality-wise very close, it is good to reflected such things in ui as well.

look at other messengers or email-apps, they're very close together most times, and that makes a lot of sense.

Issue #2209, opened 2021-04-06
No logfolder or logfile created in Windows (store) version


havet:
Hi,
wanted to have a look in the logfile because some messages got lost. Couldn't find any logs.

- Operating System (Linux/Mac/Windows/iOS/Android): Windows 10
- Delta Chat Version: 1.15.3
- Expected behavior: logfile/logfolder opened in explorer with: View - Developer - Open the log folder/current logfile
- Actual behavior: Error message: C:\users ... DeltaChat/logs is not accessible ...
- Steps to reproduce the problem: Se expected behavior
- Screenshots: see below
![Error](https://user-images.githubusercontent.com/7511145/113723462-aa194680-96f1-11eb-8adc-5bc0dd184ea0.jpg)

- Logs: NO.

Simon-Laux:
Maybe the windows store version saves its files somewhere else? we should investigate this further

havet:
I've looked in various places, but have not found anything related to Delta Chat. Maybe you'll have to look in the code? I haven't so far.

Simon-Laux:
Looking for in deltachat code would be a waste of time, because its not in it - we get the directories from https://www.npmjs.com/package/application-config#config-location, so maybe its somewhere in there or just some kind of appx-sandboxing applies by windows:
My thinking was that the windows store version might use some sandboxing and stores the apps files in a separate location - in that case we need to adjust the logic in our code that the button also works correctly in the windows store version.

Simon-Laux:
so appx does include additional sandboxing: https://docs.microsoft.com/en-us/windows/msix/desktop/desktop-to-uwp-behind-the-scenes
but I couldn't find where it saves the appdata files to on that docpage other than that it is a `private per-user, per-app location`

Simon-Laux:
found them: `C:\Users\XXXXXX\AppData\Local\Packages\merlinux.DeltaChat_v2ry5hvxhdhyy\LocalCache\Local\DeltaChat\logs`

looks like we need an if statement to change the path when on the windows store version.

Simon-Laux:
Just found out that this also affects opening blob files: they cannot be found, because the application opening them is not in the sandbox Deltachat is in (with being an appx/uwp)
HTML emails work, because they are copies in a temporary directory first.

But this particular issue will probably disappear once core moves the blobfiles to the db, because then we have to copy those files to temporary directories anyway.

CC @r10s @link2xt @hpk42

Simon-Laux:
[...]

PR #2369, opened 2021-09-20
🚧 add a function to find out if we are running as appx

Simon-Laux:
(untested)
this could solve the logfolder not found issue #2209
but there could be better ways to approach this problem

I suggest testing this when we have code signing, also when we have codesigning it should be possible to do this via changing a flag while building a signed appx in the ci that we can then test (testing unsigned appx is not easy and importing a self signed cert is annoying)


Simon-Laux:
looks like the detection part doesn't work, even worse it crashes on first startup

Simon-Laux:
also replacing the config path to the actual sandbox path crashes the application.

So maybe we should just test for it in the menu for opening the right file?
Or maybe have a special build process for apps/UWP in general? so we can customise more to make it really work without electron builder limitations?

Also it might actually make sense to change the application directory name like for GitHub dev builds, but then we need migration because we didn't do it from the start, but it would make testing easier, because the two installations (apps and normal setup) wouldn't be strangely merged like they are currently.

Simon-Laux:
so got the workaround (only changing menu open paths) to work, but the other problems still remain so we can review&merge this (probably)

Simon-Laux:
> What is v2ry5hvx...? Why is it hardcoded?

I believe it is somekind of id that microsoft gave us. I believe it's the same on multiple systems, but right we should test this first on other systems, not just on the one device I tested it on.

PR #2371, opened 2021-09-21
"remove account" fails now allways on windows, before it tries because it…

Simon-Laux:
… doesn't work anyway:

b

a

also make sure the account is deselected when removing the selected account

and display an error if it fails


Simon-Laux:
replaced by #2373

PR #2373, opened 2021-09-22
fix windows account removal bug

Simon-Laux:
(make sure to unref all contexts after use)

also make sure the account is deselected when removing the selected account

and display an error if it fails

- [x]  TODO changelog entry 

supersedes #2371


r10s:
great you figured that out, @Simon-Laux!

Issue #2343, opened 2021-08-25
option to get from memberlist to user's profile missing


r10s:
i think a click on a member in group's memberlists should open the profile of the contact as on android/ios.

this way, one can "browse" groups much better and one can also send a message to a member without much hassle.

for "browsing", however, it should be possible to go back to the group's profile - so, "Cancel" or "Ok" should bring you back there. at best, the contact-profile lies over the group-profile and the group-profile is still a bit visible, might be complicated in the current environment, not sure.

Simon-Laux:
closed by #2381

Issue #1905, opened 2020-10-03
Edit group should be readonly on left groups


Simon-Laux:
After leaving a group the "edit group" menu disappears but when clicking on the chat part of the header I can access the edit group menu and make changes that result in an error (as expected because I'm not a part of the group anymore).

PR #2381, opened 2021-09-29
Overhaul EditGroup & Settings dialog

Jikstra:
# Changes

- Bring back the back button
- Allow viewing member profile
- when removing a member we now ask for confirmation
- add member is shown in its own dialog on top
- get rid of close footer buttons

# ToDo
- [x] fix addmember dialog
- [x] fix changing contact/group name
- [ ] show qr invite code in own dialog too
- [ ] overhaul create group too
- [ ] https://github.com/deltachat/deltachat-desktop/issues/1905
- [ ] https://github.com/deltachat/deltachat-desktop/issues/1244
- [ ] https://github.com/deltachat/deltachat-desktop/issues/2384
- [ ] https://github.com/deltachat/deltachat-desktop/issues/1939
  - [ ] show contacts that are already in the group greyed out in add member dialog
  - [ ] chips for currently added members in add member dialog

# Fixes
https://github.com/deltachat/deltachat-desktop/issues/2343

https://user-images.githubusercontent.com/34889164/136045026-744be326-2cea-4f1e-b623-5e1e46b67b7a.mp4

https://user-images.githubusercontent.com/34889164/136045037-0ff314a1-2560-4f92-acf5-13a3139bb8d3.mp4

https://user-images.githubusercontent.com/34889164/136045044-32bb4660-47c2-497c-96c0-2c40f3d30921.mp4

https://user-images.githubusercontent.com/34889164/136045047-58966c1f-82ba-48ac-9578-2e3faab952a3.mp4

https://user-images.githubusercontent.com/34889164/136045050-ae0e23be-0b5b-4dc9-a86d-8eb068d8e868.mp4

https://user-images.githubusercontent.com/34889164/136045056-1f54b812-376b-4c59-8bec-97d90f4fb4ab.mp4


Simon-Laux (changes requested):
- [x] rebase to master
- [x] fix formatting
- [x] make eslint happy

src/main/deltachat/chat.ts
```diff
@@ -47,19 +47,29 @@ export default class DCChat extends SplitOut {
...
+      const remove = previousMembers.filter(m => !members.includes(m))
+      const add = members.filter(m => !previousMembers.includes(m))
+
+      remove.forEach(id =>
+        this.selectedAccountContext.removeContactFromChat(chatId, id)
+      )
+      add.forEach(id =>
+        this.selectedAccountContext.addContactToChat(chatId, id)
+      )
+    }
```
Simon-Laux, line 72: theoretically there is a race condition here when someone else modifies the members at the same time and ui hasn't updated yet. so I would rather see that split in the ui if that makes more sense?

src/renderer/components/dialogs/CreateChat.tsx
```diff
@@ -339,12 +344,16 @@ export function AddMemberInnerDialog({
     onOk(addMembers)
   }
 
+  const inputRef = useRef(null)
+  useLayoutEffect(() => inputRef?.current?.focus(), [contactsNotInGroup])
+
```
Simon-Laux, line 349: what does this do?

Jikstra: It keeps the search bar focused in the addmember dialog

src/main/deltachat/chat.ts
Jikstra: We cannot do that in the ui without a store. And there we would have the same problem. If you want to fix this, you need to do it in core, and even there it will be hard.

Simon-Laux: > i think there are some self healing measures if the member lists are out of sync.
I don't think so

It would work if we would do it in the frontend - compare what was there when the dialog was opened and whats the resulting members after the action is done -> that way no sudden groupstate state change could result in unwanted behaviour.

PR #2383, opened 2021-10-01
WIP: Strict typescript

Simon-Laux:
- [ ] wait until #2382 is merged
- [ ] fix eslint warnings & errors

the main process is not strict yet, because the deltachat controllers' current context can be null and checking for it in every method would be a pain / make the code look unnecessarily complicated, we can do that once we extracted the deltachat part more or use https://github.com/deltachat/dc_cmd_api so we don't have to bother about that aspect.


Simon-Laux:
found some bugs while rebasing:

- [ ] changing name of contact does not refresh name in dialog
- [ ] modify group memebers doesn't update the memberlist
- [X] empty quote draft always shown
- [X] remove member button in group contact list isn't shown

PR #2401, opened 2021-11-08
Dont asynchronously load quotes, preload them in JsonMessage

Jikstra:
This makes sure that quotes dont popup


Simon-Laux (changes requested):
looks like you forgot to handle the case where the quoted message was deleted

src/shared/shared-types.d.ts
```diff
@@ -129,7 +129,17 @@ export type JsonLocations = {
...
+  quote: MessageQuote | null
+}
+
+export type MessageQuote = {
+  messageId: number
+  text: string
+  displayName: string
+  displayColor: string
+  overrideSenderName: string
+}
```
Simon-Laux, line 142: this type needs to have variant without `displayName`, `displayColor` and `overrideSenderName` for the case that the quoted message is not found in the database (for example when that message was deleted.

Jikstra: And also messageId i guess?

Simon-Laux: yes, see https://js.delta.chat/classes/message.html#getquotedmessage message can be null.

Simon-Laux (approved):
code lgtm

PR #2408, opened 2021-11-13
🚧 async qr group join

### LINKED ISSUE #2370: add missing calls to setStockTranslation()
recent core requires some more strings, meanwhile they're also on transifex and partly translated.

https://github.com/deltachat/deltachat-ios/pull/1361/files gives an overview about what is probably missing.

CAVE: take care not to confuse DC_STR_EPHEMERAL_MINUTES and DC_STR_EPHEMERAL_MINUTE, same for SECOND(S), HOUR(S) and WEEK(S)

also, DC_STR_QUOTA_EXCEEDING_MSG_BODY may contain the percent sign, which, however is escaped by `%%` - core takes care to show only one, however, this may need parser/converter adaptions (on android, i needed to adapt a test).

the issue is not urgent and should not go into 1.22, core just uses an english fallback. however, at some point, it is nice to have eg. "Connectivity view" translated.

### LINKED ISSUE #2393: use core api dc_chat_can_send()
add `dc_chat_can_send()` to dc node and use it in desktop.

So that property is checked first and if it is false then the reason text is figured out.

Simon-Laux:
- [ ] Bug: Optional: disabled reason says "user is not in group", maybe we should move diabled reason to core and add one for the secure-group-join-case﻿
- [x] needs a DeltaChat node version with https://github.com/deltachat/deltachat-node/pull/528 merged

closes #2393
closes #2370

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 - remove dependencies that are now not needed anymore (`punycode`, `simple-markdown`)
 - dont load quoted messages asynchronously
 - update deltachat-node to v1.65.0
+- Join group via qr-code is now async (group already opens, no wait time)
 
 ### Fixed
 - don't show logo twice in notifications (because macOS already shows applogo)
--- a/src/main/deltachat/chatlist.ts
+++ b/src/main/deltachat/chatlist.ts
@@ -34,12 +34,6 @@ export default class DCChatList extends SplitOut {
     return this.controller.selectedChatId
   }
 
-  async onChatModified(chatId: number) {
-    // TODO: move event handling to store
-    const chat = await this.getFullChatById(chatId)
-    this.controller.sendToRenderer('DD_EVENT_CHAT_MODIFIED', { chatId, chat })
-  }
-
   _chatListGetChatId(list: ChatList, index: number) {
     return list.getChatId(index)
   }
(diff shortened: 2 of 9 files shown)


r10s:
the current info-line that would appear on qr-group-join is maybe not perfect for that case however, it is not _that_ bad. the line reads "You cannot write because you are not in this group. To join, ask another member." - both still fit somehow - you are not in the group and you can still ask another member to join - also, it appears only for a short moment and the info-messages clarify the situation

best would be to hide the info-line for qr-group-join, everything that's needed to know is already written down in info messages. there is also the option to hide the info-line unconditionally (this is what android/ios do), and improve the info-message for leave/kick.

however, to move forward, i suggest to leave the line as is, it affects only a tiny detail, it is the only blocker and is a bit more complicated, including lots of discussions. we can then decide at some later point how to proceed on that

Simon-Laux:
I didn't found the translations for the following core stock strings:
- DC_STR_MSGLOCATIONENABLED
- DC_STR_MSGLOCATIONDISABLED
- DC_STR_ERROR_NO_NETWORK
- DC_STR_SERVER_TURNED_OFF
- DC_STR_SYNC_MSG_SUBJECT
- DC_STR_SYNC_MSG_BODY

r10s:
> I didn't find the translations for the following core stock strings

afaik, they are either fading out ("No Network", "Turned off" was used for some migration only) or are in experimental state and not really settled (locations, sync). therefore. so, it is fine to rely on the english strings provided by core in this case for now.

PR #2409, opened 2021-11-14
Convert status input to textarea

Jikstra:
Screenshot_20211114_221751

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 
 ### Changed
 - use strict typescript for ui code
+
 ## [1.22.2] - 2021-09-23
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -23,9 +23,11 @@
 - don't show logo twice in notifications (because macOS already shows applogo)
 - disable editing of left/readonly groups
 - fix member list being incomplete in chat requests
+- fix setting selfstatus
```
Simon-Laux, line 26: do we need to include bugfixes for bugs that were introduced in this version in the changelog don't they cancel each other out?

src/renderer/components/Login-Styles.tsx
```diff
@@ -6,6 +6,7 @@ import {
   Intent,
   ProgressBar,
   Switch,
+  TextArea,
```
Simon-Laux, line 9: make sure sealant passes

CHANGELOG.md
Jikstra: Ah right, that should go away again

PR #2410, opened 2021-11-15
Fix repositioning on fetch more

Jikstra:


scss/message/_message-list.scss
```diff
@@ -31,6 +31,7 @@
   ul {
     list-style: none;
     min-width: 200px;
+    margin: 10px 0px;
```
Simon-Laux, line 34: fixed with a margin, could different composer sizes be still a problem then or does composer size not matter for this issue?

Jikstra:
```
------------------
[14px]
# Message
[10px]
# Message
[10px]
# Message
[14px]
------------------
[Composer]
```

Have a look at this representation. If you scroll all the way up to where a new message should get loaded, you have 14px to the top of the messagelist. And your scrollTop is 0px. Now all the new messages get loaded and displayed, and you should get to the position where you were last. There you had 14px of nothing, now putting you back gives you 10px of nothing and 4px of a message. That looks weird, it looks like something jumps in where it shouldnt be.

PR #2418, opened 2021-11-17
[wip] Make ChatList resizable

Jikstra:
# ToDo
- [ ] ChatList needs to hide certain things on a certain treshold
- [ ] fix laggyness
- [ ] save value somewhere
- [ ] the x button for the search bar floats at a wrong position

### DIFF
--- a/scss/_global.scss
+++ b/scss/_global.scss
@@ -1,3 +1,7 @@
+:root {
+  --chatListWidth: 30%;
+}
+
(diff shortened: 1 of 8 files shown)


r10s:
yeah!

Simon-Laux:
- [ ] when searching it should probably expand, or the search would need to be a drawer / overlay or sth like that

PR #2427, opened 2021-11-23
Overhaul chat store and add typings

Jikstra:
# ToDo
- [x] Sometimes scrolling jumps around?

### DIFF
--- a/src/renderer/components/Menu.tsx
+++ b/src/renderer/components/Menu.tsx
@@ -6,7 +6,6 @@ import {
   useTranslationFunction,
   SettingsContext,
 } from '../contexts'
-import { useChatStore } from '../stores/chat'
 import { Menu } from '@blueprintjs/core'
 import {
   openLeaveChatDialog,
@@ -17,6 +16,7 @@ import {
   openMuteChatDialog,
   unMuteChat,
   sendCallInvitation,
+  unselectChat,
 } from './helpers/ChatMethods'
 import { runtime } from '../runtime'
 import { Screens } from '../ScreenController'
@@ -40,7 +40,6 @@ export function DeltaMenuItem({
 
 export default function DeltaMenu(props: { selectedChat: FullChat | null }) {
   const { selectedChat } = props
-  const chatStoreDispatch = useChatStore()[1]
 
   const tx = useTranslationFunction()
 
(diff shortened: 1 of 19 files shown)


src/renderer/components/Menu.tsx
```diff
@@ -55,7 +54,7 @@ export default function DeltaMenu(props: { selectedChat: FullChat | null }) {
 
   const logout = () => {
     if (selectedChat) {
-      chatStoreDispatch({ type: 'UI_UNSELECT_CHAT' })
+      unselectChat()
     }
```
Simon-Laux, line 58: did this even make sense in the first place, do we want to deselect the chat when switching accounts?

src/renderer/stores/store.ts
```diff
@@ -50,13 +50,13 @@ export class Store<S> {
...
-      /*this.log.debug(
+      this.log.debug(
         `DISPATCHING of "${action.type}" changed the state. Before:`,
         this.state,
         'After:',
         state
-      )*/
-      this.log.debug(`DISPATCHING of "${action.type}" changed the state.`)
+      )
+      //this.log.debug(`DISPATCHING of "${action.type}" changed the state.`)
```
Simon-Laux, line 59: logging all this might be slow things down when debug logging is active.

src/renderer/stores/store.ts
```diff
@@ -80,8 +80,20 @@ export class Store<S> {
...
+  ) {
+    const modifiedState = stateReducer(this.state)
+    if (modifiedState === undefined) return
+    this.log.debug(
+      `${description} changed the state. Before:`,
+      this.state,
+      'After:',
+      modifiedState
+    )
+    //this.log.debug(`${description} changed the state`)
```
Simon-Laux, line 95: same for this verbose logging

src/renderer/stores/chat.ts
```diff
@@ -1,263 +1,356 @@
...
+        }
+      }, 'setMessageIds')
+    },
   }
-  return state
-})
+  effect = {
+    selectChat: async (chatId: number) => {
+      // these methods were called in backend before
+      // might be an issue if DeltaBackend.call has a significant delay
```
Simon-Laux, line 206: `const [a,b] = await Promise.all([a,b])` to the rescue

Simon-Laux: but that's sth for the next pr if this even makes any problems at all

src/renderer/stores/store.ts
Jikstra: That's only there for debugging

src/renderer/components/Menu.tsx
Jikstra: Je ne sais pas, feel free to remove it and see what happens ^^

Simon-Laux (approved):

PR #2455, opened 2021-12-08
download on demand / auto-download

Simon-Laux:
- [X] ~~find out why I can not get the failstate: I deleted the message on the server and then pressed download, not it says "downloading" for eternity~~ looks like a core issue

### DIFF
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -1,1 +1,5 @@
-{}
+{
+  "download_failed_please_try_again": {
+    "message": "Message download failed, please try again"
+  }
+}
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -1,5 +1,5 @@
 {
-  "download_failed_please_try_again": {
-    "message": "Message download failed, please try again"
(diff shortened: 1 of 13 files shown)


_locales/_untranslated_en.json
```diff
@@ -1 +1,5 @@
-{}
+{
+  "download_failed": {
+    "message": "Download failed"
```
Jikstra, line 3: There's no translation for this?

Jikstra:
Tests fail but otherwise looks good :)

_locales/_untranslated_en.json
r10s: not yet, that is missing somehow, and the state seems to be hard to achieve anyway, maybe there is a core bug, not sure.

however, as the missing string produces some noise, i just added it :) https://github.com/deltachat/deltachat-android/pull/2167 - however, i do not think it is super-important to pull it in for desktop, it would probably hardly be shown anyway,

Simon-Laux: nice, we can then just remove it from `_untranslated_en.json` when we pull the translation strings again as it has the same key.

Issue #2442, opened 2021-11-28
Styling for signature textarea needed 


Simon-Laux:
because it looks strange in dark theme:

<img width="496" alt="Screenshot 2021-11-28 at 22 28 30" src="https://user-images.githubusercontent.com/18725968/143786804-89cc5067-3afb-4440-9d6b-a7c09ca8f04a.png">

ejgonzalez17:
Hello, I would like to contribute by making this issue

ejgonzalez17:
Hello, I worked on this issue and in the pull request I attached an image where the "Delete Profile Image" button is shown when it is disabled in the different themes, taking care that the bug one did not happen again. I will be attentive to your comments.

Jikstra:
@ejgonzalez17 looking at this issue again, @Simon-Laux probably was talking about the textarea element, but nevertheless the disabled Delete profile image button looked horrible too ^^

ejgonzalez17:
What could be some guides of what would be the correct style for textarea?

Jikstra:
> What could be some guides of what would be the correct style for textarea?

I would say making it similiar to the "Your name" inout field. So `color: var(--bp3InputText)` and a transparent background. Maybe even adding this white border-bottom?

ejgonzalez17:
I changed the background and also added to your color var (- mp3 OutputText), Here is what it would look like, is that ok?
<img width="589" alt="Screen Shot 2022-01-06 at 10 03 19 AM" src="https://user-images.githubusercontent.com/70716664/148412283-9f19a23f-3eab-45a5-8a2b-01854715fd47.png">

Jikstra:
@ejgonzalez17 looks really good :+1:

ejgonzalez17:
Okay, I'll proceed to do the Pull Request =)

PR #2478, opened 2021-12-23
Change the color style in the "Delete Profile Image" button when it is disabled

ejgonzalez17:
**Solves #2442 (Styling for signature textarea needed)**

**Description**
The text color of the "Delete Profile Image" button when disabled was not displayed correctly for some Themes, mainly for the dark theme.

Screen Shot 2021-12-23 at 8 41 03 AM

**Why**
The color that was set for all themes in the disabled button when editing the profile was suitable only for light themes, since in dark themes it did not display the text.

**Additional information** 
The style bug in the "Delete Profile Image" button is corrected when it is disabled, the change is tested in the different themes.

- Photo of the button view in the different themes

Screen Shot 2021-12-23 at 8 59 32 AM

### DIFF
--- a/scss/misc/_delta-buttons.scss
+++ b/scss/misc/_delta-buttons.scss
@@ -17,7 +17,7 @@ button.delta-button-round {
     border-color: rgba(206, 217, 224, 0.5);
     background-image: none;
     cursor: not-allowed;
-    color: rgba(92, 112, 128, 0.6);
+    color:white;
     &:hover {
       background-color: rgba(206, 217, 224, 0.5);
     }
--- a/scss/misc/_delta-buttons.scss
+++ b/scss/misc/_delta-buttons.scss
@@ -1,3 +1,4 @@
+
 button.delta-button-round {
   margin-top: 10px;
   background-color: var(--bp3-input-focused);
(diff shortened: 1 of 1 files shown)


scss/misc/_delta-buttons.scss
```diff
@@ -17,7 +18,7 @@ button.delta-button-round {
     border-color: rgba(206, 217, 224, 0.5);
     background-image: none;
     cursor: not-allowed;
-    color: rgba(92, 112, 128, 0.6);
+    color:rgb(51, 49, 49);
```
Jikstra, line 21:
```suggestion
    color: var(--globalText);  
```

I think using this variable is a better fit, as it's respecting the theme color

scss/misc/_delta-buttons.scss
```diff
@@ -29,3 +30,5 @@ button.delta-button-round {
     background-clip: unset;
   }
 }
+
+
```
Jikstra, line 34: Some not needed newlines :)

Jikstra:
Thank you :) A small suggestion is using var(--globalText), this uses a color variable that themes can define and probably is a better fit then introducing a color.

scss/misc/_delta-buttons.scss
ejgonzalez17: Alright, I'll remove the lines, thanks for the correction =)

ejgonzalez17:
I will also change the color for var (- globalText) , for a better fit in the colors 
Screen Shot 2021-12-27 at 5 33 28 PM

ejgonzalez17:
Hi @Jikstra , I have already made the changes taking care of the code format , thanks for the review :)

Issue #2474, opened 2021-12-19
Saving a file does not report failures


adrianheine:
- Operating System (Linux/Mac/Windows/iOS/Android): Debian GNU/Linux
- Delta Chat Version: 1.22.2
- Expected behavior: An error is reported
- Actual behavior: No feedback whatsoever is provided
- Steps to reproduce the problem: Try to save a file in a location that's not writable
- Screenshots:
- Logs:

```
Error occurred in handler for 'saveFile': [Error: EACCES: permission denied, copyfile '/home/XXX/.config/DeltaChat/accounts/XXX/dc.db-blobs/XXX.jpg' -> '/x'] {
  errno: -13,
  code: 'EACCES',
  syscall: 'copyfile',
  path: '/home/XXX/.config/DeltaChat/accounts/XXX/dc.db-blobs/XXX.jpg',
  dest: '/x'
}
```

IrvinLara9:
I would like to work on this issue!

Simon-Laux:
was solved by #2479

useful hint: you can write `closes #issue-number` in your pr description so the issue gets automatically closed once the pr is merged.

PR #2479, opened 2021-12-29
Add error message when you try to download an image in a folder with no write permission

IrvinLara9:
Solves #2474 (Saving a file does not report failures)

**Description**
Now when you try to download an image in a folder where you don't have write permission, an error alert is displayed to the user, indicating the problem.

**Why**
When the error of the write permission occurred, the app didn't notify the user that an error occurred.

**Additional information**

- Photo of the current error message:
Screen Shot 2021-12-28 at 20 14 50

### DIFF
--- a/src/main/ipc.ts
+++ b/src/main/ipc.ts
@@ -140,6 +140,11 @@ export async function init(cwd: string, logHandler: LogHandler) {
             'Permission Error',
             `Cannot write in this folder. You don't have write permission`
           )
+        } else {
+          dialog.showErrorBox(
+            'Unhandled Error',
+            `Cannot copy file. Error: ${error}`
+          )
         }
       }
     }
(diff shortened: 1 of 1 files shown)


src/main/ipc.ts
```diff
@@ -132,7 +132,16 @@ export async function init(cwd: string, logHandler: LogHandler) {
...
-      await copyFile(source, filePath)
+      try {
+        await copyFile(source, filePath)
+      } catch (error) {
+        if (error.code == 'EACCES') {
+          dialog.showErrorBox(
+            'Permission Error',
+            `Cannot write in this folder. You don't have write permission`
+          )
+        }
```
Jikstra, line 143: Can you also handle the else case? Maybe something as 
```
} else {
          dialog.showErrorBox(
            'Unhandled error',
            `Cannot copy file. Error: ${error}`
          )
}
```

Jikstra:
Thank you! Looks good :) If you have time & motivation it would be nice to also have the else case handled. But already like this I would be happy to merge this pr :)

IrvinLara9:
Hi @Jikstra Thank you for the comment! I have already added your suggestion!

Jikstra:
Perfect, thank you!

Issue #2436, opened 2021-11-26
Connectivity view uses wrong title


Jikstra:
![Screenshot_20211126_140433](https://user-images.githubusercontent.com/34889164/143585458-62011997-70a4-4a95-b81d-0c3ea3e806eb.png)

IrvinLara9:
I'll work on this!

IrvinLara9:
This was solved in #2480

PR #2480, opened 2021-12-30
Fix connectivity view title

IrvinLara9:
Solves #2436 (Connectivity view uses wrong title)

**Description**
Now when you enter the connectivity view, the correct title is shown.

**Why**
When you entered the connectivity view an incorrect title was displayed.

**Additional information**

Photo of the current connectivity view:

image

### DIFF
--- a/src/renderer/components/dialogs/Settings-Connectivity.tsx
+++ b/src/renderer/components/dialogs/Settings-Connectivity.tsx
@@ -37,7 +37,7 @@ export default function SettingsConnectivityDialog({
         width: '500px',
       }}
     >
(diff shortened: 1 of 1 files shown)


Jikstra:
Thank you again :)

Can you add a line to the CHANGELOG.md file? It's a practice we try to do to keep our changelog file updated. This makes the version release easier (for the "whats new" message) and keeping an overview, both for us and users.

I don't like to annoy contributors with those nitpicks, so in case you don't feel like doing it I can do it too.

IrvinLara9:
No problem!! I already did 👍

PR #2486, opened 2022-01-03
change translation string for download on demand to pref_show_emails_all

### LINKED ISSUE #2468: rename Auto-Download "No Limit" to "All"
if auto-download is disabled, the option should read just "All", not "No Limit".

we used "No Limit" in the first iteration on android, however, decided that "All" is easier to understand then. the string "No Limit" can be removed at some point from transifex then. for the string "All", just use the same as for "Show Classic E-Mails: All".

Simon-Laux:
closes #2468

mind that I still not agree with reusing the the `pref_show_emails_all` key because of its specific name, but whatever

### DIFF
--- a/src/renderer/components/dialogs/Settings-DownloadOnDemand.tsx
+++ b/src/renderer/components/dialogs/Settings-DownloadOnDemand.tsx
@@ -15,7 +15,7 @@ export default function SettingsDownloadOnDemand(props: {
   const tx = useTranslationFunction()
 
   const options: { label: string; value: number }[] = [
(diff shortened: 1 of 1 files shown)


r10s:
> mind that I still not agree with reusing the the pref_show_emails_all key because of its specific name, but whatever

i agree, it should just be named "all". to save data and work, in general, we use unspecific strings and reuse them as far as possible. some older strings do not follow that convention by their names.

however, all in all it seems not worth the effort to rename existing keys on all platforms :)

Jikstra (approved):

Issue #2203, opened 2021-03-31
Add an Add-Attachment menu


Simon-Laux:
A menu when clicking on the paperclip 📎 icon that allow you to specify whether you want to send an file image or video hangout/chat invitation.

naomiceron:
Hello, I would like to work on this, this is my progress so far, does it seems about right?
![Screen Shot 2022-01-04 at 11 48 30](https://user-images.githubusercontent.com/58761246/148112277-10b6fd14-a4a5-40d5-86d6-58f65bd48a48.png)

naomiceron:
I haven't been able to find the extension that an Invitation would need, is it a .qr? or should it allow to select any type of attachment? Thanks in advance! :)

r10s:
@naomiceron looks great! thanks a lot for taking care!

to your question: as far as i know, there is currently no extension or file format for invitations, neither android nor ios has sth. like that. invitations are either strings that are copy+pasted around or images containing a qr-code, the latter is not supported on all platforms.

it probably makes sense to think about a separate invitation format, however, in the context of this approach, this seems out of scope. so maybe just leave the "invitation" thingie out for now and start with the more concrete things.

Jikstra:
> it probably makes sense to think about a separate invitation format, however, in the context of this approach, this seems out of scope. so maybe just leave the "invitation" thingie out for now and start with the more concrete things.

I think invitation is about video chat invitation, which surely exists. But actually I'm not so sure if it's not better to have that in the context menu.

@naomiceron what would be your idea on how to filter files for video/pictures/documents...? Because desktops dont really have that gallery like thing that android/iphones have. But maybe I'm overseeing something?

Simon-Laux:
I think that its great ui wise, also a good place for the video chat invitation.

@jikstra we just need a list of what we support to be displayed, everything other is just a file. the other ui option would be autodetection of view type from attachment and a toogle for sending it with/without core-compression (send as file or as image/video)

I'm currently reworking on the composer system so draft ids stay (also hopefully I can finally fix the rare bug where drafts sometimes stay after sending), so @naomiceron it would be enough if you just made a pr with the ui part, I can then integrate it and do the harder draft handling stuff.

r10s:
> I think invitation is about video chat invitation, which surely exists

thanks for clarification!

[...]

PR #2488, opened 2022-01-05
Add an attachment menu

naomiceron:
Solves #2203 (Add an Add-Attachment menu) 

**Description**
Adds a menu when clicking on the paperclip icon that allows specifying whether you send Files or Media. 

**Why**
An enhancement to the existing paperclip icon function.

**Additional information**
Photo of current behavior:
Screen Shot 2022-01-05 at 12 44 26


src/renderer/components/attachment/menuAttachment.tsx
```diff
@@ -0,0 +1,125 @@
...
+            'docx',
+            'xls',
+            'xlsx',
+            'ppt',
+            'ppt',
+            'pdf',
+            'txt',
+            'csv',
+            'log',
+            'zip',
```
Jikstra, line 57: I think we shouldn't filter Files by extension, but show all of them. Otherwise it wouldn't be possible to send for example a .db file.

src/renderer/components/attachment/menuAttachment.tsx
```diff
@@ -0,0 +1,125 @@
...
+      addFileToDraft(file)
+    }
+  }
+  //function for media
+  const addFilenameMedia = async () => {
+    const file = await runtime.showOpenFileDialog({
+      filters: [
+        {
+          name: 'Media',
+          extensions: ['jpg', 'jpeg', 'png', 'gif', 'mkv', 'avi', 'mp4'],
```
Jikstra, line 74: Let's put this array into a shared constant variable. This way we can use it at various places and don't have to copy it. I would propose having it as `const MEDIA_EXTENSIONS = ['jpg'..., 'mp4'] in `src/shared/constants.ts`

Jikstra (changes requested):

Jikstra:
Looks good to me besides the comments. One other thing, you removed the package-lock.json file. Can you bring it back with `git checkout origin/master package-lock.json` ?

naomiceron:
Yes! thank you for the comment, I will change it :)

src/renderer/components/attachment/menuAttachment.tsx
Simon-Laux: also we should be able to send also media files as files, so that way there is still a way to send them without core compression

naomiceron:
With that last commit I think it solves the issues, if not please let me know.

Simon-Laux (approved):

Simon-Laux:
there is still one eslint error (I think the warning is there also on master) - please run `npm run test`.
also a small entry in changelog.md about your addition would be nice.

naomiceron:
Yes of course, that's done with this new commit :)

Simon-Laux:
prettier still complains, you can fix that with `npm run fix-formatting`

naomiceron:
I ran the command `npm run fix-formatting` but it doesn´t show me anything wrong, so that is weird. Also the command `npm run check-formatting` says that all files matched Prettier code style.

Jikstra:
> I ran the command `npm run fix-formatting` but it doesn´t show me anything wrong, so that is weird. Also the command `npm run check-formatting` says that all files matched Prettier code style. I will keep looking into what is making the problem. Here I will leave a screenshot of the results. Screen Shot 2022-01-10 at 16 20 28

I currently have the same problem on https://github.com/deltachat/deltachat-desktop/pull/2490. Locally prettier doesn't complain but on the ci it does. Everything else looks good, so let's ignore this for now :)

Issue #2491, opened 2022-01-06
When selecting the image profile, it does not accept .jpeg .ico .apng format


ejgonzalez17:
- **Operating System (Linux/Mac/Windows/iOS/Android):** Mac Big Sur

- **Delta Chat Version:** v1.26.0 

- **Expected behavior:**
Accepts different image formats

- **Actual behavior:**
The image formats that are being accepted are only: .png, .jpg . When trying to select an image with .jpeg format, it does not allow to choose the image
<img width="878" alt="Screen Shot 2022-01-06 at 2 13 30 PM" src="https://user-images.githubusercontent.com/70716664/148446162-bbbcfcca-7b89-478a-a9ef-6090c5c2f099.png">

- **Steps to reproduce the problem:**
1. Go to settings
2. Edit Profile
3. Select profile picture

- **Screenshots:**
![Screen Recording 2022-01-06 at 2 23 48 PM](https://user-images.githubusercontent.com/70716664/148447316-c8a69f1b-86ca-4b87-b0fd-c5448c70a331.gif)

ejgonzalez17:
When reviewing the code I saw that it would only be necessary to add some image format.

<img width="520" alt="Screen Shot 2022-01-06 at 3 50 56 PM" src="https://user-images.githubusercontent.com/70716664/148457121-4c0465a9-1e3b-47f4-a281-f5ecae491d46.png">
s

ejgonzalez17:
Making the change and adding the formats it was already possible for me to select the images with this type of format
<img width="837" alt="Screen Shot 2022-01-06 at 4 37 39 PM" src="https://user-images.githubusercontent.com/70716664/148462144-294d79ca-d96c-4797-9723-efad6c9c9264.png">

PR #2492, opened 2022-01-06
Add format image

ejgonzalez17:
**Solves #2491 (When selecting the image profile, it does not accept .jpeg .ico .apng format)**

**Description**
Add image formats in the process of selecting a new image in the profile

**Why**
When trying to select an image in the profile photo, formats such as .jpeg or .ico are not accepted.

**Additional information** 
Formats are added in Setting-Profile.tsx and in ViewGroup.tsx
Screen Shot 2022-01-06 at 4 21 25 PM

Screen Shot 2022-01-06 at 4 21 58 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Update translations (3.1.2022)
 - Make disabled "delete profile image" button more readable in dark themes @ejgonzalez17 #2478
 - Update inApp help (3.1.2022)
+- Add image formats when selecting the image profile @ejgonzalez17 #2492
 
 ### Fixed
 - Fix overflow in long links inside quotes @naomiceron #2467
@@ -14,6 +15,7 @@
 - Do not double log core events
 - Fix Bulgarian language name (uppercase first letter)
 - Fix signature text styling @ejgonzalez17 
+- Fix Add image formats @ejgonzalez17
 
 ## [1.26.0] - 2021-12-15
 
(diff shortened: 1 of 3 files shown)


r10s:
adding `.jpeg`, should be fine. maybe also add `.jpe` at the same time, that is also used here and there for JPEG images.

for `.apng`, `.ico` and `.webp` - they should not be used for own profile images. the images are sent to other clients using other ui (android, ios) and it is not clear if all other ui support these formats or if core converts them to png/jpg (i suspect that).

ejgonzalez17:
Okay , I'll change so that only .jpeg and .jpe are added .

r10s:
@ejgonzalez17 thanks a lot for taking care!

Simon-Laux (changes requested):
@ejgonzalez17 can you run `npm run fix-formatting` and mention your change in the `CHANGELOG.md` file?
Besides these 2 trivial points the PR looks good to me, thanks for contributing!

ejgonzalez17:
Hi , sorry for the delay, I have already uploaded the marked changes , I am waiting for your corrections or comments.

Jikstra:
> Hi , sorry for the delay, I have already uploaded the marked changes , I am waiting for your corrections or comments.

Looks good. I think it makes sense to move the image formats in constant array so that it can get reused and only needs changes at one place. Similiar to what @naomiceron did over here: https://github.com/deltachat/deltachat-desktop/pull/2488#discussion_r780772205

I will do this quickly after merging. Thank you :)

PR #2497, opened 2022-01-07
Offline Toast Style Update

### LINKED ISSUE #2454: Offline Toast obscures message composer on small width window
image_2021-12-06_01-25-05

maybe have a max with so that it cannot be wider than the chatlist or maybe an icon for the retry text instead of the long text? or split it into two lines instead of one?

trujillo9616:
Issue description:
Fixes #2454 

Offline Toast obscures message composer on small width window.

Proposed Solution:

As suggested, I modified the file `_conncetivity-toast.scss` for the divs inside the ConnectivityToast class to have a display type of block, instead of line-block.

Screen Shot 2022-01-07 at 5 03 35 PM

The new behavior of the toast message can be observed here. The Archived Chats button is still clickable when the Offline Toast message appears to the user. At no moment does the toast obscure the message composer.

recording

Suggestions are welcomed! 😁

### DIFF
--- a/scss/_connectivity-toast.scss
+++ b/scss/_connectivity-toast.scss
@@ -22,7 +22,7 @@
   }
 
   div {
-    display: inline-block;
+    display: block;
     color: var(--colorPrimary);
     margin-left: 4px;
     cursor: pointer;


Jikstra:
Thank you very much! I think this issue annoyed already quite some people :)

PR #2501, opened 2022-01-10
fix: remove profile picture

### LINKED ISSUE #2472: cannot delete profile picture in DC Desktop 1.26.0
I put a profile picture for my DC desktop account and when I wanted to delete it, it was still there afterwards.

cavesdev:
Fixes #2472  

## Solution
Changed the empty string representation to 'null' when there was no profile picture. Apparently, when sending an empty string to the backend it gets interpreted as a path (which is obviously an invalid path) and throws an I/O error. 

To avoid changing many type annotations, I only added an if statement when sending the profile picture representation to the backend and added the null type on that function's parameter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - Do not double log core events
 - Fix Bulgarian language name (uppercase first letter)
 - Fix signature text styling @ejgonzalez17 
+- Fix the profile picture removal @cavesdev #2472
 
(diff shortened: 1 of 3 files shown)


cavesdev:
Updated changelog!

Jikstra:
Perfecto! Thank you very much :+1:

Yettte:
Muchas gracias amigo! :)

PR #2502, opened 2022-01-10
fix: removed horizontal scroll in autocrypt dialogs

### LINKED ISSUE #2277: Can scroll sideways in autocrypt-setup dialog
where the numbers are shown in the autocrypt setup dialog, I can scroll sideways. this should not be the case, as then some numbers are obstructed/hidden, while before scrolling to the side I could clearly read them all.

also the receiving dialog has a similar sideways scrolling issue:

cavesdev:
Fixes #2277 

## Solution
The solution was to remove the BlueprintJS's classes on parent `div`s (`Classes.BODY_DIALOG`). It does not affect in any way the styles of the dialogs apart from disabling horizontal scroll.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - Fix connectivity status hiding composer @trujillo9616
 - Fix overflow in Confirmation Dialog @Abhijnan-Bajpai 
 - Fix the profile picture removal @cavesdev #2472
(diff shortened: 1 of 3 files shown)


Jikstra:
Looks good, those nested body classes are indead weird. Normally one needs them once, but they are too much :D Thank you!

PR #2507, opened 2022-01-13
In AccountSetupScreen login by pressing enter

flipsimon:


src/renderer/components/screens/AccountSetupScreen.tsx
```diff
@@ -61,6 +61,21 @@ export default function AccountSetupScreen({
     }
   }
 
+  const onKeyDown = (ev: KeyboardEvent) => {
```
Simon-Laux, line 64: eslint's rules-of-react hooks check complains here, run `npm run test` or look bellow on gh.
You probably want a useCallback or useMemo here.

Simon-Laux (commented):
looks good, besides the small eslint warning

PR #2509, opened 2022-01-13
feat: add new contact within edit group dialog

### LINKED ISSUE #2256: cannot add new e-mail addresses directly to a group
one cannot add an e-mail-address, that is not known as a contact yet, directly to a group - neither on group creation nor on group editing.

maybe, i've oversee something, however, my path was _"New chat / enter address / New contact"_ to create a  one-to-one chat. then, _"New chat / New group / Add members / enter address [x] / Ok / Create group"_ to create group with that contact.

it would be great, if one could add a new contact already at [x]

cavesdev:
Fixes #2256 

## Solution

Added the possibility to add new contacts directly within the edit group members dialog. 

Once added, the query string changes because I wasn't able to trigger a list update without changing it.

https://user-images.githubusercontent.com/24899540/149410010-6a9db5b3-41fc-4660-b0ee-6c4ac1224b6a.mov

### DIFF
--- a/src/renderer/components/dialogs/CreateChat.tsx
+++ b/src/renderer/components/dialogs/CreateChat.tsx
@@ -5,6 +5,7 @@ import React, {
   useRef,
   useLayoutEffect,
   useEffect,
+  ChangeEvent,
 } from 'react'
 import { Card, Classes } from '@blueprintjs/core'
 import { C } from 'deltachat-node/dist/constants'
@@ -18,7 +19,6 @@ import {
 } from '../contact/ContactList'
 import {
   PseudoListItem,
-  PseudoListItemNoSearchResults,
   PseudoListItemShowQrCode,
   PseudoListItemAddMember,
   PseudoListItemAddContact,
@@ -334,6 +334,18 @@ export function AddMemberInnerDialog({
     .map(([contactId, _contact]) => contactId)
 
   const [contactIdsToAdd, setContactIdsToAdd] = useState<JsonContact[]>([])
+  const [{ queryStrIsValidEmail }, updateContacts] = useContactsNew(
+    C.DC_GCL_ADD_SELF,
+    ''
+  )
+  const [{}, onSearchChangeNewContact] = useContactSearch(updateContacts)
+
+  const onSearchChangeValidation = (query: ChangeEvent<HTMLInputElement>) => {
+    if (searchContacts.size === 0) {
+      onSearchChangeNewContact(query)
+    }
+    onSearchChange(query)
+  }
 
   const addOrRemoveMember = (contact: JsonContact) => {
     if (contactIdsToAdd.findIndex(c => c.id === contact.id) === -1) {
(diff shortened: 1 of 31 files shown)


src/renderer/components/dialogs/CreateChat.tsx
```diff
@@ -367,6 +379,35 @@ export function AddMemberInnerDialog({
...
+    await DeltaBackend.call('contacts.createContact', queryStr)
+
+    if (inputRef.current) {
+      const nativeInputValueSetter = Object.getOwnPropertyDescriptor(
+        window.HTMLInputElement.prototype,
+        'value'
+      )?.set
+      nativeInputValueSetter?.call(inputRef.current, queryStr.split('@')[0])
+      var inputEvent = new Event('input', { bubbles: true })
+      inputRef.current.dispatchEvent(inputEvent)
```
Simon-Laux, line 394: I assume this code is for updating the search results after creating the new contact, right?

Simon-Laux:
eslint has quite a few warnings, you can see them with `npm run test`

src/renderer/components/dialogs/CreateChat.tsx
Jikstra: Aaaah this writes the email into the input field. I think you could do `updateContacts()` to refetch contacts and `onSearchChange(inputEvent)` to change the search input. This is probably a bit more error prone then doing it through the dom.

src/renderer/components/dialogs/CreateChat.tsx
```diff
@@ -334,6 +334,18 @@ export function AddMemberInnerDialog({
...
+  const [{ queryStrIsValidEmail }, updateContacts] = useContactsNew(
+    C.DC_GCL_ADD_SELF,
+    ''
+  )
+  const [{}, onSearchChangeNewContact] = useContactSearch(updateContacts)
+
+  const onSearchChangeValidation = (query: ChangeEvent<HTMLInputElement>) => {
+    if (searchContacts.size === 0) {
+      onSearchChangeNewContact(query)
+    }
```
Jikstra, line 346: For what is this code needed? It looks like you don't use the useContactSearch() state that you defined in line 341?

src/renderer/components/dialogs/CreateChat.tsx
Jikstra: Hmm I think a better approach would be to call `updateContacts()` & `onSearchChange(inputEvent)`

cavesdev: I know! I was trying to use the onChange handler but it is only called when the input changes directly in the application. This is the way to trigger the onChange handler "manually" or programatically.

cavesdev: If it is not updated then after adding a new contact you have to manually change the input again to be able to find the contact

src/renderer/components/dialogs/CreateChat.tsx
cavesdev: I had to use the useContactSearch handler because it does the input validation (if it's a valid email). I figured out it was better to validate the input using existing methods than adding a new regex

cavesdev:
Added an eslint-ignore line for the empty object (line 341) because only the setter is used, the actual value is not useful.

Jikstra:
> 

If you don't use the value you can also call the variable something starting with an `_` I think and then it's fine too.

cavesdev:
Changed to `_`. Also, fixed merge conflicts

Jikstra:
@cavesdev cool! I had played around with your code and found some solutions which I think are nicer:
1. Using onSearchChange() with a faked ChangeEvent to trigger the refetch of contacts over using the dom ref. I think the code is cleaner and more clear what is happening
2. I implemented automatically selecting the new created contact, I think this is better ux

I created a pr against your branch, you can review/merge it in your own fork: https://github.com/cavesdev/deltachat-desktop/pull/3

Jikstra:
I merged your changes in https://github.com/deltachat/deltachat-desktop/pull/2532
Thank you for you time and effort :)

PR #2516, opened 2022-01-14
Image zoom for full screen media view

flipsimon:


Jikstra:
Geiloooo!
One thing, can you do `git checkout origin/master package-lock.json` and commit the original package-lock.json file?

flipsimon:
Using the original package-lock file leads to errors as it's not synced to the package.json file anymore which contains react-zoom-pan-pinch. What should I do in this case?

Jikstra:
> Using the original package-lock file leads to errors as it's not synced to the package.json file anymore which contains react-zoom-pan-pinch. What should I do in this case?

Oh I think I was stupid. Of course we need a new package-lock.json if you added a dependency. Sorry :D

Jikstra:
Awwwesome!

Issue #2448, opened 2021-11-30
Search results cannot jump to the message body and context.


YamatoRyou:
- Operating System: x64 Windows 7
- Delta Chat Version: 1.22.1 & 1.25.1
- Expected behavior: When you click the search result, you should jump to the corresponding message body and display the context of this message.
- Actual behavior: Only the message details can be displayed when you click the search result.
- Steps to reproduce the problem: Type in keywords to get search results; click on a search result.
- Screenshots: 
![{4F7B314F-6737-30F9-B4AE-F37626DD1252}](https://user-images.githubusercontent.com/48104960/144118296-5f0ed3f3-c097-45f1-b6a1-d9cb317f74b5.png)

- Logs: n/a

Jikstra:
Yes this is a missing feature. I will hopefully find time to implement this soon :)

Jikstra:
@YamatoRyou we can leave it open :)

PR #2534, opened 2022-01-22
Implement jump to message

Jikstra:
Fixes https://github.com/deltachat/deltachat-desktop/issues/2448

# Todo
- [x] implement fetchNextPage
- [x] send message


src/renderer/stores/chat.ts
```diff
@@ -364,6 +373,82 @@ class ChatStore extends Store<ChatStoreState> {
...
+
+      const message = _message[msgId] as MessageType
+      if (message === null || message === undefined) {
+        throw new Error('jumpToMessage: Tried to jump to non existing message with id: ' + msgId)
+      }
+
+      const chatId = (message as MessageType).chatId
+
+      const chat = <FullChat>(
+        await DeltaBackend.call('chatList.selectChat', chatId)
```
Simon-Laux, line 389: only needed if its not the current chat

Jikstra: No we always do it. This way we start clean every time you jump to a message.

Simon-Laux: If that makes things easier for a first step.

PR #2536, opened 2022-01-24
First remove the draft, then send the message

Jikstra:
This feels a bit better then first sending the message, then the message pops up, then the draft disappears

### DIFF
--- a/src/renderer/components/composer/Composer.tsx
+++ b/src/renderer/components/composer/Composer.tsx
@@ -110,7 +110,6 @@ const Composer = forwardRef<
           ? draftState.quotedMessageId
           : undefined,
       })
-
     } catch (error) {
       log.error(error)
     } finally {
--- a/src/renderer/components/composer/Composer.tsx
+++ b/src/renderer/components/composer/Composer.tsx
@@ -98,6 +98,11 @@ const Composer = forwardRef<
         log.debug(`Empty message: don't send it...`)
         return
       }
+      /* clear it here to make sure the draft is cleared */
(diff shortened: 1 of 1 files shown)


Simon-Laux:
as I said before the right way would be sending the draft message, then reloading

Jikstra:
> as I said before the right way would be sending the draft message, then reloading, then the deletion is not necessary anymore. If it should still be "slow" then there is probably a bug elsewhere

So we need an ipc call ala sendDraftMessage(chatId) ?

Simon-Laux:
> So we need an ipc call ala sendDraftMessage(chatId) ?

yes

Jikstra:
Just talked with @flipsimon about this and we found a problem with the sendDraftMessage idea. As the setDraft is debounced and the core takes some time, it's possible that sendDraftMessage would send an old draft. I think it makes more sense to just use the freshest message value from the ComposerInputRef, first remove the draft, then send the message.

src/renderer/components/composer/Composer.tsx
```diff
@@ -98,18 +98,18 @@ const Composer = forwardRef<
         log.debug(`Empty message: don't send it...`)
         return
       }
+      /* clear it here to make sure the draft is cleared */
+      await DeltaBackend.call('messageList.removeDraft', chatId)
+      // /* update the state to reflect the removed draft */
+      window.__reloadDraft && window.__reloadDraft()
```
Simon-Laux, line 104: do this after sending it. or just remove the await before send

Simon-Laux (approved):

PR #2546, opened 2022-02-02
Enabled 'copy image' context menu option in gallery and chat view and…

flipsimon:
… removed 'copy text' option if text is empty.

### DIFF
--- a/src/main/deltachat/login.ts
+++ b/src/main/deltachat/login.ts
@@ -221,7 +221,6 @@ export function txCoreStrings() {
   strings[C.DC_STR_AUDIO] = tx('audio')
   strings[C.DC_STR_FILE] = tx('file')
   strings[C.DC_STR_ENCRYPTEDMSG] = tx('encrypted_message')
-  strings[C.DC_STR_STATUSLINE] = tx('pref_default_status_text')
   strings[C.DC_STR_MSGGRPNAME] = tx('systemmsg_group_name_changed')
   strings[C.DC_STR_MSGGRPIMGCHANGED] = tx('systemmsg_group_image_changed')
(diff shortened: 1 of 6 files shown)


flipsimon:
Yet to fix (but does not seem urgent):
If a message contains an image and (only) a link/email, the context menu options differ if one right-clicks on the image or on the text part of the message. When clicked on the image, the context menu offers options 'Copy image' and 'Copy text', when clicked on the text, the context menu offers 'Copy image' and 'Copy link/email'.

flipsimon:
@Simon-Laux branch based on new_core_1.71. Hope that was correct?

src/main/deltachat/login.ts
```diff
@@ -221,7 +221,6 @@ export function txCoreStrings() {
   strings[C.DC_STR_AUDIO] = tx('audio')
   strings[C.DC_STR_FILE] = tx('file')
   strings[C.DC_STR_ENCRYPTEDMSG] = tx('encrypted_message')
-  strings[C.DC_STR_STATUSLINE] = tx('pref_default_status_text')
```
Simon-Laux, line 224: technically this doesn't belong in the pr as it is for making the code work with the newest core,
which is already done in #2530

Simon-Laux (approved):

src/main/deltachat/login.ts
flipsimon: Thanks for reviewing! So now I should remove the changes from this file from the PR even if the tests fail then?

Jikstra: The tests should only fail locally for you if you use a newer deltachat-node :)

PR #2561, opened 2022-02-19
Fix refresh messagelist

Jikstra:
- [ ] sometimes a lock is not cleaned up properly?

### DIFF
--- a/src/renderer/components/dialogs/DisappearingMessages.tsx
+++ b/src/renderer/components/dialogs/DisappearingMessages.tsx
@@ -50,7 +50,7 @@ function SelectDisappearingMessageDuration({
         value={String(DisappearingMessageDuration.OFF)}
       />
       <Radio
-        key={'eph-1'}
+        key={'eph-1a'}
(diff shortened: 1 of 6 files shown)


src/renderer/components/dialogs/DisappearingMessages.tsx
```diff
@@ -48,6 +49,11 @@ function SelectDisappearingMessageDuration({
         label={tx('off')}
         value={String(DisappearingMessageDuration.OFF)}
       />
+      <Radio
+        key={'eph-1a'}
+        label={'After 3 seconds'}
+        value={String(DisappearingMessageDuration.THREE_SECONDS)}
+      />
```
Simon-Laux, line 56: is this useful for users? should it stay?

Simon-Laux:
sending messages is broken for me when checking out this pr

Simon-Laux (approved):
sending works again, so fine to merge even though I got a strange bug, but master's message list is buggier at this point so merging would be an improvement

PR #2562, opened 2022-02-21
Make background image url portable

### LINKED ISSUE #2169: Background url is not portable
background url saved in `config.json` is an absolute path and thus not portable.
also it might make sense to store the value some different way than as css code. maybe `img:[path]` and `color:[color css]` so that the ui could transform the path as needed.
this would need some migration, but its not much work as really only the file name is interesting as background images are all copied to the same folder in the data dir.

flipsimon:
First take on making the background image url portable. Not sure if this is the right way. Happy about reviews and suggestions!
closes #2169

### DIFF
--- a/src/renderer/components/dialogs/Settings-Appearance.tsx
+++ b/src/renderer/components/dialogs/Settings-Appearance.tsx
@@ -11,6 +11,7 @@ import { SettingsSelector } from './Settings'
 import { SmallSelectDialog, SelectDialogOption } from './DeltaDialog'
 import { runtime } from '../../runtime'
 import { RC_Config, Theme } from '../../../shared/shared-types'
+import { join } from 'path'
 
 const enum SetBackgroundAction {
   default,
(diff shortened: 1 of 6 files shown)


missytake:
Right now, if people just update to our branch, their background image is not loaded... we need to think about a migration path. But I already have an idea :)

missytake:
Okay, this PR is ready for review :)

I would not merge it before the 1.28 release, I'm not so sure if we really got *all* corner cases and the release has a lot of bug potential already... it's not so important, let's just merge it afterwards.

Does this need a changelog item? mergeable says it needs one.

Jikstra:
> Does this need a changelog item? mergeable says it needs one.

Yes, please add one :)

Simon-Laux:
I'm not too happy about giving the frontend the config path, but we'll change that stuff again soon anyway with the command api so okay for now.

src/renderer/components/dialogs/Settings-Appearance.tsx
```diff
@@ -11,6 +11,7 @@ import { SettingsSelector } from './Settings'
 import { SmallSelectDialog, SelectDialogOption } from './DeltaDialog'
 import { runtime } from '../../runtime'
 import { RC_Config, Theme } from '../../../shared/shared-types'
+import { join } from 'path'
```
Simon-Laux, line 14: don't import node dependencies in the frontend, it's not given that they work there.
its ok for join in path I think, but stuff as `fs` won't work without bad practice electron magic that weakens security.

PR #2570, opened 2022-02-26
Webxdc keep dom storage

Simon-Laux:
- [X] clear webxdc data on account removal
- [X] option in settings to clear webxdc data for account
- ~~[ ] clear webxdc data when message is deleted (and also close the webxdc window if its still open)~~ moved to its own issue: #2630
- [x] rebase this pr

related: https://github.com/deltachat/deltachat-android/pull/2221

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased][unreleased]
 
+### Added
+- webxdc content now has access to persistent DOMStorage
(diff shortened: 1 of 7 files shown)


adbenitez:
I don't think this is necessary for a 1.28 release, Android released 1.28 without this, so unless this is something you can do quickly it is better to ship 1.28 and not to block release due to this PR

PR #2575, opened 2022-03-03
upgrade to electron version 14

Simon-Laux:
this is the oldest currently supported version of electron, electron 13 is deprecated.

the next versions of electron need a newer nodejs version﻿


Jikstra:
Do we wanna upgrade electron for the next release? And only to 14 or to a newer version?

Simon-Laux:
that's the question, but I think it makes sense

PR #2576, opened 2022-03-03
Set enterKeySends to true in default config

flipsimon:


Simon-Laux (approved):

r10s:
this change easily results in messages being sent while  (a) one  just thinks one enters a linefeed or (b) one thinks one is in draft mode.

there may be different expectations for the new-line-thingie, _Alice_ may think of a Messenger (enter-sends) and _Bob_ may think of a email-client (enter-is-newline), however, the possible harm when the default is not aligned with the expectation are not equal:

- for Alice, the message would not be sent but a new line is entered. Alice has to click the "send" button to continue and maybe change a default. **nothing bad has happen.**

- for Bob, the message starting with "fuck you, boss!" is just send while he maybe expected to be in draft mode, maybe he wants to enter a newline, maye he hit enter accidentially instead of backspace. whatever, **bad things has happened.**

even if the enter-defaults-to-send is true for _all_ other messengers (i do not think so, and digging deeper, there are also lots of discussions around that):

- we advertise Delta also as a email-app, so newlines play quite a role and the situation is different from irc and other messenger-only apps.
if you think, many ppl will change the default, maybe consider making the option in the settings more prominent, we could move them up to "Communication" eg.

so, my advice would be to leave the default as it was before :)

EDIT: in background discussions, these points are widely shared, however, these answer did not make it here ;) idea is to revert the change and maybe guide ppl better to change the option.

PR #2577, opened 2022-03-03
backend strict typescript

Simon-Laux:
- enable strict types for backend
- remove `state`, `translate` and `localeData` from `ExtendedAppMainProcess`
- remove not existing function from typedef
- log unhandled errors and rejections in frontend

- [x] reference new dc-node version with https://github.com/deltachat/deltachat-node/pull/540 merged


src/main/desktop_settings.ts
```diff
@@ -0,0 +1,73 @@
...
+        saved.lastAccount = undefined
+      }
+    } catch (error) {
+      log.debug(error)
+      log.info('Missing configuration file. Using default values.')
+    }
+    this.inner_state = Object.assign(default_state, saved)
+  }
+
+  mutate(state: Partial<DesktopSettingsType>) {
```
Jikstra, line 44: Not too happy about mutate, would prefer a simple update/set.

Simon-Laux: just naming?

PR #2603, opened 2022-03-11
Always mark messages as seen if user requests them to display

Jikstra:


src/main/deltachat/messagelist.ts
```diff
@@ -198,23 +198,23 @@ export default class DCMessageList extends SplitOut {
...
-        if (
-          getDirection(message) === 'incoming' &&
-          message.state !== C.DC_STATE_IN_SEEN
-        ) {
-          markMessagesRead.push(messageId)
+        // mark messages read triggers various mechanics in the core
+        // and should always be triggered if the user displays those
+        // messages on the screen.
+        if (getDirection(message) === 'incoming') {
+          markMessagesSeen.push(messageId)
```
Simon-Laux, line 205: I don't like this logic in the backend, also could this be responsible for the new startup lag?

link2xt: It's probably fast as there is no additional database query, but even checking direction is unnecessary. It's completely ok to just call markSeenMessages for all messages displayed, including outgoing.

link2xt (approved):

Simon-Laux:
changelog entry needed?

Jikstra:
> changelog entry needed? if this fixes a bug that should be mentioned in the changelog.

Not really, doesn't change a thing an user would notice

PR #2607, opened 2022-03-11
New design for videochat instance dialog with disable, jitsi and cust…

flipsimon:
…om options

Default is still that the option is turned off. Just the design changes.

The plan is to add some cool providers of jitsi servers to the list soon. @cyBerta already wrote to a few of them and asked.
image
image
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Changed
 - exclude more unused files from installation package
+- Improved videochat instance dialog
 
 ## [1.27.1] - 2022-03-10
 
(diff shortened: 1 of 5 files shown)


Simon-Laux:
the "new design" expectations somehow set my expectations very high, but I think its already a good step.
we can improve it when we tackle improving the settings ui as a whole.
/me dislikes the current skinny and long settings dialog

flipsimon:
Yes you're right - design might be a bit far fetched :D
We had a planning session on the new settings design here in Cuba and I wanted to start working on it tomorrow. It will definitely not be that skinny and long anymore ;)

Jikstra:
Come on, we use new design for far less :D

Simon-Laux (approved):

Issue #2610, opened 2022-03-12
Warning: validateDOMNesting(...): <div> cannot appear as a descendant of <p>


Jikstra:
```
Warning: validateDOMNesting(...): <div> cannot appear as a descendant of <p>.
    at div
    at p
    at div
    at Message2 (file:///home/kerle/Coding/deltachat-desktop/html-dist/bundle.js:98145:13)
    at li
    at MessageWrapper
    at div
    at MessagePageComponent2 (file:///home/kerle/Coding/deltachat-desktop/html-dist/bundle.js:98879:5)
    at ul
    at div
    at import_react66.default.memo.areEqual2 (file:///home/kerle/Coding/deltachat-desktop/html-dist/bundle.js:98837:7)
    at MessageList (file:///home/kerle/Coding/deltachat-desktop/html-dist/bundle.js:98668:16)
    at RecoverableCrashScreen (file:///home/kerle/Coding/deltachat-desktop/html-dist/bundle.js:98966:7)
    at div
    at div
    at MessageListAndComposer (file:///home/kerle/Coding/deltachat-desktop/html-dist/bundle.js:99056:16)
    at RecoverableCrashScreen (file:///home/kerle/Coding/deltachat-desktop/html-dist/bundle.js:98966:7)
    at div
    at div
    at MainScreen (file:///home/kerle/Coding/deltachat-desktop/html-dist/bundle.js:100389:65)
    at div
    at ScreenController (file:///home/kerle/Coding/deltachat-desktop/html-dist/bundle.js:104310:7)
    at SettingsContextWrapper (file:///home/kerle/Coding/deltachat-desktop/html-dist/bundle.js:104632:37)
    at CrashScreen (file:///home/kerle/Coding/deltachat-desktop/html-dist/bundle.js:104530:7)
    at App (file:///home/kerle/Coding/deltachat-desktop/html-dist/bundle.js:104577:70)
```

PR #2611, opened 2022-03-12
Fix div is not allowed to be descendant of p

Jikstra:
Fixes https://github.com/deltachat/deltachat-desktop/issues/2610

### DIFF
--- a/scss/message/_message.scss
+++ b/scss/message/_message.scss
@@ -385,7 +385,7 @@
   text-align: center;
   margin: 26px 0px;
 
-  p {
+  .bubble {
     display: inline-block;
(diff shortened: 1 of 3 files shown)


scss/message/_message.scss
```diff
@@ -385,7 +385,7 @@
   text-align: center;
   margin: 26px 0px;
 
-  p {
+  .bubble {
```
Simon-Laux, line 388: have you made sure that `p` does not have behavior, that this class does not overwrites?

Jikstra: No but looks similiar enough

PR #2612, opened 2022-03-12
more package size reduction and list exclusions in seperate file

Simon-Laux:
- move exclude list to extra file and reorder it
- exclude even more


Simon-Laux:
just tried the appimage it still works, so can be merged

PR #2615, opened 2022-03-12
fix two different about dialogs on macOS

### LINKED ISSUE #2280: two different "about dialogs"
on mac, there are two menu entries named "About Delta Chat", each opening different dialogs.

the one is in the "Delta Chat" menu, the other in the "Help" menu:

 

Screen Shot 2021-06-02 at 16 47 31

i think, this is a bit confusing, esp. when guiding users to some debug information. also, it looks weird :)

i think, in general, it is okayish to keep both entries, however, they should open the same dialog.

i assume there is some hook in electron, handling this special "Delta Chat / About Delta Chat" entry.

Simon-Laux:
closes #2280

for now don't remove the menu item from help, because that way its easier for us to point users to it.
when it is in the same place on all 3 platforms.﻿


Jikstra (approved):

r10s:
> for now don't remove the menu item from help

yip, that is perfectly fine, having two options to do the same thing is fine.

Issue #2257, opened 2021-05-21
MacOS: open deltachat from uri doesn't detect the uri


Simon-Laux:
This affects both `mailto` and `OPENPGP4FPR` they are correctly detected/handled once deltachat is opened.

r10s:
i just tested with 1.20.2, `openpgp4fpr` opens Delta Chat whether it is already running or not, that is great!

however, once started, the secure-join-protocol is not started, it should basiccally do the same as if a qr-code with the content was scanned, so ask the user and pass the qr-code to dc_join_securejoin().

the goal is that snippets as the following work:

```
<!DOCTYPE html>
<html>
<head>
</head>
<body>

<a href="OPENPGP4FPR:61D37E30CA7C4630C66D564977FB68391D3B7184#a=tunis4%40testrun.org&n=&i=sp2gRzrr7AA&s=QrDr3ASfZGA">
   <img src="QR code here .." />.
</a>

</body>
</html>

```

r10s:
the same issue is true if Delta Chat is the "Default E-Mail App" on mac os and the uses selectes "Share to Delta Chat" eg. from Safari - if Delta Chat is not yet started, it is started, however the things to be shared are lost.

Simon-Laux:
https://github.com/electron/electron/issues/23852

Simon-Laux:
in GitHub desktop somehow works, it uses electron 17.0.1, so maybe it is fixed in electron 17?

Simon-Laux:
so found out why it did not work: macOS only uses `app.on('open-url')` for this, but the listener needs to be registered early so it can catch the event. we set the listener too late to get the url.

PR #2616, opened 2022-03-13
several fixes to open uri

Simon-Laux:
- closes https://github.com/deltachat/deltachat-desktop/issues/2257
- show qr code content if decoding failed:
Bildschirmfoto 2022-03-13 um 01 28 43

- also add a workaround for https://github.com/deltachat/deltachat-core-rust/issues/1969
until core can accept both.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -82,3 +82,4 @@ src/shared/build-info.ts
 # ignore provisioning file
 embedded.provisionprofile
 
+stub.js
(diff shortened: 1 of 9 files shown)


.gitignore
```diff
@@ -82,3 +82,4 @@ src/shared/build-info.ts
 # ignore provisioning file
 embedded.provisionprofile
 
+stub.js
```
Simon-Laux, line 85: this is a file I needed for local testing on Mac (to skip the signing step)

PR #2622, opened 2022-03-14
fix debouncing of onChatListItemChanged

Simon-Laux:
this reduces unessesary calls to `chatList.getChatListEntryMessageIdForChatId`
also don't fetch the non-existant chatlist item for the trash chat (id 3).

With this change the call count of this function went down from 669 to 27 (importing the same backup of an test account for testing)

This should already make things faster during startup, because as you can see here that function calls `getChatList`:

https://github.com/deltachat/deltachat-desktop/blob/98ddbe9f8a854b65c72ed54e96e339a93d0b3b86/src/main/deltachat/chatlist.ts#L57-L66

As you can see in the comment we should move this function to the core and decouple it from fetching the whole chat-list.
We didn't do it back then because @r10s was worried that chatlistitems content and position in the chatlist would get out of sync.

### DIFF
--- a/src/renderer/components/chat/ChatList.tsx
+++ b/src/renderer/components/chat/ChatList.tsx
@@ -401,34 +401,47 @@ export function useLogicVirtualChatList(chatListIds: [number, number][]) {
     })
   }
 
-  const onChatListItemChanged = useCallback(
-    (chatId: number, _messageId: number | string) => {
-      const debouncingChatlistItemRequests: { [chatid: number]: number } = {}
-      const updateChatListItem = async (chatId: number) => {
-        debouncingChatlistItemRequests[chatId] = 1
-        // the message id of the event could be an older message than the newest message (for example msg-read event)
-        const messageId = await DeltaBackend.call(
-          'chatList.getChatListEntryMessageIdForChatId',
-          chatId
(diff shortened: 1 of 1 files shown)


src/renderer/components/chat/ChatList.tsx
```diff
@@ -401,34 +401,47 @@ export function useLogicVirtualChatList(chatListIds: [number, number][]) {
...
-        setChatLoading(state => ({
-          ...state,
-          [chatId]: LoadStatus.LOADED,
-        }))
-        if (debouncingChatlistItemRequests[chatId] > 1) {
-          updateChatListItem(chatId)
-        } else {
-          debouncingChatlistItemRequests[chatId] = 0
-        }
+  const onChatListItemChanged = useMemo(() => {
```
Jikstra, line 404: Why useMemo over useCallback?

adbenitez (approved):

src/renderer/components/chat/ChatList.tsx
Simon-Laux: because then I can not define the stuff outside of the function, read the next lines (where `{` begin and end) then you'll understand

PR #2623, opened 2022-03-14
Overhaul three dot menu

Jikstra:
Screenshot_20220314_174734

Screenshot_20220314_172734

Screenshot_20220314_172755


r10s:
sooooooooo much better!

i would re-add "Edit Group", even if that is also possible indirectly by a tap on the title bar. it is an quite important, it was there before as well and it is good to have a "menu of all options" in the menu :)

Simon-Laux (approved):
nice now keyboard navigation in the 3dot-menu works!

Jikstra:
> sooooooooo much better!
> 
> i would re-add "Edit Group" or better the more common "View Profile", even if that is also possible indirectly by a tap on the title bar. it is an quite important item, it was there before as well and it is good to have a "menu of all options" in the menu :)

Removed it as even if there was an Edit Group, there was never an View Profile. So for 1to1 chats there never was an item to access to the profile dialog. But can add it again/add the other one too.

r10s:
> Removed it as even if there was an Edit Group, there was never an View Profile.

yip, it is good to have either both or none.

> But can add it again/add the other one too.

i would tend to add a "View Profile" for all group types on desktop. there is enough room on, the menu is still not overloaded and it makes migration easier if ppl search for sth. to "edit a group" in the three dot menu.

on android, however, we removed that option as there is far less space. also, in contrast to desktop, the title is "much closer" to the menu and ppl are much more used to tap there, these conventions are not there on desktop.

PR #2637, opened 2022-03-22
fix react memo for message page

Simon-Laux:
this fixes webxdc summary updates


Jikstra:
I like the introduced class to alter a messagePage. But the actual problem is that the messagePage reference is not getting changed in the messageChanged reducer. See this pr https://github.com/deltachat/deltachat-desktop/pull/2639

Issue #2594, opened 2022-03-09
message input lagging


r10s:
when typing in a message to the input windows, the update of the typed in characters is slow, typically the characters appear half of a second or more after typing - where it should be instant.

the issue is not only there when typing in lots of lines of text, it is also noticeable when there is very few text. it does not happen all the time, however, pretty much often (just play around ...)

the reason is probably that on every character the draft is updated.

Simon-Laux:
> when typing in a message to the input windows, the update of the typed in characters is slow, typically the characters appear half of a second or more after typing - where it should be instant.

I can not really reproduce this, in which versions does this appear?

What I noticed is that I can not keep a character pressed to spam it, but trying in text works as expected.

> the reason is probably that on every character the draft is updated.

not in 1.27.2, currently the debouncing works (500ms after I stop typing the draft is saved to core).
But I noticed that the react state changes on every character and that could trigger re-rendering:
<img width="744" alt="Bildschirmfoto 2022-03-16 um 15 04 30" src="https://user-images.githubusercontent.com/18725968/158609016-eda72fe5-7767-48e4-93c5-ce3b6eafd412.png">

Jikstra:
@hpk42 @r10s can one of you do a performance recording? You can start it like this:
1. Open the developer console
2. Go to performance tab
3. click on the record button (its a grey circle) in the upper left
4. perform what is slow
5. stop the recording
6. Save the profile with the arrow down button in the performance tab, it's in the same row as the record button
7. Send me the saved recording

r10s:
@Jikstra sent you a recording to your riseup address, not sure if that help - but the tool is impressive, at a first glance :)

hpk42:
with recent masters i don't get lagging message input.  typed quite some things, and it seems the "setDraft" debouncing is working. please re-open if you still have this issue with recent master's of core/node/desktop .... or with the upcoming desktop test release.

PR #2640, opened 2022-03-22
increase debounce timeout to one second

adbenitez:
this makes typing and long-pressing a key like backspace more smooth

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Changed
 - Set default of enterKeySends to false again
 - remove webxdc clear domstorage settings for now until we know what we want (see https://github.com/deltachat/deltachat-desktop/issues/2638)
+- increase debounce timeout to one second
 
 ## [1.27.2] - 2022-03-15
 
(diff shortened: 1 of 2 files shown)


gerryfrancis:
Might fix https://github.com/deltachat/deltachat-desktop/issues/2594 , thank you, @adbenitez !

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Changed
 - Set default of enterKeySends to false again
 - remove webxdc clear domstorage settings for now until we know what we want (see https://github.com/deltachat/deltachat-desktop/issues/2638)
+- increase debounce timeout to one second
```
Simon-Laux, line 8: which debounce timeout?
```suggestion
- increase composer draft saving debounce timeout to one second
```

adbenitez: ups! true

Jikstra (approved):
Obviously we have to fix it properly, seems like the backend blocks or something but if it already improves things im happy to merge :)

PR #2650, opened 2022-03-27
update react-string-replace to 1.0.0

dotlambda:
This turns lodash into a devDependency, removing 4.8 MiB from the package size.


dotlambda:
In case you wonder whether this is a breaking change, it's not: https://github.com/iansinnott/react-string-replace/releases

Simon-Laux (approved):

PR #2651, opened 2022-03-27
visualize "send on enter" setting + adjust `ENTER + SHIFT behavior + add inApp keybindings cheatsheet

Simon-Laux:
- [X] `ENTER + SHIFT` and `ENTER + CTRL` are now the same in the composer they send or add a newline depending on the `enterKeySends` user preference
- [X] visualize "send on enter" setting
- [x] is Mac? (hide meta key combos if not mac, if on mac show meta key combos and maybe rename alt to option)
- [X] keyboard shortcut cheat sheet dialog
- [x] BUG: cheat sheet dialog should not be able to be opened multiple times on top of each other
- [x] translations for action titles
- [x] translations for key names (Control -> Ctrl, Strg in German)? - hardcoded for now because German seems to be the only case where keys have different labels
- [ ] shortcut popover (dismiss-able: do not show again) on hover on send button

## Preview

 

Peek 2022-03-28 01-48


src/renderer/components/KeyboardShortcutHint.tsx
```diff
@@ -0,0 +1,250 @@
...
+
+const keySymbols: { [key: string]: string } = {
+  Control: '^',
+  Shift: '⇧',
+  Alt: '⌥',
+  Meta: '⌘',
+  Enter: '↵',
+  ArrowDown: '↓',
+  ArrowUp: '↑',
+}
```
Jikstra, line 13: Maybe rewrite to const KEY_SYMBOL_{CONTROL;SHIFT;ALT...} = '^' instead of a hashmap?

scss/misc/_keyboard_hint_box.scss
```diff
@@ -0,0 +1,93 @@
...
+        transform: translatey(2px);
+      }
+    }
+  }
+  .action {
+    display: block;
+    background-color: rgba(0, 0, 0, 0.05);
+    border-radius: 4px;
+    padding: 4px 6px;
+    margin: 4px 0px;
```
Jikstra, line 28: I normally try to group css into following order:
positioning & size {position, display, width, margin...}
[flex] {flex-*, justify-*}
styling {border*, background, color...}
font {font-*}

I think having this somehow consistent helps a lot with readability. But mainly wanted to write down my "system" for this.

Jikstra (approved):
Like it a lot, looks super cool. Just some small subjective style preferences, feel free to ignore :P

Talking about aesthetics, I dislike the grey background behind the keybinding widget and in the dialog ;)

PR #2652, opened 2022-03-28
fix: webxdc allow `blob:` uri scheme

Simon-Laux:


r10s:
> CSP is a whitelist-blocklist, so If we do not use it on android we don't need to modify it there.

i do not know about really existant webxdc using `blob:` scheme - if that scheme is allowed android/ios, maybe disable that scheme there?

PR #2656, opened 2022-03-31
Master

M4ss1ck:
Add pnpm lockfile (pnpm-lock.yaml) to .gitignore
Unless it is ok to commit it

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,3 +83,6 @@ src/shared/build-info.ts
 embedded.provisionprofile
 
 stub.js
+
+# pnpm lockfile
+pnpm-lock.yaml


Simon-Laux:
yeah makes sense as npm is the default for the project and so the pnpm lockfile will get out of date quickly.

Simon-Laux (approved):

.gitignore
```diff
@@ -83,3 +83,6 @@ src/shared/build-info.ts
 embedded.provisionprofile
 
 stub.js
+
+# pnpm lockfile
+pnpm-lock.yaml
```
dotlambda, line 88:
```suggestion
pnpm-lock.yaml

```

M4ss1ck:
That doesn't go here, my bad

PR #2662, opened 2022-04-02
Quote-with-sticker

### LINKED ISSUE #2658: contact name has color in quote when repying with sticker
when someone reply in a group to a message with an sticker, the quoted message is shown in a gray transparent bubble but the sender of the message is show as usual with the sender's color, in this case the text should be always white, like in Android

M4ss1ck:
As said in #2658, when replying a message with a sticker, background is transparent, so author's displayColor could get difficult to read.
That's why I'm not using it in case of stickers, but using css defaults instead
This is the behavior seen in the Android app

close #2658


Jikstra:
@M4ss1ck niice, can you rebase this so that we can get it merged? Thank youuu :)

PR #2669, opened 2022-04-05
Update electron to v16.2.1, run npm audit fix

Jikstra:


Simon-Laux:
I would rather also upgrade the project to node 16, as its the base of the newer electron versions

Simon-Laux:
let's use #2687 instead

PR #2682, opened 2022-04-08
bundle.js.map is only needed for debugging

dotlambda:
Removing another 8.3 MiB from the build. Cubans rejoice!


obsoleszenz:
The problem is, if we remove it, that the stack traces are not really helpful and we need to manually find the real line of code. That's why we currently ship it.

Simon-Laux:
yeah the hassle in error cases is not with the space saving IMO

PR #2687, opened 2022-04-10
upgrade nodejs and electron 2022 q1

Simon-Laux:
only annoying thing is that `npm i` now requiews the `--force` argument, because the peer dependency warnings of blueprintjs and co. are now errors instead of warnings.

See also https://github.com/deltachat/deltachat-node/pull/542

### Changes

- update to nodejs 16 update testcafe version in action to match the one from our package.json
- update changelog
- update package-lock again run `npm i` with `--force` otherwise it does not work
- update electron to `18.0.3`
- upgrade dc-node to `1.77.1` (dc-node now compiles for node 16)


Simon-Laux:
locally electron builder works, so I wonder whats up with the ci preview builds

Simon-Laux:
seems it was this issue https://github.com/electron-userland/electron-builder/issues/6411

.github/workflows/test.yml
```diff
@@ -33,7 +33,7 @@ jobs:
         npm run test-unit
     - uses: DevExpress/testcafe-action@latest
       with:
-        version: "1.14.0"
+        version: "1.15.0"
```
Jikstra, line 36: This is about the testcafe version?

Jikstra (approved):

PR #2689, opened 2022-04-10
update emoji-mart

Simon-Laux:


CHANGELOG.md
```diff
@@ -18,6 +18,8 @@
 ### Changed
 - Update translations (25.03.2022)
 - Update react-string-replace to `1.0.0`
+- Update emoji-mart to `^3.0.1`
+- Update @types/emoji-mart to `^3.0.9`
```
dotlambda, line 22: should be under unreleased

dotlambda (changes requested):

CHANGELOG.md
Simon-Laux: good catch!

PR #2692, opened 2022-04-11
replace moment with dayjs

dotlambda:
Moment.js is [deprecated](https://momentjs.com/docs/#/-project-status/) and Day.js is [much smaller](https://github.com/you-dont-need/You-Dont-Need-Momentjs/blob/master/README.md#brief-comparison).


src/renderer/App.tsx
```diff
@@ -66,7 +66,8 @@ export default function App(_props: any) {
     window.localeData = localeData
     window.static_translate = translate(localeData.messages)
     setLocaleData(localeData)
-    moment.locale(localeData.locale)
+    require(`dayjs/locale/${localeData.locale}`)
```
Simon-Laux, line 69: I don't like the dynamic require here, I doubt that it will work as intended. this is frontend code, not nodejs backend code.

maybe we could copy the locales over to html-dist in a build step and import them in a different way?

dotlambda: How about we import all locales that we have translations for at the top?

Simon-Laux: > How about we import all locales that we have translations for at the top?

probably the simplest solution

dotlambda: > Or something like this? https://stackoverflow.com/a/66060403

I tried this and it seems to work. Can you have a look?

Simon-Laux: code validation complains, also still requires copying the files as you don't want dayjs left in `node_modules` in your package, right?

So I'd say either bundle all languages, or try to find out why the code checks complain and try to fix them and copy the locale files to html-dist or somehow instruct esbuild to put them there, if thats possible

dotlambda: The locales are now copied to bundle.js, that shouldn't be a problem.

dotlambda: But I have no idea what to do about it complaining about dynamic imports. It sure works when I run `npm run start`.

Simon-Laux: dynamic imports are imports at runtime, so its imported at runtime not bundled, that's why these are async functions with callbacks

dotlambda: > so its imported at runtime not bundled

but it's still added to `html-dist/bundle.js`

Simon-Laux: > Without code splitting enabled, an import() expression becomes Promise.resolve().then(() => require()) instead. This still preserves the asynchronous semantics of the expression but it means the imported code is included in the same bundle instead of being split off into a separate file.

seems your right, esbuild handles it: https://esbuild.github.io/api/#splitting

Jikstra:
I think for momentjs we injected our own locales? I think I dislike depending on dayjs for this.

dotlambda:
> I think for momentjs we injected our own locales?

You translate "today" and a few other things, but that's not injected into Moment.js but done manually. And I didn't change that functionality.

> I think I dislike depending on dayjs for this.

Why?

Simon-Laux:
the translation logic comes from momentjs, dayjs has the same feature only that dayjs does not bundle all languages by default to save on size

Simon-Laux:
Bildschirmfoto 2022-04-19 um 15 34 06

So first of all days is a regression in some languages, see above ⬆️ (left momentjs, right dayjs)

second this does not encompass all languages we use we need a script for that, see #2703

PR #2693, opened 2022-04-11
node-fetch is a devDependency

dotlambda:


package.json
```diff
@@ -120,7 +119,6 @@
     "@types/mime-types": "^2.1.0",
     "@types/mocha": "^9.0.0",
     "@types/node": "^14.14.20",
-    "@types/node-fetch": "^2.5.7",
```
Simon-Laux, line 123: I think we still need this when we convert the remaining test scripts to ts

Simon-Laux (approved):

PR #2695, opened 2022-04-15
fix: Signature field is not labeled when empty

### LINKED ISSUE #2579: Edit Profile: Signature field is not labeled when empty
per default you can not know what the field is about:

only on focus you see what the field is about:

or if it contains a value:

M4ss1ck:
In Edit Profile: Signature field was not labeled when empty. Now it shows a placeholder (like username field does).
This should close #2579


Jikstra:
Graciiiias! :)

PR #2705, opened 2022-04-19
Fix sidebar colors in various themes

flipsimon:
A few new css variables for sidebar colors. Left light theme as it was before. Based on this model I chose the background color of the account section of the sidebar in each theme to be a little lighter than the navigation bar.

image


themes/minimal.scss
```diff
@@ -14,17 +14,21 @@ $navbarBgLight: $borderColor;
 $outlineColor: #b9b9b9;
```
flipsimon, line 14: added .scss file ending. was it missing on purpose?

flipsimon, line 14: added .scss file ending. was it missing on purpose?

Simon-Laux: don't know, is there a reason behind it @Jikstra?

Simon-Laux (approved):
code looks perfect

themes/minimal.scss
Jikstra: Nope dont think so

PR #2708, opened 2022-04-20
Fix e2e tests settings

flipsimon:
e2e tests are working on ubuntu now at least

### DIFF
--- a/scss/main_screen/_navbar_wrapper.scss
+++ b/scss/main_screen/_navbar_wrapper.scss
@@ -83,7 +83,7 @@
     font-size: 18px;
   }
 
-  #main-menu-button {
+  #three-dot-menu-button {
     width: 40px;
     height: 40px;
     border-radius: 100%;
--- a/src/renderer/components/attachment/menuAttachment.tsx
+++ b/src/renderer/components/attachment/menuAttachment.tsx
@@ -127,7 +127,7 @@ const MenuAttachment = ({
         }
         position={Position.TOP_LEFT}
       >
-        <Button minimal icon='paperclip' />
+        <Button id='test-attachment-menu' minimal icon='paperclip' />
       </Popover>
     </div>
   )
--- a/src/renderer/components/screens/MainScreen.tsx
+++ b/src/renderer/components/screens/MainScreen.tsx
@@ -171,6 +171,7 @@ export default function MainScreen() {
             <div
               className='sidebar-icon'
               onClick={() => setSidebarState('visible')}
+              id='hamburger-menu-button'
             >
               <Icon icon='menu' aria-label={tx('main_menu')} iconSize={20} />
             </div>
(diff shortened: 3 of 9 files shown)


test/testcafe/messagelist_tests.ts
```diff
@@ -187,7 +189,9 @@ function endCode(msg) {
 function MSGinViewportSelector(timeout = 2000): Selector {
   return Selector(
     () => {
-      let raw_labels = document.querySelectorAll('#message-list .msg-body')
+      let raw_labels = document.querySelectorAll(
+        '#message-list .msg-body, .join-button'
+      )
```
Simon-Laux, line 194: we probably need something more advanced here as this can currently only detect video messages and does not necessarily select the last/most recent video message.

also does it work as expected? means that it only passes if the element is visible to the user without scrolling.

Simon-Laux: https://testcafe.io/documentation/402829/guides/basic-guides/select-page-elements could help here, its the documentation about testcafe selectors

scss/main_screen/_navbar_wrapper.scss
```diff
@@ -83,7 +83,7 @@
     font-size: 18px;
   }
 
-  #main-menu-button {
+  #three-dot-menu-button {
```
Jikstra, line 86: can we leave it as three-dot menu? Because it's not really a main menu ^^

src/renderer/components/attachment/menuAttachment.tsx
```diff
@@ -127,7 +127,7 @@ const MenuAttachment = ({
         }
         position={Position.TOP_LEFT}
       >
-        <Button minimal icon='paperclip' />
+        <Button id='test-attachment-menu' minimal icon='paperclip' />
```
Jikstra, line 130: ?

src/renderer/components/screens/MainScreen.tsx
```diff
@@ -240,54 +241,55 @@ export default function MainScreen() {
               </NavbarHeading>
             )}
             {selectedChat.chat && (
-              <span className='views'>
```
Jikstra, line 243: Why are these changes needed?

test/testcafe/messagelist_tests.ts
```diff
@@ -146,10 +149,31 @@ const clickChatByName = async (t, name) => {
...
+  await clickSideBarItem(t, await translate('menu_settings'))
+  await t.expect(Selector('.SettingsIconButton').withText(label).exists).ok()
+  await t.click(Selector('.SettingsIconButton').withText(label))
+}
+
+export async function sendVideoChatInvitation(t) {
+  await t.click('#test-attachment-menu')
+  await t.click(
+    Selector('a.bp3-menu-item').withText(await translate('videochat'))
+  )
```
Jikstra, line 176: You can put functions that you are reusing to helpers.js :)

Jikstra (commented):
Some queeestioooons

scss/main_screen/_navbar_wrapper.scss
Simon-Laux: ?

src/renderer/components/attachment/menuAttachment.tsx
Simon-Laux: without an id the test script can not easily find it

src/renderer/components/screens/MainScreen.tsx
Simon-Laux: to hide the chat context menu when no chat is selected, basically to fix a bug

test/testcafe/messagelist_tests.ts
Simon-Laux: maybe would make some sense to convert all tests to TS in that case?

Simon-Laux:
there are some message list tests regarding video chat that should fail, because there is a bug in desktop and the test should detect that if it is working correctly

src/renderer/components/attachment/menuAttachment.tsx
Jikstra:
```suggestion
        <Button id='attachment-menu' minimal icon='paperclip' />
```

test/testcafe/messagelist_tests.ts
Jikstra: > there are some message list tests regarding video chat that should fail, because there is a bug in desktop and the test should detect that if it is working correctly

Lets keep the needed changes for this pr simple so we can merge soon?

Jikstra:
> there are some message list tests regarding video chat that should fail, because there is a bug in desktop and the test should detect that if it is working correctly

Lets keep the needed changes for this pr simple so we can merge soon?

src/renderer/components/attachment/menuAttachment.tsx
Simon-Laux: the idea of the `test-` prefix was to mark it as only used in tests, some other id's are used also by desktop

Simon-Laux:
> Lets keep the needed changes for this pr simple so we can merge soon?

what do the tests matter if they are green, but broken? the goal of this pr was/is to fix the tests

test/testcafe/messagelist_tests.ts
flipsimon: As it is only for the test, we can decide that we won't have multiple videochat invitations in the chat. So in my opinion, it doesn't really matter if the selector theoretically selects all video chat invitations. 
For me it worked as expected. When the videochat invitation was not in sight for the user, the test failed, when I stopped and scrolled down, it passed.

flipsimon:
Added messagelist tests to the workflow. Now, on ubuntu, the e2e tests work as expected: all pass except for the videochat invitation scroll down test.

PR #2711, opened 2022-04-23
feat jump to message from gallery

### LINKED ISSUE #2592: "Send Message" does not always open chat view
in the contact-profile there is a "Send Message" button that usually opens the chat of the contact and allows sending a message to that contact.

however, this does not work when the "Gallery" was opened before:

- open a one-to-one chat
- open the gallery in that one-to-one chat
- tap on the title bar to open the contact's profile
- tap "Send Message"

-> the chat is not opened, no message can be send. instead, the windows stays in "Gallery".

probably easy to fix. the same is true for the "Map View", btw.

tested with 1.27.0 on mac.

### LINKED ISSUE #2618: Jump to message in gallery
when right clicking an image there should be an option to jump to the message.

### LINKED ISSUE #2619: Jump to message does not switch the chatview to message list
if it is in map or gallery view.
even when jumping to messages from different chats.

related to: #2592

### LINKED ISSUE #2698: cannot get to the context of images and documents (from gallery)
if you see an image or a document in the gallery, one may want to see the image in its context - esp. if you use the gallery as a search-helper, eg. if you are searching for some text nearby an image you can remember.

android and ios offers a "Show in Chat" function for that, we should offer something like that also for desktop, esp. as we now can now jump-to-messages, this seems to be not so hard as before :)

- add the option "Show in Chat" (string `show_in_chat`) to the context-menu of gallery and document items
- also add the option to the context menu of enlarged images
- on selecting that option, the chat should be opened scrolled to the given message shortly flashing (as already known from jump-to-quote)

Simon-Laux:
- move view to chatstore & jump to msg in gallery fix #2592 closes #2619 closes #2618 closes #2698
- add show in chat to fullscreenmedia view

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,13 @@
 ## [Unreleased][unreleased]
 
 ### Fixed
+- Fix "Send Message" does not always open chat view #2592
 - Fix contact name has color in quote when replying with sticker
 - Fix startup crash when spam clicking on app icon on mac.
 
 ### Added
 - Implement expandable settings
+- jump to message from gallery ("show in chat") #2618
 
 ### Changed
 - Move node-fetch to devDependencies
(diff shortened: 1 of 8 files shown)


flipsimon:
Looks fine to me and everything works as expected!

Jikstra:
Looks very good :)

PR #2713, opened 2022-04-25
Changed metadata's text color to white for media messages

### LINKED ISSUE #2666: date label's text color in image messages is hard to read
image

would be better if the label's text color is white like on Android

flipsimon:
#close #2666


Simon-Laux (approved):

Simon-Laux:
maybe small changelog entry under changed?

PR #2714, opened 2022-04-25
Added systemli and Autistici as videochat providers

flipsimon:
Very basic right now without logos or anything. But maybe nice to add now and polish later?
image


Jikstra:
Maybe we should add an url to the homepage?

flipsimon:
> Maybe we should add an url to the homepage? But I'm happy to merge it already like this

Good idea! But do you mean to add an url to https://www.systemli.org/ to further promote their services or to https://meet.systemli.org/$ROOM to disclose to the user what is happening when they click the radio button?

Jikstra (approved):

PR #2715, opened 2022-04-26
increase timeout in ci to test if it makes a difference

Simon-Laux:


Simon-Laux:
does not work too, maybe there are more serious issues on windows that we should test?

PR #2719, opened 2022-04-27
Fixed keybinding keyboard font color as black for all themes

### LINKED ISSUE #2716: Wrong font color in dark themes for keybinding hints
image
image

flipsimon:
fixes #2716
Should I even open a pr for such small changes? Or do we consider pushing on master in such cases?


Simon-Laux:
using pr's is better, multiple people pushing to same branch leads to errors and mistakes, so it gets annoying fast

Simon-Laux:
you should also like the issue it fixes: #2716
also I think it would make sense to have all 3-4 collars in css vars we can adjust in the theme, because the button background and bottom border don't look that good for dark mode, too.

Jikstra:
> you should also link the issue it fixes: #2716 also I think it would make sense to have all 3-4 collars in css vars we can adjust in the theme, because the button background and bottom border don't look that good for dark mode, too.

Yup having css variables would be cool :)

Jikstra (approved):

PR #2720, opened 2022-04-28
webxdc in gallery and some gallery fixes

### LINKED ISSUE #2626: unexpected click behavior in gallery/document view
clicking eg. a PDF when in gallery/document-view, the experience is different from clicking the same PDF in the chat:

- in the chat, the PDF viewer is opened, a right-click offers an option to export the PDF - everyhing's okay here :)
- in  gallery/document view, there is an unexpected and distinction between icon-image and icon-title:  the PDF viewer only opened when clicking the icon-image - when clicking the icon-title export is done.

this is totally unexpected - first, i thought, opening the viewer is not supported from the gallery :) then i thought there is a random bug :D

we should change this so that the primary action (opening the viewer) is not related to the position the user clicks - this is also what most other UIs are doing here, including eg. windows explorer, mac finder and Delta Chat in the chat view. 
 
a right-click can offer export and other things then. that also matches better to the gallery/image view.

### LINKED ISSUE #2696: missing .xdc files in Documents tab
Hi!

When looking for the multimedia in a chat, `.xdc` files are not in Documents tab (same for the other tabs as well, although it wouldn't make sense)

Disclaimer: I make sure that there were `.xdc` files in that chat

Simon-Laux:
- add Webxdc tab to gallery closes #2696 (missing .xdc files in Documents tab)
- fix unexpected click behavior in gallery/document view closes #2626
- fix prev/next media had seperate list for images and gifs
- fixed height for webxdc icons in messages
- better empty gallery tab messages
- fix eslint complaining
- fix formatting

### DIFF
--- a/scss/gallery/_media-attachment.scss
+++ b/scss/gallery/_media-attachment.scss
@@ -146,3 +146,52 @@
     }
   }
 }
+
+.media-attachment-webxdc {
+  display: flex;
+  flex-direction: row;
+  width: 300px;
+  margin: 5px 3px;
+
+  &:hover {
+    background-color: var(--galleryWebxdcItem);
+  }
+
+  & > .icon {
+    height: 44px;
+    width: 44px;
+  }
+
+  & > .text-part {
+    flex-grow: 1;
+    margin-left: 8px;
+    // The width of the icon plus our 8px margin
+    max-width: calc(100% - 37px);
+
+    & > .name {
+      color: var(--messageAttachmentFileInfo);
+      font-size: 14px;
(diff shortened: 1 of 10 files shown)


scss/gallery/_media-attachment.scss
```diff
@@ -146,3 +146,52 @@
     }
```
flipsimon, line 146: Non-blocking: for documents: would be nice if the cursor became a pointer when hovering over the title to make clear that it is clickable as well.
Analogically for webxdc content, pointer would be nice.

PR #2733, opened 2022-05-05
allow internal webxdc navigation and handle mailto clicking in webxdc's in DC

Simon-Laux:


r10s:
great! with https://github.com/deltachat/deltachat-android/issues/2279 internal links work on all platform now, this is an important feature. <3

PR #2734, opened 2022-05-05
Prepare for v1.29.0

Jikstra:


Simon-Laux (changes requested):
shouldn't we do an 1.29 ('cause its an testrelease?) first and bump it on release to 1.30?

package-lock.json
```diff
@@ -1,12 +1,12 @@
...
   "name": "deltachat-desktop",
-  "version": "1.30.0",
+  "version": "1.29.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "deltachat-desktop",
-      "version": "1.28.2",
+      "version": "1.29.2",
```
dotlambda, line 9:
```suggestion
      "version": "1.29.0",
```

dotlambda (changes requested):

package-lock.json
dotlambda: This is still wrong

PR #2748, opened 2022-05-10
Sidebar now closes on escape key. No focusing on sidebar menu items yet

flipsimon:


Simon-Laux (approved):
Ok for now, though we need to think about how to streamline keybindings in general.

PR #2750, opened 2022-05-10
WIP Jumping to first unread message

flipsimon:
but still marking everything below as read


missytake:
nice! I've been waiting for this for a long time... directly installed the branch :3

PR #2751, opened 2022-05-11
prevent webxdc content from setting the window title

Simon-Laux:


r10s (approved):
lgtm, as far as i can tell :)

PR #2756, opened 2022-05-14
Webxdc window title: show Webxdc name first

r10s:
this gives more focus to the app name and looks more natural.
if we add an additional document name,
that will be prepended or will replace the Webxdc Name.

cmp. deltachat/deltachat-core-rust#3317 and https://github.com/deltachat/deltachat-core-rust/issues/3316


Simon-Laux (approved):

r10s:
thanks for approving, please check if the failing tests are related to this pr and merge as needed :) i do not really have an idea about that aspect.

Simon-Laux:
I think they are just broken also on master, its another issue

PR #2760, opened 2022-05-17
add debug logging switch

Simon-Laux:
- fixing imports after moving node to core repo
- use git version of deltachat node for now
- add switch to enable experimental core logging

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "application-config": "^1.0.1",
     "classnames": "^2.3.1",
     "debounce": "^1.2.0",
-    "deltachat-node": "1.83.0",
+    "deltachat-node": "github:deltachat/deltachat-core-rust#hoc/core-logging-2",
     "emoji-js-clean": "^4.0.0",
     "emoji-mart": "^3.0.1",
     "emoji-regex": "^9.2.2",
--- a/src/renderer/components/dialogs/Settings-ExperimentalFeatures.tsx
+++ b/src/renderer/components/dialogs/Settings-ExperimentalFeatures.tsx
@@ -121,8 +121,8 @@ function DebugWebxdcSwitch(props: {}) {
   return (
     <div>
       <DeltaSwitch2
-        label={tx('debug_logging')}
-        description={tx('debug_logging_description')}
+        label={tx('pref_debug_logging')}
+        description={tx('pref_debug_logging_explain')}
         value={isActive}
         onClick={() => {
           DeltaBackend.call(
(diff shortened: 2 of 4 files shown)


src/renderer/components/dialogs/Settings-ExperimentalFeatures.tsx
```diff
@@ -87,10 +96,56 @@ export function SettingsExperimentalFeatures({
...
+    refresh()
+  }, [refresh])
+
+  console.log({ debugMsgId })
+  const isActive = !isNaN(debugMsgId) && debugMsgId > 0
+  return (
+    <div>
+      <DeltaSwitch2
+        label={tx('debug_logging')}
+        description={tx('debug_logging_description')}
```
Hocuri, line 125: Is `tx` for translating? Because, the names are probably going to be pref_debug_logging and pref_debug_logging_explain for consistency with other preferences.

See https://github.com/deltachat/deltachat-android/pull/2292/files#diff-ac392d3094c70ee4fda24e1a3a80f912c9f14d01fbf4ed7a774b8ba09c301188

Hocuri (commented):
Nice, thanks! (No need to merge this yet I think, the core PR isn't merged either and @Septias can just use this branch)

src/renderer/components/dialogs/Settings-ExperimentalFeatures.tsx
Simon-Laux: yes tx is for translation, If you want you can change those strings too or comment a suggestion

Hocuri:
```suggestion
        label={tx('pref_debug_logging')}
        description={tx('pref_debug_logging_explain')}
```

package.json
```diff
@@ -83,7 +83,7 @@
     "application-config": "^1.0.1",
     "classnames": "^2.3.1",
     "debounce": "^1.2.0",
-    "deltachat-node": "1.79.3",
```
Septias, line 86: shouldn't this be a relative path if node is moved into the repo?

Septias (commented):
looks fine to me - don't value my opinion too much yet tho

hpk42:
no, also android and ios don't have it in their 1.30 releases.

package.json
Simon-Laux: ? not sure what you mean here, node was moved into the core repo not into the desktop repo

PR #2771, opened 2022-05-20
fix glob-watcher usage

dotlambda:
It still doesn't work in `bin/build-shared-convert-translations.js` and I have no idea why, but it's better than the status quo.

### DIFF
--- a/bin/build-shared-convert-translations.js
+++ b/bin/build-shared-convert-translations.js
@@ -1,7 +1,7 @@
 const converter = require('xml-js')
 const path = require('path')
 const fs = require('fs')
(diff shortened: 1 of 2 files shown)


Jikstra:
chokidar is a good library for file watching, what exactly is not working @dotlambda ?

dotlambda:
> chokidar is a good library for file watching

Should I replace glob-watcher with it?

> what exactly is not working @dotlambda ?

The callback isn't called when translations are changed. For static it works perfectly though.

Jikstra:
@dotlambda you still want to work on this?

PR #2778, opened 2022-05-26
Added last seen to profile view

### LINKED ISSUE #2573: add Contact Profile View: last seen
sometime ago core added a last seen property to contacts, android displays that information in the profile screen of a contact.
how it looks on android:
image_2022-02-26_14-54-59

the android pr: https://github.com/deltachat/deltachat-android/pull/2144

canastillo:
For enhancement #2573 (closes #2573)

I'm having a problem with the timestamps. I'm getting dates from 1970. 

This is how it looks at the moment
image

Any suggestions?

### DIFF
--- a/src/renderer/components/dialogs/ViewProfile.tsx
+++ b/src/renderer/components/dialogs/ViewProfile.tsx
@@ -45,8 +45,7 @@ const ProfileInfoName = ({
   if (lastSeen == 0) {
     lastSeenString = tx('last_seen_unknown')
   } else {
-    const extended = false
-    const date = formatRelativeTime(lastSeen * 1000, { extended })
+    const date = formatRelativeTime(lastSeen * 1000, { extended: false })
     lastSeenString = tx('last_seen_at', date)
   }
 
--- a/src/renderer/components/dialogs/ViewProfile.tsx
+++ b/src/renderer/components/dialogs/ViewProfile.tsx
@@ -1,4 +1,5 @@
 import React, { useState, useContext } from 'react'
+import formatRelativeTime from '../conversations/formatRelativeTime'
 import {
   DeltaDialogBase,
   DeltaDialogHeader,
@@ -30,16 +31,25 @@ import { selectChat } from '../helpers/ChatMethods'
 const ProfileInfoName = ({
   name,
   address,
+  lastSeen,
 }: {
   name: string
   address: string
+  lastSeen: number
 }) => {
+  const tx = useTranslationFunction()
+  const extended = false
+  const time = formatRelativeTime(lastSeen, { extended })
+
   return (
     <div className='profile-info-name-container'>
       <div>
         <p className='group-name'>{name}</p>
       </div>
       <div className='address'>{address}</div>
+      <div>
+        <p>{`${tx('last_seen_at', `${time}`)}`}</p>
+      </div>
     </div>
   )
 }
@@ -142,6 +152,7 @@ export function ViewProfileInner({
               address={
                 isDeviceMessage ? tx('device_talk_subtitle') : contact.address
               }
+              lastSeen={contact.lastSeen}
             />
           </div>
           <div
--- a/src/renderer/components/dialogs/ViewProfile.tsx
+++ b/src/renderer/components/dialogs/ViewProfile.tsx
@@ -1,4 +1,5 @@
 import React, { useState, useContext } from 'react'
+import moment from 'moment'
 import formatRelativeTime from '../conversations/formatRelativeTime'
 import {
   DeltaDialogBase,
@@ -38,8 +39,16 @@ const ProfileInfoName = ({
   lastSeen: number
 }) => {
   const tx = useTranslationFunction()
-  const extended = false
-  const time = formatRelativeTime(lastSeen, { extended })
+  const year = moment(lastSeen).format('YYYY')
+  let lastSeenString = ''
+
+  // Dates from 1970 mean that contact has never been seen
+  if (parseInt(year) <= 1970) lastSeenString = tx('last_seen_unknown')
+  else {
+    const extended = false
+    const date = formatRelativeTime(lastSeen, { extended })
+    lastSeenString = tx('last_seen_at', `${date}`)
+  }
 
   return (
     <div className='profile-info-name-container'>
@@ -48,7 +57,7 @@ const ProfileInfoName = ({
       </div>
       <div className='address'>{address}</div>
       <div>
-        <p>{`${tx('last_seen_at', `${time}`)}`}</p>
+        <p>{`${lastSeenString}`}</p>
       </div>
     </div>
   )
--- a/src/renderer/components/dialogs/ViewProfile.tsx
+++ b/src/renderer/components/dialogs/ViewProfile.tsx
@@ -39,15 +39,15 @@ const ProfileInfoName = ({
   lastSeen: number
 }) => {
   const tx = useTranslationFunction()
-  const year = moment(lastSeen).format('YYYY')
   let lastSeenString = ''
 
   // Dates from 1970 mean that contact has never been seen
-  if (parseInt(year) <= 1970) lastSeenString = tx('last_seen_unknown')
-  else {
+  if (lastSeen == 0) {
+    lastSeenString = tx('last_seen_unknown')
+  } else {
     const extended = false
-    const date = formatRelativeTime(lastSeen, { extended })
-    lastSeenString = tx('last_seen_at', `${date}`)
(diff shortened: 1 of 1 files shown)


Simon-Laux:
> Any suggestions?

Screenshot shows 1970 date, better hide the text or say "long time ago" / "unknown" if timestamp is equal to `0`

Also we could discuss the location, like android does it in a section bellow the name, similar to the signature, but I don't have a strong opinion on that, maybe its fine as is (directly under the name)

adbenitez:
> I'm having a problem with the timestamps. I'm getting dates from 1970.

hola Ana, gracias por contribuir,

dates from 1970 mean that you have never seen that contact in that case `Last seen: Unknown` must be shown (`last_seen_unknown` string)

adbenitez:
oops! @Simon-Laux  already replied

canastillo:
Thanks for your comments. I have made some changes for that case with the 70s. I wrote the condition as `<= 1970` because I remember seing even a date from 1969. 

I just have two questions: 
- Should I consider a later year or is 1970 okay?
- When does this "unknown" case happens? Because I chatted with the person of the screenshot like two days ago and I chatted today with another one and the same thing happens. Last seen has dates from 1970.

About the location, I'm open to suggestions. I thought that spot was the right place but I can change it. No problem.

Jikstra:
> Thanks for your comments. I have made some changes for that case with the 70s. I wrote the condition as `<= 1970` because I remember seing even a date from 1969 (Dec 31st probably). I didn't write anything about `lastSeen == 0` 'cause I'm actually not getting 0s from the backend, but small numbers like 1653452517 or 1653437084.
> 
> I just have two questions:
> 
> * Should I consider a later year or is 1970 okay?
> * When does this "unknown" case happens? Because I chatted with the person of the screenshot like two days ago and I chatted today with another one and the same thing happens. Last seen has dates from 1970.
> 
> About the location, I'm open to suggestions. I thought that spot was the right place but I can change it. No problem.

I have the feeling it's something else, @adbenitez in theory 0 should be returned if there's no seen info right?

src/renderer/components/dialogs/ViewProfile.tsx
```diff
@@ -30,16 +32,33 @@ import { selectChat } from '../helpers/ChatMethods'
...
+
   return (
     <div className='profile-info-name-container'>
       <div>
         <p className='group-name'>{name}</p>
       </div>
       <div className='address'>{address}</div>
+      <div>
+        <p>{`${lastSeenString}`}</p>
+      </div>
```
Jikstra, line 61:
```suggestion
      <div className='last-seen'>{lastSeenString}`}</div>
```

src/renderer/components/dialogs/ViewProfile.tsx
```diff
@@ -30,16 +32,33 @@ import { selectChat } from '../helpers/ChatMethods'
...
+  const tx = useTranslationFunction()
+  const year = moment(lastSeen).format('YYYY')
+  let lastSeenString = ''
+
+  // Dates from 1970 mean that contact has never been seen
+  if (parseInt(year) <= 1970) lastSeenString = tx('last_seen_unknown')
+  else {
+    const extended = false
+    const date = formatRelativeTime(lastSeen, { extended })
+    lastSeenString = tx('last_seen_at', `${date}`)
```
Jikstra, line 50:
```suggestion
    lastSeenString = tx('last_seen_at', date)
```

src/renderer/components/dialogs/ViewProfile.tsx
```diff
@@ -30,16 +32,33 @@ import { selectChat } from '../helpers/ChatMethods'
...
   address: string
+  lastSeen: number
 }) => {
+  const tx = useTranslationFunction()
+  const year = moment(lastSeen).format('YYYY')
+  let lastSeenString = ''
+
+  // Dates from 1970 mean that contact has never been seen
+  if (parseInt(year) <= 1970) lastSeenString = tx('last_seen_unknown')
+  else {
```
Jikstra, line 47:
```suggestion
  if (parseInt(year) <= 1970) {
    lastSeenString = tx('last_seen_unknown')
  } else {
```

src/renderer/components/dialogs/ViewProfile.tsx
```diff
@@ -30,16 +32,33 @@ import { selectChat } from '../helpers/ChatMethods'
...
   name: string
   address: string
+  lastSeen: number
 }) => {
+  const tx = useTranslationFunction()
+  const year = moment(lastSeen).format('YYYY')
+  let lastSeenString = ''
+
+  // Dates from 1970 mean that contact has never been seen
+  if (parseInt(year) <= 1970) lastSeenString = tx('last_seen_unknown')
```
Jikstra, line 46:
```
  if (parseInt(year) <= 1970) {
    lastSeenString = tx('last_seen_unknown')
  } else {
```

Jikstra (commented):
Added some small code changes :)

Simon-Laux:
It turns out the timestamp from the core is in seconds, while the timestamp expected by momenjs is in milliseconds, so you need to do `lastSeen * 1000`:
const date = formatRelativeTime(lastSeen * 1000, { extended })

then you can also just compare to `0` instead of parsing the year. we also compare timestamps to `0` in other places, recently I added it for the chatlistitems of chats that have no messages yet.

After fixing it locally I noticed this:
image

which makes not much sense to me, so I think we should use another relative time format, so that it says "Monday" instead of just "Mon". Or we opt for just the absolute date/time? (has is it done on android @adbenitez ?)

src/renderer/components/dialogs/ViewProfile.tsx
canastillo: Can you please write this as a suggestion so I can add it to the batch? The button doesn't appear in this one

adbenitez:
> I have the feeling it's something else, @adbenitez in theory 0 should be returned if there's no seen info right?

yes, and that is what android checks for: 
https://github.com/deltachat/deltachat-android/blob/master/src/org/thoughtcrime/securesms/ProfileSettingsAdapter.java#L240

adbenitez:
> which makes not much sense to me, so I think we should use another relative time format, so that it says "Monday" instead of just "Mon". Or we opt for just the absolute date/time? (has is it done on android @adbenitez ?)

for how it is done in android check:
https://github.com/deltachat/deltachat-android/blob/b488103c43fdd5c7dfa059cc1c6e33c1f0b813dd/src/org/thoughtcrime/securesms/util/DateUtils.java#L80

```
  public static String getExtendedTimeSpanString(final Context c, final Locale locale, final long timestamp) {
    StringBuilder format = new StringBuilder();
    if      (DateUtils.isToday(timestamp))                 {}
    else if (isWithin(timestamp,   6, TimeUnit.DAYS)) format.append("EEE ");
    else if (isWithin(timestamp, 365, TimeUnit.DAYS)) format.append("MMM d, ");
    else                                              format.append("MMM d, yyyy, ");

    if (DateFormat.is24HourFormat(c)) format.append("HH:mm");
    else                              format.append("hh:mm a");

    return getFormattedDateTime(timestamp, format.toString(), locale);
  }
```

src/renderer/components/dialogs/ViewProfile.tsx
```diff
@@ -30,16 +32,31 @@ import { selectChat } from '../helpers/ChatMethods'
...
 }) => {
+  const tx = useTranslationFunction()
+  let lastSeenString = ''
+
+  // Dates from 1970 mean that contact has never been seen
+  if (lastSeen == 0) {
+    lastSeenString = tx('last_seen_unknown')
+  } else {
+    const extended = false
+    const date = formatRelativeTime(lastSeen * 1000, { extended })
```
Jikstra, line 49:
```suggestion
    const date = formatRelativeTime(lastSeen * 1000, { extended: false })
```

canastillo:
Thanks @Simon-Laux for the explanation. I now understand what is going on. I changed that conditional and added the changes that Jikstra made. 

So, do you want me to add another relative format? I think on android they use the format that is used already in here.

src/renderer/components/dialogs/ViewProfile.tsx
Jikstra: One last small code nitpick and then I'm happy to merge :P

Jikstra:
> So, do you want me to add another relative format?

I wouldn't do it in this pr, but do we need another relative format? What was the discussion on this? Probably I missed it :P

src/renderer/components/dialogs/ViewProfile.tsx
canastillo: Done! :D

Issue #2627, opened 2022-03-15
too wide clickable area on forwarded messages


r10s:
having a forwarded message, the title shows "Forwarded Message" instead of the sender's name, which is fine.

however, while the clickable area is usually only as wide as the text is (eg. if the name is displayed), in case for "Forwarded Message", the clickable area is as wide as the bubble, which raises questions if eg. the cursor moves to a "pointing hand" in the right area of the bubble:

<img width=640 src=https://user-images.githubusercontent.com/9800740/158412643-d098a1d2-6e34-4272-ae1f-c1010f2560ab.png>

the effect is larger if the bubbles have some content  and some width therefore.

maybe nitpicking, however, i came over that several times ...

andresmc98:
Hi, is someone working on this issue?
i'd like to take it

r10s:
not totally sure, but it seems unlikely that someone starts just now :)

great that you want to take care of it, @andresmc98

andresmc98:
I think the issue its solved or I can't replicate it

PR #2782, opened 2022-05-27
Fix issue #2627

andresmc98:
I make changes on two files

scss/message/_message.scss
I change the attribute of the cursor and add a span propierty, now the cursor only changes to pointer when you are over the "Forwarded message" text instead all the wide area.

src/renderer/components/message/Message.tsx
I add the onClick event to the "Forwarded message" text, now you only can click in the text instead all the wide area

I decide to made it in this way because if I change the size of the clickable area that could cause some troubles whit the translations of the text (forwarded message), breaking the div or the text not all be clickable


src/renderer/components/message/Message.tsx
```diff
@@ -105,21 +105,26 @@ const ForwardedTitle = (
...
+        reactStringReplace(
+          tx('forwarded_by', '$$forwarder$$'),
+          '$$forwarder$$',
+          () => (
+            <span
+              onClick={() => onContactClick(contact)}
+              key='displayname'
+              style={{ color: color }}
+            >
+              {overrideSenderName ? `~${overrideSenderName}` : displayName}
```
Jikstra, line 119: Little bit suprised about the reactStringReplace, can we not have the forwarded by string inside the span?

Simon-Laux: other languages can have it at other places

Jikstra: Ahh right! Gooot it

PR #2783, opened 2022-05-27
Added vertical space between webxdc start-button and message-text

### LINKED ISSUE #2762: missing vertical space between webxdc start-button and message-text
between the webxdc-start-button  and the messages text, there should be some vertical spacing.

i suggest to use the same amount of space as between webxdc-name/-summary and webxdc-start-button; however, the additional space must not be applied if there is _no_ messages-text, in this case, things are already fine.

Screen Shot 2022-05-17 at 18 44 08
red: there is some space missing

&nbsp;
as a side-effect, this may also fix the floating of big emojis into the button as reported in some other channel:
image_2022-05-17_02-16-02
this is a cool-looking bug, btw :)

gcamach0:
For bug #2762
I made a selector for adding a top margin of 10px. I had to made the span inside the message inline-block to be able to add the margin(I think that does not create any conflict in other parts but it would be good if you can confirm it).
I also made the MessageBody component not to be rendered at all if there was no text at all. I thought an alternative to rendering the MessageBody conditional might be to use the :not(:empty) selector, but at the same time I considered that if you're not going to have any text, there wasn't even a need to run the code of MessageBody in the first place but I would also love to know your input in that.

Here is an example of the spacing between the button and the message.
Screen Shot 2022-05-27 at 15 33 37

And here is an example without a message.
Screen Shot 2022-05-27 at 15 34 04

The styles are only been applied when the there's a webxdc before the message.
Screen Shot 2022-05-27 at 15 37 47

closes #2762

### DIFF
--- a/src/renderer/components/message/Message.tsx
+++ b/src/renderer/components/message/Message.tsx
@@ -346,9 +346,9 @@ const Message = (props: {
       <div dir='auto' className='text'>
         {message.isSetupmessage ? (
           tx('autocrypt_asm_click_body')
-        ) : (
-          <MessageBody text={text || ''} />
-        )}
+        ) : text ? (
+          <MessageBody text={text} />
+        ) : null}
       </div>
     )
   }
--- a/scss/message/_message.scss
+++ b/scss/message/_message.scss
@@ -140,6 +140,11 @@
         padding: 6px 13px;
       }
     }
+
+    .webxdc + .text span {
+      display: inline-block;
+      margin-top: 10px;
+    }
   }
 
   .metadata {


src/renderer/components/message/Message.tsx
```diff
@@ -346,9 +346,9 @@ const Message = (props: {
       <div dir='auto' className='text'>
         {message.isSetupmessage ? (
           tx('autocrypt_asm_click_body')
-        ) : (
-          <MessageBody text={text || ''} />
-        )}
+        ) : text ? (
+          <MessageBody text={text} />
+        ) : null}
```
Jikstra, line 351: You can write this also as `text && <MessageBody text={text} />`

Jikstra:
Looks good, thank you :)

PR #2794, opened 2022-06-01
push strings to android repo

Simon-Laux:
the two remaining strings will likely change soon so I did not add them


r10s:
note that there is an issue with "About", https://github.com/deltachat/deltachat-desktop/issues/2796

moreover, the key `jump_to_search ` is renamed to `focus_search_input` and also adapted the english string, this is more clear and we are already using the "focus" wording for the input field, so we should be consistent there.

src/renderer/components/KeyboardShortcutHint.tsx
```diff
@@ -213,34 +213,34 @@ export function getKeybindings(
...
         ],
       },
       {
-        title: tx('desktop_keybindings_action_scroll_messages'),
+        title: tx('scroll_messages'),
         keyBindings: [['PageUp'], ['PageDown']],
       },
       {
-        title: tx('desktop_keybindings_action_jump_to_search'),
+        title: tx('jump_to_search'),
```
Simon-Laux, line 227:
```suggestion
        title: tx('focus_search_input'),
```

Simon-Laux:
so I added 'about' key back to untranslated to postpone the issue so I can merge this pr quicker

Jikstra (approved):

PR #2811, opened 2022-06-13
Adapt login screen to look like android

flipsimon:
image


Jikstra:
Screenshot_20220621_013332

A little bit of padding/margin and i think it should be fine

r10s:
> A little bit of padding/margin and i think it should be fine


also consider using the button style used elsewhere on desktop (so, rounded corners) and making the last two options more "secondary" (cmp right image at https://github.com/deltachat/deltachat-android/pull/2284 or

flipsimon:
Will outsource changes to delta-button-round in separate pr soon. Planing adjustments for small screen height.

PR #2814, opened 2022-06-25
[WIP] use jsonrpc

Simon-Laux:
https://github.com/deltachat/deltachat-core-rust/pull/3463


Simon-Laux:
in the future we need to mange which context is selected in the frontend, maybe we should use an react context for it?
point is that all account/context specific functions now need an accountId there is no selected context in the jsonrpc api

src/main/deltachat/webxdc.ts
```diff
@@ -321,6 +321,10 @@ If you think that's a bug and you need that permission, then please open an issu
       )
     })
 
+    ipcMain.handle('close-all-webxdc', () => {
+      this._closeAll()
+    })
+
```
obsoleszenz, line 327: Why is this part of this pr?

Jikstra, line 327: Why is webxdc part of this pr?

src/renderer/backend-com.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+  constructor() {
+    super(new ElectronTransport())
+  }
+}
+
+export const BackendRemote: BaseDeltaChat<any> = new ElectronDeltachat()
+
+/** Functions with side-effects */
+export namespace EffectfulBackendActions {
```
Jikstra, line 46: Why the seperation? Also the other backend actions also do have side effects of course.

src/renderer/ScreenController.tsx
```diff
@@ -89,10 +90,16 @@ export default class ScreenController extends Component {
...
   async selectAccount(accountId: number) {
+    // for now we still need to call the backend function,
+    // because backend still has sleected account
     await DeltaBackend.call('login.selectAccount', accountId)
+
     this.selectedAccountId = accountId
-    const account = await DeltaBackend.call('login.accountInfo', accountId)
-    if (account.type === 'configured') {
+    ;(window.__selectedAccountId as number) = accountId
+    const account = await BackendRemote.rpc.getAccountInfo(
```
Jikstra, line 99: Not a big thing of course, but I would like if the way to call the backend methods would be simply `Backend.getAccountInfo()` or is there any specific reason for the `.rpc.`? Also I think Backend already means in our architecture that it's somewhat remote.

Jikstra:
> in the future we need to mange which context is selected in the frontend, maybe we should use an react context for it? point is that all account/context specific functions now need an accountId there is no selected context in the jsonrpc api

I think a store is the way to go. React Contexts are problematic in my opinion, because they are annoying to pass to a class component and to provide too.

src/main/deltachat/webxdc.ts
Simon-Laux: generally this pr is about moving code to the frontend and calling electron through runtime.ts
This function is ran at logout which I partly already moved to the frontend.

src/renderer/backend-com.ts
Simon-Laux: eventually those functions should be transferred to the screen context / screen controller.
They call electron / runtime stuff so I figured to put them into their own functions.
of course we can discuss naming, though that would become unnecessary if we move the functions anyway?

src/renderer/ScreenController.tsx
Simon-Laux: rpc is the raw autogenerated stuff (from `deltachat-jsonrpc/typescript/generated/client.ts`), other functions on `BaseDeltaChat` are implemented manually:
`deltachat-jsonrpc/typescript/src/client.ts#L43`: 


https://github.com/deltachat/deltachat-core-rust/blob/83d71068f7335574e9547bea441bfc5ccd81f9e0/deltachat-jsonrpc/typescript/src/client.ts#L43-L55

Simon-Laux: yeah we could rename `BackendRemote` to `Backend` (just wanted something different than `DeltaRemote`)

PR #2817, opened 2022-06-30
Add floating action button to start a new chat

flipsimon:
Bildschirmfoto 2022-06-30 um 11 31 54


Simon-Laux:
changelog entry is missing, but after that it's good to be merged IMO. Let's see how people will respond to the new button.

Issue #2582, opened 2022-03-04
`deltachat --version` does not work


Simon-Laux:
would be nice if it could print some info without starting DeltaChat

Jikstra:
Also --help would be nice

nadia12-p:
hi guys, can I work on this?

nadia12-p:
hi! the --version flag is working, but I'm not sure what info you want to put in --help, the only command right now is 
--version, so it will only print that info, is it ok or am I missing something?

Simon-Laux:
there are more options, see https://github.com/deltachat/deltachat-desktop/blob/master/docs/CLI_FLAGS.md
Also because you haven't found this file on your own, we improve it's discoverability, what do you think we could do to make that file more visible?

nadia12-p:
@Simon-Laux thanks for the help!! I think keeping it like that is fine. However, I can't find the code where these options are declared which is why I'm confused hahaha do you know how they work or where I can look at the logic of these flags?

Simon-Laux:
closed by https://github.com/deltachat/deltachat-desktop/pull/2821

PR #2821, opened 2022-07-05
add flags --version,-v and --help,-h 

nadia12-p:
Issue #2582 

To run the flags you put in the terminal: 
deltachat-desktop {flag}


Jikstra (commented):
Hey, thanks for your effors. Having command line arguments and a help is highly appreciated, it's something that annoys me since a loong time already. What is the bin/create-project script doing?

src/cli/cli.js
```diff
@@ -0,0 +1,34 @@
+import arg from 'arg';
+import { getHelpMenu } from './help_menu.js';
+
+var pjson = require('../../package.json');
```
Jikstra, line 4: @Simon-Laux where do we store the version number/retrieve it currently?

nadia12-p:
> Hey, thanks for your effors. Having command line arguments and a help is highly appreciated, it's something that annoys me since a loong time already. What is the bin/create-project script doing?

You're welcome :) ah that script allows us to use es modules without needing to transpile for different node.js versions that may not have that support and it also calls the cli function from cli.js

src/cli/cli.js
Simon-Laux:
```
import { VERSION, GIT_REF, BUILD_TIMESTAMP } from '../shared/build-info'
```

Simon-Laux:
I'd prefer if it's part of the main process in `src/main`, please ask when you have electron questions how to integrate it.
There are multiple reasons for integrating it into the existing main process:
- Already uses typescript & esm transpilation
- the `src/` directory is not included when bundling it for release
- we don't use the "bin" argument in package.json, because we don't upload deltachat-desktop to npm currently, and if we would use it it would probably just say "start electron"


> to use es modules without needing to transpile for different node.js versions that may not have that support

we already transpile/compile our typescript code to plain javascript, also we don't support node versions under 16, because we compile our native library for node 16 napi. So there is no real need for your workarounds 😉 

To build run `npm run build` then you can test it:

To test it either do:
```
npm run start -- --help
```
or
```
npx electron . --help
```

src/cli/help_menu.js
```diff
@@ -0,0 +1,27 @@
+export function getHelpMenu() {
+    console.log("\n\n", "Options:", "\n\n\n",
```
Simon-Laux, line 2: you should use a https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals here for the string as it's easy to write and read.

src/cli/cli.js
```diff
@@ -0,0 +1,34 @@
+import arg from 'arg';
```
Simon-Laux, line 1: you could also use https://www.npmjs.com/package/rc, here because we already use it in other places, though not a big deal, because `arg` is already an implicit dependency of delta-chat desktop, so adding it does not make a big difference I think

package.json
```diff
@@ -58,7 +62,9 @@
     "deltachat",
     "email",
     "messaging",
-    "desktop"
+    "desktop",
+    "cli",
+    "create-project"
```
Simon-Laux, line 67: what does `"create-project"` mean?

src/cli/cli.js
Simon-Laux: See the edit in my previous comment

src/cli/help_menu.js
```diff
@@ -0,0 +1,27 @@
...
+    "Flag","\n\n",
+    "-", "\t\t\t\t", "script read from stdin","\n",
```
Simon-Laux, line 4: deltachat-desktop does not support user scripts at this time

src/cli/help_menu.js
nadia12-p: yeah, it's way easier thanks! :D

src/cli/cli.js
nadia12-p: at the end, I extended our current rc implementation

package.json
nadia12-p: I will remove it it's not used anymore

src/cli/help_menu.js
nadia12-p: ok I changed it, thanks !

nadia12-p:
Hi! I submitted new changes now it works as you suggested, you need to run for example:

`npm run dev -- --version` (-v,--help,-h)

and once it's finished building it will run the proper command and won't run the program. I also tested it with start command and it works fine, ex:

`npm run start -- --version`

package.json
```diff
@@ -82,6 +85,7 @@
     "@deltachat/react-qr-reader": "^4.0.0",
     "@mapbox/geojson-extent": "^1.0.0",
     "application-config": "^1.0.1",
+    "arg": "^5.0.2",
```
Simon-Laux, line 88: does not look like this is still needed

package.json
```diff
@@ -72,6 +72,9 @@
     "node": ">=16.0.0"
   },
   "productName": "DeltaChat",
+  "publishConfig": {
+    "access": "public"
+  },
```
Simon-Laux, line 77: what's this needed for?

src/main/help_menu.ts
```diff
@@ -0,0 +1,33 @@
+export function getHelpMenu() {
+
+    console.log(`
```
Simon-Laux, line 3:
```suggestion
/* ignore-console-log */
    console.log(`
```
To make the tests pass which normally only allows using our own logger, except when using this special comment

Simon-Laux (approved):
Wow nice: so much simpler now!

There are still some "formal" things to do before we can merge it:
1. run `npm run fix-formatting` and then make sure `npm run test` passes
2. add an entry to the changelog.md file like: "added cli arguments `--help` and `--version`"

Anyways thanks for contributing and sorry that I didn't respond sooner

package.json
nadia12-p: oh yeah, I forgot to remove it, I'll change it

package.json
nadia12-p: same as before, I forgot to remove it haha

src/main/help_menu.ts
nadia12-p: ok, I'll change it, run the commands, and add an entry to the changeling.md thanks!

PR #2827, opened 2022-07-13
Broadcast lists desktop

flipsimon:


.vscode/settings.json
```diff
@@ -1,3 +1,4 @@
 {
-  "typescript.tsdk": "node_modules/typescript/lib"
+  "typescript.tsdk": "node_modules/typescript/lib",
+  "cmake.configureOnOpen": false
```
Simon-Laux, line 3: what does this line do? why is it in the pr?

PR #2842, opened 2022-07-30
fix display of quoted forwarded message fix typo in log message

### LINKED ISSUE #2836: forwarded title is misleading in quotes
when a message as 

```
FORWARDED MESSAGE
text bla
```

is quoted, the quote should just look same as the forwarded message, so 

```
| FORWARDED MESSAGE
| text bla

text below quote
```

however, instead of FORWARDED MESSAGE, the name of the forwarded is shown, even colored. this looks as if the forwarded has written the text "text bla" which is not true.

see ios/android how this is handled there. android shows FORWARDED BY FORWARDER, also uncolored intentionally, which is an alternative (desktop already uses this term, so that is probably also the way to go there).

to the gray: forwarded messages and deleted messages are shown both uncolered as the original sender is unknown.

Simon-Laux:
Bildschirmfoto 2022-07-30 um 19 35 58

fixes #2836
also fix changelog order

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,16 @@
 
 ## [Unreleased][unreleased]
 
+## Added
+- Broadcast lists as experimental feature
+
+## Fixed
+- fix display of quoted forwarded messages
+
 ## [1.31.0] - 2022-07-17
 
 ## Added
 - Floating action button in chatlist to start a new chat
-- Broadcast lists as experimental feature
 
 ## Fixed
 - use addr if displayname is not set for webxdc selfName #2803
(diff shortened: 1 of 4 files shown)


r10s:
to be in sync with android and as the important information is the fact the message is forwarded - and not the name - we should tune down the name and show it in the same color as the rest, cmp. https://github.com/deltachat/deltachat-android/pull/1795

r10s (commented):
i cannot say much to the code, but ui-wise:

to be in sync with android and as the important information is the fact the message is forwarded - and not the name - we should tune down the name and show it in the same color as the rest, cmp. https://github.com/deltachat/deltachat-android/pull/1795 - so all in the same gray/black.

nb: i would not see this or as a necessity for 1.32.

PR #2874, opened 2022-09-06
use more of the existing jsonrpc: accept chat

Simon-Laux:


src/renderer/backend-com.ts
```diff
@@ -71,4 +71,21 @@ export namespace EffectfulBackendActions {
...
+
+  export async function blockChat(account_id: number, chatId: number) {
+    await BackendRemote.rpc.acceptChat(account_id, chatId)
+    window.__refetchChatlist && window.__refetchChatlist()
+  }
+
+  export async function deleteChat(account_id: number, chatId: number) {
+    // TODO implement in jsonrpc
+    await DeltaBackend.call('chat.delete', chatId)
+    window.__refetchChatlist && window.__refetchChatlist()
```
Jikstra, line 89: Don't really like using a global here but whatever?

Jikstra (approved):

src/renderer/backend-com.ts
Simon-Laux: me neither, it is what it is for now, the goal is to have those events somehow in core, in essence most functions of `EffectfulBackendActions` are temporary until we moved more code or implemented the events into core.

PR #2879, opened 2022-09-09
migrate to jsonrpc (mostly message related functions)

Simon-Laux:
port more functions to jsonrpc also fix selecting chat after forwarding to it

This needs state of this core pr: https://github.com/deltachat/deltachat-core-rust/pull/3590

## Test

Generally please test everything, but these areas in particular:

- [x] gallery next & prevoius
- [x]  message details
- [x]  Save Image
- [x] message context menu
- [x] join video chat
- [x] send messages in maps
- [x]  chatlist
- [x]  chatlist send failing message
- [x]  chatlist jump to message
- [x] chat audit dialog
- [x] gallery
- [x] composer and drafts (set/delete draft)
- [x] privateReply
- [x] apply bot command suggestion from a message
- [x] mailto dialog
- [x]  getDraft
- [x]  delete message
- [x] forwardMessage
- [X] Map
- [x] ViewGroup
- [x] View broadcast list
- [x] when chat has unreads start at right place
- [x] setMuteDuration
- [x] mark messages as seen when they are loaded or come into view
- [ ] Are mailing lists displayed correctly?
- [x] show qr code for self and for group
- [x] all contact lists
- [x] block and unblock contact
- [X] connectivity in sidebar, settings and toast also maybe network

# TODO

- [x] removed maybe mark seen stuff(need to reimplement it)
- [x] port contact related functions (the ones that are used most)

## Known Bugs that need to be fixed:

- [x] Mark messages as read
    - [ ] on window focus mark noticed current chat
    - [ ] on chatselect mark noticed current chat
    - [x] on loading messages in chat mark as read?
    - ~~(or only when they come into view?)~~ no that probably too hard to get right, so right now out of scope
- [x] Fix setting draft
- [x] Fix setting quote into draft
- [x] Fix remove draft
- [x] Fix jump to last last unread (jumps to wrong message)
- [x] Strange hyphen/- at end of message - markdown is broken too maybe that’s a hint / text seems like its empty string instead of null, 
maybe core should check or comes it form ui?
- [x] Turning of padding on only image messages does not work 
- [x] Set mute doesn’t do anything = seems to be sth in jsonrpc/core, strangely unmute works, the conversion seems also to work - ah seems 
like it was absolute vs relative time
- [x] Message parser is not in affect, why do I not get any error in console? but it works in map so only message seems to be affected
- [x] block contact, gets removed from chatlist, but chattiest is not refreshed
- [x] private reply does not select the chat
- [x] don't show checkmark on incoming messages in chatlist


src/renderer/stores/chat.ts
```diff
@@ -1027,26 +1060,36 @@ class ChatStore extends Store<ChatStoreState> {
...
-        !(await DeltaBackend.call(
-          'chat.setMuteDuration',
-          payload.chatId,
-          payload.muteDuration
-        ))
-      ) {
-        return
-      }
+
+      BackendRemote.rpc.setChatMuteDuration(
```
Jikstra, line 1069: I think we should await this here

PR #2881, opened 2022-09-15
Implement intersection observer to mark messages as read that are act…

Jikstra:
…ually getting displayed on screen

### DIFF
--- a/src/renderer/components/message/MessageWrapper.tsx
+++ b/src/renderer/components/message/MessageWrapper.tsx
@@ -1,17 +1,64 @@
-import React from 'react'
+import React, { useLayoutEffect } from 'react'
 import Message from './Message'
 import { ConversationType } from './MessageList'
 import { Type } from '../../backend-com'
+import { C } from 'deltachat-node/node/dist/constants'
+import { getLogger } from '../../../shared/logger'
+
+const log = getLogger('renderer/message/MessageWrapper')
 
 type RenderMessageProps = {
+  key2: string
   message: Type.Message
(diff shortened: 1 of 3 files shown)


src/renderer/stores/chat.ts
```diff
@@ -784,8 +784,8 @@ class ChatStore extends Store<ChatStoreState> {
           return false
         }
 
-        await BackendRemote.rpc.marknoticedChat(accountId, chatId)
-        chat.freshMessageCounter = 0
+        //await BackendRemote.rpc.marknoticedChat(accountId, chatId)
+        //chat.freshMessageCounter = 0
```
Jikstra, line 788: Not sure about this part. I think we should call marknoticedChat always?

src/renderer/components/message/MessageWrapper.tsx
```diff
@@ -1,17 +1,64 @@
...
+  key2: string
   message: Type.Message
   conversationType: ConversationType
+  unreadMessageInViewIntersectionObserver: React.MutableRefObject<IntersectionObserver | null>
 }
 
 export const MessageWrapper = (props: RenderMessageProps) => {
+  const state = props.message.state
+  const shouldInViewObserve =
+    state === C.DC_STATE_IN_FRESH || state === C.DC_STATE_IN_NOTICED
```
Jikstra, line 20: Also not sure about this one? If this bool is true we observe that message if it's in the viewport or not

Simon-Laux (approved):

PR #2887, opened 2022-09-18
add mailinglist address to chat subtitle

### LINKED ISSUE #2840: show mailinglist address
while you can meanwhile write to mailinglists, the email address, a messages is actually sent to, is opaque.

with the new ffi from https://github.com/deltachat/deltachat-core-rust/pull/3524, we can get this address from the core and should show it.

android/ios show them in the profile as there is few space in the chat title, see https://github.com/deltachat/deltachat-ios/pull/1654 and https://github.com/deltachat/deltachat-android/pull/2354, for desktop, however, i suggest to show it directly in the subtitle of the chat, so right of the text "Mailing List", after an n-dash, sth. as:

Screenshot 2022-07-26 at 19 06 56

cave: while often the address is "fine", it may also be quite long - so take care not to let it break the layout.

Simon-Laux:
needs new core version with https://github.com/deltachat/deltachat-core-rust/pull/3607 merged 
closes #2840


r10s (approved):
lgtm, as far as i can tell :)

PR #2894, opened 2022-09-22
more jsonrpc progress that is already mergable

Simon-Laux:
- more jsonrpc progress that is already mergable
- fix eslint stuff
- fix formatting

@Jikstra can you look into these warnings and into the last 2 commits:
```
./src/renderer/components/message/MessageList.tsx
  106:47  warning  The ref value 
'unreadMessageInViewIntersectionObserver.current' will likely have changed by 
the time this effect cleanup function runs. If this ref points to a node 
rendered by React, copy 'unreadMessageInViewIntersectionObserver.current' to 
a variable inside the effect, and use that variable in the cleanup function  
react-hooks/exhaustive-deps
  108:6   warning  React Hook useEffect has an unnecessary dependency: 
'unreadMessageInViewIntersectionObserver.current'. Either exclude it or 
remove the dependency array. Mutable values like 
'unreadMessageInViewIntersectionObserver.current' aren't valid dependencies 
because mutating them doesn't re-render the component                 
react-hooks/exhaustive-deps
```

Lets ensure the eslint fixes are not breaking anything.

### DIFF
--- a/src/main/deltachat/chat.ts
+++ b/src/main/deltachat/chat.ts
@@ -1,15 +1,9 @@
 import { C } from 'deltachat-node'
 import { getLogger } from '../../shared/logger'
 import SplitOut from './splitout'
-import { set_has_unread } from '../tray'
-import { app } from 'electron'
 
 const log = getLogger('main/deltachat/chat')
 export default class DCChat extends SplitOut {
-  getEncryptionInfo(chatId: number) {
-    return this.selectedAccountContext.getChatEncrytionInfo(chatId)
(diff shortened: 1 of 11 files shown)


Jikstra:
You cannot have references in dependency arrays: see https://epicreact.dev/why-you-shouldnt-put-refs-in-a-dependency-array/

Simon-Laux:
> And it's also not needed, it's like a pointer that is either valid or null.

so we make eslint ignore it?

PR #2896, opened 2022-09-23
make timestamp update more Entity Component System like.

### LINKED ISSUE #2885: Refresh relative timestamp on messages more frequently
### Problem
"Sometimes I focus deltachat after an hour and it still says I sent my messages 10min ago until I renter the chat or reload the chatlist"

### Proposed solution

- refresh relative timestamps every 10min regardless if window is focused or not
- and on window focus force refresh all timestamps

Simon-Laux:
fixes #2885

Basically move away from multiple setInterval calls to one central function calling all update functions every 60s (nothing changed there) and additionally on window focus (added). I hope this makes the whole thing more reliable in order to fix #2885


Jikstra:
Good idea. I think it should be easy to implement. One singleton running, a hook to register/unregister a callback that is getting run from the component?

PR #2902, opened 2022-09-25
add donate link to settings

Simon-Laux:
donate-link

link leads to https://delta.chat/contribute#donate-anchor, which leads automatically to the localised version of the page.

So that the link is user findable with minimal effort. (we could add this to iOS and android too)
We can add inApp purchases, donation requests in device messages and other things later, this is just a start.

related to:
- https://github.com/deltachat/deltachat-pages/pull/561


Jikstra:
I like it!

PR #2913, opened 2022-10-01
add recently seen indicator

### LINKED ISSUE #2873: add recently seen indicator
the lists are sorted by "last seen" on recent master, having the "recently seen" indicator as well is a great overall improvement of the contact and chat lists.

corresponding pr:  
ios (with screenshot): https://github.com/deltachat/deltachat-ios/pull/1680
android (with screenshot): https://github.com/deltachat/deltachat-android/pull/2371
core: https://github.com/deltachat/deltachat-core-rust/pull/3560

Simon-Laux:
closes #2873


missytake (approved):
looks good and works :)

Jikstra:
Interesting way with the mask, but I think for this case we can simply use css & background-color? This way it's easier to change/modify.

Simon-Laux:
> but I think for this case we can simply use css & background-color? This way it's easier to change/modify.

It's actually harder and less flexible, background colors are different in different places and we need to keep that in mind there.

for the svg you just need to copy pastet both elements into one svg then you can edit in inkscape and directly see what your doing, after that split it into 2 seperate svg files again

PR #2918, opened 2022-10-02
move more methods to jsonrpc and delete some unused code

### LINKED ISSUE #2945: change nickname does not update the shown name in view profile dialog
should immediately update (listen on contacts changed event)

Simon-Laux:
needs core with the following prs merged:

- https://github.com/deltachat/deltachat-core-rust/pull/3641
- https://github.com/deltachat/deltachat-core-rust/pull/3645
- https://github.com/deltachat/deltachat-core-rust/pull/3653
- https://github.com/deltachat/deltachat-core-rust/pull/3671
- https://github.com/deltachat/deltachat-core-rust/pull/3672

fixes #2945

### Testing points

- [x] Sending Video chat invitations
- [x] Group view
- [x] Add contact to chat
- [x] Create new group (normal and verified)
- [x] Create new broadcast list
- [x] Set group name and picture
- [x] Set broadcast list name
- [x] Leave a group
- [x] Open mailto link with address set
- [x] Click on email address in message of existing and non-existing contact
- [x] Download full message (download on demand)
- [x] Search messages
- [x] set profile picture
- [x] set chat visibility (arrive/unarchive/pin)
- [x] disappearing messages
- [X] get locations
- [X] get Account size
- [x] remember last account
- [x] paste image from clipboard
- [x] autodeletion count
- [x] Themes
- [x] import/export keys -> import keys is still broken? or did it fail because I tried to import the same key I exported before?
- [x] login / logout - sync all accounts / not sync all accounts
- [x] change desktop settings
- [ ] securejoin
- [ ] delete contact (when creating chat, but then aborting creation it should delete contacts that were created in this process?)
- [x] change contact nickname
- [ ] send sticker
- [ ] messagelist
- [x] set language and load selected language
- [x] set chat bg img, test setting one of the defaults and custom


src/renderer/components/chat/ChatListHelpers.tsx
```diff
@@ -24,17 +24,20 @@ export function debounceWithInit<ARGS extends Array<any>>(
...
   const [ids, setIds] = useState<number[]>([])
 
   const debouncedSearchMessages = useMemo(
     () =>
       debounceWithInit((queryStr: string | undefined) => {
-        DeltaBackend.call('messageList.searchMessages', queryStr || '', 0).then(
-          setIds
-        )
+        BackendRemote.rpc
+          .searchMessages(selectedAccountId(), queryStr || '', chatId)
```
obsoleszenz, line 37: I dislike a bit that its searchMessages(accountId, query, chatId) instead of searchMessages(accountId, chatId, query)

Jikstra, line 37: I dislike a bit that it's searchMessages(accountId, queryStr, chatId) instead of searchMessages(accountId, chatId, queryStr)

Jikstra (approved):

src/renderer/components/chat/ChatListHelpers.tsx
Simon-Laux: sure that is a change we could revert, but should we? I mean I could argue that `queryStr` is more important and chatId can be null | number, in future maybe also | undefined, which means we could omit it here. https://github.com/Frando/yerpc/issues/10

Jikstra: I think it's not so important if we have to change something with yerpc

PR #2923, opened 2022-10-04
hide ephemeral timer menu options for mailing lists

### LINKED ISSUE #2920: hide ephemeral timer menu options for mailing lists
in mailing lists, "disappearing messages" currently do not work nicely.

@adbenitez on dev chat:

> at the beginning I thought it would be useful but it only causes annoying situations and then if you want it enabled it would not work because new members will probably not know it was enabled, when you join you receive the last 5 messages from group but the mailing list welcome message will probably disable the timer

ios has never enabled the options and on android they're disabled at https://github.com/deltachat/deltachat-android/pull/2394 - if we find the time, maybe we can do $things about that, but for now, best is to just disable ephemeral timer options for mailing lists.

cc @adbenitez @link2xt

Simon-Laux:
closes #2920

### DIFF
--- a/src/renderer/components/ThreeDotMenu.tsx
+++ b/src/renderer/components/ThreeDotMenu.tsx
@@ -62,10 +62,11 @@ export function useThreeDotMenu(selectedChat: Type.FullChat | null) {
       })
 
     menu = [
-      canSend && {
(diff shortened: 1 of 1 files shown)


r10s (approved):
nice, lgtm.

PR #2924, opened 2022-10-04
Reimplement jumpDown button

Jikstra:
# ToDo
- [x] jumping to newest message corrupts messagelist


src/renderer/stores/chat.ts
```diff
@@ -843,12 +849,76 @@ class ChatStore extends Store<ChatStoreState> {
...
+            // If we are not switching chats, add current jumpToMessageId to the stack
+            const currentChatId = chatStore.state.chat?.id || -1
+            if (chatId !== currentChatId) {
+              jumpToMessageStack = []
+            } else if (chatStore.state.jumpToMessageStack.indexOf(addMessageIdToStack) !== -1) {
+              jumpToMessageStack = chatStore.state.jumpToMessageStack
+            } else {
+              jumpToMessageStack = [...chatStore.state.jumpToMessageStack, addMessageIdToStack]
+            }
+          }
```
Jikstra, line 913: I'm not really happy with this part, but also don't know a way of improving that code.

src/renderer/components/message/Message.tsx
```diff
@@ -499,7 +502,8 @@ export const Quote = ({
     <div
       className='quote-background'
       onClick={() => {
-        hasMessage && jumpToMessage(quote.messageId)
+        quote.kind === 'WithMessage' &&
+          jumpToMessage(quote.messageId, true, msgParentId)
```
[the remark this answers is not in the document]
Jikstra: If you jump to a quoted message, we add the messageId from where we jumped to that mesage (msgParentId) to a stack. And every time you click on the jumpDown button, we jump to last messageId in that stack.

src/renderer/components/message/MessageList.tsx
```diff
@@ -444,6 +481,7 @@ const MessagePageComponent = React.memo(
         continue
       }
       if (messagePage.dayMarker.indexOf(messageId) !== -1) {
+        console.log('xxx', message.timestamp)
```
Jikstra, line 484:
```suggestion
```

PR #2957, opened 2022-10-17
fix results of search in chat are not ordered by newest first

Simon-Laux:


src/renderer/components/chat/ChatListHelpers.tsx
```diff
@@ -34,7 +34,13 @@ export function useMessageResults(
       debounceWithInit((queryStr: string | undefined) => {
         BackendRemote.rpc
           .searchMessages(selectedAccountId(), queryStr || '', chatId)
-          .then(setIds)
+          .then(ids => {
+            if (chatId) {
+              // fix that results of search in chat are not ordered by newest first
```
r10s, line 39:
```suggestion
              // in-chat search results need to be be ordered by newest first
```

r10s (approved):
lgtm, i would reword the comment as shown.

wondering, btw, why the reodering is not needed for global search, where, at the end, newest are shown first at well. maybe the swap takes placed and that can be unified (remove other swapping and swapping unconditional here).

PR #2960, opened 2022-10-19
Update translations (19.10.2022)

### LINKED ISSUE #2959: rename "webxdc"  to "Private Apps" for the enduser
as discussed at https://github.com/deltachat/deltachat-android/pull/2377 and at other places, the term "Webxdc" is too technical, it is the format of the more understandable term "Private App".

in UI, currently, this seems to affect  the tab "Webxdc" in the gallery only. this one should be called "Private Apps" (string `webxdc_apps`). the "empty hint" should use the string `tab_webxdc_empty_hint`.

moreover, on android/ios, we use the string `webxdc_app` if a concrete app has an empty summary - either in the bubble or as a draft. this makes a bit more clearer, what the things is about. if there _is_ a summary, however, the string is _not_ shown to not clutter UI with too many information. maybe desktop should also adapt to that behavior.

Simon-Laux:
fixes #2959

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 ### Changed
+- Update translations (19.10.2022)
 
 ### Fixed
 
--- a/_locales/az.xml
+++ b/_locales/az.xml
@@ -391,24 +391,26 @@
 
 
     <!-- system messages -->
-    <!-- deprecated -->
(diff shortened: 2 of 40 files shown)


r10s:
> fixes #2959

i do not think, this pr fixes #2959, the wording "private apps" is already on master, see eg. https://github.com/deltachat/deltachat-desktop/blob/master/_locales/de.xml#L159 and the diff of this pr.

so, the tabs mentioned at #2959  are maybe hardcoded or so.

r10s (approved):

PR #2964, opened 2022-10-21
Message reactions with emojis

Simon-Laux:
Emoji reactions for chat messages.

**Wireframes**

https://whimsical.com/reactions-at-deltachat-DK8ZMpWSZmtP19Pvs9C8Uy

**Preview**

https://github.com/deltachat/deltachat-desktop/assets/1822473/ccfaee4d-c979-4a42-b99c-98222f57cae7

**Todo**

- [x] `AbsolutePositionHelper` to help us position the reaction bar
- [x] `OutsideClickHelper` to detect clicks outside the current component
- [x] Show shortcut menu to create a reaction
- [x] Give option to use a picker to select from all available emojis
- [x] Display current reactions under each message
- [x] Show a list of all reactions by user on a message
- [x] Do not show reactions on autocrypt setup messages
- [x] Design for all new components
- [x] Reaction bar positioning for only-media messages
- [x] Make `EmojiPicker` own component and reuse it
- [x] Nice colors for all sorts of themes
- [x] Show "..." even if custom icon is selected
- [x] Bug: Show all emojis panel cuts out of screen
- [x] Bug: React button is cut
- [x] Have a "React" option in context menu

Closes: #3626 

Needs core PRs to be merged:
- https://github.com/deltachat/deltachat-core-rust/pull/3644
- https://github.com/deltachat/deltachat-core-rust/pull/3686


Simon-Laux:
Bildschirmfoto 2022-10-22 um 00 36 41

found a bug

dotlambda:
> we want one version out there that can receive but not send

Why?

adzialocha:
I've rebased this to `master` and made the component use CSS modules

adzialocha:
> i know, this is a draft and too early for review. however, as easily overseen on final review: the picker is cut off when opened from a message close to the bottom of the window: 

That's something I thought I took care of, interesting! Can't reproduce it right now, maybe it got "fixed" by a recent change I did? Definitely a good reminder for final review :+1:

adzialocha:
Okay, I think this is ready for review & discussion now! I've attached a preview video in the PR's description.

Questions from my end:

1. I need some guidance in how to approach theming for this, so far only the full emoji picker is adjusting to the themes. Should I introduce new theme variables or is there something I can re-use?
2. For "only-media" messages (no text, just image / video or sticker) we have a special layouting case as the message only looks good when it is as wide as the image itself. So far I've came up with this solution (put reactions in right corner, underneath everything):
2024-01-18-200658_331x371_scrot

adzialocha:
This is how it looks like when "too many" emojis have been selected (default is set to `5`, for this image I've set it to `2`, just so you get the idea):

2024-01-18-202004_515x71_scrot

Simon-Laux:
> Should I introduce new theme variables or is there something I can re-use?

In themebase there are already some generic variables that can be re-used such as `--bgPrimary`, though seems like for `$bgSecondary` no plain css var exists yet (it is only available under different names at the moment). 

This question ties in into https://github.com/deltachat/deltachat-desktop/issues/3018, I think it makes sense to follow what I thought about theming there, basically "Don't reuse css vars across components, define them in the themebase for all components and do the reusing of colors with sass variables in there, to have all colours defined in themebase)"

But feel free to say if you have any better ideas, though maybe rather sth for #3018 and for now we do like it says in that issue for new stuff?


> 2. For "only-media" messages (no text, just image / video or sticker) we have a special layouting case as the message only looks good when it is as wide as the image itself. So far I've came up with this solution (put reactions in right corner, underneath everything)

I don't have any strong opinions or solutions, I would try moving it to the left so it starts there, maybe?

Septias:
Great job on this feature, it really looks amazing already. Two little things though:

1. Even though the Pop-up animation is really cool, I find it a little too exaggerated. I would personally go for fewer scaling.

2. The borders look a bit hard, I would go for a little lighter color. 

These are just my two humble cents. If you don't agree, feel free to just ignore this 😊

r10s:
for 2. (hard border colors), i'd just go for the iOS colors, we tweaked them quite accordingly, see https://github.com/deltachat/deltachat-ios/pull/2043 - we aim to pick them up for android as well

adzialocha:
Alright, I've had another go on the styling! Thank you for the comments. This is what I've changed in terms of design:

* Reactions bar is changing color based on themes
* Less border in reactions bar
* Same shadow & border as in context menu
* Shadow under reactions in message
* Shortcut menus next to messages are rounded now and have a background
* Less extreme animations and hover effects

I've also fixed the issue with the overlapping emoji picker (was indeed not reproducible for me as my window was always fixed to the bottom, good lesson!) and added a "Reaction" option in the context menu for the right-clickers.

src/renderer/components/Reactions/styles.module.scss
```diff
@@ -0,0 +1,44 @@
...
+  margin-right: 10px;
+  position: relative;
+  top: 20px;
+}
+
+.emoji {
+  background-color: var(--messageIncomingBg);
+  border-radius: 24px;
+  border: solid 1px rgba(100, 100, 100, 0.5);
+  box-shadow: 0 1px 1px rgba(0, 0, 0, 0.2);
```
r10s, line 12: the border is still a bit hard,  cmp. https://github.com/deltachat/deltachat-desktop/pull/2964#issuecomment-1899250677 , 0.3 alpha fixes that

```suggestion
  border: solid 1px rgba(100, 100, 100, 0.3);
```

in addition, i removed the box-shadow here completely. it affects only areas outside the bubble - and using the same, flat style as the normal bubbles just looks better and less 

before / after (including the font change):

Screenshot 2024-01-22 at 21 03 48 
Screenshot 2024-01-22 at 21 08 20

src/renderer/components/Reactions/styles.module.scss
```diff
@@ -0,0 +1,44 @@
...
+      top: 1px;
+    }
+  }
+}
+
+.emojiCount {
+  bottom: 1px;
+  color: var(--messageText);
+  font-size: 11px;
+  font-weight: bold;
```
r10s, line 40: bold looks, well, a bit bold :) instead, use a slightly larger font size. see above for comparison

```suggestion
  font-size: 12px;
```

r10s (approved):
lgtm function-wise, i made some slight css suggestions that improve things a little in my eyes - but if you disagree, it's fine as well and still lgtm :)

all in all, a great thing! thanks a lot for making this happen!

for code-level-review, this as well lgtm, but as a non-ts-expert, this does not mean much :)

src/renderer/components/OutsideClickHelper.tsx
Simon-Laux, line 1: instead of this you could also do a slight backdrop like dialogs have, this could 
A. simplify the code (no event target stuff)
B. Visually highlight the component even more

But this is no merge blocker we can still do this later if we want, this seems to work fine right now.

src/renderer/components/dialogs/ReactionsDialog/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+      const result = await BackendRemote.rpc.getContactsByIds(
+        accountId,
+        contactIds
+      )
+
+      setContacts(
+        Object.values(result).map(contact => {
+          return {
+            ...contact,
+            emoji: reactionsByContact[contact.id][0],
```
Simon-Laux, line 72: we could also already show all emojis, would be future proof, but for now only optional. no blocker

src/renderer/components/OutsideClickHelper.tsx
adzialocha: > instead of this you could also do a slight backdrop like dialogs have

I had this actually originally implemented but decided against it as it became confusing to block the whole UI for a "quick action". For dialogs it makes sense as they're visually much more prominent and really _should_ focus the attention on them but a "shortcut bar" is something which shouldn't take away the focus from the chat itself.

Looking at other applications you can always cancel your action by clicking somewhere else and continuing there without any steps in-between. That's a) faster navigation and b) no confusing blocking.

Code-wise: you will still need to catch mouse events, surely you don't need to traverse parts of the DOM, but since we're only doing it for the observed element itself the complexity is negligible (2-3 nested tags).

src/renderer/components/dialogs/ReactionsDialog/index.tsx
adzialocha: Yeah, let me add a comment here!

adzialocha: Is the array actually sorted by something? ("latest first" or just ordered by occurrence?)

adzialocha:
I've added now keeping the hover icons visible on hover as it prevented some annoying flickering and I could make the event logic more efficient (only one global listener vs. one per message)

r10s:
🎉

Septias:
congrats!

PR #2970, opened 2022-10-31
don't show dash in mailinlist title, if there is no mailinglist address closes #2965

### LINKED ISSUE #2965: mailing list address should be added only to the subtitle if set
mailing list do not always have a mailing list address - in these cases, [`dc_chat_get_mailinglist_addr()`](https://c.delta.chat/classdc__chat__t.html#a4dd46f3330e571eb24a5e04a442f0f89) returns an empty string - and the subtitles ends with a dash. this does not look good and also it looks as if sth. is left out or buggy.

we should add the dash only if there is actually an address.

Screenshot 2022-10-25 at 13 15 20

very minor thing, no urgency :)

Simon-Laux:


r10s (approved):
lgtm

PR #2980, opened 2022-11-04
fix contact requests button to run block function and show right label (Group: delete, DM: block)

### LINKED ISSUE #2877: Contact request just offers to delete no blocking
always run block, but change label to "delete" for groups.

target this after merging #2874

Simon-Laux:
closes #2877


r10s (approved):
lgtm - btw, this now also allows blocking mailinglists!

PR #2982, opened 2022-11-04
reduce notifications

### LINKED ISSUE #2278: Too many Notifications
I just opened deltachat and had around 300 new messages. And i got wayyy to many notifications about new messages. Not sure how we can debounce this in a smart way, but definetly something has to improve.

Simon-Laux:
when many messages are received at once (for example on startup).
Also show notifications when showNotificationContent is disabled (that was broken).

Bildschirmfoto 2022-11-04 um 22 41 44

Bildschirmfoto 2022-11-04 um 22 48 19

On first start notifications are grouped as soon as there is more than one, and after that they are grouped as soon as 
there are more than 3 at a time, we could still change that behaviour, both numbers are constants that can be easialy 
changed without the need to understand the code:

https://github.com/deltachat/deltachat-desktop/blob/585faf756bca476b7d429d237cbccfd5328d5631/src/renderer/system-integration/notifications.ts#L154-L156

# TODO

- [x] translation strings

needs a core with https://github.com/deltachat/deltachat-core-rust/pull/3643/ merged

closes #2278 for now, it is not perfect (on importing backup there are multiple `IncomingMsgBunch`-events, so 
multiple grouped notifications) but it already greatly reduces the notifications on startup.

### DIFF
--- a/src/renderer/system-integration/notifications.ts
+++ b/src/renderer/system-integration/notifications.ts
@@ -1,7 +1,6 @@
 import { T } from 'deltachat-node/deltachat-jsonrpc/typescript/src/lib'
 import { appName } from '../../shared/constants'
 import { getLogger } from '../../shared/logger'
-import { DcNotification } from '../../shared/shared-types'
 import { BackendRemote } from '../backend-com'
 import { isImage } from '../components/attachment/Attachment'
 import { jumpToMessage } from '../components/helpers/ChatMethods'
@@ -14,16 +13,20 @@ function isMuted(accountId: number, chatId: number) {
   return BackendRemote.rpc.isChatMuted(accountId, chatId)
 }
 
+type queuedNotification = {
+  chatId: number
+  messageId: number
+}
+
+let queuedNotifications: {
+  [accountId: number]: queuedNotification[]
+} = {}
+
 async function incomingMessageHandler(
   accountId: number,
   chatId: number,
   messageId: number
 ) {
-  if (accountId !== window.__selectedAccountId) {
-    // notifications for different accounts are not supported yet
-    return
-  }
-
   log.debug('incomingMessageHandler: ', { chatId, messageId })
 
   if (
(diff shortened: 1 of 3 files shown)


src/renderer/system-integration/notifications.ts
```diff
@@ -47,42 +50,131 @@ async function incomingMessageHandler(
...
+          accountId,
+        })
+      } else {
+        // messages from diffent chats
+        // title: "new messages"
+        // body: "324 new messages in 6 chats"
+        const chatCount = chatIds.length
+        runtime.showNotification({
+          title: tx('notify_new_messages'), // IDEA: when we support notifications from multiple accounts, then show account displayname here?
+          body: `${msgCount} messages in ${chatCount} chats`, // TODO TX
```
Hocuri, line 141: TODO

src/renderer/system-integration/notifications.ts
```diff
@@ -47,42 +50,131 @@ async function incomingMessageHandler(
...
+        // title: chatName
+        // body: "5 new Messages in ChatName"
+        const notificationInfo = await BackendRemote.rpc.getMessageNotificationInfo(
+          accountId,
+          notifications[0].messageId
+        )
+        const { chatName, chatProfileImage } = notificationInfo
+        runtime.showNotification({
+          title: chatName,
+          body: `${msgCount} new messages`, // TODO TX
```
Hocuri, line 128: TODO

Hocuri (commented):
So, putting the message ids into the event as a JSON wasn't actually useful?

Hocuri (approved):

r10s:
> So, putting the message ids into the event as a JSON wasn't actually useful?

if that is true, we should also remove the json generation (which is not even documented, we can re-add if really needed by some UI) from https://github.com/deltachat/deltachat-core-rust/pull/3643 - that way we save some complexity and maintenance. that way we have basically the "flush" event @link2xt suggested recently :)

PR #2984, opened 2022-11-05
remove DeltaBackend

### LINKED ISSUE #2076: Seperate IPC functions
So that we have at the end in runtime:
- listener to DC events (account specific, from things inside of the `main/deltachat` folder
- call DC functions
- listeners to electron back-end ipc
- ... functions for backend ipc functions

Goal is to don't have dependencies to electron functions in the renderer except for in the runtime.

### LINKED ISSUE #2938: State of porting to JSONRPC
## EffectfulBackendActions also using runtime function
- [x] messageList.saveMessageHTML2Disk [1]
- [x] chat.modifyGroup [2] - should be done by ui, the function it's comprised of should be available via jsonrpc

## move to jsonrpc
- [x] stickers.getStickers [1] - needs core access for folder name?
- [x] backup.export [3]
- [x] backup.import [1]
- [x] login.selectAccount [1]
- [x] login.logout [1]

## easy, low-hanging fruit
- [x] joinSecurejoin [2]
- [x] contacts.deleteContact [1]
- [x] contacts.changeNickname [1]
- [x] messageList.sendSticker [1]

## move to runtime
- [x] settings.setDesktopSetting [5] - eventually we might want to move it as ui config to account manager, but that's out of scope for now
- [x] settings.getDesktopSettings [3]
- [x] settings.saveBackgroundImage [2]
- [x] extras.setLocale [1]
- [x] extras.getLocaleData [1]

## misc

- [x] handle and distribute events in frontend (over jsonrpc)
- webxdc stays in backend for now, maybe register it somehow as second jsonrpc client which it's own instance (to avoid the problem of conflicting invocation ids)

### LINKED ISSUE #2981: replace contants imports form dc-node to jsonrpc-client package
change
```
import { C } from 'deltachat-node/node/dist/constants'
```
to
```
import { C } from '@deltachat/jsonrpc-client'
```

Simon-Laux:
- move more of select account to frontend
- remove unused close function
- remove unused ready event and var
- remove unused get account dir
- inline functions
- remove more unused stuff
- avoid using selected account context in webxdc-icon
- fix eslint
- remove usage of selected account(Id) in webxdc
- rm selectedAccountContext and selectedAccountId
- remove now unused login.ts
- remove DeltaBackend 🎆 🥳
- remove stray electron usage in renderer

closes #2938
closes #2076
closes #2981

Update: fixed the bug while closing DC


Simon-Laux:
> How does desktop remember the number of last used account? I think the point of "selected account" in the core is to store the last selected account when the app is restarted uniformly across platforms (in accounts .toml file).

we could store it there if it would be possible to unselect an account in core.

its stored with the other desktop settings in config.json, we could move the rest to core too, if the account manager would support custom ui config values

PR #2994, opened 2022-11-08
handle webxdc info messages differently than normal system messages

### LINKED ISSUE #2933: clicking on system message that is not from webxdc still jumps to previous message
this should not be the case, should it?
Is there even a way to know whether a system message is from a webxdc or not?

### LINKED ISSUE #2989: webxdc messages are not different to other system messages, which allows faking system messages
solution: make them visually distinct

for this desktop needs to know wether a system message is a Webxdc Info message, #2933 needs to know that too, so might make sense to solve both in one go.

Simon-Laux:
- don't jump to parent on non webxdc messages closes #2933
- different style for webxdc info messages + show webxdc icon closes #2989
- default info style, only one icon and adjust padding

### DIFF
--- a/scss/message/_message.scss
+++ b/scss/message/_message.scss
@@ -441,13 +441,11 @@
   }
 
   &.webxdc-info > .bubble {
-    border-radius: 20px;
-    background-color: rgb(0 0 0 / 29%);
-    padding: 7px 7px;
+    padding: 7px 8px;
     img {
       height: 18px;
       width: 18px;
-      margin: 0px 5px;
+      margin-right: 5px;
       border-radius: 7%; // webxdc border radius
       vertical-align: bottom;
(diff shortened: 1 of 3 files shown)


scss/message/_message.scss
```diff
@@ -440,6 +440,18 @@
     }
   }
 
+  &.webxdc-info > .bubble {
+    padding: 7px 8px;
+    img {
+      height: 18px;
+      width: 18px;
+      margin-right: 5px;
```
r10s, line 448: at some point, we will probably add rtl-support, then we should switch to `margin-inline` or so. just saying, no need to start with that now.

r10s (approved):
nice! did not test, just had a rough look at the code and to the images in dc :)

PR #3010, opened 2022-11-16
add internet access for experimental webxdcs, see #3005

### LINKED ISSUE #3005: take webxdc `internet_access` flag into account
UIs should check the field `internet_access` from [`dc_msg_get_webxdc_info()`](https://c.delta.chat/classdc__msg__t.html#aee526949983962ed25adf33a2f5e6878) - if that is true, internet access should be allowed for that webxdc, see https://github.com/deltachat/deltachat-core-rust/pull/3516 for discussions and reasoning.

respecting the flag is concretely needed to drive some experiments forward.

cmp. https://github.com/deltachat/deltachat-android/pull/2378, https://github.com/deltachat/deltachat-ios/issues/1744

Simon-Laux:
closes #3005

image


r10s (approved):
great, lgtm, thanks a lot, this is very helpful, cc @cracker0dks - desktop 1.34 will probably support `request_internet_access` with the known limitations.

i wouldn't have bothered with the additional warning (and maybe improve core hint instead) - now we need to adapt desktop if we change the conditions for internet_access ("Arbeit zieht Arbeit nach sich.", german proverb :) but I understand the intention, sure, so it's fine :)

r10s:
> that's the point, to make it harder to break our principles ;)

;)

PR #3023, opened 2022-11-21
fix jump to message from gallery

### LINKED ISSUE #3020: "show in chat" does not work
open a chat gallery with some images -> right tap an image -> select "show in chat"

-> nothing happens (1.33.2/mac)

same for audio/files/videos/apps

also the "show in chat" from "chat audit log" does not work.

Simon-Laux:
closes #3020

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Changed
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


r10s (approved):
i've seen @Simon-Laux  in person fixing that carefully! lgtm :)

PR #3031, opened 2022-11-28
prepare 1.34.1

Simon-Laux:


src/renderer/deviceMessages.ts
```diff
@@ -18,4 +18,14 @@ export function updateDeviceChats(accountId: number) {
 
 Full Changelog: https://github.com/deltachat/deltachat-desktop/blob/master/CHANGELOG.md#1_34_0`
   )
+
+  BackendRemote.rpc.addDeviceMessage(
+    accountId,
+    'changelog-version-1.34.1-version0',
+    `What's new in 1.34.1?
+
+✨ make some webxdcs work again
```
r10s, line 27:
```suggestion
✨ Make some Private Apps work again
```

"webxdc" are "Private Apps" in UI.

i think, the 1.34.1 device message could also be omitted (and say "What's new in 1.34" above). 1.34.0  was not widely released and for bugfix releases there are not always device messages afaik.

r10s (approved):

src/renderer/deviceMessages.ts
Simon-Laux: tried not to scare apple, but sure probably makes sense to omit the message

CHANGELOG.md
```diff
@@ -6,6 +6,12 @@
...
 ### Changed
 
+### Fixed
+
+<a id="1_34_1"></a>
+
+## [1.34.1] - 2022-11-28
+
 ### Fixed
 - fix webxdc: allow `self` and `blob:` in `connect-src` in CSP
```
r10s, line 16: imu, the changelogs are written for the end users, we also point from the device message there. so, we should use a much less technical language here. saying what it actually fixes for the user. for a technical view, we have the commit messages.

this is no blocker, more a general note. i also wanted to point that out before, and never came to it, and this line is just a good example :)

r10s (approved):

CHANGELOG.md
Simon-Laux: so "fix some private apps", no not a good idea to remove all info, these entries are interesting for webxdc devs and they should know about CSP, other people can just ignore what they don't understand, it is a log not a marketing document.

PR #3043, opened 2022-12-06
make font-style of info messages consistent with iOS and Android

### LINKED ISSUE #3034: font-style of info messages
from nov22 consistency discussions:

currently, info messages are always styled in the same italic style.

to get inner consistency with android/ios, and outer consistency with telegram/whatsapp/signal, this should be changed to:
- regular for normal info messages
- bold regular for date headlines

the addition italic is a bit too much, style-wise. the info is already outstanding by a different color and background.

Simon-Laux:
closes #3034

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 ### Changed
+- make font-style of info messages consistent with iOS and Android #3034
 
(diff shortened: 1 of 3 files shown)


r10s (approved):
great you push that forward!

Issue #3039, opened 2022-12-02
Message with ID 123456 was not found in cache


dumblob:
Yesterday I switched to a Deltachat Desktop window (I do it once in couple of hours usually) and got two messages:

>Message with ID 123456 was not found in cache. This should not happen. Please contact developers.

(but funnily with all first letters capital :wink:)

But I could not see anything wrong at the first sight so I made a screenshot and then pressed PageDown to see the new messages and it showed some. Then I pressed PageUp and I could not see the errors any more (they disappeared) but instead two new messages (first being a picture, second being text) appeared.

![20221201_173827+0100-1680x1050+0+0-imlib2_grab_s](https://user-images.githubusercontent.com/2451044/205283704-2760ffea-6531-41bd-8cf1-0f7373524fa8.png)

I do not know what to do :shrug:.

[2022-11-28-11-34-01.log](https://github.com/deltachat/deltachat-desktop/files/10140676/2022-11-28-11-34-01.log)

Simon-Laux:
Its a known problem, though I don't know how to reproduce it yet

dumblob:
Ah, ok. Feel free to close this if there is an opened duplicate bug somewhere (and ideally link it to here).

PR #3053, opened 2022-12-14
friendlier error message for #3039

Simon-Laux:


Simon-Laux:
#3058 is the better solution

PR #3058, opened 2022-12-21
fix "message not found in cache" bug

### LINKED ISSUE #3039: Message with ID 123456 was not found in cache

Simon-Laux:
closes #3039

The fix consists in calling a method to load all missing messages upon rendering a “message not found in cache” message in react, would not call it clean, but it is reliable in my tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - set `ITSAppUsesNonExemptEncryption` to `false` for mac
 
 ### Fixed
+- fix "message not found in cache" bug #3039
 - fix webxdc: allow `self` and `blob:` in `connect-src` in CSP
 - indentation in update device message
 - fix "no messages" blinking up for a second when loading a chat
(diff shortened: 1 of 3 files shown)


r10s:
\o/

dumblob:
Thanks! Will test once released :wink:.

PR #3069, opened 2023-01-07
only fetch the entry of chatlist that we need, not the whole list

Simon-Laux:
depends on https://github.com/deltachat/deltachat-core-rust/pull/3929

In my subjective testing this helped reducing message replay time (after backup) import on desktop by 13-19%. backup import is a good stress test for this because it calls this code-path often.

## TODO
- [ ] maybe add tests?


Simon-Laux:
superseded by #3071

PR #3091, opened 2023-01-19
Ask for name/photo for accounts created through QR

### LINKED ISSUE #3036: let user chose name+avatar after setup account via qr code scan
from nov22 consistency discussions:

setting up at least displayname is essential for the usability of random accounts that are scanned via qr code - otherwise, the contacts have no idea what the person is.

android/ios directly ask for name/avatar after scanning an account,
this part is missing on desktop.

maxphilippov:
* new property on window __askForName if it is set to true first time the MainScreen opens the SettingsProfileDialog pops up for user to change their name, add photo
* SettingsProfileDialog accepts custom title, confirmLabel, cancelLabel and passes it down

fixes #3036

### DIFF
--- a/src/renderer/components/dialogs/Settings-Profile.tsx
+++ b/src/renderer/components/dialogs/Settings-Profile.tsx
@@ -291,7 +291,7 @@ export function SettingsProfileDialog({
   isOpen: DialogProps['isOpen']
   onClose: DialogProps['onClose']
   settingsStore: SettingsStoreState
-  title?: DialogProps['title']
+  title?: string
   cancelLabel?: string
   confirmLabel?: string
   simpleSetup?: boolean
--- a/src/renderer/components/dialogs/Settings-Profile.tsx
+++ b/src/renderer/components/dialogs/Settings-Profile.tsx
@@ -123,12 +123,14 @@ export function ProfileImageSelector({
   color,
   profilePicture,
   setProfilePicture,
+  hideDeleteButton = false,
 }: {
   displayName: string
   addr: string
   color: string
   profilePicture: string | null
   setProfilePicture: (path: string) => void
+  hideDeleteButton?: boolean
 }) {
   const tx = window.static_translate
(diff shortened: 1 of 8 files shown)


src/renderer/components/dialogs/Settings-Profile.tsx
```diff
@@ -240,21 +251,30 @@ export function SettingsEditProfileDialogInner({
...
-            onChange={(
-              event: React.FormEvent<HTMLElement> &
-                React.ChangeEvent<HTMLTextAreaElement>
-            ) => {
-              setSelfstatus(event.target.value)
-            }}
-          />
+          {simpleSetup ? (
+            <span>You can change that later in the settings</span>
+          ) : (
```
Simon-Laux, line 256: I would just not show the signature field at all in this case, otherwise we would need a really wisely worded string that can be understood by normal users. Here the context is just missing to the user and I don't think it's worth to try to explain it without causing confusion.

src/renderer/components/dialogs/Settings-Profile.tsx
```diff
@@ -167,14 +169,16 @@ export function ProfileImageSelector({
...
+        {!hideDeleteButton ? (
+          <button
+            aria-label={tx('profile_image_delete')}
+            onClick={onClickRemovePicture}
+            className={'delta-button-round'}
+            disabled={!profilePicture}
+          >
+            {tx('profile_image_delete')}
+          </button>
+        ) : null}
```
Simon-Laux, line 181: users might reconsider their choice so I would still offer the option to delete the profile image again. sure the two buttons are not that pretty, but I think improving that is something for later.

Simon-Laux (changes requested):

Simon-Laux (approved):

Simon-Laux:
changelog entry would be nice, I'm not sure why the ci passes without it...

PR #3094, opened 2023-01-24
Remove incorrect placeholder for port

link2xt:
It shows that IMAP port is going to be 993 by default, but actually the core tries multiple ports,
may get autoconfiguration XML from the server
or get the port from the provider database.

Show the string "Default" without incorrect port value instead.


Simon-Laux:
I think this deserves a small changlog entry

PR #3096, opened 2023-01-29
bump electron to 22.1.0

dotlambda:


src/main/windows/main.ts
```diff
@@ -37,7 +37,7 @@ export function init(options: { hidden: boolean }) {
       x: initialBounds.x,
       y: initialBounds.y,
       webPreferences: {
-        nodeIntegration: false,
```
Simon-Laux, line 40: If we bump electron then not with using this, rather doing the work that is required than weakening security to make it work

Simon-Laux:
I did it properly without enabling node-integration in #3097

PR #3110, opened 2023-02-02
feat: show full message in window

### LINKED ISSUE #3056: "Show full message in browser" doesn't work - "Firefox can’t find the file at /tmp/deltachat-8DEtDk/message.html"
- Operating System (Linux/Mac/Windows/iOS/Android): Xubuntu 22.04
- Delta Chat Version: 1.34.0
- Expected behavior: When I receive a message which is too long to be displayed in one message bubble, and I click on "show full message in browser, it's shown in my system browser (Firefox in this case). Alternatively it could be shown in a webview, but I expect it to be shown.
- Actual behavior: A browser tab with the following error is shown: "Firefox can’t find the file at /tmp/deltachat-8DEtDk/message.html." The file does exist though.
  - receive a large mail
  - click on "show full message in browser"

Simon-Laux:
Open HTML message in dedicated window instead of inside of the browser.

| no remote content | confirmation| with remote content|
|-------------------|--------------|----------------|
| Bildschirmfoto 2023-02-03 um 22 12 31 |Bildschirmfoto 2023-02-03 um 22 12 37|Bildschirmfoto 2023-02-03 um 22 12 44|

this will also close #3056,
because with this pr, there is no "open in browser" anymore.

## TODO:
- [X] use translation system and use the string bjoern mentioned.
- [X] checkbox for remote content with dialog on click
- [X] better design of the header
- [x] respect theming (should we keep white as default background for the email though?)
- [x] always handle mailto links in deltachat

### DIFF
--- a/src/main/windows/html_email.ts
+++ b/src/main/windows/html_email.ts
@@ -128,6 +128,10 @@ export function openHtmlEmailWindow(
           message: tx('load_remote_content_ask'),
           checkboxLabel: tx('do_not_ask_again'),
           buttons: [tx('no'), tx('yes')],
+          type: 'none',
+          icon: '',
+          defaultId: 0,
+          cancelId: 0
         })
         if (result.response === 0) {
           throw new Error('user denied')
--- a/src/main/windows/html_email.ts
+++ b/src/main/windows/html_email.ts
@@ -120,13 +120,22 @@ export function openHtmlEmailWindow(
   window.webContents.ipc.handle(
     'html-view:change-network',
     async (_ev, allow_network: boolean) => {
-      if (allow_network) {
+      if (
+        allow_network &&
+        DesktopSettings.state.HTMLEmailAskForRemoteLoadingConfirmation
+      ) {
         const result = await dialog.showMessageBox(window, {
           message: tx('load_remote_content_ask'),
+          checkboxLabel: tx('do_not_ask_again'),
           buttons: [tx('no'), tx('yes')],
         })
-        if(result.response === 0) {
-          throw new Error("user denied");
+        if (result.response === 0) {
+          throw new Error('user denied')
+        }
+        if (result.checkboxChecked) {
+          DesktopSettings.update({
+            HTMLEmailAskForRemoteLoadingConfirmation: false,
+          })
         }
       }
 
--- a/src/main/windows/html_email.ts
+++ b/src/main/windows/html_email.ts
@@ -11,6 +11,7 @@ import { join } from 'path'
 import { DesktopSettings } from '../desktop_settings'
 import { truncateText } from '../../shared/util'
 import { tx } from '../load-translations'
+import { open_url } from '../open_url'
 
 const open_windows: { [window_id: string]: BrowserWindow } = {}
 
@@ -175,8 +176,13 @@ function makeBrowserView(allow_remote_content: boolean, html_content: string) {
     (e: electron.Event, url: string) => {
       // Prevent drag-and-drop from navigating the Electron window, which can happen
       // before our drag-and-drop handlers have been initialized.
+      // Also handle clicking links inside of the message.
       e.preventDefault()
-      shell.openExternal(url)
+      if (url.startsWith('mailto:')) {
+        open_url(url)
+      } else {
+        shell.openExternal(url)
+      }
     }
   )
 
--- a/src/main/windows/html_email.ts
+++ b/src/main/windows/html_email.ts
@@ -131,7 +131,7 @@ export function openHtmlEmailWindow(
           type: 'none',
           icon: '',
           defaultId: 0,
-          cancelId: 0
+          cancelId: 0,
         })
         if (result.response === 0) {
           throw new Error('user denied')
--- a/src/main/windows/html_email.ts
+++ b/src/main/windows/html_email.ts
@@ -1,7 +1,9 @@
 import electron, {
   BrowserView,
   BrowserWindow,
+  dialog,
   MenuItemConstructorOptions,
+  nativeTheme,
   session,
   shell,
   WebContents,
@@ -12,6 +14,7 @@ import { DesktopSettings } from '../desktop_settings'
 import { truncateText } from '../../shared/util'
 import { tx } from '../load-translations'
 import { open_url } from '../open_url'
+import { loadTheme } from '../themes'
 
 const open_windows: { [window_id: string]: BrowserWindow } = {}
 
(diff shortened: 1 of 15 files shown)


missytake:
Indeed, it fixes #3056 for me!

Any idea why testcafe is unhappy?

```
ERROR Unable to open the "electron:." browser due to the following error:

Error: Unable to connect
    at NodeDebug._callee2$ (/home/runner/work/deltachat-desktop/deltachat-desktop/node_modules/testcafe-browser-provider-electron/lib/node-debug.js:158:39)
    at tryCatch (/home/runner/work/deltachat-desktop/deltachat-desktop/node_modules/babel-runtime/node_modules/regenerator-runtime/runtime.js:62:40)
    at Generator.invoke [as _invoke] (/home/runner/work/deltachat-desktop/deltachat-desktop/node_modules/babel-runtime/node_modules/regenerator-runtime/runtime.js:296:22)
    at Generator.prototype.<computed> [as next] (/home/runner/work/deltachat-desktop/deltachat-desktop/node_modules/babel-runtime/node_modules/regenerator-runtime/runtime.js:114:21)
    at step (/home/runner/work/deltachat-desktop/deltachat-desktop/node_modules/babel-runtime/helpers/asyncToGenerator.js:17:30)
    at /home/runner/work/deltachat-desktop/deltachat-desktop/node_modules/babel-runtime/helpers/asyncToGenerator.js:28:13

Type "testcafe -h" for help.
child_process.js:674
    throw err;
    ^

Error: Command failed: xvfb-run --server-args="-screen 0 1280x720x24" npx testcafe --skip-js-errors electron:. ./test/testcafe/test.js ./test/testcafe/messagelist_tests.ts
    at checkExecSyncError (child_process.js:635:11)
    at execSync (child_process.js:671:15)
    at Object.<anonymous> (/home/runner/work/_actions/DevExpress/testcafe-action/latest/index.js:64:1)
    at Module._compile (internal/modules/cjs/loader.js:999:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
    at Module.load (internal/modules/cjs/loader.js:863:32)
    at Function.Module._load (internal/modules/cjs/loader.js:708:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:60:12)
    at internal/main/run_main_module.js:17:47 {
  status: 1,
  signal: null,
  output: [ null, null, null ],
  pid: 2038,
  stdout: null,
  stderr: null
}
```

r10s:
great!

please use the existing strings, so "Load remote images" (string id `load_remote_content` :) as a checkbox or switch. this is more user friendly, there were tons of discussions about that - in fact, also android wording was "Load remote content" initially, [cmp](https://github.com/deltachat/deltachat-android/pull/1763). also this is better for consistency.

before enabling, show a dialog with a warning, `load_remote_content_ask`, this dialog can also have the option `always`/`once` instead of just `yes`/`ok`.

for the css of the header as such: i assume this is a prototype currently and this will change anyway :)

r10s (approved):
looks good to me layout and functionality-wise, great job! 

i did not test nor can say much to code details, though.

r10s:
question to the dialog, no blocker: this is the electron native one, so there are no styling options, right? otherwise, of course, it would be better if, for inner consistency, it fits to the style of the app in general. and: can we remove the electron icon? 

in general, here and in corresponding chats, there seems to be some consent, that this is a "pretty nice" dialog style, and i also agree with that. wondering if should pick up that style, that seems to be somehow settled, also for other dialogs? maybe there are already defaults for that? but, sure, this is another thing and unrelated to this pr - just a question for now :)

Simon-Laux:
> this is the electron native one, so there are no styling options, right? 

yes, no styling options

> can we remove the electron icon?

in the build it will be the deltachat icon.


As stated above it's difficult to show a custom dialog here, because it would be displayed under the email, or we need to add another browser view on top of the one for the email (wasting ram and its not the easiest thing to understand code wise.


> wondering if should pick up that style, that seems to be somehow settled, also for other dialogs? maybe there are already defaults for that?

sure we could use the native dialogs more often, but they also have several downsides (they are very limited in what they can do, look different on every OS and we have to go though electron each time, making a tauri/browser version harder, because it would increase dependency on electron, which we don't really want)

For copying the style, maybe there are some good aspects that we can copy, but in general I don't really like it as it is (looks only good on macOS). We need our own consistent dialog design anyway, also refactor the dialog system to add really good keybindings and fix the "pressing escape does not what it should" -bugs.

src/main/windows/html_email.ts
```diff
@@ -0,0 +1,295 @@
...
+    'html-view:change-network',
+    async (_ev, allow_network: boolean) => {
+      if (
+        allow_network &&
+        DesktopSettings.state.HTMLEmailAskForRemoteLoadingConfirmation
+      ) {
+        const result = await dialog.showMessageBox(window, {
+          message: tx('load_remote_content_ask'),
+          checkboxLabel: tx('do_not_ask_again'),
+          buttons: [tx('no'), tx('yes')],
```
r10s, line 130:
```suggestion
          buttons: [tx('no'), tx('yes')],
          type: "none",
          icon: null,
```

that should remove the icon according to https://www.electronjs.org/docs/latest/api/dialog and https://github.com/electron/electron/issues/13250 - did not try that out, however.

r10s: thanks for trying in https://github.com/deltachat/deltachat-desktop/pull/3110/commits/65758d2bb7844de158377b324dbae0a95bfd98e3 - then the icon is there, seems to be an electron bug.  was worth a try.

missytake (approved):
Works for me :) could be prettier, but don't let perfect get in the way of better.

src/main/windows/html_email.ts
```diff
@@ -0,0 +1,299 @@
...
+      // for debugging email
+      // sandboxedView.webContents.openDevTools({ mode: 'detach' })
+    }
+  )
+
+  // for debugging wrapper
+  // window.webContents.openDevTools({ mode: 'detach' })
+}
+
+const CSP_DENY =
```
bkil, line 164: Why do you need `blob:` without enabling JavaScript? What do you think about this?
* `CSP_DENY = "default-src 'none'; font-src 'self' data:; frame-src 'self' data:; img-src 'self' data:; media-src 'self' data:; style-src 'self' data: 'unsafe-inline'; form-action 'none'; navigate-to 'none';"`
* `CSP_ALLOW = "default-src 'none'; font-src 'self' data: http: https:; frame-src 'self' data: http: https:; img-src 'self' data: http: https:; media-src 'self' data: http: https:; style-src 'self' data: 'unsafe-inline' http: https:; form-action 'none'; navigate-to 'none';`

Simon-Laux: good catch, yeah we probably don't need it the `blob:` uri. also we don't want scripts in emails at all.

Simon-Laux:
Added an option to always load remote content of html email:
Bildschirmfoto 2023-04-07 um 19 44 11
Bildschirmfoto 2023-04-07 um 19 03 04

src/main/windows/html_email.ts
Simon-Laux: - `style-src` not many email clients support remote styles, even support for the style tag is not high, I saw many recommendations of inline styles so I wouldn't allow them.
- `frame-src` - frames open security holes, imagine some kind of nested frame trickery that has javascript enabled again (I won't trust chromium here, also frames in emails sounds like a supper bad idea in general)
- `navigate-to` csp checker site says it is an invalid option

PR #3115, opened 2023-02-07
remove dependency on `glob-watcher`

Simon-Laux:
this removes the dependency on an old version of chokidar and uses the new chokidar directly instead now (which has less dependecies and is faster)

CC @dotlambda 

closes  #2771


Simon-Laux:
I'll merge this now as it only affects development stuff (npm run watch:* scripts) and I tested it and it worked.

PR #3124, opened 2023-02-15
top evil blueprint

Simon-Laux:
The new game is to remove our dependency blueprintjs, `./bin/topEvil.sh` is our trusty companion on this 
endevour: it show us where blueprint is still used.

```
🤔Result🤔
CODE:       33 | CLASSES:       35 | SCSS:      119
```

The goal is to get all 3 numbers down, idealy down to 0, bit by bit.

- change top evil goal to blueprintjs
- top evil summary


dotlambda:
What's the rationale for getting rid of blueprint?

Simon-Laux:
we override it anyway in most places, also deps are old, updating it could be much work. Also lookup the company that makes blueprintjs, their business model does not align with our project values.

PR #3134, opened 2023-02-27
remove blueprintjs

Simon-Laux:
before
CODE:       33 | CLASSES:       35 | SCSS:      119
after:
CODE:       27 | CLASSES:       32 | SCSS:      118


Simon-Laux:
this would be a bigger effort, I think for now we should just avoid blueprints when adding new features or refactoring existing features.

farooqkz:
Avoiding blueprintjs? And what alternative are you going to use?

Simon-Laux:
we override it at many places anyway with our own styles, we can just use our own styles everywhere.

farooqkz:
How about another UI library? Or our own UI library?

Simon-Laux:
I think in the end it makes sense to make an own library. depending on what we need there could also be a good library that we can use, but we should investigate where it comes from first and it should not have so many dependencies.
But better discuss this somewhere else like in the desktop chat

PR #3144, opened 2023-03-04
update electron to version `v22.3.2`

Simon-Laux:


dotlambda:
Why didn't dependabot suggest this update?

PR #3150, opened 2023-03-20
networked multidevice setup

Simon-Laux:
- [x] receiving
- [X] sending
- [x] remove timeout hack
- [x] translation keys
- [X] troubleshooting link
- [x] adjust that it looks like on android
   - [ ] ~~receive~~ postponed, is good enough for now
   - [X] send
- [x] set new core string
- [x] some kind of success feedback on sending 

depends on new core with these prs merged:
- https://github.com/deltachat/deltachat-core-rust/pull/4007
- https://github.com/deltachat/deltachat-core-rust/pull/4198

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased][unreleased]
 
 ### Added
+- networked setup multidevice
 ### Changed
 - update electron from `v22.1.0` to version `v22.3.2`
 - use the new go based transifex cli (old one stopped working)
--- a/images/icons/devices.svg
+++ b/images/icons/devices.svg
@@ -1,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   height="24"
+   width="24"
+   version="1.1"
+   id="svg826"
+   sodipodi:docname="devices.svg"
+   inkscape:version="1.1.1 (c3084ef, 2021-09-22)"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:svg="http://www.w3.org/2000/svg">
+  <defs
+     id="defs830" />
+  <sodipodi:namedview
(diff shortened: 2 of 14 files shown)


Simon-Laux:
problem is that there is a progress-bar / it takes a while until the qr code is returned by core

hpk42:
"hash mismatch" error with Android 1.36.0 scanning a QR procuced from this branch + core stable-1.112 as of 20 minutes ago 
image
i don't get this hash mismatch when using two Anroid 1.36.0's devices.

hpk42:
just to note down here as well: it seems that my original problem 
can be traced back to an iroh issue with 30-40K files in the collection -- 
@flub was able to quite reliably reproduce it in a test.

meganoahj:
So I tried using the repl tool to test out the new transfer backup locally function from and to Windows, which only worked after disabling the windows firewall manually (could be because Delta Chat is an untrusted app and Windows blocks every non standard port or protocol?) after some research I found out that there are multiple ways to fix this issue.

I tried it on windows 11 but it should be the same from windows 7 and up because they all still use nt firewall.

The simplest fix at least in my perspective would be using batch file creating firewall rule (could be integrated into installer) see here: https://www.itninja.com/blog/view/allow-an-app-through-firewall

I created an example batch file (needs admin access) which should work on every supported winver because it uses nt commands:

dcfix.bat
`netsh advfirewall firewall add rule name="DC in" profile=domain,private,public protocol=any enable=yes DIR=In program="C:\Users\Noah\Projects\deltachat-core-rust\target\debug\deltachat-repl.exe" Action=Allow
netsh advfirewall firewall add rule name="DC out" profile=domain,private,public protocol=any enable=yes DIR=Out program="C:\Users\Noah\Projects\deltachat-core-rust\target\debug\deltachat-repl.exe" Action=Allow`


Other ways could be using a firewall api to create the rules (prob have to rewrite because of way too much code or the smaller one also using c++)

https://github.com/microsoft/windows-rs (Official Microsoft Rust API)
https://github.com/marirs/winfw-rs (Smaller Bootleg firewall only API)

flub:
Directly creating holes in the windows firewall using their API sounds interesting.  Is there any documentation from microsoft on what you're allowed to do there?  Are there any permissions to deal with?

PR #3158, opened 2023-04-03
fix #3154 also fix the bug of last chat of old account still visible in new account.

### LINKED ISSUE #3154: Burner account setup screen shows information of other account
when setting up a new account from a DC Account token it shows the dialog for setting it up, but instead of empty fields the fields contain information of the previously used account.

This appears when scanning the account creation code from the welcome screen and from the other account.

Might be cause by wrong caching of current account id?

steps to reproduce: switch account, then scan in welcome screen, somehow it still shows info of old account

Simon-Laux:
fix #3154

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 - mac: update tray icon menu on hiding/showing #3041
 - chat request deletion now asks for confirmation
 - fix updating of relative timestamps #3125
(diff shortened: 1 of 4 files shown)


missytake (approved):
Looks good as far as I can tell with my limited javascript/typescript skills :) Just encountered the #3154 bug today, thanks for fixing.

Issue #3163, opened 2023-04-05
update offline help, re-pull translations


r10s:
(not sure of there if an automatism, in this case, please just close this issue)

the offline help was not updated since some time, eg. there is the section about webxdc missing.

also, in the last some weeks, lots of things were updated, not only wrt "Add Second Device".

for android/ios, we use the script [`deltachat-pages/tools/create-local-help.py`](https://github.com/deltachat/deltachat-pages/blob/master/tools/create-local-help.py), not sure if desktop used this as well.

moreover, in the last few days, lots of new strings got translations, so also re-pulling the normal app strings are worthful.

so, if possible, it would be great to have an update for both things

Simon-Laux:
I pulled the translation strings earlier today, but seems like I forgot to upload the branch?

r10s:
> I pulled the translation strings earlier today, but seems like I forgot to upload the branch?

no, you did things right, i have overseen that :) thanks!

Simon-Laux:
> no, you did things right, i have overseen that :) thanks!

I actually forgot to make the pr, only the branch was uploaded. I did it directly  after posting my last comment ;)

so help and translations are updated now, so I close this issue now

PR #3164, opened 2023-04-05
update offline help

r10s:
targets #3163


Simon-Laux (approved):
lgtm, does the script need a shebang?

r10s:
> lgtm, does the script need a shebang?

there is really nothing special in the script, it should run on any shell, no requirements

PR #3167, opened 2023-04-08
introducing chatlist item thumbnail image

meganoahj:
### Added chatlistitem preview! now working with stickers (also animated ones), images, and webxdc

Also added xdc icon preview
delta xdc icon

needs core with pr merged
https://github.com/deltachat/deltachat-core-rust/pull/4311

### DIFF
--- a/scss/chat/_chat-list-item.scss
+++ b/scss/chat/_chat-list-item.scss
@@ -63,6 +63,8 @@
       margin-top: 3px;
 
       & > .text {
+        display: flex;
+        align-items: center;
         max-width: calc(100% - 18px);
         height: 1.3em;
         flex-grow: 1;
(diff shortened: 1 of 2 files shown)


scss/chat/_chat-list-item.scss
```diff
@@ -75,6 +77,14 @@
         color: var(--chat-list-item-summary-text);
         user-select: none;
 
+        & > img.summary_thumbnail {
+          height: 18px;
+          width: 18px;
```
Simon-Laux, line 82:
```suggestion
          height: 18px;
          width: 18px;
          min-width: 18px;
          min-height: 18px;
```

adding these two helps that the thumbnails always have the same size.

But maybe we want to switch to background image and use background-cover and size to get a less stretched result, maybe a bit zoomed in.

https://developer.mozilla.org/en-US/docs/Web/CSS/background

Simon-Laux:
`npm run test` reports that the chatlistitem file still has formatting issues.

run `npm run fix-formatting` to fix it.

meganoahj:
> `npm run test` reports that the chatlistitem file still has formatting issues.
> 
> run `npm run fix-formatting` to fix it.

done
https://github.com/deltachat/deltachat-desktop/pull/3167/commits/bf7766131b45365dec2a458eceee95e446625549

PR #3175, opened 2023-04-10
improve confirm abort dialog for setup second device

### LINKED ISSUE #3173: add-second-device abort-dialog trivia
- the dialog "Abort setting up second device?" should not be shown when the initial confirmation dialog "This creates a QR code ..." is canceled

- the additional hint "This will invalidate the QR code copied to the clipboard" is meant to be added to the abort dialog only when really sth. was copied to the clipboard - otherwise, the information is more irritating than helping and does not make much sense wording wise.

apart from that - i just managed to transfer an account from desktop to android - works great!

Simon-Laux:
fixes #3173

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased][unreleased]
 
+### Changed
+- improve confirm abort dialog for setup second device #3173
+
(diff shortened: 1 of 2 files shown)


r10s (approved):
lgtm

PR #3177, opened 2023-04-11
update offline help

r10s:
this also pulls in some fixes of broken links

nothing urgent, however, can go to 1.36.2 or later


Simon-Laux:
we should mention this in the changelog

PR #3179, opened 2023-04-11
Hardening: prohibit all DNS queries

WofWca:
(except for Mapbox)


WofWca:
I have tried setting just `host-resolver-rules` (not `host-rules`), assuming that fetching by IP would still work (I have also disabled CSP for testing), but it does not. I get `net::ERR_NAME_NOT_RESOLVED`. So, I'm not sure how exactly these switches work.

Simon-Laux (approved):

PR #3193, opened 2023-04-17
remember html email window positioning and size.

### LINKED ISSUE #3191: Posibility for thee window size be remembered in html viewer !!

Simon-Laux:
closes #3191


r10s:
ftr, i just tried out that the window is not opened outside the screen when the position was saved eg. on a larger screen - looks good, electron seems to take care of that

PR #3198, opened 2023-04-19
use core for load remote content in html emails

Simon-Laux:
depends on https://github.com/deltachat/deltachat-core-rust/commit/37503dd3e838bfcdcfd784dce705b169700e0066


WofWca:
It's also a fix for remote content not loading, which was introduced in #3179

PR #3206, opened 2023-04-21
fix: multi device: translation typo transfeRring

WofWca:
And internal variable value

I'm not sure if anything translation-related needs to be updated as well. I have not find this string in the `_locales` directory.


Simon-Laux:
it's only used internally, I was to lazy to make a typescript enum

PR #3230, opened 2023-05-18
update translations

Simon-Laux:
and add `română` to language selection (18.05.2023)


_locales/_languages.json
```diff
@@ -29,6 +29,7 @@
   "pl": "Polski",
   "pt": "Português",
   "pt_BR": "Português do Brasil",
+  "ro": "română",
```
r10s, line 32:
```suggestion
  "ro": "Română",
```

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 - update deltachat-node and deltachat/jsonrpc-client to `v1.115.0`
 - remove message id from chatlistitem
 - Improved message search results
+- Update translations (18.05.2023)
+- add `română` language
```
r10s, line 11:
```suggestion
- add Romanian translation
```

(the changelog is english, in the UI, however, we show the local name)

r10s (approved):

_locales/_languages.json
Simon-Laux: not sure how Romanian does capitalisation, but I guess you did the research

r10s: assuming that proper nouns and language names at the beginning of a phrase are written in upper case is a good and reasonable default. exceptions can be handled then :)

but yes, i am also looking is used on comparable lists - and if questions pop up, i ask the translator who is doing the translation.

PR #3232, opened 2023-05-21
fix(css): remove `display: flex` from chatlist item text

link2xt:
Otherwise `<b>` highlight in the search result
does not have the space separating it from the rest of the text.

Bug screenshot:


Simon-Laux:
Bildschirmfoto 2023-05-21 um 16 14 39

fix does not work

Issue #3231, opened 2023-05-20
superfluous information shown in "Saved Messages" profile


r10s:
the following information are not useful but clutter UI:

- "Last Seen" should be hidden - it is always "Unknown", but even showing the last message sent or so does not make sense as the affected person is just seeing themselve in person when accessing the profile :) just hide that option, we just did that on ios cmp. https://github.com/deltachat/deltachat-ios/pull/1881 , android was never showing "Last Seen" for "Saved Messages"

- "Shared Chats" does not make sense as well - these are more or less all chats, so this is basically a copy of the main chatlist. "Shared Chats" for "Saved Messages" are hidden on ios/android as well.

<img width=300 src=https://github.com/deltachat/deltachat-desktop/assets/9800740/1fa7612f-7905-47c3-9dff-f04f2f09198a>

Simon-Laux:
Isn't the whole view profile dialog useless for the self contact?

r10s:
hm, indeed there are not many information left and "gallery" etc is not part of the profile as on android/ios. otoh,  it may be confusing and look like a bug when you cannot tap on the avatar or chat name for some chats, maybe still okay for being removed, not sure

Simon-Laux:
<img width="398" alt="Bildschirmfoto 2023-05-22 um 15 26 38" src="https://github.com/deltachat/deltachat-desktop/assets/18725968/080d8bb1-4c1b-4b57-8692-014abae9ddf2">

could also be an option... but honestly I'd just make it non clickable like telegram because clicking on the self contact and then getting a profile about "saved messages" is a bit confusing...

PR #3234, opened 2023-05-22
self contact now is displayed as saved messages chat in view profile

Simon-Laux:
See https://github.com/deltachat/deltachat-desktop/issues/3231#issuecomment-1557225606

~~The image might be harder to get, because is the saved messages chat avatar and not the avatar of the self contact.~~
See comments beneath r10s had a trick to get it, works, but not the prettiest or most efficient code, I know some alternative routes but they are not prettier, just a bit more efficient (runtime, not coding/maintenance) - like passing down the profile url in react properties, now that there is only one way to get to the dialog (since #3235) and that place already has the picture, but then one extra property or a cloned component just for saving 2 jsonrpc calls and 5 lines of code for this small edge case?

Bildschirmfoto 2023-05-30 um 21 40 36

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Improved message search results
 - Update translations (18.05.2023)
 - add Romanian translation
+- better profile view for saved messages chat
 
 ### Fixed
 - fix some emojis not getting larger in emoji only messages
--- a/src/renderer/components/dialogs/ViewProfile.tsx
+++ b/src/renderer/components/dialogs/ViewProfile.tsx
@@ -76,7 +76,8 @@ export default function ViewProfile(props: {
   const tx = window.static_translate
   const { openDialog } = useContext(ScreenContext)
   const [contact, setContact] = useState<Type.Contact>(props.contact)
-  const isDeviceMessage = contact.id === C.DC_CONTACT_ID_DEVICE
+  const isDeviceChat = contact.id === C.DC_CONTACT_ID_DEVICE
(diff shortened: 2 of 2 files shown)


r10s:
good idea to add `saved_messages_explain` to the dialog, this makes the whole thing much more useful!

however, we should use the "Saved Messages" icon, not the self-avatar here - this seems less confusing and also what android/ios are doing since we changed to the "Saved Messages" icon.

ftr, this is how it looks like (yikes! i am still able to build desktop on my machine!)

Screenshot 2023-05-28 at 20 06 59

Simon-Laux:
the question is how to get the saved messages icon from core... easiest way would probably be to duplicate it into the desktop repo, but then it could get out of sync (if we ever change the image).

another option would be to make a 3rd variant of the dialog (we have contact/view profile (device message also uses view profile) and group (group has internal variants for broadcast- and mailinglist)) that only handles saved messages and uses the chat profile image (so the image from core).

r10s:
> the question is how to get the saved messages icon from core

`dc_chat_get_profile_image()` on the saved-messages-chat should do

Simon-Laux:
> dc_chat_get_profile_image() on the saved-messages-chat should do

I have the self contact id, I don't have the chat id at that place. and afaik it is not a constant.

r10s:
> I have the self contact id, I don't have the chat id at that place. and afaik it is not a constant

no, it is not a constant, but you can get it unconditionally by `self_chat_id = dc_create_chat_by_contact_id(DC_CONTACT_ID_SELF)`

PR #3243, opened 2023-06-02
make sure, we always use the same path

r10s:
even if it seems unprobable that the temp dir ever changes during the function running,
it is better style not to mix `getDraftTempDir()` and `path` randomly and also underlines the gist better.

came over this during review of https://github.com/deltachat/deltachat-desktop/pull/3240


r10s:
thanks for the formatting and linting chore!

PR #3249, opened 2023-06-04
update timestamps and reconnect after resuming after sleep

Simon-Laux:
In my experience timestamps did not update reliable when resuming from standby before. Also sometimes it did not automatically reconnect after resuming after standby.

This pr listens for resume and unlock events and then updates timestamp and calls maybe_network.


Simon-Laux:
> Did you test that these events are called at all?

yes they are called, at least on MacOS.

PR #3251, opened 2023-06-04
fix e2e tests testcafe 2023

Simon-Laux:
- move testcafe stuff in own package with own node_modules folder
- update test cafe to newest version
- don't limit test speed
- update ci workflow (remove test cafe action)

Todo:
- [ ] fix the tests
- [ ] electron/testcafe freeze-crashes when clicking on "create new contact with email" item
   - new electron (25) does not fix the issue
   - downgrading testcafe back to the version we used before does not fix the issue (the electron adapter stayed at 0.0.19, because 0.0.18 does not work at all)
   - manual intervention: changing search and then clicking on it freezes too, but then closing dialog, then reopen search and clicking on it works without any freezes.
   - `"appArgs": ["--devmode"],` does not have any effects at all


Simon-Laux:
At this point I just think we should work on making the browser version work (only for testing, dev, nothing official).
Then we should try http://cypress.io/ which seems to be more mature than testcafe.

Simon-Laux:
I disabled the workflow in GitHub again

PR #3252, opened 2023-06-04
possible fix for the missing camera permission popup

Simon-Laux:
That apple review does not reject our app.


Simon-Laux:
after testing I'm not sure anymore if this changes anything it already worked fine before, at least in my tests

PR #3255, opened 2023-06-06
add global gallery / all-media view

### LINKED ISSUE #3161: Global Gallery
accessible from the sidebar. a gallery view with media from all chats. no chat/gallery/map tabs visible, instead of chat title and  avatar a heading "Global Gallery / All Media" and keeping the 3 dot menu but only if we have some options to show there, if not hide it too.

probably needs some kind of virtualisation (only load images that are shown on the screen.

also when doing this filtering / searching for file names in the document tab of the gallery seems worthwhile.

Simon-Laux:
todo:
- [x] use virtual grid for performance, we already use react-window so I would say try it with 
https://react-window.vercel.app/#/api/FixedSizeGrid
- [x] overlay with date or date range of first visible picture.
- [x] scroll to top when switching tabs
- [x] filtering / searching for file names in the document tab
    - [x] prettier search bar (follow theming)
    - [x] empty screen when there are no results
- [x] turn documents tab into a table
- [x] integrate https://github.com/deltachat/deltachat-desktop/pull/3360 into this pr
- [ ] optional: make audio elements prettier
- ~~lazy loading list content and reloading when it changes (like chatlist, and evens: deletion, incoming, changed (like for when webxdc title changes(loosely related to also #3412)))~~ first experiments failed, see #3446 to pick this up again
- ~~Gallery tabs integrated into navbar~~ we need to further discuss this, first experiments with it speak against it.

closes #3161

### DIFF
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -82,5 +82,11 @@
   },
   "group_related_chats": {
     "message": "Related Chats"
+  },
+  "gallery_image_aspect_ratio_keep": {
+    "message": "Image aspect ratio keep"
+  },
+  "gallery_image_aspect_ratio_grid": {
+    "message": "Image aspect ratio grid"
   }
 }
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -83,10 +83,10 @@
   "group_related_chats": {
     "message": "Related Chats"
   },
-  "gallery_image_aspect_ratio_keep": {
-    "message": "Image aspect ratio keep"
+  "aspect_ratio_grid": {
+    "message": "Aspect Ratio Grid"
   },
-  "gallery_image_aspect_ratio_grid": {
-    "message": "Image aspect ratio grid"
+  "square_grid": {
+    "message": "Square Grid"
   }
 }
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -88,5 +88,8 @@
   },
   "square_grid": {
     "message": "Square Grid"
+  },
+  "search_files": {
+    "message": "Search Files"
   }
 }
--- a/src/renderer/components/Gallery.tsx
+++ b/src/renderer/components/Gallery.tsx
@@ -69,13 +69,25 @@ export default class Gallery extends Component<
     }
   }
 
+  reset(){
+    this.setState({
+      id: 'images',
+      msgTypes: MediaTabs.images.values,
+      element: ImageAttachment,
+      mediaMessageIds: [],
+      mediaLoadResult: {},
+    })
+  }
+
   componentDidMount() {
     this.onSelect(this.state.id)
   }
 
   componentDidUpdate(prevProps: mediaProps) {
     if (this.props.chatId !== prevProps.chatId) {
-      this.onSelect(this.state.id)
+      // reset
+      this.reset()
+      this.onSelect('images')
     }
   }
 
@@ -150,7 +162,10 @@ export default class Gallery extends Component<
             })}
           </ul>
           <div className='bp4-tab-panel' role='tabpanel'>
-            <div className='gallery'>
+            <div
+              className='gallery'
+              key={this.state.msgTypes.join('.') + String(this.props.chatId)}
+            >
               <div className='item-container'>
                 {mediaMessageIds.length < 1 ? (
                   <p className='no-media-message'>{emptyTabMessage}</p>
--- a/src/renderer/components/Gallery.tsx
+++ b/src/renderer/components/Gallery.tsx
@@ -220,17 +220,13 @@ export default class Gallery extends Component<
 
     const filteredMediaMessageIds = mediaMessageIds.filter(id => {
       const result = mediaLoadResult[id]
-      if (
-        result.kind === 'message' &&
-        result.fileName?.indexOf(queryText) !== -1
-      ) {
-        return true
-      } else {
-        return false
-      }
+      return (
+        result.kind === 'message' && result.fileName?.indexOf(queryText) !== -1
+      )
     })
 
-    const showDateHeader = currentTab !== 'files' && currentTab !== 'webxdc_apps'
+    const showDateHeader =
+      currentTab !== 'files' && currentTab !== 'webxdc_apps'
 
     return (
       <div className='media-view'>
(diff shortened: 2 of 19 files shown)


r10s:
> scroll to top when switching tabs

there seems to be a bug in the existing gallery as well:
- open a media view with some dozens media in "Images" and "Files"
- scroll down "images" half the way
- switch to "files"
-> files is scrolled down as well even though never opened

Simon-Laux:
> there seems to be a bug in the existing gallery as well:

both problems are, it's the same gallery component

farooqkz:
> optional: make audio elements prettier

I'd say remove it from here and open another issue for it. Could be implemented by someone else other than Simon like me. Also in general, I tend to have multiple smaller issues rather than one big one if the smaller issues can be implemented/fixed independently and this is one of those cases.

src/renderer/stores/store.ts
Simon-Laux, line 1: removing listeners was broken before, I also fixed that in this pr

_locales/_untranslated_en.json
```diff
@@ -82,5 +82,11 @@
   },
   "group_related_chats": {
     "message": "Related Chats"
+  },
+  "gallery_image_aspect_ratio_keep": {
+    "message": "Image aspect ratio keep"
+  },
+  "gallery_image_aspect_ratio_grid": {
+    "message": "Image aspect ratio grid"
```
r10s, line 90: "Square Grid"

"Aspect Ratio Grid"

farooqkz:
Some fix required

image

farooqkz:
image


I believe the search file field is small. It doesn't fit with other components there in size. Also shouldn't be a search field for the apps as well?

src/renderer/components/Gallery.tsx
```diff
@@ -140,43 +254,149 @@ export default class Gallery extends Component<
...
-                        {error.msgId} {'->'} {error.error}
-                      </li>
-                    ))}
-                  </ul>
+          <div role='tabpanel'>
+            <div
+              className={`gallery gallery-image-object-fit_${
+                GalleryImageKeepAspectRatio ? 'contain' : 'cover'
+              }`}
+              key={this.state.msgTypes.join('.') + String(this.props.chatId)}
```
farooqkz, line 278: How about unpacking `msgType` in the beginning from `this.state`?

Simon-Laux:
> I believe the search file field is small. It doesn't fit with other components there in size

Might be, though it was ok for @r10s, so I think it's fine for now, we can always improve things later

src/renderer/components/Gallery.tsx
```diff
@@ -117,15 +186,60 @@ export default class Gallery extends Component<
...
+    const filteredMediaMessageIds = mediaMessageIds.filter(id => {
+      const result = mediaLoadResult[id]
+      if (
+        result.variant === 'message' &&
+        result.fileName?.indexOf(queryText) !== -1
+      ) {
+        return true
+      } else {
+        return false
+      }
```
farooqkz, line 229:
```suggestion
      const { variant, fileName } = mediaLoadResult[id]
      return variant === 'message' && fileName?.indexOf(queryText) !== -1
```

src/renderer/components/Gallery.tsx
```diff
@@ -140,43 +254,149 @@ export default class Gallery extends Component<
...
+          <div role='tabpanel'>
+            <div
+              className={`gallery gallery-image-object-fit_${
+                GalleryImageKeepAspectRatio ? 'contain' : 'cover'
+              }`}
+              key={this.state.msgTypes.join('.') + String(this.props.chatId)}
+            >
+              {mediaMessageIds.length < 1 && !loading && (
+                <div className='empty-screen'>
+                  {/* IDEA: when we have someone doing illustrations this would be a great place to add some */}
```
farooqkz, line 282: How about we open an issue for this and add "help wanted/needed" label?

src/renderer/components/Gallery.tsx
```diff
@@ -117,15 +186,60 @@ export default class Gallery extends Component<
...
+      GalleryImageKeepAspectRatio,
+    } = this.state
     const tx = window.static_translate // static because dynamic isn't too important here
     const emptyTabMessage = this.emptyTabMessage(id)
 
+    const filteredMediaMessageIds = mediaMessageIds.filter(id => {
+      const result = mediaLoadResult[id]
+      if (
+        result.variant === 'message' &&
+        result.fileName?.indexOf(queryText) !== -1
```
farooqkz, line 224: We should open an issue to find a more efficient algorithm for the checking.

farooqkz: And finding a good algorithm is not far away. I'm pretty sure an npm package already exists for it :)

src/renderer/components/Gallery.tsx
```diff
@@ -117,15 +186,60 @@ export default class Gallery extends Component<
...
   render() {
-    const { medias, id, errors } = this.state
+    const {
+      mediaMessageIds,
+      mediaLoadResult,
+      id,
+      loading,
+      queryText,
+      GalleryImageKeepAspectRatio,
+    } = this.state
```
farooqkz, line 216: How about unpacking `id`, too?

farooqkz (changes requested):

farooqkz:
I haven't reviewed changed files other than Gallery.

@Simon-Laux  Do you think it's a good idea you give me a week to see if I can remove BP parts?

Simon-Laux:
> @Simon-Laux Do you think it's a good idea you give me a week to see if I can remove BP parts?

not right now, we currently want to make a new test release and then a release, so a bigger refactoring is not good timing right now. Also only in this file or in the whole app? Generally I think we can just avoid using it in the future and remove it when we work on a component anyway.

src/renderer/components/Gallery.tsx
Simon-Laux: indexOf of is already one of the fastest methods... anyway this needs to be done in core, because currently this is one of the things that forces us to load all file informations from the DB (not just their message ids). If you have many images you might feel the delay when you go into the image tab, because currently it loads every message object from core, the chatlist on the other hand only loads visible chatlistentry objects from core. see also #3446

src/renderer/components/Gallery.tsx
Simon-Laux: would get closed immediately because we don't have an designer/illustrator, I think we can keep this comment for now.

PR #3260, opened 2023-06-09
Settings: Move Experimental Features into Advanced

### LINKED ISSUE #3256: Experimental Features 🧪should become an entry in "Advanced"
currently, the "Experimental Features" are pretty much in the face of the "normal" users, which is not the gist of these features, that, well, are experimental and may even do potential harm 😈 (we're at least saying that at several places and chats for our own insurance :)

(the idea of "Advanced" is already that the majority of users should never have the need to open the settings - this applies even more to "Experimental Features")

"Advanced / Experimental Features"  is also the place where they're found on android - and meanwhile also ios - so it is also better for communicating these features, eg. in the help (just got an complain about the inconsistency from a translator).

to be clear: this is not urgent in any way, i think, ppl will find the experiments also on desktop even when we're saying "go to advanced settings"  :)

Simon-Laux:
fixes #3256

and as a bonus it also moves the connectivity view down now:
Bildschirmfoto 2023-06-09 um 20 00 27

docs pr: https://github.com/webxdc/webxdc_docs/pull/57


r10s (approved):
nice!

PR #3263, opened 2023-06-11
only paste the file, if clipboard contains a file and text

Simon-Laux:


src/renderer/components/composer/Composer.tsx
```diff
@@ -195,6 +195,9 @@ const Composer = forwardRef<
     if (!e.clipboardData.files.length) {
       return
     }
+    // when there is a file then don't paste text
+    // https://github.com/deltachat/deltachat-desktop/issues/3261
+    e.preventDefault()
```
r10s, line 200: this prevents the `handlePaste` of the input textarea to be called, right?

isn't that missing as well in the case `if (file.path) { ddFileToDraft(file.path); return }` below?

if so, maybe better call ` preventDefault()` at the end of the event handler once, so all now and future cases are caught

r10s: > there is no benefit in calling it multiple times

but the the `preventDefault` in the try() block is probably also superfluous

r10s (approved):
lgtm, i'd have a look at the the maybe superfluous `preventDefault()`, but it would not harm anyways

src/renderer/components/composer/Composer.tsx
Simon-Laux: yeah it is, looks like I forgot to remove it

PR #3264, opened 2023-06-13
add dc as handler for xdc files

### LINKED ISSUE #2886: register for `.xdc` file extension then ask to which chat to attach as draft
register for .xdc file extension then ask to which chat to attach as draft.
kinda like mailto, we could probably reuse logic from there.

android: https://github.com/deltachat/deltachat-android/issues/2532
iOS: https://github.com/deltachat/deltachat-ios/issues/1733

Simon-Laux:
- file icon does not look very good IMO, maybe more a package or document look
  - only tested on mac (windows needs something different)
  - mac: does only work when the app is already open
  - current implementation reuses the internal api also used for webxdc.sendToChat

TODO:

- [ ] better file icon
- [x] Mac: fix that it also works on start, not only if DC is already open
- [x] windows (docs say needs to be parsed from `process.argv`)
- [x] linux (needs testing)
- [x] changelog entry

would close #2886 if done


Simon-Laux:
windows APPX works, but icon is DC icon:
Bildschirmfoto 2023-11-06 um 03 58 34
windows setup also works, including the icon:
Bildschirmfoto 2023-11-06 um 04 11 13

farooqkz:
I've tested it on Linux with i3. It prompts me to send the opened app to some chat. Is this how it is supposed to work?

Simon-Laux:
> I've tested it on Linux with i3. It prompts me to send the opened app to some chat. Is this how it is supposed to work?

yes

> But in Thunar(and using XFCE), the WebXDC file is still shown as ZIP archive:

how does the .desktop file of deltachat look like?

Simon-Laux:
So I investigated a bit and can reproduce what @farooqkz said above. The file extension is not registered in the system.
Looks like we need to look into adding a custom file to the installer packages: https://help.gnome.org/admin/system-admin-guide/stable/mime-types-custom.html.en

maybe electron builder does already support that, needs further investigation as the documentation is not really clear or easy to understand (for me atleast).  There is this closed issue hinting that it might be supported: https://github.com/electron-userland/electron-builder/issues/2872

PR #3271, opened 2023-06-16
add confirm dialog to webxdc send to chat closes #3269

### LINKED ISSUE #3269: `sendToChat()` should be cancelable at least on first screen?
the current implementation of `sendToChat()` does not allow to "cancel" the action and return to the app (as on android and [now also on ios](https://github.com/deltachat/deltachat-ios/pull/1904)).

this results in apps being terminated quite unexpectedly and undoable, which is esp. surprising if the user taps around to check functinalities or hits a "Share" button accidentally.

best option for desktop would be that the "forward/share" dialog appears floating atop of the app, however, that seems very hard to do with the current code. second best would be sth. as on android, maybe just a confirm() in the webxdc wrapper?

Simon-Laux:
Personally I'm still not convinced, IMO we should just tell app devs to implement a confirmation dialog themselves.
Bildschirmfoto 2023-06-16 um 13 19 51

translations need to be pulled before releasing this. also this is a system dialog, the electron icon will be the DC logo on Mac, on other OS the dialogs will look different.

closes #3269

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased][unreleased]
 
+### Changed
+- Added confirmation dialog to `webxdc.sendToChat` api #3269
+
 <a id="1_37_1"></a>
 
 ## [1.37.1] - 2023-06-14
--- a/src/main/deltachat/webxdc.ts
+++ b/src/main/deltachat/webxdc.ts
@@ -4,7 +4,7 @@ import SplitOut from './splitout'
 import { getLogger } from '../../shared/logger'
 const log = getLogger('main/deltachat/webxdc')
 import Mime from 'mime-types'
-import { Menu, ProtocolResponse, nativeImage, shell } from 'electron'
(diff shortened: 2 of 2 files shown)


r10s (approved):

r10s:
> Personally I'm still not convinced, IMO we should just tell app devs to implement a confirmation dialog themselves.

sharing may be very lightweight - and best UX would be if you tap share - and have the option to select a chat directly or hit "cancel" - if we say, app developers shall take care, we cannot reach this UX. also, we need to rely on app developers not making a mistake here :)

r10s:
thinking it over a bit and playing around: maybe we can just do nothing here and not merge this pr and close #3271 without action.

while the non-cancelling is not optimal, also the dialog is not - and @adbenitez made a point, when saying that it is harder to mess up with a mouse, also the apps usually do not take the full screen, so it is clearer what is going on. so, maybe no confirmation on _desktop_ is the better option, indeed (not better than having a "cancel" in the chatlist, but maybe better than the dialog)

so, i am +/-0 here meanwhile, thanks for doing the pr so that we have a choice now, @Simon-Laux!

PR #3278, opened 2023-06-21
fix: if systemPreferences.askForMediaAccess is not available, then don't call it

### LINKED ISSUE #3279: Scanning QR code makes an exception pop up (but things continue to work)
When i use master Desktop and go to qr-code scanning i get an exception but on second attempt it works. 
Not sure it needs to be considered a blocker, also good if someone else could verify this problem on a properly installed ".deb" package or so.  I am using master core/desktop from github. 

paste-2347861183

Simon-Laux:
(broke qr scan under linux (and maybe also under windows, was not tested))

fixes #3279

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased][unreleased]
 
+
+### Fixed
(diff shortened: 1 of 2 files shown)


hpk42 (approved):
yes, this fixes the problem.

PR #3280, opened 2023-06-21
don't close webxdc on `sendToChat()`

### LINKED ISSUE #3269: `sendToChat()` should be cancelable at least on first screen?

Simon-Laux:
closes #3271, closes #3269

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased][unreleased]
 
+### Changed
+- don't close webxdc on `sendToChat()`
 
 ### Fixed
 - fix: if systemPreferences.askForMediaAccess is not available, then don't call it (broke qr scan under linux (and maybe also under windows, was not tested))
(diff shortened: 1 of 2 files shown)


r10s (approved):
great, directly played around with the pr!

it feels much better, even though the user may "lose" track to the webxdc window as it is lowered and may be completely covered by the main window.

but even than, it is an improvement to "just closing"

hpk42:
agreed, thanks for the quick fix!

PR #3282, opened 2023-06-21
fix: only allow one instance of `sendToChat()`

### LINKED ISSUE #3281: sendToChat() dialog can have multiple instances
It should not be able to be opened multiple times on top of each other.

should we just ignore requests to open it when it is already open? or do sth different?

it gets more visible with #3280, but was already a problem before that. (though you needed to have multiple webxdc apps open to trigger it.

Simon-Laux:
the old dialog is replaced by the new one

closes #3281


hpk42 (approved):
works for me

PR #3293, opened 2023-06-30
add usability improvements to sticker selector UI

### LINKED ISSUE #3258: Improve sticker selector UI
At the moment it looks like this:
Bildschirmfoto 2023-06-09 um 14 31 59

I have 2 easy improvement ideas:

- make the "Open Sticker Folder" - Button nicer and move it to the bottom?
- If there are no stickers then show an "empty-screen" that says "no stickers yet" and tells the user they can add stickers with the "Open Sticker Folder" button. ("Non-ideal UI states" is also a term you can search for for inspiration, maybe similar to https://blueprintjs.com/docs/#core/components/non-ideal-state, but don't use blueprint, because we want to get rid of that dependency)

Maybe you (whoever reads this and ideally wants to make a pr) have other **easy** style improvement ideas, too.

achou11:
Closes #3258 

Notes:

- non-ideal state for when no sticker packs exist
- makes each sticker pack title sticky when scrolling until the next pack is reached. 
- ability to tab through the stickers and select using <kbd>Enter</kbd> or <kbd>Space</kbd>
- wasn't entirely sure about how to go about translations so did the best I could given the instructions in the development docs.
- not entirely sure about styling for the button that opens the sticker folder. just used an existing button styling that I figured is used elsewhere

---

Preview:

- Non-ideal state

  image
  
  
- Populated state

  image

- Interaction with populated state (first mouse, then keyboard):

  - Mouse:
    
    3258-sticker-selector-interaction-mouse

  - Keyboard (not sure why the GIF tool I used decided to make this grainy 😅 ):
    
    3258-sticker-selector-interaction-keyboard


src/renderer/components/composer/Composer.tsx
```diff
@@ -323,7 +323,10 @@ const Composer = forwardRef<
             <span />
           </div>
           <div className='send-button-wrapper' onClick={composerSendMessage}>
-            <button aria-label={tx('menu_send')} />
+            <button
+              aria-label={tx('menu_send')}
+              tabIndex={showEmojiPicker ? -1 : undefined}
```
achou11, line 328: This is to prevent the send button from being tab-focusable while the emoji/sticker picker is showing. not sure if this is desired or not so wanted to higlight

src/renderer/components/composer/EmojiAndStickerPicker.tsx
```diff
@@ -39,27 +38,30 @@ export const StickerDiv = (props: {
...
-            <div className='sticker' key={index}>
-              <img
-                src={filePath}
-                onClick={onClickSticker.bind(null, filePath)}
-              />
-            </div>
-          )
-        })}
+        {stickerPackImages.map((filePath, index) => (
+          <button
```
achou11, line 42: should I add an aria-label for this?

src/renderer/components/composer/EmojiAndStickerPicker.tsx
```diff
@@ -68,22 +70,46 @@ export const StickerPicker = (props: {
...
+                chatId={chatId}
+                key={name}
+                stickerPackName={name}
+                stickerPackImages={stickers[name]}
+                setShowEmojiPicker={setShowEmojiPicker}
+              />
+            ))}
+          </div>
+          <div className='sticker-actions-container'>
+            <button
```
achou11, line 91: should I add an aria-label for this?

src/renderer/components/composer/EmojiAndStickerPicker.tsx
```diff
@@ -68,22 +70,46 @@ export const StickerPicker = (props: {
...
+              {tx('open_sticker_folder')}
+            </button>
+          </div>
+        </>
+      ) : (
+        <div className='sticker-container'>
+          <div className='no-stickers'>
+            <h2 className='title'>{tx('no_stickers_yet')}</h2>
+            <p className='description'>{tx('add_stickers_instructions')}</p>
+            <button
```
achou11, line 104: should I add an aria-label for this?

src/renderer/components/composer/Composer.tsx
Simon-Laux: Maybe rather split that into a dedicated pr, so discussing it does not block this pr. I would like to have more opinions on this. I myself are neither for nor in favour of this, I'm neutral on this, both is fine to me.

src/renderer/components/composer/EmojiAndStickerPicker.tsx
Simon-Laux: with the filename? or what did you have in mind here?

src/renderer/components/composer/EmojiAndStickerPicker.tsx
Simon-Laux: I don't think it is required when the content of the button is already the text that would be in the aria-label?

src/renderer/components/composer/EmojiAndStickerPicker.tsx
Simon-Laux: same as above, I think it is not needed for text buttons?

src/renderer/components/composer/Composer.tsx
achou11: makes sense! removed via fd76a9cfb30fb961cbc598908d15e32e52624067

Simon-Laux (commented):
changelog still needs an entry about this, otherwise code looks good.

Works fine.

"Add some stickers bellow" does not tell how to do it, maybe we need a more detailed explanation (about add a folder, name is name of pack and content is image files)? @r10s what do you think translation wise?

Also stickers don't refresh when I change something with the folder until I close and reopen the whole emoji&sticker popup, but that minor issue is out of scope here.

src/renderer/components/composer/EmojiAndStickerPicker.tsx
achou11: don't have anything specific in mind, but yeah maybe something like "Send FILENAME" would work?

achou11:
> "Add some stickers bellow" does not tell how to do it, maybe we need a more detailed explanation (about add a folder, name is name of pack and content is image files)?

yeah agreed. definitely one of the things I intended to update based on review :)

> Also stickers don't refresh when I change something with the folder until I close and reopen the whole emoji&sticker popup, but that minor issue is out of scope here.

yeah i noticed that too and was tempted to figure out how that could work, but agree that it's out of scope for this

_locales/_untranslated_en.json
```diff
@@ -61,5 +61,14 @@
   },
   "later": {
     "message": "Later"
+  },
+  "no_stickers_yet": {
+    "message": "No stickers yet"
+  },
+  "add_stickers_instructions": {
+    "message": "Add some using the button below."
```
r10s, line 69: not sure, how the button works, but i assume it just opens the folder?

i would then go for sth. like the following following message:

```
To add stickers, 
tap "Open Sticker Folder"
and drag image and sticker files to that folder.
```

to have less text, the error "No stickers yet" can also be removed, that seems obvious.

src/renderer/components/composer/EmojiAndStickerPicker.tsx
```diff
@@ -68,22 +70,46 @@ export const StickerPicker = (props: {
...
+                setShowEmojiPicker={setShowEmojiPicker}
+              />
+            ))}
+          </div>
+          <div className='sticker-actions-container'>
+            <button
+              className='delta-button-round'
+              onClick={onOpenStickerFolder}
+            >
+              {tx('open_sticker_folder')}
```
r10s, line 95: this button is too bold if there is already content.

the look of a secondary button is totally sufficient here (otherwise it looks as if it needs to be clicked to proceed)

(secondary button = transparent background, border in button color, if at all)

achou11:
After the some recent commits to address feedback:

- Empty:

image

---

- Non-empty:

image

_locales/_untranslated_en.json
achou11: addressed via https://github.com/deltachat/deltachat-desktop/pull/3293/commits/3ac757a5906cea2f105abdc37f7d3455cb3a8049

src/renderer/components/composer/EmojiAndStickerPicker.tsx
achou11: addressed via https://github.com/deltachat/deltachat-desktop/pull/3293/commits/11443b40e1ac7dabf039e04f5229f683165a249e

lmk if you think the border should also be removed? currently just uses the existing style definition for the secondary button

Simon-Laux:
> ```
> To add stickers, 
> tap "Open Sticker Folder"
> and drag image and sticker files to that folder.
>```

Are stickers that are just in the sticker folder recognised or do they need to be into a "sticker pack" folder?
If not we can either change that in core, adjust the text to say that, or create and open a default "sticker pack" folder.

Simon-Laux (approved):

PR #3300, opened 2023-07-11
CreateChat dialog: Trim the input value to fix #3299

### LINKED ISSUE #3299: Sanitize Input in New Chat Dialog
- Operating System (Linux/Mac/Windows/iOS/Android): Arch
- Delta Chat Version: Latest
- Expected behavior: When I copy-paste an e-mail address into the "New Chat" dialog, it is accepted.
- Actual behavior: Sometimes the copy-paste adds an extra space, which trips Delta up.

farooqkz:
This is a trivial solution to fix #3299. I simply trim the value of the `<input>` upon each change.


src/renderer/components/dialogs/CreateChat.tsx
```diff
@@ -194,7 +194,7 @@ export function useContactSearch(
   }
 
   const onSearchChange = (e: React.ChangeEvent<HTMLInputElement>) =>
-    updateSearch(e.target.value)
+    updateSearch(e.target.value.trim())
```
Simon-Laux, line 197: I would instead do this before sending to core and allow users to type spaces there

Simon-Laux: the function to create a new contact is in Line 412

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 
 ### Fixed
 - fix: clicking start on an already open webxdc app now opens it again even when it was minimised #3294
+- fix: Email address is now trimmed in create chat dialogue(#3299)
```
Simon-Laux, line 8:
```suggestion
- fix: whitespaces are now removed from Email address when creating a contacts from the Create Chat dialog(#3299)
```

Simon-Laux (changes requested):
Ok just tested it, looks we also need to trim it before giving it to the email validation function

farooqkz:
I just remembered for the same reason I trimmed it in there lol

farooqkz:
@Simon-Laux What's the problem with trimming them the first place I did?

Simon-Laux:
> @Simon-Laux What's the problem with trimming them the first place I did?

user can not type spaces which feels restricted, say for example you have multiple contacts with the same first name but different surnames and you want to just type it in, then notice that you can not type a space.

farooqkz:
I've updated the hook used for searching for contacts to trim the provided `queryStr` before using it as a search query for contact list or checking validity of it as an email address.

Also, when the user queries a name, all the contacts with that name *and* the "New contact" item are shown but clicking on the "New contact" button does nothing. It's a bug I suppose?

src/renderer/components/contact/ContactList.tsx
```diff
@@ -118,7 +118,9 @@ export function useContactsMap(listFlags: number, queryStr: string) {
   return [contacts, updateContacts] as [typeof contacts, typeof updateContacts]
 }
 
-export function useContactsNew(listFlags: number, initialQueryStr: string) {
+
+// if trim is provided and is true, the queryStr is trimmed for contact search and for checking validity of the string as an Email address
+export function useContactsNew(listFlags: number, initialQueryStr: string, trim?: boolean) {
```
Simon-Laux: I think you can just trim it in all cases, this does not need to be optional, as @link2xt said we can also implement that in core, which would mean the input is trimmed on every call to `getContacts` and `checkEmailValidity`

farooqkz: As link2xt said, I also feel doing it here rather than in the core is a better idea. I agree with `useContactsNew` always trimming as it is called only for CreateChat dialogue.

src/renderer/components/contact/ContactList.tsx
```diff
@@ -131,10 +132,10 @@ export function useContactsNew(listFlags: number, initialQueryStr: string) {
         const contacts = await BackendRemote.rpc.getContacts(
           accountId,
           listFlags,
-          queryStr
+          trim ? queryStr.trim() : queryStr
```
Simon-Laux, line 135: trim is not defined so this will never be trimmed, will it?

src/renderer/components/dialogs/CreateChat.tsx
```diff
@@ -61,7 +61,8 @@ export default function CreateChat(props: {
 
   const [{ contacts, queryStrIsValidEmail }, updateContacts] = useContactsNew(
     C.DC_GCL_ADD_SELF,
-    ''
+    '',
+    true
```
Simon-Laux, line 65: trim does not exist anymore so don't set it to true

Simon-Laux (approved):

PR #3306, opened 2023-07-15
make links unclickable only if inside quote to fix #3305

### LINKED ISSUE #3305: Recently merged #3304 prevents clickable links everywhere
One commit before  #3304 was merged, clicking links was possible. 
Afterwards links are not clickable anymore (quoted or unquoted). 
Tested and reproducable in my from-source desktop install. 
Before providing a fixing PR please test it yourself (that quoted links don't work but links elsewhere work), thanks.

farooqkz:


farooqkz:
I suppose updating changelog is not needed in this case?

Simon-Laux (approved):

Simon-Laux:
> I suppose updating changelog is not needed in this case?

yes, exactly, the only thing you could potentially add is the number of this pr to the existing changelog entry

PR #3307, opened 2023-07-15
fix duplicated status updates race condition bug in webxdc

### LINKED ISSUE #3296: race-bug: webxdc status updates get received multiple times
when there are multiple status updates at the same time, sometimes they get send multiple times to the webxdc (also in wrong order?):
Bildschirmfoto 2023-07-01 um 19 13 40

It is probably that each update causes sending all new update from the update's serial onwards, we should fix this:
- easy: only send the update which serial is in the event
- more efficient / smarter: track what serial was send last and only send them on update if their serial is above the listed serial (also make sure that they have the right order if my suspicion about wrong order should be correct

I found this while using the time tracking webxdc: https://github.com/webxdc/timetracking-webxdc/issues/15

Simon-Laux:
closes #3296

The solution was to make sure only one instance of the getAllUpdates function can run at a time.

You can test this easialy with https://github.com/webxdc/webxdc-test/pull/18 (link at the bottom of the xdc app).


static/webxdc-preload.js
```diff
@@ -9,19 +9,37 @@
   let callback = null
   var last_serial = 0
   let setUpdateListenerPromise = null
+  let is_running = false
+  let scheduled = false
```
farooqkz, line 13: Why these names are in snake case?

Simon-Laux: no particular reason, there are no real/clear rules for casing in desktop, I guess camel case is preferred generally, but if you look above `last_serial` is also snake case.

adbenitez (approved):

PR #3309, opened 2023-07-17
move sync all button in account list screen to bottom to fix #2963

### LINKED ISSUE #2963: Title in the account list screen is not fully displayed in German.
When _DC-Desktop 1.33_ is running in German, the title “_Bekannte E-Mail-Konten_” is not fully displayed in the start screen.

Startbildschirm Fehler

One could make the "_Alle(s) synchronisieren_" button smaller, or just omit the word "_Bekannte_" in the title since the accounts are known anyway. Or of course both.

Startbildschirm Korrektur

farooqkz:
Per [this suggestion](https://github.com/deltachat/deltachat-desktop/issues/2963#issuecomment-1636885713), the sync all button is moved to bottom.


r10s:
> there is so much space there, maybe there is a way to make it look more like @r10s mockup

i agree, i would do it similar to the mockup.

ppl will wonder otherwise, what the switch is - or what the label is :)

having the switch on the left side is totally fine in this case, the "key left - value right" alignment does not make sense in this case and just leads to confusion.

if it is really hard to implement the mockup, as a second-good variation, we can have the switch on the right, however, with a right-aligned title. but i would prefer the idea from the mockup.

Simon-Laux (approved):

PR #3310, opened 2023-07-18
Fix edit button in a contact's profile not shown when selected from a group member list(fix #3267)

### LINKED ISSUE #3267: View Profile: "Edit Name" is not available when accessing the contact from the members section in a group view
`Edit Name` is not available when accessing the contact from the members section in a group view.
It is confusing to have 2 slightly different versions of the same view, depending on how it was opened.

farooqkz:
This uses `ViewProfile` component to show someone's profile when they are selected from group member list. This fixes #3267 

Note: currently, when the X button is pressed in the user's profile which is shown this way, the entire dialogue is closed. Should it show the previous dialogue instead which is the group member list?

### DIFF
--- a/src/renderer/components/dialogs/ViewGroup.tsx
+++ b/src/renderer/components/dialogs/ViewGroup.tsx
@@ -79,7 +79,6 @@ export default function ViewGroup(props: {
   isBroadcast: boolean
 }) {
   const { isOpen, onClose, isBroadcast } = props
-  const [viewMode, setViewMode] = useState('main')
 
   const chat = useChat(props.chat)
 
@@ -93,13 +92,7 @@ export default function ViewGroup(props: {
         height: 'calc(100vh - 50px)',
       }}
     >
-      <ViewGroupInner
-        viewMode={viewMode}
-        setViewMode={setViewMode}
-        onClose={onClose}
-        chat={chat}
-        isBroadcast={isBroadcast}
-      />
+      <ViewGroupInner onClose={onClose} chat={chat} isBroadcast={isBroadcast} />
     </DeltaDialogBase>
   )
 }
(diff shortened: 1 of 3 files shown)


src/renderer/components/dialogs/ViewGroup.tsx
```diff
@@ -217,9 +217,15 @@ function ViewGroupInner(props: {
     null
   )
 
+  if (viewMode === 'profile' && !profileContact) {
```
Simon-Laux, line 220: you could just check for profile contact and remove the view mode, then you don't need this if statement and have simpler code?

Simon-Laux (commented):
So the back button does not exist anymore in this pr?
Point of my issue was to have the edit button in both cases, the back button should not be changed.

src/renderer/components/dialogs/ViewGroup.tsx
farooqkz: the thing is that if view mode is profile, profile contact must be non-null. If it's null, it's a bug. The code shows the main view anyway, but this error log is printed to console.

farooqkz:
> So the back button does not exist anymore in this pr? Point of my issue was to have the edit button in both cases, the back button should not be changed.

How about replacing the close button with a back button?

src/renderer/components/dialogs/ViewGroup.tsx
Simon-Laux: my idea here was to remove the view mode and make it dependent on wether `profileContact` is set or not, so one variable less

Simon-Laux:
> How about replacing the close button with a back button?

No, this issue is only about making the edit button available in both cases, the UI should not be changed.
Also the difficulty of passing down a back button callback is the same or less effort than passing down a property to change the appearance and logic of the close button.

Simon-Laux (changes requested):
as I said UI besides the edit button should not be changed, the back button should still be shown and the close button should still close the whole dialog.

farooqkz:
> as I said UI besides the edit button should not be changed, the back button should still be shown and the close button should still close the whole dialog. I can take over this pr if you want.

Nope I will do it. I think I misread your message about both having back button and close button.

Simon-Laux (approved):
Great, one nitpick, if you change the name and go back the old name is still shown, but we can fix this separately and just make a bug for it for now

PR #3312, opened 2023-07-18
Ask for passphrase when exporting backup(closes #2636)

### LINKED ISSUE #2636: Add UI to set backup passphrase
Since https://github.com/deltachat/deltachat-core-rust/pull/2980 is merged (core 1.71.0), there is a possibility to set passphrase for the backup. UI is needed to allow encrypting backup.

Related android issue: https://github.com/deltachat/deltachat-android/issues/2239

farooqkz:
This PR:

 - Adds a new dialog: TextDialog
 - Removes ViewProfile's edit contact dialog and replaces it with TextDialog instead
 - Uses TextDialog to ask for passphrase when exporting backup
 - Closes #2636 

It's a draft because:
 - The `pref_backup_export_explain` string must be changed.
 - A `passphrase` string must be added. This is used as a placeholder for TextDialog's text input

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@
 - fix: links in quoted texts should not be clickable (#3290)
 - fix: remove unsupported language code, this broke the installation from the ms store on windows #3292
 
+### Changed
+
+- Ask passphrase for exporting backup.
+
 <a id="1_38_1"></a>
 
 ## [1.38.1] - 2023-06-23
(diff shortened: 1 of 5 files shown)


hpk42:
did you think about what this export-passphrase will cause? 
If this is using core support then only the database but not the blob files (attachments) will be encrypted which might be suprising for users.  Also what happens if you try to 
import this "encrypted backup" into Android/iOS/UbuntuTouch?

farooqkz:
Good thing I marked it as draft!

farooqkz:
Do you think it's a good idea to warn the user about unencrypted blobs?

r10s:
+1, we have a "resurrection" project, also the corresponding issue should go there

if we pick that up again, we should first reconsider

PR #3313, opened 2023-07-18
Make input text in CreateChat dialogue always focused

### LINKED ISSUE #3286: “Enter name or e-mail address” bar focus
- Operating System: Kubuntu 22.04
- Delta Chat Version: 1:38 (and previous I think)
- Expected behavior: After "clicking" the plus sign for new address the address bar should be in focus when starting "Enter name or e-mail address"
- Actual behavior: This is only working if I have selected the "Device Messages" and not if I'm reading an old message.

Also see - https://support.delta.chat/t/help-testing-the-upcoming-1-37-x-release/2608/11

farooqkz:
This PR fixes #3286

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - fix: text is now smaller in the unread badge on the top of the jump down button(#3068)
 - fix: links in quoted texts should not be clickable (#3290)
 - fix: remove unsupported language code, this broke the installation from the ms store on windows #3292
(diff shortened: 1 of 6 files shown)


Simon-Laux (changes requested):
Code Looks good

But I would like to find a better name for `getOpenDialogsNumber` as it was not apparent to me what it meant. If it's not important how many dialogs are open, then maybe a simple boolean return value and a function named `hasOpenDialogs` would be ok?

Simon-Laux (approved):

PR #3316, opened 2023-07-19
Don't save whitespace-only message as draft(fixes #3220)

### LINKED ISSUE #3220: trim/strip spaces from draft before saving it
step to reproduce: 
- just open a chat and press space key some times, 
- go to another chat
- the empty spaces are saved as draft

that doesn't happen in Android where no draft is saved for empty messages with just space

farooqkz:


Simon-Laux (approved):
works

PR #3318, opened 2023-07-19
make a chat list item look hovered when its context menu is open

### LINKED ISSUE #3238: chatlist: visual context missing for chat's context menu
when right clicking a chat in the chatlist, the context, which items will be affected is easily lost as nothing is hilited.

confusion raises esp. when not clicking carefully in the center, but more to the edges - see the following screenshot: will "openai@b44t.com" or "offline-test2" will be deleted?

as the hover effect continues although the popup is open, the user may get even more confusing impressions.

things are also a bit flaky - _every second click_, at least initially the correct chat is hilited:

Screenshot 2023-05-27 at 00 11 41

would be nice if one could hilite the selected item using the "hover" colorand stop hovering while the menu is open. adding more colors or other states would probably not help all in all (more clutterung and confusion).

not sure if this is easily fixable, however - whatsapp web has the same issue and telegram does not use hiliting at all. desktop apps as the explorer or the finder, however, do better.

so, @Simon-Laux, feel free to close this issue if it is more effort than, say, an hour and/or would add much complexity. this price would be too high :)

farooqkz:
fixes #3238 

Edit: took about an hour :)

### DIFF
--- a/src/renderer/components/chat/ChatListItemRow.tsx
+++ b/src/renderer/components/chat/ChatListItemRow.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useState } from 'react'
 import ChatListItem, {
   PlaceholderChatListItem,
   ChatListItemMessageResult,
@@ -31,6 +31,8 @@ export const ChatListItemRowChat = React.memo<{
     openContextMenu,
   } = data
   const chatId = chatListIds[index]
+  const [hoverChatListItem, setHoverChatListItem] = useState(false)
+
   return (
     <div style={style}>
       <ChatListItem
@@ -39,9 +41,13 @@ export const ChatListItemRowChat = React.memo<{
         onClick={onChatClick.bind(null, chatId)}
         onContextMenu={event => {
           const chat = chatCache[chatId]
-          if (chat?.type === 'ChatListItem')
+          if (chat?.type === 'ChatListItem') {
+            setHoverChatListItem(true)
             openContextMenu(event, chat, selectedChatId)
+            setHoverChatListItem(false)
+          }
         }}
+        hover={hoverChatListItem}
       />
     </div>
   )
--- a/src/renderer/components/chat/ChatListItemRow.tsx
+++ b/src/renderer/components/chat/ChatListItemRow.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React from 'react'
 import ChatListItem, {
   PlaceholderChatListItem,
   ChatListItemMessageResult,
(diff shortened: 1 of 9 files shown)


src/renderer/components/chat/ChatListItemRow.tsx
```diff
@@ -39,9 +41,13 @@ export const ChatListItemRowChat = React.memo<{
         onClick={onChatClick.bind(null, chatId)}
         onContextMenu={event => {
           const chat = chatCache[chatId]
-          if (chat?.type === 'ChatListItem')
+          if (chat?.type === 'ChatListItem') {
+            setHoverChatListItem(true)
             openContextMenu(event, chat, selectedChatId)
+            setHoverChatListItem(false)
```
Simon-Laux, line 47: while this could work in theory, `openContextMenu` is not blocking (which would be bad, because blocking would mean that the UI would be frozen ;). This would be a case for async, like `await openContextMenu`, but for that to work the context menu needs to be patched to return a promise when it was clicked away, closed, or once an action was triggered.

My first idea was building in the promise into `useChatListContextMenu`, but that would be really ugly / hard to maintain code, like resolving the promise in each action callback, so I think it would make more sense to build it directly into `openContextMenu`. So change `openContextMenu`  to return a promise that gets resolved in all cases where it is closed (also when an action was chosen it is also closed, so maybe in the function that closes it? but I haven't looked at the code in a while so take it with a grain of salt.).

farooqkz: oh sorry I thought it was working but it seems I was mistaken.

farooqkz:
It is going to take more than an hour for this to solve. But the problem's not that implementing this is big but because my lack of knowledge about async JS. I am claiming this to learn anyway.

r10s:
unless i messed up some building (would be easier to have a feature branch inside deltachat-desktop repo) this does not fix  #3238 for me.

when i right-click a chatlist item, the context menu opens, but it is still unclear to which chatlist item it refers, see #3238 for details an screenshots.

having a rough look at the changes of this pr: i doubt that the issue can be fixed with css `hover` as that changes once the mouse moves - and for the chatlist item, the selection should stay stable for the time the conext menu is opened, cmp windows-explorer or mac-finder.

as already mentioned in the issue: if this is to hard to get right, it may also be reasonable to leave things as is and do not fix this comparable small issue

Simon-Laux:
> having a rough look at the changes of this pr: i doubt that the issue can be fixed with css hover as that changes once the mouse moves - and for the chatlist item, the selection should stay stable for the time the conext menu is opened, cmp windows-explorer or mac-finder. but maybe hover in electron/react/dc works differently.

 `&:hover,
   &.context-menu-active`, the `,` means OR, I guess you either missed that or looked at an older version of the code. 

For me it worked at the time that I was working on it.

r10s (approved):
i messed up branches during checkout, sorry for that, as said, on-org branches are far easier with git :) i could test the pr finally using github cli `gh pr checkout 331` - thanks @Simon-Laux 

on-topic: looks good to me ui-wise, thanks a lot! code-wise, i think, @Simon-Laux already checked the relevant parts.

PR #3324, opened 2023-07-22
fix problem of infinite exceptions when it can't write to stdout(fixes #3323)

### LINKED ISSUE #3323: Desktop shows alerts when started from the terminal and terminal is closed
To reproduce, run `npm run start` from `bash`, then press `^Z` to pause, then type `bg` to continue the process in the background, then close the terminal.

Desktop than constantly fails while trying to write the logs to standard output which is now closed:

Closing one message immediately result in opening the next one.

The only way to recover is to use `killall electron` and restarting the desktop.

farooqkz:
It seems on node v16, we haven't got `stdout.closed` to check it for being true. [that's for v18](https://nodejs.org/dist/latest-v18.x/docs/api/stream.html#writableclosed).

According to docs of v16, we've got [writable](https://nodejs.org/dist/latest-v16.x/docs/api/stream.html#writablewritable) but it is `true` when app is in background and the crash still happens. So this solution I've committed is the only one I've found working.

Fixes #3323

### DIFF
--- a/src/shared/logger.ts
+++ b/src/shared/logger.ts
@@ -3,9 +3,7 @@ import StackFrame from 'stackframe'
 import { RC_Config } from './shared-types'
 import { stdout } from 'process'
 
-
 stdout.on('error', () => {})
-
 
 const startTime = Date.now()
 
--- a/src/shared/logger.ts
+++ b/src/shared/logger.ts
@@ -1,6 +1,11 @@
 import errorStackParser from 'error-stack-parser'
 import StackFrame from 'stackframe'
 import { RC_Config } from './shared-types'
+import { stdout } from 'process'
+
+
+stdout.on('error', () => {})
+
 
 const startTime = Date.now()
 
--- a/src/shared/logger.ts
+++ b/src/shared/logger.ts
@@ -1,9 +1,6 @@
 import errorStackParser from 'error-stack-parser'
 import StackFrame from 'stackframe'
 import { RC_Config } from './shared-types'
-import { stdout } from 'process'
-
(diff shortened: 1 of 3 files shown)


src/shared/logger.ts
```diff
@@ -1,6 +1,9 @@
 import errorStackParser from 'error-stack-parser'
 import StackFrame from 'stackframe'
 import { RC_Config } from './shared-types'
+import { stdout } from 'process'
+
+stdout.on('error', () => {})
```
Simon-Laux, line 6: this file is shared between renderer and main process. we can not use node dependencies in the renderer process.

Simon-Laux: I guess you meant to edit this file instead: https://github.com/deltachat/deltachat-desktop/blob/master/src/main/log-handler.ts

but also then your pr description tells about a different solution, so maybe you also forgot to commit some files?

farooqkz: okay I will put the error handler in that file, there are other better solutions for v18 but for v16 which Desktop is currently using, this is the best I accomplished. Also according to linkxt in the issue, on UNIX(-like), you have to try writing and see if you get error or not.

Simon-Laux: we can also upgrade the node version to 18 in a dedicated pr: updating all docs, readme, github action ci files and so on.

farooqkz: Not sure if upgrading to v18 and using `stdout.closed` will solve this crash issue.

farooqkz:
@Simon-Laux @link2xt May I get approvals?

PR #3325, opened 2023-07-22
add: `webxdc_custom.desktopDragFileOut` api

Simon-Laux:
this is an api that we added to make the UX of the "store" webxdc nicer on desktop, it allows for drag and drop sending.
It is not yet "official" or really documented as it is just for this one app.

- https://github.com/webxdc/store/pull/197

### DIFF
--- a/src/main/deltachat/webxdc.ts
+++ b/src/main/deltachat/webxdc.ts
@@ -15,6 +15,7 @@ import { tx } from '../load-translations'
 import { DcOpenWebxdcParameters } from '../../shared/shared-types'
 import { DesktopSettings } from '../desktop_settings'
 import { window as main_window } from '../windows/main'
+import { writeTempFileFromBase64 } from '../ipc'
 
 const open_apps: {
   [instanceId: string]: {
@@ -172,7 +173,8 @@ export default class DCWebxdc extends SplitOut {
                   mimeType,
                   data: Buffer.from(
                     `window.parent.webxdc_internal.setup("${selfAddr}","${displayName}")
-                  window.webxdc = window.parent.webxdc`
+                  window.webxdc = window.parent.webxdc
+                  window.webxdc_custom = window.parent.webxdc_custom`
                   ),
                 })
               } else {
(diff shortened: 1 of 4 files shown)


farooqkz (commented):
Why the data is being transferred using b64 rather than a Blob?

farooqkz (changes requested):
In the preload javascript file, according to jsdoc, `icon_data_url` is always present and `string`. But in `webxdc.ts` in handler for `webxdc:custom:drag-file-out` it is optional.

src/main/deltachat/webxdc.ts
```diff
@@ -430,6 +432,30 @@ If you think that's a bug and you need that permission, then please open an issu
...
+        icon_data_url?: string
+      ) => {
+        const path = await writeTempFileFromBase64(file_name, base64_content)
+        let icon: string | Electron.NativeImage = join(
+          __dirname,
+          '../../../images/electron-file-drag-out.png'
+        )
+        if (icon_data_url) {
+          icon = nativeImage.createFromDataURL(icon_data_url)
+        }
```
farooqkz, line 450:
```suggestion
        let icon: string | Electron.NativeImage = icon_data_url ? nativeImage.createFromDataURL(icon_data_url) : join(
          __dirname,
          '../../../images/electron-file-drag-out.png'
        )
```

Simon-Laux:
> What do you mean it "always" works? Like there are situations where blob does not and b64 does? Because I think using base64 for binary data is inefficient.

The sendToChat api also uses base64 under the hood, I don't know if you can even transfer Blob objects over electron ipc.
I know that base64 is on average 30% larger.

src/main/deltachat/webxdc.ts
Simon-Laux: does not really matter here, so I would keep it

farooqkz:
> > What do you mean it "always" works? Like there are situations where blob does not and b64 does? Because I think using base64 for binary data is inefficient.
> 
> The sendToChat api also uses base64 under the hood, I don't know if you can even transfer Blob objects over electron ipc. I know that base64 is on average 30% larger.

I still insist on using Blobs if possible. base64 is not only larger, but there is also decode/encode price.

Simon-Laux:
> I still insist on using Blobs if possible

If you want you can find out if it is possible, but I don't think it is worth the effort or time, the impact is super minimal and fast enough for our mall files use case. There are other areas where we can invest time that have a bigger impact on user experience.

farooqkz:
> > I still insist on using Blobs if possible
> 
> If you want you can find out if it is possible, but I don't think it is worth the effort or time, the impact is super minimal and fast enough for our small files use case. There are other areas where we can invest time that have a bigger impact on user experience.

I tested on my laptop which has got average hardware and it's trivial. Let it be as is then. But as a reference, encoding 50MiB of binary data to base64 took almost a second on Allwinner H3, an ARM processor, which is huge.

Simon-Laux:
data in this case is under 2mb also deltachat desktop probably will not anyway on such slow processors, remember it uses electron, which uses at least 200-300mb of ram because it is a full chrome browser.

PR #3326, opened 2023-07-22
add context menu entry to resend self sent messages

### LINKED ISSUE #3259: add "send message again" (resend message)
afaik it's currently implemented in android and only for webxdc (for webxdc it resends all state updates together with the xdc).

@adbenitez recently said he missed that functionality in desktop. What's the current state of the feature? I guess the usefulness for helping new members to catch up with normal messages is limited as long as there is no multi selection on desktop.

also resend_message is not implemented in the jsonrpc-api yet.

CC @hpk42 @link2xt @r10s @Hocuri

Simon-Laux:
closes #3259
depends on https://github.com/deltachat/deltachat-core-rust/pull/4567

- [x] make it possible to resend all self sent messages not only webxdc messages, that is the real current android behaviour

### DIFF
--- a/src/renderer/components/message/Message.tsx
+++ b/src/renderer/components/message/Message.tsx
@@ -189,8 +189,7 @@ function buildContextMenu(
   const showAttachmentOptions = !!message.file && !message.isSetupmessage
   const showCopyImage = !!message.file && message.viewType === 'Image'
 
-  const showResend =
-    message.viewType === 'Webxdc' && message.sender.id === C.DC_CONTACT_ID_SELF
+  const showResend = message.sender.id === C.DC_CONTACT_ID_SELF
 
   return [
     // Reply
--- a/src/renderer/components/message/Message.tsx
+++ b/src/renderer/components/message/Message.tsx
@@ -189,6 +189,9 @@ function buildContextMenu(
   const showAttachmentOptions = !!message.file && !message.isSetupmessage
   const showCopyImage = !!message.file && message.viewType === 'Image'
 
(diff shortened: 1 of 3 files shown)


src/renderer/components/message/Message.tsx
```diff
@@ -189,6 +189,9 @@ function buildContextMenu(
   const showAttachmentOptions = !!message.file && !message.isSetupmessage
   const showCopyImage = !!message.file && message.viewType === 'Image'
 
+  const showResend =
+    message.viewType === 'Webxdc' && message.sender.id === C.DC_CONTACT_ID_SELF
```
farooqkz, line 193: As, according to #3259, the goal is an option to resend own message, webxdc or not, isn't this boolean expression incorrect?

Simon-Laux:
> https://github.com/deltachat/deltachat-desktop/issues/3259 is about resending any own message not just WebXDC ones. But the title of this PR says something else :/

this issue was basically a discussion where I asked the others wether to also add it for other message types, there was no clear answer so I decided to add the same behaviour as on android: "you can only resend webxdc messages".

src/renderer/components/message/Message.tsx
Simon-Laux: no it is correct as the discussion there yielded no clear results it is better to use copy the android behaviour for now.
But it is very easy to change this in the future if we should decide otherwise.
another idea would be to also show it on failed messages that failed to send - but I still think we need to discuss this also with more people.

PR #3331, opened 2023-07-26
Fix colour of clock icon in ViewProfile

### LINKED ISSUE #3329: View profile: clock icon for last seen is not clearly visible in dark mode
Bildschirmfoto 2023-07-22 um 20 13 13

I think it should become white or inverted in darkmode. maybe there is a way to just give it the text color (the css variable which gets set by the theme).

farooqkz:
Fixes #3329

### DIFF
--- a/scss/_icons.scss
+++ b/scss/_icons.scss
@@ -45,3 +45,9 @@ img.material-icon {
   vertical-align: middle;
   transform: translate(0px, -0.07em);
 }
+
+i.material-svg-icon {
+  -webkit-mask-repeat: no-repeat;
+  background-color: var(--globalText);
+  display: inline-block;
+}
--- a/scss/_icons.scss
+++ b/scss/_icons.scss
@@ -50,4 +50,13 @@ i.material-svg-icon {
   -webkit-mask-repeat: no-repeat;
   background-color: var(--globalText);
   display: inline-block;
+  height: 1.15em;
+  width: 1.15em;
+  -webkit-mask-size: 1.15em;
+  margin-inline-start: 0.3em;
+  margin-inline-end: 0.3em;
+}
+
+i.material-icon-schedule {
+  -webkit-mask-image: url(../images/icons/schedule.svg);
 }
--- a/scss/_icons.scss
+++ b/scss/_icons.scss
@@ -60,4 +60,5 @@ i.material-svg-icon {
 
 i.material-icon-schedule {
   -webkit-mask-image: url(../images/icons/schedule.svg);
+  transform: translate(0px, -0.07em);
 }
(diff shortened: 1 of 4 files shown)


scss/dialogs/_view_profile.scss
```diff
@@ -58,5 +58,17 @@ div.profile-info-container {
     &.clickable:hover {
       cursor: pointer;
     }
+
+    & > i.material-icon {
```
Simon-Laux, line 62: we might have other material icons, so maybe better add the bulk of this calls to some class maybe named `.material-svg-icon` and then also add a class `.material-icon-schedule`? and do this in the scss file for icons. then we can also reuse the class elsewhere and would just overwrite things like size in viewProfile if even necessary, Basing it on `em` is a good choice so probably such overwrites are not necessary.

farooqkz: Agreed.

farooqkz: @Simon-Laux How is it now?

scss/dialogs/_view_profile.scss
```diff
@@ -58,5 +58,14 @@ div.profile-info-container {
...
       cursor: pointer;
     }
+
+    & > i.material-icon-schedule {
+      -webkit-mask-image: url(../images/icons/schedule.svg);
+      height: 1.15em;
+      width: 1.15em;
+      -webkit-mask-size: 1.15em;
+      margin-inline-start: 0.3em;
+      margin-inline-end: 0.3em;
```
Simon-Laux, line 68: I think you can also put this in the base class as it adapts to the current font size

Simon-Laux (changes requested):
Code looks good and works, but the icon is not entered anymore which does not look good:

before:
Bildschirmfoto 2023-07-26 um 20 59 55
after:
Bildschirmfoto 2023-07-26 um 21 00 05

scss/_icons.scss
```diff
@@ -45,3 +45,19 @@ img.material-icon {
...
+  height: 1.15em;
+  width: 1.15em;
+  -webkit-mask-size: 1.15em;
+  margin-inline-start: 0.3em;
+  margin-inline-end: 0.3em;
+  vertical-align: middle;
+}
+
+i.material-icon-schedule {
+  -webkit-mask-image: url(../images/icons/schedule.svg);
```
Simon-Laux, line 62:
```suggestion
  -webkit-mask-image: url(../images/icons/schedule.svg);
  transform: translate(0px, -0.07em);
```

adding this makes it look more centred, I have the values from eh other icons above.

farooqkz: image
image

It is fine as it is now

The first image is from this PR and the latter from master(with manually whitening the clock icon)

Simon-Laux: I think the offset looks better, I tried it because to me it looked a bit off.

Simon-Laux (approved):

PR #3332, opened 2023-07-27
center the no media text hint in gallery

farooqkz:


src/renderer/components/Gallery.tsx
```diff
@@ -156,7 +156,12 @@ export default class Gallery extends Component<
                   </ul>
                 </div>
               )}
-              <div className='item-container'>
+              <div
+                className='item-container'
+                style={
+                  medias.length < 1 ? { justifyContent: 'center' } : undefined
+                }
```
Simon-Laux, line 163:
```suggestion
                style={
                   { justifyContent: medias.length < 1 ? 'center': undefined } 
                }
```
maybe still needs code formatting, but I would try it like this instead

PR #3333, opened 2023-07-27
Esc always closes Fullscreen media screen. fix #2919

### LINKED ISSUE #2919: escape-key does not always close enlarged image
when enlarging an image from the gallery and pressing the "next" button _with the mouse,_ "escape" stops working for that image.

"escape" works in most other cases: when enlarging from the message view or when using the right arrow key for "next".

to reproduce:

- open the gallery from a chat (the chat should have some images)
- open any image
- tap the "next image" button on the right (do not use the keyboard for that)
- the next image is shown, but you cannot close the view by "escape"

farooqkz:
Fixes #2919


Simon-Laux (approved):
works for me, great job!

Issue #2268, opened 2021-05-24
esc-key in group create & -edit close the whole stack, not the last dialog


r10s:
usually the "escape"-key closes the most recent dialog; in a stack of dialogs, the underlying dialog is show again then (on can regard the "escape"-key as a kind of shortcut to the "cancel" or "close" button).

however, the "escape"-key closes all dialogs in a stack of dialog - eg. if you select "New chat / New group / Add members" and add some members by hand, entering a name and an icon - and finally go to Add member to check things again there and hit "Esc" there - _all_ your work is lost.

(this is an older issue, i checked, it exists also in 1.15.5 so, no worsening in 1.20.0. maybe the issue was introduced with https://github.com/deltachat/deltachat-desktop/issues/1743)

EDIT: the Esc-key works fine at "Settings / Show Classic mail"  and "Settings / Delete message from ..." - here the dialogs are also visually stacked. maybe we can see that approach as a pattern for the other dialogs? at least for future things? Cmp. directly with "Settings / Edit profile" to get the difference.

Simon-Laux:
note: this dialog is technically not stacked, its a single one.

r10s:
> note: this dialog is technically not stacked, its a single one.

okay. however, from the user-perspective, that does not really matter: there is a dialog opening another dialog - otherwise, also the "Cancel" button should always close all dialogs (just to make things clear, i do not propose to do so :)

Jikstra:
We need to kick out the canEscapeKeyClose logic of blueprint and do our own thing. We need to know more about the dialog stack, which we currently don't know (which dialog is at top...)

Simon-Laux:
we need to do keybindings sometime right anyway, maybe this will already come up with improving accessibility, if an dialog is open other keybindings like with chat should be disabled and you should not be able to tap around in the main app while an dialog is open and stuff like that.
But that's probably a bigger thing, for making the dialog better we could just multiple ones on top of each other or add some escape listeners with the prevent default trick/hack as workaround.

Jikstra:
Any ideas how we can do it "right"?

farooqkz:
I can reproduce this:

1. Click "+"
2. Click "New group"
3. Click "add members"
4. Add some members
5. Click "OK"
6. Press Esc

Simon-Laux:
I think this doesn't apply anymore. If anyone disagrees, then please create a new issue.

PR #3334, opened 2023-07-27
show previous dialog in CreateChat if any

farooqkz:
This tries to address #2268
Currently, it is only implemented for CreateChat dialog. Please comment on my solution. If it's good, I could think of a universal way of doing so for all such these situations.

Currently clicking outside the dialog, goes back to previous dialog. I think this is not good and should rather close every thing. Please comment on this one, too.


farooqkz:
> Dialogs stacked on top of each other are used, for example when scanning qr codes, or asking for confirmation in other places, so I think it should be possible to have multiple dialogs open, but only have the one on the top react to keyboard shortcuts.

This requires bigger change than my solution. I will give it a try on a clean web app to see if BP acts as expected.

farooqkz:
I will open another one.

PR #3338, opened 2023-08-01
update translations (01.07.2023)

Simon-Laux:


_locales/_languages.json
```diff
@@ -2,6 +2,7 @@
   "ar": "عربى",
   "az": "Azeri",
   "bg": "Български език",
+  "bqi": "بختیاری",
```
Simon-Laux, line 5: This is from https://en.wikipedia.org/wiki/Bakhtiari_dialect, I hope it's the correct local name of the language.

r10s: the added language has a latin-like alphabet while the language name looks arabic to me.

this is at least unusual and needs to be checked by a native speaker.

farooqkz: I am not a native speaker, but I know this script. And I'd say it's most likely correct.

Furthermore, the native writing of Arabic seems incorrect to me.

farooqkz:
I am happy that Persian translation is being added :D

_locales/_languages.json
```diff
@@ -2,6 +2,7 @@
   "ar": "عربى",
```
farooqkz, line 2:
```suggestion
  "ar": "العربیة",
```

_locales/_languages.json
Simon-Laux: I'm not sure what you mean @farooqkz, I agree with @r10s that it seems better to use the alphabet in the language name display that is actually used when you select the language.

_locales/_languages.json
Simon-Laux: google translate suggests me this `عربی`, the google translation of suggestion you made looks like the place/region, before it was a person that lived there. We expect the local name of the language that is spoken there. I trust you more on this than the robots, but a second look can't hurt?

farooqkz: Google translate is wrong then. Arabic language is الغة العربیة but the same word also has usage when talking about a land or people. Like البلد العربیة which means Arabic land/city. The ال in العربیة means that we are talking about something specific and what we know. The same for البلد which means we are talking about a land we already know. It's similar to "the" in English. When we say "the boy", we are talking about a specific boy.

Furthermore, you can check the Arabic Wikipedia:

https://ar.wikipedia.org/wiki/%D8%A7%D9%84%D9%84%D8%BA%D8%A9_%D8%A7%D9%84%D8%B9%D8%B1%D8%A8%D9%8A%D8%A9

_locales/_languages.json
farooqkz: I meant the Bakhtiari language use the same script as my language Persian. And your commit seems correct to me.

r10s: thank you for your explanations.

signal also uses [العربية](https://github.com/signalapp/Signal-Android/blob/main/app/src/main/res/values/arrays.xml#L12) meanwhile (i sometimes use that to double check when adding languages, so might be they've changed it as well at some point)

so, +1 for the change, we should change it on android then as well (delta-ios uses system language selector)

_locales/_languages.json
r10s: thank you for your explanations.

signal also uses [العربية](https://github.com/signalapp/Signal-Android/blob/main/app/src/main/res/values/arrays.xml#L12) meanwhile (i sometimes use that to double check when adding languages, so might be they've changed it as well at some point)

so, +1 for the change, we should change it on android then as well (delta-ios uses system language selector)

PR #3340, opened 2023-08-04
changed: make verified icon green

Simon-Laux:


r10s (approved):
lgtm, in chatlist and title, the new icon seems to be a bit less sharp than the old one (which was a png optimised for small sizes), however, still good enough

PR #3345, opened 2023-08-05
prepare 1.39.0

Simon-Laux:
- prepare 1.39.0
- add device message

### DIFF
--- a/src/renderer/deviceMessages.ts
+++ b/src/renderer/deviceMessages.ts
@@ -33,4 +33,14 @@ Full Changelog: https://github.com/deltachat/deltachat-desktop/blob/master/CHANG
 
 Full Changelog: https://github.com/deltachat/deltachat-desktop/blob/master/CHANGELOG.md#1_38_0`
   )
+
+  BackendRemote.rpc.addDeviceMessage(
+    accountId,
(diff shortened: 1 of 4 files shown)


src/renderer/deviceMessages.ts
```diff
@@ -33,4 +33,14 @@ Full Changelog: https://github.com/deltachat/deltachat-desktop/blob/master/CHANG
...
+
+  BackendRemote.rpc.addDeviceMessage(
+    accountId,
+    'changelog-version-1.39.0-version0',
+    `What's new in 1.39.0?
+
+✨ We fixed many small bugs and improved stability
+
+Full Changelog: https://github.com/deltachat/deltachat-desktop/blob/master/CHANGELOG.md#1_39_0`
+  )
```
farooqkz, line 45: I personally don't like this message. We could at least tell them what fixes we did. Additionally, shouldn't this be multilingual and from transifex?

farooqkz (changes requested):

src/renderer/deviceMessages.ts
Simon-Laux: there is nothing very big in this release, so better no message

PR #3346, opened 2023-08-06
first prototype of highlight qr codes in fullscreen image view

Simon-Laux:
- [x] icon button to toggle
- [ ] handle window resize (ideally only change scaling without looking for qrcodes again) - test in both directions height and width
- [ ] correct positioning -> still has bugs getting the correct scaling sometimes
- [ ] nice design
- [ ] context menu -> does not open, somehow the other context menu blocks it (maybe the pan library?? or react?)
- [x] click to open link
- [ ] if user opens internal link then close the fullscreen media view when user takes an action that justifies it. (join group, create new account and so on)
- [ ] for the enum in the wasm qr code scanner I might need it in another serde variant  representation, ts does not like the current one
- [ ] bonus: 3D perspective for qr code highlight
- [ ] better detection with other or hybrid (use 2 libs at the same time) library
- [ ] worker thread (don't block ui)

### DIFF
--- a/bin/build-frontend-ts.mjs
+++ b/bin/build-frontend-ts.mjs
@@ -52,6 +52,10 @@ async function bundle(production, minify = false) {
     'node_modules/@deltachat/message_parser_wasm/message_parser_wasm_bg.wasm',
     'html-dist/message_parser_wasm_bg.wasm'
   )
+  await copyFile(
+    'node_modules/quircs-wasm/quircs_wasm_bg.wasm',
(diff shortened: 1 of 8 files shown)


hpk42:
could you state a little summary/rationale for closing?

Simon-Laux:
> could you state a little summary/rationale for closing?

out of scope for now. Qr scanner is not good enough yet in my opinion. Also see todos.

PR #3348, opened 2023-08-07
Introduce sub-levels & make "mute chat" options configurable in context menu

### LINKED ISSUE #3116: Require less clicks to mute the chat
Currently in Delta Chat Desktop to mute the chat one has to
1) right-click the chat in the chatlist
2) select "mute notifications"
3) select radio button
4) click ok

Compare this to Signal Desktop which uses submenus for options:

maxphilippov:
Tries to address #3116 

- ContextMenu now handles sublevels
- ContextMenuItem has subitems
- New icon for ContextMenuItem with subitems

closes #3116

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 - ContextMenu now handles multiple sublevels #3116
   - ContextMenuItem has a subitems array
   - Hovering outside of the menu closes last opened sublevel
-  - ChatListContextMenu mute option now opens a submenu with duration options insead of a dialog
+  - ChatListContextMenu mute option now opens a submenu with duration options instead of a dialog
 
 ### Changed
 - Update `deltachat-node` and `deltachat/jsonrpc-client` to `v1.132.1`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,16 @@
 
 ## [Unreleased][unreleased]
 
+### Added
+- ContextMenu now handles multiple sublevels #3116
+  - ContextMenuItem has a subitems array
+  - Hovering outside of the menu closes last opened sublevel
+  - ChatListContextMenu mute option now opens a submenu with duration options insead of a dialog
+
 ### Changed
-- update deltachat-node and deltachat/jsonrpc-client to `v1.132.1`
+- Update `deltachat-node` and `deltachat/jsonrpc-client` to `v1.132.1`
+- Update `electron` from `v26.6.0` to version `v26.6.3`
+- Update `filesize` dependency from `8.0.6` to `10.1.0`
 - Use native watch mode and CSS modules of esbuild #3571
   - Use native watch mode and CSS modules of esbuild
   - Support SCSS in modules
@@ -14,17 +22,13 @@
 - Refactor dialog components, improved styling for settings #3575
 - Refactor and improve design of search input components #3572
 - Refactor dialog logic and add additional helper methods #3512
-- Upgrade `filesize` dependency from `8.0.6` to `10.1.0`
 - Prefer light theme for the help and webxdc loading pages
 - Helper method to easily use confirmation dialogs #3601
 - Refactor using new `useConfirmationDialog` hook #3602
-- update `electron` from `v26.6.0` to version `v26.6.3`
 
 ### Fixed
 - Silently fail when notifications are not supported by OS #3613
 - Fix uncaught Exception when dismissing notifications on windows #3593
-
-### Fixed
 - Introduce own React context for context menus & fix regression #3608
 
 <a id="1_42_2"></a>
@@ -140,7 +144,7 @@
 > If you are using DC on these platforms you have the following options:
 > - Keep using 1.40.4
 > - update windows to 10 or 11
-> - switch to linux (or dual boot it) 
+> - switch to linux (or dual boot it)
 > - contact delta@merlinux.eu if you have money and want to sponsor a special legacy build of deltachat desktop 1.41-1.42.
 >
 > - Core is now built on Debian 10: now requires glibc 2.28, ubuntu 18 is not supported anymore.
@@ -189,7 +193,7 @@
 - fix webxdc title not updated in document title changes #3412
 - fix: remove duplicated search button on "search in chat" #3014
 - fix "Verified by" is shown weirdly for contacts that were verified directly #3421
-- open help, webxdc and html email windows with always on top flag, if main window has that flag. 
+- open help, webxdc and html email windows with always on top flag, if main window has that flag.
 - fix copy text inserts extra linebreaks
 - improve video message - wide enough to show controls
 - gallery: fix scroll to top when switching tabs
@@ -637,7 +641,7 @@ Also make opening devtools with F12 more reliable.
 - prevent double context menu on macOS
 - fix setting color chat background color #2659
 - Fix that results of search in chat are not ordered by newest first
-- Fix messagelist overscrolling #2956 
+- Fix messagelist overscrolling #2956
 - Fix messagelist not jumping to correct message #2953
 - fix chat audit log #2967
 - fix that results of search in chat are not ordered by newest first
@@ -813,7 +817,7 @@ Also make opening devtools with F12 more reliable.
 - fixed height for webxdc icons in messages
 - Better empty gallery tab messages
 - allow internal webxdc navigation (multiple html files)
-- fix random search result position (#2631) 
+- fix random search result position (#2631)
 - fix: signature field in Edit Profile is not labeled when empty (#2579)
 - Keyboard shortcut preview for "Send on Enter" setting
 - Keyboard shortcut reference dialog (cheatsheet), accessible via `Ctrl + /` or `Cmd + /`
@@ -956,11 +960,11 @@ Also make opening devtools with F12 more reliable.
 - Fix connectivity view title @IrvinLara9 #2480
 - Do not double log core events
 - Fix Bulgarian language name (uppercase first letter)
-- Fix signature text styling @ejgonzalez17 
+- Fix signature text styling @ejgonzalez17
 - Fix missing image formats @ejgonzalez17
 - momentjs isn't localized on first startup
 - Fix connectivity status hiding composer @trujillo9616
-- Fix overflow in Confirmation Dialog @Abhijnan-Bajpai 
+- Fix overflow in Confirmation Dialog @Abhijnan-Bajpai
 - Fix the profile picture removal @cavesdev #2472
 - Fix the horizontal scroll in autocrypt dialogs @cavesdev #2277
 - Add logging in by pressing enter in AccountSetupScreen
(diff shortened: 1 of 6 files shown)


src/renderer/components/ContextMenu.tsx
```diff
@@ -1,15 +1,24 @@
...
+import classNames from 'classnames'
 import { useRef } from 'react'
 import { ScreenContext, unwrapContext } from '../contexts'
 
-export type ContextMenuItem = { label: string; action: () => void }
+export type ContextMenuItem = {
+  label: string
+  action: () => void
+  subitems?: (ContextMenuItem | undefined)[]
+}
```
Simon-Laux, line 10: I think it makes sense to make `action` and `subitems` mutually exclusive:
I did not test it, but sth like this should do the trick
```suggestion
export type ContextMenuItem = 
{ label: string } & (
 {action: () => void} | 
 {subitems?: (ContextMenuItem | undefined)[]}
)
```

src/renderer/components/ContextMenu.tsx
```diff
@@ -1,15 +1,28 @@
 import React, { useState, useEffect, useLayoutEffect, useContext } from 'react'
+import classNames from 'classnames'
 import { useRef } from 'react'
 import { ScreenContext, unwrapContext } from '../contexts'
 
-export type ContextMenuItem = { label: string; action: () => void }
+type ContextMenuItemActionable = { action: () => void; subitems?: never }
+type ContextMenuItemExpandable = {
+  action?: never
```
Simon-Laux, line 8: I think this will mean `never | undefined`, better change `?:` to just `:`
same for subitems above.

maxphilippov: This won't work. I don't remember exact reasoning, but I did try it and from the error output I believe it works like this: since it's impossible to assign any value to a `never` type compiler will never pick `ContextMenuItemActionable` and it then complains about either `subitems` not specified in cases where it's absent or that it cannot assign a function to `never` in case where `action` is specified. It's a common trick to use optional never for this.

src/renderer/components/ContextMenu.tsx
```diff
@@ -197,15 +247,20 @@ export function ContextMenu(props: {
     }
```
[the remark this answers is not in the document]
Simon-Laux: switch cases in js are ugly and easy to mess up if you forget the break statement for example.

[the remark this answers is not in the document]
maxphilippov: I went with the original code style of else-if, me personally I have an issue with switch not having a scope for each case by default. But in this case I'm fine with both options.

src/renderer/components/ContextMenu.tsx
```diff
@@ -147,27 +143,68 @@ export function ContextMenu(props: {
...
+  // always one item less than menuLevelEls
+  const [openSublevels, setSublevels] = useState<number[]>([])
+  // Which one of the last sublevel items the keyboard is focused on
+  const keyboardFocus = useRef<number>(-1)
+
+  let items = props.items.filter(val => val !== false) as ContextMenuItem[]
+
+  const levelItems: ContextMenuLevel[] = [{ items }]
+
+  for (let i = 0; i < openSublevels.length; ++i) {
```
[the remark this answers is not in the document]
maxphilippov: Makes sense, thank you. I tend to go for the implicit index by default and change it later.

src/renderer/components/ContextMenu.tsx
```diff
@@ -147,27 +143,68 @@ export function ContextMenu(props: {
...
+    levelItems.push({
+      items,
+    })
+  }
+
   useLayoutEffect(() => {
-    if (!menuRef.current) {
-      throw new Error()
+    if (menuLevelEls.current.length == 0) {
+      throw new Error('Somethings wrong with menu elements')
```
adzialocha, line 170: There's a mini typo here ("something IS" and I'd add a "the" before "menu elements").

Or, how about being a bit more explicit as well? For example: "No context menu elements available for display"?

src/renderer/components/ContextMenu.tsx
```diff
@@ -197,15 +247,20 @@ export function ContextMenu(props: {
...
+      for (const menuEl of menuLevelEls.current) {
+        isOnMenu = ev.target === menuEl
+        if (!isOnMenu) {
+          for (const child of menuEl.children) {
+            if (ev.target === child) {
+              isOnMenu = true
+              break
+            }
+          }
+        }
```
adzialocha, line 261: This is a fairly high degree of nesting which makes the code hard to read, do you mind refactoring it if possible?

src/renderer/components/ContextMenu.tsx
```diff
@@ -228,32 +283,55 @@ export function ContextMenu(props: {
...
+                  item.action()
+                }
+              }}
+              onMouseOver={() => {
+                if (item.subitems) {
+                  setSublevels(l => [...l.slice(0, levelIdx), index])
+                } else {
+                  setSublevels(l => l.slice(0, levelIdx))
+                }
+              }}
```
adzialocha, line 325: How about moving the event handler functions outside of the React render method?

It's easier to read when the view- and control layer is separated, then I can read "what the component does" in one place - and "how it looks" in another. This also gives a nice opportunity to name our functions with something descriptive which helps me to understand what this is doing.

src/renderer/components/chat/ChatListContextMenu.tsx
```diff
@@ -110,7 +109,6 @@ export function useChatListContextMenu(): {
         openLeaveChatDialog(openDialog, chatListItem.id)
       const onBlockContact = () =>
         openBlockFirstContactOfChatDialog(openDialog, chatListItem)
-      const onMuteChat = () => openMuteChatDialog(openDialog, chatListItem.id)
```
adzialocha, line 113: We probably want to use this change in the `ThreeDotMenu` component as well and then I think we can remove the then unused `openMuteChatDialog` method and then dialog component altogether?

src/renderer/components/ContextMenu.tsx
```diff
@@ -182,12 +219,25 @@ export function ContextMenu(props: {
...
+        keyboardFocus.current = openSublevels[openSublevels.length - 1]
+      } else if (ev.key == 'ArrowRight') {
+        if (current) {
+          const index = +((current as HTMLDivElement).dataset
+            .expandableIndex as string)
+          if (!isNaN(index)) {
+            setSublevels(l => [...l, index])
+            keyboardFocus.current = 0
+          }
+        }
```
adzialocha, line 233: I find this code a little hard to read. Some suggestions for a refactoring:

1. Every time you change the value of `sublevels` some slicing, spreading etc. is used which are hard to read. Why not move these `setSublevels` into separate methods and give them an explicit names what they are doing? For example:
    ```js
    const expandSubMenu = () => {
      // ...
    }
    
    const collapseSubMenu = () => {
      // ...
    }
    ``` 
    This approach might also fit nicely into the suggestion below to move the `onClick` etc. event handler logic into explicit methods (for example `closeMenu`) as well. They could all live next to each other and show how we can interact with this "sublevel" data structure
2. Type annotations and code in the same line can be hard to read, I'd separate things like this into separate lines, otherwise one almost misses the actual crucial bit which is that this type got converted by the `+`, which I would also make explicit
    ```js
    const index = +((current as HTMLDivElement).dataset
      .expandableIndex as string)
    
    // vs.
    
    const current = parent?.querySelector(':focus') as HTMLDivElement
    const index = parseInt(current.dataset.expandableIndex, 10)
    
    // .. or similar
    ```

CHANGELOG.md
```diff
@@ -17,6 +16,10 @@
 - Prefer light theme for the help and webxdc loading pages
 - Helper method to easily use confirmation dialogs #3601
 - Refactor using new `useConfirmationDialog` hook #3602
+- ContextMenu now handles multiple sublevels (#3116)
+  - ContextMenuItem has a subitems array
+  - Hovering outside of the menu closes last opened sublevel
+  - ChatListContextMenu mute option now opens a submenu with duration options insead of a dialog
```
adzialocha, line 22: Typo "insead"

adzialocha (changes requested):
Very nice addition! I've tested it a bit and it works & looks good.

My review consists mostly of readability nitpicks, I think with some small changes the code could be a little bit more readable. Consider them more as suggestions.

More crucially maybe is bringing this change also into the "three dot menu" in the upper right corner? Then we can remove the unused code which is always a good thing.

Simon-Laux:
> More crucially maybe is bringing this change also into the "three dot menu" in the upper right corner? Then we can remove the unused code which is always a good thing.

The ThreeDotMenu is already a context menu, the component exists just to separate logic (maybe not necessary) and to change the coordinates so it opens always at exactly the same place.

src/renderer/components/chat/ChatListContextMenu.tsx
Simon-Laux: ah now I get what you mean, sure that's a good idea.

src/renderer/components/ContextMenu.tsx
Simon-Laux: I like the parseInt solution more, it is more clear than the `+` which is close to ninja code

adzialocha:
@maxphilippov Small "force push" warning from me 🐱 - I've did it to bring this PR back to the latest version after touching the context menu stuff in another PR

src/renderer/components/ContextMenu.tsx
farooqkz: How about using `map` and/or `filter`? I think they are also faster?

src/renderer/components/ContextMenu.tsx
maxphilippov: Thanks for the feedback. I do agree with other things, but this one item I'd argue is better the way it is right now.

To provide a handler I'd have to do write a function producing another function just to pass additional parameters like this:
```
const makeMenuClickHandler = (
    item: ContextMenuItemActionable | ContextMenuItemExpandable,
    index: number,
    levelIdx: number
  ) => {
    return (ev: React.MouseEvent) => {
      if (item.subitems) {
        expandMenu(index, levelIdx)
        keyboardFocus.current = 0
...
```

And I feel it's a bit too much of an indirection. I do agree with adding helper functions expandSubMenu and collapseSubMenu, but that alone already add one step of indirection. What I like about current code it's in one place and it's clear about what's happening:
1. We iterate over a list of menu sublevels and then over the items to add these to DOM
2. Based on mouse input we either cut or extend this list. onMouseOver is pretty clean in that regard, onClick is more convoluted based on keyboard highlight behavior and how we handle the event, but I find it easier to digest in terms of how it all works and not what we were intending to do.

As a new contributor I'm obviously open to change it if it doesn't fit the general code style, I just feel that for now it's a better way to handle this exact case.

src/renderer/components/chat/ChatListContextMenu.tsx
maxphilippov: I did the ThreeDotMenu part, I also did a quick fix for the submenus that don't fit on the right side of the screen, but since we don't have any that might not fit on left side I didn't bother with that for now and I feel like it's a bit more complicated issue if we have couple of sublevel that don't fit, so I just went for the easiest thing.

Simon-Laux (approved):
Works, code LGTM, I think it's ok to merge now, we can still discuss refactoring outside of this pr, but I think it makes sense to not endlessly discuss these things as long as it works, there are other areas where our time is better invested ;)

Issue #3178, opened 2023-04-11
add-second-device strings missing


r10s:
the idea of the "Add as Second Device" button in the "Welcome" screen is to let ppl not run into problems when setting up a second device.

while the general functionality is there and working nicely, a hint is missing what ppl should do. currently the screen looks like the following:

<img width=360 src=https://user-images.githubusercontent.com/9800740/231149547-4b15208b-d1f3-4814-a41e-542ca3260857.png>

unless ppl have read instructions somewhere else, they are left completely alone.

as a minimum, we should replace the title and the text below the qr code with `multidevice_receiver_title` resp. `multidevice_open_settings_on_other_device`. ftr, is how android/ios look:

<img width=550 src=https://user-images.githubusercontent.com/9800740/231152194-02fee36d-f643-4cfe-a73e-b84198b8908a.jpg>

related: in the subsequent progress dialog, the title string is missing:

<img width=360 src=https://user-images.githubusercontent.com/9800740/231152964-2fa6afb1-7643-43b1-a2fe-aed90e89c50d.png>

that should be `multidevice_receiver_title` as well.

r10s:
i assume, the milestone is not a prerequisite? i'd change these strings as soon as possible, not a blocker and maybe not the reason for a 1.36.1+, but still should be part of it when it is needed for other reasons

to be cleaer:  no need to change the whole layout, just change the hint would already be a huuuuge ux improvement

Simon-Laux:
the problem is the qr code scanner (which is inside of a dialog), we would need a custom ui there and if we're already doing that then we can do the new scanner as well, it's not that hard.

PR #3349, opened 2023-08-09
Add translation for receive_backup during multi device setup

ralphtheninja:
Fixes

image_2023-08-09_14-59-17

### DIFF
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -56,6 +56,9 @@
   "show_webxdc_in_chat": {
     "message": "Show Private App in Chat"
   },
+  "receive_backup": {
+    "message": "Receiving backup from second device"
+  },
   "save_sticker": {
     "message": "Save Sticker"
   },
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Changed
 
 ### Fixed
+- fix missing translation string in setup second device progress dialog
 
 
 <a id="1_39_0"></a>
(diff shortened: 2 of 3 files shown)


_locales/_untranslated_en.json
```diff
@@ -56,6 +56,9 @@
   "show_webxdc_in_chat": {
     "message": "Show Private App in Chat"
   },
+  "receive_backup": {
+    "message": "Receiving backup from second device"
```
ralphtheninja, line 60: Maybe some other wording is better? Idk.

farooqkz: Dunno about other wording but "a second device" is correct.

r10s (changes requested):
no need to add a new string, there is already a string, but that is just not used on desktop.

so instead of adding a new string, use `multidevice_receiver_titlle` instead of `receive_backup` as described in https://github.com/deltachat/deltachat-desktop/issues/3178

things are in sync with android/ios then as well.

maybe also target the other open points of https://github.com/deltachat/deltachat-desktop/issues/3178 when on that?

Simon-Laux:
> maybe also target the other open points of https://github.com/deltachat/deltachat-desktop/issues/3178 when on that?

that is a big bigger change as it needs a new qr code dialog.

ralphtheninja:
> no need to add a new string, there is already a string, but that is just not used on desktop.
> 
> so instead of adding a new string, use `multidevice_receiver_titlle` instead of `receive_backup` as described in #3178
> 
> things are in sync with android/ios then as well.
> 
> maybe also target the other open points of #3178 when on that?

I can tweak this PR to do that instead.

r10s:
> that is a big bigger change as it needs a new qr code dialog.

i do not really buy that changing a string requires a larger refactoring 🧐 :)

did a pr that fixes the long standing issue and adds missing guidance so that ppl starting "Add Second Device" from the welcome screen have a  chance to succeed: https://github.com/deltachat/deltachat-desktop/pull/3350

Simon-Laux:
> @Simon-Laux however, see that as an experiment from my side, feel free to close or do sth else as you like 😀

I might be too much of an perfectionist, but I can see that this basic subtitle already improves things a lot, so thanks for the pr.

PR #3356, opened 2023-08-16
Fix Gallery errors cannot be closed

### LINKED ISSUE #3339: old errors eat up screen space
if, for whatever reason, some files are missing in the blob directory, all these error is shown permanently atop of the gallery, with no way for the user to get rid of that - even if the files in question and the corresponding error are several years old.

to be clear: the error is real, however the error should not stick forever atop of the gallery; instead, just show them at the place of the image - that way, the user also get some context.

btw. if the list becomes "virtual" (so, load images as needed on scrolling, i think this is planned for the global gallery), this would also be the natural behaviour - as you just do not know that image # 1000 is missing when showing the first tens.

maybe core could also help somehow, however, eg. altering the database in case of partly errors may also be dangerous. and even then, in any case it is better to show the error in place in the UI (as also eg. done on the web).

farooqkz:
This PR fixes #3339 

The error in Gallery when shown has two buttons to close the error, or to close it and never show it again for this chat. If the latter is chosen, a property in desktop settings is set to `true` for this chatId so it won't be shown for this chat ever again.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Fixed
 
+ - Fixed error in Gallery cannot be closed(#3339)
 
 <a id="1_39_0"></a>
 
--- a/src/renderer/components/Gallery.tsx
+++ b/src/renderer/components/Gallery.tsx
@@ -81,7 +81,7 @@ export default class Gallery extends Component<
       showErrors: true,
     }
     runtime.getDesktopSettings().then((settings: DesktopSettingsType) => {
-      let showErrorsSettings = settings.DontShowMissingFilesError
+      const showErrorsSettings = settings.DontShowMissingFilesError
       if (showErrorsSettings && showErrorsSettings[props.chatId]) {
         this.setState({ showErrors: false })
       }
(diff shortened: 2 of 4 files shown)


Simon-Laux:
> to be clear: the error is real, however the error should not stick forever atop of the gallery; instead, just show them at the place of the image - that way, the user also get some context.

The solution is showing errors in place of the images, showing them on the top was just a workaround because I was lazy sometime ago.

Also storing message ids in `config.json` is not that good, we want to avoid to save account specific stuff there.

If you want to do this then you can take https://github.com/deltachat/deltachat-desktop/pull/3255 over (or better do a new  pr based on it that could request being merged into my pr branch.), there is not much left and you could learn the virtual list stuff with react window (grids). I thing the solution is to just put the errors back to the images and other media entries.

farooqkz:
> > to be clear: the error is real, however the error should not stick forever atop of the gallery; instead, just show them at the place of the image - that way, the user also get some context.
> 
> The solution is showing errors in place of the images, showing them on the top was just a workaround because I was lazy sometime ago.
> 
> Also storing message ids in `config.json` is not that good, we want to avoid to save account specific stuff there.
> 
> If you want to do this then you can take #3255 over (or better do a new pr based on it that could request being merged into my pr branch.), there is not much left and you could learn the virtual list stuff with react window (grids). I thing the solution is to just put the errors back to the images and other media entries.

Can you draw a sketch on how the errors should be shown?

Simon-Laux:
> Can you draw a sketch on how the errors should be shown?

the errors are about images that failed to load, so where an image failed to load there should be an "image broken" or error icon with the error message or a shorter error that you can hover over to see the full error.
Same thing in principle for the other media types.

PR #3360, opened 2023-08-23
correct way of showing gallery errors

farooqkz:
@Simon-Laux  This adds the correct way of showing gallery errors to your PR branch.

image

There is also a tooltip.


farooqkz:
Maybe it's better to merge into master and after that you rebase?

Simon-Laux (changes requested):
it is shown at the end, not inline where the image should have been.
Also I'm not sure about the big red cross as design.

farooqkz:
> it is shown at the end, not inline where the image should have been. Also I'm not sure about the big red cross as design.

I will solve the former. Regarding the latter, is it the colour or the shape or both?

Simon-Laux:
I have an idea:

First an icon depending on media type:
- image: https://fonts.google.com/icons?selected=Material%20Symbols%20Outlined%3Aimage_not_supported%3AFILL%400%3Bwght%40400%3BGRAD%400%3Bopsz%4048
- video: https://fonts.google.com/icons?selected=Material+Symbols+Outlined:videocam_off:FILL@0;wght@400;GRAD@0;opsz@48&icon.query=image
- audio: https://fonts.google.com/icons?selected=Material+Symbols+Outlined:videocam_off:FILL@0;wght@400;GRAD@0;opsz@48&icon.query=image

under this icon a translated text "Failed to load" or some similar text, on hover (title attribute) there could be the full error text, also we could think of showing an dialog with the error text when a user clicked on it.


Things to note:
- download the icons as svg, put them in the images/icons folder and mention them inside of images/licenses
- audio and files have a different design compared to image and video which both are squares.
- `alert()` does not exist in electron, so the dialog would need to use our dialog system.

r10s:
i agree with simons suggestion about the image, the red cross is too bold. however, a single "missing" image will do, no need to provide, test etc. 3 different images for an error cornercase. the first image is good enough and can be read as "media missing"

farooqkz:
> it is shown at the end, not inline where the image should have been. Also I'm not sure about the big red cross as design.

How can I find where the image should be?

Simon-Laux:
> How can I find where the image should be?

you need to change more of the code. the errors are extracted from the result, you need to remove that code and have successes and errors in the same array.

PR #3362, opened 2023-08-23
simon/webxdc titlebar menu

Simon-Laux:
- fix: show webxdc titlebar also on mac -> make show sourcecode link accessible on macOS
- fix: hide devtool option in webxdc titlebar menu when it is not enabled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Changed
 
 ### Fixed
+- fix: show webxdc titlebar also on mac -> make show sourcecode link accessible on macOS
 
 <a id="1_40_0"></a>
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,14 @@
 ## [Unreleased][unreleased]
 
 ### Added
+- add: "always on top" option to webxdc titlebar menu.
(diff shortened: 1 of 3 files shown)


r10s:
nice!

"Delta Chat / Quit",  however, seems to close the webxdc only, not the app now. this is unusual. convention for "Quit" in the app menu is to exit the whole app.

so, i suggest to leave "Delta Chat / Quit" as before and instead add "File / Close" (or more concrete "File / Close Webxdc Window"

Simon-Laux:
> "Delta Chat / Quit", however, seems to close the webxdc only, not the app now. this is unusual. convention for "Quit" in the app menu is to exit the whole app.

Only on macOS or on other platforms too?

r10s:
> Only on macOS or on other platforms too?

windows usually does not have "Quit" but "Exit", however, as the menu is used more per-window, this is less of an issue there (not sure if "Exit" even exists on Delta Chat Windows.

for linux: yes, i'd also say there "Quit" means "exit the app". we can be more on-point by renaming it to "Quit Delta Chat" vs. "Close Window" - this is also how things usually read on mac

r10s (approved):

PR #3364, opened 2023-08-28
fix contact creation in new chat dialog not working with leading or trailing spaces.

### LINKED ISSUE #3357: fix: New email adresses should be stripped
When adding an email address that has a leading whitespace, nothing happens when clicking on "New Contact". From the console, it can be taken that an invalid email address was given to the core. If this is the case, then an error message should probably be shown. For this issue to close it is enough to just strip the entered email address though.

Simon-Laux:
closes #3357

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Changed
 
 ### Fixed
+- fix contact createion in new chat dialog not working with leading or trailing spaces. #3357
 
 <a id="1_40_0"></a>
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 ### Changed
 
 ### Fixed
+- fix contact createion in new chat dialog not working with leading or trailing spaces. #3357
```
Simon-Laux, line 10:
```suggestion
- fix contact creation  in new chat dialog not working with leading or trailing spaces. #3357
```

PR #3368, opened 2023-08-31
remove jitsi as a default Video Chat instance

### LINKED ISSUE #3366: Disable meet.jit.si as a Video Chat instance
As of now, in the experimental video chat settings, we offer 3 default instances for video chats: vc.autistici.org, meet.systemli.org, and meet.jit.si.

Since August 24, meet.jit.si requires authentication for creating rooms: https://jitsi.org/blog/authentication-on-meet-jit-si/

This hurts our workflow of "just setup a call with one click" a lot. I suggest removing them as a default video chat provider.

It would be nice to offer a 3rd default option though. We could offer https://jitsi.random-redirect.de, which redirects to one of 113 jitsi servers: https://github.com/tosterkamp/random-redirect/blob/develop/res/jitsi_servers.lst
I don't know after which criteria it is curated though, e.g. meet.jit.si is still in the list. Still an improvement though.

Simon-Laux:
because they added a sign-in requirement

closes #3366

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 ### Changed
+- remove jitsi as a default Video Chat instance, because they added a sign-in requirement #3366
 
 ### Fixed
 - fix contact creation in new chat dialog not working with leading or trailing spaces. #3357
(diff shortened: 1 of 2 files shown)


missytake (approved):
It would be nicer to replace it with another one, but good enough for now.

farooqkz (approved):

PR #3374, opened 2023-09-01
prepare 1.40.1

Simon-Laux:


Simon-Laux:
maybe we'll wait for @achou11's sticker pr to get merged before releasing?

PR #3395, opened 2023-09-12
Offer to copy non-HTTP links to the clipboard instead of trying to open them in webxdc source code link and inside of html emails.

Simon-Laux:
- [x] test it

### DIFF
--- a/src/main/windows/html_email.ts
+++ b/src/main/windows/html_email.ts
@@ -17,6 +17,7 @@ import { open_url } from '../open_url'
 import { loadTheme } from '../themes'
 import { getDCJsonrpcClient } from '../ipc'
 import { getLogger } from '../../shared/logger'
+import { clipboard } from 'electron/common'
 
 const log = getLogger('html_email')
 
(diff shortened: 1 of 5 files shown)


src/main/windows/html_email.ts
```diff
@@ -387,7 +395,24 @@ function makeBrowserView(
       if (url.startsWith('mailto:')) {
         open_url(url)
       } else {
-        shell.openExternal(url)
+        if (
+          url.startsWith('http:') ||
+          url.startsWith('https:') ||
+          url.startsWith('mailto:')
+        ) {
+          shell.openExternal(url)
```
r10s, line 403: in  the bubbles, `mailto:` links are passed to  `processOpenQrUrl(link)` instead of `shell.openExternal()` - is that difference on purpose?

in general, bubble-view and full-view should behave quite similar from the user perspective - a link fit randomly to the bubbles or not, but they should act the same (it is a pity that we cannot use the same dialogs in both szenarios, btw)

Simon-Laux: just a duplicated artefact that comes from copy pasting see 2 lines above. also `processOpenQrUrl` exist only in the frontend, so the `open_url` function is for forwarding the request to the frontend.

PR #3398, opened 2023-09-13
Show an unread badge when there are unread notifs from other accounts

farooqkz:
This PR adds an unread badge next to hamburger icon and "switch account" button in side bar when there are unread notifications in other account(s).


farooqkz:
It turns out I've included an unwanted commit. Will open another PR...

farooqkz:
> It turns out I've included an unwanted commit. Will open another PR...

Reverted :D

src/renderer/components/UnreadBadge.tsx
Simon-Laux, line 1: if this is only used for the sidebar move better move this stuff in the sidebar file (

src/renderer/components/screens/MainScreen.tsx
```diff
@@ -42,14 +44,46 @@ import SettingsStoreInstance, { useSettingsStore } from '../../stores/settings'
...
+    }
+    update()
+
+    BackendRemote.rpc
+      .getAllAccountIds()
+      .then(accountIds =>
+        accountIds.map(id => onDCEvent(id, 'IncomingMsg', update))
+      )
+
+    return update
```
Simon-Laux, line 84: there is already much happening in the MainScreen file, I propose you move the code to the unread badge component and rename that badge to sth like OtherAccountsUnreadMessagesBagde or sth that reflects that it shows that it's about wether the other accounts have unread messages.

I think the small runtime code duplication is not an issue, also we could add a timed debounce to limit how often it is updated. (instead of only making sure it is only run once at a time)

Simon-Laux (changes requested):
- [x] the check if "sync all accounts" is activated is missing, it is always shown, which does not make sense if only the active account is synced the unread count of the others can never change
- [ ] I would like to have the unread badge rather on the right in the sidebar and displaying the count of unread messages (the badge on the hamburger menu would be to small for a number, but in the sidebar is enough space for it)
- [ ] fix event listener leak
- [x] move special animation css colors to themebase if we want them
- [ ] move update code into unread badge itself and have two variants (one small one for the hamburger menu and a bigger one with number for the sidebar.)

src/renderer/components/OtherAccountsUnreadBadge.tsx
```diff
@@ -0,0 +1,62 @@
+import React, { useEffect, useState } from 'react'
+import { BackendRemote, onDCEvent } from '../backend-com'
+import { selectedAccountId } from '../ScreenController'
+import { runtime } from '../runtime'
+import { DesktopSettingsType } from '../../shared/shared-types'
+
+// Tip of the Day: Keep as many ice creams as you can in your fridge
+// for the ice cream outage days. FREE ICE CREAM FOR EVERY1! --Farooq
```
Simon-Laux, line 8: funny easteregg

src/renderer/components/OtherAccountsUnreadBadge.tsx
```diff
@@ -0,0 +1,62 @@
...
+    BackendRemote.rpc
+      .getAllAccountIds()
+      .then(accountIds =>
+        accountIds.map(id => onDCEvent(id, 'IncomingMsg', update))
+      )
+
+    return update
+  }, [isSyncAllEnabled])
+
+  if (haveOtherAccountsUnread) {
```
Simon-Laux, line 57:
```suggestion
  if (isSyncAllEnabled && haveOtherAccountsUnread) {
```

src/renderer/components/screens/MainScreen.tsx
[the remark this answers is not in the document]
farooqkz: Can you explain more please?

src/renderer/components/OtherAccountsUnreadBadge.tsx
```diff
@@ -0,0 +1,57 @@
...
+    update()
+
+    BackendRemote.rpc
+      .getAllAccountIds()
+      .then(accountIds =>
+        accountIds.map(id => onDCEvent(id, 'IncomingMsg', update))
+      )
+
+    return update
+  }, [settings])
```
Simon-Laux, line 50:
```suggestion
  }, [settings?.desktopSettings.syncAllAccounts])
```

that settings is not set is rare, we are more interested in wether `syncAllAccounts` changed

farooqkz:
@Simon-Laux I think I've solved all the problems except for the listener leak thing. I will be grateful if you do a suggestion so that I can learn how should I handle this as it's my first time using `useEffect`. Thanks in advance.

src/renderer/components/Sidebar.tsx
```diff
@@ -181,7 +181,10 @@ const Sidebar = React.memo(
           </div>
           <div key='logout' className='sidebar-item' onClick={onLogout}>
             {tx('switch_account')}
-            <OtherAccountsUnreadBadge top='-38px' left='-14px' />
+            <OtherAccountsUnreadBadge
+              style={{ top: '-38px', left: '110px' }}
```
Simon-Laux, line 185: I would prefer having these offsets in they (s)css unless the are either for prototyping or dynamic.

src/renderer/components/screens/MainScreen.tsx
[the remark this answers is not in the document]
Simon-Laux: https://react.dev/reference/react/useEffect#useeffect

the return value of the use effect function/callback must be a callback that does cleanup, in this case a callback that removes the event listeners again.

src/renderer/components/Sidebar.tsx
farooqkz: The thing is that the parent component has to determine the offsets according to its use. The offsets are different for the hamburger menu icon and in the sidebar.

Simon-Laux:
Anyways I would postpone this as there is no clear consensus of the other dc team members if this feature is actually effective enough for it's intended use case.

hpk42:
Not sure myself -- would need to use it for a while to know. 
Didn't follow much discussion here -- was there a support forum thread for it?  or some other discussed/articulated need?

farooqkz:
> Not sure myself -- would need to use it for a while to know. Didn't follow much discussion here -- was there a support forum thread for it? or some other discussed/articulated need?

It was purely my idea to enhance multi account usage. Wasn't discussed or requested in the forum.

Simon-Laux:
Let's rather go for the full sidebar solution soon. closing this for now

PR #3409, opened 2023-09-19
fix `target=_blank` links in html emails don't work

### LINKED ISSUE #3408: some links in full-msg-html-view not working
it happens from time to time to me, that some links are not working in the "Show full message..." view - where the same message's links work on iOS (not tested android in this case)

here is an example as an `.eml` file: https://gist.github.com/r10s/19f027782070c1688ece74621332dc73 (the 3D= is the quoted printable, core should clean this up for ui), the rendered, not working links look as follows:

thunderbird, apple-mail, delta-chat-ios all work. also, you can copy+paste the link from deltachat-deskop - but you cannot open it.

the mail is generated by https://aktion.campact.de/verkehr/49-euro-ticket/teilnehmen , maybe they're doing sth wrong, however i have seen the issue also before in other contexts, so even then, deltachat-desktop should probably be more graceful

Simon-Laux:
fixes #3408

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 - fix clipboard not working in webxdc apps
+- fix `target=_blank` links in html emails don't work #3408
 
 <a id="1_40_4"></a>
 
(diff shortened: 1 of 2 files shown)


r10s (approved):
yeah! works like a charm now! thanks for the quick fix.

code lgtm as far as i can tell - and i also tested a bit and played around. the mail in question works now :)

PR #3411, opened 2023-09-24
style: add dark theme for help and webxdc wrapper

WofWca:
With webxdc it affects the "loading" stage.

I tested both light and dark mode for both the help page and the webxdc. Both look ok.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ### Changed
 - update deltachat-node and deltachat/jsonrpc-client to `v1.123.0`
+- add a dark theme for the "Help" and the webxdc loading screen
 
 ### Fixed
 - fix clipboard not working in webxdc apps
--- a/static/help/help.css
+++ b/static/help/help.css
@@ -3,7 +3,12 @@
 html {
   font-family: Roboto, Apple Color Emoji, NotoEmoji, Helvetica Neue, Arial,
     Helvetica, NotoMono, sans-serif !important;
-  background-color: white;
+  color-scheme: dark light;
+}
+@media (prefers-color-scheme: light) {
+  html {
+    background-color: white;
+  }
 }
 
 .back-btn {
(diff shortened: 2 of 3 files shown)


r10s:
i just tried out, for me, the help on android supports dark theme:



is that missing for you? maybe on some android version?

also, the modified files are generated by https://github.com/deltachat/deltachat-pages/blob/master/tools/create-local-help.py - this is also true for desktop - so, if needed, we should adapt the script

static/help/cs/help.html
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html><head><meta charset="UTF-8" /><meta name="viewport" content="initial-scale=1.0" /><link rel="stylesheet" href="../help.css" /></head><body><ul id="top">
+<html><head><meta charset="UTF-8" /><meta name="color-scheme" content="dark light"><meta name="viewport" content="initial-scale=1.0" /><link rel="stylesheet" href="../help.css" /></head><body><ul id="top">
```
r10s, line 2: these files are generated by https://github.com/deltachat/deltachat-pages/blob/master/tools/create-local-help.py , so changes should go there, maybe conditional for desktop only

Simon-Laux (approved):

Simon-Laux:
can you add an changelog entry so we can merge this?

Simon-Laux (approved):

WofWca:
I believe so. See step 3 https://drafts.csswg.org/css-color-adjust/#determine-the-used-color-scheme

> Otherwise, if the user agent supports at least one of the listed [color schemes](https://drafts.csswg.org/css-color-adjust/#color-scheme), the [used color scheme](https://drafts.csswg.org/css-color-adjust/#used-color-scheme) is the first supported color scheme in the list.

> Should be reordered to avoid surprises?

Idk, if the default theme of Delta Chat is light in all other places, then yeah, makes sense. Shall I do it?

PR #3413, opened 2023-09-25
feat: webxdc: make `isSecureContext === true`

WofWca:
...for the apps.

Related test: https://github.com/webxdc/webxdc-test/pull/24

I have tested it with https://github.com/webxdc/webxdc-test/pull/24 and it works.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -35,6 +35,28 @@ protocol.registerSchemesAsPrivileged([
   {
     scheme: 'webxdc',
     privileges: {
+      // This gives apps access to APIs such as
+      // - Web Cryptography
+      // - Web Share
+      // , also see https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts/features_restricted_to_secure_contexts
+      //
+      // To give a brief explanation of what "secure context" is:
+      // Generally all websites served thorugh `https` (and not through `http`)
+      // are in a "secure context".
+      //
+      // For reference:
+      // - https://support.delta.chat/t/allow-access-to-camera-geolocation-other-web-apis/2446?u=wofwca
+      //
+      // Note that APIs requiring explicit user permission (such as camera)
+      // still don't work, see
+      // https://github.com/deltachat/deltachat-desktop/blob/455a4d01501ed82f9d8e0a36064ffbc3981722ee/src/main/deltachat/webxdc.ts#L457-L473
+      //
+      // In terms of `isSecureContext`, webxdc apps are similar to files,
+      // extensions, and FirefoxOS apps, i.e. ["Packaged Applications"]
+      // (https://w3c.github.io/webappsec-secure-contexts/#packaged-applications),
+      // so `secure: true` is applicable.
+      secure: true,
+
       allowServiceWorkers: true,
       standard: true,
       supportFetchAPI: true,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased][unreleased]
 
+### Added
+- mark webxdc app context as secure #3413
+
 ### Fixed
 - fix clipboard not working in webxdc apps
 - fix `target=_blank` links in html emails don't work #3408


src/main/index.ts
```diff
@@ -35,6 +35,25 @@ protocol.registerSchemesAsPrivileged([
...
+      // , also see https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts/features_restricted_to_secure_contexts
+      //
+      // For reference:
+      // - https://support.delta.chat/t/allow-access-to-camera-geolocation-other-web-apis/2446?u=wofwca
+      //
+      // Note that APIs requiring explicit user permission (such as camera)
+      // still don't work, see
+      // https://github.com/deltachat/deltachat-desktop/blob/455a4d01501ed82f9d8e0a36064ffbc3981722ee/src/main/deltachat/webxdc.ts#L457-L473
+      //
+      // In terms of `isSecureContext`, webxdc apps are similar to files,
```
Simon-Laux, line 51: files get less rights than https in reality, at least that's what electron docs says when I understood it correctly

WofWca: Do you have a link?

Simon-Laux: > A standard scheme adheres to what RFC 3986 calls [generic URI syntax](https://tools.ietf.org/html/rfc3986#section-3). For example http and https are standard schemes, while file is not.

nvm seems like that was about the `standard` property.

still in browser the file scheme is heavily restricted at least in my experience.

WofWca: Can you give an example? Is it about `isSecureContext` or something else?

Simon-Laux: if `isSecureContext` is true when you open the html file directly in the browser(no local server), then keep the comment about "file", if not remove "file" from the comment, after that we can merge it.

WofWca: Yes, it's true, at least for mainstream browsers.

Simon-Laux:
changelog missing

WofWca:
Added a changelog entry

src/main/index.ts
```diff
@@ -35,6 +35,25 @@ protocol.registerSchemesAsPrivileged([
   {
     scheme: 'webxdc',
     privileges: {
+      // This gives apps access to APIs such as
+      // - Web Cryptography
+      // - Clipboard
```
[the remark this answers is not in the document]
Simon-Laux: probably not, but I guess we should test against that before merging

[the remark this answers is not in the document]
WofWca: I have tested it. It's already the case, even without `isSecureContext`. Try

```javascript
window.addEventListener('DOMContentLoaded', () => {
  const textarea = document.createElement('textarea');
  document.body.appendChild(textarea);
  textarea.focus();
  setTimeout(() => {
    document.execCommand('paste');
  });
});
```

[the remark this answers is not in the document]
WofWca: So I removed the "Clipboard" point because yeah, it requires permission, which is currently denied (the `write()` API too)  (but `document.execCommand('paste')` still works, but it's not directly related to this MR).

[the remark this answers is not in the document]
WofWca: That is to say, clipboard-related security is not affected by this MR.

r10s:
did not follow all discussions, bjt only looki g at the pr, it is not clear to me, which real-world-issue  this pr will fix. 

esp. as it seems there are some potential side effects,  it would be helpful to elaborate the gist in the pr description a bit more

WofWca:
> it is not clear to me, which real-world-issue this pr will fix

If you want an example, `crypto.randomUUID()` will become available, which, for example, is used in the calendar app, with a fallback to `Math.random()`. Web Crypto API also offers some cryptography stuff, which might be useful for a "secret chat" webxdc app we talked about, like https://github.com/DavidSM100/CryptoText.
I have looked for a bit for other examples besides Web Share and Web Crypto API that require a secure context but don't require a permission but couldn't find one.

The main point of this MR as I see it is to bring us a step closer to giving apps access to camera, geolocation and other stuff that requires permissions. See [the forum post I linked in the comments](https://support.delta.chat/t/allow-access-to-camera-geolocation-other-web-apis/2446?u=wofwca).

Also FYI `isSecureContext` is already `true` for DC iOS and Android.

WofWca:
To give a background for what's "secure context": `isSecureContext` is `true` for all websites served with `https` (well, except for those that are opened in an iframe on a website that is not served securely (i.e. `http`), and some other cases).

r10s:
thanks for more detailed explanations!

PR #3414, opened 2023-09-25
Fix #3363

### LINKED ISSUE #3363: When creating contacts when creating a new group they are not shown until the group was created.
Steps to reproduce:

1. open create chat dialog
2. create new group
3. add members
4. type in email address for a contact you don't have yet and add it
5. click ok and observe that the contact is not shown in the member list
6. create the group and look open the group profile, observe that the contact now was created an is shown.

Solutions:

A. create the contact directly and don't do these pseudo contacts that get removed again at all.
B. create contact but remove it when group creation is canceled also see #3008
C. show the pseudo contacts in the member list of the group (in the create group dialog)

farooqkz:
Hello. This PR fixes #3363 
If the create broadcast component has the same problem, I could do the same for it, too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 - fix html mail getting restrictions of webxdc window
 - fix: when clicking on mailto link in html email show main window even when it was hidden.
 - Display the toast after successful key import.
+- fix the bug where non-contact email addresses are not shown when creating a group #3363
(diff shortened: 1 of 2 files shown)


Simon-Laux (approved):
works for me :+1:

Simon-Laux (approved):

PR #3419, opened 2023-09-28
style: webxdc: better "loading" screen

WofWca:
~~Add a label to say what's going on~~, apply better styles.

~~![image] (https://github.com/deltachat/deltachat-desktop/assets/39462442/383259c0-a0fa-4e77-b3af-c4da2b1fee9d)~~

image

TODO:
- [x] This is based on #3411 - rebase after the merge.
~~- [ ] Add i18n (I don't know how).~~ Not needed for now


r10s:
thanks for picking that up, i agree the loading screen can be nicer.

however, i really do not like the additional text. the additional text will open more question to the user than it answers. most users won't even have questions at that point.

therefore, i suggest to go for what eg. github and others are doing:

- remove top padding
- remove rounded corners
- make the bar a bit thinner (2/3rd)
- remove gray border from the bar
- remove "unfilled grey" color

that way, it will look much nicer and modern

Simon-Laux:
I wanted to subtly blame chromium like "securing your context..." or "isolating chromium"...
anyways Idea is to add an advanced option to disable it. my idea was that text should redirect blame away from the webxdc app

hpk42:
if anything, i think we should have a clear FAQ entry where we can link our blog post from earlier this year, and answer the question of "why does starting a webxdc app sometimes take longer?" in the FAQ for those who are curious.  I concur with bjoern that there is no point in exposing this in the primary UX, or trying to explain things as it only raises more questions.  Moreover, i hope we can at some point introduce a convenient way to let users avoid fill500 under certain conditions (i.e. the app verifiably comes from xstore but that probably requires signed webxdcs).

Simon-Laux:
Ok so no label, improve styles and then we can merge it?

r10s:
> Ok so no label, improve styles and then we can merge it?

yip.

when improving style, however, the loading bar should be more restrained than before, see suggestions above

it is not that bad, however. when doing only one tiny thing: remove the top padding and call it a day

WofWca:
What do you mean by top padding? Do you mean that the bar needs to be at the very top, not in the center?

r10s:
> What do you mean by top padding? Do you mean that the bar needs to be at the very top, not in the center?

yip, this is my suggestion. compare modern on-site loading bars on the web, eg. here on github the blue bar at the very top:

IMG_4309

sth like that looks unobtrusive, modern - and is known by users.

Simon-Laux:
> yip, this is my suggestion. compare modern on-site loading bars on the web, eg. here on github the blue bar at the very top:

I think it's only common when you have content or skeleton elements. But maybe it works for the webxdc case too I don't know.

r10s:
> But maybe it works for the webxdc case too I don't know

it works. we're using it already on desktop and android since 9 months in practise :) in contrast to the initial pr, my suggestions are really only about a _little_ nicer style, as removing the not-so-nice looking upper padding

we're not introducing sth "new" by that

WofWca:
New look:

image

Simon-Laux (approved):

Simon-Laux:
light/dark mode is not set by dc theme, but by system, but I think this is ok for now.

PR #3427, opened 2023-10-09
feat: add developer option to disable IMAP IDLE

link2xt:
Depends on https://github.com/deltachat/deltachat-core-rust/pull/4803

### DIFF
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -32,6 +32,12 @@
   "pref_donate": {
     "message": "Donate"
   },
+  "pref_disable_idle_title": {
+    "message": "Disable IMAP IDLE"
+  },
+  "pref_disable_idle_explain": {
+    "message": "Do not use IMAP IDLE extension even if the server supports it. Enabling this option will delay message retrieval, enable it only for testing."
+  },
   "clear_chat": {
     "message": "Clear Chat"
   },
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 - mark webxdc app context as secure #3413
 - Experimental: Related Chats
+- Developer option to disable IMAP IDLE #4803
 
 ### Changed
 - update deltachat-node and deltachat/jsonrpc-client to `v1.123.0`
--- a/src/renderer/components/dialogs/Settings-Advanced.tsx
+++ b/src/renderer/components/dialogs/Settings-Advanced.tsx
@@ -41,6 +41,7 @@ export function SettingsAdvanced({
       <SettingsExperimentalFeatures
         settingsStore={settingsStore}
         renderDTSettingSwitch={renderDTSettingSwitch}
+        renderDeltaSwitch2={renderDeltaSwitch2}
       />
     </>
   )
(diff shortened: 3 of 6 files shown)


r10s:
is this pr meant to be merged to master and given to "all users" or only used as a branch to created apps to be given to some users?

if it is for "all users", we should consider using a wording that explains more directly that this is a debug option only and does not help on log on issues.

ppl will try out things if they have login issues  - and unless there are situations where this option helps on login, we should e more clear here. maybe it is also better to move it to the general "advanced" settings. for devs to try out fake-idle, this is still good enough - and users do not get confused

_locales/en.xml
```diff
@@ -678,6 +678,8 @@
     <string name="pref_auto_folder_moves_explain">Chat conversations are moved to avoid cluttering the Inbox</string>
     <string name="pref_only_fetch_mvbox_title">Only Fetch from DeltaChat Folder</string>
     <string name="pref_only_fetch_mvbox_explain">Ignore other folders. Requires your server to move chat messages to the DeltaChat folder.</string>
+    <string name="pref_disable_idle_title">Disable IMAP IDLE</string>
+    <string name="pref_disable_idle_explain">Do not use IMAP IDLE extension even if the server supports it. Enabling this option will delay message retrieval, enable it only for testing.</string>
```
Simon-Laux, line 682: please don't add stuff directly here it will get lost, It needs to be added to strings.xml in the android repo, if you need it now add it to the "untranslated json"-file in the same folder.

Simon-Laux:
If you want to hide advanced experimental / developer features you could also put them as function into https://github.com/deltachat/deltachat-desktop/blob/master/src/renderer/experimental.ts so you can reach them from the devTools window's js console under `window.exp.your_function`, there you would not even need translation strings.

Simon-Laux:
> Maybe it is better to keep the new setting where it is, but hide it (together with "Only fetch from DeltaChat folder") under some "expert settings" subheader or experimental option?

we could have a experimental setting that toggles the visibility of the switch in advanced? like "show dangerous Developer options"

_locales/_untranslated_en.json
```diff
@@ -32,6 +32,12 @@
   "pref_donate": {
     "message": "Donate"
   },
+  "pref_disable_idle_title": {
+    "message": "Disable IMAP IDLE"
+  },
+  "pref_disable_idle_explain": {
+    "message": "Do not use IMAP IDLE extension even if the server supports it. Enabling this option will delay message retrieval, enable it only for testing."
```
r10s, line 39: there may be some confusion wrt enable/disable and it/this. here my attempt to clarify wording:

```suggestion
    "message": "Skip IMAP IDLE"
  },
  "pref_disable_idle_explain": {
    "message": "Use polling even if the server supports IMAP IDLE. This will delay message retrieval. Use for testing only."
```

r10s:
> But we can try to move it into Experimental features if needed

"Advanced / Experimental" shows even more clear that this is not an option a "normal" user should try out (which is somehow true for "Advanced" in general, however, in practise things are a bit mixed, also in other apps)

but also here, no blocker

PR #3433, opened 2023-10-13
fix "Verified by" is shown weirdly for contacts that were verified directly

### LINKED ISSUE #3421: "Verified by" is shown weirdly for contacts that were verified directly
for directly verified contacts, the "verified by" shows a message that the other contact has verified itself:

Screenshot 2023-09-28 at 22 32 16

_what really has happend in that case, that tunis3 was verified directly by me_

the message should read "Verified by you" in this case.

the original api required to check this case explicitly by comparing the address strings, however, this was not done in desktop as well as forgotten in android as well.

therefore, the API was much simplified at https://github.com/deltachat/deltachat-core-rust/pull/4754 :

- `dc_contact_get_verifier_addr()` should no longer be used
- `dc_contact_get_verifier_id()` will just return SELF in the case from above

👉 what is needed for desktop to adapt to this change is to show the new string `verified_by_you` in case `dc_contact_get_verifier_id()` returns SELF and to prefer `dc_contact_get_verifier_id()` over `dc_contact_get_verifier_addr()`

Simon-Laux:
closes #3421

### DIFF
--- a/src/renderer/components/dialogs/ViewProfile.tsx
+++ b/src/renderer/components/dialogs/ViewProfile.tsx
@@ -28,6 +28,9 @@ import { BackendRemote, onDCEvent, Type } from '../../backend-com'
 import { selectedAccountId } from '../../ScreenController'
 import moment from 'moment'
 import { InlineVerifiedIcon } from '../VerifiedIcon'
+import { getLogger } from '../../../shared/logger'
+
+const log = getLogger('renderer/dialogs/ViewProfile')
 
 function LastSeen({ timestamp }: { timestamp: number }) {
   const tx = useTranslationFunction()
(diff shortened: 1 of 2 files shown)


src/renderer/components/dialogs/ViewProfile.tsx
```diff
@@ -133,6 +136,10 @@ export function ViewProfileInner({
     null
   )
   const [selfChatAvatar, setSelfChatAvatar] = useState<string | null>(null)
+  const [cachedVerifier, setCachedVerifier] = useState<null | {
```
r10s, line 139: why `cached` here? i would expect the getter/setter to be named just `verifier` and `setVerifier`.

r10s (approved):
apart from the wording, as far as i understand react and typescript, this lgtm code-wise.

did not test.

src/renderer/components/dialogs/ViewProfile.tsx
Simon-Laux: yeah I could remove that "cached" it's not very important.

PR #3434, opened 2023-10-13
fix: remove duplicated search button on "search in chat"

### LINKED ISSUE #3014: two "end search" buttons are unintuitive
when the new (great!) in-chat-search is started, there are two X-buttons to _end something_.

this alone is a bit confusing, however, the _upper_ X-button also appears to do different things to the end user:

- in global-search, search is ended 
- in in-chat-search, search is not ended

in both cases, search field is cleared and focus is removed from search field.

Screenshot 2022-11-16 at 14 44 58 &nbsp; &nbsp; Screenshot 2022-11-16 at 14 39 58

to end the in-chat-search, the second button is required. this is also confusing as usually X-buttons shown at higher positions rule the ones at lower positions. seeing the controls for the first time, one would think, that the second X-button just removes a filter or so.

so, in the current UI, one would expect  the second button to clear the in-chat-search filter and continue with global search.

however, i do not suggest to make the button a remove-filter-button, instead, i propose to just **remove the second X-Button and end all search when pressing the upper one:**

- re-entering search is easy - and already know to the user (clicking in the search field or using the menu)
- an explicit shortcut for "clear all" is not that import on desktop - and in the current implementation (focus lost) the gist seems to be more "end search" anyway
- no need to make the second button nicer :)

of course, the upper X-button needs to be shown also when there is no text input in case of in-chat-search then. to be clear: this is only about removal of the second X-button, not about the line showing the chat.

just my 2ct for the in-chat-search :)

Simon-Laux:
fixes #3014

and more compact design:
Bildschirmfoto 2023-10-13 um 21 30 45

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - add a dark theme for the "Help" and the webxdc loading screen
 - update `@deltachat/message_parser_wasm` to `0.8.0`, which adds linkification to links on more generic URI schemes.
 - Update translations (13.10.2023)
+- better search in chat design which shows more results (remove redundant chat info and combine both headers)
 
 ### Fixed
 - fix clipboard not working in webxdc apps
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - add description for enableChatAuditLog setting
 - fix: import key from file instead of folder, fixes #1863
 - fix webxdc title not updated in document title changes #3412
+- fix: remove duplicated search button on "search in chat" #3014
 
 <a id="1_40_4"></a>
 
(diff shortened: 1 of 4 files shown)


r10s:
> is it understandable that you need to press the small x to exit search?

i think, this is already a huge improvement to the state before, where you had to tap the _second_ "X" to exit the search-in-chat mode:

Screenshot 2023-10-13 at 20 40 10 <i>requiring to tap the second "X"  was more irritating :)</i>

---

as real "tag chips" are out of scope, and maybe also not needed, as a further improvements, what comes to my mind:

- the repetition of the chat name and the chat avatar is superfluous - and also looks like a search result; better remove it

- the "Search in Chat" subtitle can stay, however, it could also be combined with the "123 Messages" (or use it as placeholder?)

- the immediately appearing "error" of  "0 Messages found" before even typing something could maybe just be removed (once there are result or if really nothing is found, the title should appear, of course)

r10s (approved):
i think, this is an improvement - it is definitely easier for the user to exist search-in-chat

r10s (approved):

PR #3442, opened 2023-10-20
add option to save to file system to webxdc "send to chat"-dialog

Simon-Laux:
also removed unnecessary Card component and made a new css file for this dialog (don't just reuse all mailto styles), I think in the future it would make sense to give the "select chat from list with search" ui element it's own react component, then it makes more sense to just have one class for it because then we can be sure there really is only one variant of the element.

hide whitespaces for a clearer review experience.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - mark webxdc app context as secure #3413
 - Experimental: Related Chats
 - Developer option to disable IMAP IDLE #4803
+- add option to save to file system to webxdc "send to chat"-dialog
 
 ### Changed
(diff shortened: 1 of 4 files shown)


farooqkz:
I don't get what does this PR do from the title and the first comment/message.

farooqkz:
Okay the problem's that I don't understand what `WebxdcSendToChat` for :D

Simon-Laux:
> Okay the problem's that I don't understand what WebxdcSendToChat for :D

the export button in the time tracking app or xstore@testrun.org bot.

adbenitez (approved):

PR #3443, opened 2023-10-23
open help, webxdc and html email windows with always on top flag, if main window has that flag.

### LINKED ISSUE #3365: "float on top" may make show-full-message, webxdc and help virtually unusable
if "Float On Top" flag is set for the main window, tapping "Show full message..." or "Start..." for an webxdc may look to the user as if nothing works or happens at all - as the opened full-message-view/webxdc is covered completely by the main window - and eg. on mac, you do not even see the webxdc window by some icon nor is the webxdc reachable by alt-tab. the help window has the issue, maybe others.

fixes may be:
- make the full-message-view/webxdc a "parent" of the main window (not sure what electron offers here, OS supports that, cmp dialogs)
- alternatively, full-message-view/webxdc windows should get the "float on top" as well. i do not think it is needed to have the menu-option for webxdc windows nor it is needed to apply it later on, it is only important that the user sees the window after hitting "start ..."
- consider, if "float on top" is needed at all; i think, this was considered at some point, so this is probably only an option if everything else fails or is too hard
EDIT: even if some system itself allows "float on top" (mac or windows do _not_ allow that by default iirc), this would be less if an issue as the mode is less likely enabled accidentally - and  it is also clear thar this mode is not maintained by us
- ...

EDIT: the issue is true not only for webxdc, i adapted title and text

Simon-Laux:
fixes #3365

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 - fix webxdc title not updated in document title changes #3412
 - fix: remove duplicated search button on "search in chat" #3014
 - fix "Verified by" is shown weirdly for contacts that were verified directly #3421
+- open help, webxdc and html email windows with always on top flag, if main window has that flag. 
 
 <a id="1_40_4"></a>
 
--- a/src/main/deltachat/webxdc.ts
+++ b/src/main/deltachat/webxdc.ts
@@ -238,6 +238,7 @@ export default class DCWebxdc extends SplitOut {
           icon: app_icon || undefined,
           width: 375,
           height: 667,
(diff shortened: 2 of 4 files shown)


r10s (approved):

r10s:
would still be interesting, if float-on-top is set by system, if this is also returned by isAlwaysOnTop() - otherwise, that would produce a similar issue as #3365.

but i think, that option is far less available on the systems than assumed by some dev, some linux have it - but iirc not windows/mac with defaults - so maybe even here good enough

overall, this pr makes things much better for the options we offer ourselves by our menus, this is the most important thing

Issue #2869, opened 2022-08-29
images are not recoded


r10s:
meanwhile, we have an attach-menu where you can explicitly send out "media", such as images.

however, currently "media" are _not_ recoded, so sending a jpeg can result easily in sending out 8 mb or so. 

fix is probably simple, the media should be sent out as such, then core will recode ("files" are left alone) 

(to save everyone's bandwidth, in general, Delta Chat should make things small _by default_. if one _really_ wants to send raw data, there is still the "file" option. this is what android/ios are doing here, to keep things simple and make things work, i'd just go for that :)

(tested with 1.32.0 on mac)

farooqkz:
@Simon-Laux What is the best way? Just downscaling and compressing would be enough maybe?

r10s:
downscaling and compression is already implemented in core; this should not be done in UI.

instead, if the user selects sth via "Media", send them with the flag DC_MSG_IMAGE (instead of DC_MSG_FILE)

if needed and/or possible, core will change the really outgoing file and type.

Issue #2838, opened 2022-07-24
name of "Media" attach button is misleading


r10s:
"Media" are not just images and videos but also audio.

so either include audio or rename the button to eg. "Images" or "Images and Videos".

see https://support.delta.chat/t/problem-with-media-option-in-choosing-attachments/2165/2

PR #3447, opened 2023-10-27
send compressed images

Simon-Laux:
rename attach media button to "image", allow only images and compress them on sending.
fixes https://github.com/deltachat/deltachat-desktop/issues/2838
fixes https://github.com/deltachat/deltachat-desktop/issues/2869

depends on https://github.com/deltachat/deltachat-core-rust/pull/4888

also use the new send message core function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
 - Add image compression
 
 ### Changed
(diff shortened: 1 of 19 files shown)


r10s (approved):
as far as i can tell, this lgtm, thanks a lot!

PR #3448, opened 2023-10-27
improve video message - wide enough to show controls

Simon-Laux:


r10s (approved):
what a huge improvement with only 8 loc! thanks a lot!

PR #3456, opened 2023-11-01
Add sending voice messages to desktop

farooqkz:


src/renderer/components/composer/Composer.tsx
```diff
@@ -81,7 +81,28 @@ const Composer = forwardRef<
...
+    }
+  
+    recorder.ondataavailable = (ev: BlobEvent) => {
+      data.push(ev.data);
+    }
+
+    recorder.onstop = () => {
+      clearInterval(updateRecordedSecondsInterval)
+    }
+  })
```
Simon-Laux, line 103: you need to use useEffect, useMemo or useCallback, otherwise this will be run every time the component rerenders

src/renderer/components/composer/Composer.tsx
```diff
@@ -534,3 +576,21 @@ export function useDraft(
...
+
+
+function RecordingDuration({ seconds }: {
+  seconds: number
+}) {
+  let minutes = 0
+  while (seconds >= 60) {
+    seconds /= 60
+    minutes++
+  }
```
Simon-Laux, line 588: depending of what you want to do https://momentjs.com/docs/#/durations/ might help, we depend on momentjs already anyway

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 - fix clipboard not working in webxdc apps
 - fix `target=_blank` links in html emails don't work #3408
 
+
+### added
+
+- voice messages
```
Simon-Laux, line 12: there is already a section for this above, you need to rebase your pr to newest master though, btw you should try to work from a recent master anyway to get less possibility of merge conflicts

src/renderer/components/composer/ComposerMessageInput.tsx
```diff
@@ -28,14 +30,44 @@ export default class ComposerMessageInput extends React.Component<
...
   constructor(props: ComposerMessageInputProps) {
     super(props)
     this.state = {
       text: '',
       chatId: props.chatId,
       loadingDraft: true,
+      recordedDuration: null,
     }
-
+    navigator.mediaDevices.getUserMedia({ audio: true, video: false }).then(stream => {
```
Simon-Laux, line 45: I have a feeling we should do the setup only if the user clicked on record. also we might need a runtime call to check and request permission?

Also later if could make some sense to ask user for which microphone they wanna use If they have multiple ones connected, or we somehow always take the default?

farooqkz: For now, I'd say merge the PR with the default microphone. For later, we could add something into Settings.

_locales/_untranslated_en.json
Simon-Laux, line 1: don't reintroduce the strings we already upstreamed

src/renderer/components/composer/ComposerMessageInput.tsx
farooqkz: I'm not sure if we should only request permission when recording starts. What is your reason? TBH, I don't see any difference.

_locales/_untranslated_en.json
farooqkz: I did a rebase but it seems I've done it wrong. will fix in a commit.

src/renderer/components/composer/ComposerMessageInput.tsx
Simon-Laux: > What is your reason?

apple design guidelines and other places, it's best practice to ask for permission when its needed.
also sometimes access to microphone is shown similar to how most webcams have a light when they are recording.
If that indication is on when they don't record voice messages we will get complains and new conspiracy theories.

I mean get the stream only when you click on the button and use it

farooqkz: Yeah it makes perfect sense. Will change it and resolve this convo

Simon-Laux:
ci complains about code formatting.

src/renderer/components/composer/ComposerMessageInput.tsx
```diff
@@ -92,6 +127,24 @@ export default class ComposerMessageInput extends React.Component<
   clearText() {
     this.setState({ text: '' })
   }
+  
+  startRecording() : any {
```
Simon-Laux, line 131: does not return anything
```suggestion
  startRecording() {
```

src/renderer/components/composer/ComposerMessageInput.tsx
```diff
@@ -28,14 +30,44 @@ export default class ComposerMessageInput extends React.Component<
...
+          log.error('duration of voice is zero while the mediarecorder has stopped. this must not happen')
+        this.setState({ recordedDuration: null })
+        // send the voice message
+        console.log(this.voiceData)
+        this.voiceData = []
+        this.updateRecordedDurationInterval && window.clearInterval(this.updateRecordedDurationInterval)
+      }
+    })
+    .catch((reason: any) => {
+      this.gumError = reason
```
Simon-Laux, line 69: please also log this error with `log.error`, maybe we should even somehow show it to the user later, maybe in an alert dialog?

Simon-Laux (changes requested):
- [x] once you type in text or have another file in the draft the voice message button should be disabled/hidden
- [x] voice message should be stored in draft and should be able to be listened to once recording is done, with option to delete it.
- [ ] advanced: think about what happens if user tries to switch the chat while recording. (ideally would stop recording and save it as draft, but might be too complex with current composer implementation?)

farooqkz:
The work is done. It works. There are two remaining tasks, however:

1. It seems Electron by default does not support most codecs and we might need a custom build: https://github.com/electron/electron/issues/633

image

Note that adding any codec randomly could cause to more bugs, possibly security bugs.

2. I haven't used Alert dialogues anywhere to show errors to the users. I was thinking about waiting till https://github.com/deltachat/deltachat-desktop/pull/3512 gets merged.

Ping @Simon-Laux @adzialocha

Simon-Laux:
Also: 
3. voice message is not saved as draft yet.

src/renderer/components/composer/Composer.tsx
```diff
@@ -93,10 +94,49 @@ const Composer = forwardRef<
...
+              : null,
+        })
+        await BackendRemote.rpc.removeDraft(selectedAccountId(), chatId)
+        clearDraft()
+        window.__reloadDraft && window.__reloadDraft()
+      } catch (e) {
+        log.error('Cannot send voice message:', e)
+      }
+      return
+    }
```
Simon-Laux, line 139: Is there any reason to duplicate this logic?

src/renderer/components/composer/Composer.tsx
```diff
@@ -366,12 +367,12 @@ const Composer = forwardRef<
             onMouseUp={() => {
               messageInputRef.current?.stopRecording()
             }}
-            aria-label={tx('voice_send')}
-            disabled={Boolean(draftState.file || messageInputRef.current?.hasText())}
+            aria-label={voiceMessageDisabled ? tx('voice_send_cannot') : tx('voice_send')}
```
Simon-Laux, line 368: dunno if this is needed, can't you just add a `aria-disabled="true"` attribute?

Simon-Laux:
It has a video mimetype (video/webm), we would need to change message rendering logic to display it as a voice message regardless, because currently ended voice messages look like this:


PR has this bug/regression for seeing normal attachments:
Bildschirmfoto 2023-12-02 um 14 49 41
(you can not add text anymore, voice icon is still shown)

UI while recording needs to be improved we can not merge it like that:
Bildschirmfoto 2023-12-02 um 14 51 18
- while recording mic icon should stay red
- waveform display would be nice, so there is some feedback on that is recorded / on the levels to give user feedback, if it doesn't move they know sth is wrong with their microphone.

After recording Voice message it looks like this:
Bildschirmfoto 2023-12-02 um 14 53 45
- hard to understand to user what the state is
- no way to pre-listen the message before sending


PS: please don't mark your pr as "ready to merge" when it is clearly not ready.

PR #3457, opened 2023-11-02
handle mailto: and openpgp4fpr: links in webxdc

### LINKED ISSUE #3355: mailto link inside webxdc causes blank page
image

Simon-Laux:
closes #3355

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -292,7 +292,7 @@ app.on('web-contents-created', (_ev, contents) => {
         ev.preventDefault()
       }
     })
-    
+
     contents.on('will-frame-navigate', ev => {
       if (ev.url.startsWith('webxdc://')) {
         // allow internal webxdc nav
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -271,30 +271,57 @@ app.on('web-contents-created', (_ev, contents) => {
     contents.session.storagePath &&
     contents.session.storagePath.indexOf('webxdc_') !== -1
   if (is_webxdc) {
-    contents.on('will-navigate', (e, navigationUrl) => {
+    const webxdc_open_url = (url: string) => {
+      if (url.startsWith('mailto:') || url.startsWith('openpgp4fpr:')) {
+        // handle mailto in dc
+        open_url(url)
+        mainWindow.window?.show()
+      }
+    }
+
+    contents.on('will-navigate', (ev, navigationUrl) => {
       if (navigationUrl.startsWith('webxdc://')) {
         // allow internal webxdc nav
         return
       } else if (navigationUrl.startsWith('mailto:')) {
         // handle mailto in dc
-        e.preventDefault()
-        open_url(navigationUrl)
-        mainWindow.window?.show()
+        ev.preventDefault()
+        webxdc_open_url(navigationUrl)
       } else {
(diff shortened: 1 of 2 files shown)


src/main/index.ts
```diff
@@ -271,30 +271,57 @@ app.on('web-contents-created', (_ev, contents) => {
...
+        ev.preventDefault()
+        webxdc_open_url(ev.url)
       } else {
         // prevent navigation to unknown scheme
-        e.preventDefault()
+        ev.preventDefault()
       }
     })
+
+    contents.setWindowOpenHandler(_details => {
```
adzialocha, line 310: Again, more a newbie question: 

I can see how this is the only path to handle `openpgp4fpr` links from here.

I understand why we need to prevent new windows of being created, but how come we're handling URLs in here and not only in the `will-navigate` or `will-frame-navigate` event handlers? Is there a case where links will only be handled here?

src/main/index.ts
```diff
@@ -271,30 +271,57 @@ app.on('web-contents-created', (_ev, contents) => {
     contents.session.storagePath &&
     contents.session.storagePath.indexOf('webxdc_') !== -1
   if (is_webxdc) {
-    contents.on('will-navigate', (e, navigationUrl) => {
+    const webxdc_open_url = (url: string) => {
+      if (url.startsWith('mailto:') || url.startsWith('openpgp4fpr:')) {
+        // handle mailto in dc
```
adzialocha, line 276: With the danger of being new here and asking annoying questions: What does this comment exactly imply? I read it as "Handle mailto: in DeltaChat", but exactly which parts of it? Where would we handle it instead if not in DeltaChat?

src/main/index.ts
```diff
@@ -271,30 +271,57 @@ app.on('web-contents-created', (_ev, contents) => {
...
       if (navigationUrl.startsWith('webxdc://')) {
         // allow internal webxdc nav
         return
       } else if (navigationUrl.startsWith('mailto:')) {
         // handle mailto in dc
-        e.preventDefault()
-        open_url(navigationUrl)
-        mainWindow.window?.show()
+        ev.preventDefault()
+        webxdc_open_url(navigationUrl)
```
adzialocha, line 289: Mini nitpick, do we have some style guides around letter casing? In JS world I'd expect method names to be camelCase but also I'm open to follow whatever style is preferred here

adzialocha (approved):
Looking good to me! Except of prettier not being happy I can't find any issues, only a few newbie questions in my comments.

I couldn't understand how to make `openpgp4fpr` links work on this branch, tried to send some to myself but they are not clickable, guess that's not implemented yet?

src/main/index.ts
Simon-Laux: > Where would we handle it instead if not in DeltaChat?

in the system handler for mailto, which can be deltachat, but is likely a traditional email MUA like thunderbird

src/main/index.ts
adzialocha: Okay cool, thanks for clarifying!

src/main/index.ts
Simon-Laux: I changed the casing for `webxdc_open_url`, we can change the rest of the projects var name casings in a different pr.

Issue #3459, opened 2023-11-02
electron exited with signal SIGTRAP


adzialocha:
- Operating System (Linux/Mac/Windows/iOS/Android): `Linux 6.5.9-arch2-1 x86_64 GNU/Linux`
- Delta Chat Version: `Deltachat Version 1.40.4 v1.33.0-433-g2ee34084 1698919631540`
- Expected behavior: Electron starting when running `npm run dev`
- Actual behavior: Error `electron exited with signal SIGTRAP`

Related to the discussion in https://github.com/deltachat/deltachat-desktop/pull/3381 I can't start electron when running `npm run dev` (after compiling the rust core manually and using NodeJS `18.14.0`).

The error I'm seeing is:

```
> deltachat-desktop@1.40.4 start
> electron . --devmode --disable-http-cache --translation-watch

Logfile: /home/adz/.config/DeltaChat/logs/2023-11-02-11-07-17.log
0s [i]main/index: Deltachat Version 1.40.4 v1.33.0-433-g2ee34084 1698919631540
/home/adz/dev/deltachat/deltachat-desktop/node_modules/electron/dist/electron exited with signal SIGTRAP
```

I've poked a bit in the `src/main/index.ts` codebase and realized it might be due to running critical processes already before electron signals its `onReady` event (got the hint from this issue here https://github.com/electron/electron/issues/38592).

Will write a bit more as soon as I've figured out which code parts cause this race condition.

adzialocha:
Alright, removing `import './resume_from_sleep'` https://github.com/deltachat/deltachat-desktop/blob/master/src/main/index.ts#L317 fixes the issue (and I stuble over others, but I'm not sure yet if they're related).

Mic92:
Looks like you figured out? Anyway here is the backtrace for completeness:

```
Starting program: /nix/store/7a78jpnjx4m1s8wzdc45ayi1rbf3pwyg-electron-unwrapped-26.4.0/libexec/electron/electron /nix/store/mpzkxg88hpdsxg69phlxah8vn1hpaswr-deltachat-desktop-1.40.4/lib/node_modules/deltachat-desktop
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/nix/store/gqghjch4p1s69sv4mcjksb2kb65rwqjy-glibc-2.38-23/lib/libthread_db.so.1".
[New Thread 0x7fffe60d96c0 (LWP 258667)]
[Detaching after fork from child process 258668]
[...]

PR #3381, opened 2023-09-06
upgrade electron to 26

Simon-Laux:
- upgrade electron to 24
- upgrade to electron 25 - make it compile
- migrate now deprecated apis
- upgrade to electron 26

Let's pospone thinking about this to october, because it woul drop support for windows 7 & 8, electron 22 is the last version that 
supports those Operating systems and it will be supported until october.

PR #3460, opened 2023-11-02
Initialise power monitor after electron signals readyness to fix SIGTRAP issue

adzialocha:
This PR fixes a race condition causing electron to directly crash returning `SIGTRAP`. Apparently we can not use electrons `powerMonitor` before the `onReady` event got fired.

Related PR: https://github.com/deltachat/deltachat-desktop/pull/3381

Closes: #3459

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@
 - improve video message - wide enough to show controls
 - gallery: fix scroll to top when switching tabs
 - fix: context menu items could take up multiple lines
+- fix: initialise power monitor after electron signals readyness to avoid electron failing with SIGTRAP
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,7 +48,7 @@
 - improve video message - wide enough to show controls
 - gallery: fix scroll to top when switching tabs
 - fix: context menu items could take up multiple lines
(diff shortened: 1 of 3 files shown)


adzialocha:
@ralphtheninja and @Simon-Laux could you check if this fixes the problem from your end as well?

ralphtheninja:
@adzialocha Will check right away! How did you find this out?

ralphtheninja:
This fixes the issue with SIGTRAP, but instead I get the following exception

image

@adzialocha Is this what you're targeting in #3461 ?

ralphtheninja (approved):

adzialocha:
> @Simon-Laux I can confirm that this PR together with #3461 unbreaks master! (Made a temporary local branch which combines them both to test it out).

Great! Thank you for checking.

> @adzialocha Is this what you're targeting in https://github.com/deltachat/deltachat-desktop/pull/3461 ?

Yes, exactly!

adzialocha:
> @adzialocha Will check right away! How did you find this out?

I've noticed that the first log line got printed before electron failed, then I've isolated what line caused that failure.

Also that comment in someone elses issue helped: https://github.com/electron/electron/issues/38592#issuecomment-1595858203

Mic92:
Ok. I can confirm now that this pull request definitely fixed the crash on my side.

PR #3463, opened 2023-11-02
Show prettier linter warnings through eslint

adzialocha:
This PR installs the `eslint-plugin-prettier` package and enables prettier-related linting warnings through eslint.

I prefer to see prettier style violations directly in my IDE (neovim in that case) and use the automatic fixing feature as a part of it, but also happy to find another approach if there are any recommendations.

Before:

screenshot-2023-11-02T15:10:43

After:

screenshot-2023-11-02T15:11:24

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,12 +1,14 @@
 module.exports = {
   root: true,
   parser: '@typescript-eslint/parser',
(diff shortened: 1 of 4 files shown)


Simon-Laux (approved):
LGTM

PR #3464, opened 2023-11-02
update deltachat-node and deltachat/jsonrpc-client to `v1.128.0`

Simon-Laux:
dc core is now build with and for node 18


ralphtheninja (approved):
Verified locally. Works fine!

Get some warning about missing translation in console:

image

Simon-Laux:
> Get some warning about missing translation in console:

maybe time to remove the ancient device message

PR #3466, opened 2023-11-02
remove outdated device messages

Simon-Laux:
also because translation for key `update_1_36_audit` was removed.

I keept the 1.38 device message that we stil have the template to modify for the 1.41 and 1.42 device messages.


ralphtheninja (approved):
Removing code! Love it :fire:

PR #3467, opened 2023-11-02
Fix centering of username in settings view

adzialocha:
Before:

2023-11-02-164456_591x353_scrot

After:

2023-11-02-164435_594x429_scrot


farooqkz:
> Before:
> 
> 2023-11-02-164456_591x353_scrot
> 
> After:
> 
> 2023-11-02-164435_594x429_scrot

hmm I don't see any difference in these two images at a first glance :/

adzialocha:
> hmm I don't see any difference in these two images at a first glance :/

The trick is that the whole component is now centered against the profile image, very small difference indeed 😹 

2023-11-03-163112_795x613_scrot

Simon-Laux (approved):

PR #3469, opened 2023-11-03
Show confirm dialogue when creating new chat from clicking mail address

adzialocha:


src/renderer/components/dialogs/MailtoDialog.tsx
```diff
@@ -99,37 +97,3 @@ export default function MailtoDialog(props: {
     </DeltaDialogBase>
   )
 }
-
-export async function doMailtoAction(chatId: number, messageText: string) {
```
adzialocha, line 103: I renamed this method to `createDraftMessage` as I thought it is a more explicit name for what the method actually does. As it is only doing things in a chat context I moved it into the "chat methods" file. Happy to get some feedback on the naming and placing!

src/renderer/components/helpers/OpenQrUrl.tsx
```diff
@@ -80,48 +78,7 @@ export default async function processOpenQrUrl(
...
-      }
-      callback && callback()
-    } catch (error) {
-      log.error('mailto decoding error', error)
-      window.__openDialog('AlertDialog', {
-        message: tx('mailto_link_could_not_be_decoded', url),
-        cb: callback,
-      })
-    }
+    processMailtoUrl(url, callback)
```
adzialocha, line 81: I've moved this logic into its own method, both for reusability (it is called in two different places) and also to reduce the complexity of the `processOpenQrUrl` method

src/renderer/runtime.ts
```diff
@@ -499,7 +499,7 @@ class Electron implements Runtime {
   }
   openLink(link: string): void {
     if (link.startsWith('mailto:')) {
-      processOpenQrUrl(link)
+      processMailtoUrl(link)
```
adzialocha, line 502: Since we're really only dealing with `mailto:` links here I thought its clearer to also address it like that

src/renderer/components/helpers/ChatMethods.ts
```diff
@@ -206,6 +206,49 @@ export async function joinCall(
...
+export async function createChatByEmail(email: string): Promise<number | null> {
+  const accountId = selectedAccountId()
+
+  let contactId = await BackendRemote.rpc.lookupContactIdByAddr(
+    accountId,
+    email
+  )
+
+  const chatId = contactId
+    ? await BackendRemote.rpc.getChatIdByContactId(accountId, contactId)
```
adzialocha, line 225: The `getChatIdByContactId` method was newly added to the JSON RPC exports and is not released yet, this is potentially a blocker

adzialocha (commented):
Just adding some comments to explain my decisions to move things around a little, happy to get feedback on naming and placing!

src/renderer/components/helpers/ChatMethods.ts
```diff
@@ -206,6 +206,49 @@ export async function joinCall(
...
+    : null
+
+  if (chatId) {
+    return chatId
+  }
+
+  // Ask user if they want to proceed with creating a new contact and / or chat
+  const continueProcess = await new Promise((resolve, _reject) => {
+    window.__openDialog('ConfirmationDialog', {
```
Simon-Laux, line 234: optional: tip: if you import the dialog and use it instead of the id here you also get type validation on the options.

src/renderer/components/helpers/ChatMethods.ts
```diff
@@ -206,6 +206,49 @@ export async function joinCall(
...
+  const chatId = contactId
+    ? await BackendRemote.rpc.getChatIdByContactId(accountId, contactId)
+    : null
+
+  if (chatId) {
+    return chatId
+  }
+
+  // Ask user if they want to proceed with creating a new contact and / or chat
+  const continueProcess = await new Promise((resolve, _reject) => {
```
Simon-Laux, line 233: I like the confirmation dialog as promise, maybe we can make a generic helper method for that. like this:
```ts
const confirmed =  await confirmationDialog({message: "are you sure", confirmLabel: "Yes"});
if (confirmed){

}
```

package.json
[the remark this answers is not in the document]
adzialocha: Agreed! I keep it as a reminder that this is not ready yet 😅

Simon-Laux (changes requested):
The mailto links of https://github.com/deltachat/interface/blob/master/user-testing/mailto-links.md don't work anymore.

To test either install deltachat from this batch and set as default mailer or easier open the raw md file and  copy the full links, then scan qrcode scan from clipboard and also test clicking them from a message

package.json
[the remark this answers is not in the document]
Simon-Laux: you can also add a wait-for-core tag then when we still wait for the new core release.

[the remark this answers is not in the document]
Simon-Laux: but anyways better to not commit it because it might clutter the git history

r10s:
hi, @adzialocha! in general, thanks a lot for the great pr description and for the nice video, giving ppl a bit outside desktop code a good overview what a pr is about. this is very welcome!

Simon-Laux (changes requested):
Just tested with the core master, works now as expected for me. Sorry for the confusion.
We can merge after you addressed my review comments and we have the new core release with the new jsonrpc api.

BTW: maybe we should have a different error when the chat stuff failed, "mailto parse failed" is rather confusing.

adzialocha:
Thank you for your review comments and confirming that all is good with the links now @Simon-Laux!

I like the idea to have an own helper method for the confirmation dialogue, would tackle that in a separate, follow up PR if that makes sense?

Also, I assume we will not merge this until the new core is released, right?

Simon-Laux (approved):

PR #3471, opened 2023-11-03
Remove `url-parse` dependency and replace it with `URL`

### LINKED ISSUE #3470: Use native `URL` instead of `url-parse` package
From the `url-parse` README:

> "url-parse was created in 2014 when the WHATWG URL API was not available in Node.js and the URL interface was supported only in some browsers. Today this is no longer true. The URL interface is available in all supported Node.js release lines and basically all browsers. Consider using it for better security and accuracy." (https://github.com/unshiftio/url-parse)

We can probably switch over to the native URL interface and remove that dependency.

farooqkz:
It should work but since I haven't tested it yet, marking it as draft.

This closes #3470


adzialocha (approved):
Great! Thank you for changing this so quickly @farooqkz! I've just tested it against my branch and it works as expected.

src/main/deltachat/webxdc.ts
```diff
@@ -111,7 +110,7 @@ export default class DCWebxdc extends SplitOut {
 
         if (!accounts_sessions.includes(accountId)) {
           accounts_sessions.push(accountId)
-          ses.protocol.handle('webxdc', async request => {
+          ses.protocol.handle('webxdc', async (request: any) => {
```
Simon-Laux, line 113: sure that marking it as any is correct here? shouldn't it be typed automatically by electron types?

Simon-Laux (approved):

PR #3474, opened 2023-11-04
reword some strings

r10s:
this pr rewords some strings and slightly shifts menu items.

see commit messages for details.

### DIFF
--- a/src/renderer/components/helpers/ChatMethods.ts
+++ b/src/renderer/components/helpers/ChatMethods.ts
@@ -217,7 +217,7 @@ export async function createChatByContactIdAndSelectIt(
   )
 
   if (!chatId) {
-    throw new Error(window.static_translate('create_chat_error_desktop'))
+    throw new Error('could not create chat')
   }
 
   const chat = await BackendRemote.rpc.getFullChatById(accountId, chatId)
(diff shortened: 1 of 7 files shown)


src/renderer/components/helpers/ChatMethods.ts
```diff
@@ -217,7 +217,7 @@ export async function createChatByContactIdAndSelectIt(
   )
 
   if (!chatId) {
-    throw new Error(window.static_translate('create_chat_error_desktop'))
+    throw new Error('could not create chat')
```
Simon-Laux, line 220: This error is in one of 3 code paths shown to the user, But anyways should not happen and is not useful, so no need to translate it (`createChatByContactId` would normally throw a more useful error before this if is reached I think).

src/renderer/components/helpers/ChatMethods.ts
```diff
@@ -217,7 +217,7 @@ export async function createChatByContactIdAndSelectIt(
...
   }
```
Simon-Laux, line 221: Anyways thinking about it this if statement is a remnant of using the cffi, it is useless as createChatByContactId does not return `0`/`null`, it's return type is `Result<u32>`, so we can remove the whole if statement here.

Simon-Laux (approved):
LGTM

src/renderer/components/helpers/ChatMethods.ts
r10s: even  better!

Simon-Laux: made a pr for it https://github.com/deltachat/deltachat-desktop/pull/3475

PR #3475, opened 2023-11-05
remove useless if statement

Simon-Laux:
which was a remnant from the cffi.
`BackendRemote.rpc.createChatByContactId` will return a more useful error if sth failed.


r10s:
just for interest: where is createChatByContactId() declared? the [call](https://github.com/deltachat/deltachat-desktop/blob/c91124ca93e1383f0289e81c16f733cdd3a8f287/src/renderer/components/helpers/ChatMethods.ts#L214) looks to me as if chat_id is returned as a number, not as a result. but i am also not into ts at all, so probably everything's correct :)

r10s (approved):

PR #3483, opened 2023-11-06
fix `null` account name when leaving the field empty bug

Simon-Laux:
Basically leaving the profile name empty resulted in the name `"null"` being chosen.

Also there is another bug where the name was not updated because this happens in while passing to jsonrpc and the settings cache just stores what you set, maybe it should reload the affected config key instead, but that's another issue/task.


hpk42 (approved):
thanks, lgtm

adzialocha (approved):

PR #3490, opened 2023-11-06
remove 40KiB download on demand option

### LINKED ISSUE #3489: remove smallest auto-download option
smallest option is 160k then, see https://github.com/deltachat/deltachat-android/issues/2823 and https://github.com/deltachat/deltachat-core-rust/pull/4946  for reasoning

Simon-Laux:
closes #3489


Simon-Laux:
> Those who already have this selected will see the dialog with no option selected?

I think so. should not be an issue.

Issue #3468, opened 2023-11-02
unify group creation, verify tweaks


r10s:
for group creation:

- [x] remove "Create Verified Group" from "New Chat" screen https://github.com/deltachat/deltachat-desktop/pull/3492
- [x] remove "QR Invite Code" from "Create Group" screen https://github.com/deltachat/deltachat-desktop/pull/3492
- [x] when _only verified contacts_ are selected in "New Group" screen, create a verified group - otherwise a normal group - no question or so https://github.com/deltachat/deltachat-desktop/pull/3492
- [x] set new stock string `DC_STR_NEW_GROUP_SEND_FIRST_MESSAGE` to `chat_new_group_hint` https://github.com/deltachat/deltachat-desktop/pull/3492

for modifying member list of existing groups:

- [x] remove verified/unverified filtering of contact lists when adding members https://github.com/deltachat/deltachat-desktop/pull/3492
- [x] when tapping a unverified contact in a verified group, show an alert with the text `verified_contact_required_explain` and the answer `learn_more`/`qrscan_title`/`ok` with scan-qr similar to what we're doing at #3328 https://github.com/deltachat/deltachat-desktop/pull/3492

for displaying verifier information in profiles:

- [x]  for the "introduced by" line, _only_ use `verifier_id` (no `ContactObject.is_verified`), see https://github.com/deltachat/deltachat-core-rust/pull/4922 for reasoning
- [x] show a green checkmark icon right of the chat name if  `chat_is_protected()` is true. this is for consistency with chat titlebar and for consistency with android/ios EDIT: cleanup
- [x] EDIT: additionally, show the same green checkmark icon in the profile in case no chat exist with that contact and `contact_is_verified()` is true (these profiles-without-chats seems to be reachable from global search for names only)
- [x] re-pull recent translations, there were some rewordings

counterpart of <https://github.com/deltachat/deltachat-ios/pull/1953> and <https://github.com/deltachat/deltachat-android/issues/2773>/<https://github.com/deltachat/deltachat-android/issues/2774>/<https://github.com/deltachat/deltachat-android/issues/2775>/<https://github.com/deltachat/deltachat-android/pull/2800>

r10s:
for the three "learn more" links, currently, for a beta, see https://github.com/deltachat/deltachat-android/pull/2828/commits/0108178f29dc7d46e6c57f6a4220325fc08c7e93

for the final version, the "learn more" links needs to be adapted, maybe use the offline help

PR #3328, opened 2023-07-22
verified 1on1 chats

Simon-Laux:
closes #3327

- [x] enable verified_one_on_one_chats
- [x] Show a "verified" or "verification broken" icon on protection changed update messages
  - https://github.com/deltachat/interface/blob/master/icons/ic_verified_broken.svg
- [x] use `is_protection_broken` in addition to can send to show dialog buttons like for contact request
    - [x] accept is buggy (input field is not usable after accepting)
    - [X] implement more info dialog
- [x] When tapping on the DC_INFO_PROTECTION_{EN|DIS}ABLED message, show more information in alert dialog, see https://github.com/deltachat/deltachat-android/pull/2621
   - [X] protection broken (more info & scan qr button) -> https://staging.delta.chat/733/en/help#verificationbroken
   - [x] protection enabled (also more info button) -> 'chat_protection_enabled_explanation', https://staging.delta.chat/733/en/help#verifiedchats
- [x] Block loading remote images in the "Full Msg View" not only for contact requests, also when the protection is broken (using the new method isHalfBlocked())
- [x] pull translations again, because the strings weren't available yet

Bildschirmfoto 2023-10-27 um 20 28 39

PR #3492, opened 2023-11-07
Automatically create verified group when all users are verified

adzialocha:
This PR simplifies the UI by removing all options to manually create a verified group chat. The idea is to automatically do so if all selected group members are known to be verified.

If an unverified member is added to an verified group a dialogue will be displayed. 

https://github.com/deltachat/deltachat-desktop/assets/1822473/5dec1da8-8f00-4cdc-a9fa-002c6c2aba1d

This entails:

* [x] Remove "Create Verified Group" from "New Chat" screen
* [x] Remove "QR Invite Code" from "Create Group" screen
* [x] When only verified contacts are selected in "New Group" screen, automatically create a verified group - otherwise a normal group
* [x] Set new stock string `DC_STR_NEW_GROUP_SEND_FIRST_MESSAGE` to `chat_new_group_hint`
* [x] Remove verified/unverified filtering of contact lists when adding members
* [x] When tapping an unverified contact in a verified group, show an alert with the text `verified_contact_required_explain` and the answer `learn_more` / `qrscan_title` / `ok` with scan-qr - similar to what we're doing at https://github.com/deltachat/deltachat-desktop/pull/3328
* [x] Verified status changes automatically in "Add new member to group" dialogue (after QR was scanned)

Tracking issue: https://github.com/deltachat/deltachat-desktop/issues/3468

### DIFF
--- a/src/renderer/components/dialogs/ProtectionStatusDialog.tsx
+++ b/src/renderer/components/dialogs/ProtectionStatusDialog.tsx
@@ -1,4 +1,5 @@
-import React, { CSSProperties, useContext } from 'react'
+import React, { useContext } from 'react'
+
 import { BackendRemote } from '../../backend-com'
 import { ScreenContext, useTranslationFunction } from '../../contexts'
 import { runtime } from '../../runtime'
@@ -8,13 +9,16 @@ import {
   DeltaDialogFooter,
   DeltaDialogFooterActions,
 } from './DeltaDialog'
-import type { DialogProps } from './DialogController'
 import QrCode from './QrCode'
 
-const WebLinks = {
-  Broken: 'https://staging.delta.chat/733/en/help#verificationbroken',
-  Enabled: 'https://staging.delta.chat/733/en/help#verifiedchats',
-}
+import type { DialogProps } from './DialogController'
+
+const VERIFICATION_BROKEN_URL =
+  'https://staging.delta.chat/733/en/help#verificationbroken'
+const VERIFICATION_ENABLED_URL =
+  'https://staging.delta.chat/733/en/help#verifiedchats'
+const VERIFICATION_REQUIRED_URL =
+  'https://staging.delta.chat/746/en/help#howtoe2ee'
 
 export function ProtectionBrokenDialog({
   name,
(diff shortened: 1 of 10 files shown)


src/renderer/components/dialogs/ViewGroup.tsx
```diff
@@ -110,22 +115,36 @@ export const useGroup = (chat: Type.FullChat) => {
...
+    const accountId = selectedAccountId()
+
+    // We can only add verified contacts to a protected group
+    if (chat.isProtected) {
+      const membersVerified = await areAllContactsVerified(
+        accountId,
+        newGroupMembers
+      )
+      if (!membersVerified) {
+        // @TODO: Show dialogue here
```
adzialocha, line 134: I'm not sure what the UX flow is for when a user tries to add an unverified contact to a verified group. I had a look at the [https://github.com/deltachat/deltachat-desktop/pull/3328](https://github.com/deltachat/deltachat-desktop/pull/3492/files#) flow. I assume we want to show the same dialogue (ProtectionBrokenDialog) then but a different QR code which allows the contact to scan it and join the group in a verified manner?

r10s: it is just not possibe to add unverified members, they must not even be selected

instead, an alert with "scan qr" and "learn more" is shown, but none of them will add unverified contacts to selected contacts

however, if the "qr code" scan allows to be opened while the add-member-dialog is still open (android/ios allow that), the list should be updated if verification-state changes

src/renderer/components/helpers/ChatMethods.ts
```diff
@@ -338,11 +338,14 @@ export async function modifyGroup(
 ) {
   const accountId = selectedAccountId()
   log.debug('action - modify group', { chatId, name, image, members })
-  await BackendRemote.rpc.setChatName(accountId, chatId, name)
+
   const chat = await BackendRemote.rpc.getFullChatById(accountId, chatId)
   if (!chat) {
     throw new Error('chat is undefined, this should not happen')
   }
```
Simon-Laux, line 345: `await BackendRemote.rpc.getFullChatById(accountId, chatId)` will already throw in that case so this if statement can be removed, probably comes from the time of using the cffi api in desktop where there is not so good error handling (check for 0 as hint there was an error, then fish the error from the core events, sometimes not even that)

getFullChatById can not return `null` or `undefined`, this is it's rust function signature:
```rs
 async fn get_full_chat_by_id(&self, account_id: u32, chat_id: u32) -> Result<FullChat>
```

src/renderer/components/dialogs/ViewGroup.tsx
adzialocha: > instead, an alert with "scan qr" and "learn more" is shown, but none of them will add unverified contacts to selected contacts

Super, thank you! This should be implemented now.

> however, if the "qr code" scan allows to be opened while the add-member-dialog is still open (android/ios allow that), the list should be updated if verification-state changes

I'm currently looking into that, not sure if it is part of this PR though as there are some edge cases to take into consideration:

1. Automatic update of newly added email address within the "add member" dialogue
2. Automatic update of existing addresses (for this I found the `ContactsChanged` event, but doesn't seem to be working, I'm still troubleshooting)

src/renderer/components/dialogs/ProtectionStatusDialog.tsx
```diff
@@ -8,13 +9,16 @@ import {
...
-  Enabled: 'https://staging.delta.chat/733/en/help#verifiedchats',
-}
+import type { DialogProps } from './DialogController'
+
+const VERIFICATION_BROKEN_URL =
+  'https://staging.delta.chat/733/en/help#verificationbroken'
+const VERIFICATION_ENABLED_URL =
+  'https://staging.delta.chat/733/en/help#verifiedchats'
+const VERIFICATION_REQUIRED_URL =
+  'https://staging.delta.chat/746/en/help#howtoe2ee'
```
Simon-Laux, line 21: maybe we should not duplicate it but move it instead to shared/constants file?

Simon-Laux: ah sorry read it wrong nvm

Simon-Laux (approved):

PR #3501, opened 2023-11-12
make sure verified icon is shown everywhere

### LINKED ISSUE #3498: "Green checkmark" is not always displayed when it should be
for little longer names or email addresses, the green checkmark is not shown.

this is eg. the member list of a guaranteed-e2ee group, where all members should have a green checkmark:

same happens in the title of guaranteed-e2ee group - and maybe at other places, when targeting this issues, all places where the  green checkmark should appear should be checked against longer names

nb: as discussed here and there: showing the green checkmark at the beginning of the title is not a fix - it is not what user know about these kind of checkmarks (external consistency), it a breask with android/ios and prior desktop versions (internal consistency), and it also clutters UI as the names begin with sth. the user did not type any may not have a good idea of

Simon-Laux:
- fix truncation so verified icon is always shown on ViewGroup and on ViewProfile
- always show verified icon in chatlist items
- fix trucation on contact list items
- fix truncation in navbar
- fix centering of verified icon
- improve spaceing in chatlistitems also trucate name in chatname in message search results
- more spacing
- center avatars
- css fixes

closes #3498

removes multiline display of display/group name in view group/profile, I couldn't get it to look good, maybe some other css wizard wants to try.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,7 +68,7 @@
 - fix window store installation (remove unknown language code from supported languages)
 - fix emojis in some html emails (force charset utf-8)
 - fix `null` account name when leaving the field empty bug
-- fix text truncation so verified icon is always shown on ViewGroup, ViewProfile and on ChatlistItem
+- fix text truncation so verified icon is always shown on ViewGroup, ViewProfile, ContactListItems and on ChatlistItem
 
 
 ### Removed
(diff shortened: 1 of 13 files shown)


missytake (approved):
works on linux.

adbenitez (approved):

PR #3512, opened 2023-11-14
Refactor dialog logic and add additional helper methods

adzialocha:
> This PR is a suggestion on how we might tackle the dialogue logic in this application, it might also serve as a "template" for further refactorings. Please see this as a first draft and suggestion, there's definitely ways to improve this approach, any feedback is welcome!

The general idea of this refactoring is to make it more clear on "first glance" _where_ something lives and _what it does_. The approach taken here was to move things into dedicated files, prefer React patterns (even though it looks sometimes more cumbersome), separate concerns better by changing  APIs and introducing dedicated contexts and hooks and disallow multiple paths to get to the same thing (for example `window.__openDialog`) / make things consistent.

A lot of changes in this PR are about changing imports, so rather uninteresting. The actual interesting bits are:

1. A new folder in `src/renderer/contexts` which became the home for all contexts which are somewhat used on the "root" level of the application, providing important functionality we want to have access to everywhere. I haven't really touched most contexts other than moving them into separate files
2. `DialogContext` is the one which was completly refactored, start there when trying to understand this PR! It derives its design from the previous implementation, is a bit slimmer and type-safe, also it doesn't require the dialogue logic to be in global scope (`window`)
3. `ScreenContext` got a bit less powerful and `DialogController` obsolete
4. A new folder in `src/renderer/hooks` which contains all important hooks related to the core functionality we want to use everywhere in the application, every hook goes into its own file
5. I've needed to touch upon the `KeybindingsContext` a bit as it used some dialogue logic as well, this is why this part also got partly refactored


Simon-Laux:
do you think we could also remove `isOpen: boolean` now? I mean if they are not open they are not in the dom anyway.

src/renderer/App.tsx
```diff
@@ -1,20 +1,16 @@
 import React, { useState, useEffect, useLayoutEffect } from 'react'
-import { i18nContext } from './contexts'
-import ScreenController from './ScreenController'
-import attachKeybindingsListener from './keybindings'
+import moment from 'moment'
```
adzialocha, line 2: I'm grouping all external and internal dependencies visually to make it clearer where things come from. I could see that in some files that already was the case. This PR is not about this little detail, but as soon as I touched a file I reorganised them.

This thing can be linted (sorted, grouped imports) + autofixed btw, happy to make a PR with the tooling for that.

src/renderer/ScreenController.tsx
```diff
@@ -61,19 +58,13 @@ export default class ScreenController extends Component {
...
     this.onShowAbout = this.showAbout.bind(this)
     this.onShowKeybindings = this.showKeyBindings.bind(this)
     this.onShowSettings = this.showSettings.bind(this)
-    this.dialogController = createRef()
     this.selectAccount = this.selectAccount.bind(this)
 
-    window.__openDialog = this.openDialog.bind(this)
     window.__userFeedback = this.userFeedback.bind(this)
-    window.__closeDialog = this.closeDialog.bind(this)
-    window.__hasOpenDialogs = this.hasOpenDialogs.bind(this)
```
adzialocha, line 76: Putting the dialog controls on the global window allowed for some neat developer hacks to access dialogues wherever one is but it also made it hard to grasp how things are tied together and where they came from

src/renderer/ScreenController.tsx
```diff
@@ -165,20 +156,25 @@ export default class ScreenController extends Component {
       }
     }
 
-    runtime.onOpenQrUrl = processOpenQrUrl
+    runtime.onOpenQrUrl = (url: string) => {
+      processOpenQrUrl(this.context.openDialog, this.context.closeDialog, url)
```
adzialocha, line 160: .. by removing `openDialog` from `window` we end up doing this now (passing `openDialog` into every method which doesn't have access to the React context / is not a component or a hook).

It doesn't necessarily need to stay like this though, some of these "helper" methods can become hooks and / or components themselves as well but I didn't wanted to touch this part as well.

src/renderer/ScreenController.tsx
```diff
@@ -211,7 +207,7 @@ export default class ScreenController extends Component {
   }
 
   showAbout() {
-    this.openDialog('About')
+    this.context.openDialog(About)
```
adzialocha, line 210: The new API disallows strings to identify dialogues and only accepts (type-checked) React elements, this was already possible before but not consistently used across the code-base

src/renderer/components/ConnectivityToast.tsx
```diff
@@ -1,21 +1,16 @@
...
+
+import { getLogger } from '../../shared/logger'
+import { KeybindAction } from '../keybindings'
 import { debounceWithInit } from './chat/ChatListHelpers'
 import SettingsConnectivityDialog from './dialogs/Settings-Connectivity'
 import { BackendRemote, onDCEvent } from '../backend-com'
 import { selectedAccountId } from '../ScreenController'
+import { useDialog } from '../hooks/useDialog'
+import { useTranslationFunction } from '../hooks/useTranslationFunction'
+import { useKeyBindingAction } from '../hooks/useKeyBindingAction'
```
adzialocha, line 13: Each hook became its own file, which hopefully makes it easier now to find the code faster when one wants to look at it / change it

src/renderer/components/LoginForm.tsx
```diff
@@ -467,10 +472,9 @@ export function ConfigureProgressDialog({
   }, [accountId])
 
   return (
-    <Dialog
-      isOpen={isOpen}
+    <DeltaDialogBase
```
adzialocha, line 475: Every dialogue uses now `DeltaDialogBase` in the end which makes it hopefully more clear how they all relate to each other:

- `Dialog` from `blueprintjs` > `DeltaDialogBase` > Any other dialogue

src/renderer/App.tsx
```diff
@@ -1,20 +1,16 @@
...
-import moment from 'moment'
 import { CrashScreen } from './components/screens/CrashScreen'
 import { runtime } from './runtime'
 import { updateCoreStrings } from './stockStrings'
 import { getLogger } from '../shared/logger'
 import { BackendRemote } from './backend-com'
 import { runPostponedFunctions } from './onready'
-
-attachKeybindingsListener()
+import { I18nContext } from './contexts/I18nContext'
```
adzialocha, line 13: Every context got its own file

src/renderer/components/dialogs/DeltaDialog.tsx
```diff
@@ -1,47 +1,52 @@
...
+    canEscapeKeyClose = true,
+    canOutsideClickClose = true,
+    showCloseButton = true,
+    ...props
+  }) => {
+    const isFixed = props.fixed
+    return (
+      <>
+        <Dialog
+          isOpen={true}
```
adzialocha, line 31: This is the only place now where we set `isOpen`. Since every other component is either rendered (opened) or not rendered (closed) we didn't really make use of this property anyhow

src/renderer/contexts.ts
```diff
@@ -1,52 +0,0 @@
-import React, { useContext } from 'react'
```
adzialocha, line 1: I moved every context into its own file

src/renderer/contexts/DialogContext.tsx
```diff
@@ -0,0 +1,91 @@
...
+  hasOpenDialogs: false,
+  openDialog: _ => '',
+  closeDialog: _ => {},
+}
+
+export const DialogContext = React.createContext<DialogContextValue>(
+  initialValues
+)
+
+export const DialogContextProvider = ({ children }: PropsWithChildren<{}>) => {
```
adzialocha, line 38: This is the "heart" of this refactor PR: A new DialogContext and Provider. It is inspired by the previous implementation which solved a couple of problems:

* Have knowledge about if there's currently a dialogue in the view or not
* Have a common place where they are rendered
* Have a nice convenient `openDialog` API with type checks on the properties

src/renderer/contexts/KeybindingsContext.tsx
```diff
@@ -0,0 +1,63 @@
...
+      }
+    }
+  })
+
+  // @TODO: This probably needs another place
+  useKeyBindingAction(KeybindAction.KeybindingCheatSheet_Open, () => {
+    if (!window.__keybindingsDialogOpened) {
+      openDialog(KeybindingCheatSheet)
+    }
+  })
```
adzialocha, line 38: These key-bindings here are important on a base-level for the application but still they don't belong here (as this context should be an agnostic of how it is used).

I will probably move them somewhere else

src/renderer/contexts/ScreenContext.tsx
```diff
@@ -0,0 +1,18 @@
...
+export const ScreenContext = createContext({
+  /**
+   * Shows a context menu
+   *
+   * @returns a promise with no return value that gets resolved when the context
+   * menu disapears again regardless what action the user took or if they
+   * canceled the dialog
+   */
+  openContextMenu: async (..._args: Parameters<OpenContextMenu>) => {},
+  userFeedback: (_message: false | userFeedback) => {},
```
adzialocha, line 15: I can see further refactorings where `openContextMenu` and `userFeedback` get moved into their own regarding contexts as well. Then it gets more clear where this logic lives / better code separation.

src/renderer/components/dialogs/AlertDialog.tsx
```diff
@@ -13,18 +16,16 @@ export default function AlertDialog({
 }: {
   cb?: () => void
   message: string | JSX.Element
-  onClose: () => void
-}) {
+} & DialogProps) {
   const tx = useTranslationFunction()
-  const isOpen = !!message
```
adzialocha, line 19: Wondering what this did before? Can it be removed like that?

adzialocha: There's similar other places like that

src/renderer/contexts/DialogContext.tsx
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  return (
+    <DialogContext.Provider value={value}>
+      {children}
+      {Object.keys(dialogs).map(id => {
+        return dialogs[id]
+      })}
+    </DialogContext.Provider>
+  )
```
adzialocha, line 90: Since the provider now comes as a custom thing we can conveniently offer a place to render the dialogues, this makes `DialogController` obsolete

src/renderer/App.tsx
Simon-Laux: > This thing can be linted (sorted, grouped imports) + autofixed btw, happy to make a PR with the tooling for that.

sure why not, though only if it does not add too many new dependencies ;)

src/renderer/utils/random.ts
```diff
@@ -0,0 +1,3 @@
+export function generateRandomUUID() {
+  return crypto.randomUUID()
+}
```
farooqkz, line 3: Why cryptographically secure UUIDs are used here?

src/renderer/contexts/DialogContext.tsx
```diff
@@ -0,0 +1,91 @@
...
+  openDialog: _ => '',
+  closeDialog: _ => {},
+}
+
+export const DialogContext = React.createContext<DialogContextValue>(
+  initialValues
+)
+
+export const DialogContextProvider = ({ children }: PropsWithChildren<{}>) => {
+  const [dialogs, setDialogs] = useState<{ [id: DialogId]: JSX.Element }>({})
```
farooqkz, line 39: I feel this can be written a bit more readable. How about `Map<DialogId, JSX.Element>`?

I know Map and Object are not the same. But why an object is used here rather than a map?

src/renderer/contexts/DialogContext.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+  const closeDialog = useCallback((id: DialogId) => {
```
farooqkz, line 41: Did you mean this?
```suggestion
  const closeDialog = useCallback<CloseDialog>((id: DialogId) => {
```

farooqkz:
I don't have much of an opinion on this PR. It is more of a learning opportunity for me.

src/renderer/components/dialogs/MuteChat.tsx
```diff
@@ -1,20 +1,20 @@
...
 
 export default function MuteChat({
-  isOpen,
   onClose,
   chatId,
 }: {
-  isOpen: boolean
   onClose: () => void
   chatId: number
 }) {
```
farooqkz, line 15: Should not it be part of `DialogProps`?

Simon-Laux: it is not needed anymore because of this pr

src/renderer/contexts/DialogContext.tsx
Simon-Laux: `Record<keytype,valuetype>` can be used for objects https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type

I personally also of the use that syntax (`{[key: type]: type}`), so it's fine for me, I don't really have a preference between the two but maybe `Record` is better?

src/renderer/utils/random.ts
adzialocha: Good question! I don't really need the cryptographically secure randomness here but it seemed an easy way to get a random enough string without a lot of effort (vs. building a counter logic or my own way to use Math.random())

src/renderer/contexts/DialogContext.tsx
adzialocha: > I feel this can be written a bit more readable. How about `Map<DialogId, JSX.Element>`?
> 
> I know Map and Object are not the same. But why an object is used here rather than a map?

Maps are really quite different than objects and we're not using them as far as I know (you can't just use `{}` as an replacement of a Map).

`Record` as @treefit suggests is the closest replacement which would work.

I'm also open to any way of writing this type down (Record or `{}` syntax), happy to change it in whatever way we decide and stick to it.

src/renderer/components/dialogs/MuteChat.tsx
adzialocha: > Should not it be part of `DialogProps`?

`onClose` should be taken from `DialogProps` though 👍🏻 I fix that!

src/renderer/contexts/DialogContext.tsx
adzialocha: Yes! Though, Typescript should already know the type, as `closeDialog` is part of `value` which again is part of the context which knows `CloseDialog` (check `DialogContextValue`). I wouldn't mind adding it here though even if it's redundant, it communicates stronger what it is about

src/renderer/utils/random.ts
farooqkz: I believe there is a UUID package we can use.

src/renderer/contexts/DialogContext.tsx
farooqkz: I strongly suggest `Record` as it's more readable and feels more standard.

src/renderer/hooks/useDialog.ts
```diff
@@ -0,0 +1,13 @@
+import { useContext } from 'react'
+
+import { DialogContext } from '../contexts/DialogContext'
+
+export function useDialog() {
```
farooqkz, line 5:
```suggestion
export default function useDialog() {
```

This makes more sense I think?

src/renderer/utils/random.ts
Simon-Laux: please don't add new packages unless they are really necessary.
more packages bring more dependencies and with more dependencies grows the risk of getting an infected/vulnerable one. Not to mention build/install time and download size.

If there is a browser api directly for this it will likely be faster than a js implementation of the uuid function: https://dev.to/galkin/crypto-randomuuid-vs-uuid-v4-47i5

src/renderer/contexts/DialogContext.tsx
Simon-Laux: note that I think it's not worth to discuss about this small cosmetic thing here...

src/renderer/hooks/useDialog.ts
Simon-Laux: would need to be changed in files that import it too I guess?

src/renderer/utils/random.ts
farooqkz: yeah it might be faster but we are using crypto api which is not needed in this case. Doesn't a simple random large enough int suffice?

src/renderer/hooks/useDialog.ts
farooqkz: yeah

src/renderer/utils/random.ts
Simon-Laux: I think it wants a string here. if you a really into it you can benchmark what is faster, but then again it does not make a difference because this is about opening dialogs, a code path that is not called that often. Also the crypto api is available at all browser targets we want to support, so I see no problem with using it.
If that should become a problem in the future, then we can still change it later.

Simon-Laux (approved):
Looks good, cleaning that up was long overdue, thanks for taking this on.

I think we can merge to master directly, 1.42 is done enough and soon fully released.

PR #3513, opened 2023-11-14
update local help

r10s:
this pulls in all help texts and anchors for guaranteed end-to-end encryption (and more)


Simon-Laux:
Thanks, changelog entry with date is still needed for it to be merged.

PR #3518, opened 2023-11-15
use local help for guaranteed end-to-end encryption more info

### LINKED ISSUE #3514: update guaranteed end-to-end encryption links or use local help
guaranteed end-to-end encryption help is merged to master, before 1.42, the following links must be used instead of the staging ones:

- https://delta.chat/help#howtoe2ee -  when the user tries to add a non-verified-contact to a guaranteed-e2ee-group and hits "learn more"
- https://delta.chat/help#e2eeguarantee - when "everything's fine" and the user taps "learn more" on a green checkmark info
- https://delta.chat/help#nocryptanymore - when the user taps "learn more" on broken encryption info

or, even better, links from the dialogs to the local help using the same anchors (after https://github.com/deltachat/deltachat-desktop/pull/3513 is merged)

Simon-Laux:
closes #3514

### DIFF
--- a/src/main/windows/help.ts
+++ b/src/main/windows/help.ts
@@ -32,7 +32,9 @@ export async function openHelpWindow(locale: string, anchor?: string) {
     win.focus()
     if (anchor) {
       win.webContents.executeJavaScript(`
-        document.getElementById(atob("${btoa(anchor)}"))?.scrollIntoView({"behavior":"smooth"})
+        document.getElementById(atob("${btoa(
+          anchor
+        )}"))?.scrollIntoView({"behavior":"smooth"})
       `)
     }
     return
@@ -76,7 +78,9 @@ export async function openHelpWindow(locale: string, anchor?: string) {
   win.once('ready-to-show', async () => {
     if (anchor) {
       await help_window.webContents.executeJavaScript(`
-      document.getElementById(atob("${btoa(anchor)}"))?.scrollIntoView({"behavior":"instant"})
+      document.getElementById(atob("${btoa(
+        anchor
+      )}"))?.scrollIntoView({"behavior":"instant"})
       `)
     }
     help_window.show()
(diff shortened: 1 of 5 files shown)


src/main/windows/help.ts
```diff
@@ -27,9 +27,16 @@ async function getHelpFileForLang(locale: string) {
...
 
-export async function openHelpWindow(locale: string) {
+export async function openHelpWindow(locale: string, anchor?: string) {
   if (win) {
     win.focus()
+    if (anchor) {
+      win.webContents.executeJavaScript(`
+        document.getElementById(atob("${btoa(
+          anchor
+        )}"))?.scrollIntoView({"behavior":"smooth"})
```
r10s, line 37: when opening the help window from "learn more", there is a small flickering for a fraction of a second, looks as if things scroll down fast.

this animation is needless and not helping, it would be better, to just open the help at the given position directly, without scrolling for flickering.

maybe `{"behavior":"instant"}`(cmp. https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView ) is the better option here?

r10s (approved):
very nice!

(tested with the green-checkmark info)

src/main/windows/help.ts
Simon-Laux: I thought it would be better if it reacts somehow, but sure we can set this to instant, though maybe users will think nothing changed then?

Simon-Laux: also if the window is not already open it is instant.

r10s: > also if the window is not already open it is instant

it is not for me (macos, m1 pro), this is the flickering i mentioned

but it is still okayish, and yes, having some feedback when the help is already open is nice.

so, let's leave things as is, it is good enough :)

Simon-Laux: There are two different cases if it is not open it scrolled instantly and if it is already open it scrolled smoothly.
Anyways now I removed the smooth scroll, but I'm not sure now if your issue is not a different one?

r10s: solved by recent commits

r10s (approved):
yeah, 💯% instant with the recent commits - and also makes sense to scroll when the window is already open

PR #3519, opened 2023-11-15
streamline headlines in profiles

r10s:
the headlines in profiles should just read what the "profile" is about - so "Contact", "Group", "Broadcast List" -
not that it is a "profile" (whatever that is from the view of the user).

also the 'Edit' prefix is not correct,
esp. confusing as there is a dialog that open the thing for editing.

these are the headlines that we're using also on android/ios, and that are used in similar ways also in similar app.

left=before / right=after:

 &nbsp; &nbsp; 
&nbsp;

 &nbsp; &nbsp; 

(nb: it is correct that "tunis1000" does not have a green checkmark, i just used that to test breaking encryption)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - update deltachat-node and deltachat/jsonrpc-client to `v1.131.2`
 - Update inApp help (15.11.2023)
 - make help's "scroll to top" button less intrusive
+- streamline profile titles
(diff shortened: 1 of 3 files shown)


Simon-Laux (approved):

Simon-Laux:
formatting still needs to be fixed with `npm run fix`

Issue #3499, opened 2023-11-11
open "device messages" in case no chat is selected just after installation


r10s:
currently, the app is quite empty directly after setting up an account.

<img width="1512" alt="Screenshot 2023-11-11 at 22 50 06" src="https://github.com/deltachat/deltachat-desktop/assets/9800740/42d703e1-45bf-4560-860c-e2f1a547c294">

we can easily make better use of all that space by selecting the "Device Messages" chat directly after first log in.

this also fits nicely to the idea of guiding the user more on-point by device-messages describing eg. to to get in contact with other ppl. but already now, with current device messages, it will be an improvement (ppl will select the device chat anyway quickly, as there is not much else to select - and then the current hint is anyways gone)

(android/ios do not have this issue as they're showing anyways the chatlist with "Device Messages" atop, which is very probably be tapped as well)

EDIT: not needed for 1.42

PR #3531, opened 2023-11-16
Select device message chat automatically to welcome first users

adzialocha:
Closes https://github.com/deltachat/deltachat-desktop/issues/3499

### DIFF
--- a/src/renderer/components/LoginForm.tsx
+++ b/src/renderer/components/LoginForm.tsx
@@ -443,7 +443,7 @@ export function ConfigureProgressDialog({
               'ui.lastchatid',
               String(deviceChatId)
             )
-            // SettingsStoreInstance is reloaded the first time the main screen in shown
+            // SettingsStoreInstance is reloaded the first time the main screen is shown
           }
 
           // Yay! We're done and ready to go
--- a/src/renderer/components/LoginForm.tsx
+++ b/src/renderer/components/LoginForm.tsx
@@ -438,10 +438,12 @@ export function ConfigureProgressDialog({
           // as a first introduction to the app after they've entered
           const deviceChatId = await getDeviceChatId(accountId)
           if (deviceChatId) {
-            SettingsStoreInstance.effect.setCoreSetting(
+            await BackendRemote.rpc.setConfig(
+              accountId,
               'ui.lastchatid',
               String(deviceChatId)
             )
+            // SettingsStoreInstance is reloaded the first time the main screen in shown
           }
 
           // Yay! We're done and ready to go
--- a/src/renderer/components/LoginForm.tsx
+++ b/src/renderer/components/LoginForm.tsx
@@ -47,6 +47,30 @@ export function defaultCredentials(credentials?: Credentials): Credentials {
   return { ...defaultCredentials, ...credentials }
 }
 
+/**
+ * Helper method to determine the chat id of the "Device Messages" read-only chat.
+ *
+ * Note that there's currently no API to retrieve this id from the backend, this
+ * is why we're iterating over all currently available chats instead.
+ */
+async function getDeviceChatId(accountId: number): Promise<number | null> {
+  const chatIds = await BackendRemote.rpc.getChatlistEntries(
+    accountId,
+    null,
+    null,
+    null
+  )
+
+  for (const chatId of chatIds) {
+    const chat = await BackendRemote.rpc.getFullChatById(accountId, chatId)
+    if (chat.isDeviceChat) {
+      return chatId
+    }
+  }
+
+  return null
+}
+
 type LoginProps = React.PropsWithChildren<{
   credentials: Credentials
   setCredentials: (credentials: Credentials) => void
@@ -369,6 +393,7 @@ export function ConfigureProgressDialog({
   const [error, setError] = useState('')
   const [configureFailed, setConfigureFailed] = useState(false)
   const accountId = selectedAccountId()
+  const tx = useTranslationFunction()
 
   const onConfigureProgress = ({
     progress,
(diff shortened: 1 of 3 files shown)


adzialocha:
@r10s, do we want this on _every_ start of DC when no chat has been selected (which is only the case when user has not been active yet or just started to use DC) or _only_ on the very first start?

r10s:
> do we want this on every start of DC when no chat has been selected (which is only the case when user has not been active yet or just started to use DC) or only on the very first start?

i'd do that on every start until the user selects some chat the first time - then, never do that again.

also, if a chat is deleted, we should show the current empty view saying "Select a chat or create a new chat" - and that is probably the moment we should take care not to re-show the device-chat on startup

src/renderer/components/screens/MainScreen.tsx
```diff
@@ -118,6 +118,11 @@ export default function MainScreen() {
       ]
       if (lastChatId) {
         selectChat(Number(lastChatId))
+      } else {
+        // Welcome the user with the "Device Messages" chat in case no chat was
+        // selected previously (for example when they've just joined Delta Chat
+        // for the first time)
+        selectChat(C.DC_CONTACT_ID_DEVICE)
```
r10s, line 125: this does not work, DC_CONTACT_ID_DEVICE is a contact-id, not a chat-id.

there is no api to get the device-chat-id directly; however, as there are only two chats on startup, just iterate over them and check for [`dc_chat_is_device_talk()`](https://c.delta.chat/classdc__chat__t.html#a560a4ccfa512cad1ecb066ac275b76b8) resp. the corresponding jsonrpc call (i think, this is good enough, the effort to create and maintain a new api is probably higher)

Simon-Laux:
> i'd do that on every start until the user selects some chat the first time - then, never do that again.

It should only be set for newly created accounts, not for existing accounts.

Create/Login Account -> Open Device chat
Existing account starts up with no account selected -> do nothing special
Chat was deleted -> do nothing special

I think while configuring we can set the ui config for last selected chat, maybe even already before calling `configure` if the device chat already was created at that stage. 🤔

r10s:
the current approach seems to work already with deleted chats (as `ui.lastchatid` is not removed with chat deletion), so, it is only up to select the correct chat

adzialocha:
> > i'd do that on every start until the user selects some chat the first time - then, never do that again.
> 
> It should only be set for newly created accounts, not for existing accounts.

Makes sense :+1:

> I think while configuring we can set the ui config for last selected chat, maybe even already before calling `configure` if the device chat already was created at that stage. 🤔

That's a good call! I was pondering how to do this without introducing another "loading" state before we show the main UI (even if its a fraction of time, it might look buggy), that would be a nice solution.

Would these implementation steps make sense for you, or am I'm missing already existing functionality?

1. Implement logic to find out chat id for device message chat
2. Add a true flag to `SettingsStore` to keep track of if I just created a new account, since we need that state
3. If app gets opened with this flag set to true we move directly to the message device chat
4. Flip that flag to false as soon as we call `selectChat` with a "real chat" for the first time

Simon-Laux:
How about instead:

1. set 'ui.lastchatid' over settings manager to `"device-chat"`
2. check in the function that selects the chat from 'ui.lastchatid' for device chat, then in that case find device chat and select it

src/renderer/components/LoginForm.tsx
```diff
@@ -398,17 +423,24 @@ export function ConfigureProgressDialog({
...
-          })
-          await BackendRemote.rpc.configure(window.__selectedAccountId)
+          }
+
+          // Select "Device Messages" chat as the initial one. This will serve
+          // as a first introduction to the app after they've entered
+          const deviceChatId = await getDeviceChatId(accountId)
+          if (deviceChatId) {
+            initialConfig['ui.lastchatid'] = String(deviceChatId)
+          }
```
Simon-Laux, line 437: this needs to be done after configuration, also might be better to directly user the settings manager as the value is also fetched from there then we don't need to care for updating the cached value in the settings manager.

adzialocha: > this needs to be done after configuration (before that device chat does not exist, just tested it)

Ah, thanks for testing! In my case it actually worked, must be a race condition? I'll move it further to avoid it.

r10s: imu, [`configure()` waits for device chat being created](https://github.com/deltachat/deltachat-core-rust/blob/main/src/configure.rs#L486) [by another thread](https://github.com/deltachat/deltachat-core-rust/blob/main/src/configure.rs#L182), so after `configure()` has finished we should be safe.

but while `configure()` runs we may or may not have the chat

Simon-Laux: (this is about the bare core calls, jsonrpc may add further pitfalls :)

in the case of configure it only calls stop_io and start_io automatically.

Simon-Laux (changes requested):
Does not work for me ;/

Simon-Laux:
I make a small pr to this pr to fix it

Simon-Laux:
> I make a small pr to this pr to fix it

was only a little thing so pushed directly yo your branch, works I think we can merge it

src/renderer/components/LoginForm.tsx
```diff
@@ -398,17 +424,29 @@ export function ConfigureProgressDialog({
...
+          // Select "Device Messages" chat as the initial one. This will serve
+          // as a first introduction to the app after they've entered
+          const deviceChatId = await getDeviceChatId(accountId)
+          if (deviceChatId) {
+            await BackendRemote.rpc.setConfig(
+              accountId,
+              'ui.lastchatid',
+              String(deviceChatId)
+            )
+            // SettingsStoreInstance is reloaded the first time the main screen in shown
```
Simon-Laux, line 446:
```suggestion
            // SettingsStoreInstance is reloaded the first time the main screen i shown
```

Simon-Laux (approved):

adzialocha:
> > I make a small pr to this pr to fix it
> 
> was only a little thing so pushed directly yo your branch, works I think we can merge it

Interesting! I was wondering if it was the store as well, thank you for looking into it!

Issue #3535, opened 2023-11-16
reply menu option not disabled in read-only mailing list, draft set


adbenitez:
the menu option "reply" when you right-click a message in a read-only mailing list should be hidden, as the composer is disabled and people can't reply, the draft shouldn't be set

- Operating System (Linux/Mac/Windows/iOS/Android): Linux
- Delta Chat Version: 1.41.2 (git: v1.33.0-512-gf81e95b2)

![image](https://github.com/deltachat/deltachat-desktop/assets/24558636/834ee63a-a445-44b5-b700-60364415b932)

![image](https://github.com/deltachat/deltachat-desktop/assets/24558636/146f846e-e7e6-4360-835c-efdb72097731)

PR #3536, opened 2023-11-16
Fix: Do not show reply option in read-only groups

adzialocha:
Closes: https://github.com/deltachat/deltachat-desktop/issues/3535

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 ### Fixed
 - fix: files search not case-insensitive
 - fix: bug in emoji detection for jumbomoji #3508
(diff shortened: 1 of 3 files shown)


Simon-Laux (approved):
LGTM

r10s:
@Simon-Laux @adzialocha if not yet done somehow if could not get (i looked only at the diff) i think, we should better use [`dc_chat_can_send()`](https://c.delta.chat/classdc__chat__t.html#acb894b6d56253effbc886299add54872) to check if one can reply to a message

eg. there are mailing lists, where one can reply to or you cannot reply to contact request or if protection is broken; this and more is checked by `can_send()`

PR #3532, opened 2023-11-16
Do not show redundant read-only message when in device messages chat

adzialocha:
This PR hides the composer view including it's little message when being in the read-only "Device Message" chat as it didn't add much information for the user.

Also, I've done a mini-refactoring, moving this logic into own component and hook files.

Closes https://github.com/deltachat/deltachat-desktop/issues/3511

PR #3537, opened 2023-11-16
Improve layout and fix unknown locale of `DisabledMessageInput`

adzialocha:
Follow up PR from https://github.com/deltachat/deltachat-desktop/pull/3532

**Minor fixes:**

* Fix warning for using an unknown translation by just returning an debug string (this is at least how it was in the version before)
* Remove `useRef` as it doesn't really have a use in this scenario

**Layout fix:**

CSS: Center text always, also in narrow window sizes, independent of padding

Before:

2023-11-16-202352_678x76_scrot

After:

2023-11-16-202413_682x64_scrot

### DIFF
--- a/src/renderer/components/composer/Composer.tsx
+++ b/src/renderer/components/composer/Composer.tsx
@@ -305,7 +305,7 @@ const Composer = forwardRef<
       </div>
     )
   } else if (isDisabled) {
-    return <DisabledMessageInput ref={ref} reason={disabledReason} />
+    return <DisabledMessageInput reason={disabledReason} />
   } else {
     return (
       <div className='composer' ref={ref}>
--- a/src/renderer/components/composer/DisabledMessageInput.tsx
+++ b/src/renderer/components/composer/DisabledMessageInput.tsx
@@ -21,7 +21,7 @@ const DisabledMessageInput = ({ reason }: Props) => {
       case DisabledChatReasons.DEVICE_CHAT:
         return tx('messaging_disabled_device_chat')
       case DisabledChatReasons.UNKNOWN:
-        // Unknown cases are likely to be caused by a bug somewhere else,
+        // Unknown cases are likely to be caused by a new case introduced by a new core update that is not yet handled here,
         // but we don't want to crash the UI
         return 'messaging_disabled_unknown'
       default:
--- a/src/renderer/components/composer/DisabledMessageInput.tsx
+++ b/src/renderer/components/composer/DisabledMessageInput.tsx
@@ -18,8 +18,6 @@ const DisabledMessageInput = ({ reason }: Props) => {
         return tx('messaging_disabled_not_in_group')
       case DisabledChatReasons.DEADDROP:
         return tx('messaging_disabled_deaddrop')
-      case DisabledChatReasons.UNKNOWN:
-        return 'UNKNOWN_DISABLED_CHAT_REASON'
     }
   }, [reason, tx])
 
--- a/src/renderer/components/composer/DisabledMessageInput.tsx
+++ b/src/renderer/components/composer/DisabledMessageInput.tsx
@@ -1,4 +1,4 @@
-import React, { forwardRef, useMemo } from 'react'
+import React, { useMemo } from 'react'
 
 import { useTranslationFunction } from '../../contexts'
 import { DisabledChatReasons } from './useIsChatDisabled'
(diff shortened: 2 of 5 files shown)


adzialocha:
We could also throw an Error here in case we arrive in the "unknown" state, though I wonder if that wouldn't be to extreme

src/renderer/components/composer/DisabledMessageInput.tsx
```diff
@@ -7,40 +7,37 @@ type Props = {
...
+  const reasonMessage = useMemo(() => {
+    switch (reason) {
+      case DisabledChatReasons.MAILING_LIST:
+        return tx('messaging_disabled_mailing_list')
+      case DisabledChatReasons.NOT_IN_GROUP:
+        return tx('messaging_disabled_not_in_group')
+      case DisabledChatReasons.DEADDROP:
+        return tx('messaging_disabled_deaddrop')
+      case DisabledChatReasons.UNKNOWN:
+        return 'UNKNOWN_DISABLED_CHAT_REASON'
```
adzialocha, line 22: We could also throw an Error here in case we arrive in the "unknown" state

src/renderer/components/composer/Composer.tsx
```diff
@@ -305,7 +305,7 @@ const Composer = forwardRef<
       </div>
     )
   } else if (isDisabled) {
-    return <DisabledMessageInput ref={ref} reason={disabledReason} />
```
adzialocha, line 308: The `ref` is used by `MessageListAndComposer` to automatically give focus on the input area. Since none input exists at this point (read-only group) we also don't need that reference

Simon-Laux: good catch

Simon-Laux:
I think I would keep the unknown, but throw the error instead when the `DisabledChatReasons` variant is not known.

Why? to not crash the whole app when core adds a new feature to disable sending in a new case. but to have the error when desktop devs forget to fully implement a `DisabledChatReasons` variant

src/renderer/components/composer/DisabledMessageInput.tsx
```diff
@@ -7,40 +7,39 @@ type Props = {
...
+    switch (reason) {
+      case DisabledChatReasons.MAILING_LIST:
+        return tx('messaging_disabled_mailing_list')
+      case DisabledChatReasons.NOT_IN_GROUP:
+        return tx('messaging_disabled_not_in_group')
+      case DisabledChatReasons.DEADDROP:
+        return tx('messaging_disabled_deaddrop')
+      case DisabledChatReasons.DEVICE_CHAT:
+        return tx('messaging_disabled_device_chat')
+      default:
```
Simon-Laux, line 23:
```suggestion
      case DisabledChatReasons.UNKNOWN:
        return tx('messaging_disabled_unknown');
      default:
```

adzialocha: Ah, sorry, I've misread your point above.

We do not have any message for `messaging_disabled_unknown`, can I just go ahead and add one for `en`?

Simon-Laux: we also put an untranslated string there, point is just that people should not get a crash screen in normal usage and english error is better than that.

Simon-Laux: we developers on the other hand should get a crash screen when we mess up in this case that's why I proposed it that way ;)

adzialocha: > we also put an untranslated string there

I receive an annoying warning in my console that the translation is missing, this is why I've created that PR actually first-hand 😅

adzialocha:
How is this now @Simon-Laux? Now as you said the unknown case is not crashing but returning a "weird" string to signal something is off.

I didn't use the `tx` method as already mentioned, to not clutter the dev console with annoying "translation missing" warnings.

src/renderer/components/composer/DisabledMessageInput.tsx
```diff
@@ -7,40 +7,43 @@ type Props = {
...
+      case DisabledChatReasons.MAILING_LIST:
+        return tx('messaging_disabled_mailing_list')
+      case DisabledChatReasons.NOT_IN_GROUP:
+        return tx('messaging_disabled_not_in_group')
+      case DisabledChatReasons.DEADDROP:
+        return tx('messaging_disabled_deaddrop')
+      case DisabledChatReasons.DEVICE_CHAT:
+        return tx('messaging_disabled_device_chat')
+      case DisabledChatReasons.UNKNOWN:
+        // Unknown cases are likely to be caused by a bug somewhere else,
```
Simon-Laux, line 24:
```suggestion
        // Unknown cases are likely to be caused by a new case introduced by a new core update that is not yet handled here,
```

Simon-Laux (approved):

Issue #3540, opened 2023-11-17
"reply" function degraded on master


r10s:
eg. for mailing lists, the "reply" function is no longer available since the attempt to fix https://github.com/deltachat/deltachat-desktop/issues/3535 in https://github.com/deltachat/deltachat-desktop/pull/3536

reason is that the menu entry is hidden because of chattype instead of checking [`dc_chat_can_send()`](https://c.delta.chat/classdc__chat__t.html#acb894b6d56253effbc886299add54872) - which is also much better in checking other states as contact-request or guaranteed-e2ee-broken

adzialocha:
Oh! Dammit. So mailing lists are actually writable?

> > [dc_chat_can_send()](https://c.delta.chat/classdc__chat__t.html#acb894b6d56253effbc886299add54872)

Where is this method reflected in the Rust implementation, RPC and / or TypeScript wrappers? I can try a deep dive but maybe faster to ask here. @Simon-Laux @r10s

adzialocha:
> Where is this method reflected in the Rust implementation, RPC and / or TypeScript wrappers? I can try a deep dive but maybe faster to ask here. @Simon-Laux @r10s

Okay, I assume it is `canSend` on the chat objects

r10s:
> Oh! Dammit. 

no worries, you could not have known that, even long-term members are struggling with state of api & co here and there :)

> So mailing lists are actually writable?

yes, most times. but there are situation where this is not possible or not wantend, for whatever reason. in these cases can_send() returns false. so UI should not worry about the "why", which may also change over time by adding features or fixing bugs. UI should just use these information. in general, the whole api is meant to be very high level, and serving the UIs concrete needs

PR #3541, opened 2023-11-17
Use `dc_chat_can_send` to show or hide reply functionality

adzialocha:
Closes https://github.com/deltachat/deltachat-desktop/issues/3540

### DIFF
--- a/src/renderer/components/message/Message.tsx
+++ b/src/renderer/components/message/Message.tsx
@@ -1,3 +1,10 @@
+import React, { useContext } from 'react'
+import reactStringReplace from 'react-string-replace'
+import { C, T } from '@deltachat/jsonrpc-client'
+import classNames from 'classnames'
+
+import MessageBody from './MessageBody'
+import MessageMetaData from './MessageMetaData'
 import {
   onDownload,
   openAttachmentInShell,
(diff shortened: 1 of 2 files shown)


r10s (commented):
lgtm, thanks a lot 🙏

 again, i leave final approval to the ppl more into the desktop code

src/renderer/components/message/Message.tsx
```diff
@@ -289,27 +282,33 @@ function buildContextMenu(
...
   const { openContextMenu, openDialog } = screenContext
 
   const showMenu: (
     event: React.MouseEvent<HTMLDivElement | HTMLAnchorElement, MouseEvent>
-  ) => void = event => {
+  ) => Promise<void> = async event => {
+    const chat = await BackendRemote.rpc.getFullChatById(
+      accountId,
+      message.chatId
+    )
```
Simon-Laux, line 302: we could probably better use the message display context for that? this might add a mini delay to the opening of context menu?

Simon-Laux (approved):

src/renderer/components/message/Message.tsx
Simon-Laux: but such things can also be in a follow up pr.

adzialocha: Yes! I was wondering about that, thanks for the pointer.

PR #3549, opened 2023-11-19
simon/i3473 fix autocrypt setup dialog

### LINKED ISSUE #3473: wrong autocrypt setup code does not show error
entering a bad autocrypt setup code does not show an error. instead, the spinning wheel is shown forever (in general, there is not much sense of the spinner; the result is instant)

Screenshot 2023-11-03 at 23 53 06

entering the correct autocrypt setup code works as expected.

Simon-Laux:
- autocrypt setup dialog handle & show error, remove spinner, refactor (unify components)
- remove blueprint js dependency from autocrypt setup enter key dialog
- code-cleanup

Bildschirmfoto 2023-11-19 um 01 08 14

now shows the error.

close #3473

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased][unreleased]
 
+### Fixed
+- fix: show error in autocrypt setup enter key dialog when the typed in key is invalid
+- fix bug: allow deleting first number in autocrypt enter setup key dialog
+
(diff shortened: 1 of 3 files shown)


r10s (approved):
tested a bad transfer as well as a good transfer - works as expected to me, nice!

cannot say much to the code, however, but lgtm :)

PR #3552, opened 2023-11-20
fix displaying sticker that has wrong mimetype

Simon-Laux:


adbenitez (approved):
thanks a lot!!!! :pray:

PR #3559, opened 2023-11-23
fix clear chat sometimes not refreshing the messagelist

Simon-Laux:


r10s (approved):
did not test - if it helps, lgtm :)

PR #3571, opened 2023-11-27
Use native watch mode and CSS modules of esbuild

adzialocha:
This PR introduces [CSS Modules](https://github.com/css-modules/css-modules) into our build pipeline. It's actually natively supported by `esbuild`, I just needed to update it to the latest version and add TypeScript support via a plugin. There's a bit of noise in this PR as I needed to bump the typescript and all plugins, so prettier got sad and I needed to autofix that, but most of the touched files can be ignored during review as they only changed formatting.

On top this uses a thin wrapper around the `css-loader` of `esbuild` and extend it with SCSS functionality, so we don't need to rewrite our .scss files to pure CSS.

To still get linter warnings I needed to add a `eslint` plugin and a `reporter` plugin which tells us about the current bundle process. On top when starting the `npm run watch` script, a parallel `tsc` process is launched in the background.

The idea behind this PR is to start using (S)CSS modules for new components and refactor old components to use them step by step. It will hopefully help us to structure the CSS better by avoiding global scope, also it will be easier to see _where_ the styles for each component live.


static/main.html
```diff
@@ -13,28 +20,21 @@
...
-
-      worker-src blob: is needed for mapbox afaik
-     -->
-    <link rel="stylesheet" href="./fonts/fonts.css">
-    <link rel="stylesheet" href="main.css" />
-    <link rel="stylesheet" href="fallback-theme.css" />
+    />
+    <link rel="stylesheet" href="./fonts/fonts.css" />
+    <link rel="stylesheet" href="./main.css" />
+    <link rel="stylesheet" href="./bundle.css" />
```
adzialocha, line 26: The `bundle.css` comes from the CSS modules via esbuild. We still keep `main.css` inside as we still have a lot of CSS from the `manifest.scss`

adzialocha: Eventually we will only have a `bundle.css` in the future

bin/build-frontend-ts.mjs
```diff
@@ -1,64 +1,129 @@
...
-    console.log(`- Executing "${command} ${args.join(' ')}"`)
-    const p = spawn(command, args, {
-      shell: true,
-      cwd: process.cwd(),
-      ...options,
-    })
-    p.stdout.pipe(process.stdout)
-    if (listener) {
-      p.stdout.on('data', listener)
+function startTypeChecker() {
```
adzialocha, line 16: Wondering if we can just spawn `tsc` in an npm script rather than doing it here

bin/build-frontend-ts.mjs
```diff
@@ -70,80 +105,79 @@ let wasmPlugin = {
...
-          .toString()
-          .indexOf('Found 0 errors. Watching for file changes.') !== -1
-      ) {
-        BundleIfNeeded()
-      }
-    }
-
-    watchTscProcess = run(
-      'npx',
-      'tsc -b src/renderer --pretty -w --preserveWatchOutput'.split(' '),
```
adzialocha, line 131: This is now running as an npm script via `watch:frontend:types`

src/main/tsconfig.json
```diff
@@ -9,7 +9,8 @@
     "allowJs": false,
     "esModuleInterop": true,
     "strictNullChecks": true,
-    "strict": true
+    "strict": true,
+    "plugins": [{ "name": "typescript-plugin-css-modules" }]
```
Simon-Laux, line 13: backend and shared don't need the css plugin, or does the LSP have a bug where it is needed in all modules/targets?

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## [Unreleased][unreleased]
 
+### Changed
+- Chore: Use native watch mode and CSS modules of esbuild #3571
```
Simon-Laux, line 6: maybe we should also mention the dependency updates?
I don't know as most of them are just dev tooling, but maybe we did it in the past atleast for the big ones as esbuild.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
...
+
```
Simon-Laux, line 7: also display eslint errors while building.

src/main/electron-context-menu.ts
```diff
@@ -35,12 +35,7 @@ const removeUnusedMenuItems = (
     | electron.MenuItem
 
   return menuTemplate
-    .filter(
-      menuItem =>
-        menuItem !== undefined &&
-        menuItem !== false &&
-        menuItem.visible !== false
-    )
+    .filter(menuItem => menuItem !== undefined && menuItem.visible !== false)
```
adzialocha, line 38: This currently breaks after this change. Apparently `menuItem` can sometimes be a `boolean`, even though this is not given in the TypeScript definition

Simon-Laux: this is a hack to make this syntax/style of menu item composing work:
```js
[
condition && MenuItem,
condition && MenuItem,
MenuItem,
]
```

package.json
```diff
@@ -117,30 +117,31 @@
     "@types/mapbox-gl": "^0.54.5",
     "@types/mime-types": "^2.1.0",
     "@types/mocha": "^9.0.0",
-    "@types/node": "^14.14.20",
-    "@types/node-fetch": "^2.5.7",
+    "@types/node": "^20.10.0",
```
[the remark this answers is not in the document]
adzialocha: Good question, it looks [complicated](https://stackoverflow.com/questions/42035263/relationship-between-the-version-of-node-js-and-the-version-of-types-node), but also apparently it's somewhat working for Node v18

src/main/electron-context-menu.ts
```diff
@@ -25,28 +25,52 @@ import { tx } from './load-translations'
...
 const webContents = (win: BrowserWindow) => win.webContents
 
 const removeUnusedMenuItems = (
-  menuTemplate: (electron.MenuItemConstructorOptions | electron.MenuItem)[]
-) => {
-  let notDeletedPreviousElement:
+  menuTemplate: (
     | electron.MenuItemConstructorOptions
     | electron.MenuItem
+    | boolean
```
Simon-Laux, line 38: might make sense to replace 
```js
   | electron.MenuItemConstructorOptions
   | electron.MenuItem
```
by
```js
| NotDeletedMenuItems
```

what do you think?

adzialocha: Think I found a better name and approach now!

src/main/electron-context-menu.ts
```diff
@@ -25,28 +25,19 @@ import { tx } from './load-translations'
 
 type todo = any
 
-type NotDeletedMenuItems =
-  | electron.MenuItemConstructorOptions
-  | electron.MenuItem
+type MenuItems = electron.MenuItemConstructorOptions | electron.MenuItem
```
Simon-Laux, line 28: name implies plural 🤔

Simon-Laux:
we might still need to make a 1.42.2 (to fix the 163 provider), so maybe we should delay merging it until the release.
Or we just YOLO it and care for fixing bugs fast as they arise and release fast updates, but generally it's good practice to test a bit before releasing.

adzialocha:
> we might still need to make a 1.42.2 (to fix the 163 provider), so maybe we should delay merging it until the release.
> Or we just YOLO it and care for fixing bugs fast as they arise and release fast updates, but generally it's good practice to test a bit before releasing.

Cool, let's keep it here or merge into a new development branch?

Simon-Laux:
a temporary development branch might be a good idea that we can then later merge into master and rebase to master.

Simon-Laux (approved):
LGTM, but maybe we should change base to master again and merge it directly to master. I'd say we are done with 1.42, if we need another patch we can open a branch from the release tag

PR #3572, opened 2023-11-27
Refactor styles for `SearchInput` component

adzialocha:
This is an example to demonstrate how we could approach refactoring styles based on CSS modules using the `SearchInput` component.

## Refactoring

- Introduce CSS module for `SearchInput`
- Removed unnecessary CSS properties
- Moved everything component-related into single file (no other dependencies)
- Remove all `!important` properties
- Remove blueprint styles

## Styling

- Reduced font size
- Added hover state for close button
- Fix minor centering issue
- Caret color is the same as text color

### Before

2023-11-29-105901_522x194_scrot
2023-11-29-105909_485x234_scrot

### After

2023-11-29-105702_420x252_scrot
2023-11-29-105818_570x241_scrot

Based on https://github.com/deltachat/deltachat-desktop/pull/3571


Simon-Laux:
image
I ❤️ the new hover effect

adzialocha:
Nice, I've removed the blueprint classes from the clear button and could also get rid of all ugly `!important` properties altogether ✨ 

Think its to early to introduce a new `IconButton` component, but we should do that as soon as this is used somewhere else again.

src/renderer/components/SearchInput/styles.module.css
```diff
@@ -0,0 +1,54 @@
+:root {
```
adzialocha, line 1: The style guide I've tried to follow here looks like that:

### General

- Prefer `padding-top`, `padding-bottom`, `padding-left etc. over `padding` to avoid bugs through order-critical dependencies between properties, except when you only need `padding` or `border` etc.
- Try to avoid `!important` as much as possible (goal is to get rid of them altogether at one point)

### Naming

- Class names and variable names are always camelCase
- Sub-classes can have short, local names (like `.active` or `.warning`)

### Ordering

- Properties are sorted alphabetically
- Order of groups: variables (`--borderRadius`), properties (`height: 10px;`), pseudo elements (`::before`), sub-classes (`.active`), always separated with a newline
- Classes (`.searchInput` etc.) are also always separated with a newline

### Variables

- Variables which are used across multiple classes in the same file are introduced at the beginning of the document, inside a `:root` group. Every variable name is prefixed with `--local-` (for example `--local-borderRadius: 10px;`)
- Global variables (used across files) are as-is (later we can think about prefixing them with `--theme-` but who knows ..)
- Variables only used inside of a single class can be named however, they are always defined at the beginning of the class and separated with a newline from the following properties

adzialocha: I don't know yet how theming comes into all of this, still learning about the theming system

r10s:
cannot say much to the refactorings, but the outcome is nice!

PR #3575, opened 2023-11-30
Refactor styles and components for all dialogues

adzialocha:
This is another large refactoring of _all_ dialogues in the application. It mostly touches component structure and styling, so no actual logic has been changed with this PR.

* New components for dialogs have been introduced, like `Dialog`, `DialogHeader`, `DialogBody`, `DialogFooter` etc.
* All places where custom CSS was injected into dialogues have been removed
* Slowly get rid of BlueprintJS by reducing the surface as much as possible, the only thing touching Blueprint is `Dialog` itself now
* Prefer explicit components with CSS modules rather than `style` hacks
* Nice new basic components like `Icon`, `Switch` or `Callout`

On top I gave dialogues a round border and started to change the styling of the settings view slightly:

2023-12-03-235209_818x1031_scrot
2023-12-03-235218_717x1310_scrot

I needed to touch a bit of styling of other dialogues as well, just to make them work again with the new components. If we also want to re-design or refactor them I suggest making separate PRs for them.


src/renderer/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,53 @@
...
+        className={classNames(styles.dialog, props.className, {
+          [styles.fixed]: props.fixed,
+        })}
+        style={{
+          width: width && `${width}px`,
+          height: height && `${height}px`,
+        }}
+      >
+        {children}
+      </BlueprintDialog>
```
adzialocha, line 48: This is the only place now where we use blueprint for dialogues. All of this will fit nicely together with the dialog logic refactor in https://github.com/deltachat/deltachat-desktop/pull/3512

src/renderer/components/Dialog/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+import DialogBody from './DialogBody'
+import DialogContent from './DialogContent'
+import DialogFooter from './DialogFooter'
+import DialogHeader from './DialogHeader'
+import DialogHeading from './DialogHeading'
+import DialogWithHeader from './DialogWithHeader'
+import EditButton from './EditButton'
+import FooterActionButton from './FooterActionButton'
+import FooterActions from './FooterActions'
+import OkCancelFooterAction from './OkCancelFooterAction'
```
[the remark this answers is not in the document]
adzialocha: Splitting up is also nice as the styles get even easier to read

src/renderer/components/Icon/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+    <span
+      className={classNames(styles.icon, className)}
+      style={{
+        WebkitMask: `url(../images/icons/${icon}.svg) no-repeat center`,
+        height: `${size}px`,
+        width: `${size}px`,
+      }}
+    />
+  )
+}
```
adzialocha, line 38: This is our new own `Icon` component, also slowly replacing blueprint

src/renderer/components/Icon/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+  | 'code-tags'
+  | 'cross'
+  | 'devices'
+  | 'favorite'
+  | 'forum'
+  | 'lead-pencil'
+  | 'list'
+  | 'open_in_new'
+  | 'person'
+  | 'swap_vert'
```
adzialocha, line 17: Some of these names could be better, but that's maybe part of another PR

src/renderer/components/LargeProfileImage/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+        <span
+          className={styles.largeProfileImageArea}
+          style={{ backgroundColor: color }}
+        >
+          {initials}
+        </span>
+      )}
+    </div>
+  )
+}
```
adzialocha, line 41: I think this could also probably be now used in other places as well, also not part of this PR though

src/renderer/components/LoginForm.tsx
```diff
@@ -470,55 +476,40 @@ export function ConfigureProgressDialog({
...
-              style={{ marginLeft: 'auto' }}
-            >
-              {tx('ok')}
-            </p>
-          </DeltaDialogFooter>
+          <DialogBody>
+            <DialogContent paddingTop>
+              <p>{error}</p>
+            </DialogContent>
+          </DialogBody>
```
adzialocha, line 505: This is a good example of how a simple dialog body looks like now

src/renderer/components/Settings/Advanced.tsx
```diff
@@ -0,0 +1,34 @@
...
+import ImapFolderHandling from './ImapFolderHandling'
+import ManageKeys from './ManageKeys'
+import SettingsHeading from './SettingsHeading'
+import SettingsSeparator from './SettingsSeparator'
+
+type Props = {
+  settingsStore: SettingsStoreState
+}
+
+export default function Advanced({ settingsStore }: Props) {
```
adzialocha, line 16: I renamed all settings-related components from `Settings-Advanced.tsx` to `Settings/Advanced.tsx`

src/renderer/components/Settings/styles.module.scss
```diff
@@ -0,0 +1,172 @@
...
+  align-items: center;
+  display: flex;
+}
+
+.profile {
+  display: flex;
+  margin-bottom: 20px;
+  margin-left: 25px;
+  margin-right: 30px;
+}
```
adzialocha, line 152: We probably want to move `Profile` into an own CSS module as well

src/renderer/components/Switch.tsx
```diff
@@ -0,0 +1,22 @@
...
+      label={props.label}
+      disabled={props.disabled}
+      onChange={event => {
+        props.onChange(event.currentTarget.checked)
+      }}
+      alignIndicator='right'
+      checked={props.checked}
+    />
+  )
+}
```
adzialocha, line 22: This is our new `Switch` component

src/renderer/components/dialogs/ForwardMessage/styles.module.scss
```diff
@@ -0,0 +1,4 @@
+// We need to force the dialog to full height to make `AutoSizer` work
+.forwardMessageDialog {
+  height: 100%;
+}
```
adzialocha, line 4: Every-time we use `AutoSizer` in a dialogue a hack is required, I wonder if this can be improved

src/renderer/components/Icon/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+}
+
+import styles from './styles.module.scss'
+
+export default function Icon({ size = 20, icon, className }: Props) {
+  return (
+    <span
+      className={classNames(styles.icon, className)}
+      style={{
+        WebkitMask: `url(../images/icons/${icon}.svg) no-repeat center`,
```
Simon-Laux, line 32: also should specify the standard `mask` css property. we should use both, maybe tauri will need webkit mask, but could also be already so old that it does not matter, I did not closely follow it.

Simon-Laux: you could also move `no-repeat` and `center` to the css file.

src/renderer/components/Dialog/DialogWithHeader.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+type Props = React.PropsWithChildren<
+  {
+    className?: string
+    fixed?: boolean
+    height?: number
+    onClickBack?: () => void
+    title: string
+    width?: number
+  } & Pick<DialogProps, 'isOpen' | 'onClose'>
```
Simon-Laux, line 16: isOpen will not longer be used with #3512

src/renderer/components/Dialog/Dialog.tsx
Simon-Laux: we should investigate using the `<dialog>` html tag instead.
Also I think we don't need the animations that blueprint comes with, but even if we do, I think they will be easy to replicate.

r10s:
for the item ordering, i tried to create a focused, top-level discussion at https://github.com/deltachat/interface/issues/57

Issue #3568, opened 2023-11-27
uncaught exception when closing full message view


Louis2747:
- Operating System EndeavourOS(linux) wayland with kde-unstable repo
- Delta Chat Version: Version 1.42.1 (git: v1.33.0-546-g80bf53e9) from chaotic-aur but also happens on regular aur 
- Expected behavior: full message view opens with visible text and closes without error
- Actual behavior: full message view opens with no contents and errors when closing
- Steps to reproduce the problem: clink the show full message button on a message and then close the message window
- Screenshots:
![Screenshot_20231127_124440](https://github.com/deltachat/deltachat-desktop/assets/52701949/b45c75de-c5d1-4078-9e90-da92aa4f9d73)

- Logs:
[2023-11-27-12-45-25.log](https://github.com/deltachat/deltachat-desktop/files/13473753/2023-11-27-12-45-25.log)

Simon-Laux:
Thanks for reporting. It does not appear on macOS nor Windows. What electron version is used and what aur packages did you try specifically?

Louis2747:
well the chaotic-aur one is [deltachat-desktop-git](https://aur.archlinux.org/packages/deltachat-desktop-git) which depends on electron25
i tried [deltachat-desktop](https://aur.archlinux.org/packages/deltachat-desktop) which depends on electron22
and [deltachat-desktop-bin](https://aur.archlinux.org/packages/deltachat-desktop-bin) which depends on electron26, but on that one it looked like this
![Screenshot_20231127_172402](https://github.com/deltachat/deltachat-desktop/assets/52701949/f0e35ada-4a5c-45ca-871c-ffdf9606c833)

Louis2747:
it appears to be fixed

PR #3577, opened 2023-12-01
try fixing the race in html view

Simon-Laux:
to adress #3568

I can not reproduce this on mac as it only happens on arch which seems to use wrong or broken versions of electron?
Anyways this change makes logical sense so it is fine to merge.


Simon-Laux:
so it does not fix #3568, but it fixes a race when starting the view that @Jikstra reported to me:
```
electron_html_email_view.html:1 Uncaught (in promise) Error: Error invoking remote method 'html-view:resize-content': No handler registered for 'html-view:resize-content'
```

adzialocha (approved):

PR #3579, opened 2023-12-02
log error when webxdc status update send fails

Simon-Laux:
Used to find this bug that is triggered when you send 2 status updates at the same time or closely together:
Bildschirmfoto 2023-12-02 um 15 32 35

I think it is fixed in the next core update, so I'm doing a pr to upgrade core next: #3580


Simon-Laux:
> Don't you log all the exceptions somewhere upwards? The error should be thrown all the way up even without this, is it just silently dropped there?

it is not shown correctly and not logged without this

PR #3584, opened 2023-12-04
Improve efficiency of retrieval of trusted domains

farooqkz:
This PR does a micro improvement in that area. The storing mechanism has changed but the retrieval mechanism is still compatible with older versions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased][unreleased]
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


Simon-Laux:
I'm not sure about this:
- Efficiency boost of removing json parse is not noticeable
- if we want ui in the future to manage or clear trusted domains it will become harder to retrieve all domains at once with this change.

farooqkz:
I agree. lemme close this for now.

PR #3587, opened 2023-12-09
Add video chat instances icons

farooqkz:
This PR adds these:

image


r10s:
does this do http-requests by just opening the dialog?

usually, we avoid http-requests in ui unless there is really a benefit thar cannot be achieved in other ways (as autoconfig - but even there we minimize requests by offline provider db)

that in mind, not sure how useful these icons are at all, however, maybe i also missed some discussions that happend before :)

Simon-Laux:
Yeah standard icons should be hardcoded, and icons for the custom provider should only be fetched while adding them, then stored to a `ui.` config setting.

> that in mind, not sure how useful these icons are at all, however, maybe i also missed some discussions that happend before :)

Basically an internal discussion where I told @farooqkz an old idea of mine that I already had since the last refactoring that @flipsimon did of this dialog.
It's does not really have priority, I would declare it as enhancement, basically the idea is to allow the user to add multiple custom video providers to the list, this would also pave the way to allow an "choose on sending" option where you could asked which one you like to use when you send an invitation. This would be not only nice to have, but a real improvement for people that use video chat services with multiple static rooms that they registered beforehand.
My idea here was to explore that idea on desktop first and I thought it would be a nice exercise for @farooqkz.

Also @farooqkz If you don't want to do everything in one branch we could make a new branch and merge this pr into the new branch instead. Why? because I want to merge complete things into master, especially for such optional enhancements they should be done by the time they are merged into the main/master branch. (to not create additional work)

farooqkz:
@hpk42 I would like to report Simon and bjoern for working on weekends. Please remember it's not my fault. They could review the thing later, Monday.

farooqkz:
I will mark this as draft for now. and wait for later to do things with higher priority.

farooqkz:
> Yeah standard icons should be hardcoded, and icons for the custom provider should only be fetched while adding them, then stored to a `ui.` config setting.


Can you explain a bit more? Don't you think the icons of the default providers are subject to change, too? Isn't it a good idea to store them the same way as well?

> Also @farooqkz If you don't want to do everything in one branch we could make a new branch and merge this pr into the new branch instead. Why? because I want to merge complete things into master, especially for such optional enhancements they should be done by the time they are merged into the main/master branch. (to not create additional work)

Yeah makes sense. I agree with that. I could merge it into another branch.

One more thing, don't you think using icons this way and leaving "Off" and "Custom" icon-less makes the UI ugly? How about using hardcoded icons for them as well? And use a question mark icon if there is no icon at all.

Simon-Laux:
> Can you explain a bit more? Don't you think the icons of the default providers are subject to change, too? Isn't it a good idea to store them the same way as well?

we can update them once in a while manually when doing a new release, I don't want to make network requests without the user doing anything related to trigger it. like when they add a provider it's ok IMO, but not then they just open the dialog.

If you want to make it easier you could make some script that can generate an array for the default providers into a file similar to the build info. I think it should just be an array anyway that gets set when you don't have it:

```
const defaultProviders = <import from static generated file that has the icons embedded as base64>
if (ui.videochatproviders does not exist) {
ui.videochatproviders = defaultProviders
}
```
So the default providers would essentially pre-fill the video chat provider list and thus could also be deleted by the user if they don't like to have them.

I would store everything as a json array (encoded to string) in one core ui config field for now, if there should be problems with the images for example we can think of a solution later (storing them in dedicated config fields, or even better build this feature into core and save them as blobs)

farooqkz:
I will open another PR.

PR #3600, opened 2023-12-20
Update developer guidelines

adzialocha:


docs/STYLES.md
```diff
@@ -4,32 +4,41 @@ We use `prettier` for CSS formatting. Use `npm run fix:format` before committing
...
+
+### Naming
+
+- Class names and variable names are always camelCase (`.searchInput`, `--borderRadius`, etc.)
+- The class name should represent the React component it relates to (for example `SearchInput` is represented with `.searchInput`)
+- Sub-classes are usually modifiers of the component and can have short, "local" names (like `.active` or `.warning`)
+
+### Ordering
+
+- Properties are sorted alphabetically
```
Simon-Laux, line 29: can we enforce this with linting?

Simon-Laux: but might not always makes sense like:
```
mask: <value>;
-webkit-mask: <value>;
```
but we could also say such polyfills for old property names should be inserted by the compilation process and not manually 🤔

adzialocha: > can we enforce this with linting?

Yeah, I've set this up for other projects as well, can dig it up.

Simon-Laux:
IMO we can merge this already.

adzialocha:
> IMO we can merge this already. we can always adjust the guidelines and refine them further later

Cool, I've meant to just link it better to the existing docs, will do it this week!

docs/STYLES.md
adzialocha: I've removed this for now, the linting tooling is very heavy

Issue #3485, opened 2023-11-06
Generic helper method for confirmation dialogues


adzialocha:
Refactor all confirmation dialoges as promises, using a  helper method like this:

```ts
const confirmed =  await confirmationDialog({message: "are you sure", confirmLabel: "Yes"});
if (confirmed){
 // ...
}
```

_Originally posted by @Simon-Laux in https://github.com/deltachat/deltachat-desktop/pull/3469#discussion_r1382039918_

Simon-Laux:
While at it we could also add that for the Alert dialog, but there you don't need to wait for the result as there is none, so that method would not be async.

adzialocha:
I think this can be closed now with the new refactoring

PR #3601, opened 2023-12-21
Helper method to easily use confirmation dialogues

adzialocha:
```typescript

const confirmed = await openConfirmationDialog({
  message: '...',
})

if (confirmed) {
  // Do something
}
```

Closes: https://github.com/deltachat/deltachat-desktop/issues/3485


Simon-Laux (approved):
LGTM

PR #3602, opened 2023-12-21
Use new `useConfirmationDialog` hook

adzialocha:
Use the new `useConfirmationDialog` hook everywhere possible, as introduced in: https://github.com/deltachat/deltachat-desktop/pull/3601

Please note that I couldn't use it everywhere as not all functions are React components (hooks are not allowed outside of them). I think there's some neat React patterns we can use to refactor these as well, but this is not part of this PR.


Simon-Laux (approved):
code LGTM

PR #3603, opened 2023-12-21
Regression: Add missing openDialog and type declaration for contact chat list items

Simon-Laux:
the types are unfortuantely not checked at the point of the data of virtual lists. so the typescript compiler did not catch this.

changelog entry not needed, because the introduced bug is not released yet.


adzialocha (changes requested):
I suggest to declare the type of the `listData` argument in `ChatListPart` (currently `any`) to prevent further similar bugs ✊🏻

adzialocha:
Since all chat list items seem to be React components we could also use the `useDialog` in all of them to prevent this prop drilling

adzialocha:
Since all chat list items seem to be React components we could also use the `useDialog` hook in all of them to prevent  prop drilling

adzialocha (approved):
I've had a go on the improvements, fine to merge now for me.

Issue #2892, opened 2022-09-21
can´t delete several pics at once


g6094199:
- Operating System (Linux/Mac/Windows/iOS/Android): Ubuntu Budgie 22.04
- Delta Chat Version: 1.32
- Expected behavior: possibility to pick several pics to take action on them, e.g. delete them at once
- Actual behavior: e.g .you can only delete one pic at a time, which is annoying when having a bunch of pics to delete
- Steps to reproduce the problem: try to pick more than one picture
- Screenshots:
![Bildschirmfoto vom 2022-09-21 11-33-06](https://user-images.githubusercontent.com/37833326/191470766-7b1b9d6c-6587-4e96-8f05-72c29d0f5d46.png)

Jikstra:
Yes multi select would be an important feature :)

Simon-Laux:
@JackJeisly suggested supporting both Ctrl+Click and Shift+Click, I assume the latter is for selecting ranges and the former for manual multi selection

r10s:
> suggested supporting both Ctrl+Click and Shift+Click, I assume the latter is for selecting ranges and the former for manual multi selection

yes. this is a 30+ years old convention, you can check that eg. in windows explorer, macos finder and probably also most linux file managers. i miss that every day in delta chat desktop, for both, the chat-list and the message-list :)

g6094199:
additionally for the desktop app: mark a message as mentioned above and delete it with the "del" (entf) key should also be possible.

RasheedAZ:
Not sure if I should open a separate issue, but I'm having similar troubles deleting multiple chats. The only way possible appears to be right click and select "delete chat".

I have hundreds of chats that I need to delete :(

Already tried shift/ctr + select and also deleting with the delete key... none of them worked.

farooqkz:
> Not sure if I should open a separate issue, but I'm having similar troubles deleting multiple chats. The only way possible appears to be right click and select "delete chat".
> 
> I have hundreds of chats that I need to delete :(
> 
> Already tried shift/ctrl + select and also deleting with the delete key... none of them worked.

You should open another issue.

WofWca:
> supporting both Ctrl+Click and Shift+Click

For reference, check out ARIA guide, ~~[the Grid pattern](https://www.w3.org/WAI/ARIA/apg/patterns/grid/#gridNav_inside)~~ [the Listbox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/):
[...]

PR #3606, opened 2023-12-28
Select multiple messages in chat for forwarding & deletion

farooqkz:
This PR aims to implement selecting multiple messages for forwarding and deleting. Reviews are welcome after christmas :)

> i know, this is a draft, bit some comments on testing:
> 
>     * the "forward/delete" bar needs a "cancel" or "X" button (i think, an "X" to the very right would be pretty straight forward)
> 
>     * it is not possible to select images as this enlages images - every tap should just select in select mode. at best, we should also handle taps in the backgound left/right of the bubbles (esp. if we hilite messages using a background color)
> 
>     * if not too hard, i'd prefer to have the "forward/delete bar" at the bottom, replacing the input bar during select mode
> 
> 
> otherwise, this is a really BIG enhancement - in comparable few lines of code :) great!

 - We've got a cancel button now
 - Fixed
 - We should get votes for that

image

this would solve one part of #2892 (that issue is about multi select in chatlist, message list and gallery)


farooqkz:
Tada! Ready for review

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 - Prefer light theme for the help and webxdc loading pages
 - Helper method to easily use confirmation dialogs #3601
 - Refactor using new `useConfirmationDialog` hook #3602
+- Add multi selecting messages for forwarding and deleting
```
adzialocha, line 20: CHANGELOG entry is missing a PR number

scss/message/_select-mode-mask.scss
```diff
@@ -0,0 +1,12 @@
+.select-mode-mask {
```
adzialocha, line 1: Is this a "mask" (https://css-tricks.com/clipping-masking-css/) or rather an "overlay"?

scss/manifest.scss
```diff
@@ -64,6 +64,8 @@
 @import 'message/_status-icon';
 @import 'message/_message-calls';
 @import 'message/_message-markdown';
+@import 'message/_selected-messages-action.scss';
+@import 'message/_select-mode-mask.scss';
```
adzialocha, line 68: It would be nice to use CSS modules here instead if possible

scss/message/_selected-messages-action.scss
```diff
@@ -0,0 +1,16 @@
...
+  align-items: center;
+  & > p {
+    width: 100%;
+    text-align: right;
+    margin: 8px;
+  }
+  & > button {
+    margin: 4px;
+    margin-bottom: 6px !important;
+  }
```
adzialocha, line 15: CSS patterns like overriding nested components are tricky and hard to read, the `!important` is indicating that you're overriding some unknown parent style which was defined somewhere else.

If we would use CSS modules and clear custom classes for each version of a button the CSS code would be much easier to read where what is defined & to maintain.

I'm not sure how easy it is to refactor the CSS into modules, but if you can see a way to avoid writing CSS like that with little changes I'd appreciate, otherwise we can look at a proper refactoring later, that's also okay as I don't want to derail your PR 😅

src/renderer/components/dialogs/ForwardMessage/index.tsx
```diff
@@ -22,14 +22,15 @@ import styles from './styles.module.scss'
 const LIST_FLAGS = C.DC_GCL_FOR_FORWARDING | C.DC_GCL_NO_SPECIALS
 
 export default function ForwardMessage(props: {
-  message: T.Message
+  message: T.Message | number[]
```
adzialocha, line 25: I think the API got a little more confusing now as `message` (singular) can now be a single message object or a list of message ids (plural).

How about:

```t
messages: T.Message[] | number[]
```

Ideally even giving `number[]` a type so it gets clearer what the number represents. Maybe we even have that type somewhere already.

```ts
type MessageId = number

// ...

messages: T.Message[] | MessageId[]
```

Or even better, just remove the object support and only accept IDs from now on:

```ts
messageIds: number[]
```

src/renderer/components/message/MessageList.tsx
```diff
@@ -92,13 +92,21 @@ function useUnreadCount(
   return freshMessageCounter
 }
 
+type MessageListProps = {
+  chatStore: ChatStoreStateWithChatSet
+  refComposer: todo
+  selectMessage: (id: number) => void
+  unselectMessage: (id: number) => void
+  selectedMessages: number[]
```
adzialocha, line 100: `MessageList` and `Message` get quite many new props here which I wouldn't see _directly_ as part of the message component itself.

Maybe we can introduce a context here instead to manage the selection state for messages? The context lives in the currently displayed chat and every message can consume the state of the context to understand if it is selected or not.

I've seen you already have a nice reducer. With the context representing the state and the reducer changing it we can have a nice isolated logic. All of this can be moved into own files.

src/renderer/components/message/MessageListAndComposer.tsx
```diff
@@ -83,6 +84,41 @@ export default function MessageListAndComposer({
...
+        case 'select':
+          return [...selectedMessages, action.messageId]
+        case 'unselect':
+          return selectedMessages.filter(id => id !== action.messageId)
+        case 'reset':
+          return []
+      }
+    },
+    []
+  )
```
adzialocha, line 105: Nice!

adzialocha (changes requested):
I can see some nice patterns here with dedicated components and reducer logic for the selection logic.

As written above I'd suggest to factor out the state as well into an own context, to avoid prop drilling and making `Message` and `MessageList` more complex.

You already have some code isolation in place and we could go even further. I'd suggest to create a new folder and move a context implementation, your new components, the reducer and maybe even the CSS module there to make it all represent the "multi select" feature, completly independent from the rest.

adzialocha:
Let me know if you need any help with my suggested changes! Also I've changed the target branch to the button refactor one as this branch includes these changes as well, hope that's okay.

scss/message/_select-mode-mask.scss
farooqkz: Sorrt it's an "overlay". Thanks.

scss/message/_selected-messages-action.scss
farooqkz: yeah a proper refactoring is needed for the buttons component.

src/renderer/components/dialogs/ForwardMessage/index.tsx
farooqkz: I agree. But this is for another PR IMO.

Simon-Laux (changes requested):
does not work for me or I don't understand how to do it, lets put it into draft state again until it is mergeable.

farooqkz:
> does not work for me or I don't understand how to do it, lets put it into draft state again until it is mergeable.

I've began to use Contexts thanks to adz. You can omit my last commit to see the working version with props drilling.

farooqkz:
@adz, I have difficulty implementing your suggestion. Some hints?

farooqkz:
> ~@adz, I have difficulty implementing your suggestion. Some hints?~
> 
> On second thought, please give me some more days.

Okay I'm stuck, man!

adz:
Different adz I think. Interesting project though.

adzialocha:
This is a really nice feature and I'd love to see it getting released, maybe in `v1.46`. @farooqkz we can look at it together whenever you are ready in the next weeks, I'm happy to make some direct code suggestions as well if that's fine with you

hpk42:
Discussed with nico and farooq -- putting this into project resurrection for when the time comes!

PR #3608, opened 2023-12-31
Introduce own React context for context menus & fix regression

adzialocha:
This fixes a bug which was introduced during the Dialog refactorings causing Dialogs to not be aware of the `ScreenContext`, not allowing them to access the `openContextMenu` method.

This fix detaches the context menu logic from the screen context as moving the screen context to a higher level (to make dialogs aware of it) would have been a much more complicated undertaking. At the same time this refactoring was planned anyhow, it makes sense to decouple the logic a little bit.


Simon-Laux (changes requested):
Code looks good, but I get this error spammed to console:
Bildschirmfoto 2023-12-31 um 20 19 40

adzialocha:
> Code looks good, but I get this error spammed to console:
> Bildschirmfoto 2023-12-31 um 20 19 40
> 

Yeah, this is not done yet, just started to move the parts. I'll request review if it's ready 😎

Simon-Laux (approved):
LGTM, though maybe we should merge after #3348? or do you want to rebase #3348 onto this?

adzialocha:
> LGTM, though maybe we should merge after #3348? or do you want to rebase #3348 onto this?

I'll rebase #:+1:

PR #3615, opened 2024-01-01
Remember file paths from file selector dialogs

### LINKED ISSUE #3596: not all folders of save/load dialogs are rememberd
the folder for "Save As"  is remembered, which allows to eg. easily export several images from a chat to some folder.

however, the folder for "Attach Image" or "Attach File" is _not_ remembered, making it hard to attach eg. several images to a chat.

we should remember most folders in general, however, still using a good default - and taking care that remembered folders may be not available later (in this case, again, use the default)

Simon-Laux:
- remember last open dialog locations
- fix lint
- do not persist last save location across sessions anymore

fixes #3596


CHANGELOG.md
```diff
@@ -18,6 +18,8 @@
 - Helper method to easily use confirmation dialogs #3601
 - Refactor using new `useConfirmationDialog` hook #3602
 - update `electron` from `v26.6.0` to version `v26.6.3`
+- remember file open dialog locations across the current session
+- do not persist last save location across sessions anymore
```
adzialocha, line 22: PR number is missing, maybe you also group the changes?

src/renderer/utils/cachedLastUsedPath.ts
```diff
@@ -0,0 +1,14 @@
+export function cachedLastUsedPath(
```
adzialocha, line 1: Is this really "caching"? You're _persisting_ the last used directory in a storage. I'd call this method `storeLastUsedPath` or similar.

src/renderer/components/Settings/ManageKeys.tsx
```diff
@@ -8,15 +8,21 @@ import useTranslationFunction from '../../hooks/useTranslationFunction'
...
+import { cachedLastUsedPath } from '../../utils/cachedLastUsedPath'
+import { dirname } from 'path'
 
 export default function ManageKeys() {
   const tx = useTranslationFunction()
   const openConfirmationDialog = useConfirmationDialog()
 
   const onKeysImport = useCallback(async () => {
+    const { defaultPath, setLastPath } = cachedLastUsedPath(
+      'last_directory:keys_import',
```
adzialocha, line 20: Could be nice to have the keys part centrally defined somewhere as constants, then there's a central place where you can see what keys exist in the store, maybe even directly as a type?

How about moving it into the same file, right next to where you define `cachedLastUsedPath`?

src/shared/shared-types.d.ts
```diff
@@ -52,6 +52,7 @@ export interface DesktopSettingsType {
   activeTheme: string
   minimizeToTray: boolean
   syncAllAccounts: boolean
+  /** @deprecated is now not persisted between settings anymore */
```
adzialocha, line 55: Will this variable be removed in the future?

Maybe this doc-string can be improved, how about "The last used file location for the save dialog is now only kept in memory and not persisted anymore."

On top, if you're planning to remove it, maybe also state why.

adzialocha (changes requested):
Code looking good, but I have some suggestions around documentation & naming.

src/renderer/utils/cachedLastUsedPath.ts
Simon-Laux: it's only stored for the current session though

src/renderer/components/Settings/ManageKeys.tsx
Simon-Laux: thought about it, but thought it wouldn't make a big difference, because it is only in memory anyway, but I agree it removes potential of messing up, I can make a type or constant enum for it.

adzialocha (approved):

PR #3616, opened 2024-01-01
message-parse summary text in chatlist item summary

### LINKED ISSUE #3476: ChatlistItem summary is not message-parsed
Should be also run thought the message parser.

Why:
- for labeled links (`[here](https://delta.chat)`) so they don't take up extra space in the summary
- also useful maybe for the experimental markdown like bold text

How:
- no interaction
- no linebreaks

Both of the "how" requirements could be reached by reusing message body parsing (used currently for messages and for contact status/signatures in ViewProfile) and disabling pointer events and line breaks via CSS. But the logic could also be cloned if we want to render the elements differently.

@adbenitez drew my attention to this.

Simon-Laux:
also simplify MessageBody component

closes #3476

preview:

| markdown disabled (default) | markdown enabled |
|------------------------------| -------------------|
|Bildschirmfoto 2024-01-01 um 06 21 58 | image|

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [Unreleased][unreleased]
 
 ### Added
-- message-parse summary text in chatlist item summary #3476
+- Introduce text formatting for chatlist item summary (use message-parser) #3476
 
 ### Changed
 - update deltachat-node and deltachat/jsonrpc-client to `v1.132.1`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased][unreleased]
 
+### Added
+- message-parse summary text in chatlist item summary #3476
+
 ### Changed
 - update deltachat-node and deltachat/jsonrpc-client to `v1.132.1`
 - Use native watch mode and CSS modules of esbuild #3571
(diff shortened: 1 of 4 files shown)


src/renderer/components/message/MessageMarkdown.tsx
```diff
@@ -103,10 +103,59 @@ function renderElement(elm: ParsedElement, key?: number): JSX.Element {
   }
 }
 
-export function message2React(message: string): JSX.Element {
+/** render in preview mode for ChatListItem summary, not interactive but looks rendered */
```
adzialocha, line 106: I'm not sure if I understand this comment, maybe the word order is off? How about something like: "Render formatted summary text of chat list item". The "interactivity" part I don't know if I understand. Maybe you want to say: "Links can not be clicked." etc.?

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## [Unreleased][unreleased]
 
+### Added
+- message-parse summary text in chatlist item summary #3476
```
adzialocha, line 6: The CHANGELOG entry / PR title can maybe be improved as I'm not sure if I understand it. How about "Introduce text formatting for chatlist item summary" or similar?

adzialocha (changes requested):
Code looks fine, just some minor changes requested for documentation

adzialocha (approved):

PR #3617, opened 2024-01-01
Disable three-dot-menu when not applicable (map, other gallery tabs)

### LINKED ISSUE #3523: "Square grid" and "Aspect Ratio Grid" must only be shown if there is a grid
these gallery options are hard to get anyways and should only be available if there is a direct, visible effect on changes.

otherwise, the user has no idea what happens - and may later notice that the images are not rendered as usual and has no idea how to switch back - bad ux all in all.

this means, the options should only be available in "Images" and "Video" and must be hidden in "Audio", "Files", "App"

Simon-Laux:
closes #3523

https://github.com/deltachat/deltachat-desktop/assets/18725968/e16b2616-95fe-4c74-8cbd-fe4646b68e51

### DIFF
--- a/src/renderer/components/screens/MainScreen.tsx
+++ b/src/renderer/components/screens/MainScreen.tsx
@@ -176,11 +176,17 @@ export default function MainScreen() {
             galleryRef.current?.state.currentTab || ''
           ))
     )
-  }, [selectedChat, selectedChat?.activeView, alternativeView, galleryRef])
+  }, [selectedChat, alternativeView, galleryRef])
 
   useEffect(() => {
     updateThreeDotMenuHidded()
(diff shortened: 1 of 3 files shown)


src/renderer/components/screens/MainScreen.tsx
```diff
@@ -163,6 +163,30 @@ export default function MainScreen() {
       ? 'gallery'
       : 'chat'
   )
+  const galleryRef = useRef<Gallery | null>(null)
+
+  const [threeDotMenuHidded, setThreeDotMenuHidded] = useState(false)
```
adzialocha, line 168: Typo: "hidden"

adzialocha (approved):
Looks good! Just that typo, otherwise it's ready to be merged.

PR #3621, opened 2024-01-06
Account Sidebar

### LINKED ISSUE #2331: Switching Accounts too cumbersome
- Delta Chat Version: 1.20.3

I have multiple E-Mail addresses and want to use multiple of them at a time.

I can't run multiple instances of Delta Chat at the same time:

```
Only one instance allowed. Quitting.
```

I know, you can switch between email addresses using the tab at the top right, but that's slow.
I think, there are multiple reasons for that:

* three clicks required to switch selected address
* small animations after clicking the button, which take just about 0.2 seconds, but slows the process down anyway
* first two clicks on the top right, third click in the center of the app
* contacts on the left, while the button to switch is on the right

Instead of fixing these problems, I'd prefer to see all contacts of all my email addresses at once, similar to how common email clients work.
The contacts could be sorted by address and have small headers inbetween, which shows the email address, they belong to.
Maybe you would still have a currently active address, so adding new contacts and groups etc. applies to a single address only.

Or the simplest quickfix: Just allow multiple instances of DeltaChat

### LINKED ISSUE #3646: Open dialogs are hard to notice
- Operating System (Linux/Mac/Windows/iOS/Android): Desktop
- Delta Chat Version: v1.42
- Expected behavior: The open dialog should be easily noticeable
- Actual behavior: It is a bit hard to notice it.
- Steps to reproduce the problem: Try to open some link with DC, it'll pop up from background with a dialog asking if you want to Ok or cancel. Since the dialog has the same colour as the rest of DC, it'll be hard to notice it.
- Screenshots: TBD
- Logs: N/A

Simon-Laux:
This pr will add a sidebar for multi account usage (like discord has for it's guilds/servers and element has for it's spaces, for us it's email accounts).
Later it will also add multi account notifications, but currently it is still in the draft stage.

possible inspirations:
- zulip (I like the shortcut)
- discord
- element

Issues solved
- #3646
- #2331
- #3668
- #3656

closes #3646, closes #2331

### Changes

Added:
- Add Account Switcher Sidebar
- Add new account deletion screen

changed:
-  badge count is now across all active accounts
- "sync all accounts" moved to settings under advanced -> experimental
- webxdc stay open when switching accounts (when you have sync all enabled, which is the default)

Removed:
- Remove old Account switcher screen
- Remove Sidebar Menu

### UI

- [x] indication of selected account
- [x] find better solution for scrollbar:
	- maybe using sth like https://github.com/Grsmto/simplebar
	- completely hiding it like discord does there? -> we don't need it (simon and bjoern say)
- [x] better new accounts button design
- [x] add unread badge so people can already use the branch
- [x] hover tooltip like discord (account name, email, maybe size and id there, though the id should be de-emphasise and the size will be in settings once we have some size cleanup settings like telegram)
	- for disabled accounts, it could show "[icon] disconnected background sync", maybe with an info button that leads to a dialog explaining it (later move the text to help) -> currently no, because no links possible/allowed in a tooltip
	- for other accounts it could show connection state in the over overlay?
	- also should show if chat is muted?

### Big Picture / place of it in the app UX

- [x] Remove option to hide accounts sidebar
- [x] Remove old account switcher
	- [x] implement delete account somewhere, but needs some way to make the confirmation accident-proof:
		- -> Right click delete account -> new screen with new background (like iceberg but hell / fire) -> maybe double confirmation needed
	- [x] Show hover information somewhere (hover info, account size and id)
		- maybe in about screen?
- [x] Screen for no selected account -> needed after deletion to replace the old account switcher that would appear
	- [x] on first start create a new account directly and thus the welcome screen is shown, ensure that this still works
- [x] Badge counter of the app (in macOS dock, taskbar on other platforms)
	- what number should it show: sum of not muted accounts messages!
- [x] multiaccount webxdc / keep webxdc open when switching accounts
	- ~~also mention the account in title bar~~ - we agreed that the existing chatname is enough context for now
- [x] refactor to component (css not in global file where possible, own folder for the component/s)
- [x] account sidebar no theming -> neutral grey with white -> always dark mode
	- will allow per account theming later
- [x] remove border radius hover -> keep radius of hovered state
	- try inverting add button instead of changing radius?
- [x] sync all -> move to advanced settings
- [x] right click -> all media, settings, remove account
	- also a button for "mark as read" when it has unread messages -> call mark notice chat on chats from all fresh messages
- [x] write entries in changelog of the things that changed (webxdc stay open across accounts, sync all moved to settings, introduce sidebar, badge count is now across all active accounts and so on.)
### Bugs or possible sources of them

- [x] make sure `SettingsStoreInstance.effect.load()` is reloaded when switching accounts
- [x] going back on new account screen does delete the account, but not update the account list in the sidebar
- [x] chat from last account is still selected while switching
	- -> maybe use the react unique key hack to force complete re-rendering?
- [x] Fullscreen media view is under account list, also stays open when switching account
- [x] Fix chat menu (3dot) (positioning)
- [x] fix gallery (increases hight of whole app content, interface looks bugged out)
- [x] fix map (increases hight of whole app content, interface looks bugged out)
- [x] clear button of search is at the wrong place (positioning)
- [x] `--als-active-indicator-color` is not themed -> active indicator invisible in light mode
- [x] put all dialogs in the foreground, when clicking the account bar the click should be on the backdrop and thus close the dialog
- [x] account badges (both unread and disabled sync) are over the dialog backdrops, which allows you to change account while you have the settings dialog open which results in a UI/react crash
- [x] changed display name or avatar is not auto updated when user changes it. - also happens on first login sometimes
- [x] welcome screen and delete account screens both use blueprint dialog and that has an invisible backdrop that blocks you from from using the sidebar -> making backdrop more visible fixed it
- [x] cancel in login screen of welcome screen does not work -> should go back to previous account in session
- [x] when no account is selected the settings store is empty and thus the background sync disabled sign is not shown and the list is buggy - access the variable differently and do some workaround, unless I want to do it in core directly
- [x] login with email credentials does not update the screen and results in a buggy state
- [x] check that deletion is not buggy anymore
- [x] change sync all accounts translation string: "Sync All" is not enough text in the new context
### optional polish:
- [ ] polish: indication of hovered account: maybe set via js instead so it is immediately updated when starting to scroll?
- [ ] on open: scroll currently active account into view

## last todo before merging
- [x] settings shortcut
- [X] rebase and remove menu sidebar
- [x] #3674
- [x] fix connectivity in tooltip update after unmounted (react error)
- [x] #3677

## More
- Anything besides this todo (excluding new found bugs of course) is now moved to #3671


Simon-Laux:
thanks for fixing the sidebar bug, but I have changes in my working copy so would have been better to do this on a separate branch and making a pr into this branch.

r10s:
first of all, this is a great PR with really huge UX impact! thanks a lot for pushing that forward!

i know, this is still "WIP", but one comment on the "Settings" availability: the dialog is now reachable in the account's context menu as well as in the main menu, which is great.

however, for both, ease and discoverability, it is better if it also stay available in the primary UI, so by just "left clicking somewhere".

one obvious idea we've seen in similar apps, is to add the icon left-bottom to the sidebar:



nb: mac-accountbar looks great ❤️

adzialocha:
I'm receiving an "Failed loading chat Chat#22 from database\n\nCaused by:\n    Query returned no rows" error when switching accounts sometimes (roughly every 2nd time), it seems to be a race condition but I can't pin down yet where it comes from.

2024-02-09-112055_1253x1396_scrot

CHANGELOG.md
```diff
@@ -10,7 +10,12 @@
 - Add shortcut to scan qr code to "New Chat" dialog #3623
 - Message reactions with emojis #2964
 - Show info dialog on tap for invalid unencrypted mails #3652
+<<<<<<< HEAD
 - Add QR scan button next to search input on main screen #3665
+=======
+- Add Account Switcher Sidebar #3621
+- Add new account deletion screen #3621
+>>>>>>> 833c4f3c (update changelog)
```
adzialocha, line 18: 🐱

docs/THEMES.md
```diff
@@ -5,6 +5,10 @@ Themes are CSS files that act as themes by containing many different css variabl
 Most of these variables are small variations of the base colors of the theme so using scss allows us to generate them from the base colors to save us work.
 But of course we can overwrite those variables manually too, we can even overwrite css classes if our theme needs it.
 
+## Notes
+
+- The account sidebar should not be themed
```
adzialocha, line 10: In "light" mode the sidebar becomes a bit disturbing with such a hard contrast, I don't think this needs to be discussed now, but are we really set on never changing this ever? 😄

scss/_global.scss
```diff
@@ -206,3 +206,31 @@ kbd,
...
 }
+
+.main-container {
+  display: flex;
+  position: absolute;
+  width: 100%;
+  height: 100%;
+  overflow: hidden;
+
+  // todo move these ones into their files
```
adzialocha, line 217: Let's do it as part of this PR!

src/renderer/components/screens/AccountDeletionScreen/AccountDeletionScreen.tsx
```diff
@@ -0,0 +1,117 @@
...
+  FooterActions,
+} from '../../Dialog'
+
+import styles from './styles.module.scss'
+import { T } from '@deltachat/jsonrpc-client'
+import { BackendRemote } from '../../../backend-com'
+import { Screens } from '../../../ScreenController'
+import { avatarInitial } from '../../Avatar'
+
+export function AccountDeletionScreen({
```
adzialocha, line 19: You can export them as default if it's the only component

src/renderer/components/screens/AccountDeletionScreen/AccountDeletionScreen.tsx
```diff
@@ -0,0 +1,117 @@
+import React, { useEffect, useState } from 'react'
+import useTranslationFunction from '../../../hooks/useTranslationFunction'
```
adzialocha, line 2: Mini optional nitpick: I'm trying to keep some sort of grouping of the imports to distinct better where they come from, separated by newlines: external, internal, styles, types

src/renderer/components/screens/AccountDeletionScreen/AccountDeletionScreen.tsx
```diff
@@ -0,0 +1,117 @@
...
+          fixed={true}
+          onClose={() => {}}
+          width={400}
+        >
+          <DialogHeader
+            onClickBack={accountInfo ? onCancel : undefined}
+            title={tx('delete_account')}
+          />
+          <DialogBody>
+            <DialogContent paddingTop={false}>
```
adzialocha, line 59: `paddingTop` is always false by default

src/renderer/components/screens/AccountDeletionScreen/styles.module.scss
```diff
@@ -0,0 +1,60 @@
+.AccountDeletionScreen {
```
adzialocha, line 1: Let's keep all class names in lowerCamelCase (https://github.com/deltachat/deltachat-desktop/blob/master/docs/STYLES.md#naming), this needs to be adjusted for all other classes in this PR as well

src/renderer/components/screens/AccountDeletionScreen/styles.module.scss
```diff
@@ -0,0 +1,60 @@
...
+  flex-grow: 1;
+  .Background {
+    background-size: cover;
+    // TODO find a better fire/lava image
```
adzialocha, line 5: Let's find the right picture as part of this PR, otherwise it'll stay forever hehe

src/renderer/components/screens/AccountDeletionScreen/styles.module.scss
```diff
@@ -0,0 +1,60 @@
...
+    background-image: url('../images/account_deletion_background.jpg');
+    height: 100vh;
+    margin-top: 0px;
+  }
+}
+
+.avatar {
```
adzialocha, line 12: Think this needs a `cursor: pointer`

src/renderer/components/screens/AccountListSidebar/styles.module.scss
```diff
@@ -0,0 +1,203 @@
...
+    object-fit: cover;
+    color: var(--avatarLabelColor);
+    text-align: center;
+    font-size: var(--local-avatar-font-size);
+  }
+
+  img.content {
+    object-fit: cover;
+  }
+}
```
adzialocha, line 37: `.avatar` is not used by `AccountListSidebar` and seems to be a duplicate as well

src/renderer/components/screens/AccountListSidebar/styles.module.scss
```diff
@@ -0,0 +1,203 @@
...
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+
+  --als-avatar-size: 42px;
+  --als-avatar-margin: 4px;
+  --als-active-indicator-color: white;
+  padding-top: 5px;
+
+  .Account {
```
adzialocha, line 57: We should avoid inherited styles like this, especially when the affected elements are not even part of the component. From here I can't tell how and where `.Account` or even deeper `.avatar` is affected.

Patterns to avoid this are to introduce variants of the sub components you want to style and pass this over as a property (for example `<MyComponent secondary />` etc.), if the styling is too far away, then consider creating a new component. Another, more hacky, but still okay approach would be to give it a `className` and pass over your customized css over like that, though I'd avoid that if it is changing the component a lot

src/renderer/components/screens/NoAccountSelectedScreen/styles.module.scss
```diff
@@ -0,0 +1,24 @@
+.NoAccountSelectedScreen {
+  flex-grow: 1;
+  .Background {
```
adzialocha, line 3: `.Background` is not a modifier of `NoAccountSelectedScreen` but its own component, you can declare the style on root level, no need to nest it here

src/renderer/components/screens/AccountListSidebar/AccountListSidebar.tsx
```diff
@@ -0,0 +1,119 @@
...
+      }
+    }
+  }, [accountForHoverInfo])
+
+  useEffect(() => {
+    updateHoverInfoPosition()
+  }, [accountForHoverInfo, updateHoverInfoPosition])
+
+  useEffect(() => {
+    window.__updateAccountListSidebar = debounce(() => {
```
adzialocha, line 85: We should try to avoid using `window`, it creates global state which gets hard to understand and to debug. For this I'd use contexts where it is clear where the state lives and who it consumes.

src/renderer/components/screens/AccountListSidebar/AccountListSidebar.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+  const refresh = useMemo(
+    () => async () => {
+      const accounts = await BackendRemote.rpc.getAllAccounts()
+      setAccounts(accounts)
+      const desktopSettings = await runtime.getDesktopSettings()
+      setSyncAllAccounts(desktopSettings.syncAllAccounts)
+    },
+    []
+  )
```
adzialocha, line 42: This all smells a little bit like logic which might live well in a context rather than in the view component itself ("state / logic required to manage the account sidebar")

adzialocha (changes requested):
Really nice work! 🥇 

Seems all to work well, except of that slightly scary race condition I've written about in another comment, not sure yet what is going on, it doesn't seem to break anything as far as I can tell.

Made a couple of comments on code structure as there's some patterns   I'd like to avoid for new code as it is hard to read and to debug (CSS inheritance, globally defined state and logic), I can help with restructuring if needed.

Lastly there's some mini nitpicks around how we style CSS documents. Hopefully the new [style guide](https://github.com/deltachat/deltachat-desktop/blob/master/docs/STYLES.md) "guides" you there 😸 It's not stopping us from merging ofc, but nice to stick to consistency if we can.

adzialocha:
> one obvious idea we've seen in similar apps, is to add the icon left-bottom to the sidebar:

💯

adzialocha:
Another bug seems to be the `Connectivity` component not cancelling some async check when it got unmounted, should be easy to fix:

2024-02-09-120530_1257x210_scrot

src/renderer/components/screens/AccountListSidebar/AccountListSidebar.tsx
```diff
@@ -0,0 +1,119 @@
...
+import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react'
+import { BackendRemote } from '../../../backend-com'
+import { AccountItem } from './AccountItem'
+
+import styles from './styles.module.scss'
+import { AccountHoverInfo } from './AccountHoverInfo'
+import { debounce } from 'debounce'
+import { runtime } from '../../../runtime'
+
+export function AccountListSidebar({
```
adzialocha, line 11: I'd be up for moving all new components just into the `components` folder (no `screens` subfolder, etc.), at least for everything new. It's hard to understand the semantics of `screens`, is `AccountListSidebar` really a screen?

src/renderer/components/screens/AccountListSidebar/AccountListSidebar.tsx
adzialocha: Oh, I see, this is deeply intertwined with the settings store .. okay, let's leave it then as a hack here as I can see how a refactor is too much work

docs/THEMES.md
Simon-Laux: depends on what we want to do, my last state of discussion with @r10s was that we want to allow different themes for different accounts.
If we stay with that idea then it makes sense to theme the sidebar separately (maybe offer a dark vs light mode switch).

If we make it themable and don't support per-account themes then we could even do crazy gradient designs like discord in the future: https://discord.com/blog/bring-your-vibe-to-discord-with-new-themes-in-nitro

src/renderer/components/screens/AccountListSidebar/AccountListSidebar.tsx
Simon-Laux: it is at least on the same upper level in code, but sure we can change the folder structure, your comment makes sense.

src/renderer/components/screens/AccountListSidebar/AccountListSidebar.tsx
Simon-Laux: desktop settings are only used for the sync state and that will be stored in core in the future for each account separately so the info will become integrated into the result of `BackendRemote.rpc.getAllAccounts`.

Though the window variable hack for triggering to update it might live better in a react context on app level? or it's own store? so it can be triggered wherever it is needed.
The actual result is only needed here and passed down to the `AccountItem` components, which are nearly pure components, the only thing not pure is the unread count (and if you are pedantic also the `useTranslationFunction` hook)

src/renderer/components/screens/AccountListSidebar/AccountListSidebar.tsx
Simon-Laux: I added it to the todo in https://github.com/deltachat/deltachat-desktop/issues/3671

src/renderer/components/screens/AccountListSidebar/AccountListSidebar.tsx
Simon-Laux: Ideally when we have the core events it will just a simple use effect like updating unread state in the accounts:
```ts
  const [unreadCount, setUnreadCount] = useState<number>(0)
  useEffect(() => {
    const update = debounce(() => {
      BackendRemote.rpc
        .getFreshMsgs(account.id)
        .then(u => setUnreadCount(u?.length || 0))
        .catch(log.error)
    }, 200)
    update()
    const cleanup = [
      onDCEvent(account.id, 'IncomingMsg', update),
      onDCEvent(account.id, 'MsgsNoticed', update),
    ]
    return () => cleanup.forEach(off => off())
  }, [account.id])
```
maybe a bit more complex to handle updating single accounts instead of all of them if we want that optimisation

Simon-Laux:
> I'm receiving an "Failed loading chat Chat#22 from database\n\nCaused by:\n Query returned no rows" error when switching accounts sometimes (roughly every 2nd time), it seems to be a race condition but I can't pin down yet where it comes from.

I think the chatstore / message list store depends too much on the global selected account id.
If there is an action running and you switch accounts it might try to complete the action with the other account. so this needs a refactoring I think making sure those stores live inside of the react tree and are not reused globally if possible.
But this is too much to do inside of this pr.

CHANGELOG.md
Simon-Laux: git is hard 😅

src/renderer/components/screens/AccountDeletionScreen/AccountDeletionScreen.tsx
Simon-Laux: Bildschirmfoto 2024-02-09 um 18 29 57
seems harder for vscode to find, but sure I could do that.

Simon-Laux: also you can do default export always, you can still access other exports like with react:
```js
Import React, {useEffect} from"react"
```

src/renderer/components/screens/AccountDeletionScreen/AccountDeletionScreen.tsx
Simon-Laux: sounds like sth a lint could enforce? I agree with this in terms that I like it, but not enough to force myself to follow it as it does not matter much.

src/renderer/components/screens/AccountDeletionScreen/styles.module.scss
Simon-Laux: I mentioned it on the todo issue, I asked some people, I don't have an answer yet, if you also have fire/lava/campfire/danger pictures that you took and have the rights to you could suggest them in the desktop group.
This does not need to block this first pr.

src/renderer/components/screens/AccountDeletionScreen/styles.module.scss
Simon-Laux: @r10s what do you say?

src/renderer/components/screens/AccountListSidebar/styles.module.scss
Simon-Laux: it is used by `AccountItem` and all sidebar components share this style file. the class on the top of the file is copied from the other avatars of delta chat desktop, cause the new styling system does not allow me to overwrite classes from different files.

src/renderer/components/screens/AccountListSidebar/styles.module.scss
Simon-Laux: account is not inherited or am I missing something? .avatar is lazily cloned because I just copied code, we can copy instead of import the react code as well to make the code prettier, but it is already somewhat independent.

with the new system we can not really overwrite existing classes (or at least I don't know how)

Simon-Laux:
> Another bug seems to be the Connectivity component not cancelling some async check when it got unmounted, should be easy to fix

It is likely because of the debounce, so not as easy to fix if we want to keep the debouncing. I'm not sure if we could use a ref for that? maybe we need to even convert this into a class component where we could keep around a "was destroyed" variable that we can check before calling set state.

adzialocha:
> > Another bug seems to be the Connectivity component not cancelling some async check when it got unmounted, should be easy to fix
> 
> It is likely because of the debounce, so not as easy to fix if we want to keep the debouncing. I'm not sure if we could use a ref for that? maybe we need to even convert this into a class component where we could keep around a "was destroyed" variable that we can check before calling set state.

Keeping an unmounted variable sounds good!

src/renderer/components/AccountListSidebar/AccountHoverInfo.tsx
```diff
@@ -0,0 +1,157 @@
...
+        )}
+      </>
+    )
+  }
+
+  return (
+    <div className={styles.accountHoverInfo} role='tooltip'>
+      {content}
+      <div className={styles.hoverInfoFooter}>
+        <AccountSize accountId={account.id} /> - {account.id}
```
r10s, line 64: this "number" looks like a bug :) we should not show debug info that are not meaningful at all for the user in primary UI.

```suggestion
        <AccountSize accountId={account.id} />
```

i am also not sure about showing "size" that prominently, but i'd give that a try. it is useful in general and also somehow clear in meaning

Simon-Laux: > i am also not sure about showing "size" that prominently, but i'd give that a try

I want to eventually have a storage size section in the settings where you can manage storage usage kinda like telegram has.

Maybe it would make sense to move the account id to the about dialog, since we also have the dc version there.
It can be useful for understanding logs that people send us and to find the account folder on disk.

Simon-Laux: The account size is also interesting to differentiate accounts when you have the same one multiple times, that was one of the main reasons why I added it as a tooltip in the old account switcher.

Simon-Laux: made a pr that updates the about dialog to show the selected account id and to show more versions (core, electron, node - wanted to do that for a longer time already anyway)

src/renderer/components/screens/AccountDeletionScreen/styles.module.scss
adzialocha: I propose that anything which is clickable should be indicated to the user with a pointer cursor.

"Settings" and "Create Account" already have one, why not also for switching accounts?

docs/THEMES.md
adzialocha: I like that!

src/renderer/components/screens/AccountDeletionScreen/AccountDeletionScreen.tsx
adzialocha: I've experimented with a linter in a separate branch

src/renderer/components/screens/AccountDeletionScreen/styles.module.scss
adzialocha: I'd remove this TODO then and just leave it as is. If there's no answer, it will be merged and likely stay forever until someone else makes another PR, but I'd not consider this a TODO anymore

src/renderer/components/screens/AccountListSidebar/styles.module.scss
adzialocha: > with the new system we can not really overwrite existing classes (or at least I don't know how)

You don't apply inheritance in the new system (which leads to complexity which is hard to follow).

Instead you do what I've proposed here:

> Patterns to avoid this are to introduce variants of the sub components you want to style and pass this over as a property (for example <MyComponent secondary /> etc.), if the styling is too far away, then consider creating a new component. Another, more hacky, but still okay approach would be to give it a className and pass over your customized css over like that, though I'd avoid that if it is changing the component a lot

adzialocha (approved):
Looking all great and working super on my machine. For me this is ready to merge!

I think there's some small "anti patterns" in how CSS modules work (inheritance, when to decide for stand-alone components, etc.), but I'm happy to refactor them in follow-up PRs.

PR #3625, opened 2024-01-08
Fix `WebxdcSendToChat` dialog regression after large dialog refactor

adzialocha:


Simon-Laux (approved):

scss/dialogs/_webxdc_send_to_chat.scss
Simon-Laux, line 1: I'm wondering whether we could merge this into the component styles and get rid of this file? or are it's classes reused elsewhere?

adzialocha: Yeah, that would be nice, and indeed it's not used anywhere else. Thing is, it would need more changes, for example in `ChatListItemNormal` and other components. Don't know if we want to do that now ..

PR #3627, opened 2024-01-08
Improve position of verified icon in profile info

### LINKED ISSUE #3620: "green checkmark" looks detached in profiles
the "green checkmark" is usually shown directly beside the group or contact name, it is somehow regarded to be a part of it. 

however, this is not true for the group profiles - when the name is split to two rows, there is quite some random looking gap:

Screenshot 2024-01-06 at 13 27 47

(i also think, in this example, a wrap should not be needed at all)

if it is not possible for whatever reason to add the icon as a "inline", i would at least adapt the padding between text and icon to not more than 1em - and defer wrapping

adzialocha:
Inline verified icon with display name and center it with text when displaying profile info in groups and contacts.

Before:

2024-01-08-115901_608x312_scrot

After:

2024-01-08-115231_618x276_scrot

Bonus-Level:

* New `ProfileInfoHeader` component which is now reused across "view profile" and "view group" dialogs

Closes #3620

### DIFF
--- a/scss/dialogs/_group-styles.scss
+++ b/scss/dialogs/_group-styles.scss
@@ -1,28 +1,15 @@
-p.group-name,
 input.group-name-input {
-  margin-left: 0px;
-  margin-bottom: 0px;
-  font-size: x-large;
-  width: 100%;
-  border: 0;
-  height: auto;
   background-color: transparent;
-  color: var(--bp4MenuText);
-  user-select: text;
-}
-
-p.group-name {
-  display: flex;
-  p.trucated-name {
-    text-overflow: ellipsis;
-    overflow: hidden;
-    margin-bottom: 0;
-  }
-}
-
-input.group-name-input {
   border-bottom: solid;
   border-color: var(--loginInputFocusColor);
+  border: 0;
+  color: var(--bp4MenuText);
+  font-size: x-large;
+  height: auto;
+  margin-bottom: 0px;
+  margin-left: 0px;
+  user-select: text;
+  width: 100%;
 }
 
 .group-separator {
@@ -40,19 +27,8 @@ input.group-name-input {
   display: flex;
   margin-bottom: 20px;
 
-  .group-name {
-    overflow: hidden;
-    text-overflow: ellipsis;
-    line-height: 1;
-  }
-
   .group-name-input-wrapper {
     margin-left: 17px;
-
-    input.group-name-input {
-      top: -7px;
-      position: relative;
-    }
 
     p.input-error {
       color: var(--colorDanger);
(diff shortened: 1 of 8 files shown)


scss/dialogs/_group-styles.scss
```diff
@@ -1,28 +1,15 @@
...
-    overflow: hidden;
-    margin-bottom: 0;
-  }
-}
-
-input.group-name-input {
+  border: 0;
   border-bottom: solid;
   border-color: var(--loginInputFocusColor);
+  color: var(--bp4MenuText);
```
farooqkz, line 6: Can we get rid of BP here? Our plan is to slowly get rid of BluePrint

src/renderer/components/Avatar.tsx
```diff
@@ -94,10 +96,11 @@ export function AvatarFromContact(
...
 
-export function ClickForFullscreenAvatarWrapper(props: {
-  filename: string | null
-  children: React.ReactNode
-}) {
+export function ClickForFullscreenAvatarWrapper(
+  props: PropsWithChildren<{
+    filename?: string
+  }>
+) {
```
farooqkz, line 103: Nice!

src/renderer/components/ProfileInfoHeader/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+    <div className={styles.profileInfoHeader}>
+      <ClickForFullscreenAvatarWrapper filename={avatarPath}>
+        <Avatar
+          displayName={displayName}
+          avatarPath={avatarPath}
+          color={color}
+          wasSeenRecently={wasSeenRecently}
+          large
+        />
+      </ClickForFullscreenAvatarWrapper>
```
farooqkz, line 35: Nice pattern here! I continue to learn from you!

farooqkz (changes requested):

scss/dialogs/_group-styles.scss
Simon-Laux: this is variable from us, many duplicate the var/color under a new name in themebase?

src/renderer/components/dialogs/ViewGroup.tsx
```diff
@@ -247,22 +243,14 @@ function ViewGroupInner(
...
-                  {chat.isProtected && <InlineVerifiedIcon />}
-                </p>
-              </div>
+            <DialogContent paddingBottom>
+              <ProfileInfoHeader
+                avatarPath={groupImage ? groupImage : undefined}
+                color={chat.color}
+                displayName={groupName}
+                isVerified={chat.isProtected}
+                wasSeenRecently={chat.wasSeenRecently}
```
Simon-Laux, line 252: groups don't have wasSeenRecently (green dot)

Simon-Laux (approved):

src/renderer/components/dialogs/ViewGroup.tsx
adzialocha: I'll make it optional to reflect that in the API

PR #3638, opened 2024-01-23
idea: reaction tooltip

Simon-Laux:
if we want this we should probably implement it more efficiently, like caching contacts and maybe implementing getting the contact ids in the jsonrpc layer


Simon-Laux:
we would need to redo this with a custom tooltip that lives in html to be able to set the fonts so it uses the same emojis as the rest of the app, this would also let us get rid of the delay when showing it.

PR #3639, opened 2024-01-24
Do not add changelog device message on fresh accounts

### LINKED ISSUE #3637: do not add "Changelog" device message for new accounts
the "Changelog" device messages should be added only for real updates

on new installations or accounts, the "What's new?" messages only clutters the device messages with unimportant information (either the user is new to the app, so everyrhing is new and nothing is changed, or an existing user has created a new account, they have seen the changelog anyways)

this is esp. important _new_ as we aim to add information wrt getting in contact for chatmail users, however, it was a bit annoying all the time. 

moreover, this makes untranslated changelogs less annoying as the user is not confronted with a language mix upfront.

when reading https://c.delta.chat/classdc__context__t.html#a1a2aad98bd23c1d21ee42374e241f389 , not added changelog on updates was always the gist, however, somehow forgotten or removed on the way :)

(technically, we cannot use the pattern from the example as the "welcome" message is added by come in some threads, so we need to track if the account was just created somehow, on android we do that by passing a flag around, see https://github.com/deltachat/deltachat-android/pull/2935) (still, `label` needs to be set with `msg=null` to avoid future adding)

Simon-Laux:
closes #3637

depends on https://github.com/deltachat/deltachat-core-rust/pull/5211 (`v1.134.0`)


src/renderer/deviceMessages.ts
```diff
@@ -1,13 +1,34 @@
...
+      await BackendRemote.rpc.addDeviceMessage(accountId, label, {
+        text: null,
+        html: null,
+        viewtype: null,
+        file: null,
+        location: null,
+        overrideSenderName: null,
+        quotedMessageId: null,
+        ...msg,
+      })
```
adzialocha, line 23: Isn't the third argument just a string here?

src/renderer/deviceMessages.ts
```diff
@@ -1,13 +1,34 @@
 import { BackendRemote } from './backend-com'
 
-export async function updateDeviceChats(accountId: number) {
+export async function updateDeviceChats(
+  accountId: number,
+  skipCurrentChangelog: boolean = false
+) {
+  const UpdateMessage = async (
```
adzialocha, line 7: Function names are usually written in lowerCamelCase and classes are in UpperCamelCase, so I'd change this to `updateMessage` but since it is only used once here, why a function at all?

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@
 - Remember file open dialog locations across the current session and do not persist last save location across sessions anymore #3615
 - Disable three-dot-menu when not applicable (map, other gallery tabs) #3523
 - move pin icon in chatlist after date #3636
+- do not add changelog device message on fresh accounts
```
adzialocha, line 36: PR no missing

src/renderer/ScreenController.tsx
```diff
@@ -241,6 +241,7 @@ export default class ScreenController extends Component {
               selectAccount: this.selectAccount,
               onAddAccount: async () => {
                 const accountId = await BackendRemote.rpc.addAccount()
+                updateDeviceChats(accountId, true) // skip changelog
                 await this.selectAccount(accountId)
```
adzialocha, line 245: More a sidenote with ideas on refactoring: This stuff is crucial "business logic" (even though I hate that word) which is buried deep down in the UI between some components, maybe we can separate these things one day and have the general "onboarding" logic disconnected from the view

adzialocha (changes requested):

src/renderer/deviceMessages.ts
```diff
@@ -1,13 +1,34 @@
...
+        file: null,
+        location: null,
+        overrideSenderName: null,
+        quotedMessageId: null,
+        ...msg,
+      })
+    }
+  }
+
   // const tx = window.static_translate
```
adzialocha, line 27: Could be removed since dead code

src/renderer/deviceMessages.ts
Simon-Laux: you might have multiple device messages you want to send, also to separate the if statement from the content, otherwise you'd need to repeat the label.

src/renderer/ScreenController.tsx
Simon-Laux: sure, but such a thing needs to be done carefully with much testing.

src/renderer/deviceMessages.ts
Simon-Laux: sure but on the other hand you might need it when translating the update message (which we currently don't do because it would require bringing the string into the android repo, waiting for translations and then pulling it again) - but iOS and android translate this string.

BTW: the `tx`, there is a script that searches the code for missing translation keys, it looks for `tx("key")`, that is the reason it is mapped everywhere to that variable name.

src/renderer/deviceMessages.ts
Simon-Laux: other option could be an array of device messages, but I think that would be overkill.

Simon-Laux: anyways not important enough to discuss

adzialocha (approved):

Simon-Laux:
Looks like I should not have merged this, the core on master is not new enough (it needs 1.134.0, master has 1.133.2), I'll make a pr to update core to 1.134.0 now. 🤦

PR #3642, opened 2024-01-24
Hide reactions UI for some messages or chats

adzialocha:


adzialocha:
@Simon-Laux I need some guidance if this covers all the cases, is there something missing?

Simon-Laux (changes requested):
Looks good besides the two small comments

Simon-Laux:
- [ ] Video chat invitations don't display it, so I would just disallow reacting on those too.

Issue #3647, opened 2024-01-29
allow creating/consuming invite-links in-app


r10s:
the QR code copy/paste options should be able to deal with invite links directly

- for copying, invite-links can be created by a one-liner as in https://github.com/deltachat/deltachat-android/pull/2943/files#diff-3cd645a2c944b39351f145cd1d6fe006de4dde3b59438c945a19a581ffcfcd8aR73 (there is no core function for that currently)

- for pasting, probably not much to do as core currentl `main` accepts these links directly - but should be tested :)

PR #3650, opened 2024-01-31
Copy invite link to clipboard instead of only QR code data

adzialocha:
Closes: https://github.com/deltachat/deltachat-desktop/issues/3647

### DIFF
--- a/src/renderer/utils/invite.ts
+++ b/src/renderer/utils/invite.ts
@@ -1,0 +1,7 @@
+const INVITE_URL_HOST = 'i.delta.chat'
+
+/** Converts an invite QR code to an HTTP invite URL */
+export function qrCodeToInviteUrl(qrData: string): string {
+  const code = qrData.split(':', 2)[1].replace('#', '&')
+  return `https://${INVITE_URL_HOST}/#${code}`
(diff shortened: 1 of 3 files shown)


adzialocha:
Tested parsing, seems to work as well against latest core version from `main`

src/renderer/utils/invite.ts
```diff
@@ -0,0 +1,7 @@
+const INVITE_URL_HOST = 'i.delta.chat'
+
+/** Converts an invite QR code to an HTTP invite URL */
+export function qrCodeToInviteUrl(qrData: string): string {
+  const code = qrData.split(':', 2)[1].replace('#', '&')
```
r10s, line 5: [double checked](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace) as i am always confused by that: replace() only replaces the first occurrence, fine :)

r10s (commented):
nice, also tested a bit.

with the next core update, one should also be able to paste these links back to the app.

r10s (approved):

r10s (approved):

PR #3660, opened 2024-02-04
remove unused css for reply hover button

Simon-Laux:
as this is just unused css this does not need a changelog entry.

We now have hover buttons again, if we want a reply button (latest discussions with adz and r10s said no), then we better add it in the new hover button component that adz made.


Simon-Laux:
ci failure is unrelated to change, most likely sth wrong with the npm package for the core/jsonrpc types or ci caching issue because it complains about the recently changed device message api.

PR #3661, opened 2024-02-04
z-index refactoring

Simon-Laux:
remove z-index where it was not needed, where it was needed tried to convert it to a lower or local/scoped z-index and put the remaining global ones into one file (`_variables.scss`).

## Commit messages
- scope the FAB z-index to the chatlist to get rid of its global z-index
- remove unessesary z-index from .group-image-edit-button
- reduce dependency on z-index in qr scanner ui
- remove dependence on z-index from fullscreen media view
- rm z-index from connectivity toast
- rm unnecessary z-indexes in jump-down-button
- put all global z-index values into a scss map in _variables.scss
- fix fmt

### DIFF
--- a/scss/_global.scss
+++ b/scss/_global.scss
@@ -144,12 +144,13 @@ code {
   image-rendering: pixelated;
 }
 
+// toast for errors and success messages
 .user-feedback {
   position: fixed;
   top: 60px;
   text-align: center;
   width: 100%;
-  z-index: 100;
+  z-index: map-get($z-index, user-feedback);
   pointer-events: none;
 
   p {
--- a/scss/chat/_chat-list-item.scss
+++ b/scss/chat/_chat-list-item.scss
@@ -327,7 +327,7 @@
   --size: 50px;
   --iconSize: 20px;
   position: fixed;
-  z-index: 20;
+  z-index: 1; // local z-index scoped to the chatlist
   bottom: 15px;
 
   width: var(--size);
--- a/scss/chat/_chat-list-item.scss
+++ b/scss/chat/_chat-list-item.scss
@@ -327,7 +327,7 @@
   --size: 50px;
   --iconSize: 20px;
   position: fixed;
-  z-index: 1; // local z-index scoped to the chatlist
+  z-index: map-get($z-index, chatlist-scope-floating-action-button);
   bottom: 15px;
 
   width: var(--size);
--- a/scss/chat/_chat-list.scss
+++ b/scss/chat/_chat-list.scss
@@ -1,5 +1,5 @@
 .chat-list {
-  z-index: 0; // this scopes the z-index of the floating action button to the chatlist
+  z-index: $z-index-new-local-scope; // needed for .floating-action-button
   width: 30%;
   height: $content-height;
   float: left;
--- a/scss/chat/_chat-list.scss
+++ b/scss/chat/_chat-list.scss
@@ -1,4 +1,5 @@
 .chat-list {
+  z-index: 0; // this scopes the z-index of the floating action button to the chatlist
   width: 30%;
   height: $content-height;
   float: left;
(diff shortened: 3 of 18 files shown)


Simon-Laux:
ci failure is unrelated to change, most likely sth wrong with the npm package for the core/jsonrpc types or ci caching issue because it complains about the recently changed device message api.

scss/_global.scss
```diff
@@ -144,12 +144,13 @@ code {
   image-rendering: pixelated;
 }
 
+// toast for errors and success messages
```
ralphtheninja, line 147: Cheers! :beers: :smile:

scss/chat/_chat-list-item.scss
```diff
@@ -327,7 +327,7 @@
   --size: 50px;
   --iconSize: 20px;
   position: fixed;
-  z-index: 20;
+  z-index: 1; // local z-index scoped to the chatlist
```
ralphtheninja, line 330: You could also move these to the variables file. Name the key accordingly to the comment, e.g. `'local-scope-chatlist'` or similar. This way we don't need these comments and we really have all z-index values in one file.

scss/dialogs/_qr-code.scss
```diff
@@ -27,7 +27,7 @@
 
 .scan-qr-description {
   position: relative;
-  z-index: 10;
+  z-index: 1;
```
ralphtheninja, line 30: Same here. This is also a locally scoped z-index?

scss/message/_message-list.scss
```diff
@@ -23,11 +23,10 @@
     bottom: calc(-100% + #{$height} + #{$marginBottom});
     height: $height;
     width: $width;
-    z-index: 10;
 
     .counter {
       position: relative;
-      z-index: 10;
+      z-index: 1;
```
ralphtheninja, line 29: Ditto.

src/renderer/components/AbsolutePositioningHelper/styles.module.scss
```diff
@@ -1,4 +1,6 @@
+@import '../../../../scss/variables';
+
 .absolutePositioningHelper {
   position: fixed;
-  z-index: 1000;
+  z-index: map-get($z-index, absolutePositioningHelper);
```
ralphtheninja, line 5: I'd use the same convention for all keys, e.g. `'absolute-positioning-helper'`. I realize this class isn't named the same way, but we can change this later.

ralphtheninja (commented):
Good that we got rid of some z-index values that weren't needed. But maybe consider moving all the ones that are needed to the variables file, now we're sort of in the middle of not centralizing the z-index values and also spreading some out.

Note that the key values for the z-index map doesn't _have_ to be named as the css class. They could be given different meaning in the key string. See my comments.

scss/chat/_chat-list.scss
```diff
@@ -1,4 +1,5 @@
 .chat-list {
+  z-index: 0; // this scopes the z-index of the floating action button to the chatlist
```
ralphtheninja, line 2: Same here

scss/dialogs/_qr-code.scss
Simon-Laux: no it isn't not really necessary with those z-index 1 values, but sure for completeness sake I can do it.

Simon-Laux:
moved all into the map, added missing local scopes and made an extra sass var for creating new local scopes: `$z-index-new-local-scope`.
I also removed the `z-index` of the map overlay, as it could also be solved by reordering the html elements.

Simon-Laux:
No time to wait for a review, I tested it myself and self-reviewed the code to ensure that it does not break anything obvious.

ralphtheninja:
Good job!

Issue #3628, opened 2024-01-08
Make setting display name mandatory when creating new account over QR Code Scan (DCACCOUNT scheme)


adzialocha:
Related Android issue: https://github.com/deltachat/deltachat-android/issues/2913

Simon-Laux:
only when creating a new account over qr code scan (DCACCOUNT scheme), I assume?

r10s:
it is reasonable to also expect a displayname when saving things on later edits

r10s:
also change the used string from `qraccount_success_enter_name` to `set_name_and_avatar_explain` (note that the latter does not require a placeholder)

if a message is needed to remind the user to fill you the name, use `please_enter_name`.

PR #3663, opened 2024-02-07
Require setting display name when creating profile for first time after scanning DCACCOUNT invite URL

adzialocha:
Closes: https://github.com/deltachat/deltachat-desktop/issues/3628

### DIFF
--- a/src/renderer/components/dialogs/EditProfileDialog/index.tsx
+++ b/src/renderer/components/dialogs/EditProfileDialog/index.tsx
@@ -10,41 +10,32 @@ import ProfileImageSelector from './ProfileImageSelector'
 import Dialog, {
   DialogBody,
   DialogContent,
+  DialogFooter,
   DialogHeader,
-  OkCancelFooterAction,
+  FooterActionButton,
+  FooterActions,
 } from '../../Dialog'
+import Callout from '../../Callout'
 import useTranslationFunction from '../../../hooks/useTranslationFunction'
+import useAlertDialog from '../../../hooks/useAlertDialog'
 
 import type { DialogProps } from '../../../contexts/DialogContext'
 
-export default function EditProfileDialog({
-  onClose,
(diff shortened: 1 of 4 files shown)


src/renderer/components/helpers/OpenQrUrl.tsx
```diff
@@ -172,7 +172,7 @@ export default async function processOpenQrUrl(
         openDialog(ConfirmationDialog, {
           message: tx(message, replacementValue),
           cb: resolve,
-          confirmLabel: tx('login'),
+          confirmLabel: tx('login_title'),
```
adzialocha, line 175: Interestingly the `login` translation key did not exist. I'm temporarily using this one now. Should we introduce a `login` key or leave it like that?

src/renderer/components/dialogs/EditProfileDialog/index.tsx
```diff
@@ -10,59 +10,40 @@ import ProfileImageSelector from './ProfileImageSelector'
...
     <Dialog canOutsideClickClose={false} onClose={onClose}>
-      <DialogHeader title={title} />
-      {EditProfileDialogInner({
-        settingsStore,
-        onClose,
-        cancelLabel,
-        confirmLabel,
-        firstSetup,
-      })}
+      <DialogHeader title={tx('pref_profile_info_headline')} />
```
adzialocha, line 34: This is the title being used on Android as well

src/renderer/components/helpers/OpenQrUrl.tsx
r10s: let's leave it as is - there are lots, esp. older strings, that do not follow our current default of "lowercase_text_with_underscores ~= key" - login_title seems one of the less weird examples here :)

r10s (approved):
thanks a lot - lgtm, also tested the PR a bit

Issue #3622, opened 2024-01-07
add QR icon to main screen


r10s:
on android and iOS, QR code scanning is available from the main screen.

for consistency, but also as QR code scanning becomes more important for chatmail, we want to have the option on the mainscreen on desktop as well, eg, as

<img width="768" alt="Screenshot 2024-01-07 at 15 32 43" src="https://github.com/deltachat/deltachat-desktop/assets/9800740/b02aa5bc-0555-4ff2-94da-37aff2ad2b98">

(we can leave the old QR code icon in the drawer for now; the drawer is subject to change anyways and may be replaced by an account sidebar anyways)

once, this is done, we can adapt the [help](https://delta.chat/en/help#howtoe2ee) and remove the on "Desktop the QR Icon is in the left-side sandwich menu" extra wording

PR #3665, opened 2024-02-08
Add QR scan button next to search input on main screen

adzialocha:
2024-02-08-120959_636x119_scrot

Closes: #3622


images/icons/qr.svg
Simon-Laux, line 1: we can probably update the references to `../images/qr_icon.svg` and remove the old file so it is not duplicated.
Or just remove the old file and remove the qr button from the sidebar.

src/renderer/components/SearchInput/SearchInputButton.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+type Props = {
+  icon: IconName
+  onClick: () => void
+  size?: number
+}
+
+export default function SearchInputButton({ icon, onClick, size = 20 }: Props) {
+  return (
+    <button className={styles.searchInputButton} onClick={onClick}>
```
Simon-Laux, line 17: aria label for accessibility would be nice (not our current focus, but we will need to fix accessibility in the future)

Simon-Laux (approved):
I like the idea of replacing the item, not sure if that will affect accessibility, but I (as a non-impaired person) like it.

images/icons/qr.svg
Simon-Laux: I will make a quick follow-up pr for this, as I already have the editor open

Simon-Laux: https://github.com/deltachat/deltachat-desktop/pull/3666

adzialocha: Nice, thanks! I've also meant to do the refactor in a separate PR.

src/renderer/components/SearchInput/index.tsx
```diff
@@ -35,11 +54,21 @@ export default function SearchInput(props: Props) {
...
+      {hasValue && (
+        <SearchInputButton
+          aria-label='Clear'
+          icon='cross'
+          onClick={handleClear}
+        />
+      )}
+      {!hasValue && (
+        <SearchInputButton
+          aria-label='Scan QR Code'
```
Simon-Laux, line 66: we should also translate these labels

Simon-Laux (approved):

PR #3670, opened 2024-02-08
Account Sidebar: No window titlebar on MacOS

Simon-Laux:
other apps like discord do this, looks quite nice

But of course might bring new bugs with it.

### DIFF
--- a/scss/main_screen/_navbar_wrapper.scss
+++ b/scss/main_screen/_navbar_wrapper.scss
@@ -70,6 +70,8 @@
   }
 }
 
-.no-drag, #hamburger-menu-button, #chat-list-search {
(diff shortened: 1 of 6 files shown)


adzialocha:
I can't test this as I don't have MacOS but I really like the idea!

Simon-Laux:
Ok let's merge this seems most people agree that it is cool.
The only thing we need to care now for is to mark elements in the navbar as `-webkit-app-region: drag` or `-webkit-app-region: no-drag`, depending on wether the click and drag should be considered moving the window or handled by the react app. So when adding a new button to the navbar don't forget to add the `.no-drag` class, so it is actually clickable.

Issue #3668, opened 2024-02-08
Remove Sidebar menu


Simon-Laux:
with the addition of the account switcher sidebar (#3621) the sidebar-menu that you access over the hamburger menu on the top left corner becomes a confusing ui element. Also most of the options are already accessible in different places.
So after moving the remaining options we can remove this menu.

All functions of the sidebar and if they are already accessible else where and could just be removed
- Header
   - [X] Avatar and email address -> already in settings
   - [X] Connectivity -> if broken already in the toast and always in settings or since #3621 when you hover over an account.
   - [X] Saved Messages -> exists by default and when you delete it you get a device message with instruction on how to re-create it
   - [X] QR-Scanner -> #3665 moved this to the navbar on the main screen.
 - Menu
   - [X] New Chat -> we already have the plus icon as floating action button on the chatlist
   - [x] Blocked Contacts -> needs to be moved to settings like on android and iOS
   - [X] Archived Chats -> the chatlistitem link is now alway on the top, so no need anymore for that button
   - [X] All Media -> in context menu on account in #3621
   - [X] Settings  -> in context menu on account in #3621
   - [x] Help -> already on `F1` and in window menu, but would make sense to also link it in the settings dialog like on android and iOS, or maybe even better: a `?` button in the account switcher bar at the  bottom as last item?
 - Footer
    - [X] version of the app -> already in about dialog
    - [x] link to about dialog -> already accessible in the window titlebar, but would be useful to have it somewhere else as well (for browser version or people that don't know about the sidebar, but it is less important for users than the help unless they need the version to report a bug to us, maybe also put it into the settings?)

Simon-Laux:
closed by https://github.com/deltachat/deltachat-desktop/pull/3621

PR #3672, opened 2024-02-08
make sidebar menu obsolete

Simon-Laux:
- move about and help as buttons to settings
- move unblock contacts to settings

addresses #3668 so we can remove the sidebar menu as soon as we have the account sidebar merged (#3621).

Bildschirmfoto 2024-02-08 um 21 45 08 Bildschirmfoto 2024-02-08 um 21 45 15

We don't need a changelog entry as when we merge this we will remove the sidebar and that removal will get the changelog entry.


r10s (approved):
nice, lgtm

PR #3677, opened 2024-02-11
improve about dialog

Simon-Laux:
- add more versions to about dialog and update order of existing information.

Bildschirmfoto 2024-02-11 um 18 28 32

### DIFF
--- a/src/renderer/components/dialogs/About.tsx
+++ b/src/renderer/components/dialogs/About.tsx
@@ -76,8 +76,6 @@ export default function About({ onClose }: DialogProps) {
 
   useEffect(() => {
     BackendRemote.rpc.getSystemInfo().then(info => {
-      console.warn({ info })
-
       setCoreVersion(info['deltachat_core_version'])
       setSqliteVersion(info['sqlite_version'])
     })
--- a/src/renderer/components/dialogs/About.tsx
+++ b/src/renderer/components/dialogs/About.tsx
@@ -41,7 +41,11 @@ export function DCInfo(_props: any) {
 
   return (
     <>
-      <h3>Version details:</h3>
+      <h3>Debug Info for System &amp; Selected Account</h3>
+      <p>
+        Local Account Id:{' '}
+        <b>{window.__selectedAccountId || '(no account selected)'}</b>
+      </p>
       <div className='dialog-about__dc-details'>
         <table>
           <tbody>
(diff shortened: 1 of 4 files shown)


farooqkz:
You must understand when you make CI unhappy, you make me unhappy. And when you make me unhappy, I nag all the day for you to death.

src/renderer/components/dialogs/About.tsx
```diff
@@ -41,7 +41,11 @@ export function DCInfo(_props: any) {
...
   return (
     <>
-      <h3>Version details:</h3>
+      <h3>Debug Info for System &amp; Selected Account</h3>
+      <p>
+        Local Account Id:{' '}
+        <b>{window.__selectedAccountId || '(no account selected)'}</b>
+      </p>
       <div className='dialog-about__dc-details'>
         <table>
```
farooqkz, line 50: Usually it's a good idea to use a CSS grid or flexbox rather than a table.

Simon-Laux: it's a debug view we shouldn't spend to much time on it right now.

Simon-Laux: also in this case the info fits well into a table IMO

Simon-Laux:
CI should be fixed now.

> But do you think it makes sense to merge it after my Button refactor PR?

it merges into the sidebar pr

src/renderer/components/dialogs/About.tsx
adzialocha: Yeah! Tables were meant for ... tables! So it's actually a perfect use case!

src/renderer/components/dialogs/About.tsx
```diff
@@ -67,6 +71,18 @@ export function DCInfo(_props: any) {
 export default function About({ onClose }: DialogProps) {
   const tx = useTranslationFunction()
 
+  const [coreVersion, setCoreVersion] = useState('')
+  const [sqliteVersion, setSqliteVersion] = useState('')
+
+  useEffect(() => {
+    BackendRemote.rpc.getSystemInfo().then(info => {
+      console.warn({ info })
```
adzialocha, line 79: I'd use `info` as a log level here. If you want to hide it from the developer by default, use `debug`, that's still not a warning but optional

adzialocha (approved):
Minor change request for the log level, otherwise all good and helpful. Thank you!

src/renderer/components/dialogs/About.tsx
Simon-Laux: ah good catch it should be removed, I just used it for debugging

PR #3678, opened 2024-02-12
Refactor message meta data component

adzialocha:


adzialocha:
Let me know if the issue disappeared for you or if it's still reproducible! Thank you!

Simon-Laux (approved):
LGTM

adzialocha:
Oupsi, still didn't fix the issue. It's some funky race condition, but I think I understand what's going on (something about the keys per message), will investigate

adzialocha:
I'll merge this for now as a nice refactor but not as a fix. For the actual fix we might need more time and I'm not sure if that's a priority

Issue #3635, opened 2024-01-18
"Search in Chat" for archived chats broken


r10s:
when opening "Search in Chat" on an archived chat, no search input field is available:

<img width="414" alt="Screenshot 2024-01-18 at 19 23 08" src="https://github.com/deltachat/deltachat-desktop/assets/9800740/d4b004fa-e118-495f-8ee7-a03cfadb7c47">

the search input field only becomes available when exiting the archive using the "up" button

r10s:
i'd call it a bug as no "search input field" pops up when you select "search in chat", but yes, whatever, i am fine without this tag as well :)

> What should we do instead to indicate that we're currently in the "Archived Chats"?

well, the indication is missing already now, when you tap "up" to finally get the "search input field".

maybe the following will do:

- never hide the "exit archive" button-icon, so there is at least some hint that one is in the archive

- when selecting "search in chat", directly show and focus the "search input field" as done for non-archived chats as well (i think, archived and non-archive should just behave similar)

this means, when typing in a search string, you'll have the "X" button to end the search as well as the button to exit archive. but i think, it is good to have stable buttons for that (btw, in case the exit-archive button shares logic with the qr-button: also for the qr-button it is probably better if it is shown stable even if there is a search)

adzialocha:
> to be clear: this issue is about the "search in chat" available from the chat's menu in the upper right corner; the issue is not about the "global search" that is not available when in archive

Aha, this was the bit I was missing! Now it makes sense :+1:

PR #3679, opened 2024-02-12
Show search bar when searching chat in archived mode

adzialocha:
Closes: #3635


Simon-Laux (approved):
works in my first tests and code looks fine

PR #3680, opened 2024-02-12
Multi Account Notifications

Simon-Laux:
While the multiaccount sidebar is cool, with multiaccount notifications it will become awesome :star:.

## Progress

- [x] enable multi account notifications and switch account when clicking on them
- [x] mute for accounts
	- [x] save property inside of the specified account "ui.desktop.muted"
	- [x] check when creating a notification if account is muted
	- [x] account item
		- [x] context menu option to mute
		- [x] visual indication (grey badge if any, if no unread then show mute icon in badge)
		- [x] hover info for muted accounts
	- [x] show option to mute account in settings and separated selected account settings visually from global settings
- [X] show notifications for other accounts also when window is in foreground
- [x] not only switch account but also open correct chat (~~maybe setting ui.lastchat before switching?~~)
- [x] Changelog entry

###### future questions:

I answered them myself for now, but we can still discuss them here or later.

- [ ] when DC is in foreground and a new message comes in for a not selected account, should there be a notification? probably yes?
- [ ] if account is muted but selected, should there be notifications? probably no?


adbenitez (approved):
Just did typo check 👀

_locales/_untranslated_en.json
```diff
@@ -11,5 +11,14 @@
   },
   "background_sync_disabled_explaination": {
     "message": "Background Sync Disabled, Account is only synced when selected"
+  },
+  "account_muted_description": {
+    "message": "Notifications are disabled for this Account"
```
adzialocha, line 16: This is a little lengthy, how about just "Muted" or "Notifications muted"?

_locales/_untranslated_en.json
```diff
@@ -11,5 +11,14 @@
...
+  },
+  "pref_for_all_accounts": {
+    "message": "For All Accounts"
```
adzialocha, line 19: The uppercase is a little confusing, I'd write "For all accounts" and "For the current account" or even better: "All accounts" and "Current account"

src/renderer/stores/accountNotifications.ts
```diff
@@ -0,0 +1,82 @@
+/// this store is to keep track of what accounts are muted, basically for caching the core ui config key.
+/// in the future it can also cache custom notifcations sounds
```
adzialocha, line 2: Why is caching necessary? Does it give us any performance benefits we can measure?

src/renderer/components/Settings/Notifications.tsx
```diff
@@ -13,6 +20,7 @@ export default function Notifications({ desktopSettings }: Props) {
 
   return (
     <>
+      <SettingsHeading>{tx('pref_for_all_accounts')}</SettingsHeading>
```
adzialocha, line 23: Some ideas here:

1. How about changing the order, making "current account" first and "all accounts" second
2. Remove "current account" heading, as similar to all other settings we can assume that they are mainly for the current account already (except when they are explicitly named to affect the application on a system-level)
3. Disable the "current account" switch as soon as the system-wide setting got turned off

adzialocha:
Regarding your questions:

> when DC is in foreground and a new message comes in for a not selected account, should there be a notification? probably yes?

I expect not application to send me notifications while I'm looking at it. I'd assume that the "inner UI" would indicate that I've received a new message somewhere, for example through the blue badge counter popping up. So I'd say no notifications as soon as the app is in the foreground

> if account is muted but selected, should there be notifications? probably no?

Agree with "no", as no selected account should cause notifications, similar to my point above

_locales/_untranslated_en.json
Simon-Laux: idea was to explain the icon, not sure if "Muted" would be enough? but I guess we could try

_locales/_untranslated_en.json
Simon-Laux: we have this strange casing for the other heading in in the settings too.

src/renderer/stores/accountNotifications.ts
Simon-Laux: it removes needless roundtrips to core and allows us to synchronously check the fields.

adzialocha: Caching is an optimization and I'd be cautious about introducing that too early, especially with it being hard to debug and to control.

To me it currently sounds like we would also be fine without it

adzialocha: > it also gives us a clean & easy way make sure all react components showing this information are updated when we update it.

That would be the role of a state handling logic but not of a caching layer

Simon-Laux:
> I expect not application to send me notifications while I'm looking at it. I'd assume that the "inner UI" would indicate that I've received a new message somewhere, for example through the blue badge counter popping up. So I'd say no notifications as soon as the app is in the foreground

telegram and WhatsApp send notifications for other chats on mobile if I remember correctly.
The notification also has the function that you see the text and when you click on it the right account is directly opened.
Just increasing some high badge number without a sound is not really noticeable. It's not like the current account where you see the preview of the last message in the chatlist.

_locales/_untranslated_en.json
r10s: yip, that would also be my suggestion, just "Muted" is enough. the user has selected "Muted" for the account before and there is an icon, so it is pretty clear

_locales/_untranslated_en.json
r10s: the "strange casing" was pushed not least from the former desktop team at https://github.com/deltachat/deltachat-android/pull/1979 :)

titles may be a corner case, but i think it is good to use the "mostly upper" also there, i also still agree with the former decision, it still matches nicer in general to most other apps and conventions.

but let's go for for "All Account"/"Current Account" then, if it is really needed to mention account scope (it is an information that may open questions to the remaining settings dialog, i'd like to get rid of that, but i do not have a good alternative yet)

src/renderer/components/AccountListSidebar/AccountItem.tsx
```diff
@@ -82,6 +98,40 @@ export default function AccountItem({
...
+      <div
+        className={classNames(styles.accountMuted, {
+          [styles.accountBadgeMuted]: muted,
+        })}
+        aria-label='Account Notifications Muted'
+      >
+        <div className={styles.accountMutedIcon} />
+      </div>
+    )
+  }
```
r10s, line 133: for muted accounts the switch from grey-badge to the mute icon is unexpected. also, the other way round, it looks as if mute is suddenly disabled.

it would be better to show the icon stable inside the avatar. this also looks more like a state and is also consisten with what we're doing on chats.

i suggest sth like the following, the icon could also be black with white drop shadow:

Screenshot 2024-02-12 at 19 18 16

r10s (commented):
this is a very nice welcome improvement! thanks a lot for pushing that forward ❤️

src/renderer/stores/accountNotifications.ts
Simon-Laux: > Caching is an optimization and I'd be cautious about introducing that too early, especially with it being hard to debug and to control.

While I agree with your statement in general, I think it is the right solution in this case.

_locales/_untranslated_en.json
Simon-Laux: > the "strange casing" was pushed not least from the former desktop team at https://github.com/deltachat/deltachat-android/pull/1979 :)

I thought it was the people writing the website, not sure who came up with this first xD

src/renderer/components/Settings/Notifications.tsx
Simon-Laux: Bildschirmfoto 2024-02-13 um 16 19 47

1. not sure, if we have 3. it makes more sense to have the global setting on top
2. we can only do that if we do 1. and we introduce a new string for the toggle switch for muting the account, otherwise the ui gets confusing.
3. you can still change it in the account sidebar and it still affects the badge color, so I think really disabling the input is more effort (if it should be consistent with the rest of the app, otherwise we could just ignore that you can change it somewhere else.).

Also there will be a new (probably account specific) setting in the future: setting a different/custom notification sound #3583, so keep that in mind when we talk about structuring the dialog. (I think it will become a select that opens a dialog, but still an extra setting that takes up space.

src/renderer/components/AccountListSidebar/AccountItem.tsx
```diff
@@ -194,3 +194,13 @@ async function markAccountAsRead(accountId: number) {
     )
   )
 }
+
+const AudioMutedIcon = () => (
```
adzialocha, line 198: Why not use the `Icon` component here?

src/renderer/stores/accountNotifications.ts
adzialocha: Thanks for clarifying, I would like to look at this closer some time!

Simon-Laux:
>  I'd recommend to not use the OS-level-notifications API for non-OS concerns.

We are not going to reimplement custom notifications inside of DC like telegram does on mobile, are we? I think that's too much work for now.

PR #3683, opened 2024-02-12
Open DC Invite Links directly inside delta chat from links clicked in messages and html emails

Simon-Laux:


src/main/windows/html_email.ts
```diff
@@ -384,7 +384,10 @@ function makeBrowserView(
     }`)
 
   const openLink = (url: string) => {
-    if (url.startsWith('mailto:')) {
+    if (
+      url.startsWith('mailto:') ||
+      (url.startsWith('https://i.delta.chat/') && url.includes('#'))
```
adzialocha, line 389: I'd use the nice `isInviteLink` helper method here as well

adzialocha (approved):

PR #3690, opened 2024-02-14
remove stale CONTRIBUTORS.md

r10s:
the file is stale since years -
this always happens sooner or later with lists no one maintains, better avoid them totally :)

also, the contributors are easily accessible using git or github.


Simon-Laux:
> this always happens sooner or later with lists no one maintains, better avoid them totally :)

was maintained by hallmark, we removed that dependency a while ago

PR #3696, opened 2024-02-19
Add repology badge to readme

link2xt:


adzialocha:
Cooool! I've just moved it a bit further down under "Installation" as it fairly large for being the first people see 😅

PR #3698, opened 2024-02-19
slightly wider account sidebar as suggested in the forum

Simon-Laux:
make account switcher 2px wider: https://support.delta.chat/t/a-few-ideas-to-make-dc-desktop-1-43-even-better/2931

(so traffic lights look more centered on macOS)

before:
Bildschirmfoto 2024-02-19 um 22 41 19

after:
Bildschirmfoto 2024-02-19 um 22 41 44

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [Unreleased][unreleased]
 
 ### Changed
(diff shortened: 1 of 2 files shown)


r10s (approved):
yeah, looks a little more relaxed and polished that way 👌

PR #3699, opened 2024-02-20
Experimental: Refactor state logic

adzialocha:
**🔬 This is heavily WIP 🧪** 

> I'm exploring refactoring the state logic by experimenting with the [Jotai](https://jotai.org/) framework. It seems to be a light way to combine state handling and side-effects without buying into complex state stores and reducers. At the same time it seems to have nice patterns with async, debouncing, abort handlers, clean-up functions etc.

**Diagram:**

```
                      Core RPC, FS, LocalStorage, ...

                                  ▲  │
                                  │  │
                                  │  │
I/O                               │  │
                                  │  │
                    ┌─────────────┴──▼────────────┐
                    │                             │
                    │          /backend           │
                    │                             │
                    └────▲──┬─────────────▲──┬────┘
                         │  │             │  │
                ┌────────┼──┼─────────────┼──┼─────────┐
                │        │  │             │  │         │
                │        │  │             │  │         │
                │   ┌────┴──▼───┐      ┌──┴──▼────┐    │
                │   │           ◄──────┤          │    │
Global State    │   │  /atoms   │      │ /effects │    │
                │   │           ├──────►          │    │
                │   └────▲──┬───┘      └──────────┘    │
                │        │  │                          │
                │        │  │                          │
                └────────┼──┼──────────────────────────┘
                         │  │
                ┌────────┼──┼───────────────────────────────────────────┐
                │        │  │                                           │
                │    ┌───┴──▼──┐                                        │
                │    │         │                                        │
                │    │ /hooks  │                                        │
                │    │         │                                        │
                │    └───▲──┬──┘           Local State                  │
                │        │  │                                           │
React           │        │  │             ┌──────────────────────────┐  │
                │        │  │             │                          │  │
                │  ┌─────┴──▼─────┐       │  /<Component>/hooks      │  │
                │  │              │◄──────┤                          │  │
                │  │ /components  │       │  /<Component>/contexts   │  │
                │  │              ├──────►│                          │  │
                │  └──────────────┘       └──────────────────────────┘  │
                │                                                       │
                └───────────────────────────────────────────────────────┘
```

**General ideas:**

* There's two kinds of states: *global state* which is sometimes dependent on external IO (for example by talking to the RPC API of the core or local storage) and *local state* which lives in a single React component or a sub-tree of them
* Every single global state we're interested in across the application is expressed in an Jotai "atom", living in a file inside a new `atoms` folder
* The actual atom / state is not exposed. With additional "read-only" and "write-only" atoms (see Jotai documentation) we allow state transitions and read access
* State transitions should be written without any large knowledge about IO
* Additionally, all state logic should never know _anything_ about React, it lives on its own. It serves an API to the outside which can be consumed by React, it uses the API of the `backend` code to interact with the core / local storage / filesystem or other IO
* Every method which talks to the backend (IO) is state-less and wrapped in an own method which lives in a file in a new `backend` folder
* Side-effects are expressed with "atom-effects" and are mounted in the getter / setter methods of each atom. Like this it gets easy to see what effects to expect when changing a value. Effects are methods which live in files in a new `effects` folder
* Side-effect methods, similar to state, should also be *as atomic as possible* and not handle multiple things which are unreleated to each other
* The React world consumes global state through hooks which wrap the atom state logic
* If no global state is required it is recommended to just use React `useState` and contexts and keep everything in the React universe

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,6 +26,7 @@
         "error-stack-parser": "^2.0.7",
         "filesize": "^10.1.0",
         "immutable": "^4.0.0",
+        "jotai": "^2.6.4",
         "mapbox-gl": "^1.12.0",
         "mime-types": "^2.1.31",
         "moment": "^2.29.2",
@@ -4738,6 +4739,26 @@
       },
       "engines": {
         "node": ">=10"
+      }
+    },
+    "node_modules/jotai": {
+      "version": "2.6.4",
+      "resolved": "https://registry.npmjs.org/jotai/-/jotai-2.6.4.tgz",
+      "integrity": "sha512-RniwQPX4893YlNR1muOtyUGHYaTD1fhEN4qnOuZJSrDHj6xdEMrqlRSN/hCm2fshwk78ruecB/P2l+NCVWe6TQ==",
+      "engines": {
+        "node": ">=12.20.0"
+      },
+      "peerDependencies": {
+        "@types/react": ">=17.0.0",
+        "react": ">=17.0.0"
+      },
+      "peerDependenciesMeta": {
+        "@types/react": {
+          "optional": true
+        },
+        "react": {
+          "optional": true
+        }
       }
     },
     "node_modules/js-tokens": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,6 +27,7 @@
         "filesize": "^10.1.0",
         "immutable": "^4.0.0",
         "jotai": "^2.6.4",
+        "jotai-effect": "^0.5.0",
         "mapbox-gl": "^1.12.0",
         "mime-types": "^2.1.31",
         "moment": "^2.29.2",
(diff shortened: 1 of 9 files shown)


Simon-Laux:
I'm not sure about keeping global state outside of the react tree at all. Also I'd start with the parts that currently make problems not with the parts that are already working, even if they are not pretty enough.

I also thought about porting the stores to a framework like `zustand` or `jotai`, but the custom store class already works fine so I don't see a priority in changing it now.

it looks interesting to move screen and selected account state outside of the react classes like that, but I think keeping global state like selected account is what leads to problems and we should not do it, maybe for desktop settings (`config.json` / `state.ts`) that are global, or for the account notification settings.

I mean we need to keep a reference to the selected account outside of react, but we should only use it where there can be no side effects like:

Use it for notifications and other system integrations but not for actions like delete chat, those actions **should always get the selected account from react where they are triggered**, so you do not get undefined/unwanted behaviour when switching the account right after triggering an action that contained multiple async interactions with core.

adzialocha:
Closing this for now, this is not meant to be merged anyhow

PR #3702, opened 2024-02-22
Add video chat instances icons

farooqkz:
This PR aims to do just these two:

 - [x] Fetch video chat instance's favicon and save them in a json file in `src`
 - [x] Show those icons in the frontend
 - [x] remove the initial render delay(about 1s on my machine).


Simon-Laux:
> Fetch video chat instance's favicon and save them in a json file in src

I think it makes not much sense to do it on every build because those won't change very often. 
This will make development slower by making "npm run build" slower, also it will fail on flatpak as you have no internet access there.

you can also save the icons of the default instances as icons to a subfolder of images/

Then the format of the video chat provider objects could be:

```ts
interface VideoChatInstance {
 // id for distinguishing between the instances
 id: string,
 // the url of the video chat provider that is given to core
 url: string,
 // path of the icon, either relative url for default providers or base64 for custom providers that user added that were saved in a core ui config
 icon: string, 
}
const VIDEO_CHAT_INSTANCES: Record<string, VideoChatInstance > = {
'systemli': { id: 'systemli', url: 'https://meet.systemli.org/$ROOM', icon: "../images/videochat_providers/systemli.ico" },
'autistici': { id: 'autistici', url: 'https://vc.autistici.org/$ROOM', icon: "../images/videochat_providers/autistici" },
}
```

farooqkz:
@Simon-Laux @adzialocha I need to investigate why the initial render takes so much time but hints are welcome.

Simon-Laux:
needs to be rebased

bin/add-video-chat-instances-icons.js
```diff
@@ -0,0 +1,48 @@
...
+    url: 'https://vc.autistici.org/$ROOM',
+    icon: '../images/videochat_instances/autistici.ico',
+  },
+}
+
+const getData = async () => {
+  let data = {}
+  for (const [id, props] of Object.entries(VIDEO_CHAT_INSTANCES)) {
+    for (const i = 1; i <= 3; i++) { // max 3 retries
+      const response = await fetch(props.url.replace('$ROOM', '/images/favicon.ico'))
```
Simon-Laux, line 24: might work, but parsing url and just taking hostname + `/favicon.ico` might be better. but ok for now as long as the default instances are jitsi 

but do not write code to parse urls yourself, just use https://nodejs.org/docs/latest-v18.x/api/url.html

Simon-Laux:
might make sense to show the icons pixelated, as blurry does not look good. (css `image-rendering: pixelated;`)
We can also try to get the icons from different places in the hope to get a higher resolution one (fetching the html and parsing the xml to get the `head` elements content:
- `<link rel="apple-touch-icon" href="images/apple-touch-icon.png">`
- icons referenced in a webmanifest: `<link rel="manifest" href="manifest.json" />`
- `<link rel="icon" type="image/png" href="images/favicon.ico?v=1">`

Simon-Laux:
But I'm not sure it's wise to prioritise this right now as there is still quite some stuff left todo to make it really useful, also I think the code structure can be much simplified.

farooqkz:
This PR is ready for review and just adds the icons to make it look prettier a bit. The option to add feature to let users add their own instances will be added through another PR.

farooqkz:
I think I did a wrong rebase :D

nicodh:
Current status:
image

Dark theme
image

@farooqkz what should we do with this PR? It needs some more improvements

Design:
- [ ] use less pixelated icons
- [ ] always white background icon to be background independant
- [ ] layout per item with proper padding and  
Code:
- [ ] define as chat instances as array, saved in "ui.dekstop.vc.instances" core config key
- [ ] option to delete and add (multiple) custom instances
- [ ] option to select/choose instance when sending a video chat invitation

r10s:
ftr, this is a successor of https://github.com/deltachat/deltachat-desktop/pull/3587

farooqkz:
Thanks for your review @nicodh I think maybe close it and come to it later?

nicodh:
Please reopen when you implemented the requested changes or if you need more feedback! :-)

PR #3704, opened 2024-02-23
fix broken html email window (CSP got broken with the recent electron update)

### LINKED ISSUE #3701: "show full message" html-view broken
the "show full message" html-view is broken at least on macOS using official 1.43.1 - the view is just empty:

i tried also with older messages, but i did not manage to get a single view. the same message can be shown in html on recent iOS.

this is what the log shows when using recent master:

<details>

```
TypeError: Headers.append: "default-src 'none';
font-src 'self' data:;
frame-src 'none';
img-src 'self' data:;
media-src 'self' data:;
style-src 'self' data: 'unsafe-inline';
form-action 'none';
script-src 'none';" is an invalid header value.
    at Object.webidl.errors.exception (node:internal/deps/undici/undici:1570:14)
    at Object.webidl.errors.invalidArgument (node:internal/deps/undici/undici:1581:28)
    at _Headers.append (node:internal/deps/undici/undici:2087:31)
    at fill (node:internal/deps/undici/undici:1965:19)
    at initializeResponse (node:internal/deps/undici/undici:5648:9)
    at new Response (node:internal/deps/undici/undici:5446:9)
    at /Users/bpetersen/projects/deltachat-desktop/src/main/windows/html_email.ts:327:12
    at AsyncFunction.<anonymous> (node:electron/js2c/browser_init:2:47399)
(node:1855) electron: Failed to load URL: email://index.html with error: ERR_UNEXPECTED
28.8s [D]main/themes: load theme file dc:light /Users/bpetersen/projects/deltachat-desktop/themes/light.css
28.8s [D]main/themes: render theme data
28.8s [D]main/themes: render theme data for theme: { name: 'Light Theme', description: 'Our default theme.' }
29.8s [i]main/state: Saving state to /Users/bpetersen/Library/Application Support/DeltaChat/config.json
```

</details>

Simon-Laux:
fixes #3701

### DIFF
--- a/src/main/windows/html_email.ts
+++ b/src/main/windows/html_email.ts
@@ -293,6 +293,8 @@ media-src 'self' data:;
 style-src 'self' data: 'unsafe-inline';
 form-action 'none';
 script-src 'none';`
+  .split('\n')
+  .join('') // since new electron version there can not be linebreaks in the header anymore: https://github.com/deltachat/deltachat-desktop/issues/3701
 const CSP_ALLOW = `
 default-src 'none';
 font-src 'self' data: http: https:;
@@ -303,6 +305,8 @@ style-src 'self' 'unsafe-inline';
 form-action 'none';
 script-src 'none';
 `
+  .split('\n')
+  .join('') // since new electron version there can not be linebreaks in the header anymore: https://github.com/deltachat/deltachat-desktop/issues/3701
 
 function makeBrowserView(
   account_id: number,
--- a/src/main/windows/html_email.ts
+++ b/src/main/windows/html_email.ts
@@ -292,9 +292,7 @@ img-src 'self' data:;
 media-src 'self' data:;
 style-src 'self' data: 'unsafe-inline';
 form-action 'none';
-script-src 'none';`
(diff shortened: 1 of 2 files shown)


src/main/windows/html_email.ts
```diff
@@ -292,7 +292,7 @@ img-src 'self' data:;
 media-src 'self' data:;
 style-src 'self' data: 'unsafe-inline';
 form-action 'none';
-script-src 'none';`
+script-src 'none';`.replace(/\n/g, '') // since new electron version there can not be linebreaks in the header anymore: https://github.com/deltachat/deltachat-desktop/issues/3701
```
r10s, line 295: not sure about this specific language, however, in general, replacing new line by spaces seems to be better in case some language needs whitespace as a token. 

(assume eg html with <tag key=value\notherkey> )

r10s (approved):

src/main/windows/html_email.ts
Simon-Laux: I think in this case the semicolon does this

PR #3707, opened 2024-02-28
remove unexpected empty space (bottom padding) from view profile dialog

Simon-Laux:


src/renderer/components/dialogs/ViewProfile/styles.module.scss
```diff
@@ -6,4 +6,5 @@
 .viewProfileDialogBody {
   display: flex;
   flex-direction: column;
+  padding-bottom: 0.6px;
```
Simon-Laux, line 9: zero padding resulted in a second scrollbar, I suspect the auto sizer rounds up the (ceil) the height it gets somewhere so it takes more than there actually is.

PR #3710, opened 2024-03-01
change export keys open directory confirm button label to "select"

### LINKED ISSUE #3709: "Open" instead of "Save" in directory selector
When exporting a key, the directory selector displays "_Open_" instead of "_Save_".
Version: 1.43.1
macOS 14.3.1

Simon-Laux:
closes #3709


farooqkz:
I agree with "Select [directory]" more.

adzialocha:
> I agree with "Select [directory]" more.

+1, should we just go for "Select" as we have that translation string already?

adzialocha (approved):

PR #3711, opened 2024-03-03
Fix: Changing display name of a contact does not change it immediately in the messages

### LINKED ISSUE #3703: Editting name of a contact does not change their name in the messages
1. Rename some of your contacts by opening their profile, edit button.
2. See that their new name does not appear in the messages.
3. Select another chat and select the chat with the contact and see that the new name appears.

farooqkz:
Fix the bug where changing display name of a contact does not change it immediately in the messages

Closes #3703

### DIFF
--- a/src/renderer/components/message/Message.tsx
+++ b/src/renderer/components/message/Message.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useContext } from 'react'
+import React, { useCallback, useContext, useState } from 'react'
 import reactStringReplace from 'react-string-replace'
 import classNames from 'classnames'
 import { C, T } from '@deltachat/jsonrpc-client'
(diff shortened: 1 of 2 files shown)


src/renderer/components/message/Message.tsx
```diff
@@ -88,7 +88,19 @@ const AuthorName = (
...
+  const update = async ({ contactId }: { contactId: number | null }) => {
+    if (contactId === id) {
+      const contact = await BackendRemote.rpc.getContact(
+        selectedAccountId(),
+        contactId
+      )
+      setDisplayName(contact.displayName)
+    }
+  }
+  onDCEvent(selectedAccountId(), 'ContactsChanged', update)
```
Simon-Laux, line 103: you need to wrap this in useEffect, `onDCEvent` returns a cleanup function that unregisters the even listener, useEffect makes sure it is called.

See the other places where `onDCEvent` is used

Simon-Laux (changes requested):

Simon-Laux (approved):

Issue #3675, opened 2024-02-09
Refactor to remove possible races when switching account in message list / chatstore


Simon-Laux:
> I'm receiving an "Failed loading chat Chat#22 from database\n\nCaused by:\n Query returned no rows" error when switching accounts sometimes (roughly every 2nd time), it seems to be a race condition but I can't pin down yet where it comes from.

I think the chatstore / message list store depends too much on the global selected account id (`window.__selectedAccountId`).
If there is an action running and you switch accounts it might try to complete the action with the other account. so this needs a refactoring I think making sure those stores live inside of the react tree and are not reused globally if possible.
But this is too much to do inside of this pr.

_Originally posted by @Simon-Laux in https://github.com/deltachat/deltachat-desktop/issues/3621#issuecomment-1936291644_

Simon-Laux:
The current idea is to remove the global chat store and move to a local chat store that lives inside of the react tree

## Where is it currently used

- ChatMethods.ts -> maybe move these functions to a chatlist functions context or take chatstore instance as parameter
	- `selectChat()`
		- ChatList
		- CreateChat dialog
		- Forward Message dialog
		- QrCode dialog
		- ViewGroup dialog
		- ViewProfile dialog
		- WebxdcSendToChat dialog
		- mailtoUrl
		- MessageMarkdown #mostly_inside_msg_list
		- messageFunctions #mostly_inside_msg_list
		- MainScreen
	- `setChatView()`
		- MessageMarkdown
		- MainScreen
	- `jumpToMessage()`
		- mediaAttachment (in gallery)
		- ChatListItemRowMessage - message search results
		- EmojiAndStickerPicker - sending sticker
		- ChatAuditLog Dialog - show message or webxdc in chat
		- Fullscreen media Dialog - show message in chat
		- Message #mostly_inside_msg_list
		- notifications.ts #outside_of_react
	- `unselectChat()`
		- Composer #mostly_inside_react_chat - blocking chat
		- MainScreen - unselect chat when going to global gallery
	- `setChatVisibility()`, `openDeleteChatDialog()` and `openBlockFirstContactOfChatDialog()` - (conditionally call `unselectChat()` internally)
		- ChatListContextMenu
		- ThreeDotMenu #mostly_inside_react_chat
	- `sendCallInvitation()` - jumps to new message
		- Composer -> Attachment Menu
	- `createChatByContactIdAndSelectIt()` - calls selectChat
[...]

PR #3725, opened 2024-03-07
Refactor chat store logic into dedicated React contexts

adzialocha:
* Remove `ChatStore`
* Move "global" chat state to `ChatContext`

Closes: https://github.com/deltachat/deltachat-desktop/issues/3675


src/renderer/ScreenController.tsx
```diff
@@ -341,17 +342,19 @@ export default class ScreenController extends Component {
...
-                selectedAccountId={this.selectedAccountId}
-                onAddAccount={this.addAndSelectAccount.bind(this)}
-                onSelectAccount={this.selectAccount.bind(this)}
-                openAccountDeletionScreen={this.openAccountDeletionScreen.bind(
-                  this
-                )}
-              />
-              {this.renderScreen()}
-            </div>
+            <ChatProvider>
```
Simon-Laux, line 345: the sidebar does not need the context, does it?

adzialocha: Yeah, that's right, it's not required. Dialogs need it though so it lives even lower now.

src/renderer/contexts/ChatContext.tsx
```diff
@@ -0,0 +1,77 @@
...
+  chatId?: number
+  selectChat: SelectChat
+  setChatView: SetView
+  unselectChat: UnselectChat
+}
+
+export const ChatContext = React.createContext<ChatValue | null>(null)
+
+export const ChatProvider = ({ children }: PropsWithChildren<{}>) => {
+  // @TODO: Explore how to refactor account state into its own context
```
Simon-Laux, line 33: maybe a a context for the screen controller, but might also make sense to  duplicate it as measure to prevent it going out of sync?
or have some use(Layout)Effect hook that removes the chat data as soon as the accountId changes?

src/renderer/components/Runtime.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { selectedAccountId } from '../ScreenController'
+
+import type { PropsWithChildren } from 'react'
+
+/**
+ * Helper component to hook React methods into the external "runtime". This
+ * allows us to interact with the underlying Electron runtime and
+ * operating system.
+ */
+export default function Runtime({ children }: PropsWithChildren<{}>) {
```
Simon-Laux, line 16: we should maybe rename this to avoid confusion with the actual runtime interface, make "RuntimeConnection" or similar, as this only sets some specific callbacks for the runtime

Simon-Laux: or RuntimeConnection or RuntimeConnectionController or whatever that's different enough from the interface name

src/renderer/hooks/usePrivateReply.ts
```diff
@@ -0,0 +1,39 @@
...
+      // retrieve existing draft to append the quotedMessageId
+      const oldDraft = await BackendRemote.rpc.getDraft(accountId, chatId)
+
+      await BackendRemote.rpc.miscSetDraft(
+        accountId,
+        chatId,
+        oldDraft?.text || null,
+        oldDraft?.file || null,
+        quotedMessageId,
+        'Text'
```
Simon-Laux, line 32: I wonder if this is correct, we should remember this pace in the code and take over the existing view type, at best we should make a new core/jsonrpc api for "add quoted message to existing draft".

with the composer refactoring this will definitely become incorrect as it recreates the message and so it would delete all status updates of a prepared webxdc message, it should modify the existing draft if it exists and not recreate the draft message with a new message id.

Simon-Laux: @adzialocha I know you just copied it over, this issue is older, I'm not blaming you ;)

src/renderer/contexts/ChatContext.tsx
adzialocha: There's different approaches we could take:

1. Call the `unselectChat` method on the `ChatContext` whenever `unselectAccount` in `ScreenController` gets called
2. Use `key={this.selectedAccountId}` to force `ChatContext` to rebuild with default values
3. Use an React effect to internally call `unselectChat` in `ChatContext` whenever account id changes

Currently I'm using the 2nd option, even though it is a little bit implicit. Probably doing 1. or 3. is better.

src/renderer/components/Runtime.tsx
adzialocha: I've changed it to `RuntimeAdapter` now, it's an "adapter" for the DeltaChat Desktop React implementation to the runtime interface

src/renderer/hooks/usePrivateReply.ts
adzialocha: Do you want to create an issue for that?

src/renderer/contexts/ChatContext.tsx
adzialocha: Looking at it again I think it's fine as it is right now (using `key`), later we can make it more explicit, for example with 1. but I think we might need to touch too much of the account logic for this then and this is out of scope of this PR

adzialocha: In any case, I've removed `accountId` from the `ChatProvider`, now it is passed in as an prop which makes more sense

src/renderer/ScreenController.tsx
```diff
@@ -354,19 +337,33 @@ export default class ScreenController extends Component {
...
+                      openAccountDeletionScreen={this.openAccountDeletionScreen.bind(
+                        this
+                      )}
+                    />
+                    {this.renderScreen()}
+                  </div>
+                </KeybindingsContextProvider>
+              </DialogContextProvider>
+            </ContextMenuProvider>
+          </ChatProvider>
```
adzialocha, line 366: This became a little cute context sandwich. Many dialogues are dependent on knowing about the chat state, while every screen needs to know about dialogues, this is why `ChatProvider` pops up in a fairly "global" position.

I don't think this is bad at all (a currently selected chat _is_ a very "global" state in this application), just wanted to note that here.

src/renderer/components/QrErrorMessage.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react'
+
+import useTranslationFunction from '../hooks/useTranslationFunction'
+
+type Props = {
+  url: string
+}
+
+export default function QrErrorMessage({ url }: Props) {
```
adzialocha, line 9: This became it's own little component as passing in JSX as an `message` value into an alert dialog component doesn't work within hooks. In any case, it should be cleaner like this now.

src/renderer/components/RuntimeAdapter.tsx
```diff
@@ -0,0 +1,42 @@
...
+  useEffect(() => {
+    runtime.onOpenQrUrl = (url: string) => {
+      const accountId = selectedAccountId()
+      processQr(accountId, url)
+    }
+
+    runtime.setNotificationCallback(async ({ accountId, msgId, chatId }) => {
+      if (window.__selectedAccountId !== accountId) {
+        await window.__selectAccount(accountId)
+      }
```
adzialocha, line 29: All of this accountId stuff smells a little bit, but I think it's fine

src/renderer/components/RuntimeAdapter.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { selectedAccountId } from '../ScreenController'
+
+import type { PropsWithChildren } from 'react'
+
+/**
+ * Helper component to hook React methods into the external "runtime". This
+ * allows us to interact with the underlying Electron runtime and
+ * operating system.
+ */
+export default function RuntimeAdapter({ children }: PropsWithChildren<{}>) {
```
adzialocha, line 16: I came up with this `RuntimeAdapter` as a glue between the "external runtime world" and React. Otherwise there's no way we get access to React hooks etc from the outside.

So far this has been solved with using callbacks on `window` which then somewhat call React code but all of this will then happen outside of the React lifetime and state cycle. Doing it like this instead could be a nice pattern for future refactorings.

Note that this is only useful for calling React code whenever an external event took place. For the other way around (calling external functions when an React event took place) we should be already set.

src/renderer/backend/chat.ts
```diff
@@ -67,6 +67,14 @@ export function markChatAsSeen(accountId: number, chatId: number) {
   clearNotificationsForChat(accountId, chatId)
 }
 
+export async function setChatVisibility(
+  accountId: number,
+  chatId: number,
+  visibility: T.ChatVisibility
+) {
+  await BackendRemote.rpc.setChatVisibility(accountId, chatId, visibility)
+}
```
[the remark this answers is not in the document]
adzialocha: Right, that might have been a little bit too much. The idea behind all of this to make it more clear whenever we talk to something which is not React related (better separation of layers), but I'm fine with not doing that

src/renderer/contexts/ChatContext.tsx
Simon-Laux: I think 2 (react `key` "hack") makes most sense

src/renderer/hooks/usePrivateReply.ts
Simon-Laux: https://github.com/deltachat/deltachat-desktop/issues/3750

src/renderer/backend/chat.ts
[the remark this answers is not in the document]
Simon-Laux: I think that layer of abstraction would be a bit too much boilerplate/overhead, I'm fine with wrapping it when it has side effects, also see `EffectfulBackendActions` in `backend-com.ts`, those also only exist for things that lack core events or are combined actions that call multiple functions, though the ones with `window.__refetchChatlist && window.__refetchChatlist()` will vanish with the upcoming chatlist events pr.

src/renderer/components/QrErrorMessage.tsx
Simon-Laux: maybe we could group the qr code related components into a qr folder

src/renderer/components/RuntimeAdapter.tsx
Simon-Laux: can't you also pass the account id and account selection function to this component?

src/renderer/components/chat/ChatListContextMenu.tsx
```diff
@@ -2,44 +2,60 @@ import React, { useContext, useState } from 'react'
...
+import useOpenViewGroupDialog from '../../hooks/useOpenViewGroupDialog'
+import useOpenViewProfileDialog from '../../hooks/useOpenViewProfileDialog'
+import useTranslationFunction from '../../hooks/useTranslationFunction'
+
+import type { T } from '@deltachat/jsonrpc-client'
+import type { UnselectChat } from '../../contexts/ChatContext'
 
 function archiveStateMenu(
-  chat: Type.ChatListItemFetchResult & { kind: 'ChatListItem' },
+  unselectChat: UnselectChat,
```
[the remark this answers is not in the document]
Simon-Laux: not really in this case as this is computed lazily only when you right click on a chat in the chatlist. Or am I missing something?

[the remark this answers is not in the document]
adzialocha: It depends where you look, but some `buildContextMenu` are not wrapped in any hook (like `useMemo` or `useCallback` etc., so they will get re-compiled on every render of the component, which you don't want for larger functions)

src/renderer/backend/chat.ts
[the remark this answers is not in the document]
adzialocha: I'll remove everything which doesn't do more logic or is a convenience thing :+1:

src/renderer/components/QrErrorMessage.tsx
adzialocha: I'll do this as part of the QR code refactor!

adzialocha:
I'm not sure why this is stuck on throwing linting errors about things which do not exist anymore, restarting the CI didn't help ..

In any case, this PR is ready for review and merging.

src/renderer/backend/chat.ts
```diff
@@ -0,0 +1,131 @@
...
+    return !contacts[contactId].isVerified
+  })
+}
+
+/**
+ * Helper method to determine the chat id of the "Device Messages" read-only chat.
+ *
+ * Note that there's currently no API to retrieve this id from the backend, this
+ * is why we're iterating over all currently available chats instead.
+ */
```
Simon-Laux, line 112: sidenote, I know the comment is not new:

I just tested performance and came to the conclusion that we do not need make a core api as long as we only use it on fresh accounts it's fast enough.

my account main account: ~163ms
a new account that only has like 20 chats: ~9-11ms
completely fresh account: ~2-3ms

This method is not exactly efficient, but we don't use it enough to justify moving it to core.

src/renderer/hooks/usePrevious.ts
```diff
@@ -0,0 +1,9 @@
+import { useEffect, useRef } from 'react'
+
+export default function usePrevious(value: any) {
```
Simon-Laux, line 3: naming nitpick: isn't this more like useInitialValue?

Simon-Laux (commented):
When switching accounts the account sidebar flickers, we should somehow prevent it from completely reloading (visually popping in unread counters) in that case.

Generally the code looks really nice now! Thanks very much for taking on this refactoring, it's a huge leap forward and was easy to review 🚀 🌟

src/renderer/hooks/useChat.ts
nicodh, line 1: Maybe put all hooks moved from ChatMethods in a folder hooks/chat/ to not overcrowd the hooks folder? Probably we should also move the hooks from components/helper/hooks.ts to the hooks folder, so there will be many soon...

Maybe comment: what's the reason for wrapping the ChatContext in a custom hook? - or is it a no-brainer for react pros? ;-)

nicodh (approved):
Looks like a good refactoring and cleanup to me!
Only some ideas for naming and comments. In general I prefer more verbose comments, they are removed anyway when compiled.

src/renderer/hooks/useChat.ts
Simon-Laux: makes it quicker to use, only one import instead of two.

src/renderer/components/MessageListView.tsx
```diff
@@ -0,0 +1,61 @@
...
+      case ChatView.Media:
+        return (
+          <Gallery
+            ref={galleryRef}
+            chatId={chat.id}
+            onUpdateView={onUpdateGalleryView}
+          />
+        )
+      case ChatView.Map:
+        return <MapComponent sendMessage={sendMessage} selectedChat={chat} />
```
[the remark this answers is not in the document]
adzialocha: `withSendMessageHook` is a container function which still needs to pass the required prop into the wrapped component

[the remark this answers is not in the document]
adzialocha: See: https://stackoverflow.com/questions/53371356/how-can-i-use-react-hooks-in-react-classic-class-component

src/renderer/contexts/ChatContext.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+      // Jump to last message if user clicked message twice
+      // @TODO: We probably want this to be part of the UI logic instead
+      if (nextChatId === chatId) {
+        window.__internal_jump_to_message?.(undefined, false, undefined)
+      }
+
+      // Already set known state
+      setActiveView(ChatView.MessageList)
+      setChatId(nextChatId)
```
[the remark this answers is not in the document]
adzialocha: `getFullChatById` is an async method and we already have informations we can change before we need to wait for it. Some other components further down rely on the chatId being set, so they can already do their work as well while the full chat data is loaded independently

src/renderer/hooks/usePrevious.ts
adzialocha: No, this is always giving the value which was set before the render cycle

src/renderer/hooks/useSelectLastChat.ts
```diff
@@ -0,0 +1,36 @@
+import { useCallback, useEffect } from 'react'
+
+import useChat from './useChat'
+import useHasChanged from './useHasChanged'
```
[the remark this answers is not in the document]
adzialocha: This might help understanding what is going on: https://legacy.reactjs.org/docs/hooks-faq.html#how-to-get-the-previous-props-or-state

src/renderer/hooks/useChat.ts
adzialocha: > Maybe put all hooks moved from ChatMethods in a folder hooks/chat/ to not overcrowd the hooks folder? Probably we should also move the hooks from components/helper/hooks.ts to the hooks folder, so there will be many soon...

I've tried to not structure files by semantics (usually more subjective) but more by how deep they are used in the component graph. If a hook is used potentially everywhere in the application then it should live in the root `hooks` folder, if it is used only within the component `MyComponent`, then it should live in the folder of that component. Another typical pattern is to organize them by "views", so if there's a `Settings` view then all hooks and contexts and components for this view can be put into that folder etc.

I'm also fine with putting it in a `hooks/chat` folder, I'm slightly worried that the semantics are a little blurry plus I don't mind having too many files in a place if they signal to me that they're all of same importance, but that's not a strong argument, I'm open for other solutions as well.

Besides of all of that I can have a scan over all hooks and see where they're actually used, maybe some of them can actually be moved further down to their only consuming components at least.

src/renderer/contexts/ChatContext.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+    const onContactsModified = (
+      eventAccountId: number,
+      { contactId }: { contactId: number | null }
+    ) => {
+      if (eventAccountId !== accountId) {
+        return
+      }
+
+      if (!chat) {
```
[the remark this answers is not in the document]
adzialocha: I'm actually not 100% sure why this update is required at all, could you explain what this is for, then I'll add it as a comment. Thank you!

src/renderer/components/RuntimeAdapter.tsx
Simon-Laux: I mean why is the hook passed:
1. in `MessageListView.tsx` to `withSendMessageHook(MapComponent)` as prop
2. in `withSendMessageHook` it is not used from the prop instead fetched from the context again and then set to the `MapComponent` as prop. technically `MapComponent` gets `sendMessage` set two times or am I missing sth?

src/renderer/backend/chat.ts
Simon-Laux: But the disadvantage to it is that it triggers multiple renders instead of just one... anyways I tried both ways and can not see a huge difference on my fast device.

src/renderer/contexts/ChatContext.tsx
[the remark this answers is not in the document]
Simon-Laux: I guess if contact name, avatar or recently seen changes on a 1on1 chat. maybe only the `recently seen` in reality I don't know in my head if the other events trigger a chat update instead or also rely on the ContactsChanged event

src/renderer/hooks/usePrevious.ts
Simon-Laux: then it makes sense to document that in a comment to avoid future confusion

src/renderer/contexts/ChatContext.tsx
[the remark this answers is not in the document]
adzialocha: I'm still confused why this update is required and why it is only for 1:1 chats. Maybe I'll leave it as-is for now (as this is how it was before as well) and we can add commentary / adjustments later?

PR #3740, opened 2024-03-31
Work on ImageCropper

### LINKED ISSUE #1779: Profile picture is distorted if not square
- Operating System (Linux/Mac/Windows/iOS/Android): Ubuntu Focal
- Delta Chat Version: 1.4.3 via deb packages
- Expected behavior:

Screenshot from 2020-07-07 10-21-03

- Actual behavior:

Screenshot from 2020-07-07 10-20-38

maxphilippov:
fixes #1779

- a component to crop any image to a certain size by zooming and dragging the original to fit a predefined shape
- this component is used in ProfileImageSelector by altering result of ImageSelector upload
- there's a circle pattern as default and square pattern, images are cropped as square

What's left to do:
- decide on cancel behavior
- decide if there should be some way to tell cropper to use the whole image besides the zooming in manually

### DIFF
--- a/src/renderer/components/ImageCropper/index.tsx
+++ b/src/renderer/components/ImageCropper/index.tsx
@@ -55,7 +55,7 @@ export default function ProfileImageCropper({
   const posX = useRef<number>(0)
   const posY = useRef<number>(0)
 
-  const speedMultiplier = 2
+  const containerScale = useRef<number>(1.0)
   const zoom = useRef<number>(1.0)
   const initialZoom = useRef<number>(1.0)
 
@@ -117,11 +117,11 @@ export default function ProfileImageCropper({
         'rect(' +
         (y - hh) +
         'px ' +
-        (x + wh) +
+        (x + hw) +
         'px ' +
         (y + hh) +
         'px ' +
-        (x - wh) +
+        (x - hw) +
         'px)'
       )
     }
@@ -132,13 +132,12 @@ export default function ProfileImageCropper({
       return
     }
 
-    container.current.style.transform =
-      'scale(' +
-      Math.min(
-        container.current?.clientWidth / targetWidth,
-        container.current?.clientHeight / targetHeight
-      ) +
-      ')'
+    containerScale.current = Math.min(
+      container.current?.clientWidth / targetWidth,
+      container.current?.clientHeight / targetHeight
+    )
+
+    container.current.style.transform = `scale(${containerScale.current})`
     initialZoom.current = zoom.current = Math.min(
       targetWidth / fullImage.current.clientWidth,
       targetHeight / fullImage.current.clientHeight
@@ -175,8 +174,10 @@ export default function ProfileImageCropper({
       (targetHeight / zoom.current - imgH) / 2
     )
 
+    // we don't multiply nX and nY by zoom.current here since image scale does that for us
     cutImage.current.style.clipPath = makeSelector(imgW / 2 + nX, imgH / 2 + nY)
 
+    // here we are in absolute coordinates, we have to multiply image offset by zoom
     const imgX = (ctxW - imgW) / 2 - nX * zoom.current
     const imgY = (ctxH - imgH) / 2 - nY * zoom.current
 
@@ -202,16 +203,11 @@ export default function ProfileImageCropper({
       ev.preventDefault()
 
       dragging.current = false
-
-      // we never go faster with zoom above 1, but we speed up zoomed-out (below 1) by inverting the zoom value
-      const zoomSpeed = Math.min(
-        speedMultiplier / zoom.current,
-        speedMultiplier
-      )
-
       ;[posX.current, posY.current] = moveImages(
-        posX.current - (ev.clientX - startX.current) * zoomSpeed,
-        posY.current - (ev.clientY - startY.current) * zoomSpeed
+        posX.current -
+          (ev.clientX - startX.current) / zoom.current / containerScale.current,
+        posY.current -
+          (ev.clientY - startY.current) / zoom.current / containerScale.current
       )
     }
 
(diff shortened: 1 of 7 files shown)


Simon-Laux:
> decide if there should be some way to tell cropper to use the whole image besides the zooming in manually

maybe a reset button to reset the zoom and have it at use full image by default (or the center part of it if the image has a different aspect ratio than 1:1)

farooqkz:
Can it wait one or two more days if my review is required?

Simon-Laux:
> Can it wait one or two more days if my review is required?

yes, also your review is optional on this one, though I'd appreciate it.

farooqkz:
> > Can it wait one or two more days if my review is required?
> 
> yes, also your review is optional on this one, though I'd appreciate it.


My reviews aren't only good for the PR authors but also for myself to learn. I've learned at list a thing or two from Max in the past PR.

src/renderer/components/ImageCropper/index.tsx
```diff
@@ -0,0 +1,284 @@
...
+    }
+  })
+  return (
+    <Dialog
+      className={styles.imageCropperDialog}
+      canEscapeKeyClose
+      onClose={onClose}
+      canOutsideClickClose={false}
+    >
+      <DialogHeader></DialogHeader>
```
adzialocha, line 246: I'd give it a title, not for every user it'll be clear what I can "do" in this dialog, especially when it will be always shown

src/renderer/components/ImageCropper/index.tsx
```diff
@@ -0,0 +1,284 @@
...
+      <DialogBody className={styles.imageCropperDialogBody}>
+        <DialogContent>
+          <div
+            ref={container}
+            style={{
+              height: '500px',
+            }}
```
adzialocha, line 253: Can this `height` style be defined in `.imageCropperContainer` instead?

src/renderer/components/ImageCropper/index.tsx
```diff
@@ -0,0 +1,284 @@
...
+      <DialogFooter>
+        <FooterActionButton
+          onClick={() => {
+            onCancel()
+            onClose()
+          }}
+        >
+          {tx('cancel')}
+        </FooterActionButton>
+        <FooterActionButton onClick={onSubmit}>{tx('save')}</FooterActionButton>
```
adzialocha, line 280: The `FooterActionButton` needs to be wrapped in `FooterActions` for correct layout, check other dialog implementations for reference

src/renderer/components/ImageCropper/styles.module.scss
```diff
@@ -0,0 +1,41 @@
+// We need to force the dialog to full height to make `AutoSizer` work
+.imageCropperDialog {
+  height: 100%;
+}
+
+.imageCropperDialogBody {
+  display: flex;
+  flex-direction: column;
+}
```
adzialocha, line 9: To me it looks like this is not really required. Also there's no use of `AutoSizer` I can spot?

src/renderer/components/dialogs/EditProfileDialog/ProfileImageSelector.tsx
```diff
@@ -1,8 +1,11 @@
 import React from 'react'
 
 import ImageSelector from '../../ImageSelector'
+import ImageCropper from '../../ImageCropper'
```
adzialocha, line 4: Can we also add this cropper for editing group images?

adzialocha (changes requested):
I have a few comments and adjustments, but the code looks really good to me!

src/renderer/components/dialogs/EditProfileDialog/ProfileImageSelector.tsx
maxphilippov: I think I've dragged this one for a while :) So maybe I'll make a new PR for the group images?

Simon-Laux (changes requested):
Trying it out I found these bugs and suggestions:

- [x] Scrolling somehow makes the crop circle stutter. (at first I suspected the pinch to zoom gesture on the macOS touchpad, but this also happens with my mouse, more noticeable when scrolling quickly with the loose/free wheel option):

https://github.com/deltachat/deltachat-desktop/assets/18725968/ca256c10-b215-47d3-8d7d-6ea72bed292a

- [ ] Dragging direction is inverted, croppers on touch screen and other drag and drop interface don't invert the movement:


https://github.com/deltachat/deltachat-desktop/assets/18725968/4d74fbf6-5cea-4b1d-871d-19aed2e39ce3


- [ ] Tweak the layout a bit
      - real dialog title in the style of other dialogs - for consistency: "Crop Image"
      - move text what do do down as subtitle for cropper like it is in the qr code scanner dialog
      -  change pagging/margin/space so buttons do not look aligned wrongly

- [ ] after cropping when I click again on it the file explorer opens in the temp folder, I think that was not intentional and is a bug


Other than that it works great, thanks for tackling this long standing issue! ❤️ 🚀

Simon-Laux (changes requested):
- [ ] there is a bug when cropping, as shown by these screenshots:


Bildschirmfoto 2024-05-10 um 04 06 34
Bildschirmfoto 2024-05-10 um 04 06 50

Simon-Laux:
- [ ] Visually sometimes the old image stays, I guess we could add some randomness to the temp file name to prevent browser caching
- [ ] on MacOS touchpad scrolling and zooming seem to be triggered at the same time, this makes it very hard to use. you might not have such a touchpad, so I could investigate that in a follow-up pr.

maxphilippov:
> * [ ]  Visually sometimes the old image stays, I guess we could add some randomness to the temp file name to prevent browser caching
> * [ ]  on MacOS touchpad scrolling and zooming seem to be triggered at the same time, this makes it very hard to use. you might not have such a touchpad, so I could investigate that in a follow-up pr.
> * [ ]  the `title` attribute tooltip is a bit unexpected.
Fixed the name.

Try a new zoom out, there's a small fidgeting when max zoomed out I don't like but looks like it's more consistent now. I also need a feedback on the direction of the zoom, cause I had to invert touchpad zoom direction on Windows to make it consistent.

You might actually need to investigate a title attribute, I couldn't reproduce it on Windows. Does it stay after the dialog opens or it's reacting somehow through the dialog popup? Browsers have a lot of bugs with these titles (or maybe even OSes do, I got it all the time on Win).

src/renderer/components/ImageCropper/index.tsx
```diff
@@ -0,0 +1,295 @@
...
+export default function ProfileImageCropper({
+  setProfilePicture,
+  filepath,
+  shape,
+  onClose,
+  onCancel,
+  targetWidth,
+  targetHeight = targetWidth,
+}: {
+  setProfilePicture: (path: string) => void
```
Simon-Laux, line 30: rename this to something more generic like `onResult` or `callback`

Simon-Laux (approved):

PR #3741, opened 2024-04-01
feat: Realtime channels

Septias:
Part of https://github.com/deltachat/deltachat-core-rust/pull/5346

TODO:
- Transport info about webxdc apis not being operational.

### DIFF
--- a/_locales/en.xml
+++ b/_locales/en.xml
@@ -504,8 +504,7 @@
     <string name="multidevice_same_network_hint">Make sure both devices are on the same Wi-Fi or network</string>
     <string name="multidevice_this_creates_a_qr_code">This creates a QR code that the second device can scan to copy the profile.</string>
     <string name="multidevice_install_dc_on_other_device">Install Delta Chat on your other device (https://get.delta.chat)</string>
-    <!-- "I Already Have a Login / Add as Second Device” should be the same text as defined by the keys onboarding_alternative_logins and multidevice_receiver_title -->
-    <string name="multidevice_tap_scan_on_other_device">Start Delta Chat, tap “I Already Have a Login / Add as Second Device” and scan the code shown here</string>
+    <string name="multidevice_tap_scan_on_other_device">Start Delta Chat, tap “Add as Second Device” and scan the code shown here</string>
     <!-- Shown inside a "QR code card" with very limited space; please formulate the text as short as possible therefore. The placeholder will be replaced by name and/or address eg. "Scan to set up second device for Alice (alice@example.org)" -->
     <string name="multidevice_qr_subtitle">Scan to set up second device for %1$s</string>
     <string name="multidevice_receiver_title">Add as Second Device</string>
@@ -587,7 +586,7 @@
     <!-- for classic email, we use the classical term "Account" -->
     <string name="login_explain">Log in with an existing e-mail account</string>
     <string name="login_subheader">For known e-mail providers additional settings are set up automatically. Sometimes IMAP needs to be enabled in the web settings. Consult your e-mail provider or friends for help.</string>
-    <string name="login_no_servers_hint">Delta Chat does not collect user data, everything stays on your device.</string>
+    <string name="login_no_servers_hint">There are no Delta Chat servers, your data stays on your device.</string>
     <string name="login_inbox">Inbox</string>
     <string name="login_imap_login">IMAP Login Name</string>
     <string name="login_imap_server">IMAP Server</string>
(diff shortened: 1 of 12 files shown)


static/webxdc-preload.js
```diff
@@ -62,6 +68,10 @@
       onStatusUpdate()
       return promise
     },
+    setEphemeralUpdateListener: (cb) => {
+      ipcRenderer.invoke('webxdc.sendGossipAdvertisement')
```
Simon-Laux, line 72: maybe we should trigger a native confirmation dialog to inform users that p2p "leaks"/exposes the users ip address to the other group memebers.

Simon-Laux: like "more info"(linking to inApp Help), "Deny", "Allow" and a checkbox "remember choice for this webxdc app"

hpk42:
On Tue, Apr 02, 2024 at 09:13 -0700, Simon Laux wrote:
> @Simon-Laux commented on this pull request.
> 
> 
> 
> > @@ -62,6 +68,10 @@
>        onStatusUpdate()
>        return promise
>      },
> +    setEphemeralUpdateListener: (cb) => {
> +      ipcRenderer.invoke('webxdc.sendGossipAdvertisement')
> 
> like "more info"(linking to inApp Help), "Deny", "Allow" and a checkbox "remember choice for this webxdc app"

would mean you need to track per-app what is allowed, and then allow folks
to change it later -- quite some UI and user-communication effort. 

whatsapp and signal enable P2P by default for voice-calls 
and have a single global option "always relay calls"/"Protect IP address in calls"
which defaults to "off" for both, and is in advanced options. 
Probably makes sense as a global option because either you care or you don't. 

ASFAIK iroh/integration is not able to do User1->relay1->relay2->User2 yet so
if i am not mistaken we can not right now offer it like this. 
I tentatively suggest to only introduce a global option to 
enable/disable ephemeral channels and have translated strings for what it means,
and probably default to off at first.

Septias:
I also think that a single configuration option should suffice. Maybe we can talk to the iroh people about our need for IP address protection.

static/webxdc-preload.js
```diff
@@ -67,8 +67,8 @@
       onStatusUpdate()
       return promise
     },
-    setEphemeralUpdateListener: cb => {
-      ipcRenderer.invoke('webxdc.sendGossipAdvertisement')
+    setEphemeralUpdateListener: async cb => {
+      await ipcRenderer.invoke('webxdc.sendGossipAdvertisement')
       ephemeralCb = cb
```
Simon-Laux, line 72:
```suggestion
    setEphemeralUpdateListener: cb => {
      ephemeralCb = cb
      return ipcRenderer.invoke('webxdc.sendGossipAdvertisement')
```

static/webxdc-preload.js
WofWca: > "leaks"/exposes the users ip address to the other group memebers

To be precise, it's not just that the IP is revealed to the members, but also that whoever is observing your connection can see who you're connecting to, which is more important when talking to people that you trust.

_locales/en.xml
```diff
@@ -755,6 +755,8 @@
     <string name="pref_edit_profile">Edit Profile</string>
     <string name="disable_imap_idle">Disable IMAP IDLE</string>
     <string name="disable_imap_idle_explain">Do not use IMAP IDLE extension even if the server supports it. Enabling this option will delay message retrieval, enable it only for testing.</string>
+    <string name="enable_realtime">Enable Realtime APIs</string>
+    <string name="enable_realtime_explain"> Enable Webxdc realtime APIs which create direct connection between devices and leak IP addresses to each other.</string>
```
Simon-Laux, line 759: these belong not here but in `_locales/_untranslated_en.json` instead, or in a pr to the android repo. this file is not to be edited by developers as it is overwritten with the next pull from transifex.

r10s, line 759: changing the strings here does not work afaik, strings will be overwritten here.

the source of the strings is at https://github.com/deltachat/deltachat-android/blob/main/res/values/strings.xml , so we need a PR there.

static/webxdc-preload.js
```diff
@@ -1,3 +1,47 @@
+/**
+ * @typedef {import('webxdc-types').RealtimeListener} RT
+ * @type {RT}
+ */
+class RealtimeListener {
+  constructor(ipc) {
+    this.listener = null
+    this.trashed = false
+    this.ipc = ipc
```
Simon-Laux, line 9: we should not expose that, it might be a security issue, do not store a reference to it in the class.

Simon-Laux (changes requested):

default.nix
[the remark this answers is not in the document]
Septias: Other nix users might need this as well, but I will add it to the .gitignore for now

static/webxdc-preload.js
Septias: I don't know why, but wrapper functions around IPC can not be transported through the electron portal...
I have to think of something else then..

Septias: nvm

_locales/en.xml
r10s: ftr, the strings `enable_realtime` and `enable_realtime_explain` can be pulled from transifex now

default.nix
Simon-Laux, line 1: this file is still part of git because it was already committed. needs some interactive rebase to remove the commit that added it. maybe it would be easier to squash everything into a single commit?

Septias: yeah lets just squash

Simon-Laux: the file would be still there, unless someone makes a commit to delete it. anyways I can do it

Simon-Laux:
So we decided to squash merge this when the time is ready.

- This depends on a core with this pr merged https://github.com/deltachat/deltachat-core-rust/pull/5607, so as of the time of writing this it is not merged yet.

Septias:
Core PR has been merged

Simon-Laux:
sorry was in a hurry and annoyed that I needed to fix to the new config key

PR #3747, opened 2024-04-04
perf: fix low FPS on chat scroll on some machines (again)

WofWca:
- [x] This is based on #3743, expecting that to get merged first

Before:
image

After:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 - fix chat audit dialog was going out of viewport on smaller screens #3736
 - fix long names breaking layout of reactions dialog #3736
 - hide "add second device" instructions when transfer has started #3748
-- improve chat scroll performance #3743
+- improve chat scroll performance #3743, #3747
 
 <a id="1_44_1"></a>
 
--- a/scss/message/_message-list.scss
+++ b/scss/message/_message-list.scss
@@ -85,37 +85,9 @@
   #message-list {
     position: absolute;
     bottom: 0;
+    // There is also `will-change: scroll-position` dynamically set on this
+    // element.
     overflow: scroll;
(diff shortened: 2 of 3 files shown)


WofWca:
Hold up, this increases scripting time significantly. Will try to optimize now.

WofWca:
> Hold up, this increases scripting time significantly. Will try to optimize now.

Ok, fixed. I didn't know React re-renders on all `setState`s and not just the ones that can actually affect DOM.

WofWca:
Ehh, I'm not sure about whether we should actually turn it off on timeout. Because if you stop for >3 seconds, then scroll again, there will be a stutter.
Maybe only enabling `will-change` when we're scrolled all the way bottom makes more sense?

But this is better than it is now anyways, so it can be merged IMO.

adzialocha (approved):
Great work! 🕵🏻 

I can't "see" or test the differences as my available machines are too fast, but the code looks good.

Simon-Laux (approved):
I have a bad feeling with the timeout, we need to pay attention to it. But let's merge for now and see how it goes.

We can disable the switching logic for high-dpi screens in a follow-up pr, because there you can not see the blurriness anyway

farooqkz:
Nice job there! I was supposed to test these performance issues on me le potato.

WofWca:
I just found out that we're also using `will-change` for the chat list:

https://github.com/deltachat/deltachat-desktop/blob/cdf2b7fccaa2f964297a4610fbafdcad266fdebd/src/renderer/components/chat/ChatList.tsx#L84

https://github.com/bvaughn/react-window/commit/98b6e3da51c4c4348ce1074920784f08b97e8b21

PR #3755, opened 2024-04-08
Save bounds on webxdc close

### LINKED ISSUE #3754: Remember last bounds of webxdc window
It would be nice if DC Desktop would remember the last position and size of the webxdc window.

nicodh:
resolves #3754

### DIFF
--- a/src/main/deltachat/webxdc.ts
+++ b/src/main/deltachat/webxdc.ts
@@ -227,8 +227,6 @@ export default class DCWebxdc extends SplitOut {
           },
           title: makeTitle(webxdcInfo, chatName),
           icon: app_icon || undefined,
-          width: lastBounds.width,
-          height: lastBounds.height,
           alwaysOnTop: main_window?.isAlwaysOnTop(),
           show: false,
         })
@@ -422,7 +420,6 @@ export default class DCWebxdc extends SplitOut {
               [id]: lastBounds,
             },
           })
-          // })
         })
 
         webxdc_windows.once('ready-to-show', () => {})
--- a/src/main/deltachat/webxdc.ts
+++ b/src/main/deltachat/webxdc.ts
@@ -413,11 +413,10 @@ export default class DCWebxdc extends SplitOut {
 
         webxdc_windows.once('close', () => {
           const lastBounds = webxdc_windows.getBounds()
-          const id = `${accountId}.${msg_id}`
           DesktopSettings.update({
             webxdcBounds: {
               ...(DesktopSettings.state.webxdcBounds ?? {}),
-              [id]: lastBounds,
+              [`${accountId}.${msg_id}`]: lastBounds,
             },
           })
         })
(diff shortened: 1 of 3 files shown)


src/main/deltachat/webxdc.ts
```diff
@@ -221,10 +227,15 @@ export default class DCWebxdc extends SplitOut {
           },
           title: makeTitle(webxdcInfo, chatName),
           icon: app_icon || undefined,
-          width: 375,
-          height: 667,
+          width: lastBounds.width,
+          height: lastBounds.height,
```
WofWca, line 231:
```suggestion
```

These can now be removed now, right? [They're optional](https://www.electronjs.org/docs/latest/api/browser-window#new-browserwindowoptions).

src/main/deltachat/webxdc.ts
```diff
@@ -402,6 +413,18 @@ export default class DCWebxdc extends SplitOut {
...
+          const lastBounds = webxdc_windows.getBounds()
+          const id = `${accountId}.${msg_id}`
+          DesktopSettings.update({
+            webxdcBounds: {
+              ...(DesktopSettings.state.webxdcBounds ?? {}),
+              [id]: lastBounds,
+            },
+          })
+          // })
+        })
```
WofWca, line 426:
```suggestion
          })
        })
```

src/main/deltachat/webxdc.ts
```diff
@@ -402,6 +413,18 @@ export default class DCWebxdc extends SplitOut {
           delete open_apps[`${accountId}.${msg_id}`]
         })
 
+        webxdc_windows.once('close', () => {
+          const lastBounds = webxdc_windows.getBounds()
+          const id = `${accountId}.${msg_id}`
```
WofWca, line 418: At this point I'd DRY `${accountId}.${msg_id}`

WofWca (approved):

WofWca:
Nice!

src/main/deltachat/webxdc.ts
```diff
@@ -200,6 +200,12 @@ export default class DCWebxdc extends SplitOut {
 
         const app_icon = icon_blob && nativeImage?.createFromBuffer(icon_blob)
 
+        const lastBounds = DesktopSettings.state.webxdcBounds?.[
+          `${accountId}.${msg_id}`
+        ] ?? {
```
Simon-Laux, line 205: we should think about wether we want to store it in that file or inside of the accounts config table under a config key with the "ui" prefix:
- advantage would be that config.json does not know about message and account ids, same as we moved the selected chat id to a ui config key
- possible disadvantage would be that it's then exported on backups

Simon-Laux: another reason is that I want to keep account specific values out of `config.json`, because my long term plan is to move the values there (global settings) to core account manager config keys (but the system for that does not exist yet) -> basically store the values of `config.json` in `accounts.toml` or a similar file that is managed by core instead of the runtime.

Anyways there is already a mechanism for account specific values with `ui.*` config keys, so let's use them.

Simon-Laux (approved):

src/main/deltachat/webxdc.ts
Simon-Laux: I'll merge this pr and then make a new one that switches it over to store it in core

Issue #3658, opened 2024-02-03
qr code scanner: also accept images from clipboard and scan from them


Simon-Laux:
Just a convenience feature, no need to mention it in the ui.

If you choose the option to paste from clipboard it should do the following:
1. try the text from clipboard as url
2. if that fails and there is an image in the clipboard scan the image and take the first detected qr code.

Why? I accidentally tried to do this multiple times already while not too thinking much, like I miss clicked one time trying to copy the link from the qr code at nine.testrun.org and I copied the image instead. I imagine there are multiple other ways to get into this situation. it is not much effort for us to make this work, so would be cool if it could.

Issue #2634, opened 2022-03-18
Cant select proper camera.


thecodewarrior2:
<!--

- Operating System (Linux/Mac/Windows/iOS/Android): windows
- Delta Chat Version:1.27
- Expected behavior: select working web cam
- Actual behavior: selects first webcam it detects
- Steps to reproduce the problem: scan a qr code
- Screenshots:
- Logs:

My built-in webcam is broken. Cant select an external webcam.

Simon-Laux:
could you elaborate?

Simon-Laux:
do you mean you have multiple cameras and want to select which one to use for the qr code scan?

thecodewarrior2:
Yes.

thecodewarrior2:
if there are multiple webcams installed, there is no way within the
application to select the correct one.

See device manager screenshot. ( attached ) DeltaChat defaults to the
laptops usb 2.0 built-in webcam. This can not be used in the docking
station it has. I have an external usb webcam.  The Logitech one in the
bottom of the screenshot.  The only way to select the Logitech camera is to
disable the usb 2 camera. There is no setting in deltachat to pick the
correct camera. On the Desktop version of deltachat, I would recommend a
dropdown selection option of webcams to pick on the scan qr code screen.

r10s:
as this is part of "QR Scanner Refactor" - is there something planned/possible that goes beyond [MediaTrackConstraints](https://developer.mozilla.org/en-US/docs/Web/API/MediaTrackConstraints) - because  this is also exposed to https://github.com/JodusNodus/react-qr-reader/#component-api and just using this might remove pressure from a  refactoring requirement

r10s:
ftr: i also ran into that an with the continuous camera of macos/iphone, however, at the end it turns out, it was more my fault as the iphone was used for "facingMode: user" - and switching this is part of the system, not even sure if an app has access to that

Simon-Laux:
I also already stated this on a local branch

Issue #1730, opened 2020-06-17
camera stays sometimes on after qr code scan


Simon-Laux:
This is not the expected behavior. The camera should turn itself of always after an qr code scan.
I saw this issue on linux and on mac, so it's likely on windows, too.

Jikstra:
I think it got solved by the overhaul, please reopen if it happens again

Simon-Laux:
still happens sometimes, I think we should take this in our own hands and build our own qr code reader component, also to solve #2634

r10s:
ftr, i cannot reproduce the issue and have not seen the described behaviour for years. was just trying hard to reproduce it on macos ventura, without success 👍 - is there a way to trigger the behaviour? is it about the welcome-screen or the in-app scanner?

r10s:
it happens to me as well now and i found a way to reproduce it on macos: it happens reproducible when you close the scanner dialog _before_ the camera is initialized (so, before the camera image is shown). 

maybe just a a forgotten deinit and no funamental issue with the scanner in use.

https://user-images.githubusercontent.com/9800740/233804598-86c8bd91-24ac-4f51-b49e-076878729050.MOV

Simon-Laux:
still an issue as of #3507

Issue #3599, opened 2023-12-20
QR Code scanner does not unwind properly when prematurely closed


adzialocha:
Closing the QR Scanner Camera too early causes an infinite error loop:

```
Uncaught TypeError: Cannot read properties of null (reading 'srcObject')
    at t.value (index.js:215:19)
```

Steps to reproduce:

1. Open DC Desktop (1.42.2)
2. Open QR Code Scanner Dialog from Sidebar
3. Click "Scan QR Code"
4. Close Dialog directly (don't wait until QR Scanner loaded)

PR #3762, opened 2024-04-09
Refactor QR code reader

adzialocha:
This PR tackles a couple of issues around scanning QR codes:

1. [x] Make `QrReader` an independent React component which can be re-used in other places in the future
2. [x] Use `jsqr` library with inlined WebWorker (before it was wrapped in https://github.com/deltachat/react-qr-reader)
3. [x] Stop camera properly when component gets unmounted (closes https://github.com/deltachat/deltachat-desktop/issues/1730 and https://github.com/deltachat/deltachat-desktop/issues/3599)
4. [x] Add UI to allow selecting a camera and facing (closes https://github.com/deltachat/deltachat-desktop/issues/2634)
5. [x] Bring "Scan from Clipboard" and "Upload image" functionality into UI
6. [x] Accept images from clipboard (closes https://github.com/deltachat/deltachat-desktop/issues/3658)
7. [x] Added a spinner to show something while camera is loading
8. [x] Added an visual error mode when something goes wrong
9. [x] Confirmation dialogs did not fire the callback when pressing ESC-keys https://github.com/deltachat/deltachat-desktop/issues/3081
10. [x] Remove progress bar dialog parsing QR code, it was almost never shown since parsing is too short and makes it look buggy due to the flickering

## Preview

2024-04-16-143404_765x788_scrot
2024-04-16-143513_767x792_scrot
2024-04-16-143648_764x788_scrot


farooqkz:
Hello there. I see you are trying to add this with WASM. Isn't it better to use native? It has better performance for sure. Not that it really matters since QR code scanning is not compute intensive nor the user would scan 100 codes a day. Just asking.

Simon-Laux:
Could make sense to make a hybrid solution of also still using `jsqr` in parallel because there are some cases where quircs does not detect the qr codes.
Also for performance processing the images in a web worker would make sense.

- `jsqr` is what the library, that we currently use, is using for qr code detection/decoding
- `quircs` is a rust port of [`quirc`](https://github.com/dlbeer/quirc) made by dignifiedquire.
   - quircs can detect multiple qr codes in one scan and can also give us the position of the qr code in the image (not that we would use it here but still nice feature, I tried adding a qr scan feature to the gallery with it (highlighted clickable qr codes in fullscreen media view), but it was not good enough in detecting and parsing qr codes (https://github.com/deltachat/deltachat-desktop/pull/3346)
   - I made the wasm wrapper you pulled in, I also experimented with apis to sharpen the image and manipulate it otherwise before passing it with `quircs`, which seemed to improve detection rate on a low res camera, but did not make any noticeable difference when I tested it with a high-res webcam. (currently it tries again after sharpening if it couldn't find a qr code: https://github.com/Simon-Laux/quircs_wasm/blob/0bc826f35b1249211908a8a352145b6ef8f9bfe1/src/lib.rs#L35)
   -  Anyways feel free to also adapt that wasm layer if you need to modify the apis when it is not good enough yet. (https://github.com/Simon-Laux/quircs_wasm/tree/master, I can also move the repo to the deltachat organisation, if we now start using it for real)

Whatever we do we should **not** use the library that we use on android, as that has detection problems sometimes (see https://support.delta.chat/t/error-qr-code-could-not-be-decoded/2923 and https://github.com/deltachat/deltachat-core-rust/issues/5256).

Simon-Laux:
> What was the reasoning to not use zxing?

`jsqr` is just what `react-qr-reader` is using.

dc android uses `zxing`, It is a very versatile library detecting all kinds of codes, there is this report of not being able to detect an image of an qr code that `quircs` can (I linked that report above). Could maybe be that the library got better and we haven't updated it in android?

Better ask some android people if that is still a problem and about their experience with `zxing`.

adzialocha:
> jsqr is just what react-qr-reader is using.

Ah, interesting. This must have been at the point where you've forked it because in upstream it also uses `zxing` now.

The QR library doesn't seem to be the main issue of this refactoring, I'll keep that in the background and maybe just use what we have already in case that's just fine.

adzialocha:
I've integrated `zbar-wasm` now and it seems to work very well!

adzialocha:
I've introduced an Camera and Facing selector to some sort of "Settings" menu which floats over the video image.

How do you feel about adding the functionality for "Paste from Clipboard" and "Upload image" somewhere there as well, either in the same menu or in separate buttons?

This is currently how it looks like (not happy with it), 

2024-04-12-122444_757x774_scrot
2024-04-12-122458_715x274_scrot

How about we have two buttons, one "Camera" icon for selecting Camera and Facing, one "Upload" icon for pasting from clipboard and uploading an image? Or even three separate buttons, one for settings, and then one for clipboard and one for uploading an image?

Pinging @Simon-Laux and @r10s

Simon-Laux:
I imagined it as a rounded square button in the top left corner with white background (in darkmode maybe different bg).
basically a "Floating action button"
I think a single button would be fine if we find the right icon, otherwise we could also do two next or beneath each other. one with a camera icon and the other one with another fitting icon (maybe a generic scanner / view finder icon).
Thinking about it we could also do three buttons, as finding a fitting/intuitive icon for both scan from image file and scan from clipboard feels hard.

Simon-Laux:
> (since invite links) advanced

I don't completely agree with the "advanced" term here yet, the invite links work most of the time on mobile directly, but on desktop the situation is not as good yet, see #3657.

r10s:
> but on desktop the situation is not as good yet, see https://github.com/deltachat/deltachat-desktop/issues/3657

but even then: the browser opens and you tap "Open Chat" with the `OPENPGP4FRP:` scheme - no need to share data via image or clipboard. or did i missed sth and the issue is larger?

to be clear: maybe the term i've chosen is confusing. i meant, that for most users, switching camera is probably needed more often than the other options. and therefore, make them first, and not hidden in a submenu. i do not suggest to hide or remove the paste/load further

for the icon: maybe just stay with the cogwheel then, just make it a bit larger

Simon-Laux:
> or did i missed sth and the issue is larger?

I wouldn't trust it too much there are many platforms (I recently fixed sth about file/url association in the flatpak package for example) and there are also different browsers like people using the tor browser which on purpose does not integrate with the system.

>  i meant, that for most users, switching camera is probably needed more often than the other options

I think it is the opposite, I do think it's not that often that you are a Mac user with connected iPhone, a live-streamed with multiple connected cameras or a linux phone user that managed to install dc desktop on their phone.

Though I still agree with your conclusion that it would be nice to not have an extra menu for the camera

r10s:
thanks for sharing thoughts, very interesting!

tbh, when we discussed the menu _ordering_, i did not expect that this raises a discussion :) so:

note, that the now _recommended workflow_ is scanning QR codes and sharing invite links.

the camera options belong to the recommended workflow. despite tor or broken systems, it is unquestionable also on desktop that load/paste, as _supported fallback_, plays less_a role with the new invite links.

ordering recommending workflow options first makes sense even if we assume for a moment, that there are more ppl using tor or have a broken system than ppl having two webcams [^1]

honestly, these are all weak arguments, but they sum up. and at the end, one has to make a decision to move forward :)

if the decision is wrong, we can correct that easily. again: we're discussing ordering of 4 options, not functionality as such

[^1]: additional webcams are often used for old/broken laptops. or just look at a random office. but also on macos, there is a good share of macos users that have an iphone as well - and get two cameras by that

adzialocha:
I think this is ready for review now, some testing with different devices / cameras etc. would be appreciated! Thank you 💖 

Happy to also work further on the design if there's any ideas.

r10s (commented):
very nice! also the new, direct camera selection

i leave code review to the experts, functionality-wise, i can say it works nicely on my mac.

for the UI, some minors, not blocking or so:

- if possible, add a separator line after the camera selection. in general, it is best-practise to group radio items in menus that way
- i'd sort "Load QR code as image" very last -  these images are not very well supported on the platforms - where "Paste" has a direct pendent already on desktop
- also, the menu flows out of screen when close to the edge - this would not happen easily when using a  a left-hand corner, which is maybe is nicer in general

and when on that: "Hold your camera over the QR code." is often not practical on desktop - when on that, we could add a more fitting string à la "Move the QR code to the camera" - if you agree, i can do a PR on android to get the string in

Simon-Laux:
> if possible, add a separator line after the camera selection. in general, it is best-practise to group radio items in menus that way

our context menu implementation does not support seperators nor radio buttons yet, so we would have to add the functionality there first:
 - seperator would just be a visual element, with logic changes to keyboard navigation to skip separator items when selecting the next context menu item.
 - I think we can do without radio buttons for now as the checkmark emoji is enough for now and we do not have a design concept for radio/checkbox context menu items yet.

> maybe sort "Load QR code as image" very last - these images are not very well supported on the platforms - where "Paste" has a direct pendent already on desktop

I don't care much about the ordering. Though I believe the last and first menu options are the easiest to click quickly, but that will not matter once there is a separator line both items will again be similarly easy to click.

> also, the menu flows out of screen when close to the edge - this would not happen easily when using a a left-hand corner, which is maybe is nicer in general

I still favour the top left corner. maybe with a background to look like a real button.

> and when on that: "Hold your camera over the QR code." is often not practical on desktop - when on that, we could add a more fitting string à la "Move the QR code to the camera" - if you agree, i can do a PR on android to get the string in

+1

_locales/_untranslated_en.json
```diff
@@ -23,5 +23,8 @@
   },
   "pref_current_account": {
     "message": "Current Account"
+  },
+  "camera_access_failed": {
+    "message": "Could not access video camera"
```
Simon-Laux, line 28: maybe add a hint about permissions though I don't know how common real errors vs permission errors are, so maybe not.

r10s:
k, if there is no separator possible currently, then it's that. as said, no blocker or so :)

> I think we can do without radio buttons for now as the checkmark emoji is enough for now

agree. i was referring to "radio buttons" only for functionality, not for design

src/renderer/components/QrReader/index.tsx
```diff
@@ -0,0 +1,447 @@
...
+      {!error && <div className={styles.qrReaderHint}>{tx('qrscan_hint')}</div>}
+      {videoDevices.length > 0 && (
+        <button className={styles.qrReaderButton} onClick={handleSelectDevice}>
+          <Icon
+            icon='settings'
+            size={24}
+            className={styles.qrReaderButtonIcon}
+          />
+        </button>
+      )}
```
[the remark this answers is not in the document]
adzialocha: You mean the file input button? This is to trigger an file open dialog

r10s:
i added a small PR for the string: https://github.com/deltachat/deltachat-android/pull/2992 - no need, however, that this PR gets blocked by that, the string can also be changed on next pull

src/renderer/components/QrReader/index.tsx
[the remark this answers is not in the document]
adzialocha: Do you mean the button for the file input? This is required to trigger opening a file open dialog.

Or do you mean something else?

src/renderer/runtime.ts
```diff
@@ -77,6 +77,7 @@ interface Runtime {
   downloadFile(pathToSource: string, filename: string): Promise<void>
   transformBlobURL(blob: string): string
   readClipboardText(): Promise<string>
+  readClipboardImage(): Promise<string>
```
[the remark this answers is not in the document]
adzialocha: > also is there a `null` or `undefined` case? those should also be expressed in the return type.

Not officially documented at least by Electron 😅 will investigate!

r10s:
one degradation i noticed in comparison with the old QR code implementation:

the QR code is no longer mirrored (it should be when using the "selfie" camera, so probably most times on desktop)

reason is, that once mirrored, the camera act like a mirror, ppl know where well how to move - moving the code left, shows them further left. also, this is what most selfie-cameras are doing

when not mirrored, there is always confusion and placing things in front of camera takes a lot longer

see https://github.com/deltachat/deltachat-desktop/pull/1613 / https://github.com/deltachat/deltachat-desktop/issues/1591

r10s:
> i hope, there is a simple flag in the new lib 🤞

ahh: maybe this is what ominous facing mode is for - in 2020, @nicodh adapted a setting with exactly that name

adzialocha:
> one degradation i noticed in comparison with the old QR code implementation:
> 
> the QR code is no longer mirrored (it should be when using the "selfie" camera, so probably most times on desktop)
> 
> reason is, that once mirrored, the camera act like a mirror, ppl know where well how to move - moving the code left, shows them further left. also, this is what most selfie-cameras are doing
> 
> when not mirrored, there is always confusion and placing things in front of camera takes a lot longer
> 
> see https://github.com/deltachat/deltachat-desktop/pull/1613 / https://github.com/deltachat/deltachat-desktop/issues/1591
> 
> i hope, there is a simple flag in the new lib 🤞

Good point! I think we can simply just apply a css transform on the video element to flip it.

src/renderer/runtime.ts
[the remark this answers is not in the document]
Simon-Laux: ah if there is no image it returns: `'data:image/png;base64,'`, that probably also triggers the bug that I can not scan text from clipboard currently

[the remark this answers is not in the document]
adzialocha: > ah if there is no image it returns: `'data:image/png;base64,'`, that probably also triggers the bug that I can not scan text from clipboard currently

Aah, alright, will fix that

src/renderer/components/QrReader/index.tsx
```diff
@@ -0,0 +1,447 @@
...
+
+  const handleSelectDevice = useCallback(
+    (event: React.MouseEvent<HTMLButtonElement>) => {
+      const [cursorX, cursorY] = [event.clientX, event.clientY]
+
+      // Allow user to select a different camera when more than one is given
+      const cameraItems: ContextMenuItem[] =
+        videoDevices.length > 1
+          ? videoDevices.map(device => {
+              const marker = device.deviceId === deviceId ? ' ✔' : ''
```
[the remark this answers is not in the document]
adzialocha: I can try a non-emoji character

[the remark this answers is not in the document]
r10s: on iOS, we're using ✓ that works in dark mode as well

src/renderer/components/QrReader/index.tsx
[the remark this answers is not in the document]
Simon-Laux: Bildschirmfoto 2024-04-16 um 22 40 15
 looks nearly identical

_locales/_untranslated_en.json
adzialocha: I've added some more advice :+1:

src/renderer/components/QrReader/index.tsx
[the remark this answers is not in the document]
adzialocha: Ah, funny, I didn't spot this one! Thank you, that must have been a copy and paste thing

src/main/ipc.ts
```diff
@@ -210,7 +210,17 @@ export async function init(cwd: string, logHandler: LogHandler) {
     return clipboard.readText()
   })
   ipcMain.handle('electron.clipboard.readImage', () => {
-    return clipboard.readImage().toDataURL()
+    const base64Data = clipboard.readImage().toDataURL()
+
+    // Electron just returns an empty base64 string (for example
+    // 'data:image/png;base64,' when no image was in the clipboard),
+    // we check that here and more conveniently return null instead
```
Simon-Laux, line 217: there is also `image.isEmpty()`: https://www.electronjs.org/de/docs/latest/api/native-image#imageisempty

maybe using that results in clearer/prettier code

adzialocha:
Okay, this is ready again for another round of reviews, from my side all good to merge. Thank you for your inputs!

The issue https://github.com/deltachat/deltachat-desktop/issues/3081 was also now properly tackled. It's a mix of:

1. Input element not firing `onChange` when image data didn't change https://github.com/deltachat/deltachat-desktop/pull/3762/commits/bfd860a01458e0cf9d6e4a80a39203d1fd5bebbd
2. Confirm dialog not firing callback when ESC-key was pressed to close it https://github.com/deltachat/deltachat-desktop/pull/3762/commits/fea3585875289554a22b91e562709705f9235f36

The latter is slightly unrelated to this PR but I fix it here now since it's still QR-code related.

adzialocha:
Hmn, I have issues scanning QR codes with my X1 laptop webcam and Arch Linux 😢 - it works when loading them as an image, but directly from the camera stream it never finds a code. Maybe we have to go back to the old library as I can't figure out where the issue lies ..

src/renderer/components/QrReader/index.tsx
```diff
@@ -0,0 +1,484 @@
...
+import {
+  ZBarConfigType,
+  ZBarScanner,
+  ZBarSymbol,
+  ZBarSymbolType,
+  getDefaultScanner,
+  scanImageData,
+} from '@undecaf/zbar-wasm'
+import classNames from 'classnames'
+import { Spinner } from '@blueprintjs/core'
```
Simon-Laux, line 17: please don't use blueprint in new code

Simon-Laux:
> Maybe we have to go back to the old library as I can't figure out where the issue lies ..

oh no... sounds a lot like my first experience with the bar wasm demo site. We could do the brute-force solution of using both scanner libraries in webwokers (web workers to use multicore cpus better).

adzialocha:
> > Maybe we have to go back to the old library as I can't figure out where the issue lies ..
> 
> oh no... sounds a lot like my first experience with the bar wasm demo site. We could do the brute-force solution of using both scanner libraries in webwokers (web workers to use multicore cpus better).

An older zbar wasm code works funnily in the browser, but the library is bad (`zbar.wasm`) and unusable for us. I guess we'd need our own wasm version of zbar to make it work.

In any case, this new scanner library experiment was fun, but not the idea of this PR, i've switched back to the old `jsqr` one now and at least things shouldn't become _worse_ like that.

src/renderer/components/QrReader/index.tsx
adzialocha: I wouldn't like to write a new spinner as part of this PR, do you have any suggestions what do do instead?

Simon-Laux: take one from https://cssloaders.github.io and mention where it comes from to comply with the MIT licence

src/renderer/components/QrReader/index.tsx
```diff
@@ -0,0 +1,457 @@
...
+      // another `onChange` again when selecting the same image
+      if (inputRef.current) {
+        inputRef.current.value = ''
+      }
+    },
+    [handleError, handleScanResult, tx, userFeedback]
+  )
+
+  // Show a context menu with different video input options to the user.
+  const handleSelectDevice = useCallback(
```
Simon-Laux, line 244: name is confusing as it is not only about device selection, but also the other options

r10s:
i think, the camera is not freed when closing the scanning dialog by tapping  _outside_; this was not the case before (on mac, open camera is very visible by led and an icon in the menu bar)

(i am not totally sure, currently i cannot double check as i cannot build desktop for $reasons currently)

adzialocha:
> i think, the camera is not freed when closing the scanning dialog by tapping _outside_; this was not the case before (on mac, open camera is very visible by led and an icon in the menu bar)
> 
> (i am not totally sure, currently i cannot double check as i cannot build desktop for $reasons currently)

I couldn't reproduce this as it probably was something else than the tapping outside (that just regularly should unwind the component), but looked at the logic again and found some cases where it might have been stuck in edge error cases. I made the logic more robust now.

src/renderer/components/QrReader/index.tsx
adzialocha: Is it okay to do this in a subsequent PR? To me it feels like it's own thing

Simon-Laux: we can do that, if we don't forget it.

src/renderer/components/QrReader/index.tsx
```diff
@@ -0,0 +1,477 @@
...
+  width: number
+  height: number
+}
+
+const SCAN_QR_INTERVAL_MS = 250
+
+const worker = new Worker()
+
+/**
+ * Convert file data to base64 string.
```
Simon-Laux, line 41:
```suggestion
 * Convert file data to base64 encoded data URL string.
```

Simon-Laux (approved):
we can merge this and then do the remaining changes in a follow up pr (namely to remove blueprint usage), I think it makes sense to get more people to test the new scanner and all the serious problems that I found are already solved now.

PR #3769, opened 2024-04-12
restrict file protocol

### LINKED ISSUE #2170: Restrict file protocol

Simon-Laux:
closes #2170


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 - Update translations (2024-04-04) #3746
 - minor improvements to "add second device" dialog #3748
 - Remove deprecated translations #3756
+- Restrict file protocol #3769
```
WofWca, line 14:
```suggestion
- Improve security: restrict file protocol #3769
```

src/main/windows/main.ts
```diff
@@ -158,6 +166,45 @@ export function init(options: { hidden: boolean }) {
       }
     }
   )
+
+  window.webContents.session.webRequest.onBeforeRequest(
+    { urls: ['file://*'] },
+    (details, callback) => {
+      const pathname = decodeURIComponent(new URL(details.url).pathname)
+
+      if (!isAbsolute(pathname) || pathname.indexOf('..') !== -1) {
```
WofWca, line 175: Just a refactor

```suggestion
      if (!isAbsolute(pathname) || pathname.includes('..')) {
```

WofWca (approved):
Just skimmed over. Didn't try to break it, didn't test.
Better than nothing at least.

Issue #3676, opened 2024-02-11
windows: add-second-device issues


r10s:
i recently got the opportunity to watch someone using "add second device" from android to desktop; there are the issues that pop up:

- for whatever reason, the QR code scanner rectangle is larger than it fits on the screen, although the window was full-height. this leads to the what-to.do-hints not being visible:

![image_2024-02-10_19-40-36](https://github.com/deltachat/deltachat-desktop/assets/9800740/56aae44f-91c2-423c-a20a-c60b23dbddb6)

with just that, the user is pretty lost when starting the process from the new device. the hint **"On the first device, go to "Settings / Add Second Device" and scan the code shown there"** is only shown when the user discovers that there can be sth. scrolled

on comparable large mac screens (the window screen was a normal 15~16" thinkpad laptop), the hint is shown, at least in full-height windows (it still disappears when the window is shrinked)

it seems better to never add scrollbars there but to scale the QR code scanner rectangle and make hint being always visible (the overlay in the camera window is nice, btw)

2. in contrast to mac, the camera image is not acting in the same way as a mirror. this leads to sort of intelligence test when moving the qr code around. this is a minor thing, it adds some initial confusion though, where ppl also have to deal with contrast and distance. if it gets complicated to fix that, it is maybe not worth fixing.
 
the detection of the QR code as such is pretty nice and did not cause issues

the used laptop was a standard lenovo one with a standard out-of-the-box windows, nothing tweaked.

finally, there was https://support.delta.chat/t/add-hint-select-on-windows-desktop-private-network-to-add-a-second-device/2921/2 , but i fixed that before and i am working on a PR to get that to the "troubleshooting" (maybe we should get that to the primary windows UI as well, however, with an iroh update, there may also be holepunching, so starting with "troubleshooting" seems fine)

Simon-Laux:
related to the troubleshooting: https://github.com/deltachat/deltachat-desktop/issues/3218

Simon-Laux:
> for whatever reason, the QR code scanner rectangle is larger than it fits on the screen, although the window was full-height. this leads to the what-to.do-hints not being visible:

[...]

PR #3773, opened 2024-04-15
New "Instant Onboarding" UX flow to create accounts

adzialocha:
Wireframes: https://whimsical.com/instant-onboarding-TrFnChnvwpRQqTSqgXoVPw

Related translation PR: https://github.com/deltachat/deltachat-android/pull/3001

**Todo**

- [x] Introduce new "Instant Onboarding" screen creating account from chatmail instance (`nine.testrun.org` as default)
- [x] Move alternative setup methods into separate dialog
- [x] Show selected chatmail instance when coming from `DC_QR_ACCOUNT` QR code
- [x] Remove dialog asking for username and profile image _after_ account got created
- [x] Instant onboarding when not having an account and using `DC_QR_ASK_VERIFYCONTACT` or `DC_QR_ASK_VERIFYGROUP` QR code using secure join afterwards
- [x] Show confirmation dialog _before_ instant onboarding when `DC_QR_ASK_VERIFYCONTACT` or `DC_QR_ASK_VERIFYGROUP` QR code
- [x] Bring back flow when scanning `DC_QR_LOGIN` QR code
- [x] Automatically select created chats after flow finished
- [x] Bring back confirmation dialog when coming from `DC_QR_ACCOUNT` QR code
- [x] Make it all look shiny and nice ✨ 
- [x] Introduce a new `Checkbox` component ✨
- [x] Update translation strings based on https://github.com/deltachat/deltachat-android/pull/3001
- [x] Find a new onboarding image ✨

**Video**

https://github.com/deltachat/deltachat-desktop/assets/1822473/e52391a2-a90a-40e4-b3f1-03b0acae22d3

**Screenshots**

> New "Welcome Screen" with slightly different button designs and new image

2024-04-24-224650_1431x1156_scrot

> Default instant onboarding dialog

2024-04-24-175215_1280x1175_scrot

> Alternative login methods dialog

2024-04-24-175338_1280x1175_scrot

> Scanning a QR code to join a group via secure join. We do not show the link here to see other chatmail instances as we're currently not supporting changing the chatmail instance in secure join QRs

2024-04-24-175416_1280x1175_scrot

> Scanning chatmail QR code

2024-04-24-175203_1190x1064_scrot


src/renderer/hooks/useInstantOnboarding.ts
```diff
@@ -0,0 +1,196 @@
...
+          !['account', 'askVerifyGroup', 'askVerifyContact'].includes(qr.kind)
+        ) {
+          throw new Error(
+            'QR code needs to be of kind `account`, `askVerifyGroup` or `askVerifyContact` for instant onboarding flow'
+          )
+        }
+
+        // Ask user to confirm creating a new account if they already have one
+        const numberOfAccounts = (await BackendRemote.rpc.getAllAccountIds())
+          .length
```
Simon-Laux, line 53: you could also ask if the current account is unconfigured, if it is use it, if it is configured ask to add a new account

src/renderer/hooks/useInstantOnboarding.ts
```diff
@@ -0,0 +1,196 @@
...
+
+      setWelcomeQr(qrWithUrl)
+      setShowInstantOnboarding(true)
+
+      setTimeout(async () => {
+        if (screen !== Screens.Welcome) {
+          // Log out first by switching to an (temporary) blank account
+          const blankAccount = await BackendRemote.rpc.addAccount()
+          window.__selectAccount(blankAccount)
+          changeScreen(Screens.Welcome)
```
Simon-Laux, line 79: selecting an unconfigured account should bring you always into the welcome screen, so no need to change to it again?

src/renderer/hooks/useInstantOnboarding.ts
Simon-Laux: I guess if the account is unconfigured you do not even need to ask the user here.

Simon-Laux:
> Bring back flow when scanning DC_QR_LOGIN QR code (did this ever exist?)

yes it does exist, it is basically setting credentials and advanced options it is documented at https://github.com/deltachat/interface/blob/master/uri-schemes.md#dclogin-

src/renderer/hooks/useInstantOnboarding.ts
adzialocha: Right, I've added it anyhow as react de-deduplicates it and on top it makes the code easier to read, but I can remove it if it's confusing

Simon-Laux: rather remove it and make a comment that says that it automatically goes there, maybe with a link to the file where it continues

src/renderer/global.d.ts
```diff
@@ -29,7 +29,6 @@ declare global {
       | undefined
     __refetchChatlist: undefined | (() => void)
     __welcome_qr: undefined | string
```
Simon-Laux, line 31: makes sense to remove `__welcome_qr` here too if you have moved it already to the new instant onboarding context

src/renderer/components/screens/WelcomeScreen/OnboardingScreen.tsx
```diff
@@ -0,0 +1,116 @@
...
+              onClick={onScanQRCode}
+            >
+              Scan QR code
+            </Button>
+            <Button
+              className={styles.welcomeScreenButton}
+              type='secondary'
+              onClick={onAlreadyHaveAccount}
+            >
+              I already have an account
```
WofWca, line 109: Why not the more straight-forward "create account"/"sign up" and "log in". I know under "log in" there is also "add as second device" and "restore from backup", but that is still "logging in", but with extra stuff.

_locales/_untranslated_en.json
```diff
@@ -23,5 +23,41 @@
   },
   "pref_current_account": {
     "message": "Current Account"
+  },
+  "onboarding_create_instant_account": {
+    "message": "Let's get started!"
```
adbenitez, line 28: I am not sure "let's get started!" fully convey "I don't have account / register"

missytake: What about "Create Account"? I guess people have already considered this wording

r10s: yeah, this is a not-so-easy question.

the issue with "Create Account" (what would be "Create Profile" then) is that ppl need to understand or may ask theirselves what an "Account" or a "Profile" is, at the very first screen already, or if the can get along without? and this even if they never used Delta Chat before and the only thing they know is that someone said "Hey, we can use that for chatting". we're talking about ppl that do not know or care, that an email address will created

therefore, the very simple wording, speaking about nothing special, just "Let me in" - this is also what happens when you install WhatsApp (no verb), Signal (no verb, "Continue"), Telegram (no verb, "Start Messaging") and probably many others

so, if you will, we make things easier to use for ppl who do not know or care, maybe at the cost of ppl knowing that an account will be created. implicitly, however, more experienced ppl get by the "third button" that "sth" is created (which, then, thinking it over  should pick up the wording of the subsequent page, so maybe better "I Already Have a Account/Profile"  - so give the thing only one name, and skip this name where possible)

_locales/_untranslated_en.json
```diff
@@ -23,5 +23,41 @@
...
+    "message": "Manual login"
+  },
+  "instant_onboarding_title": {
+    "message": "Your profile"
+  },
+  "instant_onboarding_create": {
+    "message": "Create profile"
+  },
+  "instant_onboarding_default_info": {
+    "message": "Your profile will be created on the default instance."
```
r10s, line 43: we also thought about using the other wording "Your profile will be created on the following instance: nine.testrun.org" - would have been even easier, as nothing special is done.

however, this will raise questions: what is an instance? what is nine.testrun.org? can i trust that? etc. therefore, unless it seems _really_ needed for ppl "just want to chat with this app" - without knowing or caring instances, chatmail, let alone "nine", it seems advisable to leave that out. the less questions popping up, the better.

and: there is a help button just at the side: we aim to explain things in detail there, for ppl that are interested.

r10s, line 43: if some additional policy is needed at the end, we'll probably add/link that below this text - but that seems to be an open question

_locales/_untranslated_en.json
```diff
@@ -29,5 +29,41 @@
...
     "message": "Could not access video camera. You might want to check your video permissions or if the camera is used already by another program"
+  },
+  "onboarding_create_instant_account": {
+    "message": "Let's get started!"
+  },
+  "onboarding_alternative_logins": {
+    "message": "I already have a login"
+  },
+  "manual_account_setup_option": {
+    "message": "Manual login"
```
r10s, line 40: suggestions from internal chat:

> I think email login under the name "manual Login" it too much hidden / discuoraged, how about calling it "Login with normal/standard email account" or sth like that instead? or "Manual Login with existing email account"

_locales/_untranslated_en.json
r10s: this string will be removed completely, instead a "agree to privacy policy" link will be added above the "create button"

src/renderer/components/screens/WelcomeScreen/chatmailInstances.ts
```diff
@@ -0,0 +1,22 @@
+import type { QrWithUrl } from '../../../backend/qr'
+
+// URL to list of various other chatmail instances
+export const CHATMAIL_INSTANCES_LIST_URL = 'https://delta.chat/chatmail'
```
r10s, line 4: until we have translations ready (requires pushing to transifex, but that's too soon as the page is subject to change) better use the english page explicitly

```suggestion
export const CHATMAIL_INSTANCES_LIST_URL = 'https://delta.chat/en/chatmail'
```

src/renderer/hooks/useInstantOnboarding.ts
```diff
@@ -0,0 +1,144 @@
...
+                  accountId,
+                  'selfavatar',
+                  profilePicture
+                )
+              }
+              await BackendRemote.rpc.setConfig(
+                accountId,
+                'displayname',
+                displayName
+              )
```
Simon-Laux, line 107: > Note: We need to set these profile settings _after_
> calling `configure` to make the UI aware of them already

what part of the UI does not support this? as both of these calls could fail I would prefer if we ran them before configuring (to not end up with a half configured account, or AEAP messages in device messages).

The problems I noticed when moving these two calls before calling configure (in this case opening the `ConfigureProgressDialog`):
- `window.__updateAccountListSidebar?.()` is missing in this onSuccess callback, without it the account list does not update reliably
- Core has a bug where the saved messages chat gets the own account avatar picture instead of the saved messages image, but that is a bug we should file and fix in core
- ... is there anything else that I missed in my testing?

src/renderer/components/screens/WelcomeScreen/UserAgreement.tsx
```diff
@@ -0,0 +1,52 @@
...
+function extractLinkFromQrCode(qrWithUrl: QrWithUrl) {
+  if (qrWithUrl.qr.kind !== 'account') {
+    throw new Error('QR needs to be of kind "account"')
+  }
+
+  const matches = qrWithUrl.url.match(URL_REGEX)
+  if (matches && matches.length > 0) {
+    const { origin } = new URL(matches[0])
+    return origin
+  }
```
Simon-Laux, line 24: what is this logic for? is `.domain` from core not what we want here?

src/renderer/contexts/InstantOnboardingContext.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export const InstantOnboardingContext =
+  React.createContext<InstantOnboardingContextValue | null>(null)
+
+export const InstantOnboardingProvider = ({
+  children,
+}: PropsWithChildren<{}>) => {
+  // Flag to display "Instant Onboarding" UI in welcome screen
+  const [showInstantOnboarding, setShowInstantOnboarding] = useState(false)
```
Simon-Laux, line 20: I can read/understand two different things from the comment:
- instant onboarding option is shown or hidden in welcome screen
- welcome screen gets replaced by onboarding screen

solution idea:
- change var name and/or change comment to clarify what it does

src/renderer/hooks/useInstantOnboarding.ts
```diff
@@ -0,0 +1,144 @@
...
+
+              // 4. If the user created a new account from trying to contact
+              // another user or joining the group we continue with this now
+              let chatId: number | null = null
+              if (welcomeQr) {
+                if (welcomeQr.qr.kind === 'askVerifyContact') {
+                  chatId = await secureJoinContact(accountId, welcomeQr, true)
+                } else if (welcomeQr.qr.kind === 'askVerifyGroup') {
+                  chatId = await secureJoinGroup(accountId, welcomeQr, true)
+                }
```
Simon-Laux, line 117: if we do not accept any qr code here then we should maybe log that we received an unexpected qr type here?
I guess we'll have at least one more case: opening a mailto link with DC before having a configured account.

An error log entry could help the devs in the future to figgure out that they need to add their qr code handling case here, that it works.

src/renderer/hooks/useProcessQr.ts
```diff
@@ -47,43 +67,110 @@ export default function useProcessQR() {
...
+  // a new "chatmail" profile for them and connects them with the regarding
+  // contact or group if given.
+  //
+  // We ask the user if they really want to proceed with this action and
+  // accordingly prepare the onboarding.
+  const startInstantOnboarding = useCallback(
+    async (accountId: number, qrWithUrl: QrWithUrl) => {
+      const { qr } = qrWithUrl
+      const numberOfAccounts = (await BackendRemote.rpc.getAllAccountIds())
+        .length
```
Simon-Laux, line 81: we should rather check if the current account is new (unconfigured), otherwise the message is wrong when you have exactly one configured account and scan the code.

src/renderer/hooks/useProcessQr.ts
```diff
@@ -47,43 +67,110 @@ export default function useProcessQR() {
...
+        if (!userConfirmed) {
+          return
+        }
+      }
+
+      // Start instant onboarding flow for user
+      await switchToInstantOnboarding(qrWithUrl)
+
+      // Log-out user if they already have an account
+      if (screen !== Screens.Welcome) {
```
Simon-Laux, line 129: would check directly if the current account is configured instead of relying on the screen variable.

src/renderer/hooks/useProcessQr.ts
```diff
@@ -47,43 +67,110 @@ export default function useProcessQR() {
...
+          confirmLabel: tx('ok'),
+        })
+
+        if (!userConfirmed) {
+          return
+        }
+      }
+
+      // Start instant onboarding flow for user
+      await switchToInstantOnboarding(qrWithUrl)
```
Simon-Laux, line 126: naming confusion: why not call it like the comment: `startInstantOnboardingFlow`.

Also the comment "for user" is a bit confusing as below a new account could be created, so it's not clear who this "user" is.

src/renderer/hooks/useProcessQr.ts
```diff
@@ -93,103 +180,76 @@ export default function useProcessQR() {
...
+        callback && callback()
+        return
+      }
 
-        if (screen !== Screens.Welcome) {
-          // log out first (not needed anymore as selectAccount does this automatically now)
-          window.__selectAccount(await BackendRemote.rpc.addAccount())
+      // DCLOGIN: Ask the user if they want to login with given credentials
+      if (qr.kind === 'login') {
+        await startInstantOnboarding(accountId, parsed)
```
Simon-Laux, line 193: does not work, also I don't think this needs the instant Onboarding screen, just the dialog asking whether you want to login with those credentials (if account is un-configured ask wether you want to login, if it is configured ask if it should create another account and login there; just as it was previously.)

Simon-Laux:
I think it makes sense to also have a "scan qr code" button inside of the "already have an account" dialog.
If you have an qr code, you might click there to look for the qr code scanner.

Like if someone prints out personalised DCLOGIN qr codes for a school class and hands them out, or a sport club sends out a letter with a personalised dclogin account code to their members.

It is still different to the "add as second device" button, as I plan to customise the scanner dialog for that option to address the missing points from #3178 and #3676 (add instructions and trouble shooting button and disallow other qr code types).

src/renderer/hooks/useInstantOnboarding.ts
adzialocha: > what part of the UI does not support this?

That was my question as well. It just didn't work but I couldn't find out yet why, thanks for investigating.

src/renderer/components/screens/WelcomeScreen/UserAgreement.tsx
adzialocha: The core unfortunately doesn't give us the URL schema

src/renderer/contexts/InstantOnboardingContext.tsx
adzialocha: > I can read/understand two different things from the comment:
> 
>     * instant onboarding option is shown or hidden in welcome screen
>     * welcome screen gets replaced by onboarding screen

This sounds to me like an implementation detail this state variable doesn't need to know about ..

src/renderer/hooks/useInstantOnboarding.ts
adzialocha: > if we do not accept any qr code here then we should maybe log that we received an unexpected qr type here?

If I understand your question then this is already taken care of in other places. See `switchToInstantOnboarding` and the `useProcessQR` logic

src/renderer/hooks/useProcessQr.ts
adzialocha: This is mostly copied from the old codebase, can change it ofc

src/renderer/components/screens/WelcomeScreen/OnboardingScreen.tsx
r10s: let's leave wording/flow discussions outside this PR for now, even if things are not 100% perfect, we need to move forward

_locales/_untranslated_en.json
r10s: let's leave wording/flow discussions outside this PR for now, even if things are not 100% perfect, we need to move forward

src/renderer/hooks/useInstantOnboarding.ts
adzialocha: Filed the core issue here: https://github.com/deltachat/deltachat-core-rust/issues/5509

src/renderer/hooks/useProcessQr.ts
adzialocha: I brought back the original code now, that should hopefully reflect the correct flow now for `DCLOGIN`

_locales/_untranslated_en.json
r10s: let's leave wording/flow discussions outside this PR for now, even if things are not 100% perfect, we need to move forward

src/renderer/components/screens/WelcomeScreen/UserAgreement.tsx
Simon-Laux: I think https is safe to assume and require here

src/renderer/hooks/useInstantOnboarding.ts
Simon-Laux: I think it still makes sense to log the unexpected qr code as error or warning to help future devs.

Simon-Laux (changes requested):
DCLOGIN code does AEAP now when run from an configured account it changes your email credentials after showing a dialog that promises you that your existing account will stay as it is.

thoughts for future:
But thinking about that a bit we should probably show a custom dialog in this case, you scanned an account qr code to you want move your account to this new address or do you want to login to a new account. But allowing AEAP to do with these codes is probably nothing that we should suggest to users as long as AEAP is still considered experimental

src/renderer/hooks/useProcessQr.ts
Simon-Laux: we could also have the onboarding screen, sounds like @r10s did that for iOS and android?
Anyways however the UI flow, the main thing that I care about is that it works nicely and at all in both cases (scanning from unconfigured account and scanning from configured account).

src/renderer/hooks/useProcessQr.ts
```diff
@@ -138,6 +139,48 @@ export default function useProcessQR() {
...
+  // We ask the user if they really want to proceed with this action and log
+  // them in automatically.
+  const startLogin = useCallback(
+    async (accountId: number, qrWithUrl: QrWithUrl) => {
+      const { qr, url } = qrWithUrl
+      if (qr.kind !== 'login') {
+        throw new Error('QR code needs to be of kind "login"')
+      }
+
+      const numberOfAccounts = (await getConfiguredAccounts()).length
```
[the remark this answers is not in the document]
adzialocha: > Or is there some good logical reason for making it based on account count that I missed?

The method returns a list of all configured accounts, if it's length is `0` then it means there's no configured account as you say.

adzialocha:
> DCLOGIN code does AEAP now when run from an configured account it changes your email credentials after showing a dialog that promises you that your existing account will stay as it is.

This should be fixed now by making sure the user is logged out

src/renderer/contexts/InstantOnboardingContext.tsx
adzialocha: I find it hard to come up with something else here without any clear suggestion _how_ you understand this part of code - ("misunderstandings" are well .. misunderstandings, so any input would be helpful to improve). Feel free, to do the changes directly.

My nitpick take on your comment: This component doesn't need to be super-aware of _how_ it is used in the UI, it just needs to handle this flag which makes something visible or not.

src/renderer/components/screens/WelcomeScreen/UserAgreement.tsx
adzialocha: Since this code is already working and fine I wouldn't change it further

src/renderer/hooks/useProcessQr.ts
adzialocha: Similar to my other comment, not sure if I can follow, feel free to do changes directly

r10s (commented):
functionality-wise this lgtm, i tested quite a bit - very nice, what an improvement, thanks a lot 🤘

there may be issues and cornercases we did not encounter yet - however, esp. for that i would merge more sooner than later, so that we can do beta-versions and testing - and have smaller PRs to review then

code-level-wise - to me, it looks perfectly fine - also due to iterations and suggestions from react-experts  - but well, i cannot really judge as i am too far away from react & co

adbenitez:
> Default instant onboarding dialog

what happens if you click the scan QR code here after having setting name and avatar??? will desktop remember this? I assume it will then show the "set profile name and image" again after the QR is scanned and account is created, this could cause confusion and frustration for an user that doesn't onboard in the default/main instance: "I already set name and avatar and now get asked again and I lost the configured data!?"

r10s:
>  will desktop remember this? 

yes, it is remembered, just tested.

we can achieve the same probably relatively easily on android/ios by creating an unconfigured context only as needed (so, reusing it when the Activity is reopened)

adzialocha:
> EDIT: but i noticed that avatar-selection in general stops working, there seems to be sth lost on recent changes :) maybe related to [deltachat/deltachat-core-rust#5509](https://github.com/deltachat/deltachat-core-rust/issues/5509) ? (if it requires core, we can still merge this PR and file an blocking issue for that) - but the flow is perfectly fine, the broken avatar is preserved when switching instances :)

I've made an issue here: https://github.com/deltachat/deltachat-desktop/issues/3797

src/renderer/hooks/useInstantOnboarding.ts
Simon-Laux: Doing it with typescript is an even better idea, thanks. I made a commit to make it fail when the case is not handled: be0c1a6e9ed03d43a10016330224c6c26c73b5e6

src/renderer/contexts/InstantOnboardingContext.tsx
Simon-Laux: I changed the comment in 6919869875642558c42106bbdd50a1a77aef0fe9

r10s:
yeah, first platform with instant onboarding! congrats! 🎉👯‍♀️👯‍♂️

PR #3780, opened 2024-04-18
Update `deltachat-node` and `deltachat/jsonrpc-client` to `v1.137.3`

Simon-Laux:


hpk42:
is  the issue with the windows test failure related to this PR?

Simon-Laux:
about windows ci failure:
- blueprint npm package (failed to remove directory dureing install)
- it builds core instead of using the prebuild, I don't think the runner switched cpu architecture, so must be the blueprint thing disrupted the prebuild step
- can not find the code formatter prettier, which indicates an incomplete/broken npm install process

This will make me put "removing blueprintjs" on a higher priority. I restarted the ci, let's see if it is only flaky, seems to be unrelated to dc node.

Simon-Laux:
windows ci Issue still persists, still blueprint. still unrelated to the dc node update. so ignoring for now.

Issue #3776, opened 2024-04-16
Switching accounts re-renders everything visibly including sidebar


Simon-Laux:
account sidebar should not re-render completely when switching account:
- it should keep scroll position
- accounts should not pop up again
- unread badges should not be hidden and pop up again

PR #3787, opened 2024-04-24
Prevent re-rendering of all components on account id change

adzialocha:
Closes https://github.com/deltachat/deltachat-desktop/issues/3776


Simon-Laux:
Trying this pr I found caching issues, I made a smaller pr that fixes the issue and does not have the caching issues. (seems we still need to use the react key - hack to force pre-rendering of some components, I basically moved it down to not include the sidebar anymore)

- https://github.com/deltachat/deltachat-desktop/pull/3789

PR #3789, opened 2024-04-24
Prevent re-rendering of account sidebar when switching account

### LINKED ISSUE #3776: Switching accounts re-renders everything visibly including sidebar

Simon-Laux:
Closes #3776

Alternative to #3787 (which showed caching issues when switching accounts)

The only thing that is a bit annoying in terms of flickering now are the gallery tabs (when the experimental feature location streaming is enabled), though that is only minor, we could solve that by splitting the setting store up in one that caches account setting and another one that caches config.json/DesktopSettings, the latter stores the experimental option for location streaming and does not need to be reloaded everytime the account id changes.

### DIFF
--- a/src/renderer/ScreenController.tsx
+++ b/src/renderer/ScreenController.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { createRef } from 'react'
 import { Component } from 'react'
 import { DcEventType } from '@deltachat/jsonrpc-client'
 import { debounce } from 'debounce'
@@ -19,7 +19,7 @@ import SettingsStoreInstance from './stores/settings'
 import { NoAccountSelectedScreen } from './components/screens/NoAccountSelectedScreen/NoAccountSelectedScreen'
 import AccountDeletionScreen from './components/screens/AccountDeletionScreen/AccountDeletionScreen'
 import RuntimeAdapter from './components/RuntimeAdapter'
-import { ChatProvider } from './contexts/ChatContext'
+import { ChatProvider, UnselectChat } from './contexts/ChatContext'
 import { ContextMenuProvider } from './contexts/ContextMenuContext'
 import { InstantOnboardingProvider } from './contexts/InstantOnboardingContext'
 
@@ -43,6 +43,7 @@ export default class ScreenController extends Component {
   state: { message: userFeedback | false; screen: Screens }
   selectedAccountId: number | undefined
   lastAccountBeforeAddingNewAccount: number | null = null
+  unselectChatRef = createRef<UnselectChat>()
 
   constructor(public props: {}) {
     super(props)
@@ -130,6 +131,9 @@ export default class ScreenController extends Component {
     if (this.selectedAccountId === undefined) {
       return
     }
+
+    this.unselectChatRef.current?.()
+
     const previousAccountId = this.selectedAccountId
 
     SettingsStoreInstance.effect.clear()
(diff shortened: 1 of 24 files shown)


adzialocha:
This might cause the chat state to be temporarily invalid, right? I can see how switching the account id will not reset the chat id to `undefined`. Ideally we would call `unselectChat` whenever we unselect an account (similar to how I did it in my other PR https://github.com/deltachat/deltachat-desktop/pull/3787)

Simon-Laux:
I used a ref instead of refactoring and moving the component up.

WofWca (commented):
Ehh, looks a little fragile. We reset all state of `ChatProvider` with `unselectChat`, but `<ContextMenuProvider>`, `<DialogContextProvider>` etc also have their state that could become invalid, don't they?

WofWca:
Maybe just keeping the scroll position is good enough for now? With [`getSnapshotBeforeUpdate`](https://react.dev/reference/react/Component#getsnapshotbeforeupdate)

src/renderer/ScreenController.tsx
```diff
@@ -129,6 +130,9 @@ export default class ScreenController extends Component {
     if (this.selectedAccountId === undefined) {
       return
     }
+
+    this.unselectChatRef.current?.()
```
farooqkz, line 134: What does this mean here? Looks weird to me :)

WofWca: The function is set from inside [`ChatContext`](https://github.com/deltachat/deltachat-desktop/pull/3789/files#diff-e9605783d283da9708373aa54111f85308a01ee55595a58be274bb20f34000a7). Read it as `chatContext.unselectChat()`

Simon-Laux:
> `<ContextMenuProvider>`, `<DialogContextProvider>` etc also have their state that could become invalid, don't they?

I guess yes in theory: a dialog like settings could still be open when you switch account from a notification that was for another account.

> Maybe just keeping the scroll position is good enough for now? With [getSnapshotBeforeUpdate](https://react.dev/reference/react/Component#getsnapshotbeforeupdate)

no, because:
1. the sidebar is a functional component not a class component
2. the react "key" hack forces complete re-rendering, so this won't help as the component does not exist anymore
3. it would still flicker which is annoying.

Simon-Laux:
The only way to switch accounts without clicking in the interface (which means no context menu / dialog can be open at the same time), is to click on a notification that is for another account (I just added a commit to close all dialogs in that case).

(At least to according to my current knowledge.)

src/renderer/ScreenController.tsx
adzialocha: An alternative to doing it like that is to introduce the `unselectChat` as a prop to the classic class-style React component:

https://github.com/deltachat/deltachat-desktop/blob/adz/i3776-fix-rerendering/src/renderer/ScreenController.tsx#L330-L337

WofWca:
This came up in https://github.com/deltachat/deltachat-desktop/pull/5824, i.e. removing React warnings. The one remaining is here, "cannot access refs during render":

https://github.com/deltachat/deltachat-desktop/blob/973d33b30fb879c1b5ac1f7ee0c2db831e353aa7/packages/frontend/src/contexts/ChatContext.tsx#L172

I tried to take a look at this again.
Apparently React re-creates not only the context/component with `key`, but all its children, even if they're defined somewhere above, and not inside the `key`ed component, which is TBH a little weird to me, but maybe I'm just missing something.

Things that I considered:
- Apply `key` to `ChatContext` again, but also move the context such that the accounts sidebar is not inside of the `ChatContext`. However, accounts sidebar relies on `useChat()`, so some more refactoring would be needed.
- Move the `unselectChat` to inside of `ChatContext`. Declare that `ChatContext` can handle account switching.

But maybe simply ignoring the warning would work for now.

PR #3791, opened 2024-04-25
remove disabled composer reason, now composer is just always hidden when `chat.canSend` is `false`

### LINKED ISSUE #3790: message when following an invitelink is misleading
when following an invitelink, or when scanning a qr code, the chat shows a hint saying "To join, ask another member" - even though this was just done or one cannot know members as one followed an invitelink

i assume, the idea of the hint is when one left a group. however, with invitelines etc. nowadays it does more harm than help and i suggest to just remove it (similar to https://github.com/deltachat/deltachat-desktop/issues/3511) - it also clutters UI, so refining is not only more work, but also does not necessarily help more

we also add hints as info messages, these are more on-point; ftr android/ios were never showing such hints

Screenshot 2024-04-24 at 22 53 25

Simon-Laux:
closes #3790


r10s (approved):
lgtm, functionality wise.

simon and me just had a chat and agreed that it is better to remove it completely for now than to show misleading (as in #3790) or wrong texts (currently mailinglists say we do not support writing)  
in theory, the latter could have been changed easily, but we just never came to it, so, it is also a maintenance problem.

after the next release, we can think over if we want an api that brings the "reason bar" back for situations where the info-messages are not fully sufficient - these are only read-only-mailinglists and maybe group-left (the latter is somehow clear by info-message). one-to-ones will get detailed info-messages by https://github.com/deltachat/deltachat-core-rust/issues/5475

EDIT: for read-only-mailinglists, an easy and less-cluttering alternative is to show them in the title bar as "Read-Only Mailinglist" or "Mailinglist (read only)" so (it is better to not show the "disabled bar" permanently during normal usage)  
this is the place where we also otherwise show "info" of the chat, eg. for mailinglists on desktop, the address is shown there already. broadcast-lists, which are very similar to read-only-mailinglists, are annouced there as well.
(maybe a minor, but still: by that, we also have less complexity in the message-input-area on all platforms)

however, these are all cornercases, it is fine to remove it completely now. thanks for fixing

PR #3792, opened 2024-04-25
investigate preview build ci failure

Simon-Laux:


Simon-Laux:
- so windows has problems deleting files while doing "npm install"
- and macOS runner is suddenly an arm/M1 runner which makes the action fail (I guess output folder is different)

Simon-Laux:
windows was fixed meanwhile, macOS seems to still have an error with the output filename, but let's figure that out in a new pr

PR #3793, opened 2024-04-26
fix Clicking notification does not bring Delta Chat to foreground on Windows

### LINKED ISSUE #3764: Clicking notification doesn't bring Delta Chat to foreground on Windows
- Operating System (Linux/Mac/Windows/iOS/Android): Windows
- Delta Chat Version: 1.44.0
- Expected behavior: The main Delta Chat window is brought to foreground (focused?) when a notification is clicked
- Actual behavior: The account and the chat get switched and the icon in the task bar gets highlighted, but the window stays on the foreground
    1. Open Delta Chat
    2. Open another window and focus is. _DO NOT_ close the Delta Chat window (to minimize it to tray, because then it works fine)
    3. Wait for a new message to arrive (you can one yourself from a second account)
    4. Click the notification
- Screenshots:
- Logs:

The code is here, I believe

https://github.com/deltachat/deltachat-desktop/blob/631d60354ea06654d98ee976c1436dd4e048124d/src/main/notifications.ts#L63-L68

Related:
- Similar: #3609
- A bunch of workarounds: #2867

Simon-Laux:
fixes #3764

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,7 @@
 - fix: clear notifications for a contact request when blocking it #3268
 - Unmount QR scanner and disable camera correctly on abort or exit #3762
 - Close reactions bar on emoji selection #3788
-- fix Clicking notification does not bring Delta Chat to foreground on Windows 
+- fix Clicking notification does not bring Delta Chat to foreground on Windows #3793
 
 <a id="1_44_1"></a>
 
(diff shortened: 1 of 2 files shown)


Simon-Laux:
maybe this will also fix it for linux/gnome, I only tested it on windows and macOS

WofWca (approved):
Unfortunately I don't have a Windows dev setup. And I don't know if this is the "proper" fix, but hopefully it's better.

PR #3795, opened 2024-04-29
change importing location of constants to jsonrpc client

Simon-Laux:
I don't think this mini change of import location in one file is worth a changelog entry


Simon-Laux:
> Windows and macOS build failed

windows prebuild is missing in dc-node and on macOS there are currently issues because GitHub's runner switched to arm. we can  investigate it further https://github.com/deltachat/deltachat-desktop/pull/3792

PR #3796, opened 2024-04-30
Refactoring for ESM module support

link2xt:
This is an attempt to make https://github.com/deltachat/deltachat-core-rust/pull/4971 works.
It works without ESM modules currently, but with ESM module PR there are still problems:
```
Error [ERR_REQUIRE_ESM]: require() of ES Module /home/user/src/deltachat/deltachat-core-rust/node/dist/index.js from /home/user/src/deltachat/deltachat-desktop/tsc-dist/main/deltachat/controller.js not supported.
Instead change the require of index.js in /home/user/src/deltachat/deltachat-desktop/tsc-dist/main/deltachat/controller.js to a dynamic import() which is available in all CommonJS modules.
    at c._load (node:electron/js2c/node_init:2:13672)
    at Object.<anonymous> (/home/user/src/deltachat/deltachat-desktop/tsc-dist/main/deltachat/controller.js:39:39)
    at c._load (node:electron/js2c/node_init:2:13672)
    at Object.<anonymous> (/home/user/src/deltachat/deltachat-desktop/tsc-dist/main/ipc.js:56:38)
    at c._load (node:electron/js2c/node_init:2:13672)
    at Object.<anonymous> (/home/user/src/deltachat/deltachat-desktop/tsc-dist/main/index.js:134:26)
    at c._load (node:electron/js2c/node_init:2:13672)
    at Object.<anonymous> (/home/user/src/deltachat/deltachat-desktop/index.js:11:1)
    at c._load (node:electron/js2c/node_init:2:13672)
```

I don't understand where this `require` comes from. Maybe it is because `main` module is a CommonJS module:
https://github.com/deltachat/deltachat-desktop/blob/e603178e5eae0f933138e3e7cbfe399bf70f3c06/src/main/tsconfig.json#L6

### DIFF
--- a/src/main/ipc.ts
+++ b/src/main/ipc.ts
@@ -8,7 +8,7 @@ import {
   shell,
 } from 'electron'
 import { getLogger } from '../shared/logger'
-import { getDraftTempDir, getLogsPath } from './application-constants'
+import { getDraftTempDir } from './application-constants'
 import { LogHandler } from './log-handler'
 import { ExtendedAppMainProcess } from './types'
 import * as mainWindow from './windows/main'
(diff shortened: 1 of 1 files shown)


Simon-Laux:
the point of the dynamic require/import is that we can present the errors (when native module is not found) to the user in a dialog without crashing the whole app immediately.

Simon-Laux:
> why would deltachat-node be missing and not some other module?

because it is the only native dependency

PR #3800, opened 2024-04-30
whitelist files chosen by file chooser

### LINKED ISSUE #3797: Selected image can not be displayed when src is `file://`
Selecting an image via `runtime.showOpenFileDialog` and then displaying the file using the selected path leads to the following error:

```
GET file:///<path_to_image>.png net::ERR_BLOCKED_BY_CLIENT
```

This seems to be related to the changes made in https://github.com/deltachat/deltachat-desktop/commit/5dfd93798889414e9a1ca84aa73740838265c1ee 

I assume we have to base64 encode the file contents and display this in the element instead to comply with the new restrictions.

Simon-Laux:
closes #3797

### DIFF
--- a/src/main/windows/main.ts
+++ b/src/main/windows/main.ts
@@ -26,7 +26,13 @@ import { fileURLToPath } from 'url'
 
 const log = getLogger('main/mainWindow')
 
-export let window: (BrowserWindow & { hidden?: boolean }) | null = null
+type ExtendedBrowserWindow = BrowserWindow & {
+  hidden?: boolean
+  /** whitelist of file paths that user selected that should the ui be able to also load via the file:/// scheme */
+  filePathWhiteList: string[]
+}
+
+export let window: ExtendedBrowserWindow | null = null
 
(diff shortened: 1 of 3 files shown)


maxphilippov (approved):
Looks good to me, tested this on Windows

src/main/windows/main.ts
```diff
@@ -26,7 +26,13 @@ import { fileURLToPath } from 'url'
 
 const log = getLogger('main/mainWindow')
 
-export let window: (BrowserWindow & { hidden?: boolean }) | null = null
+type ExtendedBrowserWindow = BrowserWindow & {
+  hidden?: boolean
+  /** whitelist of file paths that user selected that should the ui be able to also load via the file:/// scheme */
```
WofWca, line 31:
```suggestion
  /**
  * whitelist of file paths that user selected that the UI should be able to also load via the file:/// scheme
  * example: when changing avatar, we need to display the selected image before it is uploaded to core
  */
```

WofWca (approved):

PR #3801, opened 2024-05-01
follow up pr to instant onboarding  & add as second device scanner improvements

### LINKED ISSUE #3676: windows: add-second-device issues

### LINKED ISSUE #3799: DCLOGIN broken when logged in
It does not work when you scan a dclogin code in a configured account.

This bug was introduced by merging #3773

### LINKED ISSUE #3803: Help does not open for missing languages
actual: help does not open

expected: help should open in the English version

error:
```
node:16376) UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory, stat '~/Coding/deltachat-desktop/html-dist/help/nl_NL/help.html'
(node:16376) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
```

Simon-Laux:
- fixes #3799
- also add steps and troubleshooting link to " add as second device qr code scan" dialog #3178
- fix help not opening for languages that have no localised help

closes #3676
closes #3803


r10s:
thanks a lot for taking care! ❤️

some high-level comments:

- it seems fine to regard the "My Profile" screen as a confirmation dialog, indeed, nothing has happend if the user taps "back". thanks for thinking that over!

- the new "troubleshooting" button is fine, indeed, as you said, "more options" is gone, so it's better to have that directly there

- for the new anchor [#multiclient-troubleshooting](https://delta.chat/en/help#troubleshooting) vs. [#multiclient](https://delta.chat/en/help#multiclient), i do not think it is _that_ clear, which is better, regarding the current help. if you do not know much about "add second device" at all, a little introduction is needed, and counts to troubleshooting as well, even if the headline is not literally correct,  the "Troubleshooting" immediately follows. maybe we should have named the button just "Help" and/or organize the paragraphs differently.  
for the sake of ease and consistency, however, for now, let's stay with the anchor and the button label. at some point we will iterate over the section anyways, no need to change things before (also as adding anchors adds noise to translations etc.)  
esp, as there was just nothing before :)

- for the "add second device" screen, i'd unclutter the above area as follows:

Simon-Laux:
Sure the troubleshooting section is visible, but not that much, otherwise I wouldn't have raised it as an issue:
Bildschirmfoto 2024-05-02 um 00 13 20

Though we can say that people looking for it will surely find it (provided the other sections above it do not grow too much (content updates and. other languages might need less or more characters/space to express the concepts).

About renaming to "help", I wouldn't do that because of the following reasoning: To give some stereotype, some group of people try around stuff and would never "read the manual" first. I have the feeling that the word "troubleshooting" is less of a problem than admitting to needing "help" for those people. (I personally also have that tendency of trying out things before reading the manual, though I don't know if that is an effect of genetics/biology, my soul, the society/patriarchy or whatever).
But my position on this aspect is debatable, you could also argue that more people want assistant to understand what second device setup is at all, though I doubt that many people would stumble across that dialog by accident.

CHANGELOG.md
```diff
@@ -9,7 +9,8 @@
 - add quick key `Cmd+W`/`Ctrl+W` to close webxdc-, html_email- and help-window #3770 #3778
 - Accept images from clipboard in QR reader #3762
 - Introduce new `Spinner` component #3786
-- Instant Onboarding #3773
+- Instant Onboarding #3773 #3801
+- Add steps to add as second device qr code scan dialog #3801
```
r10s, line 13: this is hard to understand, maybe the following:

```suggestion
- Add instructions and troubleshooting button to "Add as Second Device" dialog #3801
```

src/renderer/hooks/useInstantOnboarding.ts
```diff
@@ -81,22 +74,33 @@ export default function useInstantOnboarding(): InstantOnboarding {
...
-      // Use custom chatmail instance if given by QR code
-      if (welcomeQr && welcomeQr.qr.kind === 'account') {
-        instanceUrl = welcomeQr.url
+      let configurationQR = `dcaccount:${DEFAULT_CHATMAIL_QR_URL}`
+
+      if (welcomeQr) {
+        // Use custom chatmail instance if given by QR code
+        if (welcomeQr.qr.kind === 'account') {
+          // 1. In this "Instant Onboarding" account creation flow the user is not
+          // asked to manually insert any mail server credentials.
```
r10s, line 83:
```suggestion
          // asked to manually enter any mail server credentials.
```

r10s (approved):
yeah, this is very nice improvement wrt layout and functionality now 👌

code lgtm - as far as i can tell :)

PR #3810, opened 2024-05-07
Add map as internal webxdc widget

nicodh:
Depends on https://github.com/deltachat/deltachat-core-rust/pull/5541 !!

Depends on non public package @deltachat/stdio-rpc-server !!

Added in chat view and as global map (new)

MR to stdio branch, since the new methods 
_setWebxdcIntegration
initWebxdcIntegration_
do not exist in deltachat-node

tbd: naming of Map Widgets
The title of the Map popup is relevant as it is possible to open multiple wirndows for multiple accounts
We have now:
Maps - Global map for all chats (Account 'AccountName')
Maps - Map for chat 'ChatName'


Simon-Laux:
> MR to stdio branch, since the new methods
setWebxdcIntegration
initWebxdcIntegration
do not exist in deltachat-node

you could also still have used locally compiled `deltachat-node`

nicodh:
Since https://github.com/deltachat/deltachat-core-rust/pull/5541 is in core now, this can be tested now by following the steps "**Use a local core git checkout**" described in https://github.com/deltachat/deltachat-desktop/blob/master/docs/UPDATE_CORE.md

nicodh:
Moved to #3836

PR #3811, opened 2024-05-08
Adding icons to the context menu

maxphilippov:
- add icon to ContextMenuItem and render those
- use our ContextMenu for AttachmentMenu


src/renderer/components/composer/menuAttachment.tsx
```diff
@@ -111,59 +102,50 @@ export default function MenuAttachment({
...
-      {itemsArray.map((item: MenuAttachmentItemObject) => (
-        <MenuItem
-          key={item.id}
-          icon={item.icon}
-          text={item.text}
-          onClick={item.onClick}
-        />
-      ))}
-    </>
+    <Button
```
Simon-Laux, line 144: use a normal button element here and adjust styling with the `attachment-button` class.

we want to get rid of blueprintjs and it currently looks bad in dark theme:
Bildschirmfoto 2024-05-10 um 04 15 42

(I know this is just a draft, but wanted to give that feedback already anyway)

farooqkz:
Nice doing here!

src/renderer/components/composer/menuAttachment.tsx
maxphilippov: Pushed an update for this, there only paperclip styling left (change color and make it a bit smaller).

I made copies of image.svg for icons used before ('document', 'media', 'phone'), should I just copy old svg data there or we need to make new ones?

Simon-Laux: just download icons from https://fonts.google.com/icons that look similar / have the same message if there are no matching icons in `images/icons` yet. `images/icons` is the folder for the icons.

Simon-Laux: I think we should not keep using the blueprint icons unless we want to specify that they have a different license than the google/material-design ones.

src/renderer/components/composer/menuAttachment.tsx
```diff
@@ -111,59 +102,48 @@ export default function MenuAttachment({
...
 
   // item array used to populate menu
-  const items: MenuAttachmentItemObject[] = [
-    settings?.settings.webrtc_instance && {
-      id: 0,
-      icon: 'phone' as IconName,
-      text: tx('videochat'),
-      onClick: onVideoChat,
+  const menu: (ContextMenuItem | false)[] = [
+    settings?.settings.webrtc_instance !== undefined && {
```
Simon-Laux, line 106: Seems like this needs to check for empty string

nicodh:
Looks good to me, except for the irregular margins of menu items (which is not related to this PR)
I can provide styles for the left example if wanted

attachment-menu-comparison

nicodh (approved):

maxphilippov:
> Looks good to me, except for the irregular margins of menu items (which is not related to this PR) I can provide styles for the left example if wanted
I'm up for that, can change this today

PR #3817, opened 2024-05-12
Fix the bug where new message arriving makes the reactions bar and dialogue closed

farooqkz:
The bug was with scrolling. On scroll, the reactions bar closes. I'm preventing scrolling to bottom when the reactions bar is open/shown.

### DIFF
--- a/src/renderer/components/message/MessageList.tsx
+++ b/src/renderer/components/message/MessageList.tsx
@@ -120,7 +120,7 @@ export default function MessageList({ accountId, chat, refComposer }: Props) {
     fetchMoreBottom,
     fetchMoreTop,
   } = useMessageList(accountId, chat.id)
-  const { hideReactionsBar } = useReactionsBar()
+  const { hideReactionsBar, isReactionsBarShown } = useReactionsBar()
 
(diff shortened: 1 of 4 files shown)


Simon-Laux (approved):
fixes the issue

src/renderer/components/message/MessageList.tsx
```diff
@@ -236,20 +236,20 @@ export default function MessageList({ accountId, chat, refComposer }: Props) {
...
-        Event?.stopPropagation()
+        evt?.preventDefault()
+        evt?.stopPropagation()
         return false
       } else if (distanceToTop < 200) {
         log.debug('onScroll: Scrolled to top, fetching more messages!')
         setTimeout(() => fetchMoreTop(), 0)
-        Event?.preventDefault()
-        Event?.stopPropagation()
+        evt?.preventDefault()
```
Simon-Laux, line 245: on other places we use only `ev` as abbreviation.

farooqkz: okay so I will change it to `ev` here as well.

PR #3821, opened 2024-05-15
fix open about dialog via menu

### LINKED ISSUE #3820: "About" menu entry opens "Settings" instead of "About" dialog
both "About" menu entries (macos, recent master) open the "Settings" dialog instead of the "About" dialog.

in 1.42, "About" was opened.

not a biggie, but still sth. that gives a wrong impression :)

Simon-Laux:
no changelog entry because it was introduced in this unreased version.
fixes #3820


r10s (approved):
lgtm, thanks for fixing!

PR #3823, opened 2024-05-15
hide irrelevant advanced settings for chatmail users

### LINKED ISSUE #3819: remove some advanced options if `is_chatmail=1`
recent core has a new config-value `dc_get_config(ctx, "is_chatmail")` that returns `1` in case the profile is a chatmail profile.

in case of chatmail profiles, we want to hide the following options from "Settings / Advanced":

- Prefer End-To-End-Encryption
- Watch Sent Folder
- Send Copy to Self
- Move Automatically to DeltaChat Folder
- Only Fetch from DeltaChat Folder

counterpart of https://github.com/deltachat/deltachat-android/pull/3050 and https://github.com/deltachat/deltachat-ios/pull/2166 (see also for screenshots there)

Simon-Laux:
fix #3819

### DIFF
--- a/src/renderer/components/Settings/Encryption.tsx
+++ b/src/renderer/components/Settings/Encryption.tsx
@@ -19,17 +19,24 @@ import useDialog from '../../hooks/dialog/useDialog'
 
 import type { DialogProps } from '../../contexts/DialogContext'
 import ManageKeys from './ManageKeys'
+import { SettingsStoreState } from '../../stores/settings'
 
-export default function Encryption() {
+type Props = {
+  settingsStore: SettingsStoreState
+}
+
+export default function Encryption({ settingsStore }: Props) {
   const { openDialog } = useDialog()
   const tx = useTranslationFunction()
 
   return (
     <>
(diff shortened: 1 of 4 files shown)


r10s:
sth went wrong with the rebase?

src/renderer/components/Settings/Encryption.tsx
```diff
@@ -19,17 +19,24 @@ import useDialog from '../../hooks/dialog/useDialog'
...
   const { openDialog } = useDialog()
   const tx = useTranslationFunction()
 
   return (
     <>
-      <CoreSettingsSwitch
-        settingsKey='e2ee_enabled'
-        label={tx('autocrypt_prefer_e2ee')}
-      />
+      {settingsStore.settings.is_chatmail == "0" && (
```
WofWca, line 34: Nitpick: I prefer `!= "1"` to show that it's "hidden in specific cases" rather than "shown in specific cases"

WofWca (commented):
Did not test, the code is good at a glance.

PR #3840, opened 2024-05-21
Show vcard in messages

### LINKED ISSUE #3829: Add support for Viewtype::Vcard
https://github.com/deltachat/deltachat-core-rust/pull/5536 introduces a new viewtype `Vcard` (`Viewtype::Vcard` in JsonRPC, `DC_MSG_VCARD` in the C FFI). In order to show it, call the JsonRPC function `parse_vcard(filePath)` where `filePath` is the path to the message's attachment.

`parse_vcard()` will return a JSON struct with fields `addr`, `display_name`, `profile_image`, and `color` (which you can use to create a placeholder image in case there is no profile_image). You can ignore the other returned fields. Note that `parse_vcard()` returns a list of contacts; simply ignore all except for the first contact.

Android implements viewing Vcards here: https://github.com/deltachat/deltachat-android/pull/3043

Can be done now (depends on https://github.com/deltachat/deltachat-core-rust/pull/5536):

- [x] display vcard messages in-chat
- [x] display in quotes/replies
- [ ] display in draft/staging
- [x] When a vCard is clicked, ask for confirmation and then call `import_vcard()` (depends on https://github.com/deltachat/deltachat-core-rust/pull/5582/)

These TODO-items should be postponed:
- [ ] ~~set new core stock string "Contact"~~ (looks like we won't keep this stock string)

Preview from Android:
image

nicodh:
resolves #3829

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@
 ## [Unreleased][unreleased]
 
 ### Added
-- show VCard attachement as VCard in message list
+- show VCard attachement as VCard in message list #3840
+- add contact from VCard & start chat on click #3840
 
 ### Changed
 - Update `@deltachat/stdio-rpc-server` and `deltachat/jsonrpc-client` to `v1.139.6`
@@ -42,8 +43,6 @@
 - Use openstreetmap in map
 - Open map in a separate window
 - Add global map for all chats in account
-- show VCard attachement as VCard in message list
-- add contact from VCard & start chat on click
 
 ### Changed
 - Update translations (2024-05-20) #3746 #3802 #3827 #3837
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,8 @@
 - Use openstreetmap in map
 - Open map in a separate window
 - Add global map for all chats in account
-
+- show VCard attachement as VCard in message list
+- add contact from VCard & start chat on click
 
 ### Changed
 - Update translations (2024-05-20) #3746 #3802 #3827 #3837
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased][unreleased]
 
+### Added
+- show VCard attachement as VCard in message list
+
 ### Changed
 - Update `@deltachat/stdio-rpc-server` and `deltachat/jsonrpc-client` to `v1.139.6`
 - Always show `msg.overrideSenderName` even when the message is sent by yourself
--- a/src/renderer/components/message/Message.tsx
+++ b/src/renderer/components/message/Message.tsx
@@ -43,14 +43,12 @@ import Reactions from '../Reactions'
 import ShortcutMenu from '../ShortcutMenu'
 import InvalidUnencryptedMailDialog from '../dialogs/InvalidUnencryptedMail'
 import Button from '../Button'
+import VCardComponent from './VCard'
 
 import styles from './styles.module.scss'
 
 import type { OpenDialog } from '../../contexts/DialogContext'
 import type { PrivateReply } from '../../hooks/chat/usePrivateReply'
-import useChat from '../../hooks/chat/useChat'
-import ConfirmationDialog from '../dialogs/ConfirmationDialog'
-import { createChatByContactId } from '../../backend/chat'
 
 const Avatar = (
   contact: T.Contact,
@@ -658,7 +656,9 @@ export default function Message(props: {
           {message.viewType === 'Webxdc' && (
             <WebxdcMessageContent message={message}></WebxdcMessageContent>
           )}
-          {message.viewType === 'Vcard' && VCardContent(message)}
+          {message.viewType === 'Vcard' && (
+            <VCardComponent message={message}></VCardComponent>
+          )}
           {content}
           {hasHtml && (
             <div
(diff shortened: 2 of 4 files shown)


nicodh:
Something seems to be wrong with the prettier check: local check shows no errors, the errors displayed here like 
`"app.asar"` are not in the code base?

src/renderer/components/message/Message.tsx
```diff
@@ -542,6 +542,36 @@ export default function Message(props: {
...
+    const codepoint = displayName && displayName.codePointAt(0)
+    const initial = codepoint
+      ? String.fromCodePoint(codepoint).toUpperCase()
+      : '#'
+    content = (
+      <div className={styles.vcard}>
+        <div
+          className={classNames('avatar', styles.avatar)}
+          aria-label={displayName}
+        >
```
farooqkz, line 556: I think it makes sense to use the same Avatar component as chatlist?

src/renderer/components/message/Message.tsx
```diff
@@ -542,6 +542,36 @@ export default function Message(props: {
         </div>
       </div>
     )
+  } else if (message.vcardContact) {
+    const { profileImage, color, displayName, addr } = message.vcardContact
```
farooqkz, line 546: Maybe it makes sense to introduce a new component named "VCardContact"?

src/renderer/components/message/Message.tsx
```diff
@@ -542,6 +542,36 @@ export default function Message(props: {
         </div>
       </div>
     )
+  } else if (message.vcardContact) {
```
Simon-Laux, line 545: Isn't vcard an attachment that can still have a text associated with it like webxdc messages?
I don't think it is as special of a message as video chats that get fully custom UI

Simon-Laux: cc @r10s @adbenitez ?

iequidoo: Yes, it can have a text as well. Moreover, this is already possible in Android

src/renderer/components/message/Message.tsx
nicodh: Yes I agree, but then we have to change the line  there, since the profileImage is base64 encoded. Will do that later...

nicodh:
Display in stage, replies, drafts etc. uses just the currently provided stuff from core, which might be improved in future

src/renderer/components/message/Message.tsx
```diff
@@ -586,6 +589,11 @@ export default function Message(props: {
   const hasText = text !== null && text !== ''
   const fileMime = (!isSetupmessage && message.fileMime) || null
   const isWithoutText = isMediaWithoutText(fileMime, hasText, message.viewType)
+  const showAttachment = (message: T.Message) =>
```
Simon-Laux, line 592: `showAttachment` as function sounds like an action, maybe rename it to sth like `shouldShowAttachment` or make it a variable/const instead of a function?

src/renderer/components/message/Message.tsx
```diff
@@ -819,3 +828,70 @@ function WebxdcMessageContent({ message }: { message: T.Message }) {
     </div>
   )
 }
+
+/**
+ * displays a VCard attachement with avatar, mail & name
+ * onClick imports VCard and creates chat
+ * (only first if multiple contacts are included in VCard)
+ */
+function VCardContent(message: T.Message) {
```
Simon-Laux, line 837: this component should get it's own file

src/renderer/components/message/Message.tsx
```diff
@@ -819,3 +828,70 @@ function WebxdcMessageContent({ message }: { message: T.Message }) {
...
+  if (message.vcardContact) {
+    const { profileImage, color, displayName, addr } = message.vcardContact
+    const startChatWithContact = async (addr: string) => {
+      const continueProcess = await new Promise((resolve, _reject) => {
+        openDialog(ConfirmationDialog, {
+          message: tx('ask_start_chat_with', displayName),
+          confirmLabel: tx('ok'),
+          cb: resolve,
+        })
+      })
```
Simon-Laux, line 851: there is a hook you can use for this:
```

const confirmed = await openConfirmationDialog({
        message: tx('ask_start_chat_with', displayName),
        confirmLabel: tx('ok'),
      })
```

src/renderer/components/message/Message.tsx
```diff
@@ -819,3 +828,70 @@ function WebxdcMessageContent({ message }: { message: T.Message }) {
...
+ * displays a VCard attachement with avatar, mail & name
+ * onClick imports VCard and creates chat
+ * (only first if multiple contacts are included in VCard)
+ */
+function VCardContent(message: T.Message) {
+  const { selectChat } = useChat()
+  const accountId = selectedAccountId()
+  const { openDialog } = useDialog()
+  const tx = useTranslationFunction()
+  if (message.vcardContact) {
```
Simon-Laux, line 842:
```suggestion
  if (message.vcardContact) {
    return null
  }
```

src/renderer/components/message/Message.tsx
```diff
@@ -819,3 +828,70 @@ function WebxdcMessageContent({ message }: { message: T.Message }) {
...
+        })
+      })
+
+      if (!continueProcess) {
+        return null
+      }
+
+      const contactIds = await BackendRemote.rpc.importVcard(
+        accountId,
+        message.file ?? '' // vCard messages always have a file property
```
Simon-Laux, line 859:
```suggestion
        message.file! // vCard messages always have a file property
```
would also be possible

src/renderer/components/message/Message.tsx
```diff
@@ -819,3 +828,70 @@ function WebxdcMessageContent({ message }: { message: T.Message }) {
...
+              alt={displayName}
+              className='content'
+              src={'data:image/jpeg;base64,' + profileImage}
+            />
+          ) : (
+            <div style={{ backgroundColor: color }} className='content'>
+              {initial}
+            </div>
+          )}
+        </div>
```
Simon-Laux, line 889: is there a reason not to use the avatar component?

Simon-Laux (approved):
Works nicely, I just have some code suggestions. before merging we would need to create an issue for the composer preview, so that todo does not get lost

src/renderer/components/message/Message.tsx
Simon-Laux: but does not matter probably, your solution ensures that it is a string, that's probably better

src/renderer/components/message/Message.tsx
```diff
@@ -819,3 +828,70 @@ function WebxdcMessageContent({ message }: { message: T.Message }) {
...
+ * (only first if multiple contacts are included in VCard)
+ */
+function VCardContent(message: T.Message) {
+  const { selectChat } = useChat()
+  const accountId = selectedAccountId()
+  const { openDialog } = useDialog()
+  const tx = useTranslationFunction()
+  if (message.vcardContact) {
+    const { profileImage, color, displayName, addr } = message.vcardContact
+    const startChatWithContact = async (addr: string) => {
```
Simon-Laux, line 844: the logic in this function should be probably wrapped into a try/catch with a log statement and an alert dialog in the catch statement so the user also gets the error

Simon-Laux: maybe we should make a helper function for this at some point, this is also a problem in other places.

Some hook like
```js
function useCallbackWithUserAlert(cb, dependencies, alertMessage) {
const openAlertDialog = useAlertDialog()
return useCallback(()=>{
return async (...args) => {
try {
await cb(...args)
} catch (error) {
log.error(alertMessage, error)
openAlertDialog({message: alertMessage + error && error.message ?? error.toString()})
}
}, [openAlertDialog,...dependencies])
}
```
that we could use in these places:

```js
const startChatWithContact = useCallbackWithUserAlert(async (addr: string) => {}, [])
```

Simon-Laux: transferred this to it's own issue: #3883

CHANGELOG.md
```diff
@@ -31,7 +31,8 @@
 - Use openstreetmap in map
 - Open map in a separate window
 - Add global map for all chats in account
-
+- show VCard attachement as VCard in message list
+- add contact from VCard & start chat on click
```
Simon-Laux, line 35: needs to be rebased / not be in the old/already released version

Issue #1830, opened 2020-07-30
Use virtual scrolling for other contact lists to account for slow loading


Simon-Laux:
Some users (probably users with many contacts) observe slow loading of the dialogue when creating a new chat, this is caused by the contact list component.

There are two separate fixes which will help with lowering the loading time:

* Use faster `getContactIds` RPC call (as pointed out by @link2xt)
* Use virtual scrolling as we already use it in the global search (as pointed out by @Simon-Laux)

Simon-Laux:
I think the only place where it would make some sense is the new chat dialog, UnblockContact and ViewGroup don't likely have so much items that it would be worth it.

Simon-Laux:
merging in comments from #3632:

> Contact list is using slow getContacts JSON-RPC call which returns a list of objects rather than calling fast getContactIds and then loading contacts on demand.
> 
> This result in 150-200ms delay when clicking on "+" button in the chatlist before any contacts are displayed.
>
> ~ link2xt in #3632


> Actually measured with:
> 
> ```
> console.time("foo"); exp.rpc.getContacts(1, 2, ''); console.timeEnd("foo")
> ```
> 
> and it shows at most 27ms :/
> 
> So react debounce/memo wrappers might be adding something to it.
>
> ~ link2xt in #3632


> You forgot an `await`:
> 
> ```js
> console.time("foo"); await exp.rpc.getContacts(window.__selectedAccountId, 2, ''); console.timeEnd("foo")
> ```
> 
> In my tests: `exp.rpc.getContactIds`: is 1,5-25ms `exp.rpc.getContacts`: is 65-77ms
> 
> The debouncing might add a delay, especially if it waits the initial time, haven't looked at the code, but I think using a virtual list as the already existing issue #1830 says is the real solution.
>
> ~ simon-laux in #3632

WofWca:
Uh-oh, not great news. What I didn't realize is that [`BackendRemote.rpc.getContacts()`](https://github.com/deltachat/deltachat-desktop/blob/bf11419376a1263b62699c770f6ed7b6c12a059c/src/renderer/components/contact/ContactList.tsx#L96) is async...
Itself it takes ~3 seconds to finish for 5000 contacts:

![image](https://github.com/deltachat/deltachat-desktop/assets/39462442/1f2dd0ae-6fc9-48c7-9a57-408b6c9b0a03)

So after all it's not _only_ the UI, as I initially said. @link2xt and @Simon-Laux were both right.

[...]

PR #3842, opened 2024-05-21
perf: use 'react-window' for contact list

WofWca:
- [x] Need to clean up, address the TODOs
- [x] Would also need to verualize it in [`CreateChatMain`](https://github.com/deltachat/deltachat-desktop/blob/ad70318c5e7d00089d9cefa6c4702159d0a8486a/src/renderer/components/dialogs/CreateChat.tsx#L209)
- [x] And perhaps separate this into a new component, since we might need to implement lazy loading later (with 'react-window-infinite-loader') (but I guess that's a problem for a future us)
- [x] Personally test it

For huge contact lists (5000 contacts), rendering took ~5 seconds.
This solves this problem by only actually rendering the
part of the list that is currently scrolled to.

We already have this implemented for `ChatList.tsx`

Parially addresses https://github.com/deltachat/deltachat-desktop/issues/1830
Why partially? Because `BackendRemote.rpc.getContacts()`
still takes 3 seconds for 5000 contacts

This commit has side effects:
- a negative: the affected contact lists on non-default app themes
    where the item height is not exactly 64px
    are now slightly more ugly. See TODO about `itemSize={64}`
- a (positive?) side effect: `AddMemberChipsWrapper`
    is now outside the scrollable region (i.e. it's always on top)
    It works fine since it's `max-height: 33%`

### DIFF
--- a/src/renderer/components/contact/ContactList.tsx
+++ b/src/renderer/components/contact/ContactList.tsx
@@ -6,6 +6,13 @@ import { debounceWithInit } from '../chat/ChatListHelpers'
 import { BackendRemote, Type } from '../../backend-com'
 import { selectedAccountId } from '../../ScreenController'
 
+/**
+ * Make sure not to render all of the user's contacts with this component.
+ * Some users might have thousands of contacts, which can result in
+ * this component taking seconds to render.
+ * Instead consider using 'react-window', like we do in the `CreateChat`
+ * component.
+ */
 export function ContactList(props: {
   contacts: Type.Contact[]
   onClick?: (contact: Type.Contact) => void
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@
 - fix long names breaking layout of reactions dialog #3736
 - hide "add second device" instructions when transfer has started #3748
 - improve chat scroll performance #3743, #3747
+- faster load of contact lists in "New Chat" and "New Group" #3842
 - reduce CPU load when moving mouse over chat #3751
 - fix chatlistitem background when context menu for it is shown it is now highlighted correctly on pinned chats #3766
 - fix add missing top padding to confirm sending files dialog #3767
--- a/src/renderer/components/dialogs/CreateChat/index.tsx
+++ b/src/renderer/components/dialogs/CreateChat/index.tsx
@@ -1,6 +1,5 @@
 import React, {
   ChangeEvent,
-  Fragment,
   useCallback,
   useContext,
   useEffect,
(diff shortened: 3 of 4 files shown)


src/renderer/components/contact/ContactList.tsx
```diff
@@ -6,6 +6,13 @@ import { debounceWithInit } from '../chat/ChatListHelpers'
 import { BackendRemote, Type } from '../../backend-com'
 import { selectedAccountId } from '../../ScreenController'
 
+/**
+ * Make sure not to render all of the user's contacts with this component.
+ * Some users might have thousands of contacts, which can result in
+ * this component taking seconds to render.
+ * Instead consider using 'react-window', like we do in the `CreateChat`
+ * component.
+ */
```
Simon-Laux, line 15: I think it would make sense to have the list of Create chat also as component.

In the long term it would also make sense to move this `ContactList` component to use react-window and lazy loading, though it might be a bit more involved because of the extra features.

maybe we should rename `ContactList` to sth like `ContactListForEdit` or some better name that reflects the extra features when we have a dedicated virtual contact list.

Though maybe it's too much work for now to generalise the virtual list from CreateChat to it's own component because there is the question of how to best represent the pseudo/fake contact items in the api of such an component.

Simon-Laux: Maybe we could have some contact list component with some api like

```tsx
interface PseudoItem {
   icon: IconName,
   label: string,
   clickAction: ()=>void,
   contextMenu?: ContextMenuItem
}

return <ContactList
   extraItemsInBeginning={PseudoItem[]}
   contactIds={number[]}
>
   { (contact: Contact) => <Contact contact={contact }/> }
   // so you could replace the contact item with alternative versions
   // that support either checkboxes or removal buttons:
   { 
      (contact: Contact) =>
      <RemovableContact
         contact={contact} 
         onRemove= {(contactId:number)=>{/* removal logic */}} />
   }
</ContactList>

```

I think sth like that could be an api improvement. But definitely out of scope for this pr.

WofWca: Yeah, I thought it's too early to make an abstraction just for these two places, especially considering that neither 'react-window' nor `ContactList` are too thick to inline.

Simon-Laux: well technically there at least 3 cases/variants:

- normal: new chat dialog
- checkbox: add members to group
- remove button: group members

I think the contact list with it's many props is kinda hard to read code, like it could be simpler with such an abstraction.
react-window + infini-loader add quite some code we shouldn't copy paste into every component. anyways this is still general thoughts nothing for this pr.

Simon-Laux (approved):
Works nicely

PR #3845, opened 2024-05-23
Fix the problem where Ok Button does not close EditAccountAndPassword dialogue

farooqkz:
I am not sure about my solution. Perhaps a better solution would be removing `disableUpdate` state?


Simon-Laux:
maybe rename `disableUpdate` to `nothingChanged`?

Simon-Laux (approved):

PR #3850, opened 2024-05-23
remember profile picture and displayname by saving it to core earlier

Simon-Laux:
profile picture is saved as soon as it is set,
displayname is set when you leave the dialog (by clicking other options, login or going back) or loose focus on the input field


r10s (approved):
nice, tested and it works as expected. cannot say much to the code.

PR #3862, opened 2024-05-26
fix: message text "wobbling" on hover

WofWca:
Introduced in a99910eaee

Before ("Layer Borders" in dev tools is turned on):

https://github.com/deltachat/deltachat-desktop/assets/39462442/8102a210-d33c-4f63-9840-67d375a7495e

Watch the text wobble (it's not just a compression thing).

Does not require a CHANGELOG entry because that commit is unreleased

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ### Fixed
 - fix that map tab is highlighted when you are in the media/gallery tab #3867
+- fix message text "wobbling" on hover #3862
 
 
 <a id="1_45_0"></a>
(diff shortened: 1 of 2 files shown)


Simon-Laux:
> Does not require a CHANGELOG entry because that commit is unreleased

I think it is already released in the test release.


Also would it make sense to put this on another layer with z-index / position absolute or things like that?

Simon-Laux (approved):

CHANGELOG.md
```diff
@@ -77,6 +77,7 @@
 - fix problem of focus when opening create chat dialogue #3816
 - fix the bug where EditAccountAndPassword dialog does not close with OK button
 - fix the problem of pressing ENTER when logging in and prompting doing additioanl dialogs #3824
+- fix message text "wobbling" on hover
```
Simon-Laux, line 80: wrong position

Simon-Laux (changes requested):
changelog is at the old release

WofWca:
Sorry, didn't notice it

Simon-Laux (approved):

Issue #3858, opened 2024-05-24
Map doesn't open


missytake:
- Operating System (Linux/Mac/Windows/iOS/Android): xubuntu 22.04
- Delta Chat Version: 1.45.0, installed via .deb
- Expected behavior: when I click on the map icon on the top right of the selected chat window, the map opens.
- Actual behavior: nothing happens.
- Steps to reproduce the problem:
  - open desktop
  - click on the map icon
- Logs:
```
Error occurred in handler for 'open-maps-webxdc': {
  code: -1,
  message: 'failed to open file /opt/DeltaChat/resources/app.asar/html-dist/xdcs/maps.xdc\n' +
    '\n' +
    'Caused by:\n' +
    '    Not a directory (os error 20)'
}
```

missytake:
If I start deltachat-desktop with `npm run start`, it works.

missytake:
The path doesn't work:

```
$ file /opt/DeltaChat/resources/app.asar
/opt/DeltaChat/resources/app.asar: data
```

There is `/opt/DeltaChat/resources/app.asar.unpacked/`, but it just contains:

```
.
└── node_modules
    └── @deltachat
        └── stdio-rpc-server-linux-x64
            ├── deltachat-rpc-server
            └── package.json

3 directories, 2 files
```

r10s:
just another data point: on mac/sonoma with arm-build (thanks for that, btw!), the map shows up

missytake:
> When I click on the gallery icon it is always highlighted along with the map icon.
> 
[...]

PR #3867, opened 2024-05-26
fix that map tab is highlighted when you are in the media/gallery tab

Simon-Laux:
https://github.com/deltachat/deltachat-desktop/issues/3858#issuecomment-2131190083


WofWca (commented):
Did not test. Looks trivial.

PR #3882, opened 2024-05-30
ImageCropper support smaller images

maxphilippov:
- rename options targetWidth/Height to desiredWidth/Height
- if any of the dimensions is smaller than desiredWidth/Height we set new targetWidth/Height and work with smaller dimensions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - fix message text "wobbling" on hover #3862
 - fix: close alternative options dialog when setting up as a second device #3873
 - fix map not opening #3876
+- fix ImageCropper not working on smaller images
 
 
 <a id="1_45_0"></a>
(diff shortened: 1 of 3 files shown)


Simon-Laux (approved):
Much better, it does not allow zooming, but I think that's fine for now (also makes not much sense to allow too small profile images)

r10s:
> Much better, it does not allow zooming

it allows zooming, but it is nearly impossible to discover for the casual user, and bad ux therefore: _use the mouse wheel or a corresponding scrolling (!) gesture_

we should change that to the well known gestures _pinch_ and _punch_ (same for the normal image view, btw) or, if a PR is too hard, create at least an issue for that.

Simon-Laux:
> it allows zooming, but it is nearly impossible to discover for the casual user, and bad ux therefore: use the mouse wheel or a corresponding scrolling (!) gesture

This pr was about fixing too small images, that broke the cropper entirely, now you can not zoom the image if its to small, you can just select the square area you want if it is not square.

Please file an issue for your other suggestions, we have planned + & - buttons, pinch+zoom is also supported by the cropper

PR #3884, opened 2024-05-30
fix right click in ContactList should open a context menu

Simon-Laux:
- **fix: right click in ContactList should open a context menu**
- **don't show deletion option for selfcontact**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - fix: close alternative options dialog when setting up as a second device #3873
 - fix map not opening #3876
 - fix ImageCropper not working on smaller images
+- fix: right click in ContactList should open a context menu
 
 
 <a id="1_45_0"></a>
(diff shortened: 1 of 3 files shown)


nicodh (commented):
I just stumbled over the fact that many clickable items miss the cursor:pointer class: The contact context menu (Delete contact), the contact items and the chat list while the accounts have a pointer cursor...

Is that intentionally?

r10s:
in general, we try to follow the unagitated desktop-approach for the mouse curser.

for desktop-apps, there is no rule "everything that is tappable" gets the pointing hand. in dc-desktop, we got some mix:

so, changing the cursor to a pointing hand makes more sense when the items are smaller, not directly beside each other and and not "well known" each other - so, eg "start" button for webxdc, links, icons in the main window.

changing the cursor makes less sense, when items are beside each other without gaps, or at well known positions. eg. for the chatlist, half of the screen would be a pointing hand otherwise. also for window-tools as close/fullscreen - and for menus, the pointing hand makes less sense and is usually also not used by the os

some "hard rules" may be easier, however, eg. for the chatlist it also looks worse, and for the menus unexpected, so it stays a bit vague.

nicodh:
Ok,  at least in Slack or VSCode every clickable item changes the cursor to a pointer.
Anyway fine for me, I just wanted to know if it is intentionally

nicodh (approved):

PR #3889, opened 2024-06-01
replace `momentjs` with `luxon`

### LINKED ISSUE #3866: A bug with date of audio messages in Russian
- Operating System: Linux
- Delta Chat Version: 1.45
- Actual behavior: Month and day not properly appearing for audiofiles
deltachat-2024-05-26-114950 (1)

### LINKED ISSUE #3886: Chinese shows `N_MINUTES:ONE` instead of age of message
from the forum, https://support.delta.chat/t/help-testing-the-upcoming-v1-45-x-release/3089/7

one can easily reproduce that with any of the chinese translations, the wrong text is shown at least if the message is between one and two minutes old.

Screenshot 2024-05-31 at 12 16 30

Simon-Laux:
`luxon` makes use of the browser's Internationalization API. 

also fix update timestamps when language changes

closes #3866
closes #3886

Let's move merging this to the next release cycle. (1.48) - the two issues are already semi fixed by the translation update


Simon-Laux:
I also changed the format of some dates. Might need a follow up pr if the format is not 100% right in some places.

src/renderer/components/conversations/formatRelativeTime.ts
```diff
@@ -1,59 +1,74 @@
...
+          ...DateTime.TIME_SIMPLE,
+          weekday: 'short',
+        }) /* Saturday, 5:51 PM */
+      : timestamp.toLocaleString({
+          ...DateTime.TIME_SIMPLE,
+          weekday: 'short',
+        }) /* Sat 5:51 PM */
+  } else {
+    timestamp.toRelative()
+  }
```
WofWca, line 47: But even with that, this format looks wonky, too verbose, takes too much space. This formats as "20 minutes ago". This is used both in the chat list and on messages, and it takes too much space IMO. It's the "ago" word I think.
I have checked out the docs of [`Intl.RelativeTimeFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat) and of `timestamp.toRelative()` and have not found an option that gets rid of "ago". `{ style: 'narrow' }` formats it as "1m ago", and `{ style: 'short' }` as "1 min. ago".

Perhaps using just the time (13:37) is good, as most messengers (including DC Android) do. But this could be done in a separate MR I guess.

src/renderer/components/conversations/formatRelativeTime.ts
```diff
@@ -1,59 +1,74 @@
...
-    return tx(key, String(diff.minutes()), {
-      quantity: diff.minutes() === 1 ? 'one' : 'other',
+export function DayMarkerLabel(tx: getMessageFunction, timestamp: DateTime) {
+  const now = DateTime.now()
+  const diff = DateTime.now().diff(timestamp).shiftToAll()
+
+  if (diff.days <= 1 && now.ordinal === timestamp.ordinal) {
+    return tx('today')
+  } else if (now.ordinal === timestamp.ordinal - 1) {
+    return tx('yesterday')
```
WofWca, line 64:
```suggestion
  // Yes, this will be false on Jan 1 for yesterday's messages, but it's fine.
  } else if (now.ordinal === timestamp.ordinal - 1) {
    return tx('yesterday')
```

src/renderer/components/conversations/formatRelativeTime.ts
```diff
@@ -1,59 +1,74 @@
...
+  } else if (diff.days >= 1 || now.ordinal !== timestamp.ordinal) {
+    return extended
+      ? timestamp.toLocaleString({
+          ...DateTime.TIME_SIMPLE,
+          weekday: 'short',
+        }) /* Saturday, 5:51 PM */
+      : timestamp.toLocaleString({
+          ...DateTime.TIME_SIMPLE,
+          weekday: 'short',
+        }) /* Sat 5:51 PM */
```
WofWca, line 44:
```suggestion
    return extended
      ? timestamp.toLocaleString({
          ...DateTime.TIME_SIMPLE,
          weekday: 'long',
        }) /* Saturday, 5:51 PM */
      : timestamp.toLocaleString({
          ...DateTime.TIME_SIMPLE,
          weekday: 'short',
        }) /* Sat 5:51 PM */
```

WofWca (changes requested):
I have not considered whether we can go without a library at all, or whether `luxor` is the best choice, I haven't compared it with, say, `Day.js`.

Also I was worried that `luxor` might be big, but it does not include almost any localization strings itself, it only uses native `Intl` browser API.

I have confirmed that this fixes #3866 and #3886 (although in a way that I'm not sure I like).

I have QA'ed it lightly for a couple of minutes (with my suggested changes) and it looks fine.

Having looked at the code, I have not managed to find cases where we'd show something plain wrong or ambiguous, although I might have missed something.

With the suggested changes applied, it's fine to merge, though the relative time format might need to be reconsidered.

src/renderer/components/conversations/formatRelativeTime.ts
```diff
@@ -1,59 +1,74 @@
...
+import { DateTime } from 'luxon'
 
-const getExtendedFormats = () => ({
-  y: 'lll',
-  M: `${window.static_translate('timestamp_format_m_desktop') || 'MMM D'} LT`,
-  d: 'ddd LT',
-})
-const getShortFormats = () => ({
-  y: 'll',
-  M: window.static_translate('timestamp_format_m_desktop') || 'MMM D',
```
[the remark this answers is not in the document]
Simon-Laux: no we will need to remove it from the translations in the android repo. Bjoern does that from time to time (`grep`-ing for potentially unused strings across all projects)

farooqkz:
I think we definitely need to move from `moment` on to something else. But I am not sure `luxon` is the best choice. For starters, dayjs has the following advantages:

 - It seems more supported and more in use.
 - It has Jalali calendar plugin. It will be extremely useful once we decide to add Jalali calendar support for users who use it. Typically some Kurds(who are spread in different countries including Turkey, Iran, Iraq and Syria), people from Iran, Tajikistan, Afghanistan and Uzbakistan.
 - `dayjs` is only 2KB, not even half `luxon` (minified)

Simon-Laux:
Luxon is just a wrapper around the Browser `Intl` api, what calendars are supported is dependent on the browser and can be found out like this:

```
console.log(Intl.supportedValuesOf('calendar'));
```
output for electron:
```
[
    "buddhist",
    "chinese",
    "coptic",
    "dangi",
    "ethioaa",
    "ethiopic",
    "gregory",
    "hebrew",
    "indian",
    "islamic",
    "islamic-civil",
    "islamic-rgsa",
    "islamic-tbla",
    "islamic-umalqura",
    "iso8601",
    "japanese",
    "persian",
    "roc"
]
```

persian calendar seems to be supported

Simon-Laux:
I'm not sure which `day.js` plugin you mean, but https://github.com/alibaba-aero/jalaliday/tree/master for example does only seem to support Jalali and Gregorian calendars, so less calendars compared to the browser api that luxon uses.

farooqkz:
> Luxon is just a wrapper around the Browser `Intl` api, what calendars are supported is dependent on the browser and can be found out like this:
> 
> ```
> console.log(Intl.supportedValuesOf('calendar'));
> ```
> 
> output for electron:
> 
> ```
> [
>     "buddhist",
>     "chinese",
>     "coptic",
>     "dangi",
>     "ethioaa",
>     "ethiopic",
>     "gregory",
>     "hebrew",
>     "indian",
>     "islamic",
>     "islamic-civil",
>     "islamic-rgsa",
>     "islamic-tbla",
>     "islamic-umalqura",
>     "iso8601",
>     "japanese",
>     "persian",
>     "roc"
> ]
> ```
> 
> persian calendar seems to be supported, though `Jalali` seems to be different, maybe it hides in some of the other options? otherwise it might be better to push for official support in browsers.



> I'm not sure which `day.js` plugin you mean, but https://github.com/alibaba-aero/jalaliday/tree/master for example does only seem to support Jalali and Gregorian calendars, so less calendars compared to the browser api that luxon uses.

Hmm interesting. Let's do more research.

farooqkz:
Cheers! When I switch the language to Persian, it also switches to Persian calendar :D

src/renderer/App.tsx
```diff
@@ -56,11 +57,26 @@ export default function App(_props: any) {
...
+      )
+      LuxonSettings.defaultLocale = 'system'
+    }
     window.localeData = localeData
     window.static_translate = translate(localeData.messages)
     setLocaleData(localeData)
-    moment.locale(localeData.locale)
     updateCoreStrings()
+    // make sure the string timestamps are updated as well.
+    setTimeout(() => updateTimestamps(), 0)
```
farooqkz, line 79: Why not `debounce`?

src/renderer/components/message/MessageMetaData.tsx
```diff
@@ -56,7 +56,7 @@ export default function MessageMetaData(props: Props) {
       {hasLocation && <span className={'location-icon'} />}
       <Timestamp
         timestamp={timestamp}
-        extended
+        extended={true}
```
farooqkz, line 59: Why?

farooqkz (changes requested):
If you address my questions and also commit @WofWca's suggestions, you have my approval.

farooqkz:
Upon merging this PR, I will make a small PR to add a setting to use Gregorian calendar regardless of locale.

WofWca (commented):
Please note that the referenced issues have already been fixed.
So in practice this only changes the "message age" format (and perhaps some other formats), and removes the dependency on the on `\`${window.static_translate('timestamp_format_m_desktop') || 'MMM D'} LT\`,`

r10s:
>  I will make a small PR to add a setting to use Gregorian calendar regardless of local

usually, we [avoid adding options wherever we can](https://github.com/deltachat/deltachat-desktop/blob/main/CONTRIBUTING.md#rough-ux-philosophy). as locales usually provide a reasonable default and a messenger is no calendar, without additional reasoning, i do not see why to make an exception here.

i think, much about the discussion is about Arabic/Persian - but before considering adding additional switches about how a date is displayed, wouldn't the far larger impact be to support right-to-left properly? for _that_ there are user requests :)

Simon-Laux:
I'd need to rebase the pr.
It would also remove dependency on momentjs which is mostly deprecated since browsers ship the `Intl` api. (so not only get rid of the translation string)

farooqkz:
> > I will make a small PR to add a setting to use Gregorian calendar regardless of local
> 
> usually, we [avoid adding options wherever we can](https://github.com/deltachat/deltachat-desktop/blob/main/CONTRIBUTING.md#rough-ux-philosophy). as locales usually provide a reasonable default and a messenger is no calendar, without additional reasoning, i do not see why to make an exception here. also, we're struggling in that area already without options :)
 

I think there are some Persian users who prefer Gregorian calendar for whatever reason(e.g. living in a Foreign country where the default calendar is Gregorian there).

> but let's leave that bit aside: i think, much about the discussion is about Arabic/Persian - before considering adding additional switches about how a date is displayed, wouldn't the far larger impact be to support right-to-left properly? for _that_ there are user requests :)

Very reasonable.

nicodh:
will be reopened when ready for review

PR #3904, opened 2024-06-05
Show error to user if core process exits unexpectedly

### LINKED ISSUE #3902: no error shown when the app cannot be opened
downloaded1.45.2 and 1.45.3 ARM builds from https://support.delta.chat/t/help-testing-the-upcoming-v1-45-x-release/3089/13 - on mac os 14.4.1 the main window does not open.

the app is not really crashed, however, menu and about work somehow (see video)

this is reproducible and does not work all the time. when downgrading to 1.45.0, things work again

https://github.com/deltachat/deltachat-desktop/assets/9800740/c82612ad-9447-4d17-bba7-8f1a811a4323

Simon-Laux:
closes #3902

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased][unreleased]
 
+### Fixed
+
+- Show error to user if core process exits unexpectedly
+
 <a id="1_45_3"></a>
 
 ## [v1.45.3] - 2024-06-05
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


Simon-Laux:
> Are the errors actually printed to stderr in case of failing to open an account? I suspect the log is empty in the case described in https://github.com/deltachat/deltachat-desktop/issues/3902, no?

The stderr of the core is indeed not captured by our logging system, it is printed to the terminal if you run the app from the terminal though.

r10s:
this is how it looks like with the error of #3902 - much better i'd say, one gets a clue now

Screenshot 2024-06-05 at 19 56 45


on hitting "ok", Delta Chat exits, this is also better i'd say

Issue #3830, opened 2024-05-17
Allow to share contact as vCard


adbenitez:
with the new core API it is possible to share contacts as vcard, there should be an option in the attachment button called "Contact" on click the contact list view is opened, user select one contact that gets send as vcard using the make_vcard() JSON-RPC API

- [ ] add "contact" option to attachment menu to select contact
- [ ] on clicking the "contact" button a "contact selection" list appears, to select contact to share as vcard (using https://github.com/deltachat/deltachat-core-rust/pull/5571)

nicodh:
@adbenitez  
Do I get it right: the current core implementation would require these steps:
1. after click on "contact" in attachemnt menu show a dialog to select oner or more contacts
2. once contacts were selected, get VCard including these contacts by calling rpc.makeVcard
3. save this VCard as temporary file (!)
4. attach this file to the draft like a normal file attachement

I would have expected there is a direct way to set draft.vCardContact but it seems that is not supported and that is why step 3 and 4 are neccessary.

adbenitez:
> 1. after click on "contact" in attachemnt menu show a dialog to select oner or more contacts

it is not select but to click a single one to be selected for sharing, sharing multiple at once (single vcard) is not supported atm

> I would have expected there is a direct way to set draft.vCardContact but it seems that is not supported and that is why step 3 and 4 are neccessary?

I think it is not necessary, just setting the file core should be able to recognize that this is a vcard and set the type to vcard, same happens if you attach a webxdc file or a gif image, core takes care to set the proper obvious view type, not sure if this works for draft tho, it would need to be fixed in core then

adbenitez:
> * when a shared contact/vcard is clicked, it should be imported and the 1:1 chat

this was already added in #3840

adbenitez:
the main task here is allowing to share a contact as vcard, sending directly would simplify things greatly and avoid ugly or complex stagging, I think it is the best way to go for now and anyways sending vcard + text was already from the beginning a "nice to have", so it could come at a later point and is not strictly necessary for this issue

iequidoo:
> I think it is not necessary, just setting the file core should be able to recognize that this is a vcard and set the type to vcard, same happens if you attach a webxdc file or a gif image, core takes care to set the proper obvious view type, not sure if this works for draft tho, it would need to be fixed in core then

[...]

PR #3906, opened 2024-06-05
#3830 send contact as vcard

nicodh:
Works for now but needs some improvements in core for the future as mentioned in some comments

1. reduce this overhead:  call rpc.makeVcard by contact id, save the Vcard in a tmp file and attach it to the message which is converted into a vcardContact again - would be easier just to add a vcardContact directly in draft message
2. avoid duplicate code in contact list with react-window: currently several contact list dialogs are using the CreateChat dialog (broadcast, addGroupMember etc.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,10 +44,10 @@
 
 ### Added
 - Added a Small Screen Mode, when you rezise the window to be small it will only show the chatlist with account sidebar or the Chat View with a back button.
-- show VCard attachement as VCard in message list #3840
+- show VCard attachment as VCard in message list #3840
 - add contact from VCard & start chat on click #3840
 - Webxdc realtime support #3741
-- send contact as VCard form attachment context menu #3906
+- send contact as VCard from attachment context menu #3830
 
 ### Changed
 - Update `@deltachat/stdio-rpc-server` and `deltachat/jsonrpc-client` to `v1.140.0`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased][unreleased]
 
+### Added
+- send contact as VCard from attachment context menu #3830
+
 ### Fixed
 - Do not set min window dimensions on screens that are smaller than those min dimensions (such as linux phones)
 - packaging: respect `NO_ASAR` env var in `afterPackHook`
@@ -47,7 +50,6 @@
 - show VCard attachment as VCard in message list #3840
 - add contact from VCard & start chat on click #3840
 - Webxdc realtime support #3741
-- send contact as VCard from attachment context menu #3830
 
 ### Changed
 - Update `@deltachat/stdio-rpc-server` and `deltachat/jsonrpc-client` to `v1.140.0`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@
 - show VCard attachement as VCard in message list #3840
 - add contact from VCard & start chat on click #3840
 - Webxdc realtime support #3741
+- send contact as VCard form attachment context menu #3906
 
 ### Changed
 - Update `@deltachat/stdio-rpc-server` and `deltachat/jsonrpc-client` to `v1.140.0`
--- a/src/main/cleanup_temp_dir.ts
+++ b/src/main/cleanup_temp_dir.ts
@@ -25,7 +25,7 @@ export async function cleanupDraftTempDir() {
         `found old ${files.length} temporary draft files, trying to delete them now`
       )
       const promises = []
-      for (const file in files) {
+      for (const file of files) {
         log.debug('delete', join(path, file))
         promises.push(rm(join(path, file)))
       }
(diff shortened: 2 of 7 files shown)


src/main/cleanup_temp_dir.ts
```diff
@@ -25,7 +25,7 @@ export async function cleanupDraftTempDir() {
         `found old ${files.length} temporary draft files, trying to delete them now`
       )
       const promises = []
-      for (const file in files) {
+      for (const file of files) {
```
nicodh, line 28: without this I had on Linux errors: 
delete /tmp/deltachat-draft/0 - file not exists
delete /tmp/deltachat-draft/1 - file not exists
etc.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - show VCard attachement as VCard in message list #3840
 - add contact from VCard & start chat on click #3840
 - Webxdc realtime support #3741
+- send contact as VCard form attachment context menu #3906
```
iequidoo, line 10: typo: from

src/renderer/components/composer/menuAttachment.tsx
```diff
@@ -101,6 +105,36 @@ export default function MenuAttachment({
     tx,
   ])
 
+  const selectContact = async () => {
+    const { setLastPath } = rememberLastUsedPath(LastUsedSlot.Attachment)
+    let dialogId = ''
+    /**
+     * TODO: reduce the overhead: just provide a vcardContact to draft/message
```
iequidoo, line 112: Maybe add to core a function like `miscSetDraft()` but taking an array of contact ids instead of file path? Say, `setDraftVcard()`?

Simon-Laux:
> avoid duplicate code in contact list with react-window: currently several contact list dialogs are using the CreateChat dialog (broadcast, addGroupMember etc.)

On that topic: https://github.com/deltachat/deltachat-desktop/pull/3842#discussion_r1611815417

src/renderer/components/composer/menuAttachment.tsx
Simon-Laux: I'd imagine more like a `setVcard(draft_msgid, contactId/s)` and doing that when we do the new apis during the upcoming composer rewrite, so not in this version.

CHANGELOG.md
```diff
@@ -4,9 +4,10 @@
 
 ### Added
 - Added a Small Screen Mode, when you rezise the window to be small it will only show the chatlist with account sidebar or the Chat View with a back button.
-- show VCard attachement as VCard in message list #3840
+- show VCard attachment as VCard in message list #3840
 - add contact from VCard & start chat on click #3840
 - Webxdc realtime support #3741
+- send contact as VCard from attachment context menu #3830
```
Simon-Laux, line 10: branch needs to be rebased to current `main` before merging and changelog entry needs to be moved up to current unreleased section

src/renderer/components/composer/menuAttachment.tsx
```diff
@@ -101,6 +105,36 @@ export default function MenuAttachment({
     tx,
   ])
 
+  const selectContact = async () => {
+    const { setLastPath } = rememberLastUsedPath(LastUsedSlot.Attachment)
```
Simon-Laux, line 109: I don't get why you remember the last path here, makes no sense to me to open the file picker to the tmp directory when it's opened the next time.

src/renderer/components/composer/menuAttachment.tsx
```diff
@@ -101,6 +105,36 @@ export default function MenuAttachment({
...
+          selectedAccountId(),
+          [selectedContact.id]
+        )
+        const cleanDisplayname = selectedContact.displayName.replace(
+          /[^a-z_A-Z0-9]/gi,
+          ''
+        )
+        const tmp_file = await runtime.writeTempFileFromBase64(
+          `VCard-${cleanDisplayname}.vcf`,
+          btoa(vCardContact)
```
Simon-Laux, line 128: `btoa` does not work with emojis.

This method works better:
```
{ var reader = new FileReader(); reader.readAsDataURL(new Blob([await exp.rpc.makeVcard(120, [10])])); reader.onloadend = () => console.log(reader.result) }
```

Simon-Laux (changes requested):
sharing contacts with emojis in display name does not work currently, because of the `btoa` function

src/renderer/components/composer/menuAttachment.tsx
iequidoo: Anyway i've created a PR for it, so you can check if the API fits the needs: https://github.com/deltachat/deltachat-core-rust/pull/5677

src/renderer/components/composer/menuAttachment.tsx
nicodh: This doesn't work. Any solution trying to convert utf8 to base64 does that by doing an additional encoding which needs to be decoded again after converting it back from base64.
So if I get it right, the solution in https://github.com/deltachat/deltachat-core-rust/pull/5677 does not help, since composer recreates the draft message without keeping the message id?
So I suggest to implement a writeTempFile similar to writeTempFileFromBase64 and remove that again onve Composer is refactored

src/renderer/components/composer/menuAttachment.tsx
nicodh: Just a leftover from copy paste. I removed it

src/renderer/components/composer/menuAttachment.tsx
Simon-Laux: I guess it would work if we would always use blob/buffer instead of atob & btob, but file also works

Simon-Laux (approved):
needs to be rebased before merging to put the changelog to the unreleased section

PR #3910, opened 2024-06-06
fix: i18n: wrong pluralization for many languages

### LINKED ISSUE #3870: Plural rules are not followed for many languages
- Delta Chat Version: 1.44
- Expected behavior: "2 hours" is pluralized as "2 часов"
- Actual behavior: "2 hours" is pluralized as "2 часов"
- Steps to reproduce the problem: Set language to "Russian" and look at a message that was sent 2 hours ago.
- Screenshots:
    image
- Logs: None

I noticed that in many places the plural rule used is in the form of `count === 1 ? 'one' : 'other'`. A few (hehe) examples:

- https://github.com/deltachat/deltachat-desktop/blob/035cfd0f38467ac7c3df6058df742e617e8280a1/src/renderer/components/conversations/formatRelativeTime.ts#L48
- https://github.com/deltachat/deltachat-desktop/blob/5dad3478b8fa7b336ffaa4f2a8141370b1ef4481/src/renderer/components/dialogs/CreateChat/index.tsx#L353
- https://github.com/deltachat/deltachat-desktop/blob/b342a1d47b505e68caaec71f79c381c3f304405a/src/shared/localize.ts#L49-L51

This might be true for German, English and many other languages, but it is not for as many other languages, such as Arabic and Russian. E.g. in Russian 1 is 'one' and 2 is 'few'. You can check the rules for all (or almost all?) languages [here](https://www.unicode.org/cldr/charts/43/supplemental/language_plural_rules.html)

The solution I believe is to use [Intl Browser API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/PluralRules):

```js
const arCardinalRules = new Intl.PluralRules("ar-EG");
console.log(arCardinalRules.select(0)); // "zero"
console.log(arCardinalRules.select(1)); // "one"
console.log(arCardinalRules.select(2)); // "two"
console.log(arCardinalRules.select(6)); // "few"
console.log(arCardinalRules.select(18)); // "many"
```

I wanted to make an MR, but I wasn't sure which exact places need to be adjusted.

FYI I checked a few (heh) places in DC Android and strings seem to be pluralized alright there.

WofWca:
Closes #3870

Also fix some strings being displayed as keys and not actual strings, e.g. "1 minute" message age in Indonesian, which would show as "n_minutes:one" instead of "1 menit".

However, some strings are still not pluralized at all, such as `autodel_device_ask`, but let's consider that a different issue.

This is somewhat of an alternative to #3889 as a fix to the "message age" part of #3870

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@
 - fix tray icon explaination in settings that appears when started with `--minimized` #3949
 - performance: memorize MessageBody, don't run message parser multiple times for the same message #3951
 - performance: add limits for MessageBody text generally and for quotes, core already has limits on text size, but for the cases where core has a bug it's still useful to have a failsave #3951
+- Fix some strings not being translated on some locales (e.g. "1 minute" message age in Indonesian) #3910
+- Fix strings being incorrectly pluralized for many locales (such as "2 members" in Russian) #3910
 
 ### Removed
 - removed unused Roboto font variants #3949
@@ -64,8 +66,6 @@
 ### Fixed
 
 - Show error to user if core process exits unexpectedly #3904
-- Fix some strings not being translated on some locales (e.g. "1 minute" message age in Indonesian) #3910
-- Fix strings being incorrectly pluralized for many locales (such as "2 members" in Russian) #3910
 
 <a id="1_45_3"></a>
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,8 @@
 ### Fixed
 
 - Show error to user if core process exits unexpectedly #3904
+- Fix some strings not being translated on some locales (e.g. "1 minute" message age in Indonesian) #3910
+- Fix strings being incorrectly pluralized for many locales (such as "2 members" in Russian) #3910
 
 <a id="1_45_3"></a>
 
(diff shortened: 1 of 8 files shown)


WofWca:
FYI this is not critical for the next (1.46) release.

r10s:
> However, some strings are still not pluralized at all, such as autodel_device_ask, but let's consider that a different issue.

see comment above the string. one can assume plural/lots here - if rare languages have different forms for eg. 100 vs 101, the string needs to use sth. as `item(s)` or `item/items` - but i assume these are really rare cases which languages we do not support yet and/or have not been reported as an issue yet. so things are fine here

WofWca:
> but i assume these are really rare cases

I'm not sure if you can call it rare. 100 vs 101 is incorrect for many Slavic languages, including Russian, as well as Arabic (though 99 vs 100). See https://www.unicode.org/cldr/charts/43/supplemental/language_plural_rules.html#Cardinal-Integer

But yeah, it's not too big of a deal I guess.

r10s:
point :) but still, compared to the effort, it still seems good enough. in general, for new strings, we use plurals as needed.

PR #3916, opened 2024-06-07
fix: packaging: respect `NO_ASAR` env var in `afterPackHook`

Simon-Laux:
I need this for flatpak build


r10s (approved):
approving pro-forma :)

would be nice if someone else can have a look as well.

Issue #3533, opened 2023-11-16
modify group is called even when nothing was modified


Simon-Laux:
https://github.com/deltachat/deltachat-desktop/blob/46a13ff8f0c8ac85ee462e936c9b4105ec223fa4/src/renderer/components/dialogs/ViewGroup.tsx#L117-L119

This effect fires one time initially, while nothing is changed yet. We should change the code to not do that. (react hooks are sometimes hard ;)

Even though core now ignores the core set name https://github.com/deltachat/deltachat-core-rust/issues/5011, it still makes sense to fix this.

adzialocha:
> react hooks are sometimes hard ;)

💯

PR #3924, opened 2024-06-08
ndh/fix3807 refresh member list

### LINKED ISSUE #3807: Removing member from group does not refresh member-list
Removing member from group does not refresh member-list. 

going to chat and then to member list shows the member removed. 

platform: sk/ephemeral branches on cvore and desktop which i believe are rebased on master and should not influence this screen so it's basically main.

nicodh:
First commit is only refactoring without code changes (extract methods to own files since CreateChat/index is too long)

Fix happens in second commit
resolves #3807 

resolves https://github.com/deltachat/deltachat-desktop/issues/3533


Simon-Laux (approved):

WofWca (commented):
Glanced over but didn't really get the details of what's going on.
Nothing too crazy at least.

PR #3946, opened 2024-06-17
skip `requestSingleInstanceLock` on mac appstore builds (mas), because it made it unable to start the app on older macOS devices.

### LINKED ISSUE #3938: Apple Store package not starting on MacOS Silicon; DMG works fine
- Operating System (Linux/Mac/Windows/iOS/Android): MacOS 13.6 on Apple Silicon
- Delta Chat Version: 1.44.1 App Store
- Expected behavior: Should start
- Actual behavior: Icon bounces for a few moments in the dock and then stops.
- Steps to reproduce the problem: Run the app
- Screenshots: None, nothing appears
- Logs: None that I can find...

Downloading the DMG and installing the old-fashioned way works without problems.

Simon-Laux:
closes #3938

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased][unreleased]
 
+### Fixed
+- skip `requestSingleInstanceLock` on mac appstore builds (mas), because it made it unable to start the app on older macOS devices.
+
 <a id="1_46_0"></a>
 
 ## [v1.46.0] - 2024-06-10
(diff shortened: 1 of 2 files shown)


r10s (approved):
code/logic lgtm,  could not test, however

Simon-Laux:
I think the release will be testing for this one, sure there might be other ways to try it before with special provisioning certificates from apple, but I'm too busy with other issues to investigate those.

PR #3949, opened 2024-06-17
add option to use system fonts

Simon-Laux:
- **rm unused `$roboto-light` scss var**
- **unused Roboto font variants & use italic variants of Roboto font correctly**
- **fix tray icon explaination in settings that appears when started with `--minimized`**
- **add option to use sytem ui font in appearance settings**

Feature was requested in the forum, it is especially important for other scripts as the noto font is not pretty enough in some cases:
https://support.delta.chat/t/windows-desktop-client-use-systems-default-font/3134

The feature is opt-in and can be toggled in Settings -> Appearance -> "Use System UI Font"


r10s:
what is eg. signal-electron or whatsapp-web doing there? do they have a switch as well?

wondering if we cannot just always use the system font unconditionally

Simon-Laux:
> wondering if it is an option to always use the system font unconditionally (maybe not nice on linux, iirc system fonts are quite nice on windows/Mac)

no as you might have messed up emoji/unicode fonts and I don't want to deal with such bugs, also we can not guarantee it looks nice on all systems with all fonts, also the person in the forum set a custom font as system font if I'm not mistaken.

PR #3951, opened 2024-06-17
performance: memorize MessageBody, don't run message parser multiple times for the same message

Simon-Laux:
- performance: memorize MessageBody, don't run message parser multiple times for the same message
- performance: add limits for MessageBody text generally and for quotes, core already has limits on text size, but for the cases where core has a bug it's still useful to have a failsave


WofWca (approved):
Did not test, or measure performance, but looks nice and simple, the idea makes sense.

PR #3961, opened 2024-06-19
consolidate wording for "Message Info"

r10s:
gist is to use an easier, less cluttering language - before, we had context menu items named 'Info', 'Message Details' or just 'i', opening then 'Message Details'.

together with https://github.com/deltachat/deltachat-android/pull/3137 , the menu entry will be named 'Info' everywhere and open a 'Message Info' dialog, this has the following benefits:

- inner consistency to other Delta Chat apps all apps have recognizable 'Info' in their context menu, opening a 'Message Info' dialog

- inner consistency to other desktopn menu entries: 'Message Details' is a little too much as the context 'Message' is already clear. (we also do not say 'Forward Message' or 'Reply to Message'; as an exception, we regard only 'Delete Message').

- also, 'Info' is shorter and fits better to the otherwise short menu titles.

- outer consistency: 'Info' is very well known for that kind of, well, info

to take full effect, strings need to be re-pulled,
however, this is no blocker and can be done also after merging.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Changed
 - reword advanced setting "Disable Background Sync For All Accounts" -> "Only Synchronize the Currently Selected Account" #3960
+- use 'Info' and 'Message Info' consistently #3961
 
 ### Fixed
 - Fix crash on "Settings" click when not on main screen (e.g. no account selected): hide the "settings" button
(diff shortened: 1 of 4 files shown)


Simon-Laux (approved):

Simon-Laux:
changelog entry is missing though

Simon-Laux:
now there is a changelog entry modifying the already release version...

r10s:
CHANGELOG noise :)

PR #3960, opened 2024-06-19
improvement: wording for background_sync_disabled

WofWca:
It was recently updated in #3717, and became a little awkward. The title could be interpreted as "disable sync for all accounts, including the current one".

PR #3963, opened 2024-06-19
consolidate 'Profile' wording

r10s:
there are no longer 'Accounts' :)

moreover, this shortens 'Only Synchronize Selected Profile' so that it is less often breaked into two lines,
which is not nice for a setting options

(/me wanted to add that to https://github.com/deltachat/deltachat-desktop/pull/3960 , but it was already merged :)


Simon-Laux:
would also make sense to upstream those untranslated strings

Simon-Laux (approved):

PR #3965, opened 2024-06-19
feat: add Telegram-style Ctrl+Up shortcut to reply

WofWca:
This feels quite slow at the current state, but I guess it's still useful, so maybe it's merge-able?
I haven't tested it on chats with large history.
`jumpToMessage` also feels a but slow itself (you can see on the video, it lags behind the quoted text a bit).
I wasn't sure how it's best to go without RPC calls using the data we already store somewhere, so I wrote it this way, it was just faster to write it like this.

https://github.com/deltachat/deltachat-desktop/assets/39462442/7bd8ed30-0ecc-41f5-a8ff-03ac9ee868c5


Simon-Laux:
`jumpToMessage` probably because jump to message often reloads the chat.
How about introducing this as an opt-in experimental setting?

WofWca:
So you think it's fine to keep the logic as is for now and make it an advanced setting?
If so, I'll do it.

src/renderer/components/message/MessageListAndComposer.tsx
```diff
@@ -188,6 +190,67 @@ export default function MessageListAndComposer({ accountId, chat }: Props) {
...
+  const onCtrlArrow = useCallback(
+    async (e: KeyboardEvent) => {
+      if (!e.ctrlKey) {
+        return
+      }
+      if (e.key !== 'ArrowUp' && e.key !== 'ArrowDown') {
+        return
+      }
+      const quoteMessage = (messageId: number) => {
+        setQuoteInDraft(messageId)
```
Simon-Laux, line 205: maybe you could move that logic to the composer? then there would be no need for `__getQuoteInDraft`

Simon-Laux:
yeah logic is ok on first glance. would rather have it inside of the composer, though, to avoid adding another react bypass to the `window` object and because I think composer is the better place for it anyway.

src/renderer/components/message/MessageListAndComposer.tsx
WofWca: I thought it's right to place it here based on the fact that in Telegram you can execute this shortcut even if the composer is not focused.
Plus we need access to the list of messages which composer doesn't really have (of course this can be changed but I think it's cleaner this way despite the fact that we mutate the window object)

WofWca: If this doesn't change your mind, I'll move it to the composer.

WofWca:
Force-pushed a few adjustments:
- made this an opt-in experimental setting
- added "meta" key for Mac
- smol refactors

Also I realized there is
https://github.com/deltachat/deltachat-desktop/blob/54cd76c3628e9add851ef50cf9716728a50f1964/src/renderer/keybindings.ts#L62

which stores close to all shortcuts. I personally don't think it _needs_ to move there, but some might disagree.

src/renderer/components/message/MessageListAndComposer.tsx
Simon-Laux: so the shortcut is global? Then it needs to go to keybindings.ts and become a `KeybindAction`

Simon-Laux: and the composer would get the handler for it with `useKeyBindingAction`.

handler function could be cached in an `useCallback`-hook for more efficiency I guess.

WofWca: Well, I mean it doesn't have to be global. It is only active when a chat is active and perhaps when the chat is focused.

WofWca: Currently it is only active when that `window.` function exists.

Simon-Laux: in those cases there is just no handler defined. I'd say.

The keybings system will need a rework in the future, to make a system that we can use to bring keybindings to everything in the app, maybe something stack based, where the top most dialog/context menu has the keyboard shortcut focus.

WofWca:
I moved the code to Composer.

Also I noticed a bug. If you execute the shortcut rapidly, it gets weird. The messages start getting selected seemingly at random on follow-up shortcut executions, and you have to click "remove quote" cross several times to remove the quote. Each time you click, there is a different quote.

WofWca:
Also I think later it'd be nice to add an "Esc" shortcut to remove the quote, also Telegram-style.

WofWca:
About this
> If you execute the shortcut rapidly, it gets weird

https://github.com/deltachat/deltachat-desktop/assets/39462442/90caf6e7-a85f-4f55-a202-3e5c7b837884

Simon-Laux:
Maybe we should really postpone this pr and build it into the new composer? but I guess being a little buggy is fine if it is just opt-in experimental anyway...

Simon-Laux (approved):

Simon-Laux:
The message highlight of jump to message is a bit confusing as it fades away to slowly for this feature. it could be more visually pleasing if we had a different design/way to highlight messages we jumped to.

PR #3966, opened 2024-06-19
Optimzing images

farooqkz:
I've optimized some of the images to reduce their size.


WofWca (commented):
I guess it's fine, but IDK, they look rough, e.g. the checkmark's checkmark is not made out of straight lines. Maybe it's better to re-generate them from sources (SVG) with higher compression.
Oh, I just noticed that it's only used in "help". Well, the tray icon is rough still. I don't know how it looks on iOS.

Also `README_ASSETS/screenshot.png` is misisng.

Also this branch is not based on `main`.

Simon-Laux (changes requested):
- dirty commit history, contains stuff from react pr
- some svg images look worse on first glance

what kind of optimisation did you use? lossy or lossless?

Simon-Laux:
I'm also not sure about additional lossy (looks like detail is lost) compression on theocrat  bg image.

r10s (changes requested):
there are lots of degradations not worth the few kb saved, esp. when looking at an appsize of 100-200 mb.

it is very obvious at the "guaranteed encrpytion" image, pixelized, wrong colors. but also tray logo and others are much worse.

is suggest to close this PR unmerged. one can do another attempt using eg. `oxipng -Z`

Screenshot 2024-06-19 at 23 14 19

PR #3976, opened 2024-06-25
fix mailto dialog

### LINKED ISSUE #3972: mailto: handler asks to type a contact recipient, but never finds a match
- Operating System: Windows
- Delta Chat Version: 1.46.1 (git: v1.33.0-881-g899f196d) (winstore install)
- Expected behavior: Popup appears with recent chats, filter by typing
- Actual behavior: Popup is empty, typing into filter does not suggest matches or otherwise allow user to proceed, even when typing literal addresses
- Steps to reproduce the problem: I've been using firefox's "email image" context menu. 
image

Simon-Laux:
closes #3972


src/renderer/components/dialogs/MailtoDialog/styles.module.scss
```diff
@@ -0,0 +1,10 @@
+.MailtoDialog {
+  height: 100%;
+}
+
+.MailtoDialogBody {
+  display: flex;
+  flex-direction: column;
+  padding-bottom: 0px;
```
farooqkz, line 8: I think these should be in snake case to be consistent with other styles we've got?

src/renderer/components/dialogs/MailtoDialog/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+import styles from './styles.module.scss'
+
+import type { DialogProps } from '../../../contexts/DialogContext'
+
+type Props = {
+  messageText: string
+}
+
+export default function MailtoDialog(props: Props & DialogProps) {
+  const { onClose, messageText } = props
```
farooqkz, line 24:
```suggestion
export default function MailtoDialog({ onClose, messageText }: Props & DialogProps) {
```

src/renderer/components/dialogs/MailtoDialog/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { selectedAccountId } from '../../../ScreenController'
+import Dialog, { DialogBody, DialogContent, DialogHeader } from '../../Dialog'
+import useCreateDraftMessage from '../../../hooks/chat/useCreateDraftMesssage'
+import useTranslationFunction from '../../../hooks/useTranslationFunction'
+
+import styles from './styles.module.scss'
+
+import type { DialogProps } from '../../../contexts/DialogContext'
+
+type Props = {
```
farooqkz, line 19: TS docs recommend preferring `interface` over `type` for defining object shapes. I am not requesting any change here. But let's get back to this later. This is a reminder.

src/renderer/components/dialogs/MailtoDialog/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const { isChatLoaded, loadChats, chatCache } =
+    useLogicVirtualChatList(chatListIds)
+
+  const onChatClick = async (chatId: number) => {
+    createDraftMessage(accountId, chatId, messageText)
+    onClose()
+  }
+
+  const onSearchChange = (e: React.ChangeEvent<HTMLInputElement>) =>
+    setQueryStr(e.target.value)
```
farooqkz, line 40: What is the reason one is `async` and the other is not?

farooqkz (changes requested):
Other than these "nitpicks", LGTM.

src/renderer/components/dialogs/MailtoDialog/index.tsx
WofWca: It was there before this MR

src/renderer/components/dialogs/MailtoDialog/index.tsx
WofWca: It was there before this MR

src/renderer/components/dialogs/MailtoDialog/index.tsx
WofWca: It was this way before this MR

PR #3980, opened 2024-06-27
ImageCropper controls

maxphilippov:
- panel with on-screen controls: plus, minus, reset, rotate 90 deg
- reset button removed from the dialog controls of ImageCropper


Simon-Laux:
Bildschirmfoto 2024-06-27 um 23 56 30

This is how it currently looks.

my feedback from testing:

- Everything seems to work.
- I think the zoom steps could be greater, but I don't know by how much
- the buttons should have a bigger click target, like be round and have a background (like the gear button in the qr code scanner) or be under the circle cutout, either way they should have an uniform background (reset icon is hard too see clearly, rotate button is also too close to the circle.

src/renderer/components/ImageCropper/index.tsx
```diff
@@ -103,11 +114,71 @@ export default function ImageCropper({
...
+    }
+
+    canvas.width = fullImage.current.clientHeight
+    canvas.height = fullImage.current.clientWidth
+
+    rotation.current += Math.PI * 0.5
+
+    // FIXME: we now only assume Math.PI * 0.5 rotations
+    if (rotation.current > Math.PI * 1.5) {
+      rotation.current = 0
```
WofWca, line 162: Although it appears not to, this looks like this condition could be `true` for when the rotation is actually 1.5 * π, although maybe I don't know enough about floating point numbers. Either way, the suggestion is better IMO.

```suggestion
    // FIXME: we now only assume Math.PI * 0.5 rotations
    if (rotation.current >= Math.PI * 1.999) {
      rotation.current = 0
```

src/renderer/components/ImageCropper/index.tsx
```diff
@@ -103,11 +114,71 @@ export default function ImageCropper({
...
+      return
+    }
     posX.current = 0
     posY.current = 0
     zoom.current = initialZoom.current
     moveImages(0, 0)
+    if (rotation.current > 0) {
+      cutImage.current.src = filepath
+      fullImage.current.src = filepath
+    }
```
WofWca, line 138: IDK if `onZoomReset` is also supposed to reset rotation (then it needs a name change), but it does. But `rotation.current` here remains unchanged, so it gets out of sync with the actual rotation of the image, so sometimes the "reset" button doesn't actually reset rotations, and sometimes it does.

WofWca (changes requested):
Pointed out a minor bug and a few other suggestions, but overall this improves things and works.

r10s:
very nice!  thanks a lot for iterating over in that area!

some high-level comments:

- the "-" icon looks wrong. it should have the same stroke width as "+"

- the hint "Select an area on  the image" is not really helpful in that broadness - it is obvious - but: how? either remove it (eg. "telegram" and others do not have such an hint as well) or make it more useful (eg. whats app shows "Drag the image to adjust")

- "rotate" and "reset" button are too equal.  
for "rotate", i suggest to use sth. as the following: Screenshot 2024-06-28 at 13 01 09 (taken from telegram, but used widely also elsewhere, colors and direction need to be adapted, of course)

maybe that is already different enough. otherwise, if possible, i would prefer the word "reset" instead  (this is also done by others); there is just no good, well known icon for "reset". but if that's tricky, i am also fine with an icon.

maybe the one or the other point can be improved, but already as is, it is a great improvement!

farooqkz (changes requested):
Shouldn't you use React's `useCallback` for `onZoomOut`, `onZoomReset` and other such these functions?

src/renderer/components/ImageCropper/index.tsx
```diff
@@ -289,9 +360,40 @@ export default function ImageCropper({
...
+              className={styles.imageCropperControlsButton}
+              onClick={onRotateImages}
+              aria-label={tx('ImageEditorHud_rotate')}
+            >
+              <Icon coloring='navbar' icon='rotate-right' size={24} />
+            </button>
+            <button
+              className={styles.imageCropperControlsButton}
+              onClick={onZoomReset}
+              aria-label={tx('reset')}
```
farooqkz, line 388: How about creating a new component for all of these?

farooqkz (changes requested):

src/renderer/components/ImageCropper/index.tsx
maxphilippov: I'm not an expert on floating too, but I believe this is a pretty robust solution:
- JS actually uses double precision
- since we reset it to 0 on every full rotation nothing is accumulated, so if it's true once it will be true every time
- I don't think testing for equality gives us anything

I think if we someday go for the crazy idea of allowing free rotation, we might need a more informed solution, for now I believe current one is robust for the case we're going for.

src/renderer/components/ImageCropper/index.tsx
maxphilippov: Thanks, good catch!

src/renderer/components/ImageCropper/index.tsx
maxphilippov: Not a big fan of decomposition for the sake of decomposition. I think it's much simple to understand what's happening in the code, doesn't introduce a lot of repetitive function names (when we pass them to this new component via props) when you search text in a project.
I treat it like a "send" button in a message dialog, just UI to control the component.

src/renderer/components/ImageCropper/index.tsx
```diff
@@ -46,6 +48,8 @@ export default function ImageCropper({
   const shade = useRef<HTMLImageElement>(null)
   // a wrapper inside which we drag an image
   const container = useRef<HTMLImageElement>(null)
+  // a temporary canvas for rotation and result output so we don't create
+  const tmpCanvas = useRef<HTMLCanvasElement>(null)
```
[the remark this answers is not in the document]
maxphilippov: I was looking at this initially but from what I've understood this neither helps to force canvas to be run on GPU neither on the other thread (we have to explicitly make a Worker for that). So I've decided on a more stable feature for now. 

After using the rotation for a while I think it's better to improve that with CSS transform, that would be much faster, it just required to do some math rewriting and I went for the lower hanging fruit.

maxphilippov:
> Shouldn't you use React's `useCallback` for `onZoomOut`, `onZoomReset` and other such these functions?

I don't think it helps and from what I know it might make things worse. useCallback isn't like using memo for an element. The difference between re-rendering a big element cost and just allocating a function object is huge. Wherever I've read about useCallback the recommendation is to use it if you have problem implementing useEffect where the function is a dependency or if you're using React.memo on this element. Otherwise it might just hurt performance, since memoization isn't a free process either.

src/renderer/components/ImageCropper/index.tsx
WofWca: > if it's true once it will be true every time

Good point. Though I think my suggestion is more robust and less eyebrow-raising, so I'd still apply it, but I won't insist.

farooqkz:
> > Shouldn't you use React's `useCallback` for `onZoomOut`, `onZoomReset` and other such these functions?
> 
> I don't think it helps and from what I know it might make things worse. useCallback isn't like using memo for an element. The difference between re-rendering a big element cost and just allocating a function object is huge. Wherever I've read about useCallback the recommendation is to use it if you have problem implementing useEffect where the function is a dependency or if you're using React.memo on this element. Otherwise it might just hurt performance, since memoization isn't a free process either.

Let's merge it with the state it currently has got, then we could profile.

farooqkz (approved):
My objections has all been addressed

maxphilippov:
> very nice! thanks a lot for iterating over in that area!
> 
> some high-level comments:
> 
> * the "-" icon looks wrong. it should have the same stroke width as "+"
> * the hint "Select an area on  the image" is not really helpful in that broadness - it is obvious - but: how? either remove it (eg. "telegram" and others do not have such an hint as well) or make it more useful (eg. whats app shows "Drag the image to adjust")
> * "rotate" and "reset" button are too equal.
>   for "rotate", i suggest to use sth. as the following: Screenshot 2024-06-28 at 13 01 09 (taken from telegram, but used widely also elsewhere, colors and direction need to be adapted, of course)maybe that is already different enough. otherwise, if possible, i would prefer the word "reset" instead  (this is also done by others); there is just no good, well known icon for "reset". but if that's tricky, i am also fine with an icon.
> 
> maybe the one or the other point can be improved, but already as is, it is a great improvement!

We've discussed this privately, here's the result ("reset" went to the bottom row, we dropped the hint for now [example is Telegram], but I left the CSS definition in case we want to bring it back)
image

maxphilippov:
UPD: I've messed up DialogFooter layout, fixing it, not ready for merge

Simon-Laux (changes requested):
https://github.com/deltachat/deltachat-desktop/assets/18725968/458a7ffa-4871-4506-bbc0-4264c993f89f

video shows zooming in and out by clicking, then rotating, then zooming in with scroll wheel (first buggy because of locked scrollwheel - special feature of my mouse then normal with unlocked scroll wheel)

- I still think the zoom behaviour is too slow, also it gets slower the further you are zoomed in.
- also I think the floating buttons should have a background like the button in the qr code scanner for consistency sake and to clearly see the click target: 
Bildschirmfoto 2024-07-02 um 15 20 34

nicodh:
When I tested a 3MB image the rotation took something like 5 to 6 secondes. That means after click nothing seems to happen, after 5 seconds the image gets black for a second and then the image is rotated.
Since there is no animation visible after the first click it will cause more clicks which again delays the next visible rotation even more...
So there should be some "prpocessing..." animation.

Zoom was fine for me!

nicodh (changes requested):
As mentioned it should show some "processing..." notification while the image is rotated.

A 4.5 MB image even takes something like 10-12 seconds...

Besides that functionality is fine!

nicodh:
Better would be to reduce the rotation processing time of course... :-)

maxphilippov:
So now everything should operate smoothly and pretty fast. The biggest JPG I've tested was 12Mb. 

This solution "projects" mouse position on "non-rotated" image and I kinda messed up the solution the first time with explicit direction altering variables, that thing ended up as a free flip horizontally and vertically. I didn't pick any icons, so it's just two crosses for now (2 last buttons on the floating control panel). Is it too bloated, do we need that, what do you think guys? @r10s @nicodh @farooqkz

farooqkz (changes requested):
It's good but as you picked up something big in size, there are some stuff left out still which you need to fix. I think I will make suggestions. However on the UI side the RESET button is ambiguous. With it's current functionality, it only resets rotation and zoom. But as it says(or shouts) reset, I expected it to reset the flip thing as well. Moreover, I feel like it's out of place and should be an Icon button like zoom in out and so on.

IMO, you should make it reset only zoom, and place it as an icon button between zoom in and zoom out. I think change plus and minus buttons to magnifier icons with plus and minus inside. And the reset button would be a magnifier icon with a numeric one (`1`) inside. As what you have is "kind of" an image editor, and image editors already use magnifier icons for zoom in and out, the user will have a more familiar feeling. Plus you can't put a `1` there to reset zoom as it will confuse the user.

src/renderer/components/ImageCropper/index.tsx
```diff
@@ -55,79 +70,116 @@ export default function ImageCropper({
...
-      )
 
-      setLastPath(dirname(filepath))
-      onResult(tempfilepath)
-      onClose()
-    })()
+    const tempfilename = `profile_pic_${Date.now()}.png`
+    const tempfilepath = await runtime.writeTempFileFromBase64(
+      tempfilename,
+      canvas.toDataURL('image/png').split(';base64,')[1]
```
farooqkz, line 130: I don't think it's a good idea to convert any image to PNG. Normal photography images should be something like JPG as converting them to PNG will drastically increase their size.

src/renderer/components/ImageCropper/index.tsx
```diff
@@ -55,79 +70,116 @@ export default function ImageCropper({
   const posX = useRef<number>(0)
   const posY = useRef<number>(0)
 
+  // multiplier to flip direction on X or Y
+  // either positive or negative
+  const flipDirX = useRef<number>(1)
+  const flipDirY = useRef<number>(1)
+
```
farooqkz, line 77:
```suggestion
  const posX = useRef<number>(0)
  const posY = useRef<number>(0)

  // multiplier to flip direction on X or Y
  // either positive or negative
  type FlipDir = -1 | 1
  const flipDirX = useRef<FlipDir>(1)
  const flipDirY = useRef<FlipDir>(1)

```

farooqkz (changes requested):
Overall opinion: Nice job! Keep it going!

r10s:
it was discussed already to keep the old RESET button at its position in the lower right. also because of the lack of a good image, this is fine. this is better than inventing some icon and add it to some position as "between zoom in and zoom out".

however, i agree that RESET needs to reset  everything, if we keep FLIP, that needs to be reset as well.

that said, i would leave FLIP outside of this PR. to move forward, it does not make sense to open more ends :) also, a real need, if at all, is a HORIZONTAL FLIP is sufficient, but that can go to a subsequent PR. but not even sure about that.

so my suggestions to move forward and to get it in:
- remove FLIP for now
- controls and functionality are fine and a huge improvement - let's focus on code-level review

src/renderer/components/ImageCropper/index.tsx
r10s: afaik, core scales down images to a few kb. but that needs to be double checked. not totally sure therefore if PNG or JPG is better here, if sizes are some 1000x1000 pixels, JPG is probably still better.

@iequidoo did lots of things in that area and will know :)

maxphilippov:
> however, i agree that RESET needs to reset everything, if we keep FLIP, that needs to be reset as well.
> 
> that said, i would leave FLIP outside of this PR. to move forward, it does not make sense to open more ends :) also, a real need, if at all, is a HORIZONTAL FLIP is sufficient, but that can go to a subsequent PR. but not even sure about that.
> 
> * remove FLIP for now
> * the PR needs a rebase
> * controls and functionality are fine and a huge improvement - let's focus on code-level review

Can we leave HORIZONTAL FLIP then? It's a pretty small amount of code and doesn't affect anything. Reset will set it back to 1, that's just a thing I forgot to add.

r10s:
> Can we leave HORIZONTAL FLIP in then?

sure, my suggestion was only to cut discussions short :)

maxphilippov:
> * I still think the zoom behaviour is too slow, also it gets slower the further you are zoomed in.
I agree on this one, but I don't know exactly how fast that should go, maybe I'll do another PR when you come back from holidays

maxphilippov:
Rebased version is up, also fixed some bugs in RESET. flipDirY is still there because we operate on 'non-rotated' images so when it's rotated 90 or 270 degrees we use Vertical flip under the hood.

src/renderer/components/ImageCropper/index.tsx
maxphilippov: > afaik, core scales down images to a few kb. but that needs to be double checked. not totally sure therefore if PNG or JPG is better here, if sizes are some 1000x1000 pixels, JPG is probably still better.
> 
> @iequidoo did lots of things in that area and will know :)

We've discussed that with Simon, decided to stay with PNG since it's lossless. Right now the result is limited to 256x256 so savings on size I believe are pretty small.

farooqkz:
> it was discussed already to keep the old RESET button at its position in the lower right. also because of the lack of a good image, this is fine. this is better than inventing some icon and add it to some position as "between zoom in and zoom out".

As I can't find a suitable icon for the purpose, I'm inclined to agree with this.

src/renderer/components/ImageCropper/index.tsx
farooqkz: Wait you did but not through suggestion.

src/renderer/components/ImageCropper/index.tsx
```diff
@@ -187,49 +260,95 @@ export default function ImageCropper({
...
+        ev.clientY,
+        -(rotation.current * Math.PI) / 180,
+        window.innerWidth / 2,
+        window.innerHeight / 2
+      )
+    }
+
+    const handleMouseCoords = (ev: MouseEvent) => {
+      // const dx = ev.clientX, dy = ev.clientY
+      const [dx, dy] = rotate(
```
farooqkz, line 329: These `dx` and `dy` look gorgeous xD

src/renderer/components/ImageCropper/index.tsx
```diff
@@ -55,79 +70,119 @@ export default function ImageCropper({
...
+  const onRotateImages = () => {
+    rotation.current = Math.round(rotation.current + 90) // Math.PI * 0.5
+
+    if (rotation.current >= 360) {
+      rotation.current = 0
     }
+
+    let t = targetWidth.current
+    targetWidth.current = targetHeight.current
+    targetHeight.current = t
```
farooqkz, line 179:
```suggestion
    [targetWidth.current, targetHeight.current] = [targetHeight.current, targetWidth.current]
```

r10s (commented):
looks all good to me now.

thanks a lot for taking care!

for the speed on zooming: surely, faster would be better. 
but compared to the fact that things seems not be be blocking, it is gets slower on large zoom levels only, and it is not _super_-slow as eg. opening large galleries, i do not think, this should block merging or other things from being improved and fixed first :)

the PR needs a rebase however, and i cannot say much to code-level things, not sure if there are open ends

farooqkz (changes requested):

farooqkz:
> The code looks almost good. However, there are some performance concerns, as always(!), from me. I'm using React's `Profiler` component to see if I can make changes to make it more performant. For starters, quoting React's `useCallback` docs:
> 
> > In JavaScript, a function () {} or () => {} always creates a different function, similar to how the {} object literal always creates a new object. Normally, this wouldn’t be a problem, but it means that ShippingForm props will never be the same, and your [memo](https://react.dev/reference/react/memo) optimization won’t work. This is where useCallback comes in handy:
> 
> As you are defining functions like `onZoomIn` like this, and then passing them to child components, this means at every re-render, they will get re-rendered, too. This is my hypothesis, not sure if it's really slower(significantly).

Nope I was wrong. The point is that your component hasn't got a state to change, and therefore it doesn't re-render often.

farooqkz (changes requested):
I think it makes change the `shape` parameter to be `'circle' | 'rect'` or something

farooqkz:
@maxphilippov Other than the nitpicks, it looks good to me. We can merge it. Note that adz is away now. I think we  can merge it with two approvals. You have mine, then let's go for @nicodh's

nicodh (approved):
Looks good for me! Very fluent now even with big images!

Only small point you could (not must) improve: when starting with an image wich is not in quadrat format  but in landscape the image starts being horizontally aligned in the center. When you click the reset button the image is aligned left:
Inital view:
image
After click on reset button:
image

After reset, if you click somewhere outside the image is centered again like in the initial view...

maxphilippov:
> the PR needs a rebase however. and i cannot say much to code-level things, open ends seem to be resolved, however.
Someone juts took "let's get rid of all the issues" too seriously, I just rebased it and there's a new commit on main ;)

> After reset, if you click somewhere outside the image is centered again like in the initial view...
> for the speed on zooming: surely, faster would be better. but compared to the fact that things seems not be be blocking, it gets slower on large zoom levels only, and it is not _super_-slow as eg. opening large galleries, i do not think, this should block merging or other things from being improved and fixed first :)

Yeah, if you don't mind I'll do a follow-up PR later, cause I have 2 PRs on small things to push right now and I want to do a last check on Blueprint PR so it's at least doesn't have some silly mistakes and is ready for review

src/renderer/components/ImageCropper/index.tsx
maxphilippov: > These `dx` and `dy` look gorgeous xD
> 
> (Edit: it was from mathematician point of view!)

Told you, I lost my mind debugging this :D Thanks!

PR #3984, opened 2024-06-27
feat: more shortcuts to switch between chats

WofWca:
`Ctrl + PageDown`, `Ctrl + PageUp`,
`Ctrl + Tab`, `Ctrl + Shift + Tab`

These are used by Telegram: https://github.com/telegramdesktop/tdesktop/wiki/Keyboard-Shortcuts

Related MR: https://github.com/deltachat/deltachat-desktop/pull/3965

I tested it a bit. Appears to work, and not break existing "PageUp" shortcut. ~~Those on Mac, please verify that these shortcuts are alright.~~ see [comment about Mac below](https://github.com/deltachat/deltachat-desktop/pull/3984#issuecomment-2196586587)

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased][unreleased]
 
+## Added
+- More shortcuts to switch between chats: `Ctrl + PageDown`, `Ctrl + PageUp`, `Ctrl + Tab`, `Ctrl + Shift + Tab` #3984
+
 ### Changed
 - reword advanced setting "Disable Background Sync For All Accounts" -> "Only Synchronize the Currently Selected Account" #3960
 - use 'Info' and 'Message Info' consistently #3961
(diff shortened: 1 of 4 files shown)


Simon-Laux:
I'm not cure about ctrl+tab, we might need it to be a more generic way to navigate through the whole app in the future, also I do not know whether and what key combinations are used or required by screen readers. 🤔 

The ones with PageUp and PageDown are fine by me

WofWca:
Update about Mac: Telegram has different shortcuts there https://github.com/overtake/TelegramSwift/wiki , only  ⌥⌘↑

Simon-Laux:
just checked linux Firefox and macOS safari. seems like I confused it with alt+tab on macOS and tab & shift+tab everywhere else so its fine.

Simon-Laux (approved):

Issue #3990, opened 2024-07-01
'realtime channel' experimental option not persisted


r10s:
enabling experimental 'realtime webxdc channels' option is not persisted for me on mac using dc 1.46.1 - had a hard time until i realise that, i was always thinking the [ongoing iOS implementation](https://github.com/deltachat/deltachat-ios/pull/2225) is doing sth wrong 😂

is it only me? i cannot really believe this is a general issue :)

https://github.com/deltachat/deltachat-desktop/assets/9800740/9501bdcb-17b8-4b46-986b-07cd11af4e62

hpk42:
fixed by #3992

PR #3992, opened 2024-07-01
fix: `webxdc_realtime` toggle incorrect value

WofWca:
The toggle would show `false` after every UI reload event though the actual value is `true` ("1")

For reference, `disable_idle` was added correctly in 147d95624110a


hpk42 (approved):
tested the branch and it fixes the issue, thanks!

r10s:
thanks for fixing! ❤️

WofWca:
Funny that I didn't know that you opened #3990 😀

PR #3994, opened 2024-07-01
Fix bug where after clear chat, new self sent messages are not shown in MessageView

### LINKED ISSUE #3991: message view not always updated
the message view is not always updated; at least not after "clear" in a reproducible way, not sure about other situations.

to reproduce:

- chose "Clear Chat" from the three-dot-menu in the upper right corner
- send a message to that chat from another device (without doing anything on the cleared chat)

-> the message appears in the chatlist, but the chat stays empty.

dc 1.46.1 on mac

https://github.com/deltachat/deltachat-desktop/assets/9800740/908bf7bd-ac1d-48a4-b78d-00f8f99f3af2

farooqkz:
This fixes the bug. But I suspect there are far more bugs here because of usage of `React.memo`. I might be wrong, however so let's wait and see.

Fixes #3991

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,38 +2,41 @@
 
 ## [Unreleased][unreleased]
 
-### Changed
+## Added
+- More shortcuts to switch between chats: `Ctrl + PageDown`, `Ctrl + PageUp`, `Ctrl + Tab`, `Ctrl + Shift + Tab` #3984
 
+### Changed
 - reword advanced setting "Disable Background Sync For All Accounts" -> "Only Synchronize the Currently Selected Account" #3960
 - use 'Info' and 'Message Info' consistently #3961
 - consolidate 'Profile' wording #3963
 - Update local help (2024-06-19)
+- refactor: safer types #3993
+- keep aspect ratio in quoted images #3999
 
 ### Fixed
-
 - Fix crash on "Settings" click when not on main screen (e.g. no account selected): hide the "settings" button
 - code: comply with react hook rules #3955
 - fix mailto dialog #3976
 - "Realtime Webxdc Channels" toggle not reflecting actual setting value #3992
-- Fix bug where after clear chat, new self sent messages are not shown
+- even faster load of contact lists in "New Chat" and "New Group" #3927
+- really hide 3dot menu when it is hidden #3998
+- fix react crash when downloading a video message on demand #4000
+- fix bug that showed placeholders while searching in chat forward dialog until you scrolled #4001
 
 <a id="1_46_1"></a>
 
 ## [v1.46.1] - 2024-06-17
 
 ### Added
-
 - add option to use sytem ui font in appearance settings #3949
 - pretty preview for vcard draft #3948
 
 ### Changed
-
 - use `SOURCE_DATE_EPOCH` environment var for build timestamp instead of `Date.now()` if set.
 - use italic variants of Roboto font correctly #3949
 - show chat name when searching in chat #3950
 
 ### Fixed
-
 - skip `requestSingleInstanceLock` on mac appstore builds (mas), because it made it unable to start the app on older macOS devices. #3946
 - fix tray icon explaination in settings that appears when started with `--minimized` #3949
 - performance: memorize MessageBody, don't run message parser multiple times for the same message #3951
@@ -43,7 +46,6 @@
 - make search ui visible when searching in a chat when in small screen mode #3950
 
 ### Removed
-
 - removed unused Roboto font variants #3949
 
 <a id="1_46_0"></a>
@@ -51,11 +53,9 @@
 ## [v1.46.0] - 2024-06-10
 
 ### Changed
-
 - Update translations (2024-06-09) #3925
 
 ### Fixed
-
 - refresh member list after changes #3807
 
 <a id="1_45_5"></a>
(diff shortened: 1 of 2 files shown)


hpk42:
There seem to be a lot of unrelated changes in changelog whitespacing.

hpk42:
also which bug is fixed by this?  Please properly reference it with "fixes #..."  so it automatically is referenced bi-directionally?

CHANGELOG.md
```diff
@@ -2701,21 +2847,13 @@ This section is only relevant to contributors.
 
 [unreleased]: https://github.com/deltachat/deltachat-desktop/compare/v1.46.1...HEAD
 
-[1.46.1]: https://github.com/deltachat/deltachat-desktop/compare/v1.46.0...v1.46.1
 
-[1.45.5]: https://github.com/deltachat/deltachat-desktop/compare/v1.45.5...v1.46.0
```
WofWca, line 2706: Links need to be fixed instead. Plus it'd be better if it was a separate commit, better yet a separate MR.

src/renderer/components/message/MessageList.tsx
```diff
@@ -450,31 +450,31 @@ export type ConversationType = {
...
+    onScroll,
+    messageListItems,
+    messageCache,
+    activeView,
+    messageListRef,
+    chat,
+    loaded,
+    unreadMessageInViewIntersectionObserver,
+    loadMissingMessages,
+  }: MessageListInnerProps) => {
```
WofWca, line 477: This here is just a refactor, right? I'd put it in a separate commit next time.

src/renderer/components/message/MessageList.tsx
```diff
@@ -644,7 +644,8 @@ export const MessageListInner = React.memo(
       prevProps.messageCache === nextProps.messageCache &&
       prevProps.oldestFetchedMessageIndex ===
         nextProps.oldestFetchedMessageIndex &&
-      prevProps.onScroll === nextProps.onScroll
+      prevProps.onScroll === nextProps.onScroll &&
+      prevProps.messageListItems.length === prevProps.messageListItems.length
```
WofWca, line 648: I guess this will fix the particular bug, but, as you mentioned, I believe there is a bigger underlying problem with `memo`.
It was added here https://github.com/deltachat/deltachat-desktop/pull/2535 and was just a way to save memory, if I understand it correctly. After #2987 we are no longer using pages, but messageCache, which is id-addressed, so I suspect we no longer need this. Either way, the manual `memo` condition checking doesn't seem to be saving us too many re-renders - I tried to do a few things and it re-renders practically every time something happens. Perhaps need to just re-render it when any of the props change.

WofWca (changes requested):

WofWca (commented):
In short, I'd revisit / remove CHANGELOG changes. Otherwise it's merge-able.

CHANGELOG.md
farooqkz: I didn't. LazyVim did xD

src/renderer/components/message/MessageList.tsx
farooqkz: Let's already merge this PR and discuss about it later.

ralphtheninja: Is the problem fixed if the `React.memo()` is removed altogether?

WofWca: It should, though I did not check.
I'd keep `React.memo` and only remove `arePropsEqual`, but need to check whether it really is not needed anymore.

farooqkz: Removed manual check

nicodh:
I just tested this PR and it didn't fix the problem. The new message is still shown only in Chatlist but the message list stays empty...

nicodh:
> > I just tested this PR and it didn't fix the problem. The new message is still shown only in Chatlist but the message list stays empty...
> 
> You sure. I just tested and it solves the problem. @Simon-Laux @WofWca What about you?

Tested again and still the new message is not shown if the chat was muted before. The chatlist shows the batch for a new message but messageList stays empty

WofWca:
Indeed, it doesn't appear to fix the bug.
Though I tested it differently, instead of sending a message from a second device of the active account, I send it from the account with whom the chat is open on DC desktop.

WofWca:
I discovered that when the bug happens, when a new message arrives, the `MessageList` re-renders, but `messageListItems` is always empty here

https://github.com/deltachat/deltachat-desktop/blob/9cb8c0d76ca680ef9086e7384c41868e1e3f41f9/src/renderer/components/message/MessageList.tsx#L115

So it doesn't loo like it has to do with `MessageListInner` being memoized.

farooqkz:
> Indeed, it doesn't appear to fix the bug. Though I tested it differently, instead of sending a message from a second device of the active account, I send it from the account with whom the chat is open on DC desktop.

then it's a different bug?

farooqkz:
@WofWca @nicodh I think we are talking about different bugs. This PR with its current state fixes the bug in the mentioned issue. However, I think maybe fix other related bugs in the same PR, too. Can you post exactly how do you produce the bugs, please?

WofWca:
Indeed, those look like different bugs, and for some reason [I cannot reproduce the original bug even on `main` branch](https://github.com/deltachat/deltachat-desktop/issues/3991#issuecomment-2227288348) XD.

Simon-Laux:
Superceded by #4037

PR #3998, opened 2024-07-01
really hide 3dot menu when it is hidden

### LINKED ISSUE #3987: three-dot menu disabled in chat view of audio, files, or apps
- OS: Debian stable (12.5) amd64
- Delta Chat Version: 1.46.1 (installed via .deb)

in a per-chat view, click on the media icon.  from the sub-tabs, choose either "Audio", "Files", or "Apps".  with any of those sub-tabs selected, the three-dot menu (⋮) in the upper right is disabled.  with the other sub-tabs ("Images" or "Video") the three-dot menu (⋮)  is _not_ disabled.

If there's no action choices to take there, just hide the three-dot menu.  making it disabled makes me wonder what i need to do to enable it.

Simon-Laux:
closes #3987

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - fix mailto dialog #3976
 - "Realtime Webxdc Channels" toggle not reflecting actual setting value #3992
 - even faster load of contact lists in "New Chat" and "New Group" #3927
+- really hide 3dot menu when it is hidden
 
 <a id="1_46_1"></a>
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
 - fix mailto dialog #3976
 - "Realtime Webxdc Channels" toggle not reflecting actual setting value #3992
 - even faster load of contact lists in "New Chat" and "New Group" #3927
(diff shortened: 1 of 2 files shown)


r10s (approved):
thanks for taking care!

~~did not test,  but~~ EDIT: tested and code lgtm.

this fixes the issue at hand and is better than disabling, for reasons expained in the issue. 

we'll see if ppl regard the missing kebap menu as an issue. if so, we can iterate and consider to unconditionally show the chat menu as before we added the aspect ratio command - which can be come a button or integrated to the chat-menu then. but let's see, maybe ppl understand the context is the "tab" and not the "chat" now.

but for now, things are fine :)

Simon-Laux:
> we'll see if ppl regard the missing kebap menu as an issue. if so, we can iterate and consider to unconditionally show the chat menu as before we added the aspect ratio command - which can be come a button or integrated to the chat-menu then. but let's see, maybe ppl understand the context is the "tab" and not the "chat" now.

There is also still the idea of redesigning the gallery part of the app, we had an idea of merging chat info and gallery into one big tab at one point for example.

PR #3999, opened 2024-07-01
keep aspect ratio in quoted images

### LINKED ISSUE #3931: wrong aspect ratio for images shown in quotes
when quoting a message with an image, the quote gets enriched with the original image, which is great!

however, the quoted image is always force to be a square, without respecting the aspect ratio:

 &nbsp; 

instead, either:

1. display the whole image - however, the height should still not be larger than the quote. the image can be wider, however. displaying the whole image is also what desktop does otherwise, maybe one can reuse some code

2. crop the image. this is done on mobile, as anyways already done for the normal image display (due to smaller screen and for $reasons). in case of desktop, however, i would show the image as its original, so go for 1.

all in all, this is a minor issue with low priority

Simon-Laux:
closes #3931

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - consolidate 'Profile' wording #3963
 - Update local help (2024-06-19)
 - refactor: safer types #3993
+- keep aspect ratio in quoted images
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


farooqkz (approved):
LGTM

farooqkz (approved):
Also works

farooqkz:
unless someone objects, I will merge in an hour.

PR #4000, opened 2024-07-01
refactor into proper function components, otherwise we would not be allowed to use hooks in those functions if they just return JSX without being a component.

### LINKED ISSUE #3945: React crash/error in message list: Rendered fewer hooks than expected. This may be caused by an accidental early return statement.
```
Error: Minified React error #300; visit https://reactjs.org/docs/error-decoder.html?invariant=300 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
    at Ch (node_modules/react-dom/cjs/react-dom.production.min.js:157:334)
    at li (node_modules/react-dom/cjs/react-dom.production.min.js:180:154)
    at ck (node_modules/react-dom/cjs/react-dom.production.min.js:269:343)
    at bk (node_modules/react-dom/cjs/react-dom.production.min.js:250:347)
    at ak (node_modules/react-dom/cjs/react-dom.production.min.js:250:278)
    at Tj (node_modules/react-dom/cjs/react-dom.production.min.js:250:138)
    at c (node_modules/react-dom/cjs/react-dom.production.min.js:243:163)
    at b (node_modules/react-dom/cjs/react-dom.production.min.js:123:115)
    at Nf (node_modules/scheduler/cjs/scheduler.production.min.js:18:343)
    at gr (node_modules/react-dom/cjs/react-dom.production.min.js:122:325)
```

happened after I clicked on download on a download on demand message, I'll open desktop in dev mode to find out more.

the message I downloaded was a video message.

Simon-Laux:
fixes #3945

### DIFF
--- a/src/renderer/components/message/Message.tsx
+++ b/src/renderer/components/message/Message.tsx
@@ -1,4 +1,10 @@
-import React, { useCallback, useContext, useEffect, useState } from 'react'
+import React, {
+  useCallback,
+  useContext,
+  useEffect,
+  useLayoutEffect,
+  useState,
+} from 'react'
 import reactStringReplace from 'react-string-replace'
 import classNames from 'classnames'
 import { C, T } from '@deltachat/jsonrpc-client'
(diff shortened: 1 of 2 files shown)


src/renderer/components/message/Message.tsx
```diff
@@ -555,9 +574,7 @@ export default function Message(props: {
     )
   }
 
-  // we need this typeconversion, if we don't have it esbuild tries bundling deltachat-node again,
-  // which fails because it imports stuff only available in nodejs
-  const downloadState = message.downloadState
```
Simon-Laux, line 560: outdated comment, so I removed it

WofWca (approved):
Changes make sense, though did not check if the bug is fixed.

PR #3927, opened 2024-06-10
Lazy-load contacts

WofWca:
It's easier to review the changes commit-by-commit (also I'd prefer "rebase and merge" instead of "squash"). Also, for the last commit, GitHub is not good at showing whitespace-only changes, which it mostly is.
I did a bit of QA (but maybe not enough), and looked at the code. It doesn't look like anything should break.

TODO:
- [x] There are 2 more places where `<InfiniteLoader>` needs to be added. For now I just want to show the general approach and see if anyone has objections to it.
- [x] Now the only problem is `Invalid list ref; please refer to InfiniteLoader documentation.` in the console. I think we might need to set it on `<AutoSizer` instead? not sure. Should not be a blocker for reviews. Update: solved by putting `<FixedSizeList` as a direct child of `<InfiniteLoader`

PR #4001, opened 2024-07-01
fix bug that showed placeholders while searching in chat forward dialog until you scrolled

### LINKED ISSUE #3921: Forward Message: Search shows many placeholders
it should load all visible items everytime the chat ids change.

Bildschirmfoto 2024-06-08 um 00 28 54

Simon-Laux:
fixes #3921


src/renderer/components/chat/ChatList.tsx
```diff
@@ -72,11 +72,20 @@ export function ChatListPart({
     | MessageChatListItemData
   itemHeight: number
 }) {
+  const r = useRef<InfiniteLoader | null>(null)
```
farooqkz, line 75: Baaad variable name

WofWca (approved):
Ah crap. #3927 also has this bug...

WofWca:
Did you manage to consistently reproduce #3921 before, and confirmed that it's not reproducible anymore?

src/renderer/components/chat/ChatList.tsx
Simon-Laux: yeah correct

PR #4003, opened 2024-07-02
Found some snake_case var names and converted them to camelCase counterparts

farooqkz:
`camelCase` is the standard in TS and in Desktop. Will merge upon the first approval


nicodh (approved):

WofWca (approved):
I'm not familiar with the entire code base, so don't know if there is a reason for them to be this way. But looks ok.

Simon-Laux (approved):

PR #4004, opened 2024-07-02
fix: more `<InfiniteLoader>` fixes

### LINKED ISSUE #3208: random scroll position and missing redraws when opening "Archive"
- outside the archive, in the normal chatlist scroll some pages down and select a chat
- scroll up in the chatlist and tap "Archive"

-> the archive scrolling position is somewhere some pages down, but not atop (as one would expect eg.  to access just archived chats) 

moreover, when doing this after a fresh app start, the "archive" also not shown, but only some placeholders (they also are not updated after minutes, only after scrolling):

Screenshot 2023-04-21 at 15 34 10

(desktop mac, 1.36.1)

### LINKED ISSUE #3921: Forward Message: Search shows many placeholders

WofWca:
See commits

Please verify that the old skeleton issues didn't come back after the removal of those `useEffect`s.

fixes #3921
fixes #3208

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 - fix bug that showed placeholders while searching in chat forward dialog until you scrolled #4001
 - Fix the problem of Quit menu item on WebXDC apps closes the whole DC app #3995
 - minor performance improvements #3981
+- fix chat list items (e.g. Archive) and contacts not showing up sometimes #4004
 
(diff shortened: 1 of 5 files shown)


Simon-Laux (approved):
works

PR #4005, opened 2024-07-03
fix: a11y: keyboard navigation: add outline and focusability

WofWca:
See commit messages

I tested it. Appears to be working good, but some rarely used elements might get ugly even for mouse-only navigation.

This by no means makes DC fully keyboard-accessible, but it's an improvement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ## Added
 - More shortcuts to switch between chats: `Ctrl + PageDown`, `Ctrl + PageUp`, `Ctrl + Tab`, `Ctrl + Shift + Tab` #3984
+- Better keyboard accessibility: make more elements focusable, add outline them #4005
 
 ### Changed
 - reword advanced setting "Disable Background Sync For All Accounts" -> "Only Synchronize the Currently Selected Account" #3960
--- a/scss/_global.scss
+++ b/scss/_global.scss
@@ -71,6 +71,12 @@ a {
 
 $unread-badge-size: 21px;
 
+// Let's not add `:not(:focus-visible)` since the blinking cursor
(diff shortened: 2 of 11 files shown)


Simon-Laux:
seems to work though the majority of the app still isn't focusable:
- accounts in sidebar
- tabs (qr code scanner dialog, gallery
- items in gallery
- Images in messages
- toggle switches in settings
- appearance settings, theme selector button and wallpapers

We can also track those in a new issue.

I don't really like the visual style of the outline (orange + white) yet, but maybe I'll get used to it.

also outline should probably have the highest z-index so it is always fully visible:
Bildschirmfoto 2024-07-05 um 00 44 02
(in the settings it's currently cut off)

Simon-Laux (approved):

r10s:
is the outline always visible or only after tab is hit?

is there a selector for high-contrast or accessibility maybe? what do others (signal, whatsapp) do here?

WofWca:
> highest z-index so it is always fully visible

I'm not sure if there is a way to set Z index on the outline, but you can try to set negative `outline-offset` for such cases. Or just highlight the item some other way, e.g. with the blackground color change.
The way I think about this change is to have accessibility by default for all elements.

> is the outline always visible or only after tab is hit?

Only after Tab is hit.

> is there a selector for high-contrast or accessibility maybe?

I believe so.

WofWca:
> what do others (signal, whatsapp) do here?

Signal appears to behave the same, but they do it programmatically, by [dynamically adding `.keyboard-mode` class](https://github.com/signalapp/Signal-Desktop/blob/b34f2fbb99d68bd462715f8519524100580fb372/ts/services/InteractionMode.ts#L24).

Issue #3868, opened 2024-05-26
Settings should be closable in sub-pages with escape key


Simon-Laux:
Or go back first to main settings page and then close on the second press, I assume.

r10s:
> and then close on the second press, I assume.

yip, the "Esc"-key should hit the topmost's dialog "back button"

and if if there is none, close the topmost dialog

maxphilippov:
Is #2268 related? In terms of we need to manage how dialogs are stacked and which one is closed by ESC. I can take that after resolving more pressing bugs since I'm reworking Dialog as part of BlueprintJS removal

friedmag:
I was looking at this as a first issue, and can confirm it's fixed by the commit above.

Simon-Laux:
meanwhile this was solved in #4702

PR #3977, opened 2024-06-25
Remove BP dep from Gallery

farooqkz:
This PR aims to make Gallary working without BluePrintJs. Also refactors are done as needed.

PR #4006, opened 2024-07-04
blueprintjs removal

maxphilippov:
This is an attempt to remove BlueprintJS from the project. This is pretty huge in terms of what it changes so I might have missed something, but I believed that the things came out pretty simple and manageable. What I saw during this cleanup is that we've redefined a lot of BP styling etc. making it almost our own implementation.

BlueprintJS is gone from package.json so you have to reinstall node_modules to check this properly.

Things missing:
1) I didn't handle the Escape key bind to close new Dialog. I saw other discussions (#2268, #3868) regarding how we should handle a stack of open windows and close only the latest one.
2) I didn't remove bp4- class overrides (almost all of them done in either `dev_rocket` or `dev_minimal` themes), so this is definitely an open discussion (should we make up new names, add these classes to new elements or we can come up with something else). If you worked on these themes, please hit me up.

It also opens up an easier way to update React since we don't have to manage BlueprintJS too.


WofWca:
About dialogs: not sure if you're aware, but now HTML has a native [`<dialog>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) which handles all these dismiss and stack things pretty well.

src/renderer/components/Login-Styles.tsx
```diff
@@ -205,45 +194,51 @@ export const DeltaPasswordInput = React.memo(
...
+      </>
     )
   }
 )
 
 export const DeltaProgressBar = function (
   props: React.PropsWithChildren<{
     progress: number
-    intent?: Intent
+    intent?: string
```
maxphilippov, line 230: I forgot to actually implement that, commit incoming

farooqkz:
I can't do a review on the PR today(can't concentrate due to bad sleep pattern). But my opinion is mixed. You are doing BP removal which is desired but I am not in favor of big PRs as they need extensive testing and review. IMO, in such a bug codebase, a big PR might do more bad than good unless the review and testing is really extensive. Nevertheless, As I said removal of BP is very much wanted as a major upgrade of React and many other changes are softly dependent on it.

farooqkz:
BTW, It's good that you haven't touched the Gallery component as there is another PR for it.

maxphilippov:
> I can't do a review on the PR today(can't concentrate due to bad sleep pattern). But my opinion is mixed. You are doing BP removal which is desired but I am not in favor of big PRs as they need extensive testing and review. IMO, in such a bug codebase, a big PR might do more bad than good unless the review and testing is really extensive. Nevertheless, As I said removal of BP is very much wanted as a major upgrade of React and many other changes are softly dependent on it.

Yeah, I don't think it's a good idea to dive into code straight away, I'd rather have people checking all the interfaces they use daily and look at what broke in terms of layout.

I know it looks like a huge change but from what I saw we used BP just for markup, the only thing that has some kind of interactive behavior is Dialog (we lost a lot of interactive things after replacing map with a new one opening in a popup). So it should be pretty obvious if something broke by just the look of it. I'm trying to do a screen by screen comparison, but that still requires some time.

Simon-Laux (changes requested):
Bugs from testing:
- [ ] clicking on any page in settings just closes the dialog and does not work.
- [ ] new group in create chat dialog has the same issue
- [ ] clicking on contacts in group edit/view dialog also closes the dialog and does not do what it is supposed to do

- [ ] Icons in gallery look different
Bildschirmfoto 2024-07-04 um 22 42 29
Bildschirmfoto 2024-07-04 um 22 42 53

should not be blue

- [ ] tabs are invisible in dark theme
Bildschirmfoto 2024-07-04 um 22 44 05

- [ ] Dialog footer items in add as second device are not aligned at the start and end.
Bildschirmfoto 2024-07-04 um 22 45 49

scss/_icons.scss
```diff
@@ -14,19 +14,6 @@
...
-  }
-}
-
-.remove-icon:hover {
-  .bp4-icon {
-    & > svg {
-      fill: #bf0000;
-    }
-  }
 }
```
Simon-Laux, line 17: was the removal of the hover effect intentional?

Simon-Laux:
I think we should forget about using Overlay for dialog and try to use the `<dialog>` element that the browser already provides, it even already has a `::backdrop` css selector.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog

I guess we should first try to use them with the open attribute without showModal, that we don't need to rewrite the dialog manager, but if that doesn't work we might need to rewrite it. also I wonder if the html5 dialog element supports multiple dialogs on top of each other.

If the `<dialog>` element doesn't work then we need to adjust the overlay component:
- make sure the `onClick` handler is on the backdrop and renamed to `onBackdropClick`
- make sure the overlay content is centered in a way that it does not block the backdrop from being clicked.

farooqkz:
@maxphilippov  after you addressed the bugs Simon has found, ping me to do a review by using.

maxphilippov:
> I think we should forget about using Overlay for dialog and try to use the `<dialog>` element that the browser already provides, it even already has a `::backdrop` css selector. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog
> 
> I guess we should first try to use them with the open attribute without showModal, that we don't need to rewrite the dialog manager, but if that doesn't work we might need to rewrite it. also I wonder if the html5 dialog element supports multiple dialogs on top of each other.
> 
> If the `<dialog>` element doesn't work then we need to adjust the overlay component:
> 
> * make sure the `onClick` handler is on the backdrop and renamed to `onBackdropClick`
> * make sure the overlay content is centered in a way that it does not block the backdrop from being clicked.
> 
> (thats the bug that prevents me from using the settings dialog)
also replying to @WofWca 

So I've looked into that and I personally find the <dialog> very rough and actually bad to use.

I pushed a reference implementation I used to figure things out.

My biggest problem with it is the desing which is kinda orthogonal to React idea of 'you specify element visibility by its presence in the DOM'.

This one is stateful and that makes it do excessive work which we don't even use.

Our current system (and I would argue it's a sane one) has a simple principle:
* there's no 'closed' state for dialog, if it's in DOM it's open

1. There's only one way to open <dialog> through markup and it's the only property 'open':
that gives us neither backdrop nor 'close on escape'
So we have to do useEffect and call showModal() every time, cause we don't have a 'closed' state.

Do we need a 'closed' state? I don't think so. That would require either to have all dialogs present in DOM or to have a manager to cleanup unused dialogs. I don't see any of these being useful.

We also need to have CSS-display: flex on the dialog, which conflicts with how <dialog> manages display.

2. We already have a stack of opened dialogs in Dialog Context and it's implemented as natural as possible with array.
So <dialog> storing it's own order is again excessive work. 

What I'm doing in reference impl. is using it's onClose callback to send a message to our DialogContext it's time to delete this one right after it get's hidden.

3. I'm not an expert on accessibility but it seems like it doesn't even set aria-modal="true"

maxphilippov:
> Bugs from testing:
> 
> * [ ]  clicking on any page in settings just closes the dialog and does not work.
> * [ ]  new group in create chat dialog has the same issue
> * [ ]  clicking on contacts in group edit/view dialog also closes the dialog and does not do what it is supposed to do
> * [ ]  Icons in gallery look different
> 
> Bildschirmfoto 2024-07-04 um 22 42 29 Bildschirmfoto 2024-07-04 um 22 42 53
> should not be blue
> 
> * [ ]  tabs are invisible in dark theme
> 
> Bildschirmfoto 2024-07-04 um 22 44 05
> * [ ]  Dialog footer items in add as second device are not aligned at the start and end.
> 
> Bildschirmfoto 2024-07-04 um 22 45 49
> * [ ]  tabs in scan qr code dialog look different
> 
> Bildschirmfoto 2024-07-04 um 22 57 23

Most of the dialog things should be fixed, but there's also a clear discrepancy between our styles. I didn't have a problem with flex dialog footer/header. But I had issues with basic text color in dialog. Can you re-check those please?

scss/_icons.scss
maxphilippov: Not intentional, but I moved it to `<Icon coloring` in latest commit

nicodh:
@maxphilippov could you please go through the comments and answer Simons in detail (unresolved conversations)
For example which of the reported bugs are fixed.
Would be helpful for reviews...

farooqkz:
@maxphilippov I think it's a good idea that I finish my Gallery BP removal and that you base your PR on it? Because one of the bugs @Simon-Laux has found directly is part of Gallery. Plus I'm doing some refactors there and introducing "Tabs" component which you could use.

Simon-Laux:
max pr is bigger so I guess the other way around would make more sense?

maxphilippov:
Rebased towards current main (to the monorepo version). Got some minor issues with colors, gonna look at these later and post here an update.

packages/frontend/src/components/Dialog/Dialog.tsx
```diff
@@ -29,13 +27,52 @@ const Dialog = React.memo<Props>(
     height,
     ...props
   }) => {
+    canEscapeKeyClose
```
WofWca, line 30: What's this line

packages/frontend/src/components/Dialog/Dialog.tsx
```diff
@@ -29,13 +27,52 @@ const Dialog = React.memo<Props>(
...
+    const onCancel = (ev: React.BaseSyntheticEvent) => {
+      if (!canEscapeKeyClose) {
+        ev.preventDefault()
+      }
+    }
+
+    useEffect(() => {
+      // calling showModal is "only" the way to have ::backdrop
+      dialog.current?.showModal()
+      dialog.current!.style.display = 'flex'
```
WofWca, line 67:
```suggestion
      dialog.current!.style.display = ''
```

So that CSS styles take over.

packages/frontend/src/components/Dialog/Dialog.tsx
```diff
@@ -29,13 +27,52 @@ const Dialog = React.memo<Props>(
...
+    const onClose = () => {
+      dialog.current!.style.display = 'none'
+      props.onClose()
+    }
+
+    const onCancel = (ev: React.BaseSyntheticEvent) => {
+      if (!canEscapeKeyClose) {
+        ev.preventDefault()
+      }
+    }
```
WofWca, line 62: Isn't this also supposed to be invoked when the "Cancel" button is pressed?

Simon-Laux:
Bildschirmfoto 2024-10-03 um 12 32 28
create chat dialog / dialog footer actions - should be space between the items.

Will review the rest later

nicodh:
@maxphilippov ist this final now for review? You mentioned on monday you will fix some issues?

maxphilippov:
> @maxphilippov ist this final now for review? You mentioned on monday you will fix some issues?

@Simon-Laux yeah, this is probably not ready to test yet, I had some things messed up during rebase, I did force-push some fixes but there are couple issues left. I'll change it from draft when I'm done. Hope to do that by the end of Friday/before Monday morning

maxphilippov:
So, that's my fault of not investigating <dialog> enough. It introduces us to the world of 'top-layer', where there's no way to bypass dialogs other than Popover API, Fullscreen element, and some specific animations. Fullscreen doesn't make sense for elements I'm gonna talk about later and Popover API isn't officially in React type-system (it's supposedly coming in v19).

Here's the chain effect of introducing `<dialog>`:

it immediately broke `FullscreenMedia` and `FullscreenAvatar` (when you open your profile picture it opens behind profile dialog) so I had to implement these using <dialog> too.

Which in turn broke `ContextMenu` (again it always gonna be behind either dialog or `FullscreenMedia`) so it's implemented using <dialog> too.

I still decided to go for the <dialog> for the promise of accessibility. 

But there's a catch again:

Dialog picks an element to focus on when it's opened. Looks like it cannot be disabled and the only thing left is to either unfocus or disable styling [here's a relevant discussion](https://stackoverflow.com/a/79028606). As a keyboard user I'm pretty fine with that, the only questionable thing for me is how it's rendered in Settings. 
image

So the bigger question is "can we integrate focus organically in UI or that's confusing?"

packages/frontend/src/components/Dialog/Dialog.tsx
maxphilippov: I didn't go over all the dialogs to change how buttons behave with `formmethod="dialog"` (and I suppose we'll have to wrap dialog content inside `<form>`), so cancel doesn't matter much here, the dialog closes anyway.

Also there's this fine line between close and cancel, where we probably have to discuss our policy, but after we clean up all the issues with dialog chaining. What I also didn't touch is the dialog return value.

packages/frontend/src/components/Dialog/Dialog.tsx
maxphilippov: It's been a while I'll look it up again, but I believe there were some forms which didn't play nicely with the default 'block' display, I'll recheck that. And my bad for not leaving a comment here.

WofWca: What I'm saying is that `flex` should be defined statically in CSS, and when the dialog needs to be hidden, let's apply `style.display = 'none'`.

maxphilippov:
> > I had to implement these using too
> 
> What does this mean?
> Shouldn't this be fixed by putting `FullscreenMedia` and `FullscreenAvatar` inside the dialog element?

Yeah, these are wrapped inside a special "fullscreen transparent" dialog.

> How is this a problem? Why do we need to unfocus?

image

I'm talking about this image for example.

maxphilippov:
> Bildschirmfoto 2024-07-04 um 22 42 29 Bildschirmfoto 2024-07-04 um 22 42 53
> should not be blue
> 
> * [ ]  tabs are invisible in dark theme

I've actually missed this one, gonna fix a bit later.

Simon-Laux:
About dialog focus: I think this is also what operating systems and other programs do: they focus a dialog button on confirmation and alert dialogs.

But that is not what we want for dialogs like fullscreenmedia view and settings dialog. There we probably want to focus the entire dialog element instead.

See https://github.com/whatwg/html/wiki/dialog--initial-focus,-a-proposal#use-case-analysis

maxphilippov:
> operating systems and other programs do: they focus a dialog button on confirmation and alert dialogs.
> 
> But that is not what we want for dialogs like fullscreenmedia view and settings dialog. There we probably want to focus the entire dialog element instead.

Yeah, but the issue I guess is how to display that. That's why I'm asking if we are willing to integrate that in an UI, cause I can see how surrounding a close button with a thick orange rectangle might not be appealing.

Shouldn't we though focus "close" button in fullscreenmedia if there's only one image and "next/prev" if there are multiple?

packages/frontend/scss/misc/_context_menu.scss
```diff
@@ -1,22 +1,30 @@
 .dc-context-menu-layer {
   width: 100vw;
   height: 100vh;
```
WofWca, line 3:
```suggestion
  width: 100vw;
  height: 100vh;
  // Built-in Chromium styles set `max-height` and `max-width`
  // on `dialog:-internal-dialog-in-top-layer` to `calc(100% - 2em - 6px)`.
  // Let's override that, otherwise the "backdrop" area doesn't
  // fully cover the screen.
  max-height: unset;
  max-width: unset;
```

WofWca (changes requested):
I'm not finished reviewing yet, but this is what I have so far.

Some of my comments are nitpicky, so I'm not gonna insist on them getting tackled right now.

Another few issues I found:
- In fullscreen media, a scrollbar appears (see on the right):
    image
- If you open settings, then Tab-navigate to "Chats and Media" and press spacebar, the settings dialog will simply close.  
    I checked whether this is related to the absence of `type="button"` on the buttons, but it's not the case.

packages/frontend/src/components/Dialog/Dialog.tsx
```diff
@@ -27,25 +27,74 @@ const Dialog = React.memo<Props>(
...
+            rect.top <= ev.clientY &&
+            ev.clientY <= rect.top + rect.height &&
+            rect.left <= ev.clientX &&
+            ev.clientX <= rect.left + rect.width
+          if (!isInDialog) {
+            const cancelEvent = new Event('cancel')
+            dialog.current.dispatchEvent(cancelEvent)
+            // cancel event doesn't trigger dialog close
+            dialog.current.close()
+          }
```
WofWca, line 54: > If you open settings, then Tab-navigate to "Chats and Media" and press spacebar, the settings dialog will simply close.

This is happening because of this code. When you use `spacebar` to click, `clientY` and `clientX` would be unset, and it will close the dialog.

packages/frontend/scss/login/_login.scss
```diff
@@ -82,45 +87,84 @@ div.delta-form-group {
...
+        inset 0 1px 1px rgba(17, 20, 24, 0.5);
+      color: var(--bp4InputText);
+      font-size: 14px;
+      font-weight: 400;
+      height: 30px;
+      line-height: 30px;
+      outline: none;
+      padding: 0 10px;
+      transition: box-shadow 100ms cubic-bezier(0.4, 1, 0.75, 0.9);
+      vertical-align: middle;
```
WofWca, line 165: A lot of the newly added styles are overridden right below, e.g. `font-size`.

packages/frontend/scss/message/_message-detail.scss
```diff
@@ -3,10 +3,7 @@
 .module-message-detail {
   margin-top: -20px;
 
-  .bp4-callout {
-    max-height: 50vh;
-    overflow: auto;
-  }
+  max-height: 55vh;
```
WofWca, line 6: This is weird. `max-height` is set, but then the element overflows anyway, which I think results in padding-bottom not being respected. Why do we apply it to the parent element now, instead of the `callout`?

image

packages/frontend/scss/misc/_context_menu.scss
```diff
@@ -1,22 +1,30 @@
...
   user-select: none;
 
-  &.active {
-    pointer-events: all;
+  pointer-events: all;
+
+  // default backdrop has almost transparent black background
+  &::backdrop {
+    pointer-events: none;
+    background: none;
```
WofWca, line 21:
```suggestion
  &::backdrop {
    display: none;
```

Wouldn't just this be simpler? I tested it a bit. It seems to still work alright.

packages/frontend/scss/manifest.scss
```diff
@@ -2,12 +2,9 @@
 
 // import node_module dependencies (without ".css" means include)
 @import '../node_modules/normalize.css/normalize';
-// @import '../node_modules/@blueprintjs/icons/lib/css/blueprint-icons'; - did not exist anymore?
-@import '../node_modules/@blueprintjs/core/lib/css/blueprint';
 
 // Overwrites for third party node modules.
 // Above our styles, so that our styles take priority by being defined after.
```
WofWca, line 7:
```suggestion
```

packages/frontend/src/components/screens/AccountSetupScreen.tsx
```diff
@@ -78,23 +80,23 @@ export default function AccountSetupScreen({
     }
   }, [onKeyDown])
 
+  const onCancel = () => selectAccount(accountId)
+
   return (
     <ImageBackdrop variant='welcome'>
-      <Dialog onClose={() => {}} fixed>
+      <Dialog canOutsideClickClose={false} onClose={onCancel} fixed>
```
WofWca, line 87: Isn't `onClose` also invoked when the dialog is closed with a success? I believe originally `onClose={() => {}}` meant to prevent the dialog from being closed when clicking outside. Why was `onClose={onCancel}` added?

packages/frontend/src/components/screens/WelcomeScreen/index.tsx
```diff
@@ -41,7 +41,7 @@ export default function WelcomeScreen({ selectedAccountId, ...props }: Props) {
 
   return (
     <ImageBackdrop variant='welcome'>
-      <Dialog fixed onClose={onClose} width={400}>
+      <Dialog fixed onClose={onClose} width={400} canOutsideClickClose={false}>
```
WofWca, line 44:
```suggestion
      <Dialog fixed width={400} canOutsideClickClose={false}>
```

Also need to remove the `onClose` listener above.

WofWca (changes requested):
I finished reviewing the current version.

I found a few more problems.

Also let's remember that the Gallery is undone and looks quite ugly. This is supposed to be handled by a separate MR: #3977

And I shall note that dialog transitions are gone. I am fine with this. This can be done later.

Apart from the issues I pointed out, all seems to be working nicely, and I tested a lot of things. Good work!

packages/frontend/src/components/Dialog/Dialog.tsx
```diff
@@ -38,6 +34,10 @@ const Dialog = React.memo<Props>(
           if (!dialog.current) {
             return
           }
+          // this is to filter out when pressing keys trigger `onClick`
```
WofWca, line 37:
```suggestion
          // pressing a button with Spacebar inside a dialog
          // triggers the `onClick` event.
          // Let's ignore such "clicks" here.
```

packages/frontend/scss/manifest.scss
WofWca: What about this?

packages/frontend/src/components/Switch/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const inputRef = useRef<HTMLInputElement>(null)
+
+  const toggle = () => {
+    if (!inputRef.current) {
+      return
+    }
+    const prev = !!inputRef.current.checked
+
+    inputRef.current.checked = !prev
+  }
```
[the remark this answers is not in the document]
WofWca: It still doesn't work.

packages/frontend/src/components/RadioGroup.tsx
```diff
@@ -17,9 +17,10 @@ export default function RadioGroup({
   return (
     <form>
       <fieldset className='radiogroup'>
-        {children.map((radio: any) => {
+        {children.map((radio: any, index: number) => {
           return (
             <Radio
+              key={index}
```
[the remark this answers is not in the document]
WofWca: Why is this still unchanged?

packages/frontend/scss/gallery/_media.scss
```diff
@@ -14,12 +14,28 @@ $tab-height: 46px;
...
+      font-size: 1.1rem;
+      padding-bottom: 6px;
+      &.selected,
+      &:hover {
+        color: #42a5f5;
+        border-bottom: 3px solid #42a5f5;
+        cursor: pointer;
+      }
+    }
+  }
```
WofWca, line 37: The tabs jump in size.

https://github.com/user-attachments/assets/737e2881-885e-4327-a8a5-2effd083182e

Though it is bearable I suppose.

WofWca (changes requested):
Some issues that I pointed out before are still unresolved.

packages/frontend/scss/gallery/_media.scss
nicodh: I added a flex grow 1 element to avoid layout jump and flex gap to have independant margins between the tab titles

packages/frontend/src/components/Switch/index.tsx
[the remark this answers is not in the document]
nicodh: Works now

nicodh (approved):

Issue #3416, opened 2023-09-27
Make sure WebXDC cannot access the clipboard


link2xt:
See discussion at https://github.com/deltachat/deltachat-desktop/pull/3413#discussion_r1337826506

hpk42:
On Wed, Sep 27, 2023 at 05:48 -0700, link2xt wrote:
> See discussion at https://github.com/deltachat/deltachat-desktop/pull/3413#discussion_r1337826506

i don't understand -- when i am opening the editor i *want* to be able to paste something into it.
Are you suggesting to disable this ability completely?

hpk42:
On Thu, Sep 28, 2023 at 02:15 -0700, link2xt wrote:
> WebXDC app should not be able to access the clipboard without explicit user action. Otherwise it is possible to make an app that posts your clipboard to the public board every time you open it.
> 
> `execCommand('paste')` should not work: https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand
> 
> New [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API) require a `clipboard-read` permission, this should be asked for each app or disabled.

if i go to any https-website, can they get at the content of my clipboard in default browser? 
Or is that only possible for browser-controlled input/edit-fields ?

Simon-Laux:
we can try to overwrite it, but then we have the iframe problem again:
```
iframe.contentWindow.document.execCommand("paste")
```

btw it still works in electron 26.

also the clipboard api does not work, because it's permission is denied, somehow `execCommand` works regardless.

WofWca:
I looked through Blink features ([here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/public/common/features.h) and [here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/runtime_enabled_features.json5) and was unable to find anything related. And I couldn't find anything useful in Electron docs. So I created an issue: https://github.com/electron/electron/issues/45237.

r10s:
how is the electron update related to this issue? 

at a first glance, i could not find things related to the clipboard in the linked electron pages

ell1e:
> if i go to any https-website, can they get at the content of my clipboard in default browser?

[...]

PR #4008, opened 2024-07-04
WIP: feat: a way to grant permissions to webxdcs

WofWca:
Closes https://support.delta.chat/t/allow-access-to-camera-geolocation-other-web-apis/2446?u=wofwca

image
image

TODO:
- [x] Move this behind an experimental setting

### DIFF
--- a/packages/target-electron/src/deltachat/webxdc.ts
+++ b/packages/target-electron/src/deltachat/webxdc.ts
@@ -260,12 +260,83 @@ export default class DCWebxdc extends SplitOut {
         internet_access: webxdcInfo['internetAccess'],
       }
 
+      type setPermissionRequestHandler =
+        typeof webxdcWindow.webContents.session.setPermissionRequestHandler
+      type permission_arg = Parameters<
+        Exclude<Parameters<setPermissionRequestHandler>[0], null>
+      >[1]
+      // TODO some (poorly written?) apps might require a refresh
+      // after a permission has been granted,
+      // but we don't support it because
+      // 1. There is no way to refresh a webxdc
+      // 2. There is the WebRTC 500 exhaustion hack which doesn't allow the app
+      //     to load. We should wait there instead of immediately closing.
+      const grantedPermissions = new Set<permission_arg>([
+        // because games might lock the pointer
+        'pointerLock',
+        // games might do that too
+        'fullscreen',
+      ])
+      const userControllablePermissions: permission_arg[] = [
+        // TODO should add more permissions, I added just a few for a POC.
+        //
+        // Or, we should instead dynamically add items here as the app
+        // makes permission requests (see `permission_handler`).
+        // Would we have to `window.setMenu()` each time?
+        // Can we utilize `visible: boolean` attribute instead?
+        'media',
+        // TODO `navigator.geolocation.getCurrentPosition(console.log, console.error)`
+        // fails because
+        // `Network location provider at 'https://www.googleapis.com/' : ERR_NAME_NOT_RESOLVED.`
+        'geolocation',
+        'notifications',
+        'display-capture',
+        // These are allowed by default, see `grantedPermissions = ` above.
+        // TODO Should we remove them from here as to not confuse the user?
+        'fullscreen',
+        'pointerLock',
+      ]
+
       const isMac = platform() === 'darwin'
 
       const menu = Menu.buildFromTemplate([
         ...(isMac ? [getAppMenu(webxdcWindow)] : []),
         getFileMenu(webxdcWindow, isMac),
         getEditMenu(),
+        // TODO consider whether putting this inside some other menu
+        // (say, "Edit") is reasonable, though I don't think so.
+        ...(DesktopSettings.state.enableWebxdcPermissionManagement
+          ? [
+              {
+                label: tx('menu_webxdc_permissions'),
+                submenu: userControllablePermissions.map(permissionName => ({
+                  // TODO proper names and tooltips for permissions + i18n
+                  label:
+                    permissionName === 'media'
+                      ? 'Camera and microphone'
+                      : permissionName,
+                  type: 'checkbox',
+                  // toolTip:
+                  // visible:
+                  // sublabel:
+
+                  checked: grantedPermissions.has(permissionName),
+                  id: `webxdc_permission_${permissionName}`,
+                  click: () => {
+                    const newIsGranted = !grantedPermissions.has(permissionName)
+                    if (newIsGranted) {
+                      grantedPermissions.add(permissionName)
+                    } else {
+                      grantedPermissions.delete(permissionName)
+                    }
+                    menu.getMenuItemById(
+                      `webxdc_permission_${permissionName}`
+                    )!.checked = newIsGranted
+                  },
+                })),
+              } as MenuItemConstructorOptions,
+            ]
+          : []),
         {
           label: tx('global_menu_view_desktop'),
           submenu: [
@@ -359,6 +430,9 @@ export default class DCWebxdc extends SplitOut {
       webxdcWindow.once('close', () => {
         const lastBounds = webxdcWindow.getBounds()
         setLastBounds(this, accountId, msg_id, lastBounds)
+        // TODO shall we save permissions for next session?
+        // Probably when we implement explicit user consent for this,
+        // like in browsers.
       })
 
       webxdcWindow.once('ready-to-show', () => {})
@@ -380,11 +454,6 @@ export default class DCWebxdc extends SplitOut {
         ev.preventDefault()
       })
 
-      type setPermissionRequestHandler =
-        typeof webxdcWindow.webContents.session.setPermissionRequestHandler
-      type permission_arg = Parameters<
-        Exclude<Parameters<setPermissionRequestHandler>[0], null>
-      >[1]
       const loggedPermissionRequests: { [K in permission_arg]?: true } = {}
       /** prevents webxdcs from spamming the log */
       const logPermissionRequest = (permission: permission_arg) => {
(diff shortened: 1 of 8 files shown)


src/main/deltachat/webxdc.ts
```diff
@@ -248,13 +248,80 @@ export default class DCWebxdc extends SplitOut {
...
+        // because games might lock the pointer
+        'pointerLock',
+        // games might do that too
+        'fullscreen',
+      ])
+      const userControllablePermissions: permission_arg[] = [
+        // TODO should add more permissions, I added just a few for a POC.
+        //
+        // Or, we should instead dynamically add items here as the app
+        // makes permission requests (see `permission_handler`).
```
Simon-Laux, line 272: would rather make the app specify it in its manifest.toml file

src/main/deltachat/webxdc.ts
```diff
@@ -248,13 +248,80 @@ export default class DCWebxdc extends SplitOut {
...
+        // Can we utilize `visible: boolean` attribute instead?
+        'media',
+        // TODO `navigator.geolocation.getCurrentPosition(console.log, console.error)`
+        // fails because
+        // `Network location provider at 'https://www.googleapis.com/' : ERR_NAME_NOT_RESOLVED.`
+        'geolocation',
+        'notifications',
+        'display-capture',
+        // These are allowed by default, see `grantedPermissions = ` above.
+        // TODO Should we remove them from here as to not confuse the user?
```
Simon-Laux, line 283: I guess yes

Simon-Laux:
we could show an electron/native dialog in` webxdcWindow.webContents.session.setPermissionRequestHandler` when the permission is requested the first time.
Also would make sense to persist webxdc permissions, maybe for now in `ui.*` config key. like `ui.webxdc_permissions.<messageId>`. later it would be nice to have a core api for storing permissions per webxdc instance, so that core can automatically delete the permission config when the webxdc is deleted.

src/main/deltachat/webxdc.ts
```diff
@@ -248,13 +248,80 @@ export default class DCWebxdc extends SplitOut {
...
+        // makes permission requests (see `permission_handler`).
+        // Would we have to `window.setMenu()` each time?
+        // Can we utilize `visible: boolean` attribute instead?
+        'media',
+        // TODO `navigator.geolocation.getCurrentPosition(console.log, console.error)`
+        // fails because
+        // `Network location provider at 'https://www.googleapis.com/' : ERR_NAME_NOT_RESOLVED.`
+        'geolocation',
+        'notifications',
+        'display-capture',
```
Simon-Laux, line 281: I'd remove `notifications` and `display-capture` for now:
- `notifications` -> webxdc will have it's own api for notifications and I think this notification api want network access
-  `display-capture` - probably does not work without making a custom ui for choosing what to share?

hpk42:
1) It's not agreed that we want to generically allow access to such APIs
as it can open a can of worms and cause further work on all platforms. 
Eying video/audio calls is certainly interesting but maybe better achieved
by an agreed cross-platform "integrations" approach instead of 
a generic permissions-one for all apps. 

2) As long as there are 53 open bugs (as of now on Desktop) 
and we need to maintain 1.46.X releaseability,
let's please focus on maintenance and stabilization first.

src/main/deltachat/webxdc.ts
Simon-Laux: That would allow xdc app stores to show permissions.
And we could also say you need to specify a reason string there that we can show in the permission request dialog.

WofWca: Hmmmm idk idk. I think a "website" model is more suitable. Where the website (webxdc in this case) itself would explain why it's going to request a permission, in its UI.

This "manifest" model is more suitable for when there is no way to request permissions dynamically, where you have to assess them prior to installation, but with all permissions defaulting to "denied" I think it's not needed.

WofWca:
Pushed a few changes:
- hid this behind an experimental toggle
- more possible permissions to grant
- don't show permissions that were never checked or requested
- security: check `requestingUrl` origin just in case
- adjust log text

src/main/deltachat/webxdc.ts
```diff
@@ -248,97 +249,189 @@ export default class DCWebxdc extends SplitOut {
...
+            | typeof webxdcWindow.webContents.session.setPermissionCheckHandler
+            | typeof webxdcWindow.webContents.session.setPermissionRequestHandler
+          >[0],
+          null
+        >
+      >[1]
+      // TODO some (poorly written?) apps might require a refresh
+      // after a permission has been granted,
+      // but we don't support it because
+      // 1. There is no way to refresh a webxdc
```
Simon-Laux, line 264: app can call `window.location.reload()` in it's iframe, you can only not reload the parent page.

Simon-Laux:
What I would like to see in this pr:
- [ ] clipboard read access as toggle-able permission, also loosely related to #3416 (when we find a way to disable the `execCommand("paste")` method)
- [ ] persist the permissions you have selected in ui config key (probably sth like key=`webxdc.permissions.<webxdc_id/msg_id>`)
- [ ] ask user automatically when app requests permission in an electron native dialog (because electron native dialog is the least effort) -> only ask one time for each permission, if the user has denied it don't ask again.

permission storage format could be
```ts
enum PermissionState {
 GRANTED,
 USER_DENIED, // denied, don't ask again
 UNDEFINED // denied, but ask on first request, not really in that enum, use the normal undefined instead
}

type PermissionStore = {[permission:string]: PermissionState | undefined}
```

Simon-Laux:
For other platforms:

I don't think waiting for integrations makes much sense, they come with their own questions and challenges.

The ui does not need to be complicated, we can just start with the following:

- Experimental setting to turn on the permission system
- For now only implement the web standard permissions such as camera/microphone/media and clipboard read.
- UI:
   - in the 3 dot menu on iOS and Android webxdc add a new entry "Permissions" that opens a native dialog with checkboxes that is the ui to change the preferences (on iOS it could also be an action sheet, like how we did the experimental settings there in the past)
   - additionally when the app requests permission for something the first time show an confirmation dialog asking the user to allow / deny it -> only show the dialog if the user has not denied that permission for that app before.
 - Store settings in ui config for now as I described in my last post in this thread.

- implementation: on both android and electron there is a permission handler function/callback that is called when the website requests a web permission, the handler can then accept or deny the request:
   - android: [`WebChromeClient.onPermissionRequest`](https://developer.android.com/reference/android/webkit/WebChromeClient#onPermissionRequest(android.webkit.PermissionRequest)) and  [`WebChromeClient.onPermissionRequestCanceled `](https://developer.android.com/reference/android/webkit/WebChromeClient#onPermissionRequestCanceled(android.webkit.PermissionRequest)) API 21
- iOS only supports gyroscope and media (camera & microphone) permissions, there are 2 specific handlers instead of the one that android/electron has:
  - [webView(_:requestDeviceOrientationAndMotionPermissionFor:initiatedByFrame:decisionHandler:)](https://developer.apple.com/documentation/webkit/wkuidelegate/3795606-webview)
  - [webView(_:requestMediaCapturePermissionFor:initiatedByFrame:type:decisionHandler:)](https://developer.apple.com/documentation/webkit/wkuidelegate/3763087-webview)

Later we can improve it:
- define what web permissions the app can request, by manifest, like there needs to exist a key with a reason string, like in iOS.
- store the permissions somewhere else, like with a new core api?
- if we do crypto signing we could think about sharing permission settings for a signed app version instead of having it separate for each instance.

Simon-Laux:
I rebased it to main (after merging monorepo, I offered to rebase all prs that were made before that)

r10s:
sorry to be a killjoy:

this PR and this discussion here is tricky timing-wise.

there is no agreement at all about the way to proceed with permissions. and the PR does not help on any open, much more important issue - but it adds issues and work on all platforms. distracting development resources from far more important things.

(also, the [original idea of a "share location"](https://support.delta.chat/t/allow-access-to-camera-geolocation-other-web-apis/2446/5?u=wofwca) webxdc seems to be a bit lost on the way - how useful are _desktop_ permissions to make a “share location” thing where most desktop do not even know their locations. this underlines, that it is not just this single PR, but will keep lots of devs busy)

i am not saying that it does not make sense, but i doubt it makes sense _now_.

i suggest to move this PR to resurrection and reconsider later.

WofWca:
Do you think it would be okay to merge if we were to remove the toggle from the UI completely? so that it's only possible to enable for developers.

r10s:
i would postpone the pr fully, freeing development resources. otherwise there will still be discussion, pressure on other os etc., removing focus from other things - let alone bugs :)
developler can still use this PR for experiments

WofWca:
Closing because ain't nobody got time for it right now

src/main/deltachat/webxdc.ts
ell1e: Isn't it unsafe if the permission is asked inside the app? (Since the app shouldn't be allowed to enable this with a UI that isn't clearly outside of it.)

WofWca: I'm suggesting to implement what normal browsers and websites do. The way this MR is currently implemented, you don't get any indication that the app is requesting a permission. This is not the case for websites, where you get a popup.

ell1e: Web sites are not allowed to create popups without address bars for security reasons however, while  the permission popups in browsers commonly are positioned to overlay areas outside the website canvas (which a website can't do) to indicate that it can't be possibly some fake popup by the website. Otherwise, the app could just conceal the permission and pretend you're clicking something else.

So the permission dialog should never be possibly created or handled by whatever runs inside the app window. This didn't sound like your "the website itself would explain" sounded like.

WofWca: I think I see you point. I'm not suggesting to put the dialog inside the app. What I meant is that apps should say "I am going to ask for this permission", after which they request it, which pops up the Delta Chat dialog where the user ultimately decides whether to grant the permission.

Either way the comment is outdated, I implemented the suggestion of dynamically adding the permissions to the menu as the app requests them.

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -118,6 +118,7 @@ export default class DCWebxdc extends SplitOut {
 
       const ses = sessionFromAccountId(accountId)
       const appURL = `webxdc://${accountId}.${msg_id}.webxdc`
+      const appOrigin = new URL(appURL).origin
```
WofWca, line 121: This is not good. This is `'null'`

Issue #3982, opened 2024-06-27
Notifications don't always disappear when message is read


WofWca:
- Operating System (Linux/Mac/Windows/iOS/Android): Windows 10
- Delta Chat Version: 1.46.1
- Expected behavior: The notification disappears when its message is read.
- Actual behavior:
    - Sometimes the notification stays there.
    - And clicking on it sometimes does nothing. It disappears, but if you close and open the notifications panel again, the notifications are there again.
- Steps to reproduce the problem:
    1. 1. Open one account on Delta Chat desktop.
        2. Receive a message on another account.
        3. Wait for the notification to "slide to the right".
        4. Through the Delta Chat UI navigate to the new message and read it.
    2. 1. Open a chat in DC Desktop.
        2. Switch to another window.
        3. Receive a message on the aforementioned chat.
        4. Alt+Tab to the Delta Chat window.
        5. The notification will stay there (as described), but if you switch to another chat and back to the current one, it will disappear.
- Screenshots:
    
    https://github.com/deltachat/deltachat-desktop/assets/39462442/22c85292-918a-4619-b612-02f0f718ac16

- Logs:

It seems that notifications that just appeared and stay on the screen for a few seconds behave differently from notifications that are in the "notifications panel".
So I might be describing several separate bugs there, but it all comes down to notifications not working well (at least on Windows)

WofWca:
This could be related #2438

WofWca:
Let's close as a duplicate of https://github.com/deltachat/deltachat-desktop/issues/3937

PR #4010, opened 2024-07-05
Ignore notification close event & clear notifications when marking accounts unread chat as noticed

### LINKED ISSUE #4002: "Mark All as Read" does not remove notifications
the gist of  "Mark All as Read" is to get chats out of ones way. therefore, and for consistency, also notifications should be removed when things are marked as read.

this issue may be one cause of stale notifications of https://github.com/deltachat/deltachat-desktop/issues/3991 , however, i doubt it is the only one as i was not using "Mark All as Read" until recently (i was not aware it exist :)

cmp https://github.com/deltachat/deltachat-desktop/issues/3979

nicodh:
fixes #4002

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 - minor performance improvements #3981
 - fix chat list items (e.g. Archive) and contacts not showing up sometimes #4004
 - fix bug notifications not being removed on Mac  #4010
+- fix bug "Mark All as Read" does not remove notifications #4002
 
 <a id="1_46_1"></a>
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 - Fix the problem of Quit menu item on WebXDC apps closes the whole DC app #3995
 - minor performance improvements #3981
 - fix chat list items (e.g. Archive) and contacts not showing up sometimes #4004
+- fix bug notifications not being removed on Mac  #4010
 
 <a id="1_46_1"></a>
 
--- a/src/main/notifications.ts
+++ b/src/main/notifications.ts
@@ -79,12 +79,14 @@ function showNotification(_event: IpcMainInvokeEvent, data: DcNotification) {
       notify.close()
     })
     notify.on('close', () => {
-      // this can be triggered by system out when the message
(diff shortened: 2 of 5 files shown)


nicodh:
I will test on Windows to resolve #3982

nicodh:
I also found an entry in the ChatListItemArchiveLink context menu "Mark al as read" which makes no sense imho since new messages in archived chats are not displayed as new at all.
@Simon-Laux is there a case where this makes sense? Otherwise I will remove it
ArchiveLink-Context-Menu

Simon-Laux:
mute a chat then archive it. then receive a message for it. archived chats show a badge counter that tells the number of archived muted chats with unread messages.
muted chats stay archived when a new message is received.
So this context menu is intentional and useful.

src/main/notifications.ts
```diff
@@ -79,8 +79,8 @@ function showNotification(_event: IpcMainInvokeEvent, data: DcNotification) {
       notify.close()
     })
     notify.on('close', () => {
-      notifications[chatId] =
-        notifications[chatId]?.filter(n => n !== notify) || []
+      // this is triggered when the message is moved to notification center
+      // so we don't remove the instance from the list
```
Simon-Laux, line 83: Electron docs says:

> Emitted when the notification is closed by manual intervention from the user.
> ~ https://www.electronjs.org/docs/latest/api/notification#event-close

So I think it makes sense to mention in the comment that this is a bug.

nicodh:
Tested on Windows, Mac & Linux

Works as expected, except a bug in Linux which is not related to deltachat (notifications are never removed from notification)
#/3982 ist not solved with this MR but not really related

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@
 - Fix the problem of Quit menu item on WebXDC apps closes the whole DC app #3995
 - minor performance improvements #3981
 - fix chat list items (e.g. Archive) and contacts not showing up sometimes #4004
+- fix bug notifications not being removed on Mac  #4010
```
Simon-Laux, line 29: you forgot to mention the fix for clear account notifications

PR #4011, opened 2024-07-05
fix: webxdc: CSP bypass

WofWca:
This doesn't appear to fix an exploitable vulnerability because `host-rules` is already set to disable network access, so "XDC-01-002 WP1" from the Cure53 audit has not been brought back ever since the initial fix.

Initially fixed in a9e5242acb
Reintroduced in fd1f8ce27b6e6

Related refactor commit 2cd310ec08f8f

Also bring back and improve the PDF comment about "XDC-01-005 WP1"

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 - Fix the problem of Quit menu item on WebXDC apps closes the whole DC app #3995
 - minor performance improvements #3981
 - fix chat list items (e.g. Archive) and contacts not showing up sometimes #4004
+- fix CSP bypass in webxdc (not a vulnerability) #4011
 
 <a id="1_46_1"></a>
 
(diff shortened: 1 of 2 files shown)


Simon-Laux (approved):

nicodh:
Concerning the comment:
"This doesn't appear to fix an exploitable vulnerability because host-rules is already set to disable network access,"

We should be aware that there is an ongoing discussion wether or not we should restrict internet to predefined URLs by host rules. It blocks some useful scenarios.

So we should not rely on host rules concerning XDC-01-002 WP1

Simon-Laux:
I disagree with removing host rules restriction

nicodh:
I just wanted to mention that there is a discussion. No reason to discuss it here.

nicodh:
At least it means that this fix is useful anyway :-)

nicodh (approved):

PR #4012, opened 2024-07-05
WIP: fix: chat "scrolling up" upon reaction

### LINKED ISSUE #3753: Adding a reaction makes the chat "scroll up" (then you have to scroll to see new messages)
- Operating System (Linux/Mac/Windows/iOS/Android): Windows, Linux
- Delta Chat Version: 1.44.0, 828f472a433a56bd77c25cb6f2aebcc556828b36
- Expected behavior: The scroll position remains at the very bottom of the chat
- Actual behavior: The chat becomes "scrolled up", gets "unlocked" from the bottom position
  1. Open a chat with someone, scroll to bottom
  2. Let them add a reaction, or add one yourself
  3. Wait for a new message to arrive
- Screenshots:

  https://github.com/deltachat/deltachat-desktop/assets/39462442/1f0adec3-6b10-4e21-a529-a5e522b199e6

- Logs: [none]

I think there should be a proper "declarative" fix. For example, I have noticed that in Chromium if you scroll to the bottom of the page and then increase the height of an element that is somewhere above, out of view, then the scroll height of the page will change, but the scroll position will remain at the bottom, unlike if the element was in view.

Update: oh lord oh please don't tell me this has to be fixed through scroll locking

https://github.com/deltachat/deltachat-desktop/blob/0380993ba6f65f7d52ef2acbb64a3779f9ebca22/src/renderer/stores/messagelist.ts#L309-L310

WofWca:
Closes #3753

TODO:
- [x] The biggest issue I noticed is that when the chat is scrolled to bottom and a new message appears, it causes a "scroll" event (which doesn't feel like it should be the case, since the all we did is just changed the scroll anchor), which causes the reaction bar to disappear if it was open (see `hideReactionsBar`).
Personally I think it is ok to get rid of this feature at least temporarily, because this scrolling thing is far more annoying that that feature is good

Otherwise I haven't noticed things that you could call issues.

This removes `scrollToBottom` JS on each new message, which is now unnecessary. However, the old code used `ifClose: true`, which would still scroll the chat if it was scrolled up less than 7 pixels up. The new implementation still has similar behavior as the new scroll anchor is not at the very bottom
of the scrollable area, so the chat will still show the new message as long as the bottom edge of the previous oldest message was still visible when the new one arrived.

### DIFF
--- a/scss/message/_message-list.scss
+++ b/scss/message/_message-list.scss
@@ -91,6 +91,60 @@
     width: 100%;
     padding: 0 0.5em;
 
+    // The following code block is to prevent
+    // https://github.com/deltachat/deltachat-desktop/issues/3753
+    // which is 'Adding a reaction makes the chat "scroll up"
+    // (then you have to scroll to see new messages)'.
+    // Signal messenger implements the same measure:
+    // https://github.com/signalapp/Signal-Desktop/blob/29c404266863491a3b26a73b24602d36d7a3ac46/stylesheets/_modules.scss#L5573-L5587
+    // which they have looked up here:
+    // https://css-tricks.com/books/greatest-css-tricks/pin-scrolling-to-bottom/
+    // Spec:
+    // https://drafts.csswg.org/css-scroll-anchoring
+    //
+    // TODO there is a concern: whether this makes the chat list jumpy
+    // when it is scrolled up, making the scroll anchor
+    // (which is at the bottom of chat list) go way off screen.
+    // This doesn't appear to be the case.
+    // Also whether we're gonna jump to bottom if the user is reading old
+    // messages as is waiting for the newer ones to load
+    // while scrolled to bottom. This would be unexpected.
+    // I suppose this is why Signal has `--have-newest`.
+    // But I also haven't managed to see this happen.
+    // But, for example, if you scroll up a bit, and then someone
+    // adds a reaction to a message that is rendered, but is above the messages
+    // that are currently in view, the view will shift down.
+    //
+    // Perhaps a more reasonable approch would be to set the last message
+    // _in view_ as the anchor, and only explicitly scroll to latest message
+    // when it is freshly received,
+    // but this will (probably?) require quite some JS.
+    // Or simply only alternate scroll anchoring behavior when
+    // we're scrolled to bottom.
+    // Ah, if only there was a way to "flip" overflow anchor selection
+    // algorithm to just select bottom-most visible element
+    // and not the top-most.
+    // Maybe we could though? E.g. with `order` CSS property?
+    //
+    // Anyways, it's not often that already sent messages change in Delta Chat,
+    // because there is no "delete for everyone" and no "Edit".
+    & > ul {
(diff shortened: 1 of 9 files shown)


WofWca:
It is worth examining the Signal code more closely

- https://github.com/signalapp/Signal-Desktop/commit/4b6b9cce011dfc7e4ac077a7d58c0d3b4745c01f
- https://github.com/signalapp/Signal-Desktop/commit/f115ba58739aae80b4a7b2b433fe129320beb9fa

WofWca:
Wait, actually the fact that the chat "scrolls" to the new message when the reaction bar is open, is an issue.

Maybe we should just disable this anchoring alternating behavior when the reaction bar is open.

WofWca:
The more I look at this the more it feels like it's just an ugly hack... In a perfect world, the scroll anchor _ought not_ change its position relative to actually visible elements, but here this is what's happening.
I noted this in the comments, about explicit `scrollTo` and anchor selection algorithm.

This fixes the issue, and does not appear to introduce consistent / obvious bugs. But, as I said, I think this is just a hack that we'll likely regret introducing. The `scrollToBottom` on `fetchedIncomingMessages` needs to be kept, for example.

Rather simpler workarounds I can suggest:
- Just keep message height the same even when a reaction is added
- When a reaction is added, just scroll to last message if we're scrolled close to the bottom of the message list

scss/message/_message-list.scss
```diff
@@ -91,6 +91,60 @@
...
+    // Perhaps a more reasonable approch would be to set the last message
+    // _in view_ as the anchor, and only explicitly scroll to latest message
+    // when it is freshly received,
+    // but this will (probably?) require quite some JS.
+    // Or simply only alternate scroll anchoring behavior when
+    // we're scrolled to bottom.
+    // Ah, if only there was a way to "flip" overflow anchor selection
+    // algorithm to just select bottom-most visible element
+    // and not the top-most.
+    // Maybe we could though? E.g. with `order` CSS property?
```
WofWca, line 127: Uhhhhhh guys I think this is working. I'll experiment...
Will need to take accessibility into account, and other potential things that rely on messages being in order.

src/renderer/components/message/MessageWrapper.tsx
```diff
@@ -64,8 +68,13 @@ export function MessageWrapper(props: RenderMessageProps) {
   ])
 
   return (
-    <li id={props.key2} className='message-wrapper'>
-      <Message {...props} />
+    <li
+      id={props.key2}
+      className={classNames('message-wrapper', {
+        'overflow-anchor': reactionBarShownForMessageId === props.message.id,
```
WofWca, line 74: This is made to not hide the reactions bar when a new message arrives.
But there is still an issue with this. If you open a reactions bar and someone leaves a reaction on a message that is above it, the reactions bar will hide.

WofWca:
Closing for now because, as I said, I think the approach is hacky.

Feel free to reopen though.

PR #4025, opened 2024-07-11
Preset lastChatId before switching accounts

### LINKED ISSUE #3983: Clicking notification for another account doesn't open the target chat
- Operating System (Linux/Mac/Windows/iOS/Android): Windows 10
- Delta Chat Version: 1.46.1
- Expected behavior: The chat for which the notification appeared opens
- Actual behavior: The account is switched, but the chat selection remains unchanged
    1. Select chat with A on your first account
    2. Switch to your second account
    3. Wait for someone except A send you a message to your first account
    4. Click the notification
- Screenshots:
- Logs:

```
Uncaught (in promise) Error: accountid of ChatProvider context is not equal to nextAccountId
    at selectChat (ChatContext.tsx:70:15)
    at useMessage.ts:61:15
```

Here

https://github.com/deltachat/deltachat-desktop/blob/88e65f4b3fa014460dfca84d8c9f8246836e4c06/src/renderer/components/RuntimeAdapter.tsx#L41-L56

when `jumpToMessage` is called, `accountId` remains the current account ID and not the new one

https://github.com/deltachat/deltachat-desktop/blob/54cd76c3628e9add851ef50cf9716728a50f1964/src/renderer/contexts/ChatContext.tsx#L69-L73

This was probably introduced in #3789

nicodh:
resolves #3983

The error came from the wrong accountId in selectChat which still had the state from the time it was set via runtime.setNotificationCallback

This change also avoids the loading of (the possibly wrong) lastChat, which is triggered by switching account, but might be superfluous if the chat from notification is not the lastChat

### DIFF
--- a/src/renderer/components/RuntimeAdapter.tsx
+++ b/src/renderer/components/RuntimeAdapter.tsx
@@ -9,6 +9,8 @@ import type { PropsWithChildren } from 'react'
 import { ActionEmitter, KeybindAction } from '../keybindings'
 import useDialog from '../hooks/dialog/useDialog'
 import WebxdcSaveToChatDialog from './dialogs/WebxdcSendToChat'
+import { saveLastChatId } from '../backend/chat'
+import useChat from '../hooks/chat/useChat'
 
 type Props = {
   accountId?: number
@@ -25,6 +27,7 @@ export default function RuntimeAdapter({
 }: PropsWithChildren<Props>) {
   const processQr = useProcessQr()
   const { jumpToMessage } = useMessage()
+  const { selectChat } = useChat()
 
   const { closeDialog, openDialog, closeAllDialogs } = useDialog()
   const openSendToDialogId = useRef<string | undefined>(undefined)
@@ -42,15 +45,17 @@ export default function RuntimeAdapter({
       async ({ accountId: notificationAccountId, msgId, chatId }) => {
         if (accountId !== notificationAccountId) {
           closeAllDialogs()
+          // selectAccount always loads the last used chat
+          // by setting saveLastChatId we force to load immediately
+          // the chat holding the notification message
+          await saveLastChatId(notificationAccountId, chatId)
           await window.__selectAccount(notificationAccountId)
-        }
-
-        if (chatId !== 0) {
+        } else if (chatId !== 0) {
+          await selectChat(accountId, chatId)
           clearNotificationsForChat(notificationAccountId, chatId)
-
-          if (msgId !== 0) {
-            jumpToMessage(notificationAccountId, msgId, true)
-          }
+        }
+        if (msgId) {
+          window.__internal_jump_to_message?.(msgId)
         }
       }
     )
@@ -64,7 +69,7 @@ export default function RuntimeAdapter({
         ActionEmitter.emitAction(KeybindAction.Settings_Open)
       }
     }
-  }, [accountId, jumpToMessage, processQr, closeAllDialogs])
+  }, [accountId, jumpToMessage, processQr, selectChat, closeAllDialogs])
 
   useEffect(() => {
     runtime.onWebxdcSendToChat = (file, text) => {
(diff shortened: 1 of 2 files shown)


WofWca (approved):
Makes sense.
Thoug I can't really test if this fixes it now.

nicodh:
> Makes sense. Thoug I can't really test if this fixes it now.

why not? just send a message from one account to another and click on the notification. It should open the other account and select the corect chat

WofWca:
Because I'm a Windows pleb and I don't have the setup to make a Windows build from WSL.

nicodh:
> Because I'm a Windows pleb and I don't have the setup to make a Windows build from WSL.

So you don't have a local installation of Desktop?

WofWca:
I do, but it's from https://download.delta.chat/
I don't have the setup to build from sources myself. I could try, though.

Simon-Laux:
> Because I'm a Windows pleb and I don't have the setup to make a Windows build from WSL.

you don't need that, you don't even need rust installed anymore, prebuilds work

src/renderer/components/RuntimeAdapter.tsx
```diff
@@ -42,15 +45,17 @@ export default function RuntimeAdapter({
       async ({ accountId: notificationAccountId, msgId, chatId }) => {
         if (accountId !== notificationAccountId) {
           closeAllDialogs()
+          // selectAccount always loads the last used chat
+          // by setting saveLastChatId we force to load immediately
+          // the chat holding the notification message
```
Simon-Laux, line 50: I don't understand this comment.

Simon-Laux: ah is this meant as an optimisation? save it there first so it's faster?

WofWca:
> you don't need that, you don't even need rust installed anymore, core prebuilds work out of the box. if you got a dev environment with nodejs installed you are ready to build desktop

Nice. Though I still get an error

```
⨯ ENOENT: no such file or directory, scandir '......../deltachat-desktop/dist/win-unpacked/resources/app.asar.unpacked/node_modules/@deltachat'  failedTask=build stackTrace=Error: ENOENT: no such file or directory, scandir '............../deltachat-desktop/dist/win-unpacked/resources/app.asar.unpacked/node_modules/@deltachat'
```

in the `afterPack` hook. The `app.asar.unpacked` directory is missing at that point

src/renderer/components/RuntimeAdapter.tsx
nicodh: When a new account is selected - the first thing in MainScreen is useSelectLastChat which is not needed since we load (possibly) another chat (of the notification message)
So if we set LastChat ID before, the correct chat will be loaded there

It is a hack, but the alternative would be to set some jumpToMessage in ScreenContext that would be accessible in MainScreen before loading lastChat

If you have ideas how to clarify that in the comment please add them

nicodh:
> > 
> 
> Nice. Though I still get an error

That error is not related to this PR

src/renderer/components/RuntimeAdapter.tsx
```diff
@@ -42,15 +45,17 @@ export default function RuntimeAdapter({
...
+          await saveLastChatId(notificationAccountId, chatId)
           await window.__selectAccount(notificationAccountId)
-        }
-
-        if (chatId !== 0) {
+        } else if (chatId !== 0) {
+          await selectChat(accountId, chatId)
```
Simon-Laux, line 54: jumpToMessage also selects chat, so not sure if this is needed here.

Simon-Laux (approved):

src/renderer/components/RuntimeAdapter.tsx
Simon-Laux: Ok thanks for paying attention. looks like I confused this with select chat, sorry for being wrong.

PR #4028, opened 2024-07-12
WIP: AddMemberChip needs no remove handler

### LINKED ISSUE #3935: group memberlist: just tapping pills should not clear members
when modifing a group's memberlist, added members are staged as "pills", which is nice.

however, when tapping these pills, they are just deleted.

this is unexpected.

instead, there should be a little "x" right of the pills - only if that is tapped, the pill is removed.

https://github.com/deltachat/deltachat-desktop/assets/9800740/91f72c68-a5bc-4347-8aa9-0b4b688b1d2b

nicodh:
fixes #3935

### DIFF
--- a/src/renderer/components/dialogs/AddMember/AddMemberDialog.tsx
+++ b/src/renderer/components/dialogs/AddMember/AddMemberDialog.tsx
@@ -56,17 +56,10 @@ export function AddMemberDialog({
   )
 }
 
-export const AddMemberChip = (props: {
-  contact: T.Contact
-  onRemoveClick: (contact: T.Contact) => void
-}) => {
-  const { contact, onRemoveClick } = props
+export const AddMemberChip = (props: { contact: T.Contact }) => {
+  const { contact } = props
   return (
-    <div
-      key={contact.id}
-      className='AddMemberChip'
-      onClick={() => onRemoveClick(contact)}
-    >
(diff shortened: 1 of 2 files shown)


nicodh:
@Simon-Laux I have no ideas why there was a onClick toggleMember in AddMemberChip since I found no other usage than in addMember Dialog and there is no UI element like a delete cross etc.

Simon-Laux:
> instead, there should be a little "x" right of the pills - only if that is tapped, the pill is removed.
> ~ r10s on original issue

that "X"/cross icon is missing and should be added, as far as I understood this.
Also I See no changes that could fix the visual misalignment

I don't know about toggle member inside of the chip, can be removed if unused IMO.

I imagine the cross button to remove to look like this:
Bildschirmfoto 2024-07-13 um 16 05 43
 this is the chip component from the ionic framework https://ionicframework.com/docs/api/chip#slotting-components-and-icons.

the avatar could still go to the border in our case, but it should be aligned, so it's neither smaller nor bigger than the height of the chip.

nicodh:
Ok. Then I will add a delete icon and restore the onClick remove. Might be useful in long lists...

nicodh:
Replaced by https://github.com/deltachat/deltachat-desktop/pull/4044

PR #4029, opened 2024-07-12
Select chat after forwarding

### LINKED ISSUE #3834: chat does not scroll down after forwarding
steps to reproduce:
- forward message to different chat (that is not saved messages, because saved messages is intentionally not selected)
- observer that you need to scroll down to see the message

nicodh:
fixes #3834

### DIFF
--- a/src/renderer/components/dialogs/ForwardMessage/index.tsx
+++ b/src/renderer/components/dialogs/ForwardMessage/index.tsx
@@ -45,15 +45,14 @@ export default function ForwardMessage(props: Props) {
     const chat = await BackendRemote.rpc.getFullChatById(accountId, chatId)
     onClose()
     if (!chat.isSelfTalk) {
-      selectChat(accountId, chat.id)
       const yes = await confirmForwardMessage(
         openDialog,
         accountId,
         message,
         chat
       )
-      if (!yes) {
-        selectChat(accountId, message.chatId)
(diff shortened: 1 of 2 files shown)


src/renderer/components/dialogs/ForwardMessage/index.tsx
```diff
@@ -45,15 +45,14 @@ export default function ForwardMessage(props: Props) {
     const chat = await BackendRemote.rpc.getFullChatById(accountId, chatId)
     onClose()
     if (!chat.isSelfTalk) {
-      selectChat(accountId, chat.id)
```
Simon-Laux, line 48: this is there because then you see the chat you are going to forward to, to prevent user mistakes.

nicodh: Ok - that should be worth a line of an explaining comment in the code :-)

So I have to find why the already opened chat is not scrolling down I guess

Simon-Laux: maybe calling select chat again could already work. or whatever happens when you click again on the chat in the chat-list if you have already selected the chat,

Simon-Laux (approved):

Simon-Laux (approved):
Code LGTM, but I didn't test it

PR #4030, opened 2024-07-12
WIP: Add new draft styles

### LINKED ISSUE #3968: vcard staging polishing needed
thanks for making vcard nicer at https://github.com/deltachat/deltachat-desktop/pull/3948

there are some minors that did not make it to the PR review:

- staged vcards show the "pointing finger cursor", which does not make sense as nothing happens when tapping the vcard

- there is no padding left of the vcard's avatar, which looks bad. i'd add same padding as above the vcard's avatar

Screenshot 2024-06-20 at 15 21 01

EDIT: same is true for staged files (the missing padding does not look so bad in case of files, the cursor is wrong as well), so, in case both share the same code, that can be fixed together

nicodh:
resolves #3968

### DIFF
--- a/scss/composer/_composer.scss
+++ b/scss/composer/_composer.scss
@@ -1,12 +1,10 @@
 .composer {
-  background-color: var(--composerBg);
   border-left: 1px solid rgba(16, 22, 26, 0.1);
 
   &.contact-request {
     display: flex;
     justify-content: space-around;
     padding: 15px 0px;
-
     & > .contact-request-button {
       background-color: var(--chatListItemSelectedBg);
       border-radius: 10px;
@@ -25,7 +23,11 @@
     & > .attachment-quote-section {
       display: flex;
       justify-content: space-between;
-      padding: 3px 0;
+      padding: 5px 0 0 5px;
+      background-color: #efefef;
+      border-radius: 8px;
+      margin: 0px 12px 10px;
+      opacity: 0.85;
 
       .quote {
         // this is just overwrites for the quote styles defined in _message.scss
@@ -59,6 +61,7 @@
   .lower-bar {
     display: flex;
     align-items: flex-end;
+    background-color: var(--composerBg);
 
     .attachment-button {
       width: 40px;
--- a/scss/composer/_composer.scss
+++ b/scss/composer/_composer.scss
@@ -46,7 +46,6 @@
       }
 
       .clear-quote-icon {
-        margin-left: auto;
         margin-right: 5px;
         min-width: 32px;
         &::before {
--- a/scss/composer/_composer.scss
+++ b/scss/composer/_composer.scss
@@ -31,6 +31,10 @@
         margin: 4px;
       }
 
+      button {
+        margin-left: auto;
+      }
+
       .message-attachment-media {
         flex-grow: 1;
         margin: 4px;
@@ -46,7 +50,6 @@
       }
     }
   }
-
   .lower-bar {
     display: flex;
     align-items: flex-end;
--- a/scss/composer/_composer.scss
+++ b/scss/composer/_composer.scss
@@ -44,14 +44,6 @@
       .message-attachment-generic {
         margin: 4px;
       }
-
-      .clear-quote-icon {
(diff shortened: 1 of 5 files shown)


nicodh:
TBD: should we scroll down when the attachement draft is added?

Styling is just a proposal - reduced opcity should show the "draft" state

Draft Image
image

VCard
image

Datei
image

r10s:
looks nice - at a first glance, however, i think the old "white outer box" makes things more clear, that the bubble is not sent yet, what the staging area is etc.  still, inside the white box, the nicely styled card/image can be shown.

but i am afk, only looking at the screenshots, i can have a closer look tomorrow.

Simon-Laux:
My ideas on this:
- keep in in the composer (white background), I agree with r10s on that point
- change to 100% opacity
- use some subtle color as background that you can still see the rounded corners, but the color is subtle that it still looks good

scss/composer/_composer.scss
```diff
@@ -1,12 +1,10 @@
 .composer {
-  background-color: var(--composerBg);
```
r10s, line 2: let's keep this line, whatever we do.

it is much clearer when "quote" + "image or other attachment" + "text and send button" is recognoised as one things.

this is also better for inner consistency (and for quotes also for external consistency)

scss/composer/_composer.scss
```diff
@@ -25,7 +23,11 @@
     & > .attachment-quote-section {
       display: flex;
       justify-content: space-between;
-      padding: 3px 0;
+      padding: 5px 0 0 5px;
+      background-color: #efefef;
+      border-radius: 8px;
+      margin: 0px 12px 10px;
+      opacity: 0.85;
```
r10s, line 30: as a minimal thing to solve the issues at hand, the following would do:

```suggestion
      padding: 9px;
      border-bottom: 1px solid #eee;      
```

for quotes, i would not do "more".

by more "room" around the quotes, things look nicer already.

for vcards, a rounded box and a slighly greyish background would be nice, however.

but i would leave the "X" button outside the box and also let the box appear not at 100% wide (this does not look like a "card" at the end)

but as said, also without, it fixes the issue at hand and is an improvement.

src/renderer/components/message/styles.module.scss
```diff
@@ -37,7 +37,8 @@ $reaction-transition-duration: 100ms;
 
 .vcard {
   display: flex;
-  cursor: pointer;
+  padding-left: 8px;
+  margin-bottom: 12px;
```
r10s, line 41: the padding/margin here could be removed again when going for the general "padding: 9px;" suggestion above.

having simple, horizontally symmetric spacings  is also better wrt to the rtl issue that needs to be targeted at some point, btw

r10s (commented):
i reviewed in a little more detail and played around with things on my desktop.

scss/composer/_composer.scss
nicodh: @r10s  Applied your suggestions. Feedback welcome

nicodh:
Image

image

VCard

image

Datei

image

Next proposal...

scss/composer/_composer.scss
r10s: thanks a lot! ❤️

r10s (approved):
tested - and it is a huge improvement that can be merged :)

also, the "close" button and its hover-effect is much nicer now (as no longer a long, vertical bar when hovered)

what i would streamline is the different position of the "close" button - for images/quotes is it on the right edge - for contacts it is directly after the contact card.

i think, the old approach, "right edge" is better as there are often combinations with quotes and we have two "X" - and it looks much nicer and is better usable if they're at the same position.

so, i would change that bit for the contact-cards/files as well (so moving the "x" to the edge).  but that is no blocker :)

nicodh:
That how I implemented it before, and Simon asked me to keep the icon nearer to the VCard & File since on bigger screen the X is pretty far from the left aligned VCard :-)

But I agree if there are multiple delete icons it looks strange if they are positioned different

PR #4033, opened 2024-07-17
fix: account settings not restoring on "Cancel"

WofWca:
...if the user attempted to set invalid
credentials by pressing "OK" prior to pressing "Cancel"

TODO:
- [x] This is based on https://github.com/deltachat/deltachat-desktop/pull/4032. Merge that first.
- [ ] This issue is also present on DC Android. Make an issue before we forget?
- [ ] This means that there are users for whom the "Password and Account" dialog shows credentials that are actually invalid. This commit does not address this, neither does https://github.com/deltachat/deltachat-desktop/pull/4032
    Make an issue?

But there are still cases where invalid credentials are still stored,
see comments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@
 - fix update unread badge on receiving device messages #4020
 - fix target chat was not opened on notification click #3983
 - fix "Password and Account" dialog not indicating invalid credentials, making it seem that you can change password like this #4032
+- fix "Password and Account" not restoring original credentials on "Cancel" sometimes #4033
 
 <a id="1_46_1"></a>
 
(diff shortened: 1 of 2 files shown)


Simon-Laux:
Isn't that sth that core should do? the variables it actually uses to connect are prefixed with `configured_`. I think we should rather restore from those.

WofWca:
To be honest, I don't know much about the core and its API.
But yeah, it sounds weird that you can commit invalid credentials and then it keeps working on the old credentials while still showing the invalid ones.

I made this merge request with an assumption that the core API doesn't need to be changed.

WofWca:
OK, since not so long ago we do actually make sure to perform re-login after the credentials or other configuration changes. If we fail to log in, then the credentials do get restored when you press "Cancel". So the issue that this MR tries to solve is already solved. Also on Android.

PR #4037, opened 2024-07-18
fix show new incoming messages after clearing chat

### LINKED ISSUE #3991: message view not always updated

Simon-Laux:
fixes #3991
closes #3994


src/renderer/stores/messagelist.ts
```diff
@@ -740,20 +740,23 @@ class MessageListStore extends Store<MessageListState> {
...
-        } else {
-          if (item.kind === 'message') {
-            return item.msg_id === (last_item as any).msg_id
-          } else {
-            return item.timestamp === (last_item as any).timestamp
-          }
-        }
-      })
+      let indexStart =
+        last_item === undefined
```
farooqkz, line 747: I am in favor of `lastItem` instead of `last_item`.

farooqkz (changes requested):

src/renderer/stores/chat/chat_scheduler.ts
```diff
@@ -119,7 +119,7 @@ export class ChatStoreScheduler {
       try {
         returnValue = await effect(...args)
       } catch (err) {
-        log.error(`Error in queuedEffect ${effectName}: ${err}`)
+        log.error(`Error in queuedEffect ${effectName}:`, err)
```
farooqkz, line 122: Who was blaming me for unrelated changes? :D

Simon-Laux: it is related, I needed this to get the real error. basically I reproduced it, got `Error in queuedEffect ...` but only he text of the error without stack trace, this change pointed me to the location in the message list store file.

src/renderer/stores/messagelist.ts
Simon-Laux: not really topic of this pr, this pr should just fix the bug, the name is old.

Simon-Laux: we should do the rename in a dedicated pr, otherwise it's hard to see the relevant changes,

WofWca (approved):
So it's just about `last_item` being possibly `undefined`.

This fixes the issue I described in https://github.com/deltachat/deltachat-desktop/issues/3991 I don't know about the original one.

Simon-Laux:
> This fixes the issue I described in https://github.com/deltachat/deltachat-desktop/issues/3991 I don't know about the original one.

the initial issue by @r10s is: other account sends you messages. his description of "other device" is a bit misleading because you can misunderstand it as multi device (self sent message from same account on other device).

WofWca:
You're right.

PR #4039, opened 2024-07-19
Test: Use bottom div ref

nicodh:
This could be a much simpler implementation of scrolling down. Maybe we could expose the scrollToBottom function to Context to allow usage from "outside"...


Simon-Laux:
might miss messages that are not loaded

PR #4040, opened 2024-07-19
fix: QR scanning not showing errors on failure

WofWca:
`useTranslationFunction` hook cannot be used when we create the component instance like this.

This, however, removes an extra new line, but it's not bad.

To reproduce, send this (invalid) link in a chat and click it: https://i.delta.chat/#dummy&a=example%40testrun.org&n=test&i=test&s=test

image


Simon-Laux (approved):
Code looks good, didn't test

CHANGELOG.md
```diff
@@ -40,6 +40,7 @@
 - fix update unread badge on receiving device messages #4020
 - fix target chat was not opened on notification click #3983
 - fix CSP bypass in webxdc (not a vulnerability) #4011
+- fix: QR scanning not showing errors on failure #
```
WofWca, line 43:
```suggestion
- fix: QR scanning not showing errors on failure #4040
```

PR #4041, opened 2024-07-19
feat: paste invite link to search to add contact

WofWca:
Adds a pseudo item similar to "add contact" when you type an email

The "standard" way of pasting the invite link under the gear button
on the second tab of the "scan QR" dialog is not very straigtforward.

The first reflex is to paste it in the search field,
so let's handle this.

image

### DIFF
--- a/src/renderer/components/helpers/PseudoListItem.tsx
+++ b/src/renderer/components/helpers/PseudoListItem.tsx
@@ -4,6 +4,7 @@ import { PseudoContact } from '../contact/Contact'
 import { QRAvatar } from '../Avatar'
 import useTranslationFunction from '../../hooks/useTranslationFunction'
 import { useSettingsStore } from '../../stores/settings'
+import useProcessQR from '../../hooks/useProcessQr'
 
 export function PseudoListItem(
   props: PropsWithChildren<{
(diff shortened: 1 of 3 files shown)


src/renderer/components/helpers/PseudoListItem.tsx
```diff
@@ -109,3 +110,25 @@ export const PseudoListItemAddContact = ({
...
+
+  return (
+    <PseudoListItem
+      id='newcontactfrominvitelink'
+      cutoff='+'
+      text={tx('menu_new_contact')}
+      onClick={() => processQr(accountId, inviteLink.trim())}
+    />
+    // Perhaps we could also parse username from the link
+    // and put it into `subText`.
```
Simon-Laux, line 132: great idea

Simon-Laux: I think rpc.checkQr should work for that, like it gets the name for the confirmation dialog when it asks you wether you want to chat with `<Name>`

Simon-Laux (approved):
code LGTM, didn't test

src/renderer/components/helpers/PseudoListItem.tsx
nicodh: This could be improved in another PR. It would be nice to show both, the displayname & the mail address since the random testrun.org addresses are usually not known, only the displayname

PR #4048, opened 2024-07-24
Show displayname in ask start chat with

nicodh:
image

Only showing the mail address can be confusing, since often people don't know the Maildress only the displayname


farooqkz (commented):
LGTM

Issue #3959, opened 2024-06-19
Crash when trying to go back from "Add Account" or "Delete Account" on Microsoft Store version


WofWca:
This works fine on a manually downloaded 1.46.1, which is also built from 899f196d. Note that the manually installed one is 32 bit, and the Mircosoft Store is 64. Idk if this matters.

- Operating System (Linux/Mac/Windows/iOS/Android): Windows 10
- Delta Chat Version: 1.46.1 (git: v1.33.0-881-g899f196d)
- Expected behavior: Can go back from "Add account", and can delete account.
- Actual behavior: Nothing happens when I click the "back" arrow, and nothing happens when I click "delete". After I press another button, the app crashes.
- Steps to reproduce the problem:
    1. Install the Microsoft Store version of Delta Chat
    2. Create an account, send a "note to self" idk
    3. Click "Add Account" in the accounts bar.
    4. Click "back" arrow.
- Screenshots:
    ![image](https://github.com/deltachat/deltachat-desktop/assets/39462442/585d8b73-8a86-4808-9f2e-71ce21a329c2)
- Logs:

Here I clicked "Add Account" at 08:43:00, then clicked "back" at 08:44:09, then "back" again at 08:46:00, then "back" again at 08:49:00, then "Let's get started" at 08:49:19, after which the crash occurs:

```log
2024-06-19T08:42:36.599Z	core/event            	INFO	""	1	"src/scheduler.rs:674: IMAP session supports IDLE, using it."
2024-06-19T08:42:36.662Z	core/event            	INFO	""	1	"src/imap/idle.rs:64: INBOX: Idle entering wait-on-remote state"
2024-06-19T08:43:00.709Z	core/event            	INFO	""	64	"src/sql/migrations.rs:950: Created new database; [migration] v68-v114"
2024-06-19T08:43:00.710Z	core/event            	INFO	""	64	"src/sql.rs:294: Opened database \"C:\\\\Users\\\\wofwc\\\\AppData\\\\Local\\\\DeltaChat\\\\accounts\\\\redacted-aaaaaaaa-aaaaaaa-aaaaaaa\\\\dc.db\"."
2024-06-19T08:43:00.724Z	renderer/stores/SettingsStore	INFO	""	"cleared settings store"
2024-06-19T08:43:00.731Z	core/event            	INFO	""	1	"src/scheduler.rs:108: stopping IO"
[...]

PR #4054, opened 2024-07-30
Update `@deltachat/stdio-rpc-server` and `deltachat/jsonrpc-client` to 1.42.1 and minor npm dependency updates

nicodh:
* fix crash on windows #3959

for core updates see [Changelog](https://github.com/deltachat/deltachat-core-rust/blob/main/CHANGELOG.md)


nicodh:
@Simon-Laux not sure what we should add here into CHANGELOG

nicodh:
TODO: check wether we should handle the new property "quotedText" see [Just text quote](https://github.com/deltachat/deltachat-core-rust/issues/5695)

Simon-Laux (changes requested):
package.json update and changelog updates are missing. for change log just look to previous versions and copy the line about the core update, if it could be interesting for  users also mention some of the changes of core below (in a nested list item.

Simon-Laux:
> TODO: check wether we should handle the new property "quotedText" see https://github.com/deltachat/deltachat-core-rust/issues/5695

I think we just need to explicitly set it to `null` because of https://github.com/deltachat/yerpc/issues/10

PR #4055, opened 2024-07-30
Fix order of css rules

nicodh:
since v1.77.7 the behaviour of compiling css from nested rules changed: https://sass-lang.com/documentation/breaking-changes/mixed-decls/

To avoid warnings and keep the old structure of the compiled css all properties should occur before nested rules

Follow up of #4054

### DIFF
--- a/src/renderer/components/screens/WelcomeScreen/OnboardingScreen.tsx
+++ b/src/renderer/components/screens/WelcomeScreen/OnboardingScreen.tsx
@@ -38,12 +38,8 @@ export default function OnboardingScreen(props: Props) {
         props.selectedAccountId
       )
       if (acInfo.kind === 'Unconfigured') {
-        try {
-          await props.onUnSelectAccount()
-          EffectfulBackendActions.removeAccount(props.selectedAccountId)
-        } catch (error) {
-          log.error('error on removeAccount', error)
-        }
+        await props.onUnSelectAccount()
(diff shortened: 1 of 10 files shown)


WofWca:
So, is this just a refactor? Have you seen if something is bugged without this, but with the upgrade of SASS?

src/renderer/components/screens/WelcomeScreen/OnboardingScreen.tsx
```diff
@@ -38,8 +38,12 @@ export default function OnboardingScreen(props: Props) {
...
       )
       if (acInfo.kind === 'Unconfigured') {
-        await props.onUnSelectAccount()
-        await EffectfulBackendActions.removeAccount(props.selectedAccountId)
+        try {
+          await props.onUnSelectAccount()
+          EffectfulBackendActions.removeAccount(props.selectedAccountId)
+        } catch (error) {
+          log.error('error on removeAccount', error)
+        }
```
WofWca, line 46: Unrelated?

nicodh: Oops - yes. Will remove it

nicodh:
There were warnings when building. In future the generated css will look different then before. With these changes the generated css should stay the same, at least according to the documentation

WofWca (approved):
Except for the unrelated change, looks good.

WofWca (approved):

PR #4056, opened 2024-07-30
Add Clone Group functionality

### LINKED ISSUE #3933: add "clone group"
"Clone Group" was added to [android](https://github.com/deltachat/deltachat-android/pull/2797)/[ios](https://github.com/deltachat/deltachat-ios/pull/1967) for 1.41 on the side without clear requirements for some version.

meanwhile, it turned out several times, it is a very handy feature - and that it is really missing on desktop.

from the original issue:

> cloning group offers a pragmatic way to add unverified ppl to verified groups or upgrade to verified groups.
> 
> also the "birthday issue", where a 30 ppl group wants to discuss a present for one member is fixed by that.
> no need to put the option too much in the way, however [...] when tapped [...] "new group" is opened with a prefilled member list and prefilled name

as it is probably also on desktop low effort, it makes sense to add that even when going for maintenance mode.

ui-wise, right-tap in the chatlist or the three-dot-menu of a chat come to mind; if one is easier than the other, i'd go for that on the first iteration.

see [android](https://github.com/deltachat/deltachat-android/pull/2797) and [ios](https://github.com/deltachat/deltachat-ios/pull/1967) for some reference (only few lines of changed code)

maxphilippov:
- useCreateGroup returned function (CreateGroup) now accepts optional groupTemplateId same way Android/iOS version do, this is a group chat id to base new chat on

- resolves #3933

### DIFF
--- a/src/renderer/components/dialogs/CreateChat/index.tsx
+++ b/src/renderer/components/dialogs/CreateChat/index.tsx
@@ -60,16 +60,53 @@ type ViewMode = 'main' | 'createGroup' | 'createBroadcastList'
 
 export default function CreateChat(props: DialogProps) {
   const { onClose } = props
+  const tx = useTranslationFunction()
+
   const [viewMode, setViewMode] = useState<ViewMode>('main')
 
   return (
     <Dialog width={400} onClose={onClose} fixed>
       {viewMode == 'main' && <CreateChatMain {...{ setViewMode, onClose }} />}
       {viewMode == 'createGroup' && (
-        <CreateGroup {...{ setViewMode, onClose }} />
+        <>
+          <DialogHeader title={tx('clone_chat')} />
+          <CreateGroup {...{ setViewMode, onClose }} />
+        </>
       )}
       {viewMode == 'createBroadcastList' && (
         <CreateBroadcastList {...{ setViewMode, onClose }} />
+      )}
+    </Dialog>
+  )
+}
+
+export function CloneChat(props: { chatTemplateId: number } & DialogProps) {
+  const { chatTemplateId, onClose } = props
+  const accountId = selectedAccountId()
+  const tx = useTranslationFunction()
+
+  const [chat, setChat] = useState<T.FullChat | null>(null)
+
+  useMemo(() => {
+    BackendRemote.rpc
+      .getFullChatById(accountId, chatTemplateId)
+      .then(c => setChat(c))
+  }, [accountId, chatTemplateId])
+
+  return (
+    <Dialog width={400} onClose={onClose} fixed>
+      {chat && (
+        <>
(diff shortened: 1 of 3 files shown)


src/renderer/components/dialogs/CreateChat/index.tsx
```diff
@@ -647,37 +647,61 @@ const useCreateGroup = (
...
+      if (groupTemplateId) {
+        const chat = await BackendRemote.rpc.getFullChatById(
+          accountId,
+          groupTemplateId
+        )
 
-    if (groupImage && groupImage !== '') {
-      await BackendRemote.rpc.setChatProfileImage(accountId, chatId, groupImage)
-    }
+        groupMembers.push(...chat.contactIds)
```
WofWca, line 660: Is it OK to mutate the parameter? Would it not add the same members twice to the same array if you cloned the group twice?
I'd just clone the array.

src/renderer/components/dialogs/CreateChat/index.tsx
```diff
@@ -647,37 +647,61 @@ const useCreateGroup = (
...
 
-    await Promise.all(
-      groupMembers.map(contactId => {
-        if (contactId === C.DC_CONTACT_ID_SELF) {
-          return
-        }
-        return BackendRemote.rpc.addContactToChat(accountId, chatId, contactId)
-      })
-    )
+        finalGroupName = `Copy: ${chat.name}`
```
WofWca, line 662: Not internationalized, but perhaps it's fine.

src/renderer/components/dialogs/CreateChat/index.tsx
```diff
@@ -647,37 +647,61 @@ const useCreateGroup = (
   const accountId = selectedAccountId()
   const { selectChat } = useChat()
 
-  const createGroup = useCallback(async () => {
-    const isVerified = await areAllContactsVerified(accountId, groupMembers)
+  const createGroup = useCallback(
+    async (groupTemplateId?: number | null) => {
```
WofWca, line 651:
```suggestion
    async (groupTemplateId?: number) => {
```

src/renderer/components/dialogs/CreateChat/index.tsx
```diff
@@ -647,37 +647,61 @@ const useCreateGroup = (
...
+          )
+        })
+      )
+
+      return chatId
+    },
+    [accountId, groupImage, groupMembers, groupName]
+  )
+
+  return async (groupTemplateId?: number | null) => {
```
WofWca, line 699:
```suggestion
  return async (groupTemplateId?: number) => {
```

WofWca (commented):
LGTM apart from parameter mutation, did not test.

nicodh (approved):

PR #4057, opened 2024-07-31
show device message only once as unread

### LINKED ISSUE #3934: show device message only once as unread
on switching or adding profiles,
the device messages (welcome and update messages) are added - they are maked as unread even though maybe read already several times, being unnecessary noise, not only for us devs.

we should add them as "read" instead, if they're read in other profiles already.

(this started on android with the idea to not add the messages, however, it turned out that it is easier and less confusing to mark the messages as read)

implementation-wise, both, android and ios:
-  check if a global user-setting is equal to the deviceMessageLabel just added 
- if so, the chat is marked as read completely
-  after that, unconditionally, deviceMessageLabel  is set

reference implementations: https://github.com/deltachat/deltachat-ios/pull/2204/files, https://github.com/deltachat/deltachat-android/pull/3069/files ("deviecMsgId" should be named "deviceMsgLabel" there [cmp](https://github.com/deltachat/deltachat-android/pull/3121))

this issue has low priority

meganoahj:
fixes #3934

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@
   - Case-insensitive search for non-ASCII messages
 - make ImageCropper use CSS-transforms for UI and canvas API to cut the result #3893
 - update stock translations #4051
-- show device message only once as unread
+- show device message only once as unread #4057
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


nicodh (approved):
LGTM

Reviewed by reading only

PR #4060, opened 2024-08-02
Fix empty context menu in HTML viewer

### LINKED ISSUE #4053: HTML email viewer: content context menu has no items without selection
Observed on linux flatpak gnome with dc version 1.46.1:
image
On macOS with dc version 1.46.1 it looks like this:
Bildschirmfoto 2024-07-28 um 21 22 49
But also selects the word under the mouse.
With a selection it also works under linux, but it still shows the extra separator item, which shouldn't be there if nothing comes after it.

I think the problem is the separator item(s) which make the context menu appear even then there is no selection.

maxphilippov:
- don't disply HTML mail viewer context menu if it's empty
- hide unnecessary separators
- resolves #4053

### DIFF
--- a/src/main/windows/html_email.ts
+++ b/src/main/windows/html_email.ts
@@ -3,6 +3,7 @@ import electron, {
   BrowserWindow,
   dialog,
   Menu,
+  MenuItem,
   MenuItemConstructorOptions,
   nativeTheme,
   session,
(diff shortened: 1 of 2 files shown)


maxphilippov:
I was wandering, do we really need those separators in-between?

Here's how it currently looks (with separators):
image

Here's without
image

src/main/windows/html_email.ts
```diff
@@ -518,61 +519,65 @@ const createContextMenu = (win: BrowserWindow, webContents: WebContents) => {
...
-        id: 'copyImage',
-        label: tx('menu_copy_image_to_clipboard'),
-        visible: props.mediaType === 'image',
-        click() {
-          webContents.copyImageAt(props.x, props.y)
-        },
-      }),
+    const menuItems = []
+
+    if (props.isEditable || hasText) {
```
Simon-Laux, line 524: what does `isEditable` mean here?

Simon-Laux (approved):
code looks good, didn't test yet

Simon-Laux:
> I was wondering, do we really need those separators in-between?

I don't know, I don't really care in this case, we can remove them.

meganoahj (approved):
Nice, it works

src/main/windows/html_email.ts
maxphilippov: I honestly didn't dig that deep and kinda missed this one. There was also a Paste command I removed, maybe leftovers from Signal?

Simon-Laux: more like leftovers from the context menu npm package I used to add context menus to deltachat. anyway paste makes no sense, because emails are read only.
thats also why I wonder what isEditable means, because in my mind editing means writing which goes against readonly.
But @meganoahj tested it and it worked for him, so might not matter besides being mysterious/confusing.

PR #4061, opened 2024-08-02
Fix empty space before HTML view content

### LINKED ISSUE #4052: HTML email view: strange blank space/offset until email content begins.
sandboxed view doesn't begin where header ends, instead there is some strange space.

Observed both on Linux (gnome flatpak) and MacOS with long deltachat text message.

The email content is displayed in a sandboxed browser view that is overlayed with an offset on the ui window.

I suspect the current account either has a pixel bug or does not account for display scaling.

maxphilippov:
- resolves #4052

### DIFF
--- a/src/main/windows/html_email.ts
+++ b/src/main/windows/html_email.ts
@@ -132,9 +132,6 @@ export function openHtmlEmailWindow(
     window.show()
   })
 
-  window.once('focus', () => {
-    window.webContents.setZoomFactor(DesktopSettings.state.zoomFactor)
-  })
   window.on('close', () => {
     context_menu_handle?.()
     delete open_windows[window_id]
@@ -218,6 +215,10 @@ export function openHtmlEmailWindow(
     window
   )
   window.setBrowserView(sandboxedView)
+  sandboxedView.webContents.setZoomFactor(
+    DesktopSettings.state.zoomFactor *
+      Math.pow(1.2, window.webContents.getZoomLevel())
+  )
   let context_menu_handle = createContextMenu(window, sandboxedView.webContents)
 
   window.webContents.ipc.handle('html-view:more-menu', (_ev, { x, y }) => {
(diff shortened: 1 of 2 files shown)


maxphilippov:
This is a bit of a guess from me, cause it looks like there's no need to recompute bounds.y of the html view since it's computed correctly on first launch. Was there any special case why this was needed?

Simon-Laux:
> Was there any special case why this was needed?

it changes when you zoom in or out. (in the view menu of the window title-/toolbar.

meganoahj:
I tested it and normal size works but zooming in HTML Viewer doesn't.

WofWca:
Maybe everything I'm gonna describe belongs to a different issue, but here it is:

This MR makes things much better than what I described in https://github.com/deltachat/deltachat-desktop/issues/4052#issuecomment-2266660940 , but apparently still has issues.
If on the main window you set zoom to 0.6, then open an HTML email, there is still a bit of blank space between the header bar and the contents of the email:

image

And if you set main window zoom to 1.4, then the contents of the email start to overlap the header.

In addition, "window" of the email contents seems to be smaller - you can see the scrollbar appear

image

Also I discovered (probably a separate) issue: "Zoom out" and "Zoom in" behaves weird. When you just opened the email, it will zoom only the header bar. Once you click on the contents of the email, it starts zooming the contents of the email:


https://github.com/user-attachments/assets/0b49d0c8-099c-4da7-84b6-db22f15f015f

maxphilippov:
> Maybe everything I'm gonna describe belongs to a different issue, but here it is:

Sorry for troubling you, I meant to turn it into draft after I realized it has to zoom, I pushed a fixed version now, can you look at it again?

WofWca:
IDK, it still acts funky on 7425a625

https://github.com/user-attachments/assets/4340888b-cfc7-4e29-a850-3437fc688a43

https://github.com/user-attachments/assets/06b0d762-5e1f-4bcd-8dc8-37cbb70166b1

maxphilippov:
Okay, looks like everything is fixed for me. The only thing left is a weird bug where zoom changed by the menu (Zoom In/Zoom Out) stops scaling text in header above content (while still scaling content text size).

I didn't have time yet to find a 100% reproducable scenario, but it goes something like that:
- change zoom via menu (Zoom In or Zoom Out)
- reset zoom from the same menu
- resize the html view window
If you do it couple of times Zoom In/Zoom Out will stop scaling header. After messing with zoom for some time you can make it work again.

src/main/windows/html_email.ts
```diff
@@ -218,6 +215,9 @@ export function openHtmlEmailWindow(
     window
   )
   window.setBrowserView(sandboxedView)
+  sandboxedView.webContents.setZoomFactor(
+    DesktopSettings.state.zoomFactor * (1.2 ^ window.webContents.getZoomLevel())
```
WofWca, line 219:
```suggestion
    DesktopSettings.state.zoomFactor *
      Math.pow(1.2, window.webContents.getZoomLevel())
```

Though this does not appear to affect behavior as `window.webContents.getZoomLevel()` seems to be always 1, so the value of the expression is always just `DesktopSettings.state.zoomFactor` and then the resize handler runs and overrides it.

WofWca (approved):
Apart from the mentioned issue, there are two more:

1. If you click on the contents of the email and then zoom starts only affecting the contents of the email. I think this is not too important.
2. Sometimes if you resize the window vertically rapidly, the header bar hides. This can be fixed by resizing the window again. This appears to happen on all zoom levels of the main DC window. It looks similar to https://github.com/deltachat/deltachat-desktop/issues/4052#issuecomment-2266660940

https://github.com/user-attachments/assets/cc94e873-bfdc-4e3a-916d-081a3d79551b

I think those can be tackled in a separate issue or later. This is already much better than it used to be.
Code-wise this looks simple enough, but I do not understand what is going on in detail.
I think it's time to merge, when the my suggestion is accepted.

PR #4067, opened 2024-08-05
Ndh/fix file paths

### LINKED ISSUE #4058: APPX / ms store: open sticker folder does not work
Original reported on the fediverse (includes video): https://lgbtqia.space/@m/112869304274378557

solution:
apps apps are storing their deltachat data folder inside a sandbox / different directory.
We need to tell windows to open this special directory, we already had that problem with log files in the past: #2209 

and is solved here, so just transfer/copy this solution for the sticker button:

https://github.com/deltachat/deltachat-desktop/blob/567964706534263bdbb73832248c17bc3f811444/src/main/menu.ts#L365-L374

Further thoughts related to this:
- I wonder if that path modification for appx still missing in other places..
- we could probably remove our workaround to find out if it is an appx (currently it modifies a file during the build/packaging process) and instead ask electron at runtime: `process.windowsStore` (https://www.electronjs.org/de/docs/latest/api/process#processwindowsstore-readonly)

nicodh:
resolves #4058

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,7 +61,7 @@
 - fix HTML email content not being zoomed #4052
 - fix Icon preview of latest WebXDC displayed when summary is reaction event #4062
 - fix stretched summaryPreviewIcon #4064
-
+- fix wrong paths in some Window installations #4058
 
 <a id="1_46_1"></a>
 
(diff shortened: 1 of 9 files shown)


nicodh:
Simon approved in real world...

PR #4071, opened 2024-08-08
ReactionsDialog: drop email, make rows clickable

### LINKED ISSUE #4066: remove email address from reaction details & make details tappable
as at other places,
we should tune down showning email addresses
also in the primary reaction UI.

ppl will recognize name and avatar,
and if in doubt, you should be able to tap the row to go to the full profile.

this also makes the dialog easier to read, brings things to the point
and adds consistency to iOS (and now [Android](https://github.com/deltachat/deltachat-android/pull/3207))

so,  concretely:

- remove the address in the reaction details dialog (as stroked in the following image):<br>Screenshot 2024-08-05 at 15 02 57<br>the remaining name should be centered vertically and use getDisplayName() (which will use the address if there is no name set)

- when tapping the row, open the corresponding contact's profile

maxphilippov:
- don't show email in reactions dialog unless user has no display name
- clicking on a row in reactions dialog opens a user profile
- resolved #4066

### DIFF
--- a/src/renderer/components/dialogs/ReactionsDialog/index.tsx
+++ b/src/renderer/components/dialogs/ReactionsDialog/index.tsx
@@ -1,4 +1,5 @@
 import React, { useEffect, useState } from 'react'
+import classNames from 'classnames'
 
 import Dialog, {
   DialogBody,
@@ -19,8 +20,11 @@ import styles from './styles.module.scss'
 import type { DialogProps } from '../../../contexts/DialogContext'
 import type { T } from '@deltachat/jsonrpc-client'
 
+import useOpenViewProfileDialog from '../../../hooks/dialog/useOpenViewProfileDialog'
+
 export type Props = {
   reactionsByContact: T.Reactions['reactionsByContact']
+  onClose?: () => void
 }
 
 type ContactWithReaction = T.Contact & {
@@ -38,7 +42,10 @@ export default function ReactionsDialog({
       <DialogHeader title={tx('reactions')} />
       <DialogBody>
         <DialogContent>
-          <ReactionsDialogList reactionsByContact={reactionsByContact} />
+          <ReactionsDialogList
+            reactionsByContact={reactionsByContact}
+            onClose={onClose}
+          />
         </DialogContent>
       </DialogBody>
       <DialogFooter>
@@ -50,9 +57,10 @@ export default function ReactionsDialog({
   )
 }
 
-function ReactionsDialogList({ reactionsByContact }: Props) {
+function ReactionsDialogList({ reactionsByContact, onClose }: Props) {
   const accountId = selectedAccountId()
   const [contacts, setContacts] = useState<ContactWithReaction[]>([])
+  const openViewProfileDialog = useOpenViewProfileDialog({ onAction: onClose })
 
   useEffect(() => {
     const resolveContacts = async () => {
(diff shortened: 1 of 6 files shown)


src/renderer/components/dialogs/ReactionsDialog/index.tsx
```diff
@@ -81,19 +84,26 @@ function ReactionsDialogList({ reactionsByContact }: Props) {
...
+          >
             <div className={styles.reactionsDialogAvatar}>
               <AvatarFromContact contact={contact} />
             </div>
             <div className={styles.reactionsDialogContactName}>
-              <ContactName
-                displayName={contact.displayName}
-                address={contact.address}
-              />
+              <ContactName displayName={contact.displayName} />
```
WofWca, line 106: What's the point of removing the address? I don't think it's better.

src/renderer/components/dialogs/ReactionsDialog/index.tsx
```diff
@@ -81,19 +84,26 @@ function ReactionsDialogList({ reactionsByContact }: Props) {
...
   return (
     <ul className={styles.reactionsDialogList}>
       {contacts.map(contact => {
         return (
-          <li key={contact.id} className={styles.reactionsDialogListItem}>
+          <li
+            key={contact.id}
+            className={styles.reactionsDialogListItem}
+            onClick={contactClickGen(contact.id)}
+          >
```
WofWca, line 101: We should use `<button>` so it's accessible.
But I guess this can be done later, as part of Tauri.

src/renderer/components/dialogs/ReactionsDialog/index.tsx
```diff
@@ -81,19 +84,26 @@ function ReactionsDialogList({ reactionsByContact }: Props) {
     resolveContacts()
   }, [accountId, reactionsByContact])
 
+  const contactClickGen = (contactId: number) => {
+    return () => {
+      openViewProfileDialog(accountId, contactId)
+    }
+  }
```
WofWca, line 91: Why not just

```typescript
const onContactClick = (contactId: number) => {
  openViewProfileDialog(accountId, contactId)
}
```

WofWca (commented):
The "click to open contact" is nice, but I'm not sure if removing the address is good.

src/renderer/components/dialogs/ReactionsDialog/index.tsx
maxphilippov: > We should use `<button>` so it's accessible.

I'm a bit rusty on that, will 'role="button"' be enough? I'd love to not increase hierarchy depth

src/renderer/components/dialogs/ReactionsDialog/index.tsx
maxphilippov: That's the original issue, I guess we need to discuss that with @r10s. While on this, when I look at the new version (if we remove email), maybe make the name bigger?

r10s:
> but I'm not sure if removing the address is good

tuning down email adresses is an ongoing process and needed for chatmail, this PR and reactions is only one small step in that effort.

for reasoning, see eg. [https://delta.chat/en/2024-05-31-instant-onboarding#are-we-de-emphasizing-e-mail-addresses-in-the-user-interface ](https://delta.chat/en/2024-05-31-instant-onboarding#are-we-de-emphasizing-e-mail-addresses-in-the-user-interface)- the approach was recently confirmed several times by user get stuck using email adresses. it is better to push to other ways to get in contact, as using invite codes or qr codes. but if user see an email address before, they will try with that and fail :)

src/renderer/components/dialogs/ReactionsDialog/index.tsx
WofWca: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role

> If using role="button" instead of the semantic <button> or <input type="button"> elements, you will need to make the element focusable and define event handlers for [click](https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event) and [keydown](https://developer.mozilla.org/en-US/docs/Web/API/Element/keydown_event) events. This includes handling the Enter and Space keypresses in order to process all forms of user input.

r10s (commented):
- if is better when tapping the "Me" self-reaction just does nothing; opening "Saved messages" is not helpful and is unexpected. on Android, we also just do nothing

- when doing an action from the contact, eg. tap a shared chat or tap "send-message", it is weird that the "reaction" dialog stays opened (cmp corresponding discussion about that in the linked issue)

otherwise, functionalty-wise, this lgtm. thanks a lot for pushing that foward

src/renderer/components/dialogs/ReactionsDialog/index.tsx
r10s: i commented below

r10s: for size of the name: it is fine. and re-using the same style and size as in other contact list makes sense as well.

src/renderer/components/dialogs/ReactionsDialog/index.tsx
r10s: i would keep such semantic changes out of this PR as well

nicodh (approved):

nicodh (approved):

PR #4072, opened 2024-08-08
Rephrase english 'delete chat' translation

### LINKED ISSUE #4050: Improve confirmation message when trying to delete a 1:1 chat
Now it says `Are you sure you want to delete "email@address"?` One could think that the contact will be deleted, not the chat. Secondly, chatmail addresses don't tell much. Btw, for group chats the message looks good, it uses the chat name. As for DC Android, it asks `Delete 1 chat? It will no longer be shown in the chat list; its messages will remain on the server.` I.e. no contact name as well, but it's clear what will happen at least.

maxphilippov:
- resolves #4050

### DIFF
--- a/_locales/en.xml
+++ b/_locales/en.xml
@@ -224,7 +224,7 @@
     <string name="menu_delete_chat">Delete Chat</string>
     <!-- Command to delete all messages in a chat. The chat itself will not be deleted but will be empty afterwards, so make sure to be different from "Delete Chat" here. "Clear" is a verb here, "Empty Chat" would also be fine (eg. in German "Chat leeren") -->
     <string name="clear_chat">Clear Chat</string>
-    <string name="ask_delete_named_chat">Are you sure you want to delete \"%1$s\"?</string>
+    <string name="ask_delete_named_chat">Delete chat with \"%1$s\"? It will no longer be shown in the chat list; its messages will remain on the server.</string>
     <string name="menu_delete_messages">Delete Messages</string>
     <string name="delete_contact">Delete Contact</string>
     <string name="menu_delete_location">Delete this Location?</string>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 ### Changed
+- rephrase english translation for 'Delete Chat' #4050
 
 ### Fixed
 


maxphilippov:
I'm not sure about the translation process. Should make sense to update that for other translations.

_locales/en.xml
```diff
@@ -224,7 +224,7 @@
     <string name="menu_delete_chat">Delete Chat</string>
     <!-- Command to delete all messages in a chat. The chat itself will not be deleted but will be empty afterwards, so make sure to be different from "Delete Chat" here. "Clear" is a verb here, "Empty Chat" would also be fine (eg. in German "Chat leeren") -->
     <string name="clear_chat">Clear Chat</string>
-    <string name="ask_delete_named_chat">Are you sure you want to delete \"%1$s\"?</string>
+    <string name="ask_delete_named_chat">Delete chat with \"%1$s\"? It will no longer be shown in the chat list; its messages will remain on the server.</string>
```
iequidoo, line 227: Maybe remove email address at all? DC Android doesn't show it f.e. And the address doesn't always tell much, particularly if it's a chatmail one, more useful would be to have the contact name here

maxphilippov: > Maybe remove email address at all? DC Android doesn't show it f.e. And the address doesn't always tell much, particularly if it's a chatmail one, more useful would be to have the contact name here

It's weird, I haven't checked that yet in the core, but for me it displayed what's actually a `contact.displayName`. It prints `chat.name`, maybe something changed in core?

r10s: - for 1:1 chats without known names, the email addresses is displayed.
- for 1:1 chats with names, for groups and broadcasts, the name is displayed.

i just checked that on master, this is the current state and also how it should work.

but i agree that the wording can be improved, esp. `Are you sure you want to delete "Alice"?` may sound as if one is a killer :)

not sure, if the proposed wording `Delete chat with ...?` however, fits nicely to group names. maybe a simple `Delete chat "..."? It will no longer be shown in the chat list; its messages will remain on the server.` is better - and also fits the multi-select `Delete N chats? It will no longer be shown in the chat list; its messages will remain on the server.`

```suggestion
    <string name="ask_delete_named_chat">Delete chat \"%1$s\"? It will no longer be shown in the chat list; its messages will remain on the server.</string>
```

finally, once we agree on the wording (it makes sense to continue here :) this PR shoulds be closed and another one should be opened in deltachat-android, where the strings sources are placed.

r10s:
> I'm not sure about the translation process. Should make sense to update that for other translations.

the english sources for all of android/iOS/desktop are located at:

https://github.com/deltachat/deltachat-android/blob/main/src/main/res/values/strings.xml

from there, they are pushed to transifex for translations and the pulled from android/iOS/desktop on a regular base (i think in case of desktop, this is even automated)

_locales/en.xml
iequidoo: > It's weird, I haven't checked that yet in the core, but for me it displayed what's actually a `contact.displayName`. When I first looked at the isssue it indeed did display `address`. It prints `chat.name`, maybe something changed in core? I can't find it yet

I also can't reproduce this anymore even in DC Desktop 1.46.1 release, now the contact name is displayed. Maybe it happened under some unusual conditions or even was my mistake, unfortunately i didn't save any screenshots.

> maybe a simple `Delete chat "..."? It will no longer be shown in the chat list; its messages will remain on the server.` is better

+1, this fits well for both 1:1 and group chats.

maxphilippov: > I also can't reproduce this anymore even in DC Desktop 1.46.1 release, now the contact name is displayed. Maybe it happened under some unusual conditions or even was my mistake, unfortunately i didn't save any screenshots.

Yeah, I saw that one too, I don't exactly remember, it might have been 1.46.1 or 1.44 cause I had an outdated portable version for a while and I used current dev builds and 1.44 interchangeably

r10s:
i created a PR at https://github.com/deltachat/deltachat-android/pull/3219 , let's close this one.

once https://github.com/deltachat/deltachat-android/pull/3219 is merged, the string will be updated on desktop sooner or later

PR #4080, opened 2024-08-15
Always run configuration when user clicks "OK" in password dialog

link2xt:
If user wants to skip configuration,
there is a "Cancel" button.
If user clicks "OK" accidenatlly, worst case
configuration fails and does not change anything.

Rerunning configuration may be useful
if autoconfig XML got new servers that work.
This will also enable "automatic reconfiguration"
for accounts that use providers not in provider database once core PR <https://github.com/deltachat/deltachat-core-rust/pull/5866> is merged.

On Android opening configuration dialog and clicking confirm button is already enough, but on Desktop it requires adding and removing a symbol in the address or similar change before clicking OK.


nicodh (approved):
Just removed a now unused dependency...

PR #4082, opened 2024-08-16
Prepare 1.46.4

nicodh:


CHANGELOG.md
```diff
@@ -8,6 +8,16 @@
...
 
+<a id="1_46_4"></a>
+
+## [1.46.4] - 2024-08-16
+
+### Changed
+- Update `@deltachat/stdio-rpc-server` and `deltachat/jsonrpc-client` to `1.142.6`
+  - Still try to create "INBOX.DeltaChat" if couldn't create "DeltaChat" #5870
+  - Increase timeout for QR generation to 60s #5882
+  - Fix default to strict TLS checks if not configured #5888
```
dotlambda, line 19: Shouldn't we more explicitly ask people to change their settings if their account was created on version 1.46.3?

CHANGELOG.md
```diff
@@ -8,6 +8,16 @@
 
 ### Fixed
 
+<a id="1_46_4"></a>
+
+## [1.46.4] - 2024-08-16
+
+### Changed
+- Update `@deltachat/stdio-rpc-server` and `deltachat/jsonrpc-client` to `1.142.6`
+  - Still try to create "INBOX.DeltaChat" if couldn't create "DeltaChat" #5870
```
iequidoo, line 17: This matters only for some providers that require prefixes in folder names like "INBOX.". There was one more fix which seems more important:
fix: store_seen_flags_on_imap: Skip to next messages if couldn't select folder (https://github.com/deltachat/deltachat-core-rust/issues/5870)

CHANGELOG.md
nicodh: These changes were not made in dc desktop but in core and are mentioned here only for convenience...

iequidoo: Maybe they should rerun the configuration because "fix: default to strict TLS checks if not configured" might affect which IMAP/SMTP servers are chosen during the configuration. But it doesn't look critical, the new default should apply anyway. CC @link2xt

CHANGELOG.md
iequidoo: Yes, but mentioning both is also fine

dotlambda (approved):

CHANGELOG.md
nicodh: Hmm: when I look at https://github.com/deltachat/deltachat-core-rust/issues/5870 I see it is still open? So why is it mentioned in CHANGELOG? 
The issue worth to mention here would be "fix 'Couldn't select folder DeltaChat' log entry", since that's what desktop users may see
But the issue is not yet solved so I would not show both 

Anyway: this is just an additional information in the CHANGELOG so not worth to discuss in too much detail.
Users who are interested may look into the core CHANGELOG. So I added a direct link here

iequidoo: It's still open because there's one more improvement PR still unmerged. But i agree that not mentioning it at all is fine, i think it didn't affect many users

Simon-Laux (approved):

nicodh:
Updated core to 1.142.7
Added new introduced Contact property e2eeAvail https://github.com/deltachat/deltachat-core-rust/commit/fbf66ba02b73a9f9c77ee084872789fc660dbdf3

iequidoo (approved):

PR #4083, opened 2024-08-18
Add fallback to displayName

nicodh:
When editing the name in the EditProfile Dialog the form shows sometimes the name of the user sometimes not:

Bildschirmfoto vom 2024-08-18 18-55-02

The reason is: sometimes a user has a name besides the displayName sometimes not.

Is there a reason why we have 3 name properties for a Contact:
authName
displayName
name

### DIFF
--- a/src/renderer/components/dialogs/ViewProfile/index.tsx
+++ b/src/renderer/components/dialogs/ViewProfile/index.tsx
@@ -75,7 +75,7 @@ export default function ViewProfile(
 
   const onClickEdit = () => {
     openDialog(EditContactNameDialog, {
-      contactName: contact.name,
+      contactName: contact.name === '' ? contact.displayName : contact.name,
       onOk: async (contactName: string) => {
         await BackendRemote.rpc.changeContactName(
           accountId,


iequidoo:
> Is there a reason why we have 3 name properties for a Contact:
> authName
> displayName
> name

`name` is a name given by the user, so if the user didn't modify the contact name, it's empty. `authName` is a name authorised by the contact itself. So, maybe instead of defaulting to `displayName` better to make it clear why it's empty? E.g. show grayed out `(unmodifed)`/`(keep original)`/etc

nicodh:
Ok I just checked it: It's the same behaviour on Android app. But usability wise it is confusing to see a name of the contact, click on edit and have an empty input field. Is there any other usage of the display name? For the user there is just one contact name, so he won't differentiate between these names - or am I missing something?

iequidoo:
Now when i edit a contact name i see grayed out "Name" which jumps and becomes blue when the edit field gets focus. My suggestion is to replace this "Name" with the displayname, this way no translation is needed etc.

Simon-Laux:
Let's close this for now and discuss this UX problem in https://support.delta.chat/t/ux-for-setting-alias-name-for-contacts-is-not-obvious/3224, I came up with 2 possible solution ideas, please also share your ideas if you got any, see you over there

r10s:
> When editing the name in the EditProfile Dialog the form shows sometimes the name of the user sometimes not

to fix that issue at hand, and to not spend too much resources here and go for bigger fishes, is a simple one-liner to show "authName" as the placeholder. this is what ios is doing:

IMG_6216

an additional explaining text could be helpful as well, sure. 

that way, the dialog always shows the name that is actually used (tbh, not totally sure what the PR initially did, maybe exactly that :)

Simon-Laux:
> to not spend too much resources here and go for bigger fishes

Placeholder is a reasonable improvement that also came to my mind first, but it does not explain what this alias is.
That's why I spent some time to come up with 2 solution ideas in the forum.

> not totally sure what the PR initially did

preset the value to the auth name if display name is not set, so hiding issues like the (recently fixed?) bug where core or desktop (can't remember) set display names automatically and then when people changed their name others wouldn't see the new name.

PR #4093, opened 2024-08-24
Fix if Contact has long bio, then shared chats are hidden

### LINKED ISSUE #4092: Profile View: If Contact has long bio, then shared chats are hidden
Because shared chats just take the remaining space and if the bio/signature/status of a contact is long, then there is no space left.

Solution would be to set a minimum height:

```js
/* how many items should be shown atleast if they exist */
const minHeightMin = 1
const minHeightMax = 5
const chatlistitemHeight = /* the code to extract that from the css variable, look into chatlist for that */

const minHeight = chatlistitemHeight * Math.min(Math.max(minHeightMin, itemCount), minHeightMax)

// then in react code:
<div className='mutual-chats' style={{ flexGrow: 1, minHeight }}>
```

Simon-Laux:
fixes #4092


packages/frontend/src/components/dialogs/ViewProfile/index.tsx
```diff
@@ -220,6 +220,11 @@ export function ViewProfileInner({
     avatarPath = selfChatAvatar
   }
 
+  const mutualChatsMinItems = 5
```
WofWca, line 223: Isn't this supposed called `maxItems`? Or rather `maxMinWidthItems`

WofWca (approved):
Didn't check for weird interactions with `AutoSizer`, but the change makes sense.

packages/frontend/src/components/dialogs/ViewProfile/index.tsx
Simon-Laux: yeah `maxMinWidthItems` would be a more accurate name, thanks

WofWca: *height, sorry XD

WofWca (approved):

PR #4097, opened 2024-08-29
prepare 1.46.6

Simon-Laux:


WofWca (approved):
I only tested the Telegram change, but otherwise the MR looks fine.

PR #4103, opened 2024-09-03
Update core to 1.142.12

link2xt:


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 
 ### Changed
 
+- Update `@deltachat/stdio-rpc-server` and `deltachat/jsonrpc-client` to `1.142.12`
```
iequidoo, line 9: Maybe this should go to "### Fixed"

iequidoo (approved):

CHANGELOG.md
nicodh: in ###fixed we should only list the bugs that were fixed by the update, not the update itself...

iequidoo: I wanted to comment the next line actually, "Display `Config::MdnsEnabled` as true by default." is a fix, but not very important though

PR #4106, opened 2024-09-05
Fix addMember bg

nicodh:
Before
image

After
image


maxphilippov (approved):
Forgot Changelod.md (I always do) :) Or do we skip little things like this? Weird the variable wasn't used anywhere.

nicodh:
> Forgot Changelod.md (I always do) :) Or do we skip little things like this? Weird the variable wasn't used anywhere.

Yes I considered it too small to mention it in CHANGELOG. It also didn't have an issue...

PR #4108, opened 2024-09-05
Fix text added to webxdc message doesnt show line breaks

### LINKED ISSUE #4079: Multiline text added to a webxdc message is shown w/o newlines
- Operating System: Debian GNU/Linux 12.1
- Delta Chat Version: 1.46.3 (1.46.1 as well)
- Expected behavior: Multiline text added to a webxdc message is shown as multiline
- Actual behavior: It's shown w/o newlines
- Steps to reproduce the problem: Send or receive the corresponding message
- Screenshots: 
image

For other viewtypes it works. On Android also everything works (both messages are displayed correctly).

meganoahj:
fixes #4079

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
   - Display `Config::MdnsEnabled` as true by default.
 
 ### Fixed
+- Text sent with webxdc messages now respects line breaks #4108
 
 <a id="1_46_7"></a>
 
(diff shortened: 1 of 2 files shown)


maxphilippov:
I made a fix #4109 by mistake, I didn't see you took it @meganoahj. I'd argue we have to remove this rule completely. Since without display property margin doesn't make sense. And button in WebXDC content already has a margin around. I've looked up a commit when this was added it was done to make more space between button and text.

nicodh:
I agree that @maxphilippov fix is "a bit" cleaner...

Simon-Laux:
replaced by #4109

Issue #4104, opened 2024-09-03
Do not try to start I/O before deleting account


link2xt:
Currently if an account is imported from Android nightly or DeltaLab (now ArcaneChat) into the current Desktop with stable core 1.142.11, it fails to start I/O with a toast "Failed to start I/O: Invalid configured_imap_certificate_checks value". This is expected, user will need to update to the new core or reconfigure an account to make it usable.

The problem is that it is impossible to delete such broken account in the current Desktop 1.46.7. When user right-clicks the profile and selects "Delete Profile", desktop calls `selectAccount` here:
https://github.com/deltachat/deltachat-desktop/blob/9fdd6ea790b2027e021c6026d2c6c7e6c647b4dc/src/renderer/ScreenController.tsx#L189

`selectAccount` is this function that tries to start I/O and fails:
https://github.com/deltachat/deltachat-desktop/blob/9fdd6ea790b2027e021c6026d2c6c7e6c647b4dc/src/renderer/ScreenController.tsx#L119-L142

Desktop should not try to start I/O before deleting an account. If it fails, account cannot be deleted. Also if advanced setting "Only Synchronize Selected Profile" is enabled, I/O may have been stopped on purpose and starting I/O right before deleting an account is unnecessary.

adbenitez:
> This is useful for some people that need more control about what accounts connect in public networks.

for this in ArcaneChat we have `ui.enabled` and allow users to enable/disable accounts individually, that is more flexible than "enable selected account only" since then you can use ex. 2-3 needed accounts concurrently while the rest are disabled

Simon-Laux:
>  that is more flexible than "enable selected account only" since then you can use ex. 2-3 needed accounts concurrently while the rest are disabled

thats also what the core pr does. But there is still the usability question whether account should be started automatically when it is selected despite being disabled or if it should just stay disabled until the user enables it again.

adbenitez:
it should stay disabled, and a banner could be shown at the top saying "The profile is disabled  [ENABLE]" ot so

similar to what Conversations(xmpp) does for disabled accounts

PR #4112, opened 2024-09-07
fix: unable to delete broken account

WofWca:
...imported from ArcaneChat, generated by a newer version of core.

Closes https://github.com/deltachat/deltachat-desktop/issues/4104.

I am not sure whether the rest of the discussion in #4104 was relevant to this issue, but I tested for a few minutes and didn't find a case where I can't delete the account or was unable to use the working accounts. Tried both with "Only Synchronize Selected Account" and without.

### DIFF
--- a/src/renderer/ScreenController.tsx
+++ b/src/renderer/ScreenController.tsx
@@ -137,8 +137,12 @@ export default class ScreenController extends Component {
 
     await BackendRemote.rpc.startIo(accountId)
     runtime.setDesktopSetting('lastAccount', accountId)
-    log.info('system_info', await BackendRemote.rpc.getSystemInfo())
-    log.info('account_info', await BackendRemote.rpc.getInfo(accountId))
+    BackendRemote.rpc.getSystemInfo().then(info => {
+      log.info('system_info', info)
+    })
+    BackendRemote.rpc.getInfo(accountId).then(info => {
+      log.info('account_info', info)
+    })
   }
 
   async unSelectAccount() {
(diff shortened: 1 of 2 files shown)


adbenitez (approved):
tested, works like a charm, was able to delete the account after importing it, even if IO was broken

nicodh (approved):
Tested and reviewed

src/renderer/ScreenController.tsx
```diff
@@ -135,10 +135,16 @@ export default class ScreenController extends Component {
...
+    }
     runtime.setDesktopSetting('lastAccount', accountId)
-    log.info('system_info', await BackendRemote.rpc.getSystemInfo())
-    log.info('account_info', await BackendRemote.rpc.getInfo(accountId))
+    BackendRemote.rpc.getSystemInfo().then(info => {
+      log.info('system_info', info)
+    })
+    BackendRemote.rpc.getInfo(accountId).then(info => {
+      log.info('account_info', info)
+    })
```
Simon-Laux, line 147: what's the reason for this change?
make account switching faster by not waiting for the result? or did the calls fail?

Simon-Laux (approved):
Code LGTM

src/renderer/ScreenController.tsx
Simon-Laux: acording to commit message I assume the first.

WofWca: It's both

PR #4115, opened 2024-09-09
Update electron from `30.3.1` to `32.0.2`

### LINKED ISSUE #4111: Update electron to version 32
**Why?** because arch and other distort drop support for unsupported electron versions.
Also there surely were some improvements and bugfixes to electron itself.

**When?** Either wait for monorepo branch (#4074) to be merged or make a pr on top of the monorepo branch that merges into the monorepo branch.

**How?**
read the blogposts (and maybe also changelog) of the new versions to migrate to new apis if necessary:
- https://www.electronjs.org/blog/electron-31-0
- https://www.electronjs.org/blog/electron-32-0

Simon-Laux:
closes #4111

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 ### Changed
+- Update electron from `30.3.1` to `32.0.2`
 - dev: transformed repo into monorepo
 - dev: switched from `npm` to `pnpm`
 - dev: esbuild bundling for electron main process js (+minification for releases)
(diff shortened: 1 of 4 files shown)


nicodh (commented):
How is the new clearData command related to the upgrade?
In the elctron docs it says about clearData: This method clears more types of data and is more thorough than the clearStorageData method. 
https://www.electronjs.org/docs/latest/api/session

So do we really need to call both clearStorageData() and clearData()?

Simon-Laux:
> So do we really need to call both clearStorageData() and clearData()?

the pr on electron that introduced it says that the new method cleans more kinds of data like some network history data.
https://github.com/electron/electron/pull/40983#issuecomment-1939374825

Issue #3763, opened 2024-04-09
Typing a multi-line message makes chat scroll up (then you have to scroll to see new messages)


WofWca:
- Operating System (Linux/Mac/Windows/iOS/Android): Windows
- Delta Chat Version: 1.44.0
- Expected behavior: The last message stays visible
- Actual behavior: The chat scrolls up a little and the last message is obstructed. If new messages arrive, they are also obstructed.
- Steps to reproduce the problem:
- Screenshots:

    https://github.com/deltachat/deltachat-desktop/assets/39462442/ae733a96-0b52-4d01-b472-c34b374e0b6e

- Logs: none

~~On the video I've caught a rare case where this doesn't happen for a few lines, but then it starts happening again.~~
Update: see https://github.com/deltachat/deltachat-desktop/issues/3763#issuecomment-2239170621.
I always found it about the most annoying thing about Delta Chat Desktop.

- [x] This thing also happens when you quote a message.
- [ ] And when you resize the window down.

Workaround: keep pressing "Page Down" while you're writing your message.

Related:
- https://github.com/deltachat/deltachat-desktop/issues/3753#issuecomment-2045828431 - may be a part of the fix
- #654

missytake:
Oh thanks this is annoying me so much, I tried to report it once, but then when I failed to reproduce it on purpose

WofWca:
Ok, a few discoveries. I wrote this test simple bench:

<details><summary>HTML</summary>

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>Document</title>
</head>
<body style="height: 100vh; display: flex; flex-direction: column; margin: 0;">
    <section style="overflow-y: auto; flex-grow: 1;
        max-width: 100%;">
        <ol style="margin: 0;">
            <li>sample message</li>
            <li>sample message</li>
            <li>sample message</li>
            <li>sample message</li>
            <li>sample message</li>
            <li>sample message</li>
            <li>sample message</li>
            <li>sample message</li>
            <li>sample message</li>
            <li>sample message</li>
            <li>sample message</li>
            <li>sample message</li>
            <li>sample message</li>
            <li>sample message</li>
            <li>sample message</li>
            <li>sample message</li>
            <li>sample message</li>
            <li>sample message</li>
            <li>sample message</li>
            <li>sample message</li>
            <li>sample message</li>
[...]

PR #4119, opened 2024-09-10
fix: new messages not getting scrolled into view

WofWca:
...when typing a multi-line message, quoting a message, or adding an attachment, because that would cause the chat to "scroll up" due to its resize.

Closes https://github.com/deltachat/deltachat-desktop/issues/3763

I branched off of v1.46.8 FYI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 - fix chat "scrolling up" when someone adds a reaction, resulting in new messages not getting scrolled into view when they arrive #4120
 - fix missing space between overriden sender name and image attachment #3914
 - when adding new line for a long multi-line message, the cursor would get out of view #4152
+- Chat "scrolling up" when typing a multi-line message, quoting a message, or adding an attachment, resulting in new messages not getting scrolled into view #4119
 
 <a id="1_46_8"></a>
 
--- a/packages/frontend/src/components/message/MessageList.tsx
+++ b/packages/frontend/src/components/message/MessageList.tsx
@@ -131,6 +131,12 @@ export default function MessageList({ accountId, chat, refComposer }: Props) {
   const messageListRef = useRef<HTMLDivElement | null>(null)
   const [showJumpDownButton, setShowJumpDownButton] = useState(false)
 
+  /**
+   * If scroll distance to bottom is bigger than this, we'll show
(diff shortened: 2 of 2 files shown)


src/renderer/components/message/MessageList.tsx
```diff
@@ -409,6 +409,69 @@ export default function MessageList({ accountId, chat, refComposer }: Props) {
...
+  // when the scrollable element itself resizes, and not just
+  // the content inside of it.
+  //
+  // A probably better approach would be to use
+  // `flex-direction: column-reverse;`, as in
+  // https://github.com/deltachat/deltachat-desktop/pull/4116,
+  // but it is buggy in Chromium and maybe WebKit (which we'll be using
+  // when we switch to Tauri).
+  //
+  // Should this be `useLayoutEffect`? This appears to be working fine though.
```
Simon-Laux, line 435: I think `useEffect` is better here, because the function records the first size, I guess it could not do that when the browser hasn't painted the element yet?

Simon-Laux:
I get many of these messages, maybe there is a way to prevent the loading of new messages from scrolling in this special case?
Bildschirmfoto 2024-09-10 um 23 28 30

src/renderer/components/message/MessageList.tsx
```diff
@@ -409,6 +409,69 @@ export default function MessageList({ accountId, chat, refComposer }: Props) {
...
+      }
+      const newHeight = el.clientHeight
+      if (newHeight === prevHeight) {
+        // In case only width changed or something.
+        return
+      }
+
+      const scrollDistanceToBottomBeforeResize =
+        el.scrollHeight - el.scrollTop - prevHeight
+      if (scrollDistanceToBottomBeforeResize < 7) {
```
Simon-Laux, line 457: 7 should be a named constant, that we can then ensure to be identical to the constant / threshold when the jump down button is shown.

WofWca:
> maybe there is a way to prevent the loading of new messages from scrolling in this special case?

I don't think there is an easy way to do this. And this used to happen even before this MR.
I think we'd want to switch from `onScroll` to `IntersectionObserver` in the future to detect whether we're close to the edge.

nicodh (approved):
Tested on Mac & Linux

PR #4120, opened 2024-09-10
fix: chat "scrolling up" upon reaction

WofWca:
...resulting in new messages not getting scrolled into view when they arrive.

This simply removes height changes between messages with / without a reaction.

Closes https://github.com/deltachat/deltachat-desktop/issues/3753


WofWca:
I was thinking of putting this into version 1.46.8

nicodh:
Yes but it's easy to apply it to the new package/frontend source. So hopefully not too much work :-)

nicodh:
I tested it by patching main since I have main already updated

Simon-Laux:
> I was thinking of putting this into version 1.46.8

we can still back port it afterwards if the next release takes too long.

Simon-Laux (approved):
code LGTM

PR #4125, opened 2024-09-12
improvement: scroll to new messages smoothly

WofWca:
https://github.com/user-attachments/assets/90f3a24d-80f8-4876-ae6c-3931120aa647

TODO:
- [x] Test it and see if it gets stuck often enough
  for it to be critical. I have seen it happen a few times.
  Maybe for 1 out of 100 messages.
- [ ] Think about putting behind an experimental setting?
  Ok, I have an idea. Let's merge it, and if it's not good yet, then put it behind the setting.
- [x] Add CHANGELOG entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ## Added
 - Add "Learn More" button to "Disappearing Messages" dialog #4330
 - new icon for Mac users
+- smooth-scroll to newly arriving messages instead of jumping instantly #4125
 
 ## Changed
 - enable Telegram-style Ctrl + ArrowUp to reply by default #4333
--- a/packages/frontend/src/components/message/MessageList.tsx
+++ b/packages/frontend/src/components/message/MessageList.tsx
@@ -209,6 +209,9 @@ export default function MessageList({ accountId, chat, refComposer }: Props) {
     }
   }, [jumpToMessage])
 
+  const pendingProgrammaticSmoothScrollTo = useRef<null | number>(null)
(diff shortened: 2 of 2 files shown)


nicodh:
>  Test it and see if it gets stuck often enough
> for it to be critical. I have seen it happen a few times.

Can you describe how to test and provoke "critical" behaviour? 

> Maybe for 1 out of 100 messages.
So this means for testing: it needs that many repetitions to run into an failure? What kind of failure would that be?

WofWca:
> Can you describe how to test and provoke "critical" behaviour?

I mean the behavior we observed in #4186, #3763, #3753. By "critical" I mean whether the occasional "scroll up" outweighs the niceness of the smooth scroll, and so whether we want to merge this, without making it an experimental setting.

> So this means for testing: it needs that many repetitions to run into an failure?

Yes, it results in a "scrolled up" chat once every ~100 smooth scrolls.

nicodh:
> I mean the behavior we observed in #4186, #3763, #3753. By "critical" I mean whether the occasional "scroll up" outweighs the niceness of the smooth scroll, and so whether we want to merge this, without making it an experimental setting.

These bugs are all solved, so I still don't get what this PR fixes/improves

Can you add a video like "before" and "after"

WofWca:
> These bugs are all solved, so I still don't get what this PR fixes/improves

Those are solved, but this MR might introduce them back (I mean, not those bugs precisely, but their effect of "scrolling up").

I edited the original post by adding a video.

PR #4136, opened 2024-09-16
fix: inconsistent & wrong `composer.focus()`

WofWca:
- Erasing draft text and immediately focusing another element would return the focus to the composer after 500ms, when the draft gets deleted (`clearDraft` would get invoked from `saveDraft()`.
- Composer would not get re-focused when deleting / adding attachments or removing quotation if the draft text is not empty.

`.focus()` inside of `clearDraft()` was introduced in 199ea232539e77ea9b4c903451b90a56dc880554, most likely to re-focus the input after sending a message. We now have separate code for this, see `composerSendMessage`.

This change likely needs some more testing.


nicodh (approved):
By reading and testing.

PR #4137, opened 2024-09-16
improvement: scroll selected account into view

WofWca:
...on initial render, and if needed when changing selection.

### DIFF
--- a/packages/frontend/src/components/AccountListSidebar/AccountItem.tsx
+++ b/packages/frontend/src/components/AccountListSidebar/AccountItem.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react'
+import React, { useEffect, useLayoutEffect, useRef, useState } from 'react'
 import classNames from 'classnames'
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/AccountListSidebar/AccountItem.tsx
```diff
@@ -147,6 +147,41 @@ export default function AccountItem({
...
+      behavior: 'instant',
+      // "nearest" so as to not scroll if it's already in view.
+      block: 'nearest',
+      inline: 'nearest',
+    })
+    // `window.__screen` because we display the "settings" button
+    // based on that, and whether it is displayed or not determines the
+    // scrollHeight of the accounts list, so we want to make sure
+    // to scroll after it gets displayed.
+    // TODO refactor: maybe just use `ResizeObserver`,
```
Simon-Laux, line 179: also sounds good because users may change the height of the window and thus scroll the selected account out of view.

Simon-Laux (approved):
works

PR #4138, opened 2024-09-16
Update electron from `32.0.2` to `32.1.0`

Simon-Laux:


WofWca (commented):
I think it wants an update to `package.json` as well. `pnpm install` should fix it.

WofWca (approved):
Works

PR #4142, opened 2024-09-19
Treat overriden sender name as content above attachment

### LINKED ISSUE #3914: layout bug in image message from self with overridden sender name
when the self-message has overridden sender name set:
image

for forwarded messages it is displayed properly:
image

maxphilippov:
- resolves #3914


Simon-Laux (approved):
code looks good, didn't test. I think this can be tested with adb's community experiment or using the dev console to send image messages with overriden sender name.

PR #4143, opened 2024-09-21
Change sticker reply background to make name more readable

### LINKED ISSUE #3291: hard to read color in quoted message with sticker reply
if  someone replies a message with a sticker, the author name and vertical bar should have the same color than text, otherwise it is not readable and looks bad

image

I had the impression this issue was solved in the past, maybe it is a regression in the latest version?

maxphilippov:
- resolves #3291

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@
 - accessibility: wrong positioning of some context menus and popups when activating them with keyboard #4246
 - improve keyboard and screen-reader accessibility #4210
 - "Page Up" / "Page Down" not working on scrollable elements except for messages list #4269
+- make name more readable in sticker reply #3291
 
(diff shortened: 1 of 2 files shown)


maxphilippov:
I tried making it a text color but for me on Windows but the name was blending with text. And I personally find color-coding useful to quickly search through the group dialogs. It might confuse something if that's a message not a quote, so I need feedback on this one.

_Originally posted by @maxphilippov in https://github.com/deltachat/deltachat-desktop/issues/3291#issuecomment-2365028493_

nicodh (approved):
Looks good (also tested). Concerning the colors: incomingMsgBg makes sense to me. It will have the same contrast as other incoming messages, and the contrast in the sent message is not that relevant since it shows my own name.

PR #4144, opened 2024-09-23
Save draft on chat change

### LINKED ISSUE #3733: Message draft discarded when switching to different chat quickly
- Delta Chat Version: c878fc78
- Expected behavior: Draft message saved
- Actual behavior: Draft message remains unchanged
    1. Open a chat
    2. Write something without pausing for more than a second
    3. Switch to a different chat within a second after making the last change to the draft.
- Screenshots:
image

- Logs:

Example use case: paste something in the message composer and then immediately switch to a different chat.

This has to do with this debounce

https://github.com/deltachat/deltachat-desktop/blob/584b59c1149223d04d8497f862955015248c77b7/src/renderer/components/composer/ComposerMessageInput.tsx#L49-L52

when the chat is switched, the state of the component changes and its text gets assigned the value of the other chat's text, so the draft of the previous chat is lost.
I have attemted to `updateDraftText()` with `prevState.text`, but it refuses to save the draft here

https://github.com/deltachat/deltachat-desktop/blob/584b59c1149223d04d8497f862955015248c77b7/src/renderer/components/composer/Composer.tsx#L539-L541

Then I thought I'd call `BackendRemote.rpc.miscSetDraft()` anyway, but it looks like it depends on a ton of other state, so we need `prevState` there as well I guess.

https://github.com/deltachat/deltachat-desktop/blob/584b59c1149223d04d8497f862955015248c77b7/src/renderer/components/composer/Composer.tsx#L503-L510

maxphilippov:
- resolves #3733


maxphilippov:
This is a thing we discussed a long time ago, we had a 1 second debounce for saving draft, so if you're in any way fast at switching between different chats you just lose what you were typing right before switching. We were talking about testing something around 200-300ms, I went for 100.

While I have a pretty beefy machine (some recent Ryzen 5), I tested this on full power saving both with no attached files and with 200Mb files attached as a draft. I don't see any signs of degradation in how it feels to type. With a 200Mb file attached I have some weird freezes both on `main` and here (if I switch to another contact or sometimes my mouse cursor doesn't move while typing is still perfectly smooth).

So I'm looking for feedback from people who might find a very slow machine and test it on and compare to `main` typing experience. For me 100ms is fast enough no matter how I try I don't lose any draft pieces.

maxphilippov:
I did keep the original 1 second delay between autosave

packages/frontend/src/components/composer/ComposerMessageInput.tsx
```diff
@@ -29,7 +29,7 @@ export default class ComposerMessageInput extends React.Component<
   composerSize: number
   setCursorPosition: number | false
   textareaRef: React.RefObject<HTMLTextAreaElement>
-  saveDraft: () => void
+  saveDraft: any
```
WofWca, line 32:
```suggestion
  saveDraft: ReturnType<typeof debounce<(text: string, chatId: number) => void>>
```

CHANGELOG.md
```diff
@@ -53,6 +53,7 @@
 - cancel old message highlight animations when a new message is highlighted #4203
 - fix: packaging: include architecture in filename for all appimages #4202
 - fix: make open external link scheme case insensive #4201
+- fix: save draft when you switch to another dialog #3733
```
WofWca, line 56:
```suggestion
- fix: save draft when you switch to another chat #3733
```

Another chat?

packages/frontend/src/components/composer/ComposerMessageInput.tsx
```diff
@@ -124,7 +123,7 @@ export default class ComposerMessageInput extends React.Component<
       this.resizeTextareaAndComposer()
       if (prevState.text !== this.state.text) {
         if (!this.state.loadingDraft) {
-          this.saveDraft()
+          this.saveDraft.flush(this.state.text, prevState.chatId)
```
WofWca, line 126: `flush()` takes no arguments.

WofWca (changes requested):
So do you suggest to save the draft on every key press? Because this is what's happening in this MR, and it doesn't seem intentional.

maxphilippov:
> So do you suggest to save the draft on every key press? Because this is what's happening in this MR, and it doesn't seem intentional.

Oops, my bad, rushed that a bit. Didn't notice we have a setState above. I'll draft it for now. Reverted back to 200ms, removed saveDraft from onChange then. Gonna look for a way to maybe flush via ref to messageInputComposer. I don't agree with throttle, since I don't see the point in saving first character immediately and thus pushing next save further in time. Next save is what's more likely to happen on chat change.

WofWca:
> thus pushing next save further in time

I don't think that's how throttle works. Throttle is never less frequent than debounce.

maxphilippov:
As discussed with @nicodh, I've added a simple throttle implementation and made saveDraft use this one instead of debounce.

packages/shared/util.ts
```diff
@@ -14,3 +14,30 @@ export function truncateText(text: string, max_len: number) {
...
+  return (...args: A) => {
+    if (!inThrottle) {
+      fn(...args)
+      lastTime = performance.now()
+      inThrottle = true
+    } else {
+      clearTimeout(timeout)
+      timeout = setTimeout(
+        () => {
+          if (performance.now() - lastTime >= wait) {
```
WofWca, line 34: So if this condition is false, the last call of the throttled function will not invoke `fn`?

Why not use an existing package?

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 - "Disappearing Messages" dialog not reflecting the actual current value #4327
 - accessibility: make settings keyboard-navigable #4319
 - Fix documentation for --allow-unsafe-core-replacement #4341
+- fix: save draft when you switch to another chat #3733
```
WofWca, line 16: This doesn't sound entirely accurate.

packages/frontend/src/components/composer/ComposerMessageInput.tsx
```diff
@@ -124,17 +123,14 @@ export default class ComposerMessageInput extends React.Component<
       this.resizeTextareaAndComposer()
       if (prevState.text !== this.state.text) {
         if (!this.state.loadingDraft) {
-          this.saveDraft()
+          this.saveDraft(this.state.text, prevState.chatId)
```
WofWca, line 126: Why use `prevState.chatId` here instead of `this.state.chatId`? They're the same inside this `if` block.

packages/frontend/src/components/composer/ComposerMessageInput.tsx
```diff
@@ -46,10 +46,9 @@ export default class ComposerMessageInput extends React.Component<
     this.insertStringAtCursorPosition =
       this.insertStringAtCursorPosition.bind(this)
 
-    this.saveDraft = debounce(() => {
-      const { text, chatId } = this.state
+    this.saveDraft = throttle((text: string, chatId: number) => {
       this.props.updateDraftText(text.trim() === '' ? '' : text, chatId)
-    }, 1000)
+    }, 200)
```
WofWca, line 51: Is the timer supposed to be lowered still?

packages/frontend/src/components/composer/ComposerMessageInput.tsx
```diff
@@ -46,10 +46,9 @@ export default class ComposerMessageInput extends React.Component<
     this.insertStringAtCursorPosition =
       this.insertStringAtCursorPosition.bind(this)
 
-    this.saveDraft = debounce(() => {
-      const { text, chatId } = this.state
+    this.saveDraft = throttle((text: string, chatId: number) => {
```
WofWca, line 49: Am I understanding this correctly that the switching from `const { text, chatId } = this.state` to `(text: string, chatId: number) =>` has no functional significance?

packages/frontend/src/components/composer/ComposerMessageInput.tsx
```diff
@@ -124,17 +123,14 @@ export default class ComposerMessageInput extends React.Component<
...
         }
       }
     }
   }
 
   onChange(e: React.ChangeEvent<HTMLTextAreaElement>) {
     this.setState({ text: e.target.value /*error: false*/ })
-    if (!this.state.loadingDraft) {
-      this.saveDraft()
-    }
```
WofWca, line 137: Why is this removed? Is it just a refactor?

packages/frontend/src/components/composer/ComposerMessageInput.tsx
nicodh: I think this is sufficient at least in my tests. Somehow the saveDraft seems to be blocking anyway so if I click on a chat or account immediately after typing the click has no effect.
I still managed to have a missing character in the draft or one left over after deleting but I think that's ok. I had to type very fast and click mutiple times on another chat to achieve that.

WofWca: > so if I click on a chat or account immediately after typing the click has no effect.

Uhh, is this also on the 1.48 release version?

WofWca: > I think this is sufficient at least in my tests.

What I meant is that is it supposed to be lowered to 200 instead of staying at 1000. Because switching to `throttle` is supposed to be enough.

nicodh: > Uhh, is this also on the 1.48 release version?

Yes with current master.

packages/frontend/src/components/composer/ComposerMessageInput.tsx
nicodh: These lines exist also in onComponentDidUpdate what doesn't make sense to me, since switching to another chat also triggers onComponentDidUpdate so in the end a fresh loaded draft is immediately saved again. So why not remove them there and leave them here?

packages/frontend/src/components/composer/ComposerMessageInput.tsx
nicodh: It even blocks my mouse pointer whe I try to move it
On current main
I enter fast 5-6 characters and then try to click somewhere  - it takes until the draft is saved before I can interact again

packages/frontend/src/components/composer/ComposerMessageInput.tsx
WofWca: Makes sense.
But the question is if it's directly related to what this commit is doing.
Consider this a nitpick then.

WofWca (approved):
Tested for a while. Appears to work alright.
I also looked at the `throttle` implementation again, and it appears to be OK for the current purposes.

If the emoji thing is fine, then let's merge.
Hopefully there are no bad side effects in the refactoring part of this MR.

nicodh (approved):

PR #4145, opened 2024-09-24
perf: improve message list performance

WofWca:
Recommended review (and merge) method: commit-by-commit.
I've tested the changes. Encoutered a few bugs with `jumpToMessage` not working, but fixed them.

### DIFF
--- a/packages/frontend/src/stores/messagelist.ts
+++ b/packages/frontend/src/stores/messagelist.ts
@@ -402,6 +402,8 @@ class MessageListStore extends Store<MessageListState> {
         highlight = true,
         addMessageIdToStack?: undefined | number
       ) => {
+        const startTime = performance.now()
+
         this.log.debug('jumpToMessage with messageId: ', msgId)
         const accountId = selectedAccountId()
         // these methods were called in backend before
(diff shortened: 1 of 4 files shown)


packages/frontend/src/stores/messagelist.ts
```diff
@@ -309,31 +309,33 @@ class MessageListStore extends Store<MessageListState> {
...
 
-        const firstUnreadMsgId =
-          await BackendRemote.rpc.getFirstUnreadMessageOfChat(
-            this.accountId,
-            this.chatId
-          )
+        const firstUnreadMsgId = await firstUnreadMsgIdP
         if (firstUnreadMsgId !== null) {
-          setTimeout(() => {
+          setTimeout(async () => {
```
nicodh, line 337: here a comment would be nice:
"interrupt loadChat here and switch to jumpToMessage which loads chat and messageListItems"

to explain why we return false

WofWca: Perhaps, though adding comments is not a part of this MR.

nicodh: Yes. That's why I approved it. I added it only, since you added comments in other parts of your changes which I appreciate :-)

Issue #4046, opened 2024-07-23
"Block Contact" chat menu item displayed for an already blocked contact instead of "Unblock"


iequidoo:
First of all, i believe this should be moved from the chat menu to the contact profile menu, see #4043 for discussion. At least on Android "Block/Unblock Contact" is there. (On Desktop there's no contact profile menu currently, you can only edit the contact name.) The second is that if a contact is already blocked, "Unblock Contact" should be displayed instead (as on Android). The third is that currently if i e.g. forgot that the contact is already blocked and do "Block Contact" again, the 1:1 chat is only deselected, but remains in the chatlist. But maybe this doesn't need to be fixed if "Unblock Contact" will be in the menu instead.

For reproduction: if you have a group with an already blocked contact, you can get to the 1:1 chat with it via the group members window >> click on the contact >> SEND MESSAGE.

r10s:
> the 1:1 chat is only deselected, but remains in the chatlist

i cannot reproduce that part. at least on my test on current master, blocked chats are directly removed from the chatlist.

+1 for the other points.

iequidoo:
> i cannot reproduce that part. at least on my test on current master, blocked chats are directly removed from the chatlist. maybe there is some race, however.

For me blocked 1:1 chats are removed from the chatlist only on first blocking. If i lookup the chat somehow again (e.g. via a common group with the contact), it appears in the chatlist again and doesn't go away if i retry blocking.

nicodh:
Maybe we could show a dialog: You are sending a message  to a blocked user, and if confirmed first unblock the user and then send the message

iequidoo:
> I think the best solution is: avoid chats with blocked contacts at all. So hide the "Send message" button for blocked users

So if you blocked a contact, you can't even see the chat to remember why it's blocked.

In the "Contact" window menu "Unblock Contact" is already shown as expected, why can't we do the same for the 1:1 chat menu? The issue is mainly about this. But the UI can be even unified with DC Android where there's no "Block/Unblock Contact" item in the chat menu at all.

nicodh:
> So if you blocked a contact, you can't even see the chat to remember why it's blocked.

I would expect that chats with blocked users don't show up any more. If you don't remember why you blocked someone you can unblock and look at the related chats. We also don't keept deleted chats to let you remember why you deleted a chat.

[...]

PR #4147, opened 2024-09-25
Add block contact and encryption info to profile

### LINKED ISSUE #4043: No way to do almost anything with a contact w/o having a 1:1 chat
If i only have some group with a contact, but not a 1:1 chat, i can't do almost anything with it, e.g. see "Encryption Info" or "Block Contact" (which is possible in DC Android f.e.). I only can "Edit Name". Ok, i created a 1:1 chat with the contact, now i have the extended menu (btw, i still see "Encryption Info" only in the chatlist ctx menu, but not in the chat one). Then i tried to "Block Contact", after that the 1:1 chat was deselected and i see the message "Select a chat or create a new chat". Why deselect the chat btw? I selected the 1:1 chat back, it's still empty, nothing has changed visually and if i open the menu again, i see "Block Contact" while expecting to have "Unblock Contact". So, i can block the contact again and if i do that, the same effects as described before occur. My suggestions:
- **Add the extended menu (with "Encryption Info" etc. -- as on Android) to the contact window (where the user only can "Edit Name" now).**
- Add "Encryption Info" to the 1:1 chat menu (looks like it's just forgotten, other menu items are there). **EDIT:** See the discussion, this is not needed, e.g. DC Android doesn't have it there.
- Reconsider deselecting the 1:1 chat after blocking a contact. **EDIT:** It's ok, the chat should be hidden after blocking the contact.
- Debug why a contact with an empty 1:1 chat can be blocked multiple times w/o any effects. **EDIT:** This happens not for a contact with an empty 1:1, but for an already blocked contact. Not a case for DC Android, see below. Created a separate issue: #4046.

maxphilippov:
- new `useViewProfileMenu` menu with two items: Block/Unblock Contact and Encryption Info
- minor refactor of `useChatDialog`, so we don't have to have a ChatListItem to block contact or get EncryptionInfo
- resolves #4043

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 - Experimental Telegram-style Esc to cancel reply (quote) #4153
+- new ViewProfile context menu for blocking/unblocking contact and checking encryption #4043
 
 ### Changed
 - Update electron from `30.3.1` to `32.1.0` #4138
--- a/packages/frontend/src/components/Dialog/HeaderButton.tsx
+++ b/packages/frontend/src/components/Dialog/HeaderButton.tsx
@@ -11,17 +11,24 @@ import styles from './styles.module.scss'
 type Props = ButtonHTMLAttributes<HTMLButtonElement> & {
   icon: IconName
   iconSize: number
+  rotation?: number
 }
 
 export default function HeaderButton({
   className,
   icon,
   iconSize,
+  rotation,
   ...props
 }: Props) {
   return (
     <button className={classNames(styles.headerButton, className)} {...props}>
-      <Icon className={styles.headerButtonIcon} icon={icon} size={iconSize} />
+      <Icon
+        className={styles.headerButtonIcon}
+        icon={icon}
+        size={iconSize}
+        rotation={rotation || 0}
+      />
     </button>
   )
 }
--- a/packages/frontend/src/components/chat/ChatListContextMenu.tsx
+++ b/packages/frontend/src/components/chat/ChatListContextMenu.tsx
@@ -107,7 +107,11 @@ export function useChatListContextMenu(): {
     openContextMenu: async (event, chatListItem, selectedChatId) => {
       const onDeleteChat = () =>
         openDeleteChatDialog(accountId, chatListItem, selectedChatId)
-      const onEncrInfo = () => openEncryptionInfoDialog(chatListItem)
+      const onEncrInfo = () =>
+        openEncryptionInfoDialog({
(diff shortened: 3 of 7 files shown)


nicodh (changes requested):
There are 3 options to block a contact now:
1. from contact context menu in chat list
2. from context menu in chat view
3. from settings menu in ViewProfile

Depending on the use case the behaviour is different:
1. if you use option 1 while the chat is not open, everything is fine
2. if you use option 1 while the chat is open, the chat stays open
3. if you use option 2 while the chat is open, the chat is unselected and closed
4. if you use option 3 while the chat is open, the ViewProfile stays open, the chat disappears from chat list, but the chat is still open when you close the ViewProfile dialog
5. in an open chat (remained after blocking a contact) the context menu still shows: "Block contact" instead of "unblock contact"

I suggest to always use the same behaviour as in  chat views context menu, that means after blocking a contact the current chat should be unselected if it is the 1:1 chat with the blocked user. If the user is blocked from "ViewProfile" dialog the dialog should be closed too. Otherwise we would have to add a flag to show that a user is blocked.

maxphilippov:
> 5. in an open chat (remained after blocking a contact) the context menu still shows: "Block contact" instead of "unblock contact"
There's a separate issue #4046 for this and from what I saw looking at this it seems more like a chatlist problem so I don't want to mix it up here, unless it's a minor fix, I'll push it as a separate PR.

Thanks for other points! I'll reiterate on that (+ menu styling) and push an update soon

nicodh:
> Thanks for other points! I'll reiterate on that (+ menu styling) and push an update soon

@maxphilippov please let me know when you updated this

maxphilippov:
@nicodh everything is ready for review. I've also moved "Edit Name" to this new menu as we discussed with @r10s. There also rises a question if we need the same menu for groups but with only "Encryption Info" + "Edit Name" but that's for another PR probably.

nicodh (approved):

r10s:
> There also rises a question if we need the same menu for groups but with only "Encryption Info" + "Edit Name" but that's for another PR probably.

for groups, editing name+image is far more important than setting a nickname, so i would not suggest to move that to the "three dot menu" without other changes (eg. being able to access "Edit" also from enlarged avatar view as on android).

still, for encryption info, the three dot menu makes sense for groups, so that this is consistent.

when we have that consistent, also in another PR, we can probably also remove  "Encryption Info" from the chatlist's context menus. 

however, that is future and shall not block merging this PR :) all in all a nice improvement, thanks a lot.

PR #4148, opened 2024-09-27
Scroll to bottom when composer input is higher than maxScrollHeight

### LINKED ISSUE #3774: Composer view doesn't show the line with the cursor if you add too many new lines
- Operating System (Linux/Mac/Windows/iOS/Android): .deb
- Delta Chat Version: 1.44.1
- Expected behavior: In the composer area, the text cursor always stays in view.
- Actual behavior: if you have too many lines already, and you add a new line in the bottom, the view doesn't follow the text cursor.
  - open a chat
  - click into the composer view
  - press enter multiple times, until the composer area doesn't get larger anymore and a scroll bar appears (shift+enter if you have enabled "Pressing the Enter Key will send text messages")
  - now the text cursor is out of view

Screencast of the bug: https://cloud.0x90.space/s/XygYRgGMpphqe2b

nicodh:
resolves #3774

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - dev: improved `./bin/test_for_missing_translations.sh` script: It is now more correct, helpful and faster
 - windows 64bit and 32bit protable and setup now have different filenames #4131
 - scroll the selected account into view in the accounts sidebar #4137
+- keep cursor visible in composer even with many linebreaks #3774
 
 ### Fixed
 - fix that you can not click header button in dialog when they are on top of the navbar #4093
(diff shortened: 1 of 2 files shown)


WofWca (changes requested):
https://github.com/user-attachments/assets/fdfb82e1-e2a8-4ff8-8b10-8f4c28f0acdd

This is acting real funky. I'd argue it's worse. This `resize` function is called quite frequently, e.g. also when the draft gets saved. So this line doesn't actually keep the cursor visible. Only in the particular case where it's a new line at the end of the input.

nicodh:
Good catch! I didn't test the case where the cursor is not at the end (which was the case where this bug occured). 
Should be fixed now with the additional condition.

WofWca (commented):
There is still a problem where if you add some text at the end of the composer, then scroll the composer up immediately, it would scroll back to bottom as the draft gets saved.

Please see if you like my approach better https://github.com/deltachat/deltachat-desktop/pull/4152

nicodh:
I think https://github.com/deltachat/deltachat-desktop/pull/4152 is a better solution! I didn't see that there was a artificial behaviour added with this.insertStringAtCursorPosition('\n')

PR #4150, opened 2024-09-28
refactor: add a docstring for `isInviteLink`

WofWca:
To clarify this is not just about contacts, but also groups.

### DIFF
--- a/packages/shared/util.ts
+++ b/packages/shared/util.ts
@@ -6,6 +6,10 @@ export function truncateText(text: string, max_len: number) {
   }
 }
 
+/**
+ * @returns Whether the `url` is a contact or a group invite link.
+ * (https://github.com/deltachat/invite/blob/1b5b2e2a9bfcaf87c4f44929fb7734819d1d9679/README.md?plain=1#L6-L7)
+ */
 export function isInviteLink(url: string) {
   return url.startsWith('https://i.delta.chat/') && url.includes('#')
 }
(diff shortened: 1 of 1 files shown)


nicodh (approved):

packages/shared/util.ts
```diff
@@ -6,6 +6,10 @@ export function truncateText(text: string, max_len: number) {
   }
 }
 
+/**
+ * @returns Whether the `url` is a contact or a group invite link.
+ * (https://github.com/deltachat/invite/blob/1b5b2e2a9bfcaf87c4f44929fb7734819d1d9679/README.md?plain=1#L6-L7)
```
Simon-Laux, line 11: maybe just like to the general page and additionally the specification for these links:

- https://github.com/deltachat/invite
- https://github.com/deltachat/interface/blob/main/uri-schemes.md#invite-links-httpsideltachat-

packages/shared/util.ts
```diff
@@ -6,6 +6,10 @@ export function truncateText(text: string, max_len: number) {
   }
 }
 
+/**
+ * @returns Whether the `url` is a contact or a group invite link.
```
Simon-Laux, line 10: wording confusing, first time I read it I thought it means that the function returns "group" or "contact"

```suggestion
 * @returns Whether the `url` is an invite link (for contact or a group).
```

packages/shared/util.ts
WofWca:
```suggestion
 * - https://github.com/deltachat/invite
 * - https://github.com/deltachat/interface/blob/main/uri-schemes.md#invite-links-httpsideltachat-
```

PR #4151, opened 2024-09-28
Improve invite link chatlist item

WofWca:
Recommended review commit-by-commit

image

image

Follow-up to https://github.com/deltachat/deltachat-desktop/pull/4041


nicodh (approved):
Looks good by reading and testing.

Would be nice to have that in new chat search field too!

PR #4154, opened 2024-09-29
fix: crash on "About" click if account unselected

WofWca:
E.g. after deleting an account.
Though the crash did not happen if you're in the process of creating an account, or on the initial "onboarding" screen, because that would make an "uninitialized" account and we'd be fine, so this fix is not super impactful.

This is similar to cee10103132a07add.

Yes, I also hate the approach I've taken.


nicodh (approved):
Looks good and I tested it successfully

Issue #4063, opened 2024-08-03
Accessibility: on systems with larger DPI scaler (e.g. tablet-like form factor or user with poor sight) start screen might be cut off


ell1e:
It seems like on systems with a larger DPI scaler given a relatively normal or smaller screen size, e.g. tablet-like form factor where people may often still run Desktop Linux, or on computers of user with poor sight, the delta chat start screen seems to be cut off. It would be nice if it was more responsive and scrollable to work around that.

- Operating System (Linux/Mac/Windows/iOS/Android): postmarketOS Edge x64 desktop (Alpine Linux based), using a regular KDE Desktop and not any kind of mobile desktop
- Delta Chat Version: v1.140.2 (latest on flathub)
- Expected behavior: start screen and about screen aren't cut off
- Actual behavior: it's cut off:
  ![Screenshot_20240803_144258](https://github.com/user-attachments/assets/b0d71599-3d67-43d2-ad0c-5fb4fd0a6220)
  The about screen can't fully scroll to the buttom either:
  ![Screenshot_20240803_144710](https://github.com/user-attachments/assets/1c5bef8a-6eaf-42a6-9582-a24b5442acd5)

- Steps to reproduce the problem:

  1. Set DPI scaler somewhat high on a normal sized or slightly smaller sized laptop or tablet screen. E.g. an ultraportable Ultrabook or something like the GPD Win or like the Steam deck or a normal laptop with poor eyesight user.
  2. Start delta chat
  3. Observe that it doesn't seem to fit the screen fully
- Screenshots: see above
- Logs:

Simon-Laux:
What's your window/screen resolution in pixels?

ell1e:
800x1280 with DPI scaler 1.75

WofWca:
I believe HTML/CSS-wise we're good. It appears like the window itself goes beyond the screen bounds in the original post.

https://github.com/user-attachments/assets/6a9521f9-0da8-485b-ab2a-06f884352f65

On Windows I cannot resize the window to shorter than 400px. I think we set the `minHeight` here:

https://github.com/deltachat/deltachat-desktop/blob/a5c02433976909fe01a64f3049c96d0d101ae9d8/packages/target-electron/src/application-constants.ts#L43-L44

Shall we just raise it? Do we have a reason to have it no lower than 450?

Simon-Laux:
[...]

PR #4156, opened 2024-10-01
fix: window overflowing small screens

WofWca:
...with high zoom level.

Closes https://github.com/deltachat/deltachat-desktop/issues/4063

I did not test this because for me `setMinimumSize` is not working. Though I played around with window sizes to determine the new numbers.


Simon-Laux:
zoom and system dpi are 2 different scaling factors. maybe we should rather just remove the min size? or we somehow find out dpi and then somehow incorporate it into the calculation. Also this affects the other windows too, if I'm not mistaken?

PR #4157, opened 2024-10-01
fix: not auto-scrolling to new messages

WofWca:
The bug was introduced recently, in https://github.com/deltachat/deltachat-desktop/commit/07cd2b2c5f9d779cb71bb8c034b31b503b348806.
That commit assumed that we do not have multiple elements
with the same id in the document.

No need for a CHANGELOG entry because there has not been a release.

### DIFF
--- a/packages/frontend/src/components/message/MessageList.tsx
+++ b/packages/frontend/src/components/message/MessageList.tsx
@@ -314,8 +314,9 @@ export default function MessageList({ accountId, chat, refComposer }: Props) {
           !domElement.classList.contains('info-message') &&
           !domElement.classList.contains('videochat-invitation'))
       ) {
-        log.debug(
-          `scrollTo: scrollToMessage, couldn't find matching message in dom: ${domElement}, returning`
+        log.error(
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/message/MessageList.tsx
```diff
@@ -308,16 +308,29 @@ export default function MessageList({ accountId, chat, refComposer }: Props) {
       log.debug('scrollTo: scrollToMessage: ' + scrollTo.msgId)
 
       const domElement = document.getElementById(scrollTo.msgId.toString())
```
Simon-Laux, line 310: maybe adding a prefix to the id attribute so it is not only a number might be a good idea?
Or we change it to a custom attribute like `x-msg-id` and look for it via `document.querySelector`, though I'm not sure what is faster.

We could also call `.querySelector` on the DOM element of the message list, so that it only searches relevant nodes and that makes `.querySelector` probably faster.
But I'm not sure if the speed difference in this case is even a big concern.

Simon-Laux (approved):

PR #4164, opened 2024-10-02
WIP: perf: avoid re-renders caused by `useEffect`

WofWca:
Having read https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes, I think we might be overusing `useEffect` to detect prop/state changes.
I'm not a huge React expert, but I would like to ask if these changes make sense.

### DIFF
--- a/packages/frontend/src/components/chat/ChatList.tsx
+++ b/packages/frontend/src/components/chat/ChatList.tsx
@@ -44,6 +44,7 @@ import type {
   MessageChatListItemData,
 } from './ChatListItemRow'
 import { isInviteLink } from '../../../../shared/util'
+import { useHasChanged2 } from '../../hooks/useHasChanged'
 
 const enum LoadStatus {
   FETCHING = 1,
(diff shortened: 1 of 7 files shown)


packages/frontend/src/components/chat/ChatList.tsx
```diff
@@ -629,17 +630,21 @@ function useLogicChatPart(
     useLogicVirtualChatList(chatListIds)
 
   // effects
-  useEffect(() => {
+  const queryStrHasChanged = useHasChanged2(queryStr)
+  // TODO why don't we just call this unconditionally?????
+  // `Object.is()` will take care of infinite loops.
```
Simon-Laux, line 635: I assume `Object.is()` is part of react's implementation of setState?

WofWca: That's right https://react.dev/reference/react/useState#setstate

> If the new value you provide is identical to the current state, as determined by an [Object.is](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) comparison, React will skip re-rendering the component and its children. This is an optimization. Although in some cases React may still need to call your component before skipping the children, it shouldn’t affect your code.

Simon-Laux: not completely clear from the comment alone, better mention that this is done by setState

WofWca: Sure, this is just a rough draft

WofWca:
Superseded by https://github.com/deltachat/deltachat-desktop/pull/4560

PR #4166, opened 2024-10-03
exit search when clicking on profile, when the selected profile is already the selected account

### LINKED ISSUE #3907: handle subsequent taps on profile icon
when tapping the profile icon of the already selected profile (so, not the first tap to switch the profile), the profile should go back to sort of "default state", which is partly done also when switching to another profile:

- ~~go out of profile (already done on profile switching)~~
- end search (already done on profile switching)
- ~~scroll up(already done on profile switching)~~

~~this makes sure, the user "sees" the unread chats when they notice the profile's badge counter.~~

and, it makes taps on the profle switcher predictable, the user learns to always end up in a known, standard view.

the selected ~~account~~ chat, however, should not be changed - if that means eg. that the archive needs to stay open, then it's that.

but if the archive is entered with no selection, subsequent taps should land in the default view.

cmp. https://github.com/deltachat/deltachat-ios/issues/2125 , https://github.com/deltachat/deltachat-ios/issues/2171 , comparable apps

this is a minor one, no urgency at all :)

Simon-Laux:
closes #3907

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@
 - crash on clicking "About" when no account is selected (e.g. after deleting an account) #4154
 - show "new group" instead of "new contact" when pasting a group invite link in the search field #4151
 - message input getting unexpectedly re-focused, and not re-focused after some actions if the draft text is not empty #4136
+- fix: exit search when clicking on profile when the selected profile is already the selected account #4166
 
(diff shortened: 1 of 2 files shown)


r10s (approved):
lgtm, as far as i can tell. did not test.

thanks a lot for taking care!

Issue #4167, opened 2024-10-03
RTL support


Simon-Laux:
Areas:
- [ ] Navbar
- [ ] macOS placeholder for traffic lights (window controls)
- [ ] Chatlist items
- [ ] Icons (directional icons such as back buttons on dialogs and checkmarks)
- [ ] Overlay position:
   - [ ] AccountListSidebar hover info (opens in wrong direction)
   - [ ] Emoji Picker
- [ ] Composer:
   - [ ] send button arrow icon needs to be flipped
   - [ ] text input field should be set to rtl
- [ ] MessageList
  - [ ] messages are on wrong side and sharp corner is on the wrong side -> also they don't fill screen

Useful resources:
- html `dir=<ltr|rtl>` attribute
- [https://developer.mozilla.org/en-US/docs/Web/CSS/:dir](https://github.com/deltachat/deltachat-desktop/issues/new#)

Questions:

- [ ] Do we enable it based on language or as user preference?
    - based on language is easy and doesn't introduce extra setting
    - extra setting allows testing with other languages 
- [ ] Should Icons be mirrored as well?
     - verified checkmarks and other checkmarks?
     - the send button should be mirrored
- [ ] Should we open context menus in the opposite direction? (this is about the preference, if there is not enough space there it will open on another side of the cursor)
     - I think it makes sense because you read in the other direction, but I don't know what other apps do in RTL mode

Misc:
- there seems to be no consistent API across browsers to find out writing direction (firefox has no api, safari and chrome have different api), so I would say we hardcode this in `_locales/_languages.json`?

CC @farooqkz

r10s:
>  Do we enable it based on language or as user preference?

if it all, it should be a debug-only-hidden setting, not even experimental. for the user, there is no reason to have this option - languages are RTL or LRT, not both. android/iOS/Windows/Mac/whatever usually do not have this option in UI accessible by the user.

so, i would not go for that at all - for testing, devs can easily just switch to an RTL language - we devs anyway know the position of buttons without reading :)

> Should Icons be mirrored as well?

makes sense esp. for the "back" button. for all other icons, that seems to be a minor issue, see https://m2.material.io/design/usability/bidirectionality.html#mirroring-layout - however, i would defer these icon discussions when all other parts are fine. is there just a CSS trick for mirroring? that would make the decision easier

> Should we open context menus in the opposite direction?

[...]

PR #4168, opened 2024-10-03
RTL Support

Simon-Laux:
Needs rebase after blueprint pr is merged. 

solves #4167

- **hardcoding writing direction and provide it in `I18nContext`**
- did some things, todolist is in #4167


ralphtheninja:
This is a great start! I think we should try and rebase it and try to get it into main asap. If not, it'll get harder and harder to merge it. This way we have the basic steps for handling a locale configured with RTL and some UI parts to look at.

nicodh:
I think this is almost working now.

Question is: do we need a switch to turn the rtl behaviour on or off. It seems not all people using rtl languages want this behaviour (according to some comments in the issue)

r10s:
> Question is: do we need a switch to turn the rtl behaviour on or off. It seems not all people using rtl languages want this behaviour (according to some comments in the issue)

some multilingual focused or pro tools may have a switch, i do not doubt that.

but it is rare in end-user-focused desktop apps as we are. i have rarely, if at all, seen such a switch in end-user-apps. the direction is chosen by the language itself, this is also what android/ios are doing. i would not spend additional resources and maintainancce  on things that are quite unusal and nerdy, and may also end up in addition support required.

however, it there are such options as "force ltr" and "force rtl" options available in the system or as an electron command line switch, one should consider to support it (maybe it just happens automatically), iirc, eg. on android you can do that and this is useful eg. for debugging.
but even that is a second step.

so, let's not add such a switch to our settings

EDIT: i had some one-to-one conversations with ppl affected, a switch is really not needed and not used elsewhere in comparable apps

farooqkz:
Thank you so much for this, Simon. I really appreciate it. I tried to build the Electron target from source and run it. I did `pnpm i && pnpm start` and it throwed and error at me:

```
0.2s [C]main/index: Fatal Error during init Error: Language not found in language list
    at loadTranslations (/home/farooqkz/playground/deltachat-desktop/packages/target-electron/src/load-translations.ts:47:11)
    at setLanguage (/home/farooqkz/playground/deltachat-desktop/packages/target-electron/src/load-translations.ts:36:22)
    at onReady (/home/farooqkz/playground/deltachat-desktop/packages/target-electron/src/index.ts:174:3) 
```

I don't have any environmental variable for locales set. e.g. `env | grep LC` prints nothing. `LANGUAGE` is empty and `LANG` is `C.UTF-8`. And `localectl status` reports that they are unset. Adding some logs, value of `DesktopSettings.state.locale` is undefined and `app.getLocale()` returns `en_US`. But `setLanguage` expects the ISO code without country code. Like `en` instead of `en_US`. I changed the code to `setLanguage("fa")` and it loads fine. I guess for this, I should open a new issue?

Anyway regarding the RTL support:

 - numbers should be Eastern not Western
 - Dates should be Jalali by default not Gregorian.
 - The plus button to create new chat/group should be on the left of the chat list not right
 - There should be option for people who want to use Gregorian calendar in Persian language as well as having LTR for Persian.

Edit: Moment doesn't seem to support Jalali calendar. But I'm wondering why do we need that anyway. I can do `date.toLocaleString("fa")` and it shows in Jalali calendar.

Simon-Laux:
@farooqkz you forgot to build: `pnpm dev` or `pnpm build && pnpm start`, so not sure if you ran the most current version. (but you may have build it already and just misspelled it here in your comment) also right now nico has taken over the pr and I personally don't know the exact state of it at the moment, better ask @nicodh

farooqkz:
> @farooqkz you forgot to build: `pnpm dev` or `pnpm build && pnpm start`, so not sure if you ran the most current version. (but you may have build it already and just misspelled it here in your comment) also right now nico has taken over the pr and I personally don't know the exact state of it at the moment, better ask @nicodh

I did `pnpm build && pnpm start`. Is that wrong?

nicodh:
> I did `pnpm build && pnpm start`. Is that wrong?

No. Btw running pnpm dev:electron from root is more convenient.

I will have a look at the issues soon

nicodh:
> numbers should be Eastern not Western

then we need at least all numbers from 0 to 9 in transifex (which we don't have yet)
> Dates should be Jalali by default not Gregorian.

What is the difference? I see ۸ اوت ۲۰۲۵ for August 8, 2025 (for example in an info message)
> The plus button to create new chat/group should be on the left of the chat list not right 

=> fixed

r10s:
for the calendar, this is no blocker. it is not changed by rtl and as good as before.

in another pr, there could be some options to tweak the date from locale, however, if moments.js does not support jalali, we should not spearhead, for same reasons as direction

to me the jalali sounds a bit weird https://en.m.wikipedia.org/wiki/Jalali_calendar  - is that the default signal/whatsapp/telegram are using?

farooqkz:
> for the calendar, this is no blocker. it is not changed by rtl and as good as before.
> 
> in another pr, there could be some options to tweak the date from locale, however, if moments.js does not support jalali, we should not spearhead, for same reasons as direction
> 
> to me the jalali sounds a bit weird https://en.m.wikipedia.org/wiki/Jalali_calendar - is that the default signal/whatsapp/telegram are using?

You call our calendar weird? I will ping other Persians to come and byte you!

This PR can be merged without Jalali calendar support. But Jalali calendar is definitely something Persian speaking people want. It's the calendar we use in Iran for everything. And in other Persian speaking countries, the same calendar is used.

Also, didn't we have some plans to remove moment.js in favor of something else? I think it was Simon's idea?

farooqkz:
> > numbers should be Eastern not Western
> 
> then we need at least all numbers from 0 to 9 in transifex (which we don't have yet)

Nope. You just need `number.toLocaleString("fa")`

r10s:
point taken, all calendars are weird.

but again, is jalali the default signal/whatsapp/telegram/other-larger-app are using for that region?

please accept that we're approaching many "boring" things from that angle - to give us headroom to be different elsewhere. ppl "want" a lot of things :)

but we already agreed it is no blocker, so let's merge the pr as is, you can create a feature proposal at https://support.delta.chat/c/features/6 then; it is in competition over resources with other things then

r10s (approved):
i cannot say much to the code, but the languages flagged as "rtl" matches the ones on iOS at a first glance, and the UI looks as expected

farooqkz:
There was a small problem which in navbar, English numbers where should instead of Persian ones. This has been fixed in my tiny commit.

farooqkz (approved):
Even though I always use stuff in English language, I really enjoyed seeing DeltaChat in Persian. I thank you guys on behalf of the Persian community!

PS: Did not review code.

nicodh:
After some discussion we decided to keep all numbers non localized, to be consistent with other clients and to avoid inconsistencies with numbers provided by core (which are not localized )

r10s:
> After some discussion we decided to keep all numbers non localized, to be consistent with other clients and to avoid inconsistencies with numbers provided by core (which are not localized )

to amend, after all, arabic numbers are pretty common in arabic areas as well. 

Delta Chat is not an app focused on a single language, but an international project. so we have to see what we can afford with our current resources. mixes come from used libraries, our core and whatnot. even inconsistencies in date formats, so os render arabic numbers some not. therefore, we decided to go the way to leave arabic numbers as is, also to avoid subsequent support and issues.

even apps with 100x more resources and money have a mix of numbers and date formats ...

PR #4169, opened 2024-10-03
trim whitepaces when reading from clipboard in qr code reader

### LINKED ISSUE #4161: "I already have a profile" > "Paste from Clipboard" only shows "Scanned QR Code text: ..." popup then does nothing
My apologies if this is somehow intended, but it does seem to be a bug: when I go to one Delta Chat instance and do "Settings" > "Add Second Device" and then use "Copy", then I go to the other device and somehow get the text into the clipboard there and do "I already have a profile" > "Paste from Clipboard" only shows "Scanned QR Code text: ..." popup then does nothing. The expected result would be that it actually allows me to somehow pair the devices at that point, or that it shows me why that's not possible.

- Delta Chat Version: Desktop 1.46.1 flathub 1719314892000
- Expected behavior: "Paste from Clipboard" on the 2nd device of the 1st's device pairing info will somehow proceed the device pairing or tell me what the issue is, if I did something wrong
- Actual behavior: "Paste from Clipboard" on the 2nd device will show some "Scanned QR Code text: ..." popup and then when I press "OK" I'm just back at the profile creation screen with nothing new having happened.
- Steps to reproduce the problem: See in description above.
- Screenshots:
Screenshot_20241001_232222

- Logs: 
[2024-10-01-23-06-14.log](https://github.com/user-attachments/files/17216635/2024-10-01-23-06-14.log)

Simon-Laux:
closes #4161

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - improved performance a bit #4145
 - show contact / group name when pasting invite link in the search field #4151
 - Update local help (2024-10-02) #4165
+- trim whitepaces when reading from clipboard in qr code reader #4169
 
 ### Fixed
 - fix that you can not click header button in dialog when they are on top of the navbar #4093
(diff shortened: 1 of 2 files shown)


WofWca (approved):

ell1e:
Thanks so much for the quick fix! :tada:

PR #4170, opened 2024-10-03
fix running with local core

Simon-Laux:


nicodh (approved):
Tested on linux

maxphilippov:
Sent it to Simon first, gonna duplicate here, cause I'm not sure if it's the problem with Desktop or my core build went wrong
image
Here's a log of my running `pnpm dev`. I'm not sure if underlined path is correct and who's to blame: desktop or core here.

Simon-Laux:
I'll merge this now, we can fix windows problem separately. Path looks correct, so I guess the issue is in the python script that builds the local dev version in the core repo.

PR #4173, opened 2024-10-04
perf: cache chat list & don't delay initial load

WofWca:
This has the following effects:
- Upon the initial load of a component that displays a list of chats,
    we start loading them instantly instead of waiting for the
    debounce timeout of 200ms.
- When emptying the query string for a chat list, all chats get
    displayed instantly, instead of getting re-fetched.

The following components are affected:
- The chat list on the main screen
- "Forward message to..."
- Mailto dialog
- View profile: chats in common
- webxdc sendToChat dialog

Please assess whether this is too ugly. Basically what we're doing is caching + removing the debounce. Perhaps a different approach would be more elegant. But IMO this isn't so bad, maybe not even half bad.

### DIFF
--- a/packages/frontend/src/components/chat/ChatListHelpers.tsx
+++ b/packages/frontend/src/components/chat/ChatListHelpers.tsx
@@ -53,6 +53,11 @@ export function useMessageResults(
   return ids
 }
 
+/**
+ * The resulting `chatListEntries` is cached for the initial query parameters
+ * (i.e. with default initial parameters that would be all chats).
+ * The said cache is also updated on `ChatlistChanged` event.
+ */
 export function useChatList(
   initialListFlags: number | null = null,
   initialQueryStr?: string,
@@ -68,6 +73,17 @@ export function useChatList(
   const [queryStr, setQueryStr] = useState<string | undefined>(initialQueryStr)
   const [queryContactId, setQueryContactId] = useState(initialQueryContactId)
   const [chatListEntries, setChatListEntries] = useState<number[]>([])
+
+  const areQueryParamsInitial: boolean =
+    initialListFlags === listFlags &&
+    initialQueryStr === queryStr &&
+    initialQueryContactId === queryContactId
+  const chatListEntriesForInitialQueryParams = useChatListNoDebounce(
+    accountId,
+    initialListFlags,
+    initialQueryStr,
+    initialQueryContactId
+  )
 
   const debouncedGetChatListEntries = useMemo(
     () =>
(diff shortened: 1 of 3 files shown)


Simon-Laux:
There is not debounce in core of the chatlist update event, so there needs to be some debounce in the ui.
the debounce function we have has a setting to fire before the timeout instead of after the timeout. Alternatively I made a `debounceWithInit`-helper function that is debounce, but the first invocation runs immediately.

WofWca:
Hmm. But when can that happen? Do we expect the `ChatlistChanged` event to fire frequently?

Simon-Laux:
>  But when can that happen? Do we expect the ChatlistChanged event to fire frequently?

on loading many messages after being offline.

packages/frontend/src/components/chat/ChatListHelpers.tsx
```diff
@@ -121,3 +165,53 @@ export function useChatList(
...
+ */
+function useChatListNoDebounce(
+  accountId: number,
+  listFlags: number | null = null,
+  queryStr: string = '',
+  queryContactId?: number
+) {
+  const [chatListEntries, setChatListEntries] = useState<number[]>([])
+
+  // Though perhaps a throttle would be more appropriate.
```
Simon-Laux, line 181: as long is it is called at the falling edge (so in the end it is always updated) anything should be fine that reduces requests.

WofWca: Well, if we get this event at a sustained rate for a prolonged period of time, we'd only load the chat list at the end. It'd be better to load the chat list at least every second in such a case.

Simon-Laux: good point. we need both:
- don't call it too often / overlapping
- call it latest every second
- call it at the end to make sure the correct state is shown in the end.

Maybe some code that waits for the request to complete (resolve or reject) before calling it again? But if core should not answer for any reason that could hang, so maybe just go for a classic time based throttle?

WofWca: Yes, I've been looking at [`async-throttle`](https://www.npmjs.com/package/@jcoreio/async-throttle), but it depends on Babel runtime. I didn't want to add bloat, or fork the package just yet as it will take some work.

But, prior to this MR, we already use `debounce`, so merging this MR wouldn't be a regression in this regard.
I'd suggest to tackle this later, because there are probably a few more places in the codebase where `throttle` makes more sense than `debounce`.

Simon-Laux: >  fork the package just yet as it will take some work.

Looks easy enough (looking at the uncompiled `@jcoreio/async-throttle/index.js.flow` we'd only need to copy that file, add a license/attribution comment and convert it to TS. No need to fork the whole npm package.

Simon-Laux (approved):

packages/frontend/src/components/chat/ChatListHelpers.tsx
WofWca: I tried, but it's not JS. It's something called "flow" XD

PR #4176, opened 2024-10-05
style: hide reactions instead of wrapping

WofWca:
on narrower screens.


Simon-Laux:
I think `SHOW_MAX_DIFFERENT_EMOJIS` should be dynamic -> if there is not enough space for all emojis show only the ones that fit and the more button.

adbenitez:
yes, so user can realize there are more reactions similar to what happens in android that you get a (+N) pill with the number of other reactions that are not displayed

nicodh:
resolves https://github.com/deltachat/deltachat-desktop/issues/4322

WofWca:
Closing this since if we want to show this "more" button, it would probably require JavaScript, but this MR mostly does CSS, i.e. it's not the right approach.

PR #4177, opened 2024-10-05
style: lower contrast for a border in dark theme

WofWca:
When quoting a message.
In the default dark theme was way too bright.

Before:

image

After:

image


r10s (approved):
very good! thanks a lot!

that UI style bug also hit my eyes, but i did not came with an issue, let alone fix yet :)
i also did a quick test, works as expected :)

Issue #4158, opened 2024-10-01
[UX] Entering some sort of join link or room id into the search after pressing the "+" button doesn't offer to let me join


ell1e:
I found a chat room that I was asked to join, and it's specified via some `OPENPGP4FPR:...` id. Alternatively, I was given a link starting with `https://i.delta.chat/#...`.

Neither will open in the regular desktop chat client if I enter them in the two obvious locations:

![Screenshot_20241001_184143](https://github.com/user-attachments/assets/8e56b1a5-7d58-4323-8e9d-da56f97d7f99)

![Screenshot_20241001_184027](https://github.com/user-attachments/assets/2f2d1c44-488e-4ec4-a20b-6918af86252c)

This makes it unnecessarily confusing to figure out how to even join a room without a mobile phone. I suggest that this is changed.

- Operating System (Linux/Mac/Windows/iOS/Android):
- Delta Chat Version: Desktop 1.46.8 Core v1.142.12 SQLite 3.45.3 electron 30.3.1 node 20.15.1
- Expected behavior: entering both this room id and the given join link into either of the two search fields works
- Actual behavior: neither the room id nor the join link seems to be recognized by any of the search fields
- Steps to reproduce the problem: try to enter a room id or join link into the input fields marked above
- Screenshots: see above
- Logs: probably not relevant!

Simon-Laux:
great idea!

For now you can also send the link in self send/saved messages and click on it.

WofWca:
I thought I added this to the chat list search in https://github.com/deltachat/deltachat-desktop/pull/4041, which was released in 1.46.2...

There was another small improvement #4151, which is not released yet.
And @nicodh suggested to add this to the contact search as well: https://github.com/deltachat/deltachat-desktop/pull/4151#pullrequestreview-2336857429

r10s:
> I thought I added this to the chat list search in https://github.com/deltachat/deltachat-desktop/pull/4041, which was released in 1.46.2...

yeah, i can confirm, that it works in general.

but it seems too implicit, to to show the generic "New Contact" button that is shown at other places as well, where the user has to enter an email address.

as a quick improvemnt, the button could read "Chat with CHATNAME".

WofWca:
An even better improvement would be to read the clipboard, detect that it's an invite link, and suggest appropriate action.

ell1e:
[...]

PR #4178, opened 2024-10-05
improvement: better search for invite link

WofWca:
Before:

image image

After:

image image

Partially addresses https://github.com/deltachat/deltachat-desktop/issues/4158. What's left I guess is adding this to the "plus button" dialog.

### DIFF
--- a/packages/frontend/src/components/helpers/PseudoListItem.tsx
+++ b/packages/frontend/src/components/helpers/PseudoListItem.tsx
@@ -182,12 +182,12 @@ export const PseudoListItemAddContactOrGroupFromInviteLink = ({
       id='newcontactorgroupfrominvitelink'
       cutoff='+'
       text={
+        parsedQr?.kind === 'askVerifyGroup' ? parsedQr.grpname : contactName
+      }
+      subText={
         parsedQr?.kind === 'askVerifyGroup'
           ? tx('menu_new_group')
           : tx('menu_new_contact')
-      }
-      subText={
-        parsedQr?.kind === 'askVerifyGroup' ? parsedQr.grpname : contactName
       }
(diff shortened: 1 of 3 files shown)


Simon-Laux:
I'd prefer if it said "join group" instead of "new group"

WofWca:
image

packages/frontend/src/components/helpers/PseudoListItem.tsx
```diff
@@ -177,19 +176,35 @@ export const PseudoListItemAddContactOrGroupFromInviteLink = ({
...
-      onClick={() => processQr(accountId, inviteLinkTrimmed)}
+      text={parsedQr.grpname}
+      subText={tx('join_group')}
+      onClick={onClick}
+    />
+  ) : (
+    // There are many kinds of QRs. For example, if you paste an invite link
+    // of a group that you created yourself, you'll get
+    // `parsedQr.kind === "withdrawVerifyGroup"`.
+    // TODO we probably want to better handle such cases.
```
Simon-Laux, line 201: do we want to handle all qrlinks in search fields now?

WofWca: You mean in this MR? I think not. I think we covered the two most common cases. The rest can be improved later. Anyways it's just a matter of proper labeling. The dialog that opens when you click the chatlist item will tell the user more.

Simon-Laux: might not be a bad idea to handle other qr code links as well there, as this is where some users seem to paste it to "scan" it. But thats for a future PR, I just asked because of the comment.

Simon-Laux (approved):

PR #4183, opened 2024-10-06
fix: `jumpToMessage` not highlighting messages

WofWca:
The bug was introduced in 07cd2b2c5f9d779cb71bb8c034b31b503b348806.


r10s (commented):
works as expected, cannot say much to the code and the reasoning why `LI` is handled differently

Simon-Laux (approved):
this looks like a workaround for a deeper issue, but as long as it works

WofWca:
You are correct, we should not have several elements with the same `id`. But I didn't want to dredge it all up.

PR #4189, opened 2024-10-08
blocked users in chatlist menu

maxphilippov:
There's an issue #4046, short summary:

1. Block contact
2. Find a group chat with this contact
3. Open blocked contact profile
4. Press 'Send Message'
5. Now you have a chat with blocked contact in `ChatList` (on the left), but when you open a context menu on this chat there's only "Block Contact"

Same scenario works for the ThreeDotMenu in chat view.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 
 ## Fixed
 - image attachments not being centered within a message #4313
+- fix blocked status of a contact in context menu for ChatListItem and ThreeDotMenu #4046
 
 
 <a id="1_47_1"></a>
--- a/packages/frontend/src/components/ThreeDotMenu.tsx
+++ b/packages/frontend/src/components/ThreeDotMenu.tsx
@@ -1,21 +1,26 @@
 import { C } from '@deltachat/jsonrpc-client'
-import React, { useContext } from 'react'
+import React, { useContext, useState, useEffect } from 'react'
+import debounce from 'debounce'
 
 import { Timespans } from '../../../shared/constants'
 import { ContextMenuItem } from './ContextMenu'
 import SettingsStoreInstance, { useSettingsStore } from '../stores/settings'
-import { BackendRemote } from '../backend-com'
+import { BackendRemote, onDCEvent } from '../backend-com'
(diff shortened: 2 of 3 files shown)


r10s:
"Block Contact" removes the one-to-one-chat on purpose. if you now re-create the one-to-one-chat, using "Send Message" and also send a messages, it seems a bit unexpected, that the contact is still blocked and you cannot see eg. the answer (imagine you ask the user to be friends again :)

not sure, if this is a desktop-only-issue, however. maybe we can also unblock in core, in case a private message is sent to a user

r10s:
> the chat's menu is "Block Contact" and the profile menu is "Unblock Content". this is somehow wrong.

per request, here is a video. this is what it looks on current master for me:


https://github.com/user-attachments/assets/6391d481-af99-48fc-ab13-b774d8c28df2

nicodh:
Hmm, I thought that was solved with https://github.com/deltachat/deltachat-desktop/pull/4147 where we had the same contradictory menu items... ?

maxphilippov:
> Hmm, I thought that was solved with #4147 where we had the same contradictory menu items... ?

No, I've mentioned that this might require more work and it was a part of another issue, so it's another PR.
https://github.com/deltachat/deltachat-desktop/pull/4147#issuecomment-2377590498

nicodh:
I just tested this again. Is it correct that you only solved the wrong menu item in the chatlist context menu but not in the 3 dot menu on the chat with the blocked user?

So then I would name this PR "Step one of two"

maxphilippov:
> I just tested this again. Is it correct that you only solved the wrong menu item in the chatlist context menu but not in the 3 dot menu on the chat with the blocked user?
> 
> So then I would name this PR "Step one of two" or at least create a scond issue for the chatlist menu after closing #4046

Sorry, missed that one, three dot was fixed, didn't that work for you? Should we close this now since there's no way to open both for blocked contacts?

nicodh:
As #4353 goes for not showing any 1:1 chat with a blocked user there is no need to adapt all meus to that case anymore imho
So I suggest to close this pr

PR #4192, opened 2024-10-08
Fix Settings/AllMedia in Account List context menu

### LINKED ISSUE #4190: crash, when trying to tap "Settings" of another profile
when right-tapping the profile image of a _not selected profile_, and then select "Settings", the app crashes.

https://github.com/user-attachments/assets/28cebcf0-1701-4ecc-947b-2e3b21b6185d

tested on 1.46.8 and on current master on macOS 14.5, [log](https://gist.github.com/r10s/a04c2e3085b64fb9d93803adce3f8d5e)

### LINKED ISSUE #4191: "All Media" does not always switch to "All Media"
when right-tapping the profile image of a _not selected profile_, and then select "All Media", the app does not show "All Media".

instead, only the normal profile switch seems to be done.

tested on 1.46.8 and on current master on macOS 14.5

maxphilippov:
- combine all settings-gathering when changing account since some functions await for SelectAccount
- give additional 50ms to switching to Gallery since there's a conflict of setState and async media query in Gallery.tsx
- resolves #4190
- resolves #4191

### DIFF
--- a/packages/frontend/src/ScreenController.tsx
+++ b/packages/frontend/src/ScreenController.tsx
@@ -121,6 +121,9 @@ export default class ScreenController extends Component {
       await this.unSelectAccount()
       this.selectedAccountId = accountId
       ;(window.__selectedAccountId as number) = accountId
+      // forcing to load settings here so when we for example switch to Settings
+      // from context menu they're already present and we avoid crashing
+      SettingsStoreInstance.effect.load()
     } else {
       log.info('account is already selected')
       // do not return here as this can be the state transition between unconfigured to configured
@@ -138,13 +141,15 @@ export default class ScreenController extends Component {
     if (!dontStartIo) {
       await BackendRemote.rpc.startIo(accountId)
     }
-    runtime.setDesktopSetting('lastAccount', accountId)
-    BackendRemote.rpc.getSystemInfo().then(info => {
-      log.info('system_info', info)
-    })
+
     BackendRemote.rpc.getInfo(accountId).then(info => {
       log.info('account_info', info)
     })
+    BackendRemote.rpc.getSystemInfo().then(info => {
+      log.info('system_info', info)
+    })
+
+    await runtime.setDesktopSetting('lastAccount', accountId)
   }
 
   async unSelectAccount() {
--- a/packages/frontend/src/components/AccountListSidebar/AccountItem.tsx
+++ b/packages/frontend/src/components/AccountListSidebar/AccountItem.tsx
@@ -100,7 +100,8 @@ export default function AccountItem({
         setTimeout(() => {
           ActionEmitter.emitAction(KeybindAction.GlobalGallery_Open)
           // NOTE(maxph): Gallery.tsx gets unmounted before receiving media data
-          // and breaks markdown, so here 50ms is a temprorary workaround for that
+          // and only partially updates chat header without changing chat view to Gallery, 
+          // so here 50ms is a temprorary workaround for that
         }, 50)
       },
     },
--- a/packages/frontend/src/components/AccountListSidebar/AccountItem.tsx
+++ b/packages/frontend/src/components/AccountListSidebar/AccountItem.tsx
@@ -100,7 +100,7 @@ export default function AccountItem({
         setTimeout(() => {
           ActionEmitter.emitAction(KeybindAction.GlobalGallery_Open)
           // NOTE(maxph): Gallery.tsx gets unmounted before receiving media data
-          // and only partially updates chat header without changing chat view to Gallery, 
+          // and only partially updates chat header without changing chat view to Gallery,
           // so here 50ms is a temprorary workaround for that
         }, 50)
       },
--- a/packages/frontend/src/components/AccountListSidebar/AccountItem.tsx
+++ b/packages/frontend/src/components/AccountListSidebar/AccountItem.tsx
@@ -99,7 +99,9 @@ export default function AccountItem({
         // set Timeout forces it to be run after react update
         setTimeout(() => {
           ActionEmitter.emitAction(KeybindAction.GlobalGallery_Open)
-        }, 0)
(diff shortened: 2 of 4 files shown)


maxphilippov:
A hacky solution for "All Media" from what I saw the rendering broke sometimes because of setting state after Gallery component unmounts. So I've temporarily added a slight delay to setTimeout

https://github.com/user-attachments/assets/b2b1e097-7674-4d37-9d31-a432b5f1910f

r10s (approved):
works as expected, and fixes the crash as well as the unexpected behaviour.

for the workaround: yes, would be good to have that without a hack, but even now, well, it fixes the issues at hand :)

packages/frontend/src/ScreenController.tsx
```diff
@@ -138,13 +138,15 @@ export default class ScreenController extends Component {
...
-    })
+    await Promise.all([
+      runtime.setDesktopSetting('lastAccount', accountId),
+      BackendRemote.rpc.getSystemInfo().then(info => {
+        log.info('system_info', info)
+      }),
+      BackendRemote.rpc.getInfo(accountId).then(info => {
+        log.info('account_info', info)
+      }),
+    ])
```
WofWca, line 149:
```suggestion
    await runtime.setDesktopSetting('lastAccount', accountId)
    BackendRemote.rpc.getSystemInfo().then(info => {
      log.info('system_info', info)
    })
    BackendRemote.rpc.getInfo(accountId).then(info => {
      log.info('account_info', info)
    })
```

The latter two are purely for debugging, there is no need to `await` them.

maxphilippov: It's weird, but if I take any of these out of Promise.all it's easier for me to reproduce settingsStore()[0] being null which is the reason for app to crash. It might be actually something else, and `await`-ing these 2 buys time for settingsStore to set up. So maybe we also need to have an increased setTimeout for Settings the same way I did for "All Media".

To be fair with my changes I got a couple of crashes too if I go crazy

maxphilippov: Looks like I've found a solution, gonna push soon

maxphilippov:
I don't have a lot of experience with SettingsStore but forcing .load() right after we change selected profile on window makes Settings crash irreproducible for me

packages/frontend/src/ScreenController.tsx
```diff
@@ -121,6 +121,9 @@ export default class ScreenController extends Component {
       await this.unSelectAccount()
       this.selectedAccountId = accountId
       ;(window.__selectedAccountId as number) = accountId
+      // forcing to load settings here so when we for example switch to Settings
```
WofWca, line 124: Shouldn't we remove [`SettingsStoreInstance.effect.load()` in `MainScreen.tsx`](https://github.com/deltachat/deltachat-desktop/blob/42ab00b45df1b0ebd9e95b08dd99e49e660f10f0/packages/frontend/src/components/screens/MainScreen.tsx#L133-L136) then?

And shouldn't we `await` it, if this is the way we want to go?

packages/frontend/src/components/AccountListSidebar/AccountItem.tsx
```diff
@@ -99,7 +99,9 @@ export default function AccountItem({
         // set Timeout forces it to be run after react update
         setTimeout(() => {
           ActionEmitter.emitAction(KeybindAction.GlobalGallery_Open)
-        }, 0)
+          // NOTE(maxph): Gallery.tsx gets unmounted before receiving media data
+          // and breaks markdown, so here 50ms is a temprorary workaround for that
```
WofWca, line 103: What does "breaks markdown" mean?

r10s (approved):
upon request, i did test another time - and it is still working for me, thanks a lot!

i cannot say much about the implementation details and the discussions, though - only that it is working and no longer crashing for me :)

packages/frontend/src/components/AccountListSidebar/AccountItem.tsx
maxphilippov: I don't have it on hand to reproduce but here's a screenshot from the video I posted above
image
It essentially inserts "All Media" in the chat header, effectively moving the chat title (with avatar to the middle of the header) and doesn't switch chat body to Gallery.

packages/frontend/src/ScreenController.tsx
maxphilippov: I don't see the point in asyncing except for future-proofing which I don't think is a good idea, since everything there inside is `await`-ed already so we go back to ScreenController only after all the async things are done.

I could test the MainScreen change I just expected that we might not call ChangeAccount on start or there's some other case where this is covered.

Simon-Laux:
@meganoahj just tested it and confirmed that it works now.

Simon-Laux:
Changelog entry is missing

packages/frontend/src/components/AccountListSidebar/AccountItem.tsx
r10s: so, it "breaks HTML" or "breaks Layout", it does not  "break markdown", which is not used here at all afaik 😅 

it is good to know what got broken, to be able to fix it 😅

maxphilippov: > so, it "breaks HTML" or "breaks Layout". it does not "break markdown", which is not used here at all afaik 😅
> 
> it is good to know what got broken, to be able to fix it 😅

Oh, sorry, tunnel vision. I meant markup, I thought HTML would be more confusing.

packages/frontend/src/ScreenController.tsx
WofWca: > except for future-proofing

IDK if it's just about future-proofind. Maybe there is another undiscovered bug already that is related to this.

Thought it's best to tackle it while we're at it. But fine. An improvement is an improvement.

Issue #4124, opened 2024-09-12
Improve proxy settings UI


link2xt:
With core release 1.143.0 old `socks5_*` settings are deprecated and `proxy_*` should be used instead.

This should be done when done is updated from 1.142.x.

https://github.com/deltachat/deltachat-android/pull/3292 gives an idea how the screen looks like in Android

Step 1
- [x] Change Setting Types:  socks5_enabled, socks5_host, socks5_port, socks5_user, socks5_password => proxy_enabled, proxy_url
- [x] Adapt form: One text field for one proxy_url (placeholder: "Enter proxy URL here"
- [x] UI: Show explaining text: "Supported proxy types: HTTP(S), SOCKS5 and Shadowsocks"

Step 2 (in separate PR)
- [x] UI: add client side validation
- [x] UI: Show connectivity status & proxy type
- [x] UI: enable multiple proxy url entries
- [x]  if a proxy is clicked/selected/added, enable "use proxy" toggle

Step 3  (in separate PR)
-  [x] Proxy UI should also be moved out of configuration and into a separate screen like done on Android: https://github.com/deltachat/deltachat-android/pull/3277

![new-proxy-settings](https://github.com/user-attachments/assets/b7585feb-5595-4930-ba8d-16b93c206ba9)

Sharing is probably not necessary on desktop

nicodh:
@link2xt : do we not support username & password for proxies anymore or can it be entered in the form username:password@some.host:8080

Simon-Laux:
isn't there also a new qr code type desktop should support?

https://github.com/deltachat/interface/pull/75
https://github.com/deltachat/deltachat-core-rust/pull/5895

r10s:
yip, there is DC_QR_PROXY, which should be passed to dc_set_config_from_qr() after confirmation, using `proxy_use_proxy_confirm`.

it supports not only SOCKS5 but also other proxies meanwhile, iirc

WofWca:
The majority (or all?) of this has been addressed in https://github.com/deltachat/deltachat-desktop/pull/5052.

>  UI: add client side validation

Is this enough?

https://github.com/deltachat/deltachat-desktop/blob/b9d42fa8532e50b1110901802641d3f71388396a/packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx#L321-L331

> Proxy UI should also be moved out of configuration and into a separate screen like done on Android:

I'm not sure what's the implication? Please see if what we have now is not enough.

[...]

PR #4194, opened 2024-10-09
Update core to 1.147

nicodh:


iequidoo (approved):

nicodh:
Seems we had to adapt settings first. See https://github.com/deltachat/deltachat-desktop/issues/4124

nicodh:
Step 1 of #4124 is included here now:

image

r10s (commented):
i did a quick functionality check and could manage to connect via `ss://YWVzLTEyOC1nY206c2hhZG93c29ja3M=@156.146.62.163:443#Switzerland%20#1270%20/%20OutlineKeys.com` - nice!

for the UI: as this is subject to change anyways and will do to a dedicated dialog, it is good enough for now.

for the strings:  they are available already translated, no need to add them to "untranslated".
for the "placeholder", i'd strike that for now, that seems more confusing and cluttering than helpful - ppl will have a proxy URL from some sources and needs and will paste them here - and not the other way round. also, already now it is incomplete as `ss` is missing.
if we think, a placeholder is  useful, let's start with that on android, where we move ahead with proxy settings and collect more experiences.


i cannot say much to the rest of the code

nicodh:
> for the strings: they are available already translated, no need to add them to "untranslated".

where can I check what is already translated?
Do I need to become member in the transifex DeltaChat project?

nicodh:
> for the "placeholder", i'd strike that for now, that seems more confusing and cluttering than helpful -

I agree  - at least as long the placeholder has the same styling than the entered text

r10s:
note that there is 1.147.1 meanwhile, 1.147.0 contained a bug that resulted in some incompatibilities with existing releases, see https://github.com/deltachat/deltachat-core-rust/issues/6038

packages/frontend/src/components/LoginForm.tsx
```diff
@@ -311,7 +311,12 @@ export default function LoginForm({ credentials, setCredentials }: LoginProps) {
             />
             {proxy_enabled === '1' && (
               <>
-                <p className='text'>{tx('login_proxy_experimental_warning')}</p>
+                <p className='text'>
+                  Proxy support is currently experimental. Please use at your
+                  own risk. If you type in an address in the e-mail field, there
+                  there will be DNS lookup that won't get tunneled through
```
iequidoo, line 317: ~~there~~ will be **a** DNS lookup that won't get tunneled through

PR #4195, opened 2024-10-09
chore: ci: fix preview download link missing

WofWca:
When an upload to our server fails.
Currently upload keeps failing for Mac and Windows for all MRs.
But we still upload to GitHub artifacts in such a case,
so let's not miss out on it and still show the download link.

TODO:
- [x] Let me see if this works.

### DIFF
--- a/bin/github-actions/postLinksToDetails.js
+++ b/bin/github-actions/postLinksToDetails.js
@@ -12,18 +12,27 @@ const GITHUB_API_URL =
 const prId = process.env['PR_ID']
 const GITHUB_TOKEN = process.env['GITHUB_TOKEN']
 
+/** May be absent */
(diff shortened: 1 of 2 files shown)


bin/github-actions/postLinksToDetails.js
```diff
@@ -12,18 +12,21 @@ const GITHUB_API_URL =
 const prId = process.env['PR_ID']
 const GITHUB_TOKEN = process.env['GITHUB_TOKEN']
 
+/** May be absent */
+const FULL_URL = process.env['FULL_URL']
```
Simon-Laux, line 16: rename it to something more descriptive, like artifact-url

Simon-Laux (approved):

PR #4204, opened 2024-10-10
improvement: improve jumpToMessage scrolling

WofWca:
Don't scroll the message list if the message is already visible.


nicodh (approved):
Reviewed by testing

PR #4205, opened 2024-10-10
Escape works as "go back" in settings

maxphilippov:
Did this as a test, cause it seemed like an easy refactor. And it was

Works, Notifications do look kinda weird.
image

There's an explicit "close settings" call when we click on "Add Second Device". @r10s @nicodh  Should that remain or no? Here's a video of how it works

https://github.com/user-attachments/assets/024c2980-089f-43ae-a1d7-8e0cf8a30c65

Depends on #4006 so I'm not asking for review right now, too many changes displayed here

### DIFF
--- a/packages/frontend/src/components/Dialog/BackButton.tsx
+++ b/packages/frontend/src/components/Dialog/BackButton.tsx
@@ -14,6 +14,8 @@ export default function BackButton(
 
   return (
     <HeaderButton
+      value='cancel'
+      formMethod='dialog'
       aria-label={tx('back')}
       icon='arrow-left'
       iconSize={24}
--- a/packages/frontend/src/components/Dialog/Dialog.tsx
+++ b/packages/frontend/src/components/Dialog/Dialog.tsx
@@ -50,45 +50,44 @@ const Dialog = React.memo<Props>(
             rect.left <= ev.clientX &&
             ev.clientX <= rect.left + rect.width
           if (!isInDialog) {
-            const cancelEvent = new Event('cancel')
-            dialog.current.dispatchEvent(cancelEvent)
-            // cancel event doesn't trigger dialog close
-            dialog.current.close()
+            dialog.current.close('cancel')
           }
         }
       : () => {}
 
-    const onClose = (value: any) => {
-      props.onClose && props.onClose(value)
-      dialog.current!.style.display = 'none'
+    const onClose = (_: any) => {
+      props.onClose && props.onClose(dialog.current?.returnValue)
     }
 
-    const onCancel = (ev: React.BaseSyntheticEvent) => {
+    const onEscapePress = (ev: React.BaseSyntheticEvent) => {
+      // by default this fires a 'close' event with undefined as value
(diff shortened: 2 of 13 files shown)


r10s:
works nicely! also the stacking is much more what one would expect.

the PR needs a rebase or so, cannot say anything to the code by the thousands of changed lines :)

> There's an explicit "close settings" call when we click on "Add Second Device". @r10s @nicodh Should that remain or no? Here's a video of how it works

if possible, it would nice if cancelling the initial "Add Second Device" alert is also shown atop of the settings dialog and canceling it just gets you back to settings. this would back trying out things and "clicking around, discovering what is possible" more predictable.

once the "Add Second Device" is started via "Continue", however, the settings dialog should disapepar so that one is full into the context of the "Add Second Device".

but that is no biggie - if it is hard to do, it is also fine as it is now.

r10s:
as i was asked: for the "X" button, indeed it would be a bit weird if it closes all dialogs if they are visually stacked. this is not what is done usually in this case.

so either:
- for the stacking, the "back" button is superfluous, remove it and use only "X"
- do not stack visually, let "back" button close one level, "X" the whole dialog. "Escape" acts as the "back" button

nicodh (approved):
Looks good to me.

Maybe the remaining focus border after closing a child dialog could also be removed?

image

nicodh:
> Works, Notifications do look kinda weird.

what do you mean?

maxphilippov:
> > Works, Notifications do look kinda weird.
> 
> what do you mean?

Just not used to it I guess, the fact that I see Settings in the background and Notifications are much smaller

maxphilippov:
> if possible, it would nice if cancelling the initial "Add Second Device" alert is also shown atop of the settings dialog and canceling it just gets you back to settings.

Rebased, added that. But it doesn't close when we press Continue, doesn't bother me, if it's not critical I'd merge. I'm looking into the whole "close everything in stack when we confirm some action" question as part of #2268

Simon-Laux:
similar issue exists on pressing escape when making a new profile, there it should never close the dialog completely

Simon-Laux:
> as i was asked: for the "X" button, indeed it would be a bit weird if it closes all dialogs if they are visually stacked. this is not what is done usually in this case.
> 
> so either:
> 
> * for the stacking, the "back" button is superfluous, remove it and use only "X"
> * do not stack visually, let "back" button close one level, "X" the whole dialog. "Escape" acts as the "back" button
> 
> i would go with the second option, to not add additional noise to UI, also it is what user already learned in Delta Chat :)

I would like to have this question from r10s addressed before we merge it. I would personally also prefer the second option.

WofWca:
To note, this pattern is also used in `ViewGroup`:

https://github.com/deltachat/deltachat-desktop/blob/a5c02433976909fe01a64f3049c96d0d101ae9d8/packages/frontend/src/components/dialogs/ViewGroup.tsx#L284-L290

nicodh:
Closed in favour of https://github.com/deltachat/deltachat-desktop/pull/4702

PR #4210, opened 2024-10-12
improvement: improve a11y

WofWca:
Most of the commits are independent, thus they can be reviewed and merged / cherry-picked separately. Let me know which particular commits you reviewed, or up to which point, and I'll make a separate MR to merge only those.

I have tried to make sure to not introduce UI / style regressions.

Things that need more testing:
- [ ] Contact lists (80604711c5b0ee339a31f37e02ce0d29e531b2f1) We have a lot of places where they are used, and I might have introduced a regression somewhere, though I tested quite a lot of places.

Issue #2784, opened 2022-05-28
Keyboard-only navigation does not work in Contact list, Settings dialog, shortcut cheatsheet dialog, etc.


dumblob:
- Operating System (Linux/Mac/Windows/iOS/Android): Linux
- Delta Chat Version: 1.29.1 (compiled from git 14743055 )

- Expected behavior: keyboard-only navigation works in all dialogs, windows, etc.

- Actual behavior: E.g. in the Contact list, Settings dialog, shortcut cheatsheet dialog, etc. scrolling works only with mouse wheel but not with page up/down nor with arrow up/down. E.g. pressing Tab does not move focus to another user-interactable item/widget in the app so it is impossible to control Delta Chat Desktop without any pointing device and second it is impossible to learn a specific quick workflow using only keyboard.

- Steps to reproduce the problem: Open DC Desktop and press Tab any number of times (at least once) and then try page up/down or arrow up/down but nothing happens. Close DC Desktop and start it again, then click on hamburger, then click on `Settings` and finally press PgUp/PgDown/ArrowUp/ArrowDown/... and nothing happens.

- Screenshots: N/A
- Logs: N/A

WofWca:
> scrolling works only with mouse wheel but not with page up/down nor with arrow up/down

Fixed in https://github.com/deltachat/deltachat-desktop/pull/4269

> Settings dialog

https://github.com/deltachat/deltachat-desktop/pull/4319

Keyboard navigation has been significantly improved in 1.48. The biggest thing remaining now is https://github.com/deltachat/deltachat-desktop/pull/4294, and I guess we have https://github.com/deltachat/deltachat-desktop/issues/2141 to track that.
So, I suggest to close this issue as resolved.

PR #4211, opened 2024-10-12
WIP: improvement: arrow-key navigation for chat list

WofWca:
https://github.com/user-attachments/assets/b43fb2c6-4c04-4080-ba15-7e644114c8d1

This is more of a show and tell rather than an MR for now. It sort of works, but there are a lot of issues:
- [ ] The tab order is such that the chat list does not immediately follow the search field, so you have to tab through the other navbar items before you get to the chat list, same for going back from the chat list to the search bar.
- [ ] chat list might get reordered. A chat is moved up when a message is sent to it, but the arrow-key order remains the same as it was on initial render. This results in the focus jumping several items
- [ ] "Home" and "End" do not work due to the fact that we use `react-window` - not all items are rendered, so we can't jump to the first and the last item
- [ ] Worst of all, while you scroll the chat list, it keeps jumping by itself, because when the currently focused chat item gets unrendered, `react-roving-tabindex` focuses the next element (which is out of view) and the browser scrolls it in view.

Related: https://github.com/deltachat/deltachat-desktop/issues/2784

This is based on #4210.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,7 +50,7 @@
 - cancel old message highlight animations when a new message is highlighted #4203
 - fix: packaging: include architecture in filename for all appimages #4202
 - fix: make open external link scheme case insensive #4201
-- improve accessibility a little #4133
+- improve accessibility a little #4133, #4210
 
 <a id="1_46_8"></a>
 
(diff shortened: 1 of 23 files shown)


Simon-Laux:
> "Home" and "End" do not work due to the fact that we use react-window - not all items are rendered, so we can't jump to the first and the last item

you can add event listeners for them and then scroll to the first or last index in them

WofWca:
> you can add event listeners for those keys and then scroll to the first or last index in them

Sure, but they're built into the library that I added in this MR. This is the problem. They're not great friends with `react-window`.

PR #4216, opened 2024-10-15
Less vertical empty space for img attachments

maxphilippov:
Make wide but short image attachments display nicer.

Before:
image
After:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,7 +53,7 @@
 - cancel old message highlight animations when a new message is highlighted #4203
 - fix: packaging: include architecture in filename for all appimages #4202
 - fix: make open external link scheme case insensive #4201
-- make wide but short message attachments have less empty space on top/bottom
+- wide but narrow images have less empty space on top/bottom
 
 <a id="1_46_8"></a>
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -53,6 +53,7 @@
 - cancel old message highlight animations when a new message is highlighted #4203
 - fix: packaging: include architecture in filename for all appimages #4202
 - fix: make open external link scheme case insensive #4201
+- make wide but short message attachments have less empty space on top/bottom
```
r10s, line 56:
```suggestion
- wide but narrow images have less empty space on top/bottom
```

Simon-Laux:
I think the remaining padding is also weird. I think the image should go right to the corners (unless text is larger).
What do you think @r10s? how do dc-iOS and dc-android handle image with caption / forwarded header?

Some background on the current state: I believe we needed fixed height at some point to prevent the message list from jumping around while loading images (when it changes message height).

maxphilippov:
> Some background on the current state: I believe we needed fixed height at some point to prevent the message list from jumping around while loading images (when it changes message height).

Oh, make sense, I'll look for a better solution then

r10s:
> I think the remaining padding is also weird. I think the image should go right to the corners

agree. usually, this also happens

> how do dc-iOS and dc-android handle image with caption / forwarded header?

looks like the following on android:

Screenshot 2024-10-16 at 12 19 36

ios is similar wrt image sizes, the text is show in an transparent overlay there.

maxphilippov:
I'll close it for now, this needs a better approach

PR #4217, opened 2024-10-16
Change archive back icon

nicodh:
resolve #4148

### DIFF
--- a/packages/frontend/scss/main_screen/_main_screen.scss
+++ b/packages/frontend/scss/main_screen/_main_screen.scss
@@ -5,23 +5,23 @@
   position: relative;
   overflow: hidden;
 
+  .backButton {
+    $backBtnText: white;
+    .backButtonIcon {
+      background-color: $backBtnText;
+    }
+
+    &:hover {
+      background-color: rgba(143, 153, 168, 0.15);
+    }
+  }
+
   &.small-screen {
     .message-list-and-composer,
     .chat-list,
     .media-view {
       width: unset;
       flex-grow: 1;
-    }
-
-    .backButton {
-      $backBtnText: white;
-      .backButtonIcon {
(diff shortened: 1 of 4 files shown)


WofWca (approved):
image

r10s (commented):
very nice to push that forward! that will be a big usability improvement!

currently, only the color is wrong - this is visible also in dark mode, but even more in light mode:

Screenshot 2024-10-16 at 11 07 41

the arrow should just get the color of the text. this would also improve the display in settings (where `<backButton>` probably comes from), where the grey looks a bit disabled, just using text color would also better there

nicodh:
Icon color is now the same as text color

r10s:
> Icon color is now the same as text color

it picks up body text color, but not title bar text color 😅 (which are maybe same in dark mode, but not in lite mode)

Screenshot 2024-10-16 at 16 35 44

nicodh:
> > Icon color is now the same as text color
> 
> it picks up body-text-color, but not title-bar-text-color 😅 (which are maybe same in dark mode, but not in lite mode) 
> (sorry if i have been unclear above)

Oh my gosh! Sorry is on my side! - I only checked it in the settings dialog, so it seems we have to declare a special prop to differentiate if we are on a dialog header or a navbar.

Seems too easy...
I will have a look tomorrow

nicodh:
> i just noticed, that we have the arrow already in the correct layout when the chatlist is hidden:
> maybe better reuse that?

where is that?

nicodh:
I now reuse the back button from small screen

No idea why it was not used also on default screen size before

r10s (approved):
very nice!

cannot say much to the css, but it works and looks as expected now :)

Simon-Laux (approved):

PR #4221, opened 2024-10-16
Buttons accept HTMLButton props

maxphilippov:
- make Button and some other components accept HTMLButton props
- rename Button.props.type to styling so we don't conflict with <button type="">

### DIFF
--- a/packages/frontend/src/components/Avatar/index.tsx
+++ b/packages/frontend/src/components/Avatar/index.tsx
@@ -5,7 +5,6 @@ import useDialog from '../../hooks/dialog/useDialog'
 import FullscreenAvatar from '../dialogs/FullscreenAvatar'
 
 import type { Type } from '../../backend-com'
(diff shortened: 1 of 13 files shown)


maxphilippov:
A lot of files touched, but pretty straightforward:
- Button accepts all properties of HTMLButton (`type` renamed to `styling` to avoid conflict with `<button type="">`)
- FooterActionButton accepts all properties of Button (`danger` replaced with `styling="danger"` where applicable)
- ClickForFullscreenAvatarWrapper also passes all of the HTMLButton props further

Rationale:
- we're working with focus in Dialog right now so it's useful to specify things like "autoFocus" and "formMethod"
- it's also much cleaner without repeating what's already defined in HTML types

Simon-Laux (changes requested):
styling of the buttons in profile creation is lost.

maxphilippov:
Okay, that was very silly of me

PR #4224, opened 2024-10-18
improvement: arrow-key navigation for chat list

WofWca:
https://github.com/user-attachments/assets/b43fb2c6-4c04-4080-ba15-7e644114c8d1

Implements the "roving tabindex" approach:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/Keyboard-navigable_JavaScript_widgets#technique_1_roving_tabindex.

Supersedes https://github.com/deltachat/deltachat-desktop/pull/4211.

This improves things, but the UX as a whole is not great yet:
- The tab order is such that the chat list does not
  immediately follow the search field,
  so you have to tab through the other navbar items
  before you get to the chat list.
  Same for going back from the chat list to the search bar.
  This can be addressed by #4225
- The initially "active" element is just the first chat item,
  and not the currently selected chat.
  I can fix it, in a separate MR. I left a TODO.
- Since the chat list is "virtualized", the currently active element
  might get removed from DOM when the user scrolls,
  thus we lose track of the item that was last selected.
  This is pretty hard to fix. Our virtualization library does not have nice API for this, but luckily IMO this drawback is not that impactful.

Related: https://github.com/deltachat/deltachat-desktop/issues/2784

TODO:
- [x] This is based on https://github.com/deltachat/deltachat-desktop/pull/4210, so only look at the last commit.
- [x] Test it
- [x] Add a CHANGELOG entry

If it looks and behaves good in general, I'll proceed with fixing the abovementioned issues, and adding this to other places, such as contact lists and the account list, in separate MRs perhaps.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased][unreleased]
 
 ## Added
+- accessibility: arrow-key navigation for the list of chats #4224
 - Add "Learn More" button to "Disappearing Messages" dialog #4330
 - new icon for Mac users
 - smooth-scroll to newly arriving messages instead of jumping instantly #4125
--- a/packages/frontend/src/components/chat/ChatList.tsx
+++ b/packages/frontend/src/components/chat/ChatList.tsx
@@ -44,6 +44,7 @@ import type {
   MessageChatListItemData,
 } from './ChatListItemRow'
 import { isInviteLink } from '../../../../shared/util'
+import { RovingTabindexProvider } from '../../contexts/RovingTabindex'
 
 const enum LoadStatus {
   FETCHING = 1,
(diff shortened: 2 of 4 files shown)


nicodh (approved):

nicodh:
Review by testing

WofWca:
Rebased and adjusted comments.

WofWca:
Rebased

WofWca:
@r10s could check the functional aspect of this MR? I want to start adding this everywhere in the app, namely in the accounts list, contact list, for various tabs (e.g. QR reader, emoji picker), in the gallery, and, most importantly, in the messages list. In separate MRs, of course. I want a green light so that I don't waste the efforts in case the entire approach is wrong.

@nicodh @maxphilippov could you also check if this is good code-wise?

WofWca:
Rebased

WofWca:
I talked to @r10s privately. He's not opposed to the general direction, and said that his approval should not be a blocker. @nicodh also looked at the code.

So, let's :shipit:! It's about time.

PR #4227, opened 2024-10-19
fix: don't overflow gallery tabs on narrow screen

WofWca:
Make the tabs list scrollable instead of expanding
the entire gallery view


nicodh:
I can't see any overflow of the tabs. At list not by moving the window size to the minimum (which seems to be 375 pixel)

WofWca:
It can overflow on different translations, and the search input actually does overflow even with the screen size limit - try typing a lot of text there.

nicodh (approved):

PR #4234, opened 2024-10-21
show "Mark All As Read" in account menu unconditionally and move it down, right over settings menu item

### LINKED ISSUE #3979: show "Mark All As Read" unconditionally
the sidebar introduced a new function "Mark All As Read", which, however, is unexpectedly not available when a _single chat is muted_  - but esp. then you want this function as one cares less about a chat anyways :)

same for muted archived chats, which also should count into "All" 

the reason for the issue be that jsonrpc-pendents of [`dc_get_fresh_msg_cnt()`](https://c.delta.chat/classdc__context__t.html#a6d47f15d87049f2afa60e059f705c1c5)/[`dc_get_fresh_msgs()`](https://c.delta.chat/classdc__context__t.html#a5dc16d0ebe4f837efb42b957948b54b0) are used to show/hide the option and to mark chats as read - however, already by its name "fresh" is not "all" :)

~~if these information are not easily available in the UI, we should consider adding dedicated core functions as   `dc_has_unnoticed_chats()` and `dc_marknoticed_all_chats()` (due to multiple meanings, we avoid the the term "read" in the api, [cmp](https://c.delta.chat/classdc__msg__t.html#a83fbf6e74d09a0b598ccefe9b48bd68c))~~

EDIT: idea is to just show the item unconditionally, above "Settings", see discussion below

_no "Mark All As Read" here although there are unread chats_

Simon-Laux:
closes #3979

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,7 @@
 - some reactions dialog items not being clickable on secondary accounts (profiles) #4228
 - target-electron: make sure log of stdio server is also logged to file
 - improve accessibility a little #4133
+- fix "Mark All As Read" in account menu mark also archived chats as read
 
 
 <a id="1_46_8"></a>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 - replace BlueprintJS Button, Icon, Radio, RadioGroup, Collapse, Dialog with our implementation
 - Update `@deltachat/stdio-rpc-server` and `deltachat/jsonrpc-client` to `1.147.1`
 - Update proxy configuration - a full url can be entered now
+- show "Mark All As Read" in account menu unconditionally #4234
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


WofWca (approved):
Did not test and did not assess whether this is what we want, but if it is, then LGTM.

r10s (approved):
very nice, this is much better than before!

the archive needs to be marked as read as well, however: "all" on account level includes "archive" already visually. just all blue for grey badges should disappear :)

but that can also go to another PR/issue if showhow complicated

Simon-Laux:
> just all blue for grey badges for that account should disappear :)

really confusing sentence, I don't get that. I assume marking archived chats as read will address that, but please say if you meant something else.

r10s:
there were words wrong in that sentence, i fixed that above.

but yes, "marking archived chats as read" was what i meant. thanks a lot!

r10s (approved):

Issue #4244, opened 2024-10-22
tab navigation: context menu opened via shortcut opens at weird position


r10s:
- navigating via `Tab` to a message's three-dot-menu to open the context menu
- hit `Space` or `Enter` to open the menu

=> the menu is opened in the upper left corner of the screen; on mac even below the window control elements

(instead the context menu should be opened beside the three-dot-menu, or at least at a position not covered by buttons)

https://github.com/user-attachments/assets/a78d41d3-bfee-4380-a956-69a7aa33cb1e

tested on macOs14.5 with deltachat1.46.8

WofWca:
Same goes for opening "react" through a message's context menu with keyboard.

PR #4246, opened 2024-10-23
fix: a11y: wrong positioning of context menus

WofWca:
...when activating them with keyboard (or via other means?).

Closes https://github.com/deltachat/deltachat-desktop/issues/4244.

Some of the affected places:
- Three-dot button besides a message.
- Reactions bar when activated from the message context menu.
- "More options" in the QR reader dialog.
- "Appearance" settings color picker
    (it is not yet keyboard-accessible though).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 - fix Delta Chat not launching on Debian sometimes due to missing package dependencies (`libasound2`) #4275
 - fix not being able to remove avatar for a mailing list #4270
 - fix deleting messages with broken video attachment from gallery #4283
+- accessibility: wrong positioning of some context menus and popups when activating them with keyboard #4246
 
 <a id="1_47_0"></a>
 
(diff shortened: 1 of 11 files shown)


nicodh:
How can I test this? For me (on Linux) Ctrl + F10 works as shortcut to open the context menu in Chrome browser, but not in DC Desktop ?

WofWca:
Weird. We use the native `contextmenu` event, so IDK. Maybe try the "context menu" button (to the right of the right Alt), if it's on your keyboard.

But for the places that I mentioned you don't really require it. You can open the three-dot menu with spacebar if you navigate to it by pressing Tab. Same for activating the "React" menu item (with Enter). Same with the gear icon in the QR scanner.

nicodh (approved):
Reviewed by testing

WofWca:
This can be postponed until the next release, keyboard accessibility is not great anyway right now.

PR #4251, opened 2024-10-24
add task to release doc to document appstore releases centrally on pages repo

Simon-Laux:


RELEASE.md
```diff
@@ -49,6 +49,8 @@
   - [ ] arch
   - [ ] nix
   - [ ] (community) snap
+  after all appstores are done:
+  - [ ] add/update entry on https://github.com/deltachat/deltachat-pages/blob/main/releases.md
```
r10s, line 53: iirc, the idea of `releases.md` is to see at one glance the status of releases for all distribution channels. so, what is really out there :)

to achieve that, it may be needed to update `releases.md` also if not all stores are done, esp. as there are often some channesl that get stuck for weeks, or do not get a release at all. that way, one can see that eg.  that core123 is available in macOS - but stock on Windows for reasons. or that android456 is out on Google Play - but not on F-droid since weeks

so, maybe a bit more neutral here:

```suggestion
  update <https://github.com/deltachat/deltachat-pages/blob/main/releases.md> on progress
```

Simon-Laux: I bet that if it is not a todo item it will get forgotten.
There is always the release progress issue on desktop, so if you want to know the state of the release, better look there directly. we could also establish that the latest release progress issue should be pinned for easier access.

r10s: > I see the use case more in finding out fast, what core version was used based on ui version without checking the long UI changelog easily

if that is the gist, indeed on desktop that information is hard to find.

on android/ios we always add a "update to core123" or "using core123" to the end of each changelog entry, so we could just add a link to the changelog from `release.md` then.

but anyways, it is also about what is doable, and yes, a concrete checkpoint is better than a vague additional sentence. 

so, feel free to merge that in, i was just wondering about the gist.

r10s (approved):

RELEASE.md
r10s: ftr, releases.md gets deleted alltogether, see https://github.com/deltachat/deltachat-pages/pull/994 , so, this PR can probably be reverted, idk, if there is sth. else involved for the releases-issue template

Issue #3018, opened 2022-11-16
Cleanup css theme variables


Simon-Laux:
_themebase.scss is messy, many of the css variables it generates are reused by other things so you don't know what they affect anymore at first glance. I think we need rules for it.

My suggestion:
- in the theme keep the simple base colours (`$bgPrimary`, `$bgSecondary`, `$danger`, `$success`, `$textColor`, ...)
- in themebase define all different css variables and don't reuse them in different components, with a comment on top describing what component it is for. (maybe even with the filename of the file using it, though that can be found out with a simple search already, point is to specify all variables, they aren't expensive anyway and not reuse them in different components, also find somewhat descriptive names for them)
- newer hardcode colors
- document new rules in `docs/STYLES.md`
- make up naming rules for css vars so the naming is consistent

other wording:
- never hardcode colors, use css vars instead
- every component gets a css var defined in theme base
- in theme base the colors get hardcoded or is applicable set dynamically by deriving them from the sass variables defined in the themes.

Simon-Laux:
we still haven't written down rules for the naming.

> never hardcode colors, use css vars instead

not true yet.

> every component gets a css var defined in theme base

we don't have that yet. 

There were some internal discussions of just going for base colors, but we haven't discussed that here nor really came to a conclusion/discussion-result. (I think the two step way of theme-base still makes sense: define base colors in the theme file and they get distributed to more specific vars in theme-base, and components use these more specific variables)

I think if the current state is "fine for now", then we could move this to project resurrection and look at it again in the future.

PR #4259, opened 2024-10-24
style: fix poor visibility of some text

WofWca:
I searched for (almost) all explicitly defined colors
and fixed the most obvious.


r10s:
did not test, just a general question:

is it a usual, modern way to define grey tones using opacity instead of a color? i would expect that to be slower, but apart from that, also not that much easier to read

WofWca:
> is it a usual, modern way to define grey tones using opacity instead of a color?

IDK whether you can call it a standard, I think there are not that many people who care about CSS XD
I'm just trying to save up on defining extra variables (which I'm not sure if that's something that needs to be done).

Simon-Laux:
I recently discussed css variables with @nicodh, basically we want to come up with some rules, that we also document in the `docs/STYLES.md`.

The idea is outlined in https://github.com/deltachat/deltachat-desktop/issues/3018
Basically:
- never hardcode colors, use css vars instead
- every component gets a css var defined in theme base
- in theme base the colors get hardcoded or is applicable set dynamically by deriving them from the sass variables defined in the themes.

WofWca:
I'll keep this open for now as "issue". If someone wants to re-do it properly, go ahead (I myself might come back to it later).

oddlid:
Not sure if this is the right place to comment, but I'd like to add that with all the dark themes, I find it quite hard to see the sender names against the message background.

WofWca:
I turned this into two MRs. The account address has been removed in b703d6f4ce428d8b3b58424552a49b55f1454e9a.
So this MR is superseded.

PR #4260, opened 2024-10-24
improvement: light theme for account list sidebar

WofWca:
Probably still not ideal, but should be better than a dark one.


r10s:
hm.  the title being dark partly but not filling the whole width, being interrupted by a vertical light bar looks clunky to me. this does not really look like an visual improvement.

also, the account and chat icons look too less separated, it is much less clear that taping a square icon changes everyrhing on the right (this is true also for dark mode, but we should not get the flaws from dark mode to light mode - also in dark mode things are expected to be less colorful)

what exactly is the issue with having  dark parts? iirc, slack and other comparable apps look similar.  light mode does not mean _everything_ needs to be dark. it is sort of a menu that is separated by color, similar to the title or other elements.

WofWca:
Makes sense. I just thought it look a bit odd, and that if it's dark in the dark theme, it should be light in the light theme. But perhaps I didn't look at the bigger picture.

PR #4265, opened 2024-10-25
fix: click on webxdc info icon not opening app

WofWca:
The feature was introduced in 7d24b5a64a5e240070aaae07a45663c065cd7eef,
but apparently it never worked.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased][unreleased]
 
+## Added
+- clicking on webxdc info message icon opens the webxdc app #4265
+
 ## Changed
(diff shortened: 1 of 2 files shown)


Simon-Laux:
I don't know if that was an intentional feature or just a copy paste artefact.
Clicking on the info message already jumps to the webxdc message, I think it is confusing if clicking on the icon does something else (because that click target is so small).

We'll soon also add the ability to add deep links to info messages, then those messages will open the app at a specific place when you click on them.

r10s:
i agree, it will be more confusing than helpful, if clicks on different areas do different things. just moving up, wherever you tap is better, for now

WofWca:
Closing in favor of https://github.com/deltachat/deltachat-desktop/pull/4268

PR #4269, opened 2024-10-26
fix: PageUp/Down only working for messages list

WofWca:
Some examples where this commit makes Page Up / Page Down work:
- chat list
- gallery
- settings
- contact lists
- accoutns list
- chat audit log


Simon-Laux (approved):
code LGTM, didn't test

WofWca:
I forgot to mention that this makes the messages list scroll smoothly when it is focused.
It will only scroll instantly when the composer is focused instead.

Simon-Laux:
scrolling in Gallery breaks as soon as the focused item scrolls out of view.

WofWca:
Oh well. I guess it's the reality of `react-window`...
Maybe it makes sense to add `PageUp` / `PageDown` shortcuts in #4291, as suggested in https://www.w3.org/WAI/ARIA/apg/patterns/grid/#datagridsforpresentingtabularinformation, or simply settle on `ArrowDown` / `ArrowUp`.

Issue #3883, opened 2024-05-30
helper function for user visible errors on exception in async action


Simon-Laux:
we should make a helper function for this:

Some hook like
```js
function useCallbackWithUserAlert(cb, dependencies, alertMessage) {
 const openAlertDialog = useAlertDialog()
 return useCallback(()=>{
  return async (...args) => {
  try {
    await cb(...args)
  } catch (error) {
    log.error(alertMessage, error)
    openAlertDialog({
      message: alertMessage + error && error.message ?? error.toString()
    })
  }
 }, [openAlertDialog,...dependencies])
}
```
that we could use in these places:

```js
const startChatWithContact = useCallbackWithUserAlert(async (addr: string) => {}, [])
```

_Originally posted by @Simon-Laux in https://github.com/deltachat/deltachat-desktop/pull/3840#discussion_r1620768005_

WofWca:
There could be a way to use a callback for all unhanded promise rejections.

But the proposed idea seems fine.

Though I am not sure how many places there are where this is applicable.

Simon-Laux:
> There could be a way to use a callback for all unhanded promise rejections.

Those are already logged. I think opening a dialog for every one of them could be annoying if we have some bug in some method that gets called often, say in `maybe_network` or similar. In those cases DC might become unusable because it shows many of those dialogs.

I think it makes most sense to do what I proposed to control when a dialog will be shown.
Sure requires more discipline by developers to not forget it, but does not have the risk of making DC unusable by opening tons of alert dialogs.

> Though I am not sure how many places there are where this is applicable.

many places where we call core in response to a user action -> mute chat for would be easy example.

Simon-Laux:
If the user triggered an action they can get an error/alert dialog when it fails. When some method throws not triggered by user action (all other uncaught exception) it just gets logged how it's currently logged.

I don't understand what speaks against this solution.

WofWca:
Just FYI, a few examples where this turned out to be useful:
- Clicking "Resend" on an info message (core returns error)
- Clicking "Resend" on a group where `!canSend`.

r10s:
closing stale issue; we usually do not keep refactoring and small code-level things open

if it turns out to be advantageous to do some refactoring, it should just be done it on adding/improving a feature; it is usually not a task on its own and, as here, tends to get stale

PR #4272, opened 2024-10-26
add missing cancel buttons to import-/export- and reveive-backup progress dialogs

### LINKED ISSUE #4267: add "cancel" button to "backup"
the "export backup" progress bar dialog lacks a "cancel" button, see https://github.com/deltachat/deltachat-desktop/issues/4263 for a screenshot

in deltachat-ffi, the progress can be stopped using [`dc_stop_ongoing_process()`](https://c.delta.chat/classdc__context__t.html#a4df8b12ccbfb482a0557d027eae43cfc), not sure how that works in deltachat-jsonrpc

EDIT: "add second device" ("send backup") has a cancel button. i did not check if "add as second device" ("receive backup") or "import backup" are cancellable

Simon-Laux:
fixes #4267


WofWca (commented):
1. Have you checked whether it's fine not to handle `stopOngoingProcess` promise rejecting? I'm just looking at `LoginForm.tsx` where it is sort of handled.

2. Also, have you checked whether `stopOngoingProcess` actually applies for all the underlying processes?

3. If so, do we actually handle cancellation well?

I'm not telling you to do this. Just yes / no is fine.

Simon-Laux:
> 1. Have you checked whether it's fine not to handle stopOngoingProcess promise rejecting? I'm just looking at LoginForm.tsx where it is sort of handled.

If it is not fulfilled/resolved then the dialog is not closed.
We can make this prettier with an error dialog as part of #3883 which involves going over all code anyway to find the places where it is needed.

> 2. Also, have you checked whether stopOngoingProcess actually applies for all the underlying processes?

no, but @r10s wrote to use it in #4267 and it worked when I tested it.

> 3. If so, do we actually handle cancellation well?

That's up to core mostly.

Though I saw something else that could be improved: when the receiver cancels backup transfer the transfer dialog on the sender side just closes without showing any error. Maybe that's worth an issue, but this problem was there before this pr, so nothing for this pr.

WofWca:
2. But he said
    > i did not check if "add as second device" ("receive backup") or "import backup" are cancellable

WofWca:
3. > That's up to core mostly.

    But at least the promise of the running process would reject. Do we handle that?

Simon-Laux:
> 2 But he said
> > i did not check if "add as second device" ("receive backup") or "import backup" are cancellable

I tested it, it appeared to have worked. And I just looked into the code in core: `context.alloc_ongoing()` is used in all cases (send/receive/import/export backup, import/export key and configuration).

> But at least the promise of the running process would reject. Do we handle that?

in most cases we should. I think yes. but it could have failed also before this pr.
we can look into it in detail later in the process of doing https://github.com/deltachat/deltachat-desktop/issues/3883

nicodh (approved):
By testing: works as expected

PR #4274, opened 2024-10-26
go back to a custom progressbar implementation

### LINKED ISSUE #4263: progress bar layout regression
- the progress bar is black in light theme, this is not what the user has learned and looks as if things are already done - although not event started :)

- the progress bar dialog has an odd yellow focus border, this does not make much sense (wondering, btw, why there is no cancel button, but that is another issue)

we should go back to the layout of 1.46.x

progress bar on 1.47 - note: this is the state of 0%:

Screenshot 2024-10-25 at 12 20 50

progress bar on 1.46.x:

Screenshot 2024-10-25 at 12 21 50

Simon-Laux:
because the browser native element is too limited.
I added the aria role and attributes, so acessibility should work in theory.

(/me wasted way too much time trying to understand how to style the native element.)

fixes #4263

preview:

| dark | light |
|--------|--------|
| Bildschirmfoto 2024-10-26 um 21 24 50 | Bildschirmfoto 2024-10-26 um 21 23 27 |

### DIFF
--- a/packages/frontend/scss/login/_login.scss
+++ b/packages/frontend/scss/login/_login.scss
@@ -201,25 +201,29 @@ div.delta-form-group {
   }
 }
 
-progress.delta-progress-bar {
+div.delta-progress-bar {
   width: 100%;
+  background-color: var(--progressBarBg);
+  height: 8px;
+  margin-block: 21pt;
+  --progressBarBorderRadius: 5px;
+  border-radius: var(--progressBarBorderRadius);
 
-  // https://stackoverflow.com/questions/18368202/how-can-i-set-the-color-for-the-progress-element
-  accent-color: var(--progress-bar-color);
-  &::-moz-progress-bar,
-  &::-webkit-progress-value {
+  div.bar {
     background: var(--progress-bar-color);
+    height: 100%;
+    border-radius: var(--progressBarBorderRadius);
   }
 
   &.delta-intent-primary {
-    --progress-bar-color: var(--colorPrimary);
+    --progress-bar-color: var(--progressBarPrimary);
   }
 
   &.delta-intent-success {
-    --progress-bar-color: green;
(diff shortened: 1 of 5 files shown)


packages/frontend/scss/login/_login.scss
```diff
@@ -201,25 +201,29 @@ div.delta-form-group {
   }
 }
 
-progress.delta-progress-bar {
+div.delta-progress-bar {
```
WofWca, line 204:
```suggestion
.delta-progress-bar {
```

I think `progress` was specified purely for the reader, to say that it is a special element. I think it's better not to specify `div` explicitly if, say, we happen to go from `div` to `span`.

Though feel free to ignore this suggestion, it's just a minor nitpick.

packages/frontend/scss/login/_login.scss
```diff
@@ -201,25 +201,29 @@ div.delta-form-group {
...
   width: 100%;
+  background-color: var(--progressBarBg);
+  height: 8px;
+  margin-block: 21pt;
+  --progressBarBorderRadius: 5px;
+  border-radius: var(--progressBarBorderRadius);
```
WofWca, line 210: Is there really a point to this variable? Why not just `99999px` or `50%`?

Again, feel free to ignore.

WofWca (approved):
Accessibility looks alright, according to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/progressbar_role , the screenshot is also good.

packages/frontend/scss/login/_login.scss
Simon-Laux: used in two places, so using var makes it easy to iterate on in the inspector.

WofWca: Oh. I missed the other place. Apologies.

PR #4276, opened 2024-10-26
Limit options for "Delete Messages from Server" for chatmail accounts

### LINKED ISSUE #4113: Limit options for "Delete Messages from Server" for chatmail accounts
Only the following options make sense for chatmail:
- Automatic (`delete_server_after=0` i.e. the server decides when to delete messages).
- At once after Download (`delete_server_after=1`).

See https://github.com/deltachat/deltachat-android/pull/3147 for the reference implementation. NB: Auto-adjusting of `delete_server_after` isn't needed, Core >= 1.143 will do that on its own (see https://github.com/deltachat/deltachat-core-rust/pull/5840).

Probably this may be done even after switching to new Core, but better before -- at least for consistency with DC Android.

Simon-Laux:
and rename option "never" to "automatic" for chatmail.

closes #4113

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ## Changed
 - style: avoid scrolling to account list items such that they're at the very edge of the list #4252
 - Update local help (2024-10-25)
+- Limit options for "Delete Messages from Server" for chatmail accounts #4276
 
 ## Fixed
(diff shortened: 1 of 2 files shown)


WofWca (approved):
Looking at the code, appears to close #4113, though I did not test.

PR #4277, opened 2024-10-26
Update `@deltachat/stdio-rpc-server` and `deltachat/jsonrpc-client` to `1.148.4`

Simon-Laux:


Simon-Laux:
does not work, for some reason package is not published to npm.
also don't restore this commit, lock file is wrong

Issue #4232, opened 2024-10-21
Browser Edition - Remaining To Do


Simon-Laux:
- [x] Background Image #4306
- [X] Themes (without the option for custom themes, as you can easily edit themes when you are using the browser version as you can only start it from development environment anyway, also `--watch-theme` is not necessary for this step) #4304
- [X] backup/key export  #4303
- [ ] support env vars
	- [x] TEST_DIR <- for testing
	- [ ] CERT_DIR <- for testing
	- [ ] DELTACHAT_LOCALE_DIR
- [x] env var / cli option to use test.html instead of main.html
- [ ] support cli options and document somewhere what is and isn't supported
- [ ] improve the hint/message that only one concurrent client is supported (reload button + and use custom dialog because safari does not present alert if you if the tab/window is not focused/active)
- [ ] well known temp folder that gets deleted/cleared when server is closed or started (for tmp file uploads).
- [ ] clear backup export folder on every web server start / shutdown
---
- [ ] express sometimes hangs after a burst of requests to blobfiles - investigate that function -> but maybe thats also a lock in core? -> can be easily triggered by sending many files with (todo insert pr number of my new pr here) - but after while it allows me to reload the page and then works again.

r10s:
we will create issues on point if needed

Issue #3595, opened 2023-12-19
drag'n'drop unreliable sometimes


r10s:
sometimes, drag'n'drop is unreliable and is just not working, took me quite some time to make this reproducible.

but here is an approach on macos13.6 with deltachat1.42.0:

1. have a **fresh** PDF-scan in the "Notes" app
2. drag'n'drop from the "Notes" app to "Delta Chat" -> **this does not work**
3. in the "Notes" app, right click the PDF and select "Save as"
4. drag'n'drop from the "Notes" app to "Delta Chat" -> **now same drag'n'drop works**

testing with other apps (as "Apple Mail" or "Find") drag'n'drop works already at step 2. - and doing that once, this also enables things to work with "Delta Chat". "Thunderbird" has the same bug, btw

maybe the format drag'n'drop offers changes once the file is exported or otherwise used - and that "Delta Chat" only supports one format.

_video showing steps 1. to 4.:_

https://github.com/deltachat/deltachat-desktop/assets/9800740/76470a21-831f-41d0-b8ac-d941bcd234d6

Simon-Laux:
works for me in macOS 14, though I can not exactly reproduce the scanning part, maybe it is a special state?

Anyways while testing it I noticed that drag and drop of selected text does not work, while it works in vscode and even from notes to vscode, so it seems possible, but out of scope for us for now.

r10s:
> though I can not exactly reproduce the scanning part

long-tap the "Notes" App icon on the ios homescreen  and hit "Scan Document" - if you're using the same apple account on macos and ios, you get the scan on macos.

however, this was just _one_ example i figured out to make the issue reproducible. i have seen that randomly before as well and there may be also other cases where it happens.

Simon-Laux:
Really seems to be sth when the file does not exist in apple's filesystem. not sure what the format then is, but I was not able to get the state yet, maybe I could get it when I connect the same Apple IDs on iPhone and macOS, but currently I don't want to do that.

Simon-Laux:
please reopen if this is still an issue in new versions (macOS and deltachat desktop / electron).

r10s:
i just retried on macos 14.5, it is still an issue and exactly reproducible in the same way as shown in the initial post.

> maybe I could get it when I connect the same Apple IDs on iPhone and macOS, but currently I don't want to do that.

i think, this is needed for reproducing the way shown above.

[...]

PR #4278, opened 2024-10-27
improvements to selecting multiple files to send

Simon-Laux:
- **Dropping multiple files onto deltachat now sends images as compressed images instead of uncompressed files**
- **Added support for selecting multiple files in the attachment file picker.**
- fix compression of images when added with Image option from attachment menu


packages/frontend/src/components/composer/menuAttachment.tsx
```diff
@@ -36,28 +39,56 @@ export default function MenuAttachment({
...
+          return
+        }
+
+        for (const filePath of filePaths) {
+          sendMessage(accountId, selectedChat.id, {
+            file: filePath,
+            viewtype: msgViewType,
+          }).then(() => {
+            // start sending other files, don't wait until last file is sent
+            runtime.removeTempFile(filePath)
```
WofWca, line 64: So this is copy-pasted from

https://github.com/deltachat/deltachat-desktop/blob/f0e4cc5ef0a8177876b322cb0753ee04b7cdece1/packages/frontend/src/components/message/MessageListAndComposer.tsx#L168-L190

but in that code we also do `writeTempFileFromFile`. Why do we have to do that there, but not here.

And, most importantly, if we decide to not create temp files, we should not `runtime.removeTempFile()` here. Currently in Electron this simply gives an error of "Path is outside of the temp folder". I did not check what the browser version does, but it [does not look good](https://github.com/deltachat/deltachat-desktop/blob/5f640db6f6b809f9ffa05822045bf13519cac3a2/packages/target-browser/src/backendApi.ts#L118-L138).

I think it'd be nice to add a docstring to `removeTempFile`, or somehow make it safer in general.

packages/frontend/src/components/message/MessageListAndComposer.tsx
```diff
@@ -175,9 +175,12 @@ export default function MessageListAndComposer({ accountId, chat }: Props) {
 
         for (const file of sanitizedFileList) {
           const path = await writeTempFileFromFile(file)
+          const msgViewType: Viewtype = file.type.startsWith('image')
+            ? 'Image'
+            : 'File'
```
WofWca, line 180: I wonder if we should keep improving this. E.g. add support for GIFs, webxdcs, etc.

WofWca (changes requested):
I did not test the browser version, but the Electron one looks alright, apart from `removeTempFile()`.

I tested it on a surface level, and looked at the code somewhat thoroughly. It appears to only affect the places where `multiSelections` is used, otherwise the behavior of `showOpenFileDialog` is the same.

packages/frontend/src/components/composer/menuAttachment.tsx
Simon-Laux: > but in that code we also do writeTempFileFromFile. Why do we have to do that there, but not here.

because here we get valid file paths.
everything would be cleaner if we could pass files directly to core without these temp files, but that api doesn't exist yet.

For drag and drop I needed to change it, because it was an electron specific api and also did not always work #3595.

> but it [does not look good](https://github.com/deltachat/deltachat-desktop/blob/5f640db6f6b809f9ffa05822045bf13519cac3a2/packages/target-browser/src/backendApi.ts#L118-L138).

I know, needs to be improved, I plan to move the used temp folder to a different place, I already added a todo item to #4232 - in that process we can make the path check even safer.

packages/frontend/src/components/message/MessageListAndComposer.tsx
Simon-Laux: maybe with the composer refactor, currently core changes stuff anyways. I don't have the full current behaviour loaded to my mind, I'm fine with it at the moment as long as it works.

Simon-Laux:
> In addition, I would assume we don't want to release it with 1.48?

works, fixes bugs and brings a usecase back that would otherwise have been removed (send multiple images files uncompressed in one action). So I would release it with 1.48.

packages/frontend/src/components/composer/menuAttachment.tsx
WofWca: > I know, needs to be improved

I'm interested in a sense that it wouldn't work great if you pass normal, non-temporary files there 😅

packages/frontend/src/components/composer/menuAttachment.tsx
```diff
@@ -36,28 +39,59 @@ export default function MenuAttachment({
...
```
nicodh, line 67: The line 
`runtime.removeTempFile(filePath)`
is definitely wrong since as Simon says, filePath is the original file path

"Why do we have to do that there," ?
The reason is that drag & drop only provides [DataTransfer files](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/files) which have no path and we need to pass a local path as file attachement to core

nicodh (changes requested):
I missed before that this includes a bugfix - so if we skip the wrong removeTempFile line I would vote to include it in the release

packages/frontend/src/components/composer/menuAttachment.tsx
Simon-Laux: it is needed for browser, for electron the call just failed because it can not delete non-temp files.

packages/frontend/src/components/composer/menuAttachment.tsx
```diff
@@ -36,28 +39,62 @@ export default function MenuAttachment({
...
+        }
+
+        for (const filePath of filePaths) {
+          sendMessage(accountId, selectedChat.id, {
+            file: filePath,
+            viewtype: msgViewType,
+          }).then(() => {
+            // start sending other files, don't wait until last file is sent
+            if (runtime.getRuntimeInfo().target === 'browser') {
+              // browser creeated temp files during upload that can now be cleaned up
```
WofWca, line 68:
```suggestion
              // browser created temp files during upload that can now be cleaned up
```

WofWca (approved):

PR #4287, opened 2024-10-28
add autocompletion for translation keys

Simon-Laux:
Bildschirmfoto 2024-10-28 um 18 05 34

Just for autocompletion not checking, for checking we have 2 ways:
- ./bin/test_for_missing_translations.sh
- the runtime errors

we can think about using it also for checking in the future. But currently I'd to say no.


WofWca (approved):
Very nice!
Looks like this doesn't have runtime effect.

Issue #4127, opened 2024-09-12
tab-navigation incomplete/erroneous


r10s:
#4005 adds "tab navigation" to improve accessibility, so to help users who cannot or do not want to use a mouse.

the idea of the "tab navigation" is that it can replace a mouse completely.

however, there are some open ends with the navigation:

- [x] you can focus "Add Profile", but you cannot focus the selected Profile
  #4133
- [x] when the selected profile is focused, one would expect to be able to use the cursor keys to move the selection up/down. see the browser or explorer-like apps tabs to get the idea (or the Delta Chat menus, where this works nicely as well)
  https://github.com/deltachat/deltachat-desktop/pull/4291
- [x] the chatlist is not focusable at all, when focused, up/down keys should change selected chat also here
  https://github.com/deltachat/deltachat-desktop/pull/4224
- [x] there is not tab-stop for the "profile"
  https://github.com/deltachat/deltachat-desktop/pull/4210
- [ ] in the message list, random "emoji picker" icons are selected - the ones the mouse is hovered. this does not help on accessibility.
    https://github.com/deltachat/deltachat-desktop/pull/4294
- [x] in the settings: you can navigate forth and back - but you cannot tab-stop on any setting and you cannot change settings that way
    https://github.com/deltachat/deltachat-desktop/pull/4319
- there may be more issues, i did not test every element. in general, each interactive element should be reachable 

sorry for nitpicking, but to make "tab navigation" useful, it should be coherent.

note, that "other shortcuts" are are more for power-users and are not a replacement for "tab navigation" issues: "tab navigation" should be useful _without_ learning additional shortcuts, as said, being used as a mouse replacement

for further reading see eg. https://www.nngroup.com/articles/keyboard-accessibility/

r10s:
> Could you clarify these?
> 
> > in the message list, random "emoji picker" icons are selected - the ones the mouse is hovered

see the following image:

<img width=320 src=https://github.com/user-attachments/assets/3f7db2c1-e4c0-457d-afe2-ea36b3cc9ae0>

the mouse is randomly over the message "gg" - and then the buttons beside the message can be reached via tab. but this does not make sense, the whole idea of the tab-navigation is that the mouse is _not_ needed. and you cannot reach the context menu of another message without mouse.

> > there is not tab-stop for the "profile"
> 
> What profile?

the one that opens when tapping here:

[...]

Issue #2141, opened 2021-02-20
Accessibility: Make the individual chat list more accessible.


MarcoZehe:
- Operating System (Linux/Mac/Windows/iOS/Android): All desktop
- Delta Chat Version: 1.14
- Expected behavior: The context menu should come up not only on links, but on all chat items, not just by right-clicking.
- Actual behavior: Context menus only work for focusable elements within each message, if any, like links.
- Steps to reproduce the problem: Open a chat in DeltaChat and try to open an individual message's context menu without using the mouse.
- Screenshots: N/A
- Logs: N/A

## Proposed solution

Implement the [listbox](https://www.w3.org/TR/wai-aria-practices-1.1/#Listbox) (https://www.w3.org/WAI/ARIA/apg/patterns/listbox/) design pattern, but with a few extras:

1. The context menu key (shift+F10 on Windows and Linux, the Applications/popup key on Windows, and also a custom implemented Shift+F10 on Mac) should open the context menu on the currently focused item.
2. Tab should move to focusable items within a message like links. If implemented correctly, this should be handled by the browser engine automatically. The tricky part is if there are multiple links in different messages, one may want to trap the tab key to move to the Attachments button instead of the next link in a different message. Right now, tab just moves through all the links that are anywhere within the current conversation.
3. Also make the divider items proper list items that can be focused, so users know when they traverse to the next/previous date.

Jikstra:
@MarcoZehe merci beaucoup for filing all the accessibility issues! I want to focus on fixing them for the 1.16 series. The message list will receive a rewrite and in this process i want to fix this issues :)

PR #4292, opened 2024-10-30
WIP: refactor: `useRovingTabindex` to utilize IDs

WofWca:
- **improvement: a11y: make chat list items focusable**
- **improvement: arrow-key navigation for chat list**
- **WIP: refactor: `useRovingTabindex` to utilize IDs**

Issue #4242, opened 2024-10-22
Activating "Message info" in context menu with keyboard does nothing


WofWca:
- Operating System (Linux/Mac/Windows/iOS/Android):
- Delta Chat Version: 9c5787410
- Expected behavior: the "message info" dialog gets opened
- Actual behavior: nothing apparently happens
- Steps to reproduce the problem:
    1. Right-click a message to open the context menu.
    2. Using the keyboard, navigate to "message info".
    3. Press Enter.
- Screenshots:
- Logs:

I found out that the dialog does get rendered, but it gets immediately closed, but _not_ by this code:

https://github.com/deltachat/deltachat-desktop/blob/0e0d0b80d598802ebb972c8d6c05b593bd896f24/packages/frontend/src/components/Dialog/Dialog.tsx#L32-L58

This issue does not appear in 1.46.8, I suspect this is a regression after #4006. This is not super important as of now since keyboard accessibility is lacking in other more important ways, but it will become important soon because we're working on accessibility now (#4210, #4224).

@maxphilippov would you like to take a look, or shall I?

Simon-Laux:
I can reproduce this bug.

WofWca:
There is another, probably related issue on 1.47.1.
1. Using keyboard, focus the "three-dot menu" button on a chat.
2. Press spacebar to open the menu
3. Select "Disappearing messages" with arrow keys.
4. Press enter to activate the menu item.

The dialog will open, but the menu will also remain open, and will retain focus.

PR #4294, opened 2024-10-30
improvement: a11y: arrow-key navigation for messages

WofWca:
https://github.com/user-attachments/assets/76391400-a379-45ac-bc6b-940b7a864853

Closes https://github.com/deltachat/deltachat-desktop/issues/2141.
Together with other previously merged MRs, closes https://github.com/deltachat/deltachat-desktop/issues/4127.

Basically what this commit comes down to:
1. Apply `useRovingTabindex` for message items
2. Set `tabindex="-1"` on all the interactive items
    inside every message that is currently not the active one,
    so that they do no have tab stops.

TODO:
- [x] Address the TODOs in the code
- [ ] Manage what's gonna be the initially active message,
    because initially they're all active, so
    tabbing to the messages list from the top selects
    the first rendered one as the active one.
    https://github.com/deltachat/deltachat-desktop/pull/4292
    could help with this.
    And remember that the messages list is dynamic (a new message can arrive), so simply setting the last one as active once won't cut it.
    This is also not great for performance: changing `tabindex`
    on a bunch of messages makes them all re-render.
    And otherwise, we probably want to update which one is
    the active one as new messages arrive.
    Perhaps we can simply implement the "End" shortcut as sort of a workaround for now: https://github.com/deltachat/deltachat-desktop/pull/4438
- [ ] Some `jumpToMessage` calls probably ought to focus the message. For example, when jumping to a message through a click on a quote, or a webxdc info message. But not all `jumpToMessage` should change focus. For example, receiving a new message shouldn't do it. This probably requires that we add a `focus` parameter to `jumpToMessage`. Perhaps even 2 parameters: whether to set the message as the active one in the messages list (without focusing it), and whether to focus it. But consider the fact that then the composer will be blurred and will have to be manually focused.
    This can be done in further MRs.
- [x] The interactive items with `onClick` must be actual semantic
    `<button>`s.
    See https://github.com/deltachat/deltachat-desktop/pull/4210
    for reference.
    Namely, this goes for [`Attachment`](https://github.com/deltachat/deltachat-desktop/blob/d1a3f8dbebe4cfa9f74ec4e53e29637eb1e7c573/packages/frontend/src/components/attachment/messageAttachment.tsx#L34), and all the `onClick` listeners in `Message.ts`.
    Update: here is an MR: https://github.com/deltachat/deltachat-desktop/pull/4429
- [x] Set `tabindex` on `<video>` and `<audio>`
- [x] Does `onFocus` work? Does clicking something inside a message make that message the active one?
- [x] Same for `onKeyDown`. Does pressing an arrow key when a child element is focused switch the active message?
- [x] Some interactive contents, namely videos and audios, utilize arrow keys, i.e. for seeking, changing volume. We must not invoke our onKeyDown handler in such cases, switching focus to a different message.
  FYI This also applies to audios in gallery, see https://github.com/deltachat/deltachat-desktop/pull/4376
- [ ] Shift + F10 on a webxdc info message, then "Reply" causes the webxdc app to open. This is similar (or the same?) as https://github.com/deltachat/deltachat-desktop/issues/4242
- [ ] Looks like we need to set `role` attributes so that things like NVDA don't override the arrow key presses to manage focus on their own. See ["How do I disable NVDA's arrow key navigation"](https://stackoverflow.com/questions/41473174/how-do-i-disable-nvdas-arrow-key-navigation). `role` is also mentioned in https://www.w3.org/WAI/ARIA/apg/patterns/listbox/.
  The example that [MDN links to](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Keyboard-navigable_JavaScript_widgets#technique_1_roving_tabindex) uses `role="application"` on the list. I tried setting `role="application"` on our `<ul>` that contains the messages and it started working, but I am not sure if this is correct.
  I created a separate issue: https://github.com/deltachat/deltachat-desktop/issues/4444
- [ ] As mentioned below, "react" doesn't work: you can't navigate the reaction picker with keyboard, and cannot close it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [Unreleased][unreleased]
 
 ## Added
-- accessibility: arrow-key navigation for gallery and sticker picker #4376, #4372
+- accessibility: arrow-key navigation for message list, gallery and sticker picker #4294, #4376, #4372,
 - accessibility: arrow-key navigation: handle "End" and "Home" keys to go to last / first item #4438
 - add show_app_in_chat option to webxdc info message context menu #4459
 
--- a/packages/frontend/scss/message/_message-list.scss
+++ b/packages/frontend/scss/message/_message-list.scss
@@ -123,9 +123,25 @@
       scroll-margin-bottom: $margin-bottom;
       // And this is purely cosmetic. For scrolling to a message
       // that is above the visible area.
-      scroll-margin-top: math.div($margin-bottom, 2);
+      $scroll-margin-top: math.div($margin-bottom, 2);
+      scroll-margin-top: $scroll-margin-top;
 
       min-width: 200px;
+
+      .roving-tabindex {
+        // The contents of the `<li>` are focusable (see `useRovingTabindex`).
+        // Focusing an element inside a scrollable region makes the browser
+        // scroll the focused element into view, so let's also apply the same
+        // margins for it.
+        // This is not as important though, but still nice.
+        //
+        // But ideally we'd probably want to refactor this in such a way that
+        // the focusabe element and the element that we `scrollIntoView()`
+        // when `jumpToMessage` is the same element, so only one
+        // needs `scroll-margin`.
+        scroll-margin-bottom: $margin-bottom;
+        scroll-margin-top: $scroll-margin-top;
+      }
 
       &::after {
         visibility: hidden;
--- a/packages/frontend/src/components/Avatar/index.tsx
+++ b/packages/frontend/src/components/Avatar/index.tsx
@@ -43,6 +43,7 @@ export function Avatar(props: {
   wasSeenRecently?: boolean
   style?: htmlDivProps['style']
   onClick?: () => void
+  tabIndex?: -1 | 0
   className?: string
 }) {
   const {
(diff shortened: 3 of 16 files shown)


WofWca:
I would say this is almost ready to be merged, now waiting for #4376, the rest of the TODOs can be addressed in separate MRs

WofWca:
OK, this is finally ready for review and for merge IMO.
I will create separate issues for the remaining TODO items. Otherwise I think this is an improvement.

nicodh (commented):
So far all my tests were fine!

There is an issue with the reactions. You can open them with key navigation but neither select a reaction nor close the picker again. But that could go into a separate issue

What is still missing imho, is to mention the new features in the keybindings documentation. WHat is the plan about that?

WofWca:
> What is still missing imho, is to mention the new features in the keybindings documentation. WHat is the plan about that?

IDK about that. I wouldn't call those "shortcuts". For example, WhatsApp doesn't have them listed as such. This is basically just an improvement to tab navigation, and I think using arrow keys for long lists is pretty standard.
Especially given the fact that you cannot navigate messages with keys other than arrow keys, so users will learn it naturally I think.

But I don't mind if we still explicitly document it.

nicodh (approved):

PR #4296, opened 2024-10-31
Allow second instance when DC_TEST_DIR is set

nicodh:
PR to e2e tests PR since it uses dotenv for convenience


nicodh:
@Simon-Laux are there other reasons to not allow 2 instances except if they use the same account dir?

packages/target-electron/.env.example
```diff
@@ -0,0 +1 @@
+TEST_DIR=ABSOLUTE_PATH_TO_TEST_DIR
```
Simon-Laux, line 1: comment what `TEST_DIR` is

PR #4300, opened 2024-11-01
fix borderless drag issues and add debug overlay to better test issues about draging area

### LINKED ISSUE #3926: message details window bugs
If the message details window contains enough information to display the side scroll bar, the clickable area of ​​the x-icon is reduced and moved down. This means that a user can no longer close the window by clicking in the middle of the x-icon.

DC-Desktop 1.45.5
macOS 14.5

### LINKED ISSUE #4018: mac: unclear window dragging areas
mac got an "integrated titlebar", which looks great.

however, there are some bugs.

- in enlarge image view (and maybe other views covering the main screen), the dragging areas are the ones from the underlying chat. this results in mostly random behaviour, when the user tries to drag the window - either the window is dragged, the image is dragged or nothing happens. see video#1 below

- in "add account", the window cannot be dragged at all. we sorted that as a "bug that should not hold the release" that time, but meanwhile it is time to get a grip on that :) (not sure if there is another issue)

- in the normal chatview, the width of the "profile open area" is the max. of title/subtitle. this leads to confusion esp. when the subtitle is longer - some parts close to the window margin drag the window, others do not. it seems better to make "profile open area" fit to the text, see video#2 below

https://github.com/deltachat/deltachat-desktop/assets/9800740/1f26dcf4-f657-4969-bd36-263c5f43ed6f

https://github.com/deltachat/deltachat-desktop/assets/9800740/939ca3e1-8fc3-4c34-b6df-b787328ae8f4

Simon-Laux:
Bildschirmfoto 2024-11-01 um 01 59 15
add `exp.showDragAreas()` development api to make drag areas visible.

red is `drag`, green is `no-drag`. 
**How to use**: go into electron dev tools, execute `exp.showDragAreas()` to toggle the overlay which makes drag areas visible.
This is a useful tool to make the issues with it more tangible.

Fix some of the bugs:

- make dialogs no-drag
- add input and button elements in case we missed cases, they should always be clickable
- make background image on "no account selected"- and welcome screen a drag area
- don't make dialog headers drag areas
- remove unused z-index css code from context menu
- make context menu not dragable

todo? or make new issue:
- [ ] Fullscreen media view -> where can I drag that?

closes #4018
closes #3926


WofWca (commented):
I have checked that the added `DragRegionOverlay` is not relevant in production. But it would also be nice to not include its code in the production build.

As to essence of this MR: I haven't checked it thoroughly, because I'm not on Mac.
But in general not so many lines have been changed.

r10s (commented):
this is an impressive layout, thanks for moving forward on that issue; in general, beside fullscreen, that seems to work on quick testing :)

> Fullscreen media view -> where can I drag that?

i would expect that to work in the upper area of the screen, where normally the title bar is - similar to welcome screen

i cannot say much to the code, however :)

r10s:
> Fullscreen media issue can be solved in a later fix imho

agree, but we should file an issue for that then

Simon-Laux:
So todo to merge this pr:
- [ ] add changelog entry
- [ ] file issue for the remaining issue with the full screen media view.

PR #4304, opened 2024-11-01
add theme support in browser edition

Simon-Laux:
- **remove unused `runtime.resolveThemeAddress`.**
- **sort event callback functions by whether we can implement them**
- **make themes work in browser edition**

I intentionally left out `--theme-watch` for now. It is semi-hard to implement here and there is not much advantage to developing a theme with the electron version.


nicodh (approved):
Reviewed by testing

PR #4309, opened 2024-11-02
Show paste button in qr scanner dialog

### LINKED ISSUE #3936: placement of "Paste" in "Scan QR code" not obvious
on some first tests, it turns out that the new position of the QR Scanner's "Paste" button is not obvious.

it even looks as if there is sth. missing as the "Show QR code" as a big "Copy" button where the "Scan QR code" has only empty space.

we should add the "Paste" button to the same position:

Screenshot 2024-06-13 at 00 05 00

the button should read just "Paste". "Past From Clipboard" is too long and too cluttering there. no need to make "Load QR code as image" easier accessible as well.

(i know, this was changed to streamline QR code scanner, and it was worth a try, but it turns out it is not flawlessly working that way)

nicodh:
resolves #3936

move "paste from clipboard" from settings context menu to own button

image

Another option would be: name the paste button "Import" and show the context menu onClick with the 2 options: 'load QR Code image' or 'paste from clipboard'
But buttons that show a context menu on click are a bit unusual

### DIFF
--- a/packages/frontend/src/components/QrReader/index.tsx
+++ b/packages/frontend/src/components/QrReader/index.tsx
@@ -13,7 +13,6 @@ import Spinner from '../Spinner'
 import useTranslationFunction from '../../hooks/useTranslationFunction'
 import { ContextMenuContext } from '../../contexts/ContextMenuContext'
 import { ScreenContext } from '../../contexts/ScreenContext'
-import { runtime } from '@deltachat-desktop/runtime-interface'
 
 // @ts-ignore:next-line: We're importing a worker here with the help of the
 // "esbuild-plugin-inline-worker" plugin
@@ -23,6 +22,8 @@ import styles from './styles.module.scss'
 
 import type { ContextMenuItem } from '../ContextMenu'
 import { mouseEventToPosition } from '../../utils/mouseEventToPosition'
+import { runtime, Runtime } from '@deltachat-desktop/runtime-interface'
+import QrCode from '../dialogs/QrCode'
 
 type Props = {
   onError: (error: string) => void
@@ -64,7 +65,7 @@ async function fileToBase64(file: File): Promise<string> {
 /**
  * Convert base64-encoded blob string into image data.
  */
-async function base64ToImageData(base64: string): Promise<ImageData> {
+export async function base64ToImageData(base64: string): Promise<ImageData> {
   return new Promise((resolve, reject) => {
     const image = new Image()
 
@@ -90,6 +91,29 @@ async function base64ToImageData(base64: string): Promise<ImageData> {
 
     image.src = base64
   })
+}
+
+export async function processClipBoard(runtime: Runtime) {
+  // Try interpreting the clipboard data as an image
+  const base64 = await runtime.readClipboardImage()
+  if (base64) {
+    const imageData = await base64ToImageData(base64)
+    const result = scanQrCode(imageData.data, imageData.width, imageData.height)
+    if (result) {
+      return result
+    } else {
+      throw new Error('no data in clipboard image')
+    }
+  }
+
+  // .. otherwise return non-image data from clipboard directly
+  const data = await runtime.readClipboardText()
+  if (!data) {
+    throw new Error('no data in clipboard')
+  }
+  // trim whitespaces because user might copy them by accident when sending over other messengers
+  // see https://github.com/deltachat/deltachat-desktop/issues/4161#issuecomment-2390428338
+  return data.trim()
 }
 
 export default function QrReader({ onError, onScan }: Props) {
@@ -164,42 +188,6 @@ export default function QrReader({ onError, onScan }: Props) {
     [onError]
   )
 
-  // Read data from clipboard which potentially can be an image itself.
-  const handlePasteFromClipboard = useCallback(async () => {
-    try {
-      // Try interpreting the clipboard data as an image
-      const base64 = await runtime.readClipboardImage()
-      if (base64) {
-        const imageData = await base64ToImageData(base64)
-        const result = scanQrCode(
-          imageData.data,
-          imageData.width,
-          imageData.height
-        )
-        if (result) {
-          handleScanResult(result)
-          return
-        } else {
-          throw new Error('no data in clipboard image')
-        }
-      }
-
-      // .. otherwise return non-image data from clipboard directly
-      const data = await runtime.readClipboardText()
-      if (!data) {
-        throw new Error('no data in clipboard')
-      }
-      // trim whitespaces because user might copy them by accident when sending over other messengers
-      // see https://github.com/deltachat/deltachat-desktop/issues/4161#issuecomment-2390428338
-      onScan(data.trim())
-    } catch (error) {
-      userFeedback({
-        type: 'error',
-        text: `${tx('qrscan_failed')}: ${error}`,
-      })
-    }
-  }, [handleScanResult, onScan, tx, userFeedback])
-
   // Read data from an external image file.
   //
   // We first trigger an "file open" dialog by automatically "clicking"
@@ -271,11 +259,6 @@ export default function QrReader({ onError, onScan }: Props) {
           action: handleImportImage,
           dataTestid: 'load-qr-code-as-image',
         },
-        {
-          label: tx('paste_from_clipboard'),
-          action: handlePasteFromClipboard,
-          dataTestid: 'paste-from-clipboard',
-        },
       ]
 
       openContextMenu({
@@ -283,14 +266,7 @@ export default function QrReader({ onError, onScan }: Props) {
         items,
       })
     },
-    [
-      deviceId,
-      handleImportImage,
-      handlePasteFromClipboard,
-      openContextMenu,
-      tx,
-      videoDevices,
-    ]
+    [deviceId, handleImportImage, openContextMenu, tx, videoDevices]
   )
 
   // Whenever a camera was (automatically or manually) selected we attempt
--- a/packages/frontend/src/components/QrReader/index.tsx
+++ b/packages/frontend/src/components/QrReader/index.tsx
@@ -120,7 +120,7 @@ export default function QrReader({ onError, onScanSuccess }: Props) {
         if (result) {
           onScanSuccess(result.data)
         } else {
-          throw Error(`${tx('qrscan_failed')}: no data in image`)
+          throw Error(`no data in image`)
         }
       } catch (error: any) {
         handleError(error)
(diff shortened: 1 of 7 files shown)


nicodh:
TODO: We have to adapt the e2e test for pasting the qr code! #4285

packages/frontend/src/components/dialogs/QrCode.tsx
```diff
@@ -235,6 +239,63 @@ export function QrCodeScanQrInner({
...
+
+      if (unmounted) {
+        return
+      }
+
+      handleScan(result.data)
+
+      return () => {
+        unmounted = true
+      }
```
WofWca, line 258: So this is copy-pasted. Because `unmounted` doesn't appear to do anything. Just saying.

WofWca (commented):
This removes the "paste from clipboard" menu item from `QrReader`, but `QrReader` is [also used in `ReceiveBackupDialog`](https://github.com/deltachat/deltachat-desktop/blob/a5c02433976909fe01a64f3049c96d0d101ae9d8/packages/frontend/src/components/dialogs/SetupMultiDevice/ReceiveBackupDialog.tsx#L71). Does it mean it's now impossible to paste the backup QR?

r10s:
> Another option would be: name the paste button "Import" and show the context menu onClick with the 2 options: 'load QR Code image' or 'paste from clipboard'
But buttons that show a context menu on click are a bit unusual

let's stay with the paste button.

having that hidden at "import" again is not  obvious, if is also about the balance copy/paste. there could be a three dot menu beside the paste button for the other options, but that is another effort, and maybe not needed.

nicodh:
> This removes the "paste from clipboard" menu item from `QrReader`, but `QrReader` is [also used in `ReceiveBackupDialog`](https://github.com/deltachat/deltachat-desktop/blob/a5c02433976909fe01a64f3049c96d0d101ae9d8/packages/frontend/src/components/dialogs/SetupMultiDevice/ReceiveBackupDialog.tsx#L71). Does it mean it's now impossible to paste the backup QR?

True, that's another use case not covered in the Issue since the left footer button is already set there. So maybe we should move the Troubleshoot link to another place? @r10s

WofWca:
A simple hack would be to keep the "paste from clipboard" menu item just for the Backup dialog. I.e. to add a prop `hidePasteFromClipboardMenuOption` to `QrReader`.

Simon-Laux:
I'd say keep the paste button in the menu or move trouble shooting up to the description.
BTW: I think we can remove the experimental label and version requirement

r10s:
i'd keep the menu as is, that seems to be the easiest approach.

for the "experimental" hint: we're staying with that for so long, let's do not remove that last minute, it is fine.  idk, what the exact state of iroh is, let's see that for 1.50 and maybe get it out on all platforms

Simon-Laux:
> also for reasons, regarding backup/backup2 hickup

wouldn't we need to upgrade the version then or is the first iroh/backup(without the 2) still supported?

r10s:
> wouldn't we need to upgrade the version then or is the first iroh/backup(without the 2) still supported?

i think it is fine, as long as desktop checks for DC_QR_BACKUP as well as DC_QR_BACKUP2 at all places, on android we forgot one place, and that resulted in a degradation, see https://github.com/deltachat/deltachat-android/pull/3370

nicodh:
I readded the "paste from clipboard" item to settings menu

I also refactored the implementation to avoid duplicate code by using [useImperativeHandle](https://react.dev/reference/react/useImperativeHandle)
I removed some existing code since - see comment above

packages/frontend/src/components/dialogs/QrCode.tsx
nicodh: That was code that already existed. I removed it since I could not create a case, where the callback was called in an unmounted component, even with clicking "through" the overlay to close the dialog immediately
@Simon-Laux do you remember why that was added? 
The comment said:
>  // Additionally we have checks in place to make sure we're not firing any
 > // callbacks when this React component has already been unmounted.

nicodh: If you consider this change to "risky" we could either readd it or postpone this PR for the current release

nicodh:
> TODO: We have to adapt the e2e test for pasting the qr code! #4285

Done

packages/frontend/src/components/dialogs/QrCode.tsx
WofWca: To me this looks like dead code. `unmounted` is never read after initialization, it's a new variable on each function execution, and it's only scoped to the function.

packages/frontend/src/components/dialogs/QrCode.tsx
```diff
@@ -228,8 +233,8 @@ export function QrCodeScanQrInner({
           openAlertDialog({
             message: error.message || error.toString(),
           })
+          processingQrCode.current = false
```
WofWca, line 236: Why has this been moved? The logic doesn't look right now.
And a copy-pasted version of this code is also in `ReceiveBackupDialog.tsx`, where it remains unchanged.

packages/frontend/src/components/QrReader/index.tsx
```diff
@@ -129,25 +158,40 @@ export default function QrReader({ onError, onScan }: Props) {
...
+    try {
+      const result = await processClipBoard(runtime)
+      if (typeof result === 'string') {
+        onScan(result)
+      } else {
+        try {
+          onScan(result.data)
+        } catch (error) {
+          console.error('Error while processing clipboard content:', error)
+        }
```
WofWca, line 171: If an error occurs here, `userFeedback` doesn't get called. I believe this is not the case in the pre-MR version, where if `handleScanResult` throws, we do call `userFeedback`. Is this what we want?

packages/frontend/src/components/QrReader/index.tsx
```diff
@@ -92,7 +94,34 @@ async function base64ToImageData(base64: string): Promise<ImageData> {
...
 
-export default function QrReader({ onError, onScan }: Props) {
+async function processClipBoard(runtime: Runtime) {
+  // Try interpreting the clipboard data as an image
+  const base64 = await runtime.readClipboardImage()
+  if (base64) {
+    const imageData = await base64ToImageData(base64)
+    const result = scanQrCode(imageData.data, imageData.width, imageData.height)
+    if (result) {
+      return result
```
WofWca, line 104: Nitpick: this function is only used in one place, and only the `data` property of this return value is used. I think it'd be simpler to make it return `result.data` only, so there is no need to check `typeof` of the return value if this function.

packages/frontend/src/components/QrReader/index.tsx
```diff
@@ -129,25 +158,40 @@ export default function QrReader({ onError, onScan }: Props) {
...
+    () => {
+      return {
+        pasteFromClipboard() {
+          handlePasteFromClipboard()
+        },
       }
     },
-    [onScan]
+    [handlePasteFromClipboard]
   )
```
WofWca, line 195: Nitpick: given how little `handlePasteFromClipboard` does (especially if you take my other commets into account), I personally wouldn't make it an imperative handle. Instead I'd just completely separate it from the `QrReader` component.

WofWca (changes requested):
With my non-nitpick comments addressed or dismissed, it's good to merge.

I tested it a bit, and looked at the code. It seems fine otherwise.
Though NGL, it's a bit hard to review, with refactoring changes and functional changes being mixed.

packages/frontend/src/components/QrReader/index.tsx
nicodh: That's true - it is a leftover of the previous handleScanResult which I don't know of why it was neccessary (see my question to simon). There the result was stripped down to result.data

I will reduce that...

packages/frontend/src/components/dialogs/QrCode.tsx
nicodh: if processQr is successfull it calls onDone which sets processingQrCode.current to false, but itf it fails it has to be set in the catch clause.

So both is correct, but I agree it is confusing to have it one time like this and one time like that

packages/frontend/src/components/QrReader/index.tsx
nicodh: onScan has its own error handling anyway so this is not needed

nicodh:
@r10s since I refactor this now anyway: when do we show the red alert (in our code is named "userFeedback") and when a normal warning dialog?
Since when a dialog is opened, the red message is behind the backdrop which is not so nice:

image

the "AlertDialog" would look like that:

image

which is not really what I would expect from a "AlertDialog" (butb that is probably another issue)

nicodh:
When scanning a qr code for backup import there are even 2 backdrops over the user feedback, so it is almost invisible

image

WofWca:
FYI that `user-feedback` toast is ancient. Last time it was changed was in 2019: b0134fef743e077a5af3f63174fa836d8ac12c24.

packages/frontend/src/components/dialogs/QrCode.tsx
nicodh: Moved it back

packages/frontend/src/components/QrReader/index.tsx
nicodh: I pass the error now always to onError so the parent component shows an AlertDialog

WofWca (approved):
So, there are a lot of changes in this MR, both functional and refactoring. Not all are strictly related to the title of the MR,

I have verified that the moved `fileToBase64` and `base64ToImageData` functions are unchanged. `qrCodeFromClipboard` was moved from `handlePasteFromClipboard` and it's also largely unchanged. The only thing different is that it errors out when the scan result's string is empty, which probably makes sense (but I'm not sure if this can happen).

About the removal of `setError(true)` from `handleError`, and the rename of it to `setCameraAccessError`. I haven't checked what it does exactly.

But what I noticed is that we now handle all scanning errors in one place, in the parent component, which is probably good. But we also invoke the callback after an error is thrown in `worker.postMessage(imageData)`, and `postMessage` gets called every 250ms. But I guess this can't happen often.

Another things is as was said above, now an `AlertDialog` is used instead of the `userFeedback` toast, which is better (at least in common cases).

As to the rest of the changes (that is the "add 'paste' button"): they look good, apart from that one suggestion I left!

nicodh:
> setCameraAccessError

this error is set if startStream fails so I named it a bit more meaningful

I think we won't merge this into the current release (which will be published tomorrow hopefully)

packages/frontend/src/components/QrReader/index.tsx
```diff
@@ -158,47 +82,18 @@ export default function QrReader({ onError, onScan }: Props) {
       } else {
         onError(error.toString())
       }
-
-      setError(true)
```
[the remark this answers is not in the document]
nicodh: Yes but that resulted in showing the camera_access_failed error in the UI which I consider wrong since it's a generic errorHandler (for example after an error in handleFileInputChange)

PR #4313, opened 2024-11-04
style: image attachments not being centered

WofWca:
- **refactor: DRY: add a CSS variable**
- **style: image attachments not being centered**


nicodh:
So this MR is just about restoring the previous situation? I think there are more issues with images. I don't think the hard coded 200px height make sense. Or do you want to address that in a different PR?

nicodh (commented):
Hmm, is adding a padding to the container and stretching the image with negative margins really the only way to achieve the previous state?

To understand: what  was changed and why in the mentioned PR?

nicodh:
Aah - I see: the idea behind is, to have no margins around image only messages...

WofWca:
> To understand: what was changed and why in the mentioned PR?

What 260fdd2839100872c9f60ad9db79aaf8d2b07805 did is it changed the container from `div` to `button`, and the said container stopped filling the message bubble because of this. This MR makes sure it _does_ fill the parent.

I did not introduce the negative margin thing, and I don't think it's worth to do a major refactor since we're approaching a stable release.

WofWca:
> I noticed that videos are shifted to the left when playing in the media library.

Could you make a screenshot? I'm not sure what you're talking about.

nicodh (approved):

WofWca:
@Raiden-GH Thanks! Looks like a separate issue. We'll take a look.

Issue #4293, opened 2024-10-30
Move webxdc-realtime setting out of experimental


Hocuri:
Similar to https://github.com/deltachat/deltachat-ios/pull/2360 and https://github.com/deltachat/deltachat-android/pull/3398, the "Realtime Webxdc Channels" setting should be moved from an experimental setting to a "normal" advanced setting. The default is bound to change then: https://github.com/deltachat/deltachat-core-rust/pull/6125

PR #4316, opened 2024-11-05
improvement: rearrange "realtime webxdc" setting

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/4293


r10s (approved):
very nice!

the translations need to  be pulled before the next release, so that the updated string, that also mention the IP address, get in

nicodh (approved):

PR #4323, opened 2024-11-08
Prepare release v1.48

nicodh:


nicodh:
> the local offline help will get a downgrade by this PR - probably as `./bin/create-local-help.sh` was executed while having `../deltachat-pages/` checked out in an old state - you need to pull there (or, add that to the script, maybe also check if `main` is checked out)
> 
I was in fresh pulled state in deltachat-pages #86291b98bea add more redirects

Is there another script to be executed in pages before copying to desktop?

nicodh:
I see the wrong chages now. I will investigate whats wrong...

r10s:
> I was in fresh pulled state in deltachat-pages #86291b98bea add more redirects
>
> Is there another script to be executed in pages before copying to desktop?

you're right, `./bin/create-local-help.sh` creates the offline, assuming  HTML-files from `../deltachat-pages/_site` are up to date. they are created using `bundle exec jekyll build` - which requires some stuff on your computer.

to move forward now - i can do a separate PR if you like, things are working for me locally

nicodh:
> i can do that in a separate PR if you like, things are working for me locally

Yes that would help, at least if we  want the newest help pages in desktop

r10s:
the last update is only a few weeks ago, but i created https://github.com/deltachat/deltachat-desktop/pull/4325 , when we're anyways on that :)

r10s (approved):
yeah, much less code also now :)

PR #4327, opened 2024-11-08
fix: "disappearing messages" dialog wrong value

WofWca:
The bug was introduced in https://github.com/deltachat/deltachat-desktop/commit/0e0d0b80d598802ebb972c8d6c05b593bd896f24,
where we switched from the Blueprint's `RadioGroup` to our
custom `RadioGroup`, which had the bug where it would not
reflect the actual value if it was changed programmatically.
This commit fixes that bug.

The bug in `Radio` existed ever since its introduction.
The reason for using `defaultChecked` instead of `checked`
probably was the fact that React would complain that
"input doesn't have `onChange` listener set,
use `defaultChecked` instead".

I have ligtly checked that "disappearing messages"
is the only affected place.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ## Changed
 
 ## Fixed
+- "Disappearing Messages" dialog not reflecting the actual current value #4327
 
 <a id="1_48_0"></a>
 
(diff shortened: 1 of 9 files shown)


WofWca:
Hold up, this prints warnings now

Simon-Laux (approved):
code LGTM, didn't test

PR #4340, opened 2024-11-14
Do not clear the draft if sending failed

link2xt:


WofWca:
This change makes sense to me. textareaRef is disabled until the message has been sent and the draft cleared.
Have you checked if the behavior is alright if `sendMessagePromise` takes some time to resolve?
Does it feel sluggish?

nicodh (approved):

nicodh:
> Have you checked if the behavior is alright if sendMessagePromise takes some time to resolve?

I think we can merge this, it is very theoretical "sendMessagePromise" is fulfilled when the draft is in database. It's not waiting fpor network (I tested it)

PR #4342, opened 2024-11-14
fix: always allow to edit the draft if it is not empty

link2xt:
Otherwise if you leave the group or get removed from the group, the draft stays forever and is shown in summary,
but it is not possible to edit it or get rid of it.

This builds on top of #4340 because there is a non-working "send" button.

### DIFF
--- a/packages/frontend/src/components/composer/Composer.tsx
+++ b/packages/frontend/src/components/composer/Composer.tsx
@@ -325,7 +325,13 @@ const Composer = forwardRef<
         </button>
       </div>
     )
-  } else if (!selectedChat.canSend) {
+  } else if (
+    !selectedChat.canSend &&
+    draftState.text === '' &&
+    draftState.file === null &&
+    draftState.quote === null &&
+    draftState.vcardContact === null
+  ) {
     return null
   } else {
     return (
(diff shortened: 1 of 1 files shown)


r10s:
> Composer is displayed without the send button if you cannot send but there is a draft

the image looks unexpected and like a bug. 

we're about worsening UX here.

anyways, discussion is mainly at https://github.com/deltachat/deltachat-android/issues/3428

r10s:
exchanging one bug by the other :)

but, why not save some money and use up the old bug first, known at least from deltachat/signal/whatsapp 🤷‍♂️

but yeah, all points have been made already :)

adbenitez:
> After removing the draft, composer disappears.

that is an unexpected UX, things appearing and disappearing without apparent reason (yes for the user it is harder to understand that it was because the draft and they would say "mmmhhh I swear there was an input bar here last time I entered this chat..." or "sometimes the send button is missing")

I agree with @r10s  that this just makes UI/UX worse, it is better to discuss first the issue before jumping into implementation on every UI

iequidoo:
For me a bad UX is that i can't send messages if i'm not a member. I'd expect that i'm not subscribed to messages from others, but i still can send and only get replies to my messages :) Basically, an app should be permissive

packages/frontend/src/components/composer/Composer.tsx
```diff
@@ -325,7 +325,13 @@ const Composer = forwardRef<
         </button>
       </div>
     )
-  } else if (!selectedChat.canSend) {
+  } else if (
+    !selectedChat.canSend &&
+    draftState.text === '' &&
+    draftState.file === null &&
+    draftState.quote === null &&
+    draftState.vcardContact === null
```
WofWca, line 333:
```suggestion
    !selectedChat.canSend &&
    // The composer should still be accessible if there is a draft,
    // even if `!canSend`.
    // Firstly, to be able to remove it, secondly to be able to copy it.
    // See https://github.com/deltachat/deltachat-android/issues/3428
    draftState.text === '' &&
    draftState.file === null &&
    draftState.quote === null &&
    draftState.vcardContact === null
```

WofWca (approved):
If it really is the case that core allows a draft to exist even when you can't send, then I think the draft should be accessible.
Maybe it's not the best way to achieve this, but I'd say this an improvement.

PR #4351, opened 2024-11-23
hardening: disable `IsolateSandboxedIframes`

WofWca:
This feature has been recently enabled by default in Chromium,
so let's toggle it back.


nicodh (approved):
Thanks for caring!

PR #4353, opened 2024-11-23
Change blocked contact profile view

nicodh:
This is how the profile of a blocked user looks now:

image

There is a new icon for blocked users in list view:
image

WIP: The ViewGroup component is not updated yet, so this is a draft only.


nicodh:
related to #4046 

alternative solution for PR #4189

packages/frontend/src/components/dialogs/ViewGroup.tsx
```diff
@@ -132,6 +132,18 @@ function ViewGroupInner(
     setGroupImage,
   } = useGroup(accountId, chat)
 
+  useEffect(() => {
```
nicodh, line 135: This should not be neccessary, since the chat passed as prop to ViewGroup should be updated but it isn't!

MainScreen -> ChatHeading(chat) // chat.contacts updated after blocking/unblocking!!

useOpenViewGroupDialog -> openViewProfileDialog(chat)

useDialog -> openDialog(ViewGroup, chat)

DialogContext -> createElement(ViewGroup, ...additionalProps)

ViewGroup(chat)

ViewGroupInner(chat)  chat.contacts not updated!!

packages/frontend/src/components/dialogs/ViewGroup.tsx
```diff
@@ -132,6 +132,18 @@ function ViewGroupInner(
...
+    onDCEvent(accountId, 'ContactsChanged', () => {
```
WofWca, line 136:
```suggestion
    return onDCEvent(accountId, 'ContactsChanged', () => {
```

The event listerner needs to be cleaned up.

packages/frontend/src/components/dialogs/ViewGroup.tsx
```diff
@@ -132,6 +132,18 @@ function ViewGroupInner(
...
+      BackendRemote.rpc
+        .getContactsByIds(accountId, group.contactIds)
+        .then(contacts => {
+          // update contacts in case a contact changed
+          // while this dialog is open (e.g. contact got blocked)
+          group.contacts = Object.values(contacts)
```
WofWca, line 142:
```suggestion
          // while this dialog is open (e.g. contact got blocked)
          // Yes, this mutates the prop. TODO refactor,
          // at least with `useState(initialGroup)`, as in `ViewProfile`.
          group.contacts = Object.values(contacts)
```

WofWca (changes requested):
That `useEffect` is super ugly, but I can live with it.
Works for now!
And otherwise looks good. Can be merged if the `onDCEvent` suggestion gets applied.

WofWca (approved):

PR #4356, opened 2024-11-25
Keep line breaks in quoted messages

### LINKED ISSUE #4345: Adding second device fails - "last error not set"
- Operating System (Linux/Mac/Windows/iOS/Android): Win11
- Delta Chat Version: 1.48.0
- Expected behavior: Adding Win PC as second device (main device is an android phone)
- Actual behavior: after QR scan the process of adding starts on both devices. After 30 seconds the android phone gives an error message: "Sending Error: <last error not set>
- Steps to reproduce the problem: I repeated the process several times, firewall is deactivated for delta chat, manual addition via backup file of the delta profile fails as well
- Screenshots: ---
- Logs:

### LINKED ISSUE #4354: lineends are lost in quotes
when quoting a message, the lineends are lost.

this is esp. worse in some statistics, but also in general. iirc, that used to be correct in the past

still, only few lines should be shown and the quote should be truncated.

Screenshot 2024-11-23 at 18 46 59

nicodh:
resolves #4354

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - "Disappearing Messages" dialog not reflecting the actual current value #4327
 - accessibility: make settings keyboard-navigable #4319
 - Fix documentation for --allow-unsafe-core-replacement #4341
+- fix missing linebreaks in quotes #4345
 
 <a id="1_48_0"></a>
 
--- a/packages/frontend/src/components/message/Message.tsx
+++ b/packages/frontend/src/components/message/Message.tsx
@@ -806,6 +806,7 @@ export const Quote = ({
               }
               disableJumbomoji
               nonInteractiveContent
+              isQuote={true}
             />
           </div>
         </div>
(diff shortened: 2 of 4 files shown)


WofWca (changes requested):
The commit message references wrong the issue. Needs to be #4354.

I didn't test this yet, but is it not possible to achieve with the `white-space` CSS property?

nicodh:
> The commit message references wrong the issue. Needs to be #4354.
It IS #4354 - maybe you mixed it up with the PR number #4356 ?

> I didn't test this yet, but is it not possible to achieve with the `white-space` CSS property?
No because without the conversion of Linebreaks to <span>{\n}</span> there is no linebreak white-space

WofWca:
> It IS https://github.com/deltachat/deltachat-desktop/issues/4354 - maybe you mixed it up with the PR number https://github.com/deltachat/deltachat-desktop/pull/4356 ?

Sorry, I didn't mention that I edited MR description, but the _commit message_ still references the wrong issue.

nicodh:
Ok, fixed the commit message

nicodh:
No need to use different functions here

PR #4361, opened 2024-11-27
fix: arrow-key navigation behavior in search

WofWca:
The "contacts" section items were not included
in the roving tabindex winget: they all had individual tab stops,
which is very confusing.

Before:

https://github.com/user-attachments/assets/05f11ce0-c7a2-44c0-b698-557f9b624e1c

After (it is not clear, but I have to press Tab to get to the "Contacts" section):

https://github.com/user-attachments/assets/60cef514-5fb4-433a-8043-adc8e720fc6e

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [Unreleased][unreleased]
 
 ## Added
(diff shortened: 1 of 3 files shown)


nicodh (approved):
Reviewed by testing & reading

PR #4369, opened 2024-11-28
improvement: add arrow-key nav everywhere

WofWca:
I have shallowly tested all the places where added it.

This completes arrow-key navigation for all the chat lists and contact lists, unless I missed something.

Follow up to #4224, #4291, #4361, #4362.

For easier review, hide white-space changes.  
This MR includes one small refactor commit, otherwise it's just boilerplate.

And about boilerplate: please tell me whether in general this looks like a good way to write this, or if this feels wrong and unmaintainable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [Unreleased][unreleased]
 
 ## Added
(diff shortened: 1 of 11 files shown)


nicodh (approved):
I didn't test all places, but compared the changes with --ignore-space-change 

LGTM!

Issue #4366, opened 2024-11-27
add webxdc selfAddr/notify/href/ratelimit APIs (M2/M3)


r10s:
the following task are open to complete M2/M3 proposal desktop-sided:

- [x] change `window.webxdc.selfAddr` to "self_addr" property returned by [dc_msg_get_webxdc_info()](https://c.delta.chat/classdc__msg__t.html#aee526949983962ed25adf33a2f5e6878)

- [x] add and set `window.webxdc.sendUpdateInterval` to "send_update_interval" property returned by [dc_msg_get_webxdc_info()](https://c.delta.chat/classdc__msg__t.html#aee526949983962ed25adf33a2f5e6878)

- [x] add and set `window.webxdc.sendUpdateMaxSize` to "send_update_max_size" property returned by [dc_msg_get_webxdc_info()](https://c.delta.chat/classdc__msg__t.html#aee526949983962ed25adf33a2f5e6878)

- [x] when tapping an webxdc-info-message ([dc_msg_get_info_type()](https://c.delta.chat/classdc__msg__t.html#ab3eeb59143e67fda31d3648bd775f77b) returns DC_INFO_WEBXDC_INFO_MESSAGE), open the webxdc-app directly, do not move up (instead, the app menu should provide "Show in Chat, cmp https://github.com/deltachat/deltachat-android/pull/3447, but that can also be done later)

- [x] when opening the webxdc-app from a webxdc-info-message, check [dc_msg_get_webxdc_href()]( https://c.delta.chat/classdc__msg__t.html#a61a8b635ee99c345d3da22fc7e635aa7). if set, open the app with `sth://base/HREF` instead of `sth://base/index.html`  
if the app is already open, make sure, the location is changed (probably the same way), in any case, JavaScript's location properties should be set appropriate and [hashchange](https://developer.mozilla.org/en-US/docs/Web/API/Window/hashchange_event) should be fired as needed

- [ ] listen to event [`DC_EVENT_INCOMING_WEBXDC_NOTIFY`](https://c.delta.chat/group__DC__EVENT.html#ga84b3259a78215ea622b6943b376675ca). on receiving, add a notification with text set to "APPNAME: TEXT"; TEXT comes from the event.  note, that the msg_id from the event may refer to a webxdc-instance _or_ a webxdc-info-message (in the latter case, use get_parent() to get the instance)

jsonrpc types should be similar; if sth. is missing there, this is another task.

[...]

PR #4373, opened 2024-11-29
Update core 1.151.2

nicodh:


packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -493,7 +493,7 @@ If you think that's a bug and you need that permission, then please open an issu
           accountId,
           msgId,
           update,
-          description
+          ''
         )
```
nicodh, line 497: It's a bit strange to set the parameter as deprecated, while at the same time we have to pass it to core to avoid an error there...

Should at least be an optional parameter in the type definitions in jsonrpc

Simon-Laux: > Should at least be an optional parameter in the type definitions in jsonrpc

maybe jsonrpc was forgotten to be updated, again.

Simon-Laux (approved):

pnpm-workspace.yaml
```diff
@@ -2,8 +2,8 @@ packages:
   - packages/*
 catalog:
   # deltachat specific
-  '@deltachat/jsonrpc-client': v1.151.1
-  '@deltachat/stdio-rpc-server': v1.151.1
+  '@deltachat/jsonrpc-client': v1.151.2
+  '@deltachat/stdio-rpc-server': v1.151.2
   webxdc-types: ^1.0.1
```
Simon-Laux, line 7: might make sense to also update those webxdc types. the package was renamed to `@webxdc/types`

https://github.com/webxdc/webxdc-types

packages/target-electron/src/deltachat/webxdc.ts
nicodh: Maybe I add this Option, since I'm adapting jsonrpc right now anyway, since it still misses `dc_msg_get_webxdc_href()` which is needed for https://github.com/deltachat/deltachat-desktop/issues/4366

packages/target-electron/static/webxdc-preload.js
```diff
@@ -138,12 +138,14 @@ class RealtimeListener {
...
-        JSON.stringify(update),
-        description
-      ),
+    sendUpdate: (update, description) => {
+      if (description) {
+        console.error(
+          'parameter description in sendUpdate is deprecated and will be removed in the future'
+        )
+      }
+      ipcRenderer.invoke('webxdc.sendUpdate', JSON.stringify(update))
```
WofWca, line 147: Don't we want to return this promise?

nicodh: This was not changed in this PR

WofWca: Umm, I don't think so.

```javascript
(update, description) =>
  ipcRenderer.invoke(
    'webxdc.sendUpdate',
    JSON.stringify(update),
    description
  )
```

returns a promise.

Simon-Laux: It was changed in this pr, but the type in types definition is void, so in that sense it is correct.

You should normally expect to wait for it to be sent out, because you get you own updates in the listener.
But internally it is async and if we don't await on it, possibly errors will always be uncaught errors, currently.

This might be a specification issue in error handling then: if we say that sendUpdate can fail in core and it should report errors.

pnpm-workspace.yaml
Simon-Laux: So because of recent type confusion, I now believe this should be done in this pr.

PR #4375, opened 2024-11-30
style: "Search in ..." divider overflow

WofWca:
Before:

image

After:

image

Reported by @adbenitez


Simon-Laux:
To keep in mind: For this search in chat divider it does not matter, but height might matter for the other dividers in the global search.

Simon-Laux (approved):

WofWca:
> but height might matter for the other dividers in the global search.

Those should not wrap because the only show "999+ messages / contacts / chats", unless the translation is super long.

PR #4380, opened 2024-12-02
New webxdc api

nicodh:
resolves (partly) to #4366 

Main changes: 
 - openWebxdc in frontend expects a message as first argument instead a messageId
 - selfAddr in webxdcInfo holds not the users address anymore
 - WebxdcInfo message has a new optional property webxdc_href
 - click on WebxdcInfo message directly opens the map and sets a event related location.url in iFrame (optional)

### DIFF
--- a/packages/frontend/src/components/message/Message.tsx
+++ b/packages/frontend/src/components/message/Message.tsx
@@ -835,7 +835,6 @@ function WebxdcMessageContent({ message }: { message: T.Message }) {
     name: 'INFO MISSING!',
     document: undefined,
     summary: 'INFO MISSING!',
-    internetAccess: false,
   }
 
   return (
@@ -853,12 +852,6 @@ function WebxdcMessageContent({ message }: { message: T.Message }) {
         {truncateText(info.name, 42)}
       </div>
       <div>{info.summary}</div>
-      {info.internetAccess && (
-        <div className='experimental'>
-          <b>EXPERIMENTAL</b> Webxdc that has full internet access, be careful!
-          (only works in saved messages)
-        </div>
-      )}
       <Button
         className={styles.startWebxdcButton}
         styling='primary'
--- a/packages/frontend/src/components/message/Message.tsx
+++ b/packages/frontend/src/components/message/Message.tsx
@@ -358,7 +358,6 @@ export default function Message(props: {
   const privateReply = usePrivateReply()
   const { openContextMenu } = useContext(ContextMenuContext)
   const openViewProfileDialog = useOpenViewProfileDialog()
-  const { jumpToMessage } = useMessage()
   const { joinVideoChat } = useVideoChat()
 
   const showContextMenu = useCallback(
@@ -459,15 +458,9 @@ export default function Message(props: {
     let onClick
     if (isInteractive) {
       onClick = async () => {
-        if (isWebxdcInfo && message.parentId) {
-          jumpToMessage({
-            accountId,
-            msgId: message.parentId,
-            msgChatId: undefined,
-            highlight: true,
-            msgParentId: message.id,
-            scrollIntoViewArg: { block: 'center' },
-          })
+        if (isWebxdcInfo) {
+          // open or focus the webxdc app
+          openWebxdc(message)
         } else if (isProtectionBrokenMsg) {
           const { name } = await BackendRemote.rpc.getBasicChatInfo(
             selectedAccountId(),
@@ -850,7 +843,7 @@ function WebxdcMessageContent({ message }: { message: T.Message }) {
       <img
         src={runtime.getWebxdcIconURL(selectedAccountId(), message.id)}
         alt={`icon of ${info.name}`}
-        onClick={() => openWebxdc(message.id)}
+        onClick={() => openWebxdc(message)}
       />
       <div
         className='name'
@@ -869,7 +862,7 @@ function WebxdcMessageContent({ message }: { message: T.Message }) {
       <Button
         className={styles.startWebxdcButton}
         styling='primary'
-        onClick={() => openWebxdc(message.id)}
+        onClick={() => openWebxdc(message)}
       >
         {tx('start_app')}
       </Button>
--- a/packages/target-electron/src/deltachat/webxdc.ts
+++ b/packages/target-electron/src/deltachat/webxdc.ts
@@ -94,7 +94,8 @@ export default class DCWebxdc extends SplitOut {
       msg_id: number,
       p: DcOpenWebxdcParameters
     ) => {
-      const { webxdcInfo, chatName, displayname, addr, accountId } = p
+      const { webxdcInfo, chatName, displayname, accountId } = p
+      const addr = webxdcInfo.selfAddr || ''
       if (open_apps[`${accountId}.${msg_id}`]) {
         log.warn(
           'webxdc instance for this app is already open, trying to focus it',
@@ -115,7 +116,6 @@ export default class DCWebxdc extends SplitOut {
         await this.rpc.getWebxdcBlob(accountId, msg_id, icon),
         'base64'
       )
-
       const ses = sessionFromAccountId(accountId)
       const appURL = `webxdc://${accountId}.${msg_id}.webxdc`
 
--- a/packages/target-electron/src/deltachat/webxdc.ts
+++ b/packages/target-electron/src/deltachat/webxdc.ts
@@ -99,27 +99,16 @@ export default class DCWebxdc extends SplitOut {
       p: DcOpenWebxdcParameters
     ) => {
       const { webxdcInfo, chatName, displayname, accountId, href } = p
-      const removeLeadingSlash = (s: string) => {
-        if (s.startsWith('/')) {
-          return s.substring(1)
-        }
-        return s
-      }
       const addr = webxdcInfo.selfAddr
       let base64EncodedHref = ''
       const appURL = `webxdc://${accountId}.${msg_id}.webxdc`
       if (href && href !== '') {
         // href is user provided content, so we want to be sure it's relative
-        if (
-          new URL(href, 'http://dummy.local').origin !== 'http://dummy.local'
-        ) {
-          log.warn('href is not relative, ignoring', { href })
-        } else {
-          // make href eval safe
-          base64EncodedHref = Buffer.from(
-            appURL + '/' + removeLeadingSlash(href)
-          ).toString('base64')
-        }
+        // relative href needs a base to construct URL
+        const url = new URL(href, 'http://dummy')
+        const relativeUrl = url.pathname + url.search + url.hash
+        // make href eval safe
+        base64EncodedHref = Buffer.from(appURL + relativeUrl).toString('base64')
       }
       if (open_apps[`${accountId}.${msg_id}`]) {
         log.warn(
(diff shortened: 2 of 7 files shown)


packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -89,12 +89,25 @@ export default class DCWebxdc extends SplitOut {
...
-      const { webxdcInfo, chatName, displayname, addr, accountId } = p
+      const { webxdcInfo, chatName, displayname, accountId, href } = p
+      const addr = webxdcInfo.selfAddr || ''
+      let locationHref = ''
+      if (href && href !== '') {
+        const url = new URL(href, 'http://localhost')
+        // make sure url is relative and eval safe
+        locationHref = Buffer.from(
+          url.toString().replace(url.origin, '')
+        ).toString('base64')
```
WofWca, line 109: This looks fragile. I feel like it's better to manually construct the final URL.

As an example, `new URL('mailto:example@example.com', 'https://dummy.com').origin === 'null'`.

Am I understanding it correctly that `message.webxdcHref` is user-controlled, that is any user of the app, and the app itself, is free to set it to whatever string they want?
So, clicking a webxdc info message is equivalent to clicking a regular link, except that we have to sanitize the origin of the final URL?
What do you think of doing exactly that? Convert the URLs to absolute URLs and ensure that the origin is the same.

WofWca (commented):
I left a bunch of questions. Hate to be that guy, but this MR affects security-critical code.

Overall I think we need more comments. Clearly document function contracts and parameters, are they sanitized or not. If not, how they need to be sanitized. The assumptions, etc.

packages/target-electron/src/deltachat/webxdc.ts
nicodh: >As an example, new URL('mailto:example@example.com', 'https://dummy.com').origin === 'null'.

I assume 'mailto:example@example.com' is an example for a passed webxdc_href
this condition is always false, origin is never null - rather the URL constructor will throw an error. We could use that, to make sure  the passed string is relative, since it throws an error if the string is relative and no base is provided. https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

`url (in the case of absolute URLs) or base + url (in the case of relative references) is not a valid URL.`

But I consider the implementation sufficient: if the string contains  a absolute url, that will be the origin, if not, the provided base is the origin, but in any case we replace the origin with an empty string.

And not to forget, the app itself can set the locationUrl directly anyway, so I consider this change not as critical as displaying a link in a message, for example. And it is a SystemMessage, so can not be easily faked. The href can only be transferred from inside a shared webxdc app.

So if an app takes the hash of a webxdc_href and uses it to "generate" any url in the app running on other chat members device, I could also do whatever I want.

packages/target-electron/static/webxdc-preload.js
```diff
@@ -291,17 +298,29 @@ class RealtimeListener {
...
+     * called via webContents.executeJavaScript
+     */
+    setLocationUrl(href) {
+      locationUrl = Buffer.from(href, 'base64').toString('utf8')
+      if (locationUrl && locationUrl !== '' && connectionsFilled) {
+        // if connectionsFilled is false, the url is loaded after
+        // the connections are filled
+        window.frames[0].window.location = locationUrl
+      }
+    },
```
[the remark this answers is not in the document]
nicodh: As said above. setLocationUrl does nothing that the app could not do directly.

But yes, the function could also run in the wrapper. I will have a look

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -89,12 +89,25 @@ export default class DCWebxdc extends SplitOut {
...
+     */
     const openWebxdc = async (
       _ev: IpcMainInvokeEvent,
       msg_id: number,
       p: DcOpenWebxdcParameters
     ) => {
-      const { webxdcInfo, chatName, displayname, addr, accountId } = p
+      const { webxdcInfo, chatName, displayname, accountId, href } = p
+      const addr = webxdcInfo.selfAddr || ''
+      let locationHref = ''
```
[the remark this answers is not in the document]
nicodh: Agree on that. I will rename it

nicodh:
> Hate to be that guy,

No worries. I appreciate critical feedback! :-)

packages/target-electron/src/deltachat/webxdc.ts
WofWca: > I assume 'mailto:example@example.com' is an example for a passed webxdc_href

Yes.

> this condition is always false, origin is never null

It is `'null'` for the provided example, try putting it in the console. And so `.replace(url.origin, '')` will do nothing.

There are a lot of wonky URLs. `file:///ada/Analytical%20Engine/README.md`, `urn:isbn:9780307476463`. For these `origin` is also `'null'`.

WofWca: > And not to forget, the app itself can set the locationUrl directly anyway

This is not gonna be super on-topic, but we need to make a distinction between the app and a chat member. While an app itself might be secure, chat members can still set this to whatever they want.

WofWca: > So, we don't care about sanitizing the href parameter in any way then, right?

I would argue that we do, for the reason that I mentioned earlier: the app might be secure, but chat members still can't be trusted.
We probably wouldn't want to try to load a URL that has a different origin than the app. Even though a malicious app can still try to do this, IMO we'd want prevent malicious users from doing this.
What do you think?

And sorry for spamming the comments.

packages/target-electron/static/webxdc-preload.js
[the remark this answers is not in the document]
WofWca: > setLocationUrl does nothing that the app could not do directly.

Yes, right now it does not look like it does, but it unnecessarily lives in a dangerous area with access to Node functions and variables defined above.

packages/target-electron/src/deltachat/webxdc.ts
nicodh: Let's move this discussion to desktop channel

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -89,12 +89,27 @@ export default class DCWebxdc extends SplitOut {
...
+      const addr = webxdcInfo.selfAddr
+      let base64EncodedHref = ''
+      const appURL = `webxdc://${accountId}.${msg_id}.webxdc`
+      if (href && href !== '') {
+        // href is user provided content, so we want to be sure it's relative
+        // relative href needs a base to construct URL
+        const url = new URL(href, 'http://dummy')
+        const relativeUrl = url.pathname + url.search + url.hash
+        // make href eval safe
+        base64EncodedHref = Buffer.from(appURL + relativeUrl).toString('base64')
```
WofWca, line 111: The current implementation looks better!

Also consider the following. I might be too paranoid and ignorant to how URLs work, but this looks more reliable to me:

```suggestion
        const url = new URL(href, 'http://dummy')

        const funalUrl = new URL(appURL)
        funalUrl.pathname = url.pathname
        funalUrl.search = url.search
        funalUrl.hash = url.hash

        // make href eval safe
        base64EncodedHref = Buffer.from(funalUrl.toString()).toString('base64')
```

nicodh: Here we rely on URL constructor to make sure noone finds a way to cheat it, since we call finally finalUrl.toString()

It is strange that the finalUrl shows 'webxdc:' as protocol but url.protocol = 'webxdc:' does not change any existing protocol and 'wbxdc://something' is not accepted as base argument. So there is some magic in the URL constructor and if that is changed it would break our implementation.

The current implementation garantuees that the locations starts with appUrl so I consider that more safe

packages/target-electron/static/webxdc-preload.js
```diff
@@ -291,17 +298,29 @@ class RealtimeListener {
...
     },
+    /**
+     * called via webContents.executeJavaScript
+     */
+    setLocationUrl(base64EncodedHref) {
+      locationUrl = Buffer.from(base64EncodedHref, 'base64').toString('utf8')
+      if (locationUrl && locationUrl !== '' && connectionsFilled) {
+        // if connectionsFilled is false, the url is loaded after
+        // the connections are filled
+        window.frames[0].window.location = locationUrl
```
WofWca, line 321:
```suggestion
        document.getElementById('frame').contentWindow.location = locationUrl
```

This is more concrete. Feel free to ignore

packages/target-electron/src/deltachat/webxdc.ts
nicodh: Yes was talking about your proposal. 
This is where URL constructor is inconsistent:
``` 
u = new URL('http://test.de')
u.toString() => 'http://test.de/'
u.procol = 'ftp:'
u.toString() => 'ftp://test.de/'
u.protocol = 'webxdc:'
u.toString() => 'ftp://test.de/'

```

WofWca: Indeed very weird.
But still, my approach guarantees that URL is a valid URL, with all the parameters properly "escaped" (if that's a term we can apply to URLs), and the base being set to a value controlled by us.
If there really is weirdness with such custom protocols, I'd also expect to see this weirdness down the line, in `window.location = url`.

If this doesn't convince you, them let's move on and resolve this thread.

packages/target-electron/src/deltachat/webxdc.ts
WofWca: Continued in https://github.com/deltachat/deltachat-desktop/pull/4380#discussion_r1869083111

packages/target-electron/static/webxdc-preload.js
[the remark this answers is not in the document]
WofWca: As was discussed in private, implementing my suggestion requires rethinking the design, because of the fill 500 thing.
I can live with the current one, as long as we're careful with future changes.

packages/target-electron/src/deltachat/webxdc.ts
nicodh: >  is a valid URL, with all the parameters properly "escaped"

that is the case also for the current implementation (as all properties come from URL), but I feel more comfortable to have control what the "origin" of the location.href is, by settings it explicit as string

packages/target-electron/static/webxdc-preload.js
nicodh: Maybe in a later change, I would like to refactor some more stuff later :-)

packages/frontend/src/components/message/Message.tsx
```diff
@@ -860,16 +852,10 @@ function WebxdcMessageContent({ message }: { message: T.Message }) {
         {truncateText(info.name, 42)}
       </div>
       <div>{info.summary}</div>
-      {info.internetAccess && (
-        <div className='experimental'>
-          <b>EXPERIMENTAL</b> Webxdc that has full internet access, be careful!
-          (only works in saved messages)
-        </div>
-      )}
```
WofWca, line 868: So, is `internetAccess` only true for the map webxdc now? Do we never want this warning?

WofWca (approved):
I left a comment about the "internetAccess" warning. Please check it out and resolve as you wish.
Otherwise it's good, though ideally I'd like more comments, and get rid of `exposeInMainWorld` for `setLocationUrl`. But this can be done later.
IMO This is ready to be merged. (but FYI a CHANGELOG entry is missing)

packages/frontend/src/components/message/Message.tsx
nicodh: It is set to true by Core if a webxdc is an "integrated" webxdc app (like map). Other apps don't have internet_access any more.

PR #4381, opened 2024-12-02
Update core to v1.151.3

nicodh:


WofWca (commented):
So, when is this piece of code relevant now?

https://github.com/deltachat/deltachat-desktop/blob/78fef8b6b45329250c369dc23bd1c4bcd02adca5/packages/frontend/src/components/message/Message.tsx#L863-L868

The core changelog states
> Remove experimental request_internet_access option from webxdc's manifest.toml.

Otherwise looks like this can be merged, no blockers, though some features are not implemented yet and are to be addressed in #4380.

nicodh:
> So, when is this piece of code relevant now?

I will remove it when rebasing #4380, where I reflect the core update changes

PR #4383, opened 2024-12-02
Readd Date styles in gallery

### LINKED ISSUE #4358: date looks like a tab
the date in the "media view" looks like a tab in 1.48.0 the last update:

Screenshot 2024-11-26 at 12 29 21
<br>
<br>
<br>

in 1.46, probably before the blueprint changes, it was more clear that it is no tab, by right-align and gray color:

Screenshot 2024-11-26 at 12 28 57

we should restore the old layout, maybe even with a slightly smaller font size, as the  "search files" placeholder (which is is right-aligned correctly, btw)

nicodh:
resolves #4358

### DIFF
--- a/packages/frontend/scss/gallery/_media.scss
+++ b/packages/frontend/scss/gallery/_media.scss
@@ -86,6 +86,7 @@ $tab-height: 46px;
       border-bottom: 1px solid grey;
       background: transparent;
       color: var(--textPrimary);
+      outline: none;
     }
   }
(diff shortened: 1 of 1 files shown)


WofWca (approved):
I'd say the date shouldn't be a `tab-item`, and I'm not sure about removing the outline (for accessibility reasons), but this should the issue, and makes the code more proper.
Although I did not test.

PR #4384, opened 2024-12-04
add dc version to fatal errors for better error reports

Simon-Laux:


packages/target-electron/src/index.ts
```diff
@@ -149,9 +149,10 @@ Promise.all([
     log.critical('Fatal Error during init', error)
     dialog.showErrorBox(
       'Fatal Error during init',
-      '' +
-        error +
-        '\n\nAlso make sure you are not trying to run multiple instances of deltachat.'
+      `[DC Version: ${BuildInfo.VERSION}]
```
ralphtheninja, line 152: While we're at it, why not add the core version here as well?

nicodh (approved):
Reviewed by testing and reading

packages/target-electron/src/index.ts
Simon-Laux: I said "not so easy", not impossible. sure you could run the rpc server with `--version`, but it's again code that could fail. especially since most errors in this dialog were "could not find or start core/rpc-server" in the past (catching this scenario is the main purpose of these dialogs, core crashes are very rare these days).

PR #4385, opened 2024-12-04
fullscreen media view: improve drag behaviour and remove visible scrollbars

### LINKED ISSUE #4368: Full screen media window not draggable on Mac
In media full screen view the mac window is not draggable since draggable regions are lacking

Follow up of #4300

Simon-Laux:
- **fix drag area problem in fullscreen media view**
- **fix: remove visible scrollbars in fullscreen media view**

closes #4368 (I know we could do even better, but this invisible drag area where the navbar would be already fixes the bug in my eyes)

The scrollbar problem: (those scrollbars are just a visual bug, they don't really do anything as far as I understood it)
Bildschirmfoto 2024-12-04 um 15 48 47

The drag area:
| before| after|
|-------|-----|
|Bildschirmfoto 2024-12-04 um 15 51 01|Bildschirmfoto 2024-12-04 um 15 51 42|

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@
 - do not clear the draft if sending failed. #4340
 - "Search in \<chat name\>" divider overflowing for long chat names #4375
 - fix startup delay on linux #4379
-- fix: remove visible scrollbars in fullscreen media view 
+- fix: remove visible scrollbars in fullscreen media view #4385
 
 
 <a id="1_48_0"></a>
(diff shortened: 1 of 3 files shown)


WofWca (approved):
Did not test the draggable thing, but the scrollbars fix is sensible. Scrolling is actually performed by mouse drag, and zooming with scroll wheel.

nicodh (approved):
Tested on Mac

PR #4387, opened 2024-12-04
action to build appx release for ms-store-release

Simon-Laux:


nicodh (approved):
Compared it with our preview action
LGTM

We have to deploy it, to see if it works

PR #4390, opened 2024-12-05
E2e GitHub action

### LINKED ISSUE #3172: Fix e2e ci testing
- [ ] move to extra folder with own package.json
- [ ] update test cafe or migrate to cypress, whatever works

nicodh:
closes #3172


.github/workflows/e2e.yml
```diff
@@ -0,0 +1,57 @@
...
+
+
+jobs:
+  test:
+    name: E2E Tests
+    environment: test
+    env:
+      CI: true
+      NODE_ENV: test
+      PRIVATE_CERTIFICATE_KEY: ${{ secrets.PRIVATE_CERTIFICATE_KEY }}
```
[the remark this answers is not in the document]
nicodh: This are just "dummy" certificates. They have the correct format to be accepted by the server but then are not validated in the Browser (ignoreHTTPSErrors: true).
They are not generated but just read from github secrets.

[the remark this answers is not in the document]
Simon-Laux: what link2xt suggests would have the advantage of also working in forks, when outside contributors want to contribute. Also it would be easier to follow and understand if it would not depend on GitHub secrets.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,57 @@
...
+      PRIVATE_CERTIFICATE_CERT: ${{ secrets.PRIVATE_CERTIFICATE_CERT }}
+      VERSION_INFO_GIT_REF: ${{ github.ref }}
+      DC_ACCOUNTS_DIR: ../../e2e-tests/data/accounts
+    timeout-minutes: 60
+    runs-on: ${{ matrix.os }}
```
Simon-Laux, line 33: I doubt operating system is relevant for browser. more interesting might be different browsers. with electron it is only chrome, but with tauri it is chrome and webkit/safari

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,57 @@
...
+    - name: create data dir
+      run: mkdir -p packages/target-browser/data
+    - name: Run e2e tests
+      run: pnpm e2e --project Chrome
+    - uses: actions/upload-artifact@v4
+      if: ${{ !cancelled() }}
+      with:
+        name: e2e-report
+        path: packages/e2e-tests/playwright-report/ packages/e2e-tests/test-results/
+        retention-days: 30
```
Simon-Laux, line 57: does nothing currently:
Bildschirmfoto 2024-12-09 um 22 19 48

.github/workflows/e2e.yml
nicodh: we can pass no project parameter to pnpm e2e, then it runs on all configured browsers. right now, (Firefox, Chrome & Chromium)

nicodh: webkit might also work, didn't try yet

Simon-Laux: may be easiest to run safari on macOS runner?

nicodh: webkit does not allow clipboard read in my test. Needs further investigation
might be related to https://github.com/microsoft/playwright/issues/13037

Simon-Laux (approved):

PR #4394, opened 2024-12-08
Avoid drafts in chats where canSend is false

### LINKED ISSUE #4389: Pressing Ctrl+Up in read-only mailing lists sets a draft
Ctrl+Up selects next message for the quote, but this should not work if the chat is read only (can send = false).

This happens in chats with github notifications.

nicodh:
resolves #4389

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ## Fixed
 - handle double escape on Dialog #4365
+- avoid drafts in readonly chats #4349
 
 <a id="1_49_0"></a>
 
--- a/packages/frontend/src/components/composer/Composer.tsx
+++ b/packages/frontend/src/components/composer/Composer.tsx
@@ -444,7 +444,8 @@ export function useDraft(
 
   const loadDraft = useCallback(
     (chatId: number) => {
-      if (!canSend) {
+      if (chatId === null || !canSend) {
+        clearDraft()
         return
       }
       BackendRemote.rpc.getDraft(selectedAccountId(), chatId).then(newDraft => {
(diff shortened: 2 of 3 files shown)


nicodh:
No changelog entry since bug was introduced in test release

WofWca (commented):
I don't know if this is the right approach. The fact that `useDraft` returns `draftState` of the previous chat after you switch to a chat to which you cannot send, seems pretty wrong.

Maybe it would be simpler to just make the final return statement return dummy functions and objects instead. Hmmm, but then, the shortcut code is still active...

Or maybe just disable the shortcut and that's it, and wait for the composer refactor.

Also this doesn't actually disable the Ctrl + Up shortcut in `!canSend` chats. The messages still get highlighted.

> No changelog entry since bug was introduced in test release

I don't think so. You can utilize the Ctrl + Up shortcut in `!canSend` chats in 1.48.0.

nicodh:
> I don't know if this is the right approach. The fact that `useDraft` returns `draftState` of the previous chat after you switch to a chat to which you cannot send, seems pretty wrong.

In a readonly chat there should be no calls to setDraft or load draft at all. So if you prefer a minimal approach we could add the !canSend condition only to the shortcut in https://github.com/deltachat/deltachat-desktop/pull/4394/files#diff-ca83809f74f6819ef5c7d61676ac6c9abd27cb5becc46cb08bc9c43bc87c0f61R596

btw: it seems clearDraft is nowhere used outside of useDraft, so no need to return it.

WofWca:
> In a readonly chat there should be no calls to setDraft or load draft at all

There was this MR https://github.com/deltachat/deltachat-desktop/pull/4342 which suggested making the draft editable even when `!canSend`.
But if we decide to make the draft non-editable when `!canSend`, we should probably make `useDraft` return a value where these functions (`updateDraftText`, `removeQuote`) can be `null` or something like that, instead of silently doing nothing.

But maybe I'm being too pedantic.

> So if you prefer a minimal approach we could add the !canSend condition only to the shortcut

In this particular case yes, I'd prefer to be more conservative, but I won't insist.

nicodh:
Returning null would mean to add non null checks in many places. I also don't like that loadDraft is called even in !canSend chats. Maybe it makes sense to add another clearDraft() call in loadDraft before returning.

 >  instead of silently doing nothing
I would say: these functions should never be called, but if called they should not add a draft to a readOnly chat

WofWca:
> Returning null would mean to add non null checks in many places.

Well yes, that's the point. You don't want to call a function that you expect to do something but that actually does nothing.

> Maybe it makes sense to add another clearDraft() call in loadDraft before returning.

Yes, that's better than keeping the previous state of the draft. And `clearDraft()` doesn't clear it in core, so it's alright.

WofWca (approved):

PR #4395, opened 2024-12-08
Check text exists before accessing it

### LINKED ISSUE #4337: 1.48.0 crashes sometimes when i try to quote-reply
image

nicodh:
resolves #4337

### DIFF
--- a/packages/frontend/src/components/message/Message.tsx
+++ b/packages/frontend/src/components/message/Message.tsx
@@ -792,15 +792,18 @@ export const Quote = ({
               )}
             </>
           )}
-          <div className='quoted-text'>
-            <MessageBody
-              text={
-                quote.text.slice(0, 3000 /* limit quoted message size */) || ''
-              }
-              disableJumbomoji
-              nonInteractiveContent
-            />
-          </div>
+          {hasMessage && quote.text && (
+            <div className='quoted-text'>
+              <MessageBody
+                text={
+                  quote.text.slice(0, 3000 /* limit quoted message size */) ||
+                  ''
+                }
+                disableJumbomoji
+                nonInteractiveContent
+              />
+            </div>
+          )}
         </div>
         {hasMessage && quote.image && (
           <img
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/message/Message.tsx
```diff
@@ -792,15 +792,18 @@ export const Quote = ({
...
-          <div className='quoted-text'>
-            <MessageBody
-              text={
-                quote.text.slice(0, 3000 /* limit quoted message size */) || ''
-              }
-              disableJumbomoji
-              nonInteractiveContent
-            />
-          </div>
+          {hasMessage && quote.text && (
```
WofWca, line 795: Doesn't this hide the quote when `quote.kind === 'JustText'`? Because `hasMessage` is defined as `quote.kind === 'WithMessage'`.

And I think we need a comment for the `quote.text &&` check, at least link the issue.

nicodh: So now I check only for the text property to exist.

> And I think we need a comment for the quote.text && check, at least link the issue

I think it is self explaining to check for the text property to exist before calling "slice" on it.

WofWca: Well, the thing is that the types say that it always exists.

packages/frontend/src/components/message/Message.tsx
```diff
@@ -792,15 +792,18 @@ export const Quote = ({
...
-          <div className='quoted-text'>
-            <MessageBody
-              text={
-                quote.text.slice(0, 3000 /* limit quoted message size */) || ''
-              }
-              disableJumbomoji
-              nonInteractiveContent
-            />
-          </div>
+          {quote.text && (
```
WofWca, line 795:
```suggestion
          {/* `quote.text` might be `undefined` due to
          https://github.com/deltachat/deltachat-desktop/issues/4337#issuecomment-2473521298 */}
          {quote.text && (
```

WofWca (approved):

PR #4396, opened 2024-12-08
Handle Chrome double escape behaviour

### LINKED ISSUE #4365: cancelling profile creation with "Esc" not fully working
when cancelling profile creation with the "Back Arrow", the dialog is closed, the account in creation is removed, and one is back in the last profile - all as expected.

however, using the "Esc" key instead, the account in creation is not removed and one is not back in the last profile. also the dialog is only removed partly, the background images stays:

Screenshot 2024-11-27 at 20 49 17

weirdly, the dialog comes back when the window is resized

nicodh:
resolves #4365

Pressing escape key twice will close open dialogs in Chrome no matter how onCancel is configured https://issues.chromium.org/issues/346597066

To handle that we move the onClickBackButton function one component higher to add it as onClose prop to the Dialog and handle the event the same way a BackButton Click would be handled

TODO: check other dialogs how the behave on forced close (should go into a new Issue though)

### DIFF
--- a/packages/frontend/src/components/screens/WelcomeScreen/index.tsx
+++ b/packages/frontend/src/components/screens/WelcomeScreen/index.tsx
@@ -6,6 +6,12 @@ import InstantOnboardingScreen from './InstantOnboardingScreen'
 import OnboardingScreen from './OnboardingScreen'
 import useInstantOnboarding from '../../../hooks/useInstantOnboarding'
 import { getConfiguredAccounts } from '../../../backend/account'
+import { BackendRemote, EffectfulBackendActions } from '../../../backend-com'
+import useDialog from '../../../hooks/dialog/useDialog'
+import AlertDialog from '../../dialogs/AlertDialog'
+import { getLogger } from '@deltachat-desktop/shared/logger'
+
+const log = getLogger('renderer/components/screens/WelcomScreen')
 
 type Props = {
   selectedAccountId: number
@@ -20,6 +26,7 @@ export default function WelcomeScreen({ selectedAccountId, ...props }: Props) {
     startInstantOnboardingFlow,
   } = useInstantOnboarding()
   const [hasConfiguredAccounts, setHasConfiguredAccounts] = useState(false)
+  const { openDialog } = useDialog()
   const showBackButton = hasConfiguredAccounts
 
   useLayoutEffect(() => {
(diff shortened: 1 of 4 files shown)


nicodh:
Also tested with Instant Onboarding (closing the dialog when in profile creation process. (The rebuilded dialog looses the contactName but works after entering a profile name, so I left it as rare edge case)

packages/frontend/src/components/screens/WelcomeScreen/index.tsx
```diff
@@ -35,20 +42,52 @@ export default function WelcomeScreen({ selectedAccountId, ...props }: Props) {
...
+  }
+
   return (
     <ImageBackdrop variant='welcome'>
       <Dialog
         fixed
         width={400}
         canEscapeKeyClose={false}
         canOutsideClickClose={false}
+        onClose={onClickBackButton}
```
WofWca, line 82: Hmm but doesn't this get invoked also when the dialog gets closed normally, e.g. with "Submit"?

WofWca:
Does this MR handle all the `canEscapeKeyClose={false}` occurences?

And shouldn't we remove / rename that prop?

packages/frontend/src/components/screens/WelcomeScreen/index.tsx
nicodh: No it's not invoked. The dialog is replaced with openDialog(ConfigureProgressDialog) and that does not trigger the onClose of the initial Dialog. (which is or is not expected behaviour?)

It seems while the double escape triggers the onClose event, just removing the old dialog from DOM does not trigger it...

nicodh:
> Does this MR handle all the `canEscapeKeyClose={false}` occurences?
> 
No it doesn't (as mentioned in the MR description)

> should we rename the property

I would not, since once we have other browser with tauri they hopefully behave as expected

WofWca (approved):
OK, this improves things. I tested it with both no accounts and with other accounts present.

But I'd say we should keep that issue open, to handle other such dialogs, and to perhaps find a more thorough approach.

Because `canEscapeKeyClose` doesn't actually do what it advertises. So looks like we ought to always handle `onClose` or `onCancel` on all dialogs. But let's keep further discussions in #4365, or a follow-up issue, I suppose.

PR #4400, opened 2024-12-10
Add webxdc notification handler

nicodh:
related to #4366 
Example for calendar notification

image

click opens webxdc app with href if provided

Internal changes:
new flag isWebxdcInfo for DcNotification

### DIFF
--- a/packages/frontend/src/components/RuntimeAdapter.tsx
+++ b/packages/frontend/src/components/RuntimeAdapter.tsx
@@ -11,6 +11,8 @@ import useDialog from '../hooks/dialog/useDialog'
 import WebxdcSaveToChatDialog from './dialogs/WebxdcSendToChat'
 import { saveLastChatId } from '../backend/chat'
 import useChat from '../hooks/chat/useChat'
+import { internalOpenWebxdc } from '../system-integration/webxdc'
+import { BackendRemote } from '../backend-com'
 
 type Props = {
   accountId?: number
@@ -42,7 +44,12 @@ export default function RuntimeAdapter({
     }
 
     runtime.setNotificationCallback(
-      async ({ accountId: notificationAccountId, msgId, chatId }) => {
+      async ({
+        accountId: notificationAccountId,
+        msgId,
+        chatId,
+        isWebxdcInfo,
+      }) => {
         if (accountId !== notificationAccountId) {
           closeAllDialogs()
           // selectAccount always loads the last used chat
@@ -52,6 +59,10 @@ export default function RuntimeAdapter({
           await window.__selectAccount(notificationAccountId)
         } else if (chatId !== 0) {
           await selectChat(accountId, chatId)
+          if (isWebxdcInfo) {
+            const message = await BackendRemote.rpc.getMessage(accountId, msgId)
+            internalOpenWebxdc(accountId, message)
+          }
           clearNotificationsForChat(notificationAccountId, chatId)
         }
         if (msgId) {
--- a/packages/frontend/src/components/RuntimeAdapter.tsx
+++ b/packages/frontend/src/components/RuntimeAdapter.tsx
@@ -11,8 +11,6 @@ import useDialog from '../hooks/dialog/useDialog'
 import WebxdcSaveToChatDialog from './dialogs/WebxdcSendToChat'
 import { saveLastChatId } from '../backend/chat'
 import useChat from '../hooks/chat/useChat'
-import { internalOpenWebxdc } from '../system-integration/webxdc'
-import { BackendRemote } from '../backend-com'
 
 type Props = {
   accountId?: number
@@ -44,12 +42,7 @@ export default function RuntimeAdapter({
     }
 
     runtime.setNotificationCallback(
-      async ({
-        accountId: notificationAccountId,
-        msgId,
-        chatId,
-        isWebxdcInfo,
-      }) => {
+      async ({ accountId: notificationAccountId, msgId, chatId }) => {
         if (accountId !== notificationAccountId) {
           closeAllDialogs()
           // selectAccount always loads the last used chat
@@ -59,10 +52,6 @@ export default function RuntimeAdapter({
           await window.__selectAccount(notificationAccountId)
         } else if (chatId !== 0) {
           await selectChat(accountId, chatId)
-          if (isWebxdcInfo) {
-            const message = await BackendRemote.rpc.getMessage(accountId, msgId)
-            internalOpenWebxdc(accountId, message)
-          }
           clearNotificationsForChat(notificationAccountId, chatId)
         }
         if (msgId) {
--- a/packages/frontend/src/system-integration/notifications.ts
+++ b/packages/frontend/src/system-integration/notifications.ts
@@ -284,12 +284,18 @@ async function flushNotifications(accountId: number) {
   if (notifications.length > notificationLimit) {
     showGroupedNotification(accountId, notifications)
   } else {
-    for (const { chatId, messageId, isWebxdcInfo } of notifications) {
+    for (const {
+      chatId,
+      messageId,
+      isWebxdcInfo,
+      eventText,
+    } of notifications) {
       await showNotification(
         accountId,
         chatId,
         messageId,
-        isWebxdcInfo ?? false
+        isWebxdcInfo ?? false,
+        eventText
       )
     }
   }
--- a/packages/frontend/src/system-integration/notifications.ts
+++ b/packages/frontend/src/system-integration/notifications.ts
@@ -20,90 +20,16 @@ const log = getLogger('renderer/notifications')
  * - sends notifications to runtime (which invokes ipcBackend)
  */
 
-/**
- * Event handler queue to avoid race conditions when a new event
- * is triggered while another event is still being processed
- *
- * Mainly to avoid, that flushNotifications is called while
- * notifications are still in creation process
- */
-class EventHandlerQueue {
-  _pendingPromise: boolean
-  _eventHandler: {
-    handlerPromise: () => Promise<void>
-  }[] = []
-  constructor() {
-    this._pendingPromise = false
-  }
-
-  add(handler: (args: any) => Promise<void>, args: any) {
-    // wrap the handler in a promise and add it to the queue
-    const p = (): Promise<void> =>
-      new Promise(resolve => {
-        handler(args).then(res => {
-          log.debug('promise resolved!', handler.name)
-          resolve(res)
-        })
-      })
-    this.enqueue(p)
-  }
-
-  /**
-   * add an event handler to the queue
-   * and immediately try to dequeue again
-   */
-  private enqueue(handlerPromise: () => Promise<void>) {
-    this._eventHandler.push({ handlerPromise })
-    this.dequeue()
-  }
-
-  /**
-   * here the actual event handler is called
-   * if there is no other handler still processing
-   */
-  private async dequeue() {
-    if (this._pendingPromise) {
-      // already processing a handler
-      // do nothing (dequeue is called again in finally block)
-      return
-    }
-    const item = this._eventHandler.shift()
-    if (!item) {
-      return
-    }
-
-    try {
-      this._pendingPromise = true
-      // now the wrapped Promise holding the final handler is called
-      await item.handlerPromise()
-      this._pendingPromise = false
-    } catch (e) {
-      this._pendingPromise = false
-    } finally {
-      this.dequeue()
-    }
-  }
-}
-
-const queue = new EventHandlerQueue()
-
 export function initNotifications() {
   BackendRemote.on('IncomingMsg', (accountId, { chatId, msgId }) => {
-    log.debug('IncomingMsg received')
-    queue.add(incomingMessageHandler, {
-      accountId,
-      chatId,
-      msgId,
-      isWebxdcInfo: false,
-    })
+    incomingMessageHandler(accountId, chatId, msgId, false)
   })
   BackendRemote.on('IncomingWebxdcNotify', (accountId, { msgId, text }) => {
-    log.debug(`IncomingWebxdcNotify received ${text}`)
-    queue.add(incomingWebxdcEventHandler, { accountId, msgId, text })
+    // we don't have the chatId yet, but it will be fetched in flushNotifications
+    incomingMessageHandler(accountId, -1, msgId, true, text)
   })
   BackendRemote.on('IncomingMsgBunch', accountId => {
-    log.debug('IncomingMsgBunch received')
-    queue.add(flushNotifications, accountId)
+    flushNotifications(accountId)
   })
 }
 
@@ -113,7 +39,7 @@ function isMuted(accountId: number, chatId: number) {
 
 type queuedNotification = {
   chatId: number
-  msgId: number
+  messageId: number
   isWebxdcInfo: boolean
   eventText?: string
 }
@@ -122,16 +48,14 @@ let queuedNotifications: {
   [accountId: number]: queuedNotification[]
 } = {}
 
-async function incomingMessageHandler(arg: {
-  accountId: number
-  chatId: number
-  msgId: number
-  isWebxdcInfo: boolean
+function incomingMessageHandler(
+  accountId: number,
+  chatId: number,
+  messageId: number,
+  isWebxdcInfo: boolean,
   eventText?: string
-}) {
-  const { accountId, chatId, msgId, isWebxdcInfo, eventText } = arg
-
-  log.debug('incomingMessageHandler: ', { chatId, msgId })
+) {
+  log.debug('incomingMessageHandler: ', { chatId, messageId })
 
   if (
     SettingsStoreInstance.state &&
@@ -163,27 +87,8 @@ async function incomingMessageHandler(arg: {
   }
   queuedNotifications[accountId].push({
     chatId,
-    msgId,
+    messageId,
     isWebxdcInfo,
-    eventText,
-  })
-  return
-}
-
-async function incomingWebxdcEventHandler(arg: {
-  accountId: number
-  msgId: number
-  eventText: string
-}) {
-  const { accountId, msgId, eventText } = arg
-  // await new Promise(resolve => setTimeout(resolve, 2000)) // for testing
-  const message = await BackendRemote.rpc.getMessage(accountId, msgId)
-  const chatId = message.chatId
-  return incomingMessageHandler({
-    accountId,
-    chatId,
-    msgId,
-    isWebxdcInfo: true,
     eventText,
   })
 }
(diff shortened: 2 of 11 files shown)


packages/frontend/src/components/RuntimeAdapter.tsx
```diff
@@ -52,6 +59,10 @@ export default function RuntimeAdapter({
           await window.__selectAccount(notificationAccountId)
         } else if (chatId !== 0) {
           await selectChat(accountId, chatId)
+          if (isWebxdcInfo) {
+            const message = await BackendRemote.rpc.getMessage(accountId, msgId)
+            internalOpenWebxdc(accountId, message)
+          }
```
WofWca, line 65: It's unfortunate that we have to do an RPC call here, and then another one inside of `internalOpenWebxdc`.

Would be nice to optimize it at some point. Doesn't have to be done in this MR I guess.

WofWca, line 65: Is it intentional that we don't open the app if the notification is for a different account?

WofWca (commented):
The idea of the feature itself is great, and this MR seems like a good implementation of it code-wise. But design-wise:

I don't like the idea of opening the app on notification click, at least by default.
At the very least a notification should have a separate "Launch app" button to do it.

The "[How private are webxdc apps?](https://delta.chat/en/help#how-private-are-webxdc-apps)" section states:
> This also means: it can be a privacy risk to open webxdc apps in chats where you don’t trust the members - as you know it from e-mail attachments, where you only open attachments from senders you trust, and not from spammers.

So this is like opening an attachment right from a notification. Especially with real-time channels [now enabled by default](https://github.com/deltachat/deltachat-core-rust/pull/6125) (which [can leak your real IP](https://support.delta.chat/t/can-realtime-chat-be-disabled/3160?u=wofwca)). I expressed a similar concern in https://github.com/deltachat/deltachat-desktop/issues/4366#issuecomment-2505895256.  
And, correct me if I'm wrong, such notifications would be displayed even from people who aren't on your contact list? Never mind the contact list: with Chatmail, people share their invite codes, which bypasses the "contact request" thing.  
And the other person can even completely control the contents of the notification, the text and the icon, judging by the screenshot, which is powerful for phishing.

Exposing your real IP and system info to anyone on the internet at a click of a button is not great.

What do you think?

nicodh:
> Exposing your real IP and system info to anyone on the internet at a click of a button is not great.
> 
> What do you think?

Well it's not "anyone" and the information on the notification is clear. The icon comes from a webxdc app in a shared chat, not just from the message. 

I remember there was a discussion on that and assumed the DC_EVENT_INCOMING_WEBXDC_NOTIFY should work like that.
Maybe @r10s knows more. 

But we also could restrict the direct access to the case where the webxdc app is already opened.

r10s:
on android/ios we're opening the chat with the webxdc-info-message or the webxdc-app-instance instead of opening the app directly.

we thought about that, but then did not came to that at the end, idea was, however, to add an explicit "Start" button, but maybe it is also fine as it is now.

so, i would suggest to do the same on desktop, sorry for not being clearer in the issue.

that would postpone that discussion about privacy (where we should also keep in mind, that every normal looking link send in a normal message leaks the IP-address easier. and this does not even discuss the real-world-attack vector of an IP-Address, VPN and so on ;)

if a chat is a contact request, so "untrusted", i would not send a notification at all (tbh, not totally sure if that is already the case on android, but otherwise, that'd be a bug)

nicodh:
Ok so instead calling openwebxdc we just jump to the message as usual.

r10s:
yes, that seems to be the easier thing for now

WofWca:
> Well it's not "anyone" and the information on the notification is clear. The icon comes from a webxdc app in a shared chat, not just from the message.

Anyone can send an email, and they can make an app with whatever name and icon they like.

> if a chat is a contact request, so "untrusted"

In my experience, contacts don't get marked as "requests" if they have your invite link, and invite links are supposedly not considered sensitive info: For example, people post their invite links in https://deltachat-bot.github.io/public-bots/#/home

> i would suggest to do the same on desktop

Yes, that's better.

packages/frontend/src/components/RuntimeAdapter.tsx
WofWca: Decided to not open apps on notification click, so this will be resolved.

WofWca: Decided to not open apps on notification click, so this will be resolved.

nicodh: The second call is always needed if we have a click on a WebxdcInfoMessag  to get the webxdcInfo from the parent message. No way to get it it in one call

nicodh: Is obsolete now

packages/frontend/src/system-integration/notifications.ts
```diff
@@ -76,19 +110,62 @@ async function showNotification(
...
+      let icon: string | null = null
+      if (isWebxdcInfo) {
+        const relatedMessage = await BackendRemote.rpc.getMessage(
+          accountId,
+          messageId
+        )
+        if (
+          relatedMessage.systemMessageType === 'WebxdcInfoMessage' &&
+          relatedMessage.parentId
+        ) {
```
WofWca, line 130: https://github.com/deltachat/deltachat-desktop/issues/4366 states

> note, that the msg_id from the event may refer to a webxdc-instance or a webxdc-info-message

But this only handles the latter case, am I right?

packages/frontend/src/system-integration/notifications.ts
```diff
@@ -76,19 +110,62 @@ async function showNotification(
...
       messageId,
       accountId,
+      isWebxdcInfo,
     })
   } else {
     try {
-      const notificationInfo =
-        await BackendRemote.rpc.getMessageNotificationInfo(accountId, messageId)
-      const { chatName, summaryPrefix, summaryText } = notificationInfo
+      let chatName = ''
```
WofWca, line 117: `chatName` is empty for the new webxdc info messages. Does `getMessageNotificationInfo` not work for such messages?

packages/target-electron/static/webxdc-preload.js
```diff
@@ -247,12 +247,14 @@ class RealtimeListener {
...
   contextBridge.exposeInMainWorld('webxdc_internal', {
-    setup: (selfAddr, selfName) => {
+    setup: (selfAddr, selfName, sendUpdateInterval, sendUpdateMaxSize) => {
       if (is_ready) {
         return
       }
       api.selfAddr = Buffer.from(selfAddr, 'base64').toString('utf-8')
       api.selfName = Buffer.from(selfName, 'base64').toString('utf-8')
+      api.sendUpdateInterval = sendUpdateInterval ? sendUpdateInterval : 1000
+      api.sendUpdateMaxSize = sendUpdateMaxSize ? sendUpdateMaxSize : 18874368
```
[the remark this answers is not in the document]
nicodh: This are the values core is providing right now. Just used them as a reasonable fallback in case these values are empty for whatever reason. 
I just see that [here](https://c.delta.chat/classdc__msg__t.html#aee526949983962ed25adf33a2f5e6878) a different values recommended.  Anyway it's just a fallback
> And is it possible that sendUpdateInterval is nullish

Not AFAIK

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -222,7 +223,9 @@ export default class DCWebxdc extends SplitOut {
             // initializes the preload script, the actual implementation of `window.webxdc` is found there: static/webxdc-preload.js
             return makeResponse(
               Buffer.from(
-                `window.parent.webxdc_internal.setup("${selfAddr}","${displayName}")
+                `window.parent.webxdc_internal.setup("${selfAddr}","${displayName}", ${Number(
+                  sendUpdateInterval
+                )}, ${Number(sendUpdateMaxSize)})
```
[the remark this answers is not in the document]
nicodh: yes just to make sure they are numbers

packages/frontend/src/system-integration/notifications.ts
```diff
@@ -76,19 +110,62 @@ async function showNotification(
...
+            if (webxdcMessage.webxdcInfo.icon) {
+              const iconName = webxdcMessage.webxdcInfo.icon
+              const iconBlob = await BackendRemote.rpc.getWebxdcBlob(
+                accountId,
+                webxdcMessage.id,
+                iconName
+              )
+              // needed for valid dataUrl
+              const imageExtension = iconName.split('.').pop()
+              icon = `data:image/${imageExtension};base64, ${iconBlob}`
```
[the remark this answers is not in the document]
nicodh: That's how I saw it when converting an image - so I just copy pasted it. It definitely works...

packages/target-electron/static/webxdc-preload.js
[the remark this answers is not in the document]
WofWca: I'd leave a comment then stating that this should not happen.
Because I guess we wouldn't want to provide values that aren't actually the true values.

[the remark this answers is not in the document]
r10s: i think as well, it is fine as is

there is no case where these values are unset in case core's or cffi's `get_webxdc_info()` return successfully.

not sure about jsonrpc - but as that probably just calls  core's `get_webxdc_info()` i assume this is not possible as well.

still, probably good to have some default (which could also be `undefined`, [though](https://github.com/webxdc/website/pull/94/files#diff-08cf571a0d83f52c2a62868c9fafed34f35bac7e5f72b73da3c52bc6c898fd02R83), but not `0` or `-1` or so)

packages/frontend/src/system-integration/notifications.ts
nicodh: Hmm, that's true, so we have to call getMessageNotificationInfo in any case, at least just to get the chatName...

packages/target-electron/static/webxdc-preload.js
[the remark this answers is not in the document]
Simon-Laux: Adding an extra fallback is not needed as https://js.jsonrpc.delta.chat/types/T.WebxdcMessageInfo.html shows that it can not be `undefined` or `null`.

packages/frontend/src/system-integration/notifications.ts
```diff
@@ -202,8 +284,13 @@ async function flushNotifications(accountId: number) {
...
-    for (const { chatId, messageId } of notifications) {
-      await showNotification(accountId, chatId, messageId)
+    for (const { chatId, messageId, isWebxdcInfo } of notifications) {
+      await showNotification(
+        accountId,
+        chatId,
+        messageId,
+        isWebxdcInfo ?? false
+      )
     }
```
WofWca, line 294:
```suggestion
    for (const {
      chatId,
      messageId,
      isWebxdcInfo,
      eventText,
    } of notifications) {
      await showNotification(
        accountId,
        chatId,
        messageId,
        isWebxdcInfo ?? false,
        eventText
      )
    }
```

Otherwise it seems like. `eventText` is unused.

packages/target-browser/runtime-browser/runtime.ts
```diff
@@ -401,6 +402,7 @@ class BrowserRuntime implements Runtime {
       title,
       icon: notificationIcon,
```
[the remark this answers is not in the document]
WofWca: I pushed a commit that should handle it, please review.

WofWca (commented):
I did some testing. There seems to be a problem with when we call `flushNotifications()`, which is the only function that calls `showNotification()`. `flushNotifications()` is only called from the `IncomingMsgBunch` event handler.

Reproduction steps:
1. Create two accounts, and create a group with those two accounts.
2. Send the Calendar app to the group
3. Send a regular message to the group. You'll get a notification on another account.
4. Create an event in the calendar.
  _You won't get a notification_, but I assume you should.
6. Create another event in the calendar. 
  You will get a notification for the previous event!
7. Send a regular message.
  You'll first get a notification for the second calendar event, and then a notification for the regular message!

packages/frontend/src/system-integration/notifications.ts
nicodh: No it s required. I oversaw it, since it is included in the getMessageNotificationInfo if the event has a related WebxdcInfoMessage - I had no use case where an event is sent, but the related message is the webxdc message itself, but it seems that is possible and should be supported.

nicodh:
> I did some testing. There seems to be a problem with when we call `flushNotifications()`, which is the only function that calls `showNotification()`. `flushNotifications()` is only called from the `IncomingMsgBunch` event handler.

Well when debugging it seems that in the group the IncomingWebxdcNotify event is not triggered at all. If you set a break point on frontend/../notifications in line 28
BackendRemote.on('IncomingWebxdcNotify', (accountId, { msgId, text }) => {
    incomingWebxdcEventHandler(accountId, msgId, text)
  })
 Then you see there is no IncomingWebxdcNotify event triggered in the group, while in a 1:1 chat it is

WofWca:
> Well when debugging it seems that in the group the IncomingWebxdcNotify event is not triggered at all

Are you sure? It's not the case for me. `IncomingWebxdcNotify` gets triggered, but the notification is not there.

https://github.com/user-attachments/assets/3b764a19-1cff-4377-9130-d05340f0ebff

I got it. It's because the handler is slow now, so pushing to the notifications queue happens _after_ the `IncomingMsgBunch` event.

```
IncomingWebxdcNotify 2 1 created "29" on Jan 25, 2025
22:22:54.855 notifications.ts:31 IncomingMsgBunch
22:22:54.856 notifications.ts:87 queuedNotifications[accountId].push
```

nicodh:
> I got it. It's because the handler is slow now, so pushing to the notifications queue happens _after_ the `IncomingMsgBunch` event.

Not for me. I found that the missing IncomingWebxdcNotify was caused by a calendar that was added to the group before the other test user was added. So that was correct.

But I can't reproduce your issue...?

nicodh:
But it is easly to imagine that fetching messages takes some time...

nicodh:
That means we have to use one handler for all events and queue them, to make sure that they are handled in right order.

WofWca:
Why not just make the handlers synchronous and push `Promise`s to the queue? Sounds better to me.

nicodh:
Added a solution for the queue. Based on https://stackoverflow.com/questions/53540348/js-async-await-tasks-queue (which is a very profound investigation of the queue topic). Would be much easier with rxjs (just call next() on a Subject() while others can subscribe to it) but not worth to introduce a new library here

packages/frontend/src/system-integration/notifications.ts
```diff
@@ -81,11 +124,50 @@ async function showNotification(
...
+        ) {
+          // we have to get the parent message
+          // (the webxdc message which holds the webxdcInfo)
+          message = await BackendRemote.rpc.getMessage(
+            accountId,
+            message.parentId
+          )
+        }
+        if (message.webxdcInfo) {
+          summaryPrefix = `${message.webxdcInfo.name}`
```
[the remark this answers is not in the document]
nicodh: Removed the template literal

nicodh:
Removed the event queue in favor of a more simple solution

packages/frontend/src/system-integration/notifications.ts
```diff
@@ -81,11 +115,50 @@ async function showNotification(
     try {
       const notificationInfo =
         await BackendRemote.rpc.getMessageNotificationInfo(accountId, messageId)
-      const { chatName, summaryPrefix, summaryText } = notificationInfo
+      let summaryPrefix = notificationInfo.summaryPrefix
+      const summaryText = eventText ?? notificationInfo.summaryText
+      const chatName = notificationInfo.chatName
+      let icon = getNotificationIcon(notificationInfo)
+      if (isWebxdcInfo && chatId === -1) {
```
Simon-Laux, line 122: why is chatID `-1` here?

packages/runtime/runtime.ts
```diff
@@ -112,6 +113,7 @@ export interface Runtime {
   showNotification(data: DcNotification): void
   clearAllNotifications(): void
   clearNotifications(chatId: number): void
+  // enables to set a callback (used in frontend RuntimeAdapter)
```
Simon-Laux, line 116:
```suggestion
  /** enables to set a callback (used in frontend RuntimeAdapter) */
```

documentation should use a doc comment so it is shown when you hover on it in the IDE

packages/frontend/src/system-integration/notifications.ts
Simon-Laux: ah because it isn't there yet.... I'd rather use `null` or `undefined` to represent that state.
Or add the chat id to the `IncomingWebxdcNotify` event in core.

Simon-Laux: thinking about it more it makes sense to add it to the event. It already has contact id.

https://github.com/deltachat/deltachat-core-rust/blob/191eb7efdd2765af0b9440a1316e850ed321b2b2/src/events/payload.rs#L111-L123

Simon-Laux (approved):
code looks ok, didn't test.

packages/frontend/src/system-integration/notifications.ts
WofWca, line 122: And `chatId` doesn't seem to get overridden even if it's `-1`. Is it fine to pass `chatId === -1` to `runtime.showNotification()`? Doesn't seem so at a glance.

I agree, it's probably better just to add `chatId` to the event.

WofWca: > I agree, it's probably better just to add chatId to the event.

Actually, it's present on `notificationInfo`, can be obtained from there.

WofWca: Oh I see. This condition is never even true, because `chatId` is never actually -1 here, because it's set inside of `flushNotifications`, in that new `for await` loop.

nicodh: > This condition is never even true, because chatId is never actually -1 here, 

right - the chatId condition is wrong and not needed here

> So yeah, I would remove that loop, and just get chatId from notificationInfo unconditionally

chatId is needed in flushNotifications to find mutedChats which should not show notifications

> showNotification doesn't need the chatId parameter even

not sure which "showNotification" you mean. frontend showNotification passes the chatId to runtime.showNotification which passes it to ipcMain.handle('notifications.show', showNotification) which needs the chatId for the callbacks

WofWca: > chatId is needed in flushNotifications to find mutedChats which should not show notifications

Whoops, forgot about that.

> not sure which "showNotification" you mean.

I mean the bare function defined in this file.

> frontend showNotification passes the chatId to runtime.showNotification which passes it to ipcMain.handle('notifications.show', showNotification) which needs the chatId for the callbacks

Yes, but inside of `showNotification` `chatId` is available on `notificationInfo`.

WofWca (approved):
Would be nice to clean up commit history in order to do a fast-forward merge so that refactoring changes and different functional changes are not mixed, but it's not super important.

I tested this for a bit. Works alright, and the issues are resolved!

Issue #4320, opened 2024-11-06
Image is cropped vertically in full size view


iequidoo:
- Operating System (Linux/Mac/Windows/iOS/Android): Debian GNU/Linux 12.1, KDE 5
- Delta Chat Version: 1.47.1
- Actual behavior: The top and the bottom of images are a bit cropped in full size view. I can see them if i move an image up or down though.
- Steps to reproduce the problem: Click on an image in a chat to open full size view
![image](https://github.com/user-attachments/assets/505749d0-4f9f-4e5b-b6be-9ab6cf5eec00)

nicodh:
I can confirm this. Probably related to blueprint removal

WofWca:
Based on testing, this seems to have been fixed by https://github.com/deltachat/deltachat-desktop/pull/4385

WofWca:
Hmmm, actually not. A few pixels still get cropped out.

PR #4402, opened 2024-12-11
fix: fullscreen images getting cropped a bit

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/4320

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - handle double escape on Dialog #4365
 - fix random crashes on quote reply #4337
 - avoid drafts in readonly chats #4349
+- fullscreen images getting cropped a little #4402, #4385
 
 <a id="1_49_0"></a>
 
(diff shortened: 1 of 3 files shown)


maxphilippov:
I couldn't reproduce the actually "cutting off" part on main, but this fixes empty space at the bottom of fullscreen image (something around 5px) I have on main. And it also doesn't make <dialog> have weird padding which I found when I first investigated this issue. Thanks!

maxphilippov (approved):
I couldn't reproduce the actually "cutting off" part on recent main, but this fixes empty space at the bottom of fullscreen image (something around 5px) I have on main. And it also doesn't make <dialog> have weird padding which I found when I first investigated this issue. Thanks!

WofWca:
> fixes empty space at the bottom of fullscreen image (something around 5px)

Yes, that's part of it (for me at least). There is padding at the bottom, and the top is cropped.

Issue #4404, opened 2024-12-11
chat "scrolls up" right after switching to it, because images load and shift height


WofWca:
- Operating System (Linux/Mac/Windows/iOS/Android): Windows
- Delta Chat Version: 1.48.0
- Expected behavior: Chat loads with the last message at the very bottom
- Actual behavior: First the chat is scrolled all the way to the bottom, but after a few milliseconds "scrolls up"
- Steps to reproduce the problem:
  1. Open a chat where one of the last messages is an image
- Screenshots:
- Logs:

@maxphilippov suggested what is likely to be the solution here https://github.com/deltachat/deltachat-desktop/pull/4315#issuecomment-2473186146, with an extra comment from me https://github.com/deltachat/deltachat-desktop/pull/4315#issuecomment-2503664887.

Update: the issue per se has been fixed in https://github.com/deltachat/deltachat-desktop/pull/4407, which sort of reverts the MR that introduced it (#4315), but we might want to come back to dynamic image sizes.

WofWca:
Turns out this was introduced recently, in https://github.com/deltachat/deltachat-desktop/pull/4315

WofWca:
Ok, looks like setting either `height` attribute or CSS `height:` to intrinsic size, as I tried in #4406 doesn't work, see https://github.com/deltachat/deltachat-desktop/pull/4406#issuecomment-2538962434.

I think we need a different approach, perhaps utilizing the `aspect-ratio` CSS property and turning the container into a flex container, and removing `object-fit`.

Completely disregarding actual message size and only looking at its aspect ration is probably not the way to go for tiny images.
Maybe we can set `max-height` and `max-width` to the intrinsic size.

We should think in terms of "we need to set the height and width of the image, and not rely on the browser to figure out the dimensions based on the image's intrinsic size.

Also, looking at Signal's implementation could be worth it.

nicodh:
Is this still an issue with current main for anyone?

WofWca:
Bruh. It's still reproducible on 1.51.0, with quoted images. I don't understand how, but somehow upon the initial paint the quote text takes 2 lines, and on the next paint it takes 3 lines. We fixed the width and height, WTF?!

https://github.com/deltachat/deltachat-desktop/blob/1907a4f3eb02aa0ce2a61dfc9fdcfa6bfec4389f/packages/frontend/scss/message/_message.scss#L498-L509

PR #4406, opened 2024-12-12
fix: chat not being scrolled to bottom (images)

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/4404.

The bug was introduced in db8327931ef.
Prior to that all attachments had a fixed height of 200px.

This should also eliminate layout shifting when scrolling the chat,
and when new messages arrive.

An alternative solution that would still have layout shifting,
but at least keep the chat "anchored" to bottom would be
https://github.com/deltachat/deltachat-desktop/issues/3753#issuecomment-2336809190,
but it's not supported well by browsers.

I tested it, and verified that images have proper dimensions now even before they get displayed, and that it wasn't the case prior to this MR.

### DIFF
--- a/packages/frontend/src/components/attachment/messageAttachment.tsx
+++ b/packages/frontend/src/components/attachment/messageAttachment.tsx
@@ -22,6 +22,7 @@ import FullscreenMedia, {
 import useTranslationFunction from '../../hooks/useTranslationFunction'
 import useDialog from '../../hooks/dialog/useDialog'
 import AudioPlayer from '../AudioPlayer'
+import { T } from '@deltachat/jsonrpc-client'
 
 type AttachmentProps = {
   text?: string
@@ -86,6 +87,11 @@ export default function Attachment({
         <img
           className='attachment-content'
           src={runtime.transformBlobURL(message.file)}
+          // Pre-set width and height to prevent layout shifts
+          // when the image finally loads.
+          // This is not supposed to affect the rendered dimensions
+          // of the media _after_ it has been loaded.
+          style={getDimensions(message)}
         />
       </button>
     )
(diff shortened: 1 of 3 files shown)


WofWca:
Wait, looks like this adds unnecessary padding for big images...

packages/frontend/src/components/attachment/messageAttachment.tsx
```diff
@@ -217,3 +228,20 @@ export function DraftAttachment({
...
+
+function getDimensions(
+  message: Pick<T.Message, 'dimensionsHeight' | 'dimensionsWidth'>
+): { height: number | undefined; width: number | undefined } {
+  // With some sanity checks, in case core couldn't determine dimensions
+  // for whatever reason, but the browser might.
+  return {
+    height:
+      message.dimensionsHeight > 0 && message.dimensionsHeight < 100_000
+        ? message.dimensionsHeight
```
nicodh, line 240: why we have a limit of 100_000 here? Would 10_000 not be enough?

nicodh (approved):
Tested and solves the problem!

packages/frontend/src/components/attachment/messageAttachment.tsx
WofWca: Probably. But it's just a sanity check.

WofWca: Maybe someone has 4 4K monitors stacked on top of each other on their wall 🤷‍♀️

nicodh:
> Wait, looks like this adds unnecessary padding for big images:
Yes since the height stays the same when rescaling the whole window while the image keeps it ratio

nicodh (changes requested):
It seems this needs more investigation... :-(

WofWca:
OK, I'll close this and let's continue the discussion in the relevant issue.

PR #4407, opened 2024-12-12
Revert dynamic image height

nicodh:
- add a portrait class to images with ratio <= 3:4 and allow at least 300px image hight for those images

basically reverts https://github.com/deltachat/deltachat-desktop/pull/4315/files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - fix random crashes on quote reply #4337
 - avoid drafts in readonly chats #4349
 - fullscreen images getting cropped a little #4402, #4385
+- fix chat "scrolls up" right after switching #4404
 
 <a id="1_49_0"></a>
 
--- a/packages/frontend/src/components/attachment/messageAttachment.tsx
+++ b/packages/frontend/src/components/attachment/messageAttachment.tsx
@@ -22,6 +22,7 @@ import FullscreenMedia, {
 import useTranslationFunction from '../../hooks/useTranslationFunction'
 import useDialog from '../../hooks/dialog/useDialog'
 import AudioPlayer from '../AudioPlayer'
+import { T } from '@deltachat/jsonrpc-client'
 
 type AttachmentProps = {
   text?: string
@@ -54,6 +55,22 @@ export default function Attachment({
       openAttachmentInShell(message)
     }
   }
+  /**
+   * we need a fix height for images to prevent reflow when the image loads
+   * till we have a better solution we allow a maximum height of 200px for
+   * all images in landscape mode and 300px for portrait mode
+   *
+   * for higher values we have to add media queries otherwise messages with
+   * images get strange paddings on small windows
+   */
+  const isPortrait = (
+    message: Pick<T.Message, 'dimensionsHeight' | 'dimensionsWidth'>
+  ) => {
+    if (message.dimensionsHeight == null || message.dimensionsWidth == null) {
+      return false
+    }
+    return message.dimensionsWidth / message.dimensionsHeight <= 3 / 4
+  }
   const withCaption = Boolean(text)
   // For attachments which aren't full-frame
   const withContentBelow = withCaption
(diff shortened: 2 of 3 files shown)


r10s:
the first image, so the status quo,  looks far better - and that not even means the bad-looking padding or the linebreak in the timestamp.

after some discussion with  @paulaluap, the reason seems to be that the _area_ covered by the images are more homogenous.

nicodh:
> i did not follow all previous discussions - what does this PR tries to fix?

I turned out that dynamic image heights (that depend on the individual image dimensions, only limited by some max-height styles) cause wrong height calculations in message list. So before the image is loaded the message has a different height than after loading, leading to wrong scroll positions.

A fixed height will cause images in portrait mode to be displayed very small. 

So this PR reverts the dyamic image dimensions and tries to address the second problem.

packages/frontend/src/components/attachment/messageAttachment.tsx
```diff
@@ -54,6 +55,22 @@ export default function Attachment({
...
+   * till we have a better solution we allow a maximum height of 200px for
+   * all images in landscape mode and 300px for portrait mode
+   *
+   * for higher values we have to add media queries otherwise messages with
+   * images get strange paddings on small windows
+   */
+  const isPortrait = (
+    message: Pick<T.Message, 'dimensionsHeight' | 'dimensionsWidth'>
+  ) => {
+    if (message.dimensionsHeight == null || message.dimensionsWidth == null) {
```
WofWca, line 69: I think this condition is never true, the types say that those are numbers, and the types are generated from Rust IMU.

packages/frontend/src/components/attachment/messageAttachment.tsx
```diff
@@ -54,6 +55,22 @@ export default function Attachment({
       openAttachmentInShell(message)
     }
   }
+  /**
+   * we need a fix height for images to prevent reflow when the image loads
+   * till we have a better solution we allow a maximum height of 200px for
+   * all images in landscape mode and 300px for portrait mode
```
WofWca, line 61: It sounds like this comment belongs to the CSS part of this MR.
And I'm not sure if the context of the comment is understandable, i.e. a better solution to what problem.

But I guess people can find the commit that introduced this later on.

packages/frontend/src/components/attachment/messageAttachment.tsx
```diff
@@ -54,6 +55,22 @@ export default function Attachment({
...
+   *
+   * for higher values we have to add media queries otherwise messages with
+   * images get strange paddings on small windows
```
WofWca, line 64: For higher values of what? Aspect ratio?

WofWca (approved):
Pretty good in-between solution.

WofWca:
> wrong height calculations in message list

It's simpler, it's about the good old layout shift, with no non-browser-internal calculations involved.

r10s:
ahh - thanks for pointing to the issue, and clarifying that this PR fixes a "more important and annoying" bug.

+1 for getting this PR in then. at some point, we should rethink the image sizes and layout, still

packages/frontend/src/components/attachment/messageAttachment.tsx
nicodh: Higher than 200px. The problem is, that with fixed height once the message column gets too narrow the message keeps the initial height while the image keeps it ratio and is thus not filling the height of the message any more - which results in the mentioned paddings

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 - fix random crashes on quote reply #4337
 - avoid drafts in readonly chats #4349
 - fullscreen images getting cropped a little #4402, #4385
+- fix chat "scrolls up" right after switching #4404
```
Simon-Laux, line 16: normally the number here refers to the pr, not the issue.

But then again, in my personal opinion this "add pr number to pr" rule is annoying.... (but I also see some tiny  use in having that info)

PR #4408, opened 2024-12-12
macOS: make area under traffic lights dragable and slightly larger

Simon-Laux:
raiden suggested this change to me in a private message.

| Before | After |
|--------|------|
| Bildschirmfoto 2024-12-12 um 21 51 37 |  Bildschirmfoto 2024-12-12 um 21 51 09 |

also fix the bug that size of that area changed based on profile acount and window height


nicodh (approved):
Reviewed by testing

Issue #4413, opened 2024-12-17
[UX/UI] QR Code is often cut off, which seems to prevent any scanning


ell1e:
I noticed that with the desktop client, the QR code is apparently often displayed with a scrollbar. While a scrollbar is fine for most other content, for a QR code it seems to defeat the purpose since then it can't be scanned:

![Screenshot_20241217_112835](https://github.com/user-attachments/assets/ef989d58-8f0e-4489-908a-70c4a289a9b7)

I'm therefore suggesting that the code should be scaled down instead, at least if that wouldn't result it in being something tiny like 32x32.

- Operating System (Linux/Mac/Windows/iOS/Android): postmarketOS Linux
- Delta Chat Version:  Delta Chat Desktop	1.48.0 (git: flathub) Delta Chat Core	v1.148.7
- Expected behavior: QR Code is scaled down as needed so that scanning remains possible
- Actual behavior: QR Code seems to often be shown with a scroll bar
- Steps to reproduce the problem: 1. Have Delta Chat window be small enough, 2. Open QR scan dialog
- Screenshots: see above
- Logs:

PR #4416, opened 2024-12-17
fix: QR dialog being scrollable

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/4413.

I have tested it in both "New Chat" and "Join Group".

On even shorter screens the QR is probably gonna become unreadable,
but there is not much we can do, other than decreasing paddings.


Simon-Laux (approved):
code and screenshots LGTM, didn't test

PR #4428, opened 2024-12-19
style: better align avatars in messages list

WofWca:


WofWca:
OK, this actually makes the respective message list items taller... IDK if it's worth it.

PR #4429, opened 2024-12-19
improvement: a11y: turn more elements into <button>s

WofWca:
- **improvement: turn clickable items into <button>s**
- **improvement: turn clickable items into <button>s**
- **improvement: a11y: turn quote into <button>**
- **refactor: add comment about a11y**
- **improvement: turn autocrypt msg into <button>**
- **improvement: turn delivery status into <button>**

### DIFF
--- a/packages/frontend/src/components/message/Message.tsx
+++ b/packages/frontend/src/components/message/Message.tsx
@@ -854,6 +854,8 @@ function WebxdcMessageContent({ message }: { message: T.Message }) {
       <img
         src={runtime.getWebxdcIconURL(selectedAccountId(), message.id)}
         alt={`icon of ${info.name}`}
+        // No need to turn this element into a `<button>` for a11y,
+        // because there is a button below that does the same.
         onClick={() => openWebxdc(message)}
       />
       <div
--- a/packages/frontend/src/components/message/Message.tsx
+++ b/packages/frontend/src/components/message/Message.tsx
@@ -479,6 +479,7 @@ export default function Message(props: {
       }
     }
 
+    const TagName = onClick ? 'button' : 'div'
     return (
       <div
         className={classNames(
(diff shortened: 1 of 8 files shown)


packages/frontend/src/components/message/Message.tsx
```diff
@@ -475,6 +479,7 @@ export default function Message(props: {
       }
     }
 
+    const Tag = onClick ? 'button' : 'div'
```
nicodh, line 482: Maybe "TagName" would be a better name for this? when reading
<Tag className=...> I would expect Tag is a custom component when it just is a string. The same for MessageTag > MessageTagName
Just a proposal...

nicodh (approved):
It is  a bit weird now that the reactions and the context menu is now reachable with key navigation but only, if the cursor is on the right height to "activate" these buttons. I can activate the context menu with the "menu" button on my keyboard but I can't navigate on the menu. Is the plan to implement these missing features also?

Nevertheless this PR is an improvement and a good step. Other optimizations can follow later...

WofWca:
> the reactions and the context menu is now reachable with key navigation but only, if the cursor is on the right height to "activate" these buttons

There are ways to activate "react" and the context menu with just the keyboard, so I think there is no need for the extra ways to do that with the controls that appear on hover. But this can be achieved by adding `:focus-within` on top of `:hover` if needed.

> but I can't navigate on the menu

Even with arrow keys?

nicodh:
> > the reactions and the context menu is now reachable with key navigation but only, if the cursor is on the right height to "activate" these buttons
> 
> There are ways to activate "react" and the context menu with just the keyboard, so I think there is no need for the extra ways to do that with the controls that appear on hover. But this can be achieved by adding `:focus-within` on top of `:hover` if needed.

For me it is like this:
I can focus (via tab key) to the message with the reaction
image

but the next tab jumps to the attach button in the message input field

image

I can't focus the other messages, (only when I go with the cursor near a message which makes the reactions button visible)

> 
> > but I can't navigate on the ,
> 
> Even with arrow keys?
My fault, I tried tab key. Works with arrow keys.

WofWca:
Making each message keyboard-navigable is taken care of in https://github.com/deltachat/deltachat-desktop/pull/4294.

Issue #4314, opened 2024-11-04
style: make composer's quote height constant


WofWca:
- Operating System (Linux/Mac/Windows/iOS/Android):
- Delta Chat Version: 1.47.1
- Expected behavior: the quote height is constant so the content doesn't jump
- Actual behavior: messages shift around and it's hard to keep track of them visually
- Steps to reproduce the problem:
    1. Enable the experimental "Telegram-style Ctrl+Up to reply"
    2. Open a chat with messages of quite different length
    3. Press Ctrl + Up a few times.
- Screenshots:

    https://github.com/user-attachments/assets/71aa116b-d85d-40ca-8cc7-060b41229d7a

- Logs:

The initial height transition as you make the first quote is fine, but it should stay constant as you switch to another message.

nicodh:
So that would mean we always set it to the maximum possible height it might have, which seems to be something around 80px.

Then it would look like this for one liner

<img width="557" height="137" alt="Image" src="https://github.com/user-attachments/assets/ef0e786f-e81b-4122-8138-5949440b69ce" />

Is that what you mean with this issue?

WofWca:
I did not have a precise plan of action when I made the issue. Setting it to max height sounds OK, but maybe we just don't need the quote to be multi-line in the first place, i.e. having just one line of the message (plus another line for the author) would be enough.

PR #4431, opened 2024-12-22
fix: chat "scrolling up" because of imgs loading

WofWca:
Visual aspect, before / after: 
image image

This, together with https://github.com/deltachat/deltachat-desktop/pull/4407,
should close https://github.com/deltachat/deltachat-desktop/issues/4404.

The bug was introduced in https://github.com/deltachat/deltachat-desktop/pull/3999.
Pior to that we used to have fixed width on quoted images as well.
This commmit practically reverts that MR, but also adds
`object-fit` and `object-position`, and is now in line with
message attachment images

### DIFF
--- a/packages/frontend/scss/message/_message.scss
+++ b/packages/frontend/scss/message/_message.scss
@@ -495,8 +495,21 @@
   }
 
   img.quoted-image {
+    // Remember that width and height need to be set prior to the image
+    // getting actually loaded by the browser, to avoid content shifts.
+    // Same goes for messages' `.attachment-content`.
+    // See https://github.com/deltachat/deltachat-desktop/issues/4404.
+    //
+    // Setting just the height is not enough, because there might be text
+    // displayed beside the image, which might wrap and thus change height.
+    //
+    // In addition, really short and wide images could make the message
+    // overflow the chat container, for non-self-sent messages at least.
     height: var(--quote-img-size);
-    // do not set width, to keep the aspect ratio (see https://github.com/deltachat/deltachat-desktop/issues/3931)
+    width: var(--quote-img-size);
+    object-fit: scale-down;
+    object-position: center;
+
     margin-left: 6px;
   }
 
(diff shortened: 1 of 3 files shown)


WofWca:
An alternative solution would be #4314.

packages/frontend/scss/message/_message.scss
```diff
@@ -470,8 +470,21 @@
...
+    // See https://github.com/deltachat/deltachat-desktop/issues/4404.
+    //
+    // Setting just the height is not enough, because there might be text
+    // displayed beside the image, which might wrap and thus change height.
+    //
+    // In addition, really short and wide images could make the message
+    // overflow the chat container, for non-self-sent messages at least.
     height: var(--quote-img-size);
-    // do not set width, to keep the aspect ratio (see https://github.com/deltachat/deltachat-desktop/issues/3931)
+    width: var(--quote-img-size);
```
nicodh, line 484: Is object-fit: scale-down in conjunction with width: 100% not enough (and better since it does not force any image to have a 36px width)?

I tested various variants with extreme landscape ratios and text and it looked better.

But I can also live with the current solutions, since we are going to find  amore adaptive generic solution for images anyway (hopefully)

nicodh (approved):

packages/frontend/scss/message/_message.scss
WofWca: Have you tried quoting a in image with a captoion or just an image?

nicodh: I tried with text and without. But it's true, with an extrem ratio (like 1300 x 30 px) it hides the text. So let's solve that hopefully soon as said above

WofWca: `width: 100%` simply makes it take up 50% of the flex container, which can be adjusted with `flex-shrink`. IDK if this is what we want.

Issue #3030, opened 2022-11-28
show contacts in forward message dialog, too


Simon-Laux:
sometimes you want to forward to a contact that you don't have a chat with yet.

In iOS you can also search for contacts and even messages in the forward dialog/screen (but I don't think messages make much sense here).

WofWca:
Implementation-wise this has to do with the fact that we use `useChatList` in such places, instead of a contact list.
Some other such places are: Mailto dialog, ViewProfile (chats in common), webxdc sendToChat dialog. Just look for references of `useChatList`.

WofWca:
Since not so long ago we have `SelectContactDialog` https://github.com/deltachat/deltachat-desktop/issues/3830 that probably can be used for this:

https://github.com/deltachat/deltachat-desktop/blob/a5c02433976909fe01a64f3049c96d0d101ae9d8/packages/frontend/src/components/dialogs/SelectContact/index.tsx#L30

Simon-Laux:
> also option to put in an email address to create a new contact to forward to would be nice.

note: this is no longer relevant, atleast not for chatmail profiles.

> But contacts there are not sorted in the "most recently chatted with" order, so it might be inconvenient.

well I proposed this feature to forward to contacts, that you haven't already chatted with (but know already, because you share a group).

maybe some button or fake chat list entry to open a contact selector?
@r10s what do you think could be a minimal effort solution to my problem here (can't forward to contacts that I don't have a chat with yet)?

On mobile it shows search results, that also has a sections for contacts, but not sure how hard it is to port/reuse the search result code from desktop. (if we port that we need to hide the message results and make sure the keyboard navigation across both lists works also fine)

r10s:
not sure if having always a "new contact" button would really be nice. it is a rarely used thing nowadays. the issue comes from a time, where we did not had chatmail, so it was possible to forward by just entering an email address.

i would prefer to have it more implicitly, offering the contacts when you search for some letters. but i understand that this may be complicated.

but i would also be fine with closing this issue and to postpone it to some later point. it seems not to be important, and there is not much request for it, i also updated [interface#39](https://github.com/deltachat/interface/issues/39#issuecomment-3052241499) so it no longer looks like a flaw to not support it.

i'd downgrade this issue to a normal feature proposal in the forum, where it is in competition with other, possibly more impactful things

Simon-Laux:
> the issue comes from a time, where we did not had chatmail, so it was possible to forward by just entering an email address.

[...]

PR #4440, opened 2024-12-24
refactor: unify "Select Chat": use one component

WofWca:
For:
- webxdc `sendToChat()`
- mailto dialog
- "Forward Message" dialog

What this commit comes down to:
- Take the ForwardTo dialog and use it to create
  a more generic, reusable SelectChat dialog.
- Make those similar dialogs use the new SelectChat component,
  instead of defining their markup and logic manually.

The dialogs have practically the same markup and the same
search logic, with very little differences.
They apparently were copy-pasted from `ForwardMessage` dialog,
which is the oldest of them all.

It's getting pretty annoying to maintain them, with all the
`RovingTabindex` and `InfiniteLoader`, so, it's about time
to unify them.

This is mainly a refactor, but causes a few small visual changes.
For example, this fixes a small visual bug
in the `WebxdcSendToChat` dialog
where the search field takes full height
when there are no search results.


nicodh (approved):
Reviewed by testing & reading

WofWca:
@Simon-Laux Could you please see if you agree with the general idea of this MR?

Simon-Laux (approved):
Good idea. removing code without removing functionality is always good 👍 


#3030 would affect all 3 dialogs, so having only one place to add contacts to is also nice for that.

PR #4441, opened 2024-12-24
fix: a11y: arrow-key navigation stopping working after ~10 keypresses

WofWca:
Reviewing commit-by-commit should be a bit easier. But basically what this does is it un-inlines the renderer functions, addressing this issue: https://github.com/bvaughn/react-window/issues/420#issuecomment-585813335.
You can check this behavior by arrow-key navigating the contact list in "new chat", as long as your contact list is at least ~30 items long. When new items get loaded (with 'react-infinite-loader'), the parent component re-renders, and all items' DOM get re-created, losing focus.

I am not 100% sure whether we should squash the commits. They're all quite similar, but it might not be easy to see separate logical changes when squashed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 ## Fixed
 - fix chat "scrolls up" right after switching (rev 2) #4431
 - when deleting a message from gallery, update gallery items to remove the respective item #4457
+- accessibility: fix arrow-key navigation stopping working after ~10 key presses #4441
 - accessibility: make more items in messages list keyboard-accessible #4429
 - fix "incoming message background color" being used for quotes of outgoing sticker messages #4456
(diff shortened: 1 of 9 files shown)


nicodh (commented):
What issue is fixed here?  How can I test if this PR solves the issue?
Which means what is the use case for the parent component being rerendered and items loosing focus?

WofWca:
I just edited the description:

> You can check this behavior by arrow-key navigating the contact list in "new chat", as long as your contact list is at least ~30 items long. When new items get loaded (with 'react-infinite-loader'), the parent component re-renders, and all items' DOM get re-created, losing focus.

nicodh (approved):
Reviewed by testing. 

Maybe squashing is fine, since most commits are applied to different files

PR #4442, opened 2024-12-25
refactor: correctly use `itemKey` in gallery

WofWca:
This is based on #4441.

This does not appear to have noticeable functional effect.


nicodh (approved):
This is just a minor change, when compared with #4441

PR #4451, opened 2025-01-05
Upgrade core to 1.53.0

nicodh:


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 ## Changed
 - dev: upgrade react to v18 and react pinch pan zoom to v3
+- Update `@deltachat/stdio-rpc-server` and `deltachat/jsonrpc-client` to `1.153.0`
```
iequidoo, line 9: IDK how this is generated, but i think these changes are worth mentioning (from the core's `CHANGELOG.md`) as they may be user-visible:
```
### Fixes

- Never change Viewtype::Sticker to Image if file has non-image extension ([#6352](https://github.com/deltachat/deltachat-core-rust/pull/6352)).
- Change BccSelf default to 0 for chatmail ([#6340](https://github.com/deltachat/deltachat-core-rust/pull/6340)).
- Don't treat location-only and sync messages as bot ones ([#6357](https://github.com/deltachat/deltachat-core-rust/pull/6357)).
- Update shadowsocks crate to 1.22.0 to avoid panic when parsing some QR codes.
- Prefer to encrypt if E2eeEnabled even if peers have EncryptPreference::NoPreference.
- Allow empty `To` field for self-sent messages.
```
Various refactorings and API changes may be omitted i think.

r10s: ftr, on Android and iOS we have a task in RELEASE.md that says to add user-visible, relevant changes from core-changelog to Android/iOS changelog. we do not expect end-user to read the core-changelog - also the core-changelog is more meant for devs using that library, not for the end-user

r10s (approved):

Issue #4453, opened 2025-01-05
wrong background color when showing "reply with sticker"


r10s:
when showing ones own "reply with sticker" done eg. on android or ios, the reply is shown using the background color of _incoming_ messages, not as usual the background color of _outgoing_ messages.

in other words, eg. in light mode, the background should be green instead of white:

<img width="553" alt="Screenshot 2025-01-05 at 22 36 03" src="https://github.com/user-attachments/assets/857054ba-2feb-4f93-8de1-52ddb094f775" />

the background color are an important UX element to recognise messages, eg. we also use that for secondary elements as the small emoji pills  - we should not degrade that UX element guarantee without good reasons

eg. it does not make much sense that a reply-with-image or reply-with-text gets a different background color as reply-with-sticker

WofWca:
Which Delta Chat version is this?

r10s:
> Which Delta Chat version is this?

the screenshot above is from 1.49.0, but i checked 1.50.1 that has the same issue:

<img width="543" alt="Screenshot 2025-01-06 at 15 40 47" src="https://github.com/user-attachments/assets/6a0b0d04-4f91-4fc4-ab47-9d9bf48633cf" />

PR #4456, opened 2025-01-06
style: outgoing bg color for outgoing stickers

WofWca:
Before:
image image

After:
image image

For incoming messages the color is also respective:

image image

Closes https://github.com/deltachat/deltachat-desktop/issues/4453.
The bug (if you can call it that) was introduced in f83ce46cbf80a4c019d3be02222b164a8d58fe56.
Similar commit: 8259aad97.

CI failures are to be fixed in https://github.com/deltachat/deltachat-desktop/pull/4426.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
   - Allow empty `To` field for self-sent messages.
 
 ## Fixed
+- fix "incoming message background color" being used for quotes of outgoing sticker messages #4456
 
 ## [1.50.1] - 2024-12-18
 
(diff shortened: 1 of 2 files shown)


r10s (approved):
nice! thanks for fixing that detail! 🙏

code lgtm as far as i can tell, but i also tested and it works as expected :)

r10s:
> For incoming messages the color is also respective

i do not really understand that hint and the image: incoming messages were always fine afaik - or do you just want to point out that that part was unchanged? :)

WofWca:
> or do you just want to point out that that part was unchanged? :)

Yes.

PR #4458, opened 2025-01-08
fix: devmode: fix logging and counting jsonrpc requests

Simon-Laux:
This was likely broken by me during refactoring when changing apis and initialisation to make the browser version possible.
The problem was that the comandline flags were checked before they were loaded.

Now the jsonrpc request are logged again in devmode. (write `-jsonrpc` in the filter to not be annoyed from it)
And you can now do `exp.printCallCounterResult()` again:
Bildschirmfoto 2025-01-08 um 04 41 16


WofWca (approved):
I didn't follow the code through, but the change looks simple, and it does fix the issue.

PR #4459, opened 2025-01-08
add show_app_in_chat option to webxdc info message context menu

Simon-Laux:
To make it easy again to reach/jump-to the webxdc message. Since the notification webxdc info messages open the app directly on click.

Bildschirmfoto 2025-01-08 um 05 26 40


WofWca (approved):
Nice little feature!
Tested, and the code looks good.

PR #4460, opened 2025-01-08
perf: don't re-render on every window resize

WofWca:


Simon-Laux (approved):
code LGTM.

WofWca:
Tests because of `pnpm install`. Unrelated.

PR #4470, opened 2025-01-10
Validate Proxy Url before enable proxy

### LINKED ISSUE #4231: Restart IO after proxy config changes
After changing the proxy URL or changing the proxy_enabled status we need to trigger a IO restart on core

To reproduce: set an invalid string as proxy url like "test"
See the error coming from core:
`Failed to parse proxy URL
Caused by:
    0: Cannot parse proxy URL
    1: relative URL without a base"`

Then reset to an empty string and set proxy_enabled is to 0. Core throws an error STARTTLS upgrade failed. Switching to another account and back does not fix that. After restarting desktop (and thus core) the config is used as expected and connection works.

See comments on https://github.com/deltachat/deltachat-core-rust/issues/6040#issuecomment-2426023052

Error with proxy_enabled set to false
error-desktop-settings
https://github.com/deltachat/deltachat-core-rust/issues/6040

nicodh:
Fixes #4231 

Restarting IO didn't help.

### DIFF
--- a/packages/frontend/src/components/dialogs/EditAccountAndPasswordDialog.tsx
+++ b/packages/frontend/src/components/dialogs/EditAccountAndPasswordDialog.tsx
@@ -110,13 +110,22 @@ function EditAccountInner(onClose: DialogProps['onClose']) {
       initial_settings.proxy_enabled !== accountSettings.proxy_enabled &&
       accountSettings.proxy_enabled === '1'
     ) {
-      const qr = await BackendRemote.rpc.checkQr(
-        selectedAccountId(),
-        accountSettings.proxy_url
-      )
-      if (qr.kind !== 'proxy') {
+      try {
+        const qr = await BackendRemote.rpc.checkQr(
+          selectedAccountId(),
+          accountSettings.proxy_url
+        )
+        if (qr.kind !== 'proxy') {
+          openDialog(AlertDialog, {
+            message: tx('proxy_invalid'),
+          })
+          return
+        }
+      } catch (error) {
+        const errorStr =
+          error instanceof Error ? error.message : JSON.stringify(error)
         openDialog(AlertDialog, {
-          message: tx('proxy_invalid'),
+          message: tx('proxy_invalid') + '\n' + errorStr,
         })
         return
       }


packages/frontend/src/components/dialogs/EditAccountAndPasswordDialog.tsx
```diff
@@ -102,15 +103,29 @@ function EditAccountInner(onClose: DialogProps['onClose']) {
...
-      if (confirmed) {
-        update()
+      if (!confirmed) {
+        return
+      }
+    } else if (
+      initial_settings.proxy_enabled !== accountSettings.proxy_enabled &&
+      accountSettings.proxy_enabled === '1'
+    ) {
+      const qr = await BackendRemote.rpc.checkQr(
```
Simon-Laux, line 113: are all proxies we support also supported to be read as qr code? Like is the format on both fields the same type?

Also this looks like a workaround, what do the other platforms do? Is there a reason core does not have a `validate_proxy` or even `test_proxy` method?

nicodh: > are all proxies we support also supported to be read as qr code? 
well it's in fact just a string containg (possibly) an URL.

we could do some URL validation ourselves, but I assumed it's better to use the same logic as core.

nicodh: > Also this looks like a workaround, what do the other platforms do?

I discussed it with @adbenitez - he proposed the QR code check.

nicodh: @link2xt is the QR code check sufficient? Is there a better core check available?

packages/frontend/src/components/dialogs/EditAccountAndPasswordDialog.tsx
```diff
@@ -102,15 +103,29 @@ function EditAccountInner(onClose: DialogProps['onClose']) {
...
+        selectedAccountId(),
+        accountSettings.proxy_url
+      )
+      if (qr.kind !== 'proxy') {
+        openDialog(AlertDialog, {
+          message: tx('proxy_invalid'),
+        })
+        return
       }
```
WofWca, line 122: `checkQr` can throw` e.g. for the string `socks5:`

```suggestion
      let qr: T.Qr | undefined = undefined
      try {
        qr = await BackendRemote.rpc.checkQr(
          selectedAccountId(),
          accountSettings.proxy_url
        )
      } catch (error) {
        const errorStr =
          error instanceof Error ? error.message : JSON.stringify(error)
        openDialog(AlertDialog, {
          message: tx('proxy_invalid') + '\n' + errorStr,
        })
        return
      }
      if (qr.kind !== 'proxy') {
        openDialog(AlertDialog, {
          message: tx('proxy_invalid'),
        })
        return
      }
```

WofWca (approved):
Works alright, apart from my suggestion. Tested with SOCKS5 proxies: reachable, unreachable, invalid URLs.

The other changes also make sense.

packages/frontend/src/components/dialogs/EditAccountAndPasswordDialog.tsx
WofWca: Oh, and `import type { T } from '@deltachat/jsonrpc-client'` would need to be added.

PR #4471, opened 2025-01-10
Always validate non empty displayname

### LINKED ISSUE #4042: force non-empty profile name also on editing
during onboarding, we enforce a profile name to be set.

for consistency, and [as same reasons apply](https://github.com/deltachat/deltachat-android/issues/3160#issuecomment-2226887687), we should not allow to clear the name later on editing.

there are old profiles where we did not have enforced a name on creation. we would enforce them now during editing, if eg. the user only wants to edit the "bio". however, that seems to be a minor, and also android does it that way since some time, without much complains or issues.

cmp https://github.com/deltachat/deltachat-android/issues/3160, https://github.com/deltachat/deltachat-ios/issues/2243

nicodh:
resolves #4042


r10s (approved):
nice! that looks straight forward :)

PR #4475, opened 2025-01-11
add experimental content protection option

### LINKED ISSUE #4347: Use Electron `setContentProtection` API on Windows and macOS
Electron has a function to prevent screenshotting the application window:
https://www.electronjs.org/docs/latest/api/browser-window#winsetcontentprotectionenable-macos-windows

On Windows and macOS this could be enabled by default and and option to disable it provided.

Delta Chat on Android has a similar "Screen security" option in Advanced settings, so it makes sense to have it on desktop.

Windows is introducing "Recall" and if screen capture is not prevented, disappearing messages can easily end up in the history stored on device, affecting both the user and all its contacts: 
https://support.delta.chat/t/opt-out-of-windows-recall-screenshotting-the-chats-on-desktop/3342

Simon-Laux:
it is off by default and and an experimental setting and hidden on linux because this api is macOS and windows only

closes #4347

On macOS the window is hidden from screenshots but not from screen recording.

The change should apply to all windows (main app, help, html email view, webxdc)


packages/target-electron/src/content-protection.ts
```diff
@@ -0,0 +1,27 @@
+import { BrowserWindow } from 'electron'
+import { DesktopSettings } from './desktop_settings'
+import { platform } from 'os'
+import { getLogger } from '@deltachat-desktop/shared/logger'
+
+const log = getLogger('contentProtection')
+
+function updateContentProtection(windows: BrowserWindow, enabled: boolean) {
+  if (platform() === 'darwin' || platform() === 'win32') {
```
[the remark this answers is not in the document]
Simon-Laux: no but this way you get a log entry mentioning why it failed, also I doubt that this function will ever be available on linux.

WofWca (commented):
It seems like this does not apply to webxdc windows?

nicodh (commented):
Tested on Windows10 - complete app including webxdc apps are excluded from screenshots when the feature is enabled!!

Couldn't test screen recoring though, due to missing hardware support (VritualBox)

packages/target-electron/src/content-protection.ts
```diff
@@ -0,0 +1,26 @@
...
+import { getLogger } from '@deltachat-desktop/shared/logger'
+
+const log = getLogger('contentProtection')
+
+function updateContentProtection(windows: BrowserWindow, enabled: boolean) {
+  windows.setContentProtection(enabled)
+  if (platform() !== 'darwin' && platform() !== 'win32') {
+    log.warn('setContentProtection not available on your platform', platform())
+  }
+}
```
WofWca, line 13:
```suggestion
function updateContentProtection(window: BrowserWindow, enabled: boolean) {
  window.setContentProtection(enabled)
  if (enabled && platform() !== 'darwin' && platform() !== 'win32') {
    log.warn('setContentProtection not available on your platform', platform())
  }
}
```

packages/target-electron/src/content-protection.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export function setContentProtection(windows: BrowserWindow) {
+  updateContentProtection(
+    windows,
+    DesktopSettings.state.contentProtectionEnabled
+  )
+}
```
WofWca, line 20:
```suggestion
export function setContentProtection(window: BrowserWindow) {
  updateContentProtection(
    window,
    DesktopSettings.state.contentProtectionEnabled
  )
}
```

WofWca (approved):
Looks alright. Tested on Linux, where this feature is not supported, and it works as usual: the setting is not displayed in the UI, and the `windows.setContentProtection(enabled)` is executed, but doesn't throw.

PR #4476, opened 2025-01-11
improve backup transfer dialog.

### LINKED ISSUE #4420: small improvements to backup transfer progress dialog
- while progress == 0 show some message like "establishing connection between devices", hide the message as soon as there's progress or replace it with a new message that says "transferring"

- set a timeout of one second, if there is no progress until then show a link/button to troubleshooting steps.

The second point is actually the more important one -> if the user likely has trouble point them to FAQ, where they can get help.

Simon-Laux:
Bildschirmfoto 2025-01-11 um 02 15 10
Bildschirmfoto 2025-01-11 um 21 22 50
Bildschirmfoto 2025-01-11 um 02 17 43

- add description of the current step (connecting devices or transfering)
- button for trouble shooting 
- and after 4 seconds of progress = 0 it shows a message telling the user to checkout trouble shooting

closes #4420


_locales/_untranslated_en.json
```diff
@@ -80,5 +80,11 @@
   },
   "contact_is_blocked": {
     "message": "This contact is blocked"
+  },
+  "multidevice_connecting": {
+    "message": "Connecting devices"
+  },
+  "multidevice_connection_takes_too_long": {
+    "message": "Devices are taking too long to connect - Please ensure they are on the same network and follow the troubleshooting steps."
```
WofWca, line 88: Just FYI we have `<string name="multidevice_same_network_hint">Make sure both devices are on the same Wi-Fi or network</string>`

WofWca (commented):
So, just to be clear, this relies on the core to show the progress of 0 when we're not connected.

The linter is complaining. Otherwise looks good!

packages/frontend/src/components/dialogs/SetupMultiDevice/ReceiveBackupProgressDialog.tsx
```diff
@@ -76,13 +84,24 @@ export function ReceiveBackupProgressDialog({
               {tx('error')}: {error}
             </p>
           )}
+          <p>
+            {importProgress === 0
+              ? tx('multidevice_connecting')
```
r10s, line 89: i'd just use the existing string `connecting...` - this is already translated and usually, we avoid having double strings until we know that there will be conflicts:

```suggestion
              ? tx('connectivity_connecting')
```

(would habe been better if the key would be more neutral, but well, that's not the worse :)

packages/frontend/src/components/dialogs/SetupMultiDevice/ReceiveBackupProgressDialog.tsx
```diff
@@ -16,19 +16,25 @@ import {
...
+import Button from '../../Button'
+import { runtime } from '@deltachat-desktop/runtime-interface'
 
 const log = getLogger('renderer/receive_backup')
 
 type Props = {
   QrWithToken: string
 }
 
+// time it takes on progress === 0 to show the rouble shooting button in ms
```
r10s, line 28:
```suggestion
// time it takes on progress === 0 to show the trouble shooting button in ms
```

packages/frontend/src/components/dialogs/SetupMultiDevice/ReceiveBackupProgressDialog.tsx
WofWca: I also thought about this but decided not to comment because you could argue that that string could be translated as "Connecting to email server", especially that there is a translation comment for it.

But I won't insist.

r10s: the "official" translation comment says "stay short", that's fine. we can separate the strings as soon as trouble appear, but in the vast majority of cases, over all the years, it is fine to reuse (not saying this approach is great for all apps, but in case of Delta Chat, it was :)

_locales/_untranslated_en.json
r10s: "too long" might be a bit harsh, i do not really know if there are circumstances of very slow, overloaded wifi where things could still be fine.

so, yes, maybe strike that part and maybe just reuse the existing string. or maybe go for sth. as:

> This takes longer than expected 🐌 make sure both devices are on the same Wi-Fi or network and check troubleshooting

in any case, mentioning Wi-Fi explicitly was considered that time to be a critical information as this is much more known to the end-user as an abstract "network". beside the added information, the suggestion is also a bit shorter. the emoji is a matter of taste, however, why not be a bit relaxed if otherwise things are not working. it also gives a bit more emphasis to the hint

r10s:
> So, just to be clear, this relies on the core to show the progress of 0 when we're not connected.

[by definition](https://c.delta.chat/group__DC__EVENT.html#ga006ea41d9c1a76ffc672752484c61e6c), 0 means "error, cannot connect, aborting", at least on rust/cffi event level: 

changing that to another meaning or having different meanings between cffi and jsonrpc here is a call for trouble.

android also shows "connecting", but does that only _before_ the first event is emitted, i think, this is the more correct approach.

also, android already today shows a hint, we can think over rewording (eg. pointing to troubleshooting, that was not always available), but maybe a better approach for desktops is to do the same:

- as long as screen shows "Connecting..." add the unconditional hint

that way, we also do not need to think about what is "slower than expected" - and we can maintain the same wording across all platforms

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
   - Prefer to encrypt if E2eeEnabled even if peers have EncryptPreference::NoPreference.
   - Allow empty `To` field for self-sent messages.
   - displayname may not be empty anymore #4471
+- improve backup transfer dialog (different message for connection step, timed message to tell user to checkout troubleshooting, button to link to trouble shooting) #4476
```
iequidoo, line 23:
```suggestion
- Improve backup transfer dialog (different message for connection step, timed message to tell user to check out troubleshooting, button to link to trouble shooting) #4476
```

_locales/_untranslated_en.json
```diff
@@ -80,5 +80,11 @@
   },
   "contact_is_blocked": {
     "message": "This contact is blocked"
+  },
+  "multidevice_connecting": {
+    "message": "Connecting devices"
```
iequidoo, line 85:
```suggestion
    "message": "Connecting devices..."
```
but indeed maybe it's not needed at all.

_locales/_untranslated_en.json
iequidoo: I wanted to say that `—` should be used instead of `-`, but `🐌` is better of corse

Simon-Laux: I think firewall or not allowing local connection in the mobile apps are also common. So mentioning just wifi is not enough, better point to trouble shooting.

Simon-Laux:
> [by definition](https://c.delta.chat/group__DC__EVENT.html#ga006ea41d9c1a76ffc672752484c61e6c), 0 means "error, cannot connect, aborting", at least on rust/cffi event level:

Good Point, we should use some different value for the connection step then if needed. Or change the logic in desktop to wait for first update.

On desktop it does not rely on the error in the progress report event, it just displays the error `getBackup` throws.

iequidoo (approved):

iequidoo:
> [by definition](https://c.delta.chat/group__DC__EVENT.html#ga006ea41d9c1a76ffc672752484c61e6c), 0 means "error, cannot connect, aborting", at least on rust/cffi event level:

Just in case i've checked that the core only reports 0 on errors or a canceled transfer.

WofWca (approved):

PR #4477, opened 2025-01-11
update `@deltachat/message_parser_wasm` from `0.11.0` to `0.12.0`

Simon-Laux:
See https://github.com/deltachat/message-parser/releases/tag/0.12.0


WofWca (approved):
While testing I noticed that it still consumes those delimiters if there are several of them, e.g.
`Check out https://example.com!!!`, and `https://example.com...`:




But the issue with a single `!` is fixed.

PR #4479, opened 2025-01-11
error dialog is deltachat-rpc-server is not found

### LINKED ISSUE #4469: Add dedicated error dialog when core executable is missing
Sometimes antivirus software decided to delete core #4209

We should make an extra error dialog for this case to inform the user to look if they can restore it from some "quarantine" and report the issue to us. Also to help with avoiding blame for delta chat.

Simon-Laux:
closes #4469

Bildschirmfoto 2025-01-11 um 04 59 16

needs to also be tested on windows (after installation rename deltachat-rpc-server or delete it to simulate the issue)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,13 +49,9 @@
 - accessibility: arrow-key navigation for message list, gallery and sticker picker #4294, #4376, #4372,
 - accessibility: arrow-key navigation: handle "End" and "Home" keys to go to last / first item #4438
 - add show_app_in_chat option to webxdc info message context menu #4459
-<<<<<<< HEAD
 - add experimental content protection option (to prevent screenshots and screenrecording the app) #4475
 - app picker for webxdc apps in attachement menu #4485
-- add special error dialog for the case that deltachat-rpc-server is not found
-=======
 - add special error dialog for the case that deltachat-rpc-server is not found #4479
->>>>>>> cf5337ce0 (add pr number to changelog)
 
 ### Changed
 - Update `@deltachat/stdio-rpc-server` and `deltachat/jsonrpc-client` to `1.154.1`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,8 +49,13 @@
 - accessibility: arrow-key navigation for message list, gallery and sticker picker #4294, #4376, #4372,
 - accessibility: arrow-key navigation: handle "End" and "Home" keys to go to last / first item #4438
 - add show_app_in_chat option to webxdc info message context menu #4459
+<<<<<<< HEAD
 - add experimental content protection option (to prevent screenshots and screenrecording the app) #4475
 - app picker for webxdc apps in attachement menu #4485
+- add special error dialog for the case that deltachat-rpc-server is not found
+=======
+- add special error dialog for the case that deltachat-rpc-server is not found #4479
+>>>>>>> cf5337ce0 (add pr number to changelog)
 
 ### Changed
 - Update `@deltachat/stdio-rpc-server` and `deltachat/jsonrpc-client` to `1.154.1`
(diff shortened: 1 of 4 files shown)


nicodh (commented):
Tested on Windows 10 

This is what I get. Which is not really what your screen shot shows

edit: this is the message created by ipc.ts obviously

image

Simon-Laux:
The error is only shown in packaged releases

r10s:
are there any reasons thinkable where this could happen to a normal user that are _not_ related to an AntiVirus program?

otherwise, maybe be a bit more broad and on point, also avoiding too technical things in the headline - the normal end-user has no idea an "core library" even  exist

packages/target-electron/src/deltachat/stdio_server.ts
```diff
@@ -16,14 +17,52 @@ export class StdioServer {
...
+          `Error with core has been detected, please contact developers: You can reach us on delta@merlinux.eu or on github.com/deltachat/deltachat-desktop/issues .
+
+          ${err.name}: ${err.message}
+
+          The Log file is located in this folder: ${getLogsPath()}\n
+          `
+        )
+      }
+      // I think we can exit in all the cases, because all errors here are serious
+      app.exit(1)
```
WofWca, line 63: Where can I see the list of errors? I'm worried that there might be non-critical errors.

WofWca (commented):
Apart from one comment, looks good.

packages/target-electron/src/deltachat/stdio_server.ts
Simon-Laux: the list is a comment at the start of the function:
```js
this.serverProcess.on('error', err => {
      // The 'error' event is emitted whenever:
      // - The process could not be spawned.
      // - The process could not be killed.
      // - Sending a message to the child process failed.
      // - The child process was aborted via the signal option.
      // ~ https://nodejs.org/api/child_process.html#event-error
```

WofWca: Oops, missed that.

> Sending a message to the child process failed.

I'm not sure about this. Are we unable to proceed when this happens? Have we seen this happen?

Simon-Laux: better exit/crash than undefined behaviour

WofWca: That's for sure. But what I'm getting at is that maybe this happens all the time right now, and Delta Chat keeps working fine, without us knowing it.

Simon-Laux: It will act weird if messages get lost -> like won't react to user action. Better to fail in those cases that we notice the problem and can fix it IMO.

WofWca: I would leave this line out until after 1.52.0. 
But if you're sure it's fine, I won't insist.

nicodh:
I thought maybe @Simon-Laux Simon will react on the suggestions. Why is this blocking stuff in the core?

nicodh (approved):

WofWca:
> You mean "may have" false positives? We don't expect that to happen, right?

I'm pretty sure we do. The false-positives pretty much have to do with Rust upgrade, see that issue.

Simon-Laux:
> I thought maybe @Simon-Laux Simon will react on the suggestions

no don't have opinions on the wording, I did my part: it works for me in packaged builds. I don't care about the exact message/wording.
The main point for me is that we get information we need when people complain and send us screenshots of the error window without sending us the log.
The other point is that people know it was their antivirus, so I agree with bjoern in principle (making message shorter and more direct). But I won't participate in the wording debate further

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@
 - add show_app_in_chat option to webxdc info message context menu #4459
 - add experimental content protection option (to prevent screenshots and screenrecording the app) #4475
 - app picker for webxdc apps in attachement menu #4485
+- add special error dialog for the case that deltachat-rpc-server is not found #4479
```
Simon-Laux, line 32: needs to be rebased before merging, so that the changelog entry is not in the past/already-released version.

PR #4485, opened 2025-01-13
App Picker for Desktop

### LINKED ISSUE #4463: Feature: Provide an app picker  similar to mobile apps
Data is available as JSON here https://apps.testrun.org/xdcget-lock.json

Picker in desktop might be implemented  like the Emoji picker to avoid duplicate implementations for Browser, Electron, Tauri.

- when an app is selected it should be shown as draft.

- we could provide an additional tab with "recently seen" apps with already used apps from chats

tbd: should we provide a "detail view" like on Android (which is basically https://webxdc.org/apps in a web view) or minimal view like on IOS

This is how it looks like in Android.
<p>
Alt-Text
</p>
<p>
Alt-Text
</p>

This the IOS view:

<p>
Alt-Text
</p>

nicodh:
Loads apps (as JSON) from https://apps.testrun.org and shows a picker

Closes #4463.

- ~~make URL configurable ?~~ not for now, see discussion at [#4463](https://github.com/deltachat/deltachat-desktop/issues/4463)
- [x] show extended details
- [x] add e2e test 

image

closes #4463


packages/frontend/src/components/composer/AppPickerWrapper.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  useEffect(() => {
+    const fetchApps = async () => {
+      try {
+        const response = await BackendRemote.rpc.getHttpResponse(
+          selectedAccountId(),
+          'https://apps.testrun.org/xdcget-lock.json'
+        )
+        const apps = await runtime.getJsonFromBase64(response.blob)
+        if (apps) {
```
nicodh, line 28: We could use the same function like in browser.runtime and thus implement that in the frontend, but I'm not really sure if that function is reliable.

Another solution would be to return a JSON string instead of a base64 encoded blob from backend, if the content-type is 'application/json'

packages/frontend/src/components/composer/Composer.tsx
```diff
@@ -390,6 +413,15 @@ const Composer = forwardRef<
             />
           </div>
         </div>
+        {showAppPicker && (
+          <OutsideClickHelper onClick={() => setShowAppPicker(false)}>
```
nicodh, line 417: This closes the Picker when a click outside happens. That is not nice for accessibility where no mouse might be involved. Not sure what is best practice here. We could also close the picker on "escape" key?

r10s:
nice! i did not expect to have sth. native here, however, it seems no to be too complex when looking at the line count only :) i also did not follow background discussions here, tbh. maybe that was the easier approach at the end

however, for

> make URL configurable

if somehow possible, it would be nice to support a "normal" website for that. otherwise the requirements for users to run alternative stores would be far higher - and would also be different for Android/iOS and Desktop.

the idea of a "configurable store" was also to make that as easy as possible - currently, it would be enough if a user throws some .xdc on a server and let apache/nginx render the directory, without thinking about how to generate a json

nicodh:
> detail view with descriptions is missing

Yes that can be added easily. There is an UI question though: is it nicer to have a 1-click to add it or a 2-click like on the website. Also if we should show just an overlay or a separat view (since Desktop is also used on small screens maybe both needs to be provided?) 

Maybe an info Icon showing the overlay - but a click on the app adds it directly?

packages/target-browser/runtime-browser/runtime.ts
```diff
@@ -367,6 +367,21 @@ class BrowserRuntime implements Runtime {
...
+      for (let i = 0; i < length; i++) {
+        bytes[i] = text.charCodeAt(i)
+      }
+      const decoder = new TextDecoder()
+      return JSON.parse(decoder.decode(bytes))
+    } catch (error) {
+      this.log.critical('String could not de decoded or parsed')
+      return null
+    }
+  }
```
WofWca, line 384: How is this different from `JSON.parse(atob(base64))`?

nicodh: If you test it with an base64 encoded json

it converts the resulting output:

Works:
```
const a = 'eyJhIjogIsOcw5bDhD8gVGVzdCDwn5GNIn0='
function getJsonFromBase64(base64) {
    try {
      const text = atob(base64)
      const length = text.length
      const bytes = new Uint8Array(length)
      for (let i = 0; i < length; i++) {
        bytes[i] = text.charCodeAt(i)
      }
      const decoder = new TextDecoder()
      return JSON.parse(decoder.decode(bytes))
    } catch (error) {
      console.log('String could not de decoded or parsed')
      return null
    }
  }
getJsonFromBase64(a)

Output: {a: 'ÜÖÄ? Test 👍'}
```
```
const a = 'eyJhIjogIsOcw5bDhD8gVGVzdCDwn5GNIn0='
 JSON.parse(atob(a))
 
 Ouput: {a: 'Ã\x9CÃ\x96Ã\x84? Test ð\x9F\x91\x8D'}
 ```

packages/frontend/src/components/composer/Composer.tsx
Simon-Laux: > We could also close the picker on "escape" key?

yes

Simon-Laux:
> otherwise the requirements for users to run alternative stores would be far higher

if you can host a web server, then you can also write a json file and host it as well. Or use xdcget to generate one.

Why "native"?
Because we don't want to maintain embedded browser views, the one in html email is annoying enough - and I'm not even sure if it can be replicated in tauri. Iframes have their own problems and don't fit nicely into the app.
And if you open an extra window then you could also just directly open a website or use the xdc store bot.
So better have an integrated picker that adheres to the UI design of the rest of desktop and does not need any native/runtime shenanigans like an embedded browser view or special iframe hardening.

Simon-Laux:
- [ ] On first open it showed "Offline" until I clicked on a tab (maybe select tools tab by default
- [ ] Icon in attachment menu is of a map, we should find sth more appropriate on https://fonts.google.com/icons (or maybe design a monochrome variant of webxdc logo? but I guess some icon symbolising puzzle, extension, tool or packet would be user friendlies)
- [ ] I believe the picker should open on the left side, as the context menu it was started from is also on the left side

packages/frontend/src/components/composer/AppPickerWrapper.tsx
nicodh: > since you didn't use atob/btoa

The solution uses atob but fixes the resulting "wrong" characters

I removed it now from runtime

nicodh:
> * On first open it showed "Offline" until I clicked on a tab (maybe select tools tab by default)

Are you sure it was related to the tab? I see it on first load if the icons are not yet loaded? I also changed the word to "Loading"

Simon-Laux (approved):

PR #4488, opened 2025-01-15
Order apps by date

### LINKED ISSUE #4487: Small refinements on app-picker
Thanks @nicodh and @Simon-Laux for landing the app-picker, great stuff!

two small nits: 

- sort order of app-list should be by `date` (even though it doesn't show in the primary list) 
- the first line of `description` is the title but it should not be shown in the popup
  (see eg https://webxdc.org/apps/#jagtalon-realtime-editor )

nicodh:
resolve #4487

### DIFF
--- a/packages/frontend/src/components/AppPicker/index.tsx
+++ b/packages/frontend/src/components/AppPicker/index.tsx
@@ -79,6 +79,9 @@ export function AppPicker({ className, onSelect, apps = [] }: Props) {
           newIcons[app.app_id] = `data:image/png;base64,${response.blob}`
         }
         app.short_description = app.description.split('\n')[0]
+        app.description = app.description.substring(
+          app.short_description.length + 1
+        )
         const url = new URL(app.source_code_url)
         app.author = url.pathname.split('/')[1]
         app.date = moment(app.date).format('LL')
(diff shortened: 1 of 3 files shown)


r10s:
playing around with this PR, it somehow breaks the descriptions; this is not the case on main branch.

to reproduce,  stage one .xdc and open the picker again - then the first description line is removed everywhere and replaced by the full description:



the short description "Lovely puzzle game" (see https://webxdc.org/apps/#r10s-packabunchas or app picker on first launch) has disappear completely:

nicodh:
> playing around with this PR, it somehow breaks the descriptions; this is not the case on main branch.

Good catch - I will fix that

nicodh:
> Looks great, but isn't the home tab missing? At least it's shown on the website and in DC Android.

Well - I considered it somehow useless since you can't select multiple categories anyway. But I have no strong opinion on that.  Now a click just toggles the tabs.

r10s:
>  But I have no strong opinion on that. Now a click just toggles the tabs.

well, it indeed would be better to have tabs that are mutually exclusive, as user is used to when it comes to tabs. allowing the user to "unselect" a tab (or a radio button) to get an overall view, is a bit weird and not what is expected normally.

it is not far fetched to assume, users do not understand how to get back to the "home"  view.

however, this is out of scope of this PR probably, and it also should not block releases. we can tweak in another iteration, it is more interesting to fix things as sorting and to release at all :)

nicodh:
Added a home category similar to the website

Simon-Laux (approved):

PR #4491, opened 2025-01-16
Webxdc draft preview

nicodh:
image

### DIFF
--- a/packages/frontend/src/components/attachment/messageAttachment.tsx
+++ b/packages/frontend/src/components/attachment/messageAttachment.tsx
@@ -23,6 +23,7 @@ import useTranslationFunction from '../../hooks/useTranslationFunction'
 import useDialog from '../../hooks/dialog/useDialog'
 import AudioPlayer from '../AudioPlayer'
 import { T } from '@deltachat/jsonrpc-client'
+import { selectedAccountId } from '../../ScreenController'
 
 type AttachmentProps = {
   text?: string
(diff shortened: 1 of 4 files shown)


r10s:
yeah! one step further towards preparing webxdc drafts (so starting apps and settings some things before sending)

packages/frontend/src/components/composer/Composer.tsx
```diff
@@ -440,12 +440,13 @@ export default Composer
...
 export type DraftObject = { chatId: number } & Pick<
   Type.Message,
-  'text' | 'file' | 'quote' | 'viewType' | 'vcardContact'
+  'id' | 'text' | 'file' | 'quote' | 'viewType' | 'vcardContact' | 'webxdcInfo'
 > &
   MessageTypeAttachmentSubset
 
 function emptyDraft(chatId: number | null): DraftObject {
   return {
+    id: 0,
```
WofWca, line 449: I'd prefer this to be nullable. But I guess it's fine.

packages/frontend/src/components/attachment/messageAttachment.tsx
```diff
@@ -217,6 +218,19 @@ export function DraftAttachment({
     )
   } else if (isAudio(attachment.fileMime)) {
     return <AudioPlayer src={runtime.transformBlobURL(attachment.file || '')} />
+  } else if (attachment.webxdcInfo) {
+    const iconUrl = runtime.getWebxdcIconURL(selectedAccountId(), attachment.id)
+    return (
+      <div className='media-attachment-webxdc'>
```
WofWca, line 224: The classes that this uses are defined in the `gallery` directory. Not super nice, but works for now I guess.

WofWca (approved):
Works alright!
The code is indeed simple. Very unlikely to break stuff.

And just FYI: the changelog entry is missing.

nicodh:
> And just FYI: the changelog entry is missing.

Thats how came up with the idea to remove that check :-)

PR #4492, opened 2025-01-16
Enable to skip changelog check for PRs

nicodh:
Try additional comment

#skip-changelog test

### DIFF
--- a/.github/mergeable.yml
+++ b/.github/mergeable.yml
@@ -3,15 +3,14 @@ mergeable:
   # see https://mergeable.readthedocs.io/en/latest/configuration.html
   - when: pull_request.*
     name: 'Changelog check'
-    filter:
-      - do: payload
-        pull_request:
-          description:
-            must_exclude:
-              regex: '\\s*skip changelog check\\s*'
     validate:
-      - do: and
+      - do: or
         validate:
+        - do: description
+          must_include:
+            regex: '#skip-changelog'
+        - do: and
+          validate:
           - do: dependent
             changed:
               file: 'packages/frontend/**'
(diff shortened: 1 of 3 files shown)


r10s:
ftr, [iOS](https://github.com/deltachat/deltachat-ios/pull/2515) and in the past the [core-rust](https://github.com/deltachat/deltachat-core-rust/pull/2999) were checking for the term `#skip-changelog` - which was inspired by some other thing iirc

this also its nicely to some added reasoning and is a bit better visible:

```
#skip-changelog as no user-visible things were changed, only variable renamed
```

just an inspiration, no blocker, i am fine with whatever :)

r10s (approved):
did not test, but if it works, it works :)

nicodh:
> did not test, but if it works, it works :)

It doesn't!! :-( 

But we are just discussing if we maybe skip the whole changelog check and add changes in the release only to avoid all these merge conflicts...

WofWca:
Just FYI you can edit the file right on GitHub to get immediate syntactic / semantic checking.

r10s:
> But we are just discussing if we maybe skip the whole changelog check and add changes in the release only to avoid all these merge conflicts...

yeah, this is what i was doing for android/iOS all the time the last years. it was quite annoying and time-consuming. meanwhile android/iOS releases are not only done by me, and i could not really convince others to do the same annoying stuff ;)

i also meanwhile think, it is better to think about the changelog line at the moment one is doing a PR

but yeah, whatever works for you or the desktop team is fine, you will know best. also on iOS, it is only an experiment that we may drop or replace by sth else :)

Issue #4430, opened 2024-12-20
Draft is not cleared after sending sometimes


missytake:
- Operating System (Linux/Mac/Windows/iOS/Android): xubuntu 22.04
- Delta Chat Version: 1.50.1
- Expected behavior: after I send a message, switch chat, and switch back to this chat again, the field should be empty.
- Actual behavior: sometimes the draft is not cleared; so while it disappears from the composer after sending, it is still visible in the chat list, and when I switch back and forth between chats, it appears in the composer again.
- Steps to reproduce the problem:
  - not always reproducible, always need to try a few times until it appears. I don't have a clear picture yet. but in principle:
  - type a message
  - make sure it appears as a draft in the chat list
  - send it
  - check whether it disappeared in the chat list or not

WofWca:
I'd say this is a duplicate of https://github.com/deltachat/deltachat-desktop/issues/3586, which got worse after https://github.com/deltachat/deltachat-desktop/pull/4144.

WofWca:
Is switching the chat a necessary step?

missytake:
> Is switching the chat a necessary reproduction step?

No, you can see it as a draft in the chat list (but not in the composer) before switching a chat, if it fails.

WofWca:
OK, I tested a bit. This is easier to reproduce if you type a lot and immediately send the message (easier with "Enter to send" as opposed to "Ctrl + Enter" to send). The draft will show up in the chats list, and then when you switch the chat back and forth, it will also appear in the composer, as @missytake said.

ell1e:
Perhaps I'm typing too fast, since this happens to me with 50% of my messages or so :eyes:

Issue #3586, opened 2023-12-07
sending a message is often delayed, sometimes buggy


hpk42:
am using a recent master of core and desktop. 

when i hit "ctrl-enter" for sending -- half the time it delays for 2+ seconds and sometimes the draft-text does not go away even after the message is posted (see screenshot).  It feels like something is blocking.  Currently, i can't offer much help in debugging but at least wanted to raise it, anyway. 

![Screenshot from 2023-12-07 13-36-11](https://github.com/deltachat/deltachat-desktop/assets/73579/fc4a298b-9db0-4daf-a1bb-82116de20c48)

Simon-Laux:
maybe send me the log after it happened, maybe there is some error somewhere?

WofWca:
This code is responsible for message sending and draft removal:

https://github.com/deltachat/deltachat-desktop/blob/a5c02433976909fe01a64f3049c96d0d101ae9d8/packages/frontend/src/components/composer/Composer.tsx#L140-L153

It calls `__reloadDraft`, which is where the input text is reset, I believe:

https://github.com/deltachat/deltachat-desktop/blob/a5c02433976909fe01a64f3049c96d0d101ae9d8/packages/frontend/src/components/composer/Composer.tsx#L448

hpk42:
Can't comment on the code but the issues still appears often for me -- on desktop i now take care especially in group chats, to very carefully check if the message went out, even if the input field continues to have the full draft message.  I then need to remove it manually.

WofWca:
Is it bad on all your Delta Chat accounts?

hpk42:
On Thu, Sep 19, 2024 at 23:59 -0700, WofWca wrote:
> Is it bad on all your Delta Chat accounts?

Only my (big) main accounts exposes the problem, and only in bigger group chats. 
In other accounts i am not really part of larger chats. 
Superficially, this points to some event-bubbling/timing problem? 
I also never had this problem on Android.

WofWca:
@Simon-Laux could you explain what's the point of `__reloadDraft()` instead of `clearDraft()` https://github.com/deltachat/deltachat-desktop/pull/2529 ?

Simon-Laux:
> @Simon-Laux could you explain what's the point of __reloadDraft() instead of clearDraft() https://github.com/deltachat/deltachat-desktop/pull/2529 ?

to show the actual state of the draft.

WofWca:
[...]

PR #4493, opened 2025-01-16
fix: draft not clearing after sending message

WofWca:
Commit-by-commit review should be easier, though anyway there are not many changes.

Closes https://github.com/deltachat/deltachat-desktop/issues/4430.

The bug was introduced in https://github.com/deltachat/deltachat-desktop/pull/4144.
There we made `throttle` accept the `text` argument
instead of reading `this.state.text`, and `this.state.text`
would get set to `''` after the message has been sent,
but then we'd still invoke the throttled function,
with the text from the last `onChange` event.

This probably doesn't address _all_ the "draft not cleared" issues,
e.g. https://github.com/deltachat/deltachat-desktop/issues/3586
could be a different issue.

This also fixes https://github.com/deltachat/deltachat-desktop/pull/4144#discussion_r1861963395.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 
 ## Fixed
 - fix draft not getting cleared after sending the message #4493
+- fix draft not getting saved after inserting an emoji #4493
 - fix chat "scrolls up" right after switching (rev 2) #4431
 - when deleting a message from gallery, update gallery items to remove the respective item #4457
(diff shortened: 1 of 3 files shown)


nicodh (approved):
Reviewed by reading & testing

Nice!

nicodh:
FYI: I was able to reproduce the bug on the main branch and could approve that this PR fixes it

PR #4494, opened 2025-01-16
Translations update

nicodh:


r10s (approved):
approving pro forma, assuming, this is the output of a script :)

PR #4495, opened 2025-01-16
Upgrade core to 1.54.1

nicodh:


WofWca (approved):
Just an upgrade.

Will need to double check the antivirus thing prior to release.

PR #4496, opened 2025-01-16
Auto focus when creating new profile

nicodh:


packages/frontend/src/components/screens/WelcomeScreen/InstantOnboardingScreen.tsx
```diff
@@ -147,6 +147,7 @@ export default function InstantOnboardingScreen({
             value={displayName}
             onChange={onChangeDisplayName}
             onBlur={saveDisplayName}
+            autoFocus={true}
```
nicodh, line 150: This has no effect. Why?

WofWca: Yes, we noticed that while working on the blueprint removal merge request. It looks like React is doing something funky.

WofWca: There are other places that use autofocus, but it has no effect there either.

nicodh: Maybe also some rerender thing...?

At least it works in the new profile form

WofWca: I'm sure that it really is autofocus that makes it work?
By default, the browser automatically focuses the first elements in the dialogue

nicodh: Ok, probably not. It was added 5 years ago by simon. Maybe obsolete since we removed blueprint?

nicodh: > Are you sure that it really is autofocus that makes it work?

Well I tested it twice. And I often experienced the missing auto focus when testing. But fact is also: there is no autofocus rendered in the source...??

WofWca: In bare HTML, Autofocus is supposed to work on dialogs.
But I think what is happening with React is it waits until it can mount and then it looks at all the autofocus elements and picks the first one using JavaScript.
But the problem is that it does not expect autofocus to have any effect after the first render.
This is what I think is happening.

nicodh: But can you confirm that this fixes the auto focus for a new profile name?

WofWca: I'll check later today.

WofWca: Indeed, for "add profile" it works, but not for "Edit profile"

WofWca (approved):
The changes make sense. The fact that some of them do not work is a separate problem perhaps.

PR #4502, opened 2025-01-17
update untranslated

r10s:
this PR removes strings from '_untranslated' that have a fitting counterpart or are not in use.

the PR can be merged unconditionally, but please test the affected dialogs quickly :)

most of the other untranslated strings were added at https://github.com/deltachat/deltachat-android/pull/3546 , in a subsequent  PR, after pulling, they can be used as well

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 ### Changed
+- add some missing translations
 
 ### Fixed
 
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -57,9 +57,6 @@
   "game": {
     "message": "Games"
   },
-  "app_author": {
-    "message": "Author"
-  },
   "app_size": {
     "message": "Size"
   },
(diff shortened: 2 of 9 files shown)


nicodh (approved):

r10s:
@nicodh i have no gool idea about how to pull translations from transifex, but i think you know that part. 

after that, `reset`, `pref_use_system_ui_font`, `muted`, `home`, `app_size`, `app_date_published` and `add_to_chat` can be deleted from untranslated.

for `games` and `tools` the key was changed to match the plural form, so that has to be adapted in the source code before delete the strings from untranslated

nicodh:
@r10s 

> after that, `reset`, `pref_use_system_ui_font`, `muted`, `home`, `app_size`, `app_date_published` and `add_to_chat` can be deleted from untranslated.
> 
> for `games` and `tools` the key was changed to match the plural form, so that has to be adapted in the source code before delete the strings from untranslated

I see https://github.com/deltachat/deltachat-android/pull/3546 is merged, but when pulling I don't get these strings. On https://app.transifex.com/delta-chat/delta-chat-app/ I don't find these strings too.

Issue #4508, opened 2025-01-18
jumpToMessage jumps to bottom instead of to message if message is in different chat


WofWca:
- Operating System (Linux/Mac/Windows/iOS/Android):
- Delta Chat Version: 1.51.0
- Expected behavior:
- Actual behavior:
- Steps to reproduce the problem:
  1. Search for messages. Make sure the search returns messages from different chats
  2. Click on a search result that is from a chat that is not the currently open chat
  3. The chat will open scrolled to bottom, regardless of where the target message is.
- Screenshots:
- Logs:

This also applies to "reply privately" quotes, and all the other `jumpToMessage` calls where the chat is different.

This is because we re-creae the `MessageListStore` every time `chatId` changes:

https://github.com/deltachat/deltachat-desktop/blob/c28d796ce5ba98d84905537c3b8514635f3680a2/packages/frontend/src/stores/messagelist.ts#L52-L56

But `jumpToMessage` claims to support jumping to message from a different chat:

https://github.com/deltachat/deltachat-desktop/blob/c28d796ce5ba98d84905537c3b8514635f3680a2/packages/frontend/src/stores/messagelist.ts#L406-L414

This appears to be a regression, caused by my performance changes, which surfaced the race condition. In 1.46 the bug is not present I think, at least sometimes.
Edit: on 1.50.1 it also works fine. Maybe a React or core upgrade surfaced this?

WofWca:
About 1/10 times this is still reproducible on 1.52.0

WofWca:
OK, after https://github.com/deltachat/deltachat-desktop/pull/4554 and https://github.com/deltachat/deltachat-desktop/pull/4562 I was not able to reproduce this. Let's test for a bit more and close. Then reopen if needed.

PR #4510, opened 2025-01-18
fix: `jumpToMessage` not working sometimes

WofWca:
...when the chatId of the message is different
from the currently open chat.

Closes https://github.com/deltachat/deltachat-desktop/issues/4508.
This is probably related to the React upgrade,
which changed the behaviour of `useEffect`.

The issue was that we called `window.__internal_jump_to_message`
_before_ `MessageListStore` has been reinstantiated,
so we'd call it on an old instance.

I think there might still be cases where this bug appears,
but at least this makes it appear less often.


WofWca:
Test failure seems unrelated.

nicodh (approved):
Review by testing & reading

PR #4511, opened 2025-01-19
fix: showing wrong chat ("Reply Privately" quote)

WofWca:
1. "Reply Privately" to a message in a group.
2. Click on the quote. This will open the original chat.
3. Click the "jump down" button.

This will load the messages from the private chat,
but still show the original group chat in the chat header.


nicodh (changes requested):
When testing I see constantly the (new) error log:

Tried to show messages from a different chat.
this.accountId === 1, this.chatId === 13, target IDs: 1, 14. jumpToMessageId === 81

I see that the correct group chat ID is fetched from backend (14) for the correct msgId from the quote (81), but when jump_to_message is performed still the chatId of the private chat is in this.chatId

nicodh (approved):
Review by Testing & Reading

PR #4514, opened 2025-01-19
feat: make daymarkers stick to the top

WofWca:
Remaining problems:
- Sometimes there is no daymarker above, i.e. when there were
  a lot of messages in a given day, and we haven't loaded
  the first message of the day yet.
- The background color could used some work on the light theme.
- The `min-width` is not assessed for different locales.


WofWca:
OK, let's close. I think we'll need to use some JS after all to solve the remaining issues.

PR #4518, opened 2025-01-20
fix: a11y: improve semantics

WofWca:
~~I suggest to merge this after 1.52 in order to not break stuff, therefore draft.~~ 1.52 is released


nicodh:
Can you add some comment to understand what this improves?

Simon-Laux (approved):

PR #4519, opened 2025-01-20
fix: fix roving tabindex for VCards

WofWca:


WofWca:
Test failure seems unrelated.

PR #4522, opened 2025-01-20
fix: improve input styles - less contrast to placeholders

### LINKED ISSUE #4520: Placeholder in input fields have too strong contrast

### LINKED ISSUE #4530: Bad UX/UI in contact QR screeen
when opening the contact-QR screen, there is a button that just says "copy" bellow the shown QR, the common user expectation is that the copy action will just copy the QR image to clipboard, we also improved this on Android, it would be better if there was a "chains" icon in front and/or the button is named "copy invite link" or so

another bad UX problem is that when user click the button, there is no feedback at all, the pop-up alert is not visible at all, hidden bellow the dialogs and by the time the user see it after closing all dialogs, it is just useless and needs to be manually clicked for it to disappear

nicodh:
resolves #4520 

Also changed the bottom-border to 1px width

Before:

Image

After:

image

#skip-changelog since it's just a small UI change

### DIFF
--- a/packages/frontend/scss/login/_login.scss
+++ b/packages/frontend/scss/login/_login.scss
@@ -157,7 +157,7 @@ div.delta-form-group {
       border-radius: unset;
       background: none;
       box-shadow: none;
-      border-bottom: 2px solid var(--deltaChatPrimaryFgLight);
+      border-bottom: 1px solid var(--deltaChatPrimaryFgLight);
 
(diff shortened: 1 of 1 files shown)


WofWca (approved):
Looks better to me.

r10s (approved):
definitely much better - so this is just the default? nice! even less lines of code :)

PR #4528, opened 2025-01-21
tauri: copy to clipboard

Septias:
Add the capability to copy images to the clipboard


packages/target-tauri/runtime-tauri/runtime.ts
```diff
@@ -348,8 +348,8 @@ class TauriRuntime implements Runtime {
   writeClipboardText(text: string): Promise<void> {
     return writeText(text)
   }
-  writeClipboardImage(_path: string): Promise<void> {
-    throw new Error('Method not implemented.20')
+  writeClipboardImage(path: string): Promise<void> {
```
Septias, line 351: @Simon-Laux What values can `path` have? From my experiments, I saw absolute paths but I guess this can also bea  relative blobdir path?

Simon-Laux: blobdir paths are always absolute - at least currently.

Simon-Laux:
could you rebase it to merge into https://github.com/deltachat/deltachat-desktop/pull/4533?
so that the rust method can be inside of `clipboard.rs`

Septias:
I changed the base to target your commit so you get a nicer diff on the web view, we can first merge it to your branch actually or wait until it is merged or merge it into main afterward.

Septias:
For testing, I have the image visible in my clipboard manager but can not paste it really. Might be Nixos though.

PR #4529, opened 2025-01-21
fix: chat scrolling up on quoting sometimes

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/3763 (again).
See https://github.com/deltachat/deltachat-desktop/issues/3763#issuecomment-2602630507.
Previous supposed fix: https://github.com/deltachat/deltachat-desktop/pull/4119.

Apparently the bug came back after the recent React upgrade, which was introduced in 1.51.0, so the bug is a regression.

I was unable to reproduce the bug on the Windows preview build from this MR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 ### Fixed
 - fix chat being scrolled up a little right after you switch to it (rev 3) #4521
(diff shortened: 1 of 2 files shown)


nicodh (approved):
Reviewed by testing

PR #4531, opened 2025-01-21
feat: display past chat members

link2xt:
Since the core now has a new group consistency algorithm, the list of past members who left the group in the last 90 days is sent around in the (encrypted) header. Displaying it makes users aware that this happens and past members are known even to the new members. Besides "security" aspect, the feature is just occasionally useful to check "have I removed the bot from the group already, or just cannot find it in the huge member list?".

For use together with https://github.com/deltachat/deltachat-core-rust/pull/6442

### DIFF
--- a/packages/frontend/src/components/dialogs/ViewGroup.tsx
+++ b/packages/frontend/src/components/dialogs/ViewGroup.tsx
@@ -120,6 +120,7 @@ function ViewGroupInner(
   const chatDisabled = !chat.canSend
 
   const groupMemberContactListWrapperRef = useRef<HTMLDivElement>(null)
+  const groupPastMemberContactListWrapperRef = useRef<HTMLDivElement>(null)
   const relatedChatsListWrapperRef = useRef<HTMLDivElement>(null)
 
   const {
@@ -133,7 +134,15 @@ function ViewGroupInner(
     setGroupImage,
   } = useGroup(accountId, chat)
 
+  const [pastContacts, setPastContacts] = useState<T.Contact[]>([])
+
   useEffect(() => {
+    BackendRemote.rpc
+      .getContactsByIds(accountId, group.pastContactIds)
+      .then(pastContacts => {
+        setPastContacts(Object.values(pastContacts))
+      })
+
     return onDCEvent(accountId, 'ContactsChanged', () => {
       BackendRemote.rpc
         .getContactsByIds(accountId, group.contactIds)
@@ -141,6 +150,12 @@ function ViewGroupInner(
           // update contacts in case a contact changed
           // while this dialog is open (e.g. contact got blocked)
           group.contacts = Object.values(contacts)
+        })
+
+      BackendRemote.rpc
+        .getContactsByIds(accountId, group.pastContactIds)
+        .then(pastContacts => {
+          setPastContacts(Object.values(pastContacts))
         })
     })
   }, [accountId, group])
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/dialogs/ViewGroup.tsx
```diff
@@ -136,6 +137,14 @@ function ViewGroupInner(
     setGroupImage,
   } = useGroup(accountId, chat)
 
+  const [pastContacts, setPastContacts] = useState([])
+
+  BackendRemote.rpc
+    .getContactsByIds(accountId, group.pastContactIds)
+    .then(pastContacts => {
+      setPastContacts(Object.values(pastContacts))
+    })
```
WofWca, line 146: This needs to be in `useEffect` in order to not call this on every re-render.
Perhaps also with `outdated` variable, as in https://github.com/deltachat/deltachat-desktop/pull/4500/files.

packages/frontend/src/components/dialogs/ViewGroup.tsx
```diff
@@ -316,6 +325,33 @@ function ViewGroupInner(
...
+                    <ContactList
+                      contacts={pastContacts}
+                      showRemove={false}
+                      onClick={contact => {
+                        if (contact.id === C.DC_CONTACT_ID_SELF) {
+                          return
+                        }
+                        setProfileContact(contact)
+                      }}
+                      onRemoveClick={showRemoveGroupMemberConfirmationDialog}
```
WofWca, line 349:
```suggestion
```

packages/frontend/src/components/dialogs/ViewGroup.tsx
```diff
@@ -316,6 +325,33 @@ function ViewGroupInner(
                 />
               </RovingTabindexProvider>
             </div>
+            {pastContacts.length > 0 && (
+              <>
+                <div className='group-separator'>
+                  {pastContacts.length} Past Members
```
WofWca, line 331: Needs to be translated.

WofWca (changes requested):
Nice feature!

To be clear, this is not yet supported by core.

packages/frontend/src/components/dialogs/ViewGroup.tsx
```diff
@@ -133,7 +134,15 @@ function ViewGroupInner(
...
+  const [pastContacts, setPastContacts] = useState<T.Contact[]>([])
+
   useEffect(() => {
+    BackendRemote.rpc
+      .getContactsByIds(accountId, group.pastContactIds)
+      .then(pastContacts => {
+        setPastContacts(Object.values(pastContacts))
+      })
+
     return onDCEvent(accountId, 'ContactsChanged', () => {
```
WofWca, line 146:
```suggestion
  useEffect(() => {
    BackendRemote.rpc
      .getContactsByIds(accountId, group.pastContactIds)
      .then(pastContacts => {
        setPastContacts(Object.values(pastContacts))
      })
  }, [accountId, group.pastContactIds])

  useEffect(() => {
    return onDCEvent(accountId, 'ContactsChanged', () => {
```

WofWca (approved):
Works nicely, after a bit of testing!
When removing members, they show up immediately in the "past members" section. Roving tabindex also works as intended. Same for clicking on members.

I left a small suggestion, but it's not important, feel free to ignore it.

PR #4535, opened 2025-01-22
refactor: apply `disableWidth` to `<AutoSizer>`s

WofWca:
Apparently it's not really necessary for vertical lists.
And we're already utilizing `disableWidth` in some other places.
So let's apply it to the rest where it's applicable,
in hopes to reduce code complexity.

After a bit of testing, this does not appear to affect behavior.

#skip-changelog


nicodh (approved):
> After a bit of testing, this does not appear to affect behavior.

For me neither

Issue #3671, opened 2024-02-08
Account Sidebar: Further Steps


Simon-Laux:
- [x] #3680 (assigned to @Simon-Laux for now)
- [x] context menu backdrop #3643 (@adzialocha)
	- this will solve the small ux bug that you can still trigger the hover menu with your mouse when a context menu is active
- [x] #3668 
- [x] #3670 
- [x] #3675
- [x] cursor pointer for account items (@adzialocha)
- [x] https://github.com/deltachat/deltachat-desktop/issues/3681 (@adzialocha)

## Beyond 1.44

- [x] on open: scroll currently active account into view
  https://github.com/deltachat/deltachat-desktop/pull/4137
- [ ] core: set sync for single accounts https://github.com/deltachat/deltachat-core-rust/pull/5314
    -  after this is done we can  remove the `window.__updateAccountListSidebar` hack, as this new api will replace the `syncAllAccounts` option.
- [X] ~~core: mute accounts? (notification is handled by ui so this would probably just store the setting/option and if we allow time based then revert once the mute time is over?)~~ -> rather "ui.*" setting
- [x] core: Drag and Drop to reorder accounts, needs core support to store position somehow
     - iOS already has move to top, which stores some sort number in the accounts -> https://github.com/deltachat/deltachat-ios/blob/dba6e6e2b3c5424d25403f488bbaf92965c901ff/deltachat-ios/DC/DcAccount.swift#L57
         - a possible downside to storing this in the account's ui config is that it will be exported in backups, I think this should be managed by the account manager and stored in `accounts.toml`
- [x] indicators on top and bottom if there are unread accounts out of screen
  Related: https://github.com/deltachat/deltachat-desktop/pull/4536
- [X] core: ensure there are core events for all cases where the account list is updated to remove the `window.__updateAccountListSidebar` hack
    - the only occurrence that is left of this is for syncAllAccounts setting, so moving the task of removing the hack to the sync setting in core
- [ ] Theming questions: should we make the sidebar themeable? should we allow each account to have it's own theme? some previous discussion can be found In: https://github.com/deltachat/deltachat-desktop/issues/3794

## nice to have
- [ ] Find an even better background image for account deletion

farooqkz:
Regarding, this:

> Drag and Drop to reorder accounts, needs core support to store position somehow

Doesn't core already support some mechanism of saving data no matter what?

Simon-Laux:
[...]

PR #4536, opened 2025-01-22
feat: make unread accounts `position: sticky`

WofWca:
https://github.com/user-attachments/assets/ad38fad5-4a87-4c73-be80-8ff2af772193

This somewhat addresses the
"indicators on top and bottom
if there are unread accounts out of screen"
point in https://github.com/deltachat/deltachat-desktop/issues/3671.

An alternative would be to put "unread" accounts at the top of the list (sort). Or to just show a "jump down" button, similar to how we do it in messages.
But this is a simple change I came up with, and it's not too often anyway that people have a dozen of accounts (profiles).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Added
 - highlight the first unread message upon opening a chat #4525
 - enable notifications on mentions in muted chats #4538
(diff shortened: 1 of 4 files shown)


nicodh (approved):
Reviewed by testing

PR #4537, opened 2025-01-23
fix: webxdc: Await calls to setUpdateListener() callback

ralphtheninja:
This allows an async callback to be passed to `webxdc.setUpdateListener()`.

Let me illustrate with an example by taking the simplest example from the webxdc.org page and tweaking it a bit. So imagine this app:

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8"/>
    <script src="webxdc.js"></script>
  </head>
  <body>
    <input id="input" type="text"/>
    <a href="" onclick="sendMsg(); return false;">Send</a>
    <p id="output"></p>
    <script>
      const numbers = []

      function sendMsg() {
        msg = document.getElementById("input").value;
        window.webxdc.sendUpdate({payload: msg}, 'Someone typed "'+msg+'".');
      }

      async function somethingReallyAsync () {
        return new Promise((resolve) => {
          setTimeout(() => {
            resolve()
          }, 100)
        })
      }

      async function doSomething (update) {
        const n = update.payload
        console.log('number payload', n)
        if (!numbers.includes(n)) {
          await somethingReallyAsync()
          numbers.push(n)
          document.getElementById('output').innerHTML += update.payload + "<br>";
        }
      }

      window.webxdc.setUpdateListener(async (update) => {
        await doSomething(update)
      }, 0);
    </script>
  </body>
</html>
```

It allows you to input numbers but only shows the ones you haven't already put in. So you can type like `1` and hit send a couple of times, then you just see a single `1`, then you type a `2` maybe and hit send a couple of times. It works.

Now, when you refresh the browser, it'll show everything you have put in. This means you can't have local state (this is just a javascript array) and do something async while updating that state (exemplified by `somethingReallyAsync()`).

If you instead `await callback()` this works.

Note that if you run this in the browser and you're using the local `webxdc.js` variant, the `setUpdateListener` method needs to change to something like:

image

### DIFF
--- a/packages/target-electron/static/webxdc-preload.js
+++ b/packages/target-electron/static/webxdc-preload.js
@@ -71,7 +71,7 @@ class RealtimeListener {
     )
     for (const update of updates) {
       last_serial = update.max_serial
-      callback(update)
+      await callback(update)
     }
     if (setUpdateListenerPromise) {
       setUpdateListenerPromise()


nicodh:
The change makes sense to me. Tthe webxdc API is not precise on the callback. In https://webxdc.org/docs/spec/setUpdateListener.html it says 
"_The returned promise resolves when the listener has processed all the update messages known at the time when setUpdateListener was called._"
That would require a "await" for async callbacks. 

@r10s  Looking at Android https://github.com/deltachat/deltachat-android/blob/main/src/main/res/raw/webxdc.js the callback is called without await.

So if we fix that in Desktop it should be updated there too.

And maybe update the API:
_With setUpdateListener() you define a  **(optional async)** callback that receives the updates sent by [sendUpdate()](https://webxdc.org/docs/spec/sendUpdate.html)_.

WofWca (commented):
- As @nicodh said, the spec needs to be updated.
- This is gonna be a breaking change for the apps. Need to consider how we want to handle this fact. One hypothetical situation where it's breaking is where one async update listener call awaits for another update listener call.
- Although this is also true for the current version of the webxdc spec, we probably need to say what should happen when an error occurs in the listener, or when the promise (thenable) rejects.

r10s:
@ralphtheninja can you add to the initial PR description why this change is needed and what is now possible what was not possible before.

esp. as this change requires quite some upfollownings and the impact to existing webxdc apps seems not 100% clear

also, this should not go to current release

ralphtheninja:
> @ralphtheninja can you add to the initial PR description why this change is needed and what is now possible what was not possible before.
> 
> esp. as this change requires quite some upfollownings and the impact to existing webxdc apps seems not 100% clear
> 
> also, this should not go to current release

I've updated the PR text to show more what the problem is.

ralphtheninja:
It's not really clear to me if just awaiting is good enough. What happens if the callback throws etc? This should be tested also. So maybe let this PR hang a bit here a while.

ralphtheninja:
Side note but related: Can the listener in `realtimeChannel.setListener(cb)` be async?

nicodh:
Closing since this needs to be decided and addressed in webxdc first.

WofWca:
I forgot to mention: there is a simple user-land workaround: just store the updates in an array, and then process them at your own pace.

Issue #4461, opened 2025-01-08
notify about "mentions", add "mention" switch


r10s:
in groups, Android and iOS meanwhile optionally notify about things that are **directly directed to you**, even if a group is otherwise muted. we call that "mentions". these are:
- replies to own messages
- reactions to own messages
- webxdc-notifications [^1]

there were all kind of additional discussions around these feature, **this issue is about bringing desktop to the level of iOS and Android.** steps needed:

- [ ] adapt the notifiy yes-or-no logic to the new flow ([flowchat here](https://github.com/deltachat/interface/pull/78#issuecomment-2536719734)). basically, to the existing conditions, add `if !isOneToOne AND isDirectedToYou and isMentionsEnabled { do_notify() }`  
see eg. [here](https://github.com/deltachat/deltachat-ios/pull/2434/commits/44e9e9bcaa362f8454ec5c758aaccec38f7b762b) and [here](https://github.com/deltachat/deltachat-ios/pull/2434/commits/cb0cd07ab6651eb573181a5c6f1510de0d417a54) for the logic changes on iOS

- [ ] add a toggle that to control `isMentionsEnabled` to the notification settings - for **screenshots** see  [here](https://github.com/deltachat/deltachat-android/pull/3456#discussion_r1866087382) and [here](https://github.com/deltachat/deltachat-ios/pull/2434)  
    - the strings are already there as `pref_mention_notifications` and `pref_mention_notifications_explain`
    - the toggle should default to ON
    - for storing the option, use `set/get_config("ui.mute_mentions_if_muted")`

- [ ] to make the profile settings easier understandable i would avoid the double negative default of "Mute Notifications = Off" and instead go for a simple "Notifcations = ON" as default, as shown on the screenshots linked above. that would also be more logic with the following "Mentions" toggle, that then can be disabled when notifications are disabled. but that point is not really a blocker, but would also help on [unify settings](https://github.com/deltachat/interface/issues/49)

finally, there is no urgency in this issue :) cc @adbenitez - please amend if i forgot something

[^1]: webxdc-notifications are currently always regarded as being a direct mention, so more important than the noise otherwise happening in groups, see discussions offline and eg. at https://github.com/deltachat/deltachat-android/pull/3456

PR #4538, opened 2025-01-23
Notify on mention

nicodh:
Depends on #4589

### DIFF
--- a/packages/frontend/src/components/Settings/Notifications.tsx
+++ b/packages/frontend/src/components/Settings/Notifications.tsx
@@ -17,18 +17,32 @@ type Props = {
 
 export default function Notifications({ desktopSettings }: Props) {
   const tx = useTranslationFunction()
+  const accountId = selectedAccountId()
+  const { accounts } = useAccountNotificationStore()[0]!
+  const isMuted = accounts[accountId]?.muted || false
+
+  function AccountNotificationMutedSwitch({
+    label,
+    description,
+  }: {
+    label: string
+    description?: string
+  }) {
+    return (
+      <SettingsSwitch
+        label={label}
+        value={isMuted}
+        description={description}
+        disabled={!desktopSettings['notifications']}
+        onChange={() => {
+          AccountNotificationStoreInstance.effect.setMuted(accountId, !isMuted)
+        }}
+      />
+    )
+  }
 
   return (
     <>
-      <SettingsHeading>{tx('pref_current_account')}</SettingsHeading>
-      <AccountNotificationMutedSwitch label={tx('menu_mute')} />
-      <DesktopSettingsSwitch
-        settingsKey='isMentionsEnabled'
-        label={tx('pref_mention_notifications')}
-        description={tx('pref_mention_notifications_explain')}
-        disabled={!desktopSettings['notifications']}
-      />
-      <SettingsSeparator></SettingsSeparator>
       <SettingsHeading>{tx('pref_all_accounts')}</SettingsHeading>
       <DesktopSettingsSwitch
         settingsKey='notifications'
@@ -39,29 +53,15 @@ export default function Notifications({ desktopSettings }: Props) {
         label={tx('pref_show_notification_content_explain')}
         disabled={!desktopSettings['notifications']}
       />
+      <SettingsSeparator></SettingsSeparator>
+      <SettingsHeading>{tx('pref_current_account')}</SettingsHeading>
+      <AccountNotificationMutedSwitch label={tx('menu_mute')} />
+      <DesktopSettingsSwitch
+        settingsKey='isMentionsEnabled'
+        label={tx('pref_mention_notifications')}
+        description={tx('pref_mention_notifications_explain')}
+        disabled={isMuted || !desktopSettings['notifications']}
+      />
     </>
-  )
-}
-
-function AccountNotificationMutedSwitch({
-  label,
-  description,
-}: {
-  label: string
-  description?: string
-}) {
-  const accountId = selectedAccountId()
-  const { accounts } = useAccountNotificationStore()[0]!
-  const isMuted = accounts[accountId]?.muted || false
-
-  return (
-    <SettingsSwitch
-      label={label}
-      value={isMuted}
-      description={description}
-      onChange={() => {
-        AccountNotificationStoreInstance.effect.setMuted(accountId, !isMuted)
-      }}
-    />
   )
 }
--- a/packages/frontend/src/components/Settings/Notifications.tsx
+++ b/packages/frontend/src/components/Settings/Notifications.tsx
@@ -21,26 +21,6 @@ export default function Notifications({ desktopSettings }: Props) {
   const { accounts } = useAccountNotificationStore()[0]!
   const isMuted = accounts[accountId]?.muted || false
 
-  function AccountNotificationMutedSwitch({
-    label,
-    description,
-  }: {
-    label: string
-    description?: string
-  }) {
-    return (
-      <SettingsSwitch
-        label={label}
-        value={isMuted}
-        description={description}
-        disabled={!desktopSettings['notifications']}
-        onChange={() => {
-          AccountNotificationStoreInstance.effect.setMuted(accountId, !isMuted)
-        }}
-      />
-    )
-  }
-
   return (
     <>
       <SettingsHeading>{tx('pref_all_accounts')}</SettingsHeading>
@@ -55,7 +35,14 @@ export default function Notifications({ desktopSettings }: Props) {
       />
       <SettingsSeparator></SettingsSeparator>
       <SettingsHeading>{tx('pref_current_account')}</SettingsHeading>
-      <AccountNotificationMutedSwitch label={tx('menu_mute')} />
+      <SettingsSwitch
+        label={tx('menu_mute')}
+        value={isMuted}
+        disabled={!desktopSettings['notifications']}
+        onChange={() => {
+          AccountNotificationStoreInstance.effect.setMuted(accountId, !isMuted)
+        }}
+      />
       <DesktopSettingsSwitch
         settingsKey='isMentionsEnabled'
         label={tx('pref_mention_notifications')}
--- a/packages/frontend/src/components/Settings/Notifications.tsx
+++ b/packages/frontend/src/components/Settings/Notifications.tsx
@@ -22,6 +22,12 @@ export default function Notifications({ desktopSettings }: Props) {
     <>
       <SettingsHeading>{tx('pref_current_account')}</SettingsHeading>
       <AccountNotificationMutedSwitch label={tx('menu_mute')} />
+      <DesktopSettingsSwitch
+        settingsKey='isMentionsEnabled'
+        label={tx('pref_mention_notifications')}
+        description={tx('pref_mention_notifications_explain')}
+        disabled={!desktopSettings['notifications']}
+      />
       <SettingsSeparator></SettingsSeparator>
       <SettingsHeading>{tx('pref_all_accounts')}</SettingsHeading>
       <DesktopSettingsSwitch
--- a/packages/frontend/src/system-integration/notifications.ts
+++ b/packages/frontend/src/system-integration/notifications.ts
@@ -1,12 +1,13 @@
 import { appName } from '../../../shared/constants'
 import { getLogger } from '../../../shared/logger'
+import { NOTIFICATION_TYPE } from '../../../shared/constants'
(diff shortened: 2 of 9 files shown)


packages/frontend/src/components/Settings/Notifications.tsx
```diff
@@ -33,29 +53,15 @@ export default function Notifications({ desktopSettings }: Props) {
...
       />
+      <SettingsSeparator></SettingsSeparator>
+      <SettingsHeading>{tx('pref_current_account')}</SettingsHeading>
+      <AccountNotificationMutedSwitch label={tx('menu_mute')} />
+      <DesktopSettingsSwitch
+        settingsKey='isMentionsEnabled'
+        label={tx('pref_mention_notifications')}
+        description={tx('pref_mention_notifications_explain')}
+        disabled={isMuted || !desktopSettings['notifications']}
+      />
```
nicodh, line 64: I changed the order here, to express the dependencies:

if notifications are turned off for all profiles, there is no need to mute the current profile etc.

image

WofWca:
This closes https://github.com/deltachat/deltachat-desktop/issues/4461, right?

packages/frontend/src/components/Settings/Notifications.tsx
```diff
@@ -33,29 +53,15 @@ export default function Notifications({ desktopSettings }: Props) {
         label={tx('pref_show_notification_content_explain')}
         disabled={!desktopSettings['notifications']}
       />
+      <SettingsSeparator></SettingsSeparator>
+      <SettingsHeading>{tx('pref_current_account')}</SettingsHeading>
+      <AccountNotificationMutedSwitch label={tx('menu_mute')} />
+      <DesktopSettingsSwitch
+        settingsKey='isMentionsEnabled'
```
WofWca, line 60: So this is not a core setting, correct?

packages/frontend/src/components/Settings/Notifications.tsx
```diff
@@ -17,12 +17,32 @@ type Props = {
...
+        label={label}
+        value={isMuted}
+        description={description}
+        disabled={!desktopSettings['notifications']}
+        onChange={() => {
+          AccountNotificationStoreInstance.effect.setMuted(accountId, !isMuted)
+        }}
+      />
+    )
+  }
```
WofWca, line 42: Apparently defining a component like this inline makes React create a new DOM element on every re-render, which is apparent when you toggle the value of this switch with keyboard: it will lose focus.

nicodh: Fixed it. There is anyway no need to render it inline

packages/frontend/src/components/Settings/Notifications.tsx
nicodh: Correct

packages/frontend/src/components/Settings/Notifications.tsx
r10s: hm, tbh, it was better before: as a rule of thumb, the whole settings dialog is about "Current Profile", apart from a few obvious exceptions (eg. task icon) and things that are not yet moved to profile (wallpaper).

therefore, it is not really fitting to put an exception to the top.

tbh, i also do not really get a user story or "Enable system notifications for new messages" that fits to user at scale at all.
- most users have only very few profiles
- most OS have a switch to disable notifications per-app (that would be the same, right?)
- if a user does not want to get disturbed, they would probably not disable delta-only, but just all or mute the sound
- _if_ we decide that is still needed, it would be more fitting eg. as a switch in the main menu, that is also easier accessible. "settings" on desktop are usually not a thing that you use on an every-day base

but maybe i also missed sth about  "Enable system notifications for new messages" :) 

just some thoughts, also to make the dialog more fitting to android/ios, 

however, this should not block this PR, i am also fine with the change, but at some point we may want to iterate over

nicodh: I agree that it's confusing to have a setting that applies to all profiles in this dialog. But I would argue as long as it is, the order should show the dependencies. If I first have the possibility to mute the notifications for the current profile and then to turn it off for all profiles, which overwrites the first setting it is even more confusing.

But that is not part of this PR (as you also mention).

So I think we should open a new issue:
Move settings for all accounts into a general settings dialog which can be opened from global menu and have only current profile related settings in the existing settings dialog which is opened from the UI

r10s: > Move settings for all accounts into a general settings dialog which can be opened from global menu and have only current profile related settings in the existing settings dialog which is opened from the UI

this was not my suggestion :)

this this known from some other apps, and that is always super-confusing. what is in which settings dialog ... in most cases, the user do not have that question and do not have the abstract view as we devs have.

but let's discuss that at some point one-to-one or in a small group in a call, all in all, the status quo is not bad, i just want to avoid it gets worse.

as said, for this PR, everything's fine

packages/frontend/src/system-integration/notifications.ts
```diff
@@ -260,41 +308,85 @@ async function flushNotifications(accountId: number) {
...
+        message.quote.messageId
+      )
+      if (quote.sender.id === C.DC_CONTACT_ID_SELF) {
+        log.info('mention detected: answer to own message')
+        return true
+      }
+    }
+    log.debug('ignoring notification on muted chat')
+    return false
+  }
```
WofWca, line 387:
```suggestion
  if (!SettingsStoreInstance.state?.desktopSettings.isMentionsEnabled) {
    log.info('allowIfMention: mentions are disabled')
    return false
  }

  if (notification.notificationType === NOTIFICATION_TYPE.WEBXDC_INFO) {
    log.info('mention detected: webxdc-info notification')
    return true
  }

  const message = await BackendRemote.rpc.getMessage(
    accountId,
    notification.messageId
  )

  if (
    notification.notificationType === NOTIFICATION_TYPE.REACTION &&
    message.sender.id === C.DC_CONTACT_ID_SELF
  ) {
    log.info('mention detected: reaction to own message')
    return true
  }

  if (
    message.quote &&
    message.quote.kind === 'WithMessage' &&
    (await BackendRemote.rpc.getMessage(accountId, message.quote.messageId))
      .sender.id === C.DC_CONTACT_ID_SELF
  ) {
    log.info('mention detected: answer to own message')
    return true
  }

  log.debug('ignoring notification on muted chat')
  return false
```

Less nesting. See if you like it more.

packages/frontend/src/system-integration/notifications.ts
```diff
@@ -260,41 +308,85 @@ async function flushNotifications(accountId: number) {
...
-    }
-  })
 
-  if (notifications.length > notificationLimit) {
+  const filteredNotifications = (
+    await Promise.all(
+      notifications.map(async notification => {
+        if (mutedChats.includes(notification.chatId)) {
+          // muted chat - only show if it's a mention and mentions are enabled
+          const isMention = await notificationIsMention(accountId, notification)
```
WofWca, line 317: I would prefer the `isMentionsEnabled` check to be here, but this is fine.

packages/frontend/src/system-integration/notifications.ts
```diff
@@ -260,41 +308,85 @@ async function flushNotifications(accountId: number) {
...
+      notification.messageId
+    )
+    if (notification.notificationType === NOTIFICATION_TYPE.REACTION) {
+      if (message.sender.id === C.DC_CONTACT_ID_SELF) {
+        log.info('mention detected: reaction to own message')
+        return true
+      }
+    }
+    if (message.quote && message.quote.kind === 'WithMessage') {
+      const quote = await BackendRemote.rpc.getMessage(
```
WofWca, line 376: TBH it feels like we're getting to a point where there are too many `await` and backend calls. This happens for every message on every muted chat. It doesn't sound like this is a job for the front end.

packages/frontend/src/system-integration/notifications.ts
```diff
@@ -260,41 +308,85 @@ async function flushNotifications(accountId: number) {
...
+    log.info('mention detected: webxdc-info notification')
+    return true
+  } else {
+    const message = await BackendRemote.rpc.getMessage(
+      accountId,
+      notification.messageId
+    )
+    if (notification.notificationType === NOTIFICATION_TYPE.REACTION) {
+      if (message.sender.id === C.DC_CONTACT_ID_SELF) {
+        log.info('mention detected: reaction to own message')
```
WofWca, line 371: Concerning my "too many backend calls" comment. I'm checking out `IncomingReaction` docs and MR: https://github.com/deltachat/deltachat-core-rust/pull/6072. It says:

> this PR adds an event for reactions received for one's own messages.

Which IMU means that we do not need to check if `message.sender.id === C.DC_CONTACT_ID_SELF`, because it's always `true`.

@r10s could you confirm?

WofWca (changes requested):

packages/frontend/src/system-integration/notifications.ts
nicodh: I tested it and it is true: IncomingReaction is only send to own messages

packages/frontend/src/system-integration/notifications.ts
nicodh: I agree the nesting be reduced a bit

packages/frontend/src/system-integration/notifications.ts
nicodh: I removed the SELF check

packages/frontend/src/system-integration/notifications.ts
nicodh: I removed the check and moved the condition up

r10s: yip, can confirm, it is also documented that way at [DC_EVENT_INCOMING_REACTION](https://c.delta.chat/group__DC__EVENT.html#gaef59044513b879b3a216e88d0dd2f5a7)

packages/frontend/src/system-integration/notifications.ts
```diff
@@ -260,41 +308,87 @@ async function flushNotifications(accountId: number) {
...
+              accountId,
+              notification
+            )
+            return isMention ? notification : null
+          }
+          return null
+        } else {
+          return notification
+        }
+      })
```
WofWca, line 328: This does not implement the flowchart https://github.com/deltachat/interface/pull/78#issuecomment-2536719734, because it doesn't perform the "is 1:1 chat" check.

packages/frontend/src/system-integration/notifications.ts
```diff
@@ -260,41 +308,87 @@ async function flushNotifications(accountId: number) {
...
-    }
-  })
 
-  if (notifications.length > notificationLimit) {
+  const filteredNotifications = (
+    await Promise.all(
+      notifications.map(async notification => {
+        if (mutedChats.includes(notification.chatId)) {
+          // muted chat - only show if it's a mention and mentions are enabled
+          if (SettingsStoreInstance.state?.desktopSettings.isMentionsEnabled) {
```
WofWca, line 317:
```suggestion
          // muted chat - only show if it's a mention and mentions are enabled
          // see https://github.com/deltachat/interface/pull/78#issuecomment-2536719734
          if (SettingsStoreInstance.state?.desktopSettings.isMentionsEnabled) {
```

packages/frontend/src/system-integration/notifications.ts
nicodh: Yes that's right. So another await Backend call... :-(

packages/frontend/src/system-integration/notifications.ts
```diff
@@ -260,41 +308,93 @@ async function flushNotifications(accountId: number) {
...
     }
   }
   notificationLimit = NORMAL_LIMIT
 }
 
+/**
+ * if isMentionsEnabled returns true
+ * if the notification is a mention
+ */
+async function notificationIsMention(
```
WofWca, line 365:
```suggestion
/**
 * returns true if the notification is a mention. See
 * https://github.com/deltachat/deltachat-desktop/issues/4461
 */
async function notificationIsMention(
```

The "if isMentionsEnabled" is no longer relevant.

WofWca (approved):
I tested (almost?) all the cases to see if they comply with the chat, and it seems they do.

I'm still not super happy with how many backend calls this introduced, so I hope we'll have all this stuff handled in core in the future. And this setting is disabled by default anyway.

The settings reorder makes sense to me. An input should not affect whether the inputs above it are disabled.
The "disabled" style change also seems good.

Please take a look at the last code suggestion, and let's merge!

PR #4539, opened 2025-01-23
fix: update unread badge on read from other device

WofWca:
We used to only update the count when calling
`BackendRemote.rpc.markseenMsgs()` or
`BackendRemote.rpc.marknoticedChat()`.


nicodh (approved):
Reviewed by reading

PR #4540, opened 2025-01-23
chore: update locales

nicodh:


WofWca (approved):
Just a translation update at a glance. Though I did not verify the translations themselves, or whether it matches the files in Delta Chat Android.

PR #4542, opened 2025-01-23
fix: scroll selected account into view on startup

WofWca:
The feature was introduced in 46d98111b69764b278f15c3dfe300e794dd64dcc,
but then broken apparently in d13a36cd10bcdb3c51d4f6cb0bd77a1a85f526a3.

As a side-effect, this will also show "placeholder" acount list items
while the accounts are loading.

And it makes the account list items interactive
while they're loading, i.e. you can select one
or activate the context menu.

Another thing this fixes is roving tabindex
while the accounts are loading, but isn't really important.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - fix log format for logging core events #4572
 - fix dragging files out
 - memory leak when opening and closing emoji picker #4567
+- fix selected account not getting scrolled into view in accounts bar on app start #4542
 - improve performance a little #4561
 
 <a id="1_52_1"></a>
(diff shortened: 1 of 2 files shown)


nicodh (commented):
Doesn't work for me. The selected account is not scrolled into view:

image

WofWca:
Is it on "clean" startup, i.e. after complete quitting?

WofWca:
If yes, could you also try the preview build? IIRC I also had trouble with this on the dev setup, but I forgot the reason.

nicodh (approved):
I tested now with the prebuild AppImage and it works as expected

PR #4551, opened 2025-01-24
refactor: rename `useChat().chatWithLinger`

WofWca:
So that the user of `useChat()` is more careful.

#skip-changelog


nicodh:
Can you suggest how to test that?

WofWca:
This is simply a rename. Testing this should be covered by the type checker.

nicodh (approved):

PR #4552, opened 2025-01-24
perf: improve performance

WofWca:
Review commit-by-commit recommended.


WofWca:
Rebased

nicodh (approved):
Reviewed by reading.

The changes for switchedChatAt from useState to useRef seem reasonable, although I did not really investigate the whole scroll behaviour

WofWca:
> The changes for switchedChatAt from useState to useRef seem reasonable, although I did not really investigate the whole scroll behaviour

Yes, the idea is that `switchedChatAt` is only read in the event listener, and does not affect the render output.

PR #4554, opened 2025-01-24
fix: improve `__jump_to_message` reliability

WofWca:
Follow-up to 6a656f968ac2bc2f67d6b9fbc1aba15d3851f053
(https://github.com/deltachat/deltachat-desktop/pull/4510).
Partially addresses https://github.com/deltachat/deltachat-desktop/issues/4508.
Why partially? Because the bug appears to still happen sometimes.
But this commit should make it more reliable, or at least be a refactor.

This should also improve performance in some cases,
thanks to the removal of `setTimeout()`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 ### Fixed
 - fix changelog message left unread not in the selected account as it should be but in another account. #4569
 - accessibility: some context menu items not working with keyboard navigation #4578
+- fix clicking on message search result or "Reply Privately" quote not jumping to the message on first click sometimes, again #4554
 - fix log format for logging core events #4572
 - fix dragging files out
 - memory leak when opening and closing emoji picker #4567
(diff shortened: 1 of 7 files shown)


nicodh (commented):
When reviewing this and the whole "jumpToMessage" logic, it seems we should clarify and/or document what is going on (perhaps improve the comment in jumpToMessage  in MessageListStore)

We have now:
a function _window.__internal_jump_to_message_
another function _window.__internal_check_jump_to_message_ 
an object _maybeJumpToMessageHack_ which is assigned to window.___internal_jump_to_message_asap_ which has a property _jumpToMessageArgs_ which holds a _addMessageIdToStack_ 

Maybe a list of use cases would help. Ideally some (for now manual) tests, to make sure changes do not break anything?

WofWca:
OK, I force-pushed the initial commit to clarify those `__internal` docs, and pushed another commit on top that adds docs to other `jumpToMessage` functions.

nicodh (approved):
Wow! Very accurate documentation! Thanks for that! Will help a lot in future development!

Tested randomly and so far everything worked as expected.

PR #4556, opened 2025-01-25
refactor: fix message list potentially being empty

WofWca:
...sometimes.
And maybe sometimes showing messages from previous chat.

I have not been able to confirm whether this fixes
an actual bug.
But at least in my understanding this fixes a semantic error.

The presence of the error is obvious if you make
`store.effect.loadChat()` synchronous.
So, what could be happening is that `loadChat` finishes _before_
we subscribe to the new store instance,
and thus the state never gets updated,
at least not until the next store update, which could be e.g.
`fetchMoreMessagesTop`.

The second commit also appears to be just a refactor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - fix dragging files out
 - memory leak when opening and closing emoji picker #4567
 - fix selected account not getting scrolled into view in accounts bar on app start #4542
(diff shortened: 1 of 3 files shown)


nicodh (commented):
Looks good to me and seems to make sense, although I would prefery to get this approved by @Simon-Laux since I'm not too familiar with the message store behaviour in detail

PR #4558, opened 2025-01-25
Add docker scripts to run e2e local

nicodh:


.dockerignore
```diff
@@ -0,0 +1,3 @@
+node_modules
```
WofWca, line 1: This only ignores the top-level `node_modules`, doesn't it? We have more `node_modules` per package.

Dockerfile.e2e
```diff
@@ -0,0 +1,20 @@
+FROM node:20.11
```
WofWca, line 1: Maybe rename this to something that includes "test"? "e2e" is not clear.

WofWca (commented):
So the idea is to add this as an extra option, and not make Docker the only way to run tests? If so, then it's alright.

.dockerignore
nicodh: Yes that's right. The other node_modules are very small (< 500k) compared to the big one in root. But maybe it's better to exclude them also.

WofWca:
```suggestion
**/node_modules
```

nicodh:
Closed in favour of #4563

PR #4560, opened 2025-01-25
perf: add useHasChanged2 (no `useEffect`)

WofWca:
Reduce the amount of `MessageListInner` re-renders,
as a fist step of using this function.

`useHasChanged2` is faster thanks to not using `useEffect`,
i.e. it doesn't cause an extra re-render. See
https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

Supersedes https://github.com/deltachat/deltachat-desktop/pull/4164.


WofWca:
> it seems to me a poor design that a library like react neewds a useHasChanged and a usePrevious hook and doesn't provide built-in functionality for that

You can argue that it's poor design on our part that we need such hooks. Fortunately there are not that many places left that need it.

PR #4561, opened 2025-01-25
perf: open chats faster (`loadChat()`)

WofWca:


packages/frontend/src/stores/messagelist.ts
```diff
@@ -354,12 +344,22 @@ class MessageListStore extends Store<MessageListState> {
...
               scrollIntoViewArg: { block: 'center' },
             })
-            ActionEmitter.emitAction(
-              chat.archived
-                ? KeybindAction.ChatList_SwitchToArchiveView
-                : KeybindAction.ChatList_SwitchToNormalView
-            )
           })
+
+          // TODO why do we only do this when `firstUnreadMsgId !== null`?
```
Simon-Laux, line 349: good question, maybe it was meant as an optimisation attempt. I think we can try to change it.

PR #4563, opened 2025-01-26
Run e2e tests in docker

nicodh:
Reduced Dockerfile

Use only Chrome as playwright project to reduce test duration: 
 - webkit does not yet support clipboard access in playwright
 - firefox will probably never be used (in tauri)


packages/e2e-tests/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+# Copy the package.json to install playwright
+# /app is overriden by the volume mount in docker-compose.yml
+# to the local WORKDIR
+COPY ./packages/e2e-tests/package.json /app/package.json
+
+WORKDIR /app
+
+USER root
+
+RUN git config --global --add safe.directory /app
```
WofWca, line 22: A comment would be nice.

packages/e2e-tests/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+RUN pnpm install
+
+USER root
```
WofWca, line 26:
```suggestion
```

We already switched to `root` a few lines above?

packages/e2e-tests/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+# move headless browsers to noone's home directory
+RUN cp -r /root/.cache /home/noone/
+
+RUN chown -R noone:noone /home/noone/.cache
+
+# Clean up to keep the image small
+RUN rm -rf /root/.cache/ms-playwright
+
+RUN rm -rf node_modules
```
WofWca, line 39: Is it just me or does this remove `node_modules` on the host machine?

packages/e2e-tests/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+RUN groupadd -g $gid -o noone
+RUN useradd -ms /bin/bash -u $uid -g $gid noone
+
+RUN npm install -g pnpm
+RUN pnpm config set store-dir /usr/local/pnpm-store
+
+# Copy the package.json to install playwright
+# /app is overriden by the volume mount in docker-compose.yml
+# to the local WORKDIR
+COPY ./packages/e2e-tests/package.json /app/package.json
```
WofWca, line 16: And doesn't this replace the root `package.json` with the `packages/e2e-tests/package.json` on the host machine?

docker-compose.e2e-tests.yml
```diff
@@ -0,0 +1,26 @@
...
+      dockerfile: packages/e2e-tests/Dockerfile
+      args:
+        gid: ${GID}
+        uid: ${UID}
+    environment:
+      - NODE_ENV=test
+      - NODE_TLS_REJECT_UNAUTHORIZED=0
+      - DC_ACCOUNTS_DIR=../../e2e-tests/data/accounts
+    volumes:
+      - ./:/app
```
WofWca, line 16: What is the point of having a bind mount? Not having to rebuild the image every time the code changes?

.dockerignore
```diff
@@ -0,0 +1,6 @@
+node_modules
+packages/target-tauri
+packages/target-electron
+packages/frontend/node_modules
+packages/shared/node_modules
+packages/e2e-tests/node_modules
```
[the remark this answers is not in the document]
nicodh: I just decided to rework this PR. We probably don't need dockerirgnore at all. The used docker image is far too gib (2.7 GB) - I will go for a smaller

packages/e2e-tests/Dockerfile
nicodh: Outdated

docker-compose.e2e-tests.yml
nicodh: Yes, the image should only be rebuilt if we change to another playwright version. And the current dir is mounted to /app/ in the container to have always the current sources

docs/E2E-TESTING.md
```diff
@@ -35,3 +35,18 @@ If you omit the project parameter the tests will be executed in all configured b
 The account dir for tests is in _packages/e2e-tests/data/accounts_
 
 It can be deleted after running tests and will be recreated in the next run.
+
+## Run e2e tests in docker
+
+Requirements: docker & docker-compose
```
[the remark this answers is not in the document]
nicodh: Good point. docker-compose is not neccessary anymore

docs/E2E-TESTING.md
```diff
@@ -35,3 +35,18 @@ If you omit the project parameter the tests will be executed in all configured b
...
+
+> ⚠ To avoid having data belonging to root in your working dir
+> you have to add you local user UID & GID to a .env file in packages/e2e-tests
+>
+> To get your UID/GID run `id`in a terminal
```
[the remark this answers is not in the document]
nicodh: Fixed

Simon-Laux:
> webkit does not yet support clipboard access in playwright

you could also do:
- ignore clipboard dependent tests on webkit but run the other tests
- provide a different runtime while the browser version is in testing -> a testing runtime.js that mocks the clipboard access (and let's you control it from playwright)

packages/e2e-tests/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+ARG uid
+ARG gid
+
+# git is needed for gatherBuildInfo.js in build process
+RUN apt update && apt upgrade -y && apt install -y git
+
+# add a user "noone" with the same uid and gid as the host user
+RUN groupadd -g $gid -o noone
+RUN useradd -ms /bin/bash -u $uid -g $gid noone
```
WofWca, line 11: I can see that this now doesn't override `node_modules`. Very nice. But still, what files does this write? Why do we need this? Is this the HTML report file that Playwright generates? Wouldn't it be better to keep it in the container instead of writing to host? Using the `[outputFolder` option](https://playwright.dev/docs/test-reporters#html-reporter).

Issue #4498, opened 2025-01-16
Adapt to file deduplication


Hocuri:
We want to deduplicate blob files by always setting the file name to be the hash of the file content. Core issue: https://github.com/deltachat/deltachat-core-rust/issues/6265, core PR: https://github.com/deltachat/deltachat-core-rust/pull/6332

To be done:
- [x] When the user opens an attachment, copy it to a temp file with the correct name first https://github.com/deltachat/deltachat-desktop/issues/4352
- [ ] Make sure that the actual name of the file isn't shown to the user or used to determine the file type (via the extension), instead `dc_msg_get_filename()` (C-FFI) or `MessageObject.file_name` (JsonRPC) needs to be used, similar to the Android PR at https://github.com/deltachat/deltachat-android/pull/3521. 
  - Search for usages of `MessageObject.file`, check where it's passed, and make sure that all it's never used to check the file extension, to pass it to an external application, or to show it to the user.
  - Look at places in the code where a filename is user-visible, like when viewing it in the UI or passing it to an external application (especially when sharing or opening it). Check where the filename comes from, and make sure that it comes from `filename`, not from `file`.
  - Test that 
  1. sharing to and from DC
  2. opening
  3. saving ("Export attachment")
  4. drafting and re-entering the chat
  5. opening the drafted file once more
  6. sending
  
     still works for a) Images b) Videos c) vCards (attached contacts) d) attached files e) webxdc's. Also that editing and previewing drafted images works, and that previewing drafted videos works. (I hope I didn't forget to test any combination)

Then Desktop will be ready for using core with https://github.com/deltachat/deltachat-core-rust/pull/6332 merged. Received files will be deduplicated then, but outgoing files not yet.
 
For deduplicating outgoing files, I will need to add a new JsonRPC API or modify the existing one to use `set_file_and_deduplicate()` instead of `set_file()` (or alternatively, one of you could do that). I will need to ask one of you how you are using the API so that I can design it in a way that works well for you. And then of course, you will need to create a Desktop PR similar to https://github.com/deltachat/deltachat-android/pull/3513 that uses this new/changed API.
- A PR to modify the existing API: https://github.com/deltachat/deltachat-core-rust/pull/6470/

nicodh:
With #4564  file deduplication is in core and now in desktop also. Which means, opening an attachment will reveal the "real" filename, but since the file extension is kept " determine the file type (via the extension" is still working.

[...]

PR #4564, opened 2025-01-26
Upgrade core to 1.155.1

nicodh:


WofWca:
To be clear, is it safe to upgrade core without addressing https://github.com/deltachat/deltachat-desktop/issues/4498?

PR #4565, opened 2025-01-27
chore(release): prepare release 1.52.1

nicodh:


nicodh:
> What about #4479? Do we skip it?

I assumed it's merged since long. I will merge it and force rebase this PR

WofWca (approved):
To recap. The new release, after 1.52.0:
- Fixes the "empty chat" bug.
- Fixes "saved messages" not getting received.
- Upgrades core to a version that might be detected as malware by some anti-viruses. But not Microsoft.
- Adds an explicit warning in case core gets quarantined.
- But also makes it exit on server process error.

I tested the build for just a bit, and saw no sudden exits.

Let's :shipit:.

PR #4567, opened 2025-01-27
chore: upgrade emoji-mart to fix memory leak

WofWca:
~1 MB per open-close cycle.

Apparently it's this issue: https://github.com/missive/emoji-mart/pull/868

I have confirmed that this fixes the memory leak. This is after ~50 clicks. Before:

image

After:

image


Simon-Laux (approved):
LGTM, didn't test

PR #4568, opened 2025-01-27
chore: upgrade Electron from 32 to 34

WofWca:
FYI breaking changes. Nothing much appears to apply to us, except perhaps macOS. Please re-check:
- https://www.electronjs.org/blog/electron-33-0#breaking-changes
- https://www.electronjs.org/blog/electron-34-0#breaking-changes

Closes https://github.com/deltachat/deltachat-desktop/issues/3416.

Appears to work after some basic testing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased][unreleased]
 
+> Due to the Electron update, macOS 10.15 (Catalina)
+> is no longer supported, macOS 11 (Big Sur) or later
+> is the new requirement.
+
 ### Added
 - highlight the first unread message upon opening a chat #4525
 - enable notifications on mentions in muted chats #4538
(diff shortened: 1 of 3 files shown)


nicodh (approved):
Reviewed by (smoke) testing

That MacOS 10.15 is not supported anymore is sad, but we can't change it.

WofWca:
> That MacOS 10.15 is not supported anymore is sad, but we can't change it.

Is this a big deal? Because we can upgrade to the latest 32, and it would still resolve the referenced issue.

nicodh:
> Is this a big deal? Because we can upgrade to the latest 32, and it would still resolve the referenced issue.

Well that would give us some weeks (maybe month) and then we have to upgrade anyway. So let's upgrade to 34 now

WofWca:
Electron fixed it https://github.com/electron/electron/pull/45472

PR #4569, opened 2025-01-27
store last account id in core by using selectAccount (which stores it in account manager/accounts.toml)

### LINKED ISSUE #4474: We still store lastAccount in desktop settings, is there a still reason why we don't use the selected_account from core's account manager?
I think in the past we didn't do it because there isn't an unselect account api yet.

we already have [`select_account `](https://github.com/deltachat/deltachat-core-rust/blob/cb43382896c3f2d5e86259b3fb293e85cab48a82/deltachat-jsonrpc/src/api.rs#L217-L219) and [`get_selected_account_id`](https://github.com/deltachat/deltachat-core-rust/blob/cb43382896c3f2d5e86259b3fb293e85cab48a82/deltachat-jsonrpc/src/api.rs#L223-L225), but at the moment they have todo comments to be removed because we don't use them.

While it is important that the jsonrpc api doesn't depend on the global selected account variable, I think it might make sense to use it to just store the last account. (to have less important state in the config.json / desktop settings file)

- This issue is similar to https://github.com/deltachat/deltachat-desktop/issues/4024

The actionable item is to find answer to the following questions:
- what would be needed besides an additional method to remove selection (like is the state that no account is selected even possible
- is what I'm proposing here even a good idea?
- are there any side effects of using the core variable, like core depending on this variable for something or core automatically setting/reading this variable?

Simon-Laux:
This fixes that the update device message is left unread in the wrong account (because that used the value from core already and it wasn't updated by desktop so it was wrong).

closes #4474


packages/frontend/src/ScreenController.tsx
```diff
@@ -99,8 +99,15 @@ export default class ScreenController extends Component {
   }
 
   private async _getLastUsedAccount(): Promise<number | undefined> {
-    const lastLoggedInAccountId = (await runtime.getDesktopSettings())
+    let lastLoggedInAccountId
+    const desktopSettingsLastAccount = (await runtime.getDesktopSettings())
       .lastAccount
+    if (desktopSettingsLastAccount) {
```
WofWca, line 105:
```suggestion
    if (desktopSettingsLastAccount != undefined) {
```

WofWca (approved):
I'm not really sure what how much core's `select_account()` is responsible for, but this MR makes sense, especially that we're already using `getSelectedAccountId()` in the codebase.

PR #4580, opened 2025-01-28
chore: disable E2E test workflows

WofWca:
This is letting other workflow failures go unnoticed because you just assume that the workflows will fail.


nicodh (approved):
I agree to disable them until the playwright upgrade from #4563 is merged, since that's the main reason for failing atm

PR #4582, opened 2025-01-29
fix: freshly sent "Saved Messages" not displaying

WofWca:
It sometimes momentarily shows "loading message" before
displaying the message properly, and sometimes,
if you send a lot of messages rapidly, doesn't load the messages
at all until you re-open the chat
or reload the message list otherwise.

Before:

https://github.com/user-attachments/assets/438bd0b9-65d2-4652-b648-0ad1d607ac81

After:

https://github.com/user-attachments/assets/0a242f20-11b8-4ec6-8543-2c077d16180f

This has probably been surfaced by the last core upgrade,
i.e. the thing that also introduced
https://github.com/deltachat/deltachat-core-rust/issues/6468.

### DIFF
--- a/packages/frontend/src/stores/messagelist.ts
+++ b/packages/frontend/src/stores/messagelist.ts
@@ -276,6 +276,27 @@ class MessageListStore extends Store<MessageListState> {
       }, 'messageChanged')
     },
     setMessageState: (messageId: number, messageState: number) => {
+      const messageLoaded = this.state.messageCache[messageId] != undefined
+      if (!messageLoaded) {
+        // This may happen when sending a message to "Saved Messages"
+        // on a new Chatmail account, where `MsgDelivered` would fire
+        // almost instantly after the send, even before `jumpToMessage`
+        // finishes for the new message.
+        // This results in jumpToMessage thinking that the message
+        // is already loaded, but in fact it would be just a husk
+        // of a Message object, with only the `state` property present.
+        //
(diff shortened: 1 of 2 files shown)


packages/frontend/src/stores/messagelist.ts
```diff
@@ -276,6 +276,27 @@ class MessageListStore extends Store<MessageListState> {
       }, 'messageChanged')
     },
     setMessageState: (messageId: number, messageState: number) => {
+      const messageLoaded = this.state.messageCache[messageId] != undefined
+      if (!messageLoaded) {
```
nicodh, line 280:
```suggestion
      if (this.state.messageCache[messageId] === undefined) {
```
nitpick: here i would consider the condition easier to read without variable

nicodh (approved):
Reviewed by testing & reading

packages/frontend/src/stores/messagelist.ts
```diff
@@ -276,6 +276,27 @@ class MessageListStore extends Store<MessageListState> {
       }, 'messageChanged')
     },
     setMessageState: (messageId: number, messageState: number) => {
+      const messageLoaded = this.state.messageCache[messageId] != undefined
+      if (this.state.messageCache[messageId] === undefined) {
```
WofWca, line 280:
```suggestion
      if (this.state.messageCache[messageId] == undefined) {
```

PR #4584, opened 2025-01-29
perf: improve `jumpToMessage` speed

WofWca:
- **refactor: remove outdated comment**
- **refactor: simplify jumpToMessage**
- **refactor: simplify jumpToMessage again**
- **perf: improve jumpToMessage speed**

### DIFF
--- a/packages/frontend/src/stores/messagelist.ts
+++ b/packages/frontend/src/stores/messagelist.ts
@@ -520,11 +520,9 @@ class MessageListStore extends Store<MessageListState> {
               )
               .filter(msgId => msgId !== C.DC_MSG_ID_LAST_SPECIAL)
             jumpToMessageId = items[items.length - 1]
-            const message = await BackendRemote.rpc.getMessage(
-              accountId,
-              jumpToMessageId
-            )
-            chatId = message.chatId
+            chatId = (
+              await BackendRemote.rpc.getMessage(accountId, jumpToMessageId)
+            ).chatId
             jumpToMessageStack = []
             highlight = false
           }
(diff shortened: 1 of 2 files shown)


packages/frontend/src/stores/messagelist.ts
```diff
@@ -493,14 +493,19 @@ class MessageListStore extends Store<MessageListState> {
...
         if (!accountId) {
           throw new Error('no account set')
         }
 
+        // As was said in this function's docstring,
+        // it should not be called for messages that are in a different chat,
+        // so we know the chatId in advance.
+        // However, let's keep the code that supports arbitrary chatId,
+        // which can be "enabled" by setting `chatIdPreset = undefined`.
+        const chatIdPreset: number | undefined = this.chatId
```
nicodh, line 506: You mean that there is (should?) be no use case at the moment, where this.chatId is not set and/or is different from the jumpToMessage.chatId ?
Since later we log an error if isMessageInCurrentChat is false.
If so, you leave the code since it might be useful if we extend the usage later or just as a fallback if this function is called "wrongly" for a message that belongs not to current chat?

WofWca: > You mean that there is (should?) be no use case at the moment, where this.chatId is not set and/or is different from the jumpToMessage.chatId ?

Unless there is a bug that I missed, currently we never call this `jumpToMessage` with a msgId chat belongs to a chat whose ID is not `this.chatId`.

> Since later we log an error if isMessageInCurrentChat is false.

Yes, that would be unexpected.

> If so, you leave the code since it might be useful if we extend the usage later or just as a fallback if this function is called "wrongly" for a message that belongs not to current chat?

I guess it's for the former case mostly.
But maybe also for the latter, because I am not sure if we'd want to just throw an error instead. In this regard I kept things as they were before me.

nicodh (approved):

PR #4587, opened 2025-01-30
improvement: better shortcuts

WofWca:
- **improvement: Ctrl + F to focus search**
- **improvement: Enter or ArrowDown to focus 1st item**
- **improvement: don't clear search on focus**
- **improvement: Ctrl+M to focus composer, not Ctrl+N**
- **feat: Ctrl + N to open "Create Chat"**
- **feat: add Ctrl + Shift + F to search in chat**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - shortcut improvements:
   - Ctrl + F now focuses search, instead of Ctrl + K
   - Ctrl + M now focuses composer, instead of Ctrl + N
+  - Ctrl + N now opens "New Chat" dialog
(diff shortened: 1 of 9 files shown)


nicodh (approved):
Good improvement!

Btw. when looking at the Shortcut Hints: on my Linux & MacBook the Shortcut to show the Hints is Strg + - not / 
(dash not slash)

UI: would be nice to have margins around the <keyboard elements and to make the dialog wider to allow 3 buttons in one row

Both can go into a separate PR of course

WofWca:
> on my Linux & MacBook the Shortcut to show the Hints is Strg + - not /
(dash not slash)

Hmmm. This is weird. I don't see how this could be the case.

PR #4588, opened 2025-01-31
refactor: improve styles in shortcut hint

nicodh:
- add alternative shortcut to open shortcut hints

Before:

image

After:

image

#skip-changelog

### DIFF
--- a/packages/frontend/scss/misc/_keyboard_hint_box.scss
+++ b/packages/frontend/scss/misc/_keyboard_hint_box.scss
@@ -64,7 +64,7 @@
     flex-wrap: wrap;
     flex-direction: column;
     padding: 4px;
-    max-height: 80vH;
+    max-height: 80vh;
 
     .action {
       width: calc(50% - 15px);
--- a/packages/frontend/scss/misc/_keyboard_hint_box.scss
+++ b/packages/frontend/scss/misc/_keyboard_hint_box.scss
@@ -47,10 +47,9 @@
 }
 
 .keyboard-hint-cheatsheet-dialog {
-  width: 100% !important;
-  max-width: 840px !important;
+  width: 80% !important;
+  max-width: 560px !important;
   .keybinding {
-    flex-grow: 1;
     font-size: 0.84em;
   }
   h4 .keybinding {
@@ -63,13 +62,12 @@
   .keyboard-hint-dialog-body {
     display: flex;
     flex-wrap: wrap;
-    padding: 4px;
-    max-height: 80vh;
-
+    padding-left: 10px;
+    padding-bottom: 10px;
     .action {
       width: 260px;
-      margin: 7px;
-      padding: 10px;
+      margin: 4px;
+      padding: 7px;
     }
 
     h2 {
--- a/packages/frontend/scss/misc/_keyboard_hint_box.scss
+++ b/packages/frontend/scss/misc/_keyboard_hint_box.scss
@@ -47,7 +47,8 @@
 }
 
 .keyboard-hint-cheatsheet-dialog {
-  width: 70% !important;
+  width: 100% !important;
+  max-width: 840px !important;
   .keybinding {
     flex-grow: 1;
     font-size: 0.84em;
@@ -62,18 +63,17 @@
   .keyboard-hint-dialog-body {
     display: flex;
     flex-wrap: wrap;
-    flex-direction: column;
     padding: 4px;
     max-height: 80vh;
 
     .action {
-      width: calc(50% - 15px);
+      width: 260px;
       margin: 7px;
       padding: 10px;
     }
 
     h2 {
-      max-width: 240px;
+      max-width: 280px;
       font-size: 1.4em;
     }
   }
--- a/packages/frontend/scss/misc/_keyboard_hint_box.scss
+++ b/packages/frontend/scss/misc/_keyboard_hint_box.scss
@@ -48,7 +48,7 @@
 
 .keyboard-hint-cheatsheet-dialog {
   width: 80% !important;
-  max-width: 560px !important;
+  max-width: 570px !important;
   .keybinding {
     font-size: 0.84em;
   }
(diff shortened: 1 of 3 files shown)


nicodh:
Of course the dialog now needs much more space, what can be considered worse...

packages/frontend/src/components/KeyboardShortcutHint.tsx
```diff
@@ -199,9 +199,19 @@ function Shortcut(action: ShortcutAction): CheatSheetEntryType {
...
+        <KeyboardShortcut elements={['Meta', '/']} />
+        <KeyboardShortcut elements={['Meta', '-']} />
+      </>
+    )
   } else {
-    return <KeyboardShortcut elements={['Control', '/']} />
+    return (
+      <>
+        <KeyboardShortcut elements={['Control', '/']} />
+        <KeyboardShortcut elements={['Control', '-']} />
```
WofWca, line 212: I don't understand this. `Ctrl + -` doesn't work for me, and I don't see how it should:

https://github.com/deltachat/deltachat-desktop/blob/c176996479646c719883a556a1f5cfc95bcbd31a/packages/frontend/src/keybindings.ts#L144-L146

WofWca (changes requested):
Looks much better on the screenshot, but on not-super-tall screens it's really weird.

https://github.com/user-attachments/assets/db5c8365-d111-4482-a4a7-875d150b8be2

I don't think it should be horizontally scrollable.

packages/frontend/src/components/KeyboardShortcutHint.tsx
nicodh: I think it's the german keyboard layout?
When testing here https://developer.mozilla.org/en-US/docs/Web/API/Element/keydown_event with a Linux Notebook, a Macbook Pro or another random german Windows Notebook, the "dash" key laways has the Slash key code. Although Wikipedia mentions the german keyboard has a ÷ sign which triggers the slash which is not the case for mine and (some others) keyboards.
Maybe we do a non-representative survey in Desktop dev channel?

WofWca: Better try it here, where both the key and the code are shown. https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code

I thought `code` is supposed to be layout-independent.

WofWca: Oh. Looks like on the Germain keyboard layout the "dash" key is where the Slash keys is on a US one. I get it now.

image

Well, this is a general issue that applies not just to this shortcut and not just to German. We should not add this hint unconditionally.

WofWca: I guess I could be to blame here, because I switched the check from `key` to `code` to ensure that we check the key location and not the printed character, so that the shortcut is independent of the layout. For example, we have `Ctrl + ,` to open the settings dialog, but on the Russian layout a comma is `Shift + Slash` and not `Comma`, so on the Russian layout the shortcut to open settings changes to `Ctrl + Shift + Slash`.

I'm for tackling this in a separate issue.

nicodh: The same in german keyboard layout: to have the key "/" you have to push Ctrl + Shift + /

So I will remove the second Key stroke for opening the shortcut hints.

nicodh:
> IDK about three columns. Seems too much. 

Well otherwise you have a narrow window but have to scroll a lot. Not what I would expect if I have a wide enough screen.

nicodh:
> This what it looks like now:

Looks weid. So you have a screen with less than 760px height, on the other hand no scrollbar?

WofWca:
> Looks weird. So you have a screen with less than 760px height, on the other hand no scrollbar?

No no, there is a scrollbar. I scrolled it to bottom, it's fine in this regard.

nicodh:
So I reduced the paddings and margins so that scrolling with 2 columns is less often used

packages/frontend/scss/misc/_keyboard_hint_box.scss
```diff
@@ -47,29 +47,31 @@
...
     display: flex;
     flex-wrap: wrap;
-    flex-direction: column;
-    padding: 4px;
-    max-height: 600px;
-    justify-content: space-evenly;
-    padding: 20px;
-
+    padding-left: 10px;
+    padding-bottom: 10px;
```
WofWca, line 66:
```suggestion
    justify-content: space-evenly;
    padding-left: 10px;
    padding-bottom: 10px;
```

Otherwise when there is one column it's kinda off IMO.

image

packages/frontend/scss/misc/_keyboard_hint_box.scss
```diff
@@ -47,29 +47,31 @@
...
+    flex: none;
+    margin-left: 20px;
+  }
+  .dialog-body {
+    padding-bottom: 0;
   }
   .keyboard-hint-dialog-body {
     display: flex;
     flex-wrap: wrap;
-    flex-direction: column;
```
WofWca, line 59: Don't you think the column order is better? To me it seems intuitive to read it top-to-bottom. And there are no sections that are extra wide:

image

WofWca (commented):
On 0.6x zoom there is always 1 column, no matter how wide the window is:

image

packages/frontend/scss/misc/_keyboard_hint_box.scss
nicodh: Ok, maybe better to have a centered view.

WofWca:
> Can't reproduce that

Maybe a different monitor DPR (zoom level) is needed.

packages/frontend/scss/misc/_keyboard_hint_box.scss
nicodh: If a set that it will result in just one long column which is definitely not better. Or we need 2 columns in the template. But tbh that (and this discussions about edge cases in a rarely used screen) seems a bit overdone for me.

But feel free to add a commit with better styling that fix your concerns :-)

WofWca: Oh. I get it. We have column order on `main`.

PR #4589, opened 2025-01-31
Upgrade core to 1.155.2

nicodh:


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 - highlight the first unread message upon opening a chat #4525
 
 ### Changed
+- Update `@deltachat/stdio-rpc-server` and `deltachat/jsonrpc-client` to `1.155.2`
+  - Deduplicate blob files in chat.rs, config.rs, and integration.rs.
```
WofWca, line 10: Is this non-breaking?

WofWca (commented):
I left a question, and the commit should say `1.155.2`. Otherwise it's just an upgrade.

nicodh:
> the commit should say `1.155.2`

Changed it

CHANGELOG.md
nicodh: It should not, since the file extensions are kept. It just might be confuse if the name is wrong (for example when downloading an attachement from a message). But that is already the case and it will be addressed in #4498 before the next release.

WofWca (approved):

PR #4594, opened 2025-02-02
perf: fewer `MessageList` re-renders

WofWca:


nicodh (approved):
So this is just a followup to #4593 but the changes are included

PR #4597, opened 2025-02-03
fix: arrow-key nav widgets being unfocusble

WofWca:
Reproduction steps:
1. Create a group with one more contact.
2. Open the "View Group" dialog.
3. Click "Add Members"
4. Arrow-key navigate to the member that is already
  part of the group.
  You'll notice that you pressed the arrow key,
  but the focus didn't change.
5. Press Tab.

Now you will not be able to focus the widget with Shift + Tab,
because the "active element"
(the only element that doesn't have tabindex="-1") is not focusable.


nicodh (approved):
Reviewed by testing & reading

Helpful comments!

PR #4600, opened 2025-02-03
refactor: very small refactor

WofWca:
- **refactor: rename var**
- **refactor: adjust `mapCoreMsgStatus2String` returns**

#skip-changelog

### DIFF
--- a/packages/frontend/src/components/chat/ChatListItem.tsx
+++ b/packages/frontend/src/components/chat/ChatListItem.tsx
@@ -84,12 +84,12 @@ const Message = React.memo<
   summaryPreviewImage,
   lastMessageId,
 }) {
-  const wasReceived =
+  const isIncoming =
     summaryStatus === C.DC_STATE_IN_FRESH ||
     summaryStatus === C.DC_STATE_IN_SEEN ||
     summaryStatus === C.DC_STATE_IN_NOTICED
 
-  const status = wasReceived ? '' : mapCoreMsgStatus2String(summaryStatus)
+  const status = isIncoming ? '' : mapCoreMsgStatus2String(summaryStatus)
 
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/chat/ChatListItem.tsx
```diff
@@ -84,12 +84,12 @@ const Message = React.memo<
   summaryPreviewImage,
   lastMessageId,
 }) {
-  const wasReceived =
+  const isIncoming =
```
nicodh, line 87: This sounds to me more like the process is not yet finished, like Loading...

What about justReceived?

If I look at the translations:
DC_STATE_IN_FRESH => 'delivered'
but the comment in https://c.delta.chat/group__DC__STATE.html says "Incoming fresh message" ?

WofWca: > This sounds to me more like the process is not yet finished, like Loading...

This is the purpose of the rename, to clear this confusion. These statuses are all the statuses that start with `IN_`, all the other ones are `OUT_`.

> If I look at the translations:

Where are you looking them up? In `mapCoreMsgStatus2String`? Because I think they're also wrong there, that's why I changed it.

nicodh: > Because I think they're also wrong there, that's why I changed it.
Ok I see

And the the icon in 

<div className={classNames('status-icon', status)} />

needs no update?

WofWca: That's right. See the commit message. The string returned from `mapCoreMsgStatus2String` is only used when the message is outgoing.

nicodh (approved):

PR #4602, opened 2025-02-04
use ci-chatmail server in e2e tests

nicodh:
#skip-changelog


WofWca (approved):
The workflow still runs, which means it's good!

Simon-Laux (approved):

PR #4604, opened 2025-02-05
Copy files to tmpdir

### LINKED ISSUE #4352: Attachmentfiles on devices are rw - changes to attachements are not signaled to the user by DeltaChat
- Delta Chat Version: 1.48

- Expected behavior:
If I click on an attachment (e.g. a LibreOffice document) sent to me via DeltaChat, DeltaChat either prompts me to save this to some folder on my device, or the attachment is opened directly in e.g. LibreOffice, but LibreOffice tells me, that the file is read-only and needs to be saved somewhere else before I can edit it.

- Actual behavior:
I click on an attachment in DeltaChat, attachment is opened directly in e.g. LibreOffice and I can edit & save directly to the attachment file in the DeltaChat storage location. Thus data between me and sender becomes inconsistent. DeltaChat does not inform me about the edited attachment.
Still worse is what happened to a friend of mine: He edited the document, saved it. The save was persistent, but only for a day or so. Then the changes in the document got reverted (by DeltaChat?) to the original version.

Open attachment in DeltaChat, edit, save (not "save as ..." but "save"). Then: reopen in DeltaChat. Edit is persistent, meaning attachment is not anymore in the state it was sent. DeltaChat does not inform user about this.
I was unable to reproduce the worse version on my machine. I believe what happened on my friends machine was due to him using a mail-server setup different from mine (him using the default DeltaChat-Server, me using my own mail provider).

- Screenshots: -
- Logs: -

- Comment: I know that one should save attachments to a different location before editing them, but there are many NOOBS out there and DeltaChats behaviour (see above) will get a lot of them in trouble. They should be able to use DeltaChat safely, I believe.

### LINKED ISSUE #4498: Adapt to file deduplication

nicodh:
resolves #4498 
resolves #4352 

Files are copied to accounts/accountId/tmp/ before opening.

Files that have an ID from deduplication as name are copied to a folder with that name, to assure multiple files with same name do not conflict _tmp/fileId/originalFilename.ext_

On any start of the app all tmp dirs are deleted again

I didn't find any cases for second point in that issue "Make sure that the actual name of the file isn't shown to the user " - but that might also be checked in reviews


packages/frontend/src/components/message/messageFunctions.ts
```diff
@@ -29,15 +29,17 @@ export function onDownload(msg: Type.Message) {
...
+  if (!msg.file || !msg.fileName) {
     log.error('message has no file to open:', msg)
     throw new Error('message has no file to open')
   }
-  if (!runtime.openPath(msg.file)) {
-    log.info(
-      "file couldn't be opened, try saving it in a different place and try to open it from there"
-    )
-  }
+  runtime.copyFileToInternalTmpDir(msg.fileName, msg.file).then(tmpFile => {
```
Simon-Laux, line 36: nitpick: you can also make this function async as it does not return anything (to avoid additional nesting that comes from using then)

Simon-Laux (approved):

nicodh:
> One possible concern is whether account deletion still works on windows if you currently have an attachment open in a different program?

The same for restarting the app with a file from tmp dir opened in some app

PR #4605, opened 2025-02-05
Upgrade core to 1.155.3

nicodh:
#skip-changelog


Simon-Laux:
would be nice to also update core for tauri in those update prs

Simon-Laux (approved):

PR #4606, opened 2025-02-05
fix: enable attach to node process

nicodh:
To be reviewed by someone who uses VScode.

Electron has to be started with pnpm debug:electron


WofWca (commented):
Is this supposed to replace the original configuration?

The original one did not work for me, looks like the path is no longer correct, now it's supposed to be `packages/target-electron/node_modules`.

But the new one works nicely, thanks!

Issue #4501, opened 2025-01-17
Fix race conditions with async `BackendRemote.rpc` & `useEffect`


WofWca:
- Operating System (Linux/Mac/Windows/iOS/Android): All
- Delta Chat Version: 1.51.0
- Expected behavior: Correct data for the respective item is shows.
- Actual behavior: The data corresponds to a wrong item.
- Steps to reproduce the problem:
- Screenshots:
- Logs:

See how https://github.com/deltachat/deltachat-desktop/pull/4500 fixes a race condition (supposedly resulting in https://github.com/deltachat/deltachat-desktop/issues/3978). We have a lot more of similar code which is also prone to races. To name a few:
- Searches. Basically all places where `queryStr` is used.
  - Global search (Ctrl + K), `useContactAndMessageLogic`
  - Search for contacts in "Create Chat" dialog

What we need to do is to check whether the promise is outdated when it gets resolved, to see whether we need to `setState`, as in https://react.dev/learn/you-might-not-need-an-effect#fetching-data and https://github.com/deltachat/deltachat-desktop/pull/4500.

How to search for _some_ occurrences:
- `useEffect\((.|\n){1,2000}?\sset\w{1,30}\(` - currently gives 56 occurrences across the codebase.
- `useEffect\((.|\n){1,2000}?\.then\(`
- `useEffect\((.|\n){1,2000}?await`
- `useEffect\((.|\n){1,2000}?BackendRemote.rpc`

r10s:
if this is not regarded as a bug, we should close this stale issue.

if it turns out to be advantageous to do some refactoring, it should just be done it on adding/improving a feature; it is usually not a task on its own and, as here, tends to get stale

PR #4615, opened 2025-02-08
fix: chat list & chat view showing different chats

WofWca:
Reproduction steps:
1. Select a chat with a lot of members, e.g. 100
  (use Ctrl + Tab for speed).
2. Immediately select another chat with few members,
  e.g. 1 on 1 chat (click it in the chat list for speed).

You'll notice that the chat list shows the chat from step 2,
but the messages list and the header shows the chat from step 1.

This issue belongs to the class of issues described in


nicodh (approved):
I couldn't really test, since I could not reproduce the race condition. At least I experienced no errors when testing.

But code looks good to me.

Issue #4559, opened 2025-01-25
message-parser bug: labeled links are parsed incorrectly


r10s:
we do not want to render markdown at the moment -- it is still too experimental, it is not our current focus, and it is not supported on all platforms. having that partly supported weakens overall UX. 

therefore, text should be rendered as entered.

however, for this is not aways true.

if one enters `this is **bold**, *italic*, [link](https://delta.chat)` one gets the following:

![Screenshot 2025-01-25 at 14 06 44](https://github.com/user-attachments/assets/f8f1d55c-e1a9-4ffb-9de2-c9b06dfbee50)

the link is rendered as mark down, but should not.

this is not a new issue, i just came over that when noticing another issue (as said, it is experimental or reasons, similar to location streaming :) - the other issue is that the upcoming dialog shows weird punycode, though the link does not require punycode:

![Screenshot 2025-01-25 at 14 10 52](https://github.com/user-attachments/assets/7a7f2dd5-070d-4d46-998f-7eb27e100766)

at some point, we may focus on markdown, but that day is not today :)

tested on mac 1.50.1

r10s:
hm, that seemed to be introduced unconditionally of the markdown switch at https://github.com/deltachat/deltachat-desktop/pull/1771 - but i find that very questionable. these kind of things affecting the whole ecosystems should be discussed platform comprehensive.

but well, i am find to leave link rendering as is then, and only fix the punycode issue

Simon-Laux:
you might mean this: https://github.com/deltachat/message-parser/issues/88
does also affect normal links, not only labeled links.

nicodh:
I think the confusion here comes from the fact that desktop ALWAYS parse labeled links, no matter if you have markup enabled in the settings or not.

There is even a test in the message parser for parse_desktop_set named _labeled_link_should_not_work_ which in contrary to the test name expects a _LabeledLink_ to be returned:
```
fn labeled_link_should_not_work() {
    assert_eq!(
        parse_desktop_set("[a link](https://delta.chat/en/help?hi=5&e=4#section2.0)"),
        vec![LabeledLink {
            label: vec![Text("a link")],
            destination: https_link_no_puny(
                "https://delta.chat/en/help?hi=5&e=4#section2.0",
                "delta.chat",
            )
        }]
    );
}
```

 https://github.com/deltachat/message-parser/blob/master/tests/text_to_ast/desktop_set.rs#L345

nicodh:
[...]

PR #4619, opened 2025-02-10
Show parts of new URL instead incoming values

nicodh:
resolves partly #4559 

This PR fixes two issues:

- punycode is shown to the user
- detect capital letters looking similar to other lower case letters like capital i looking like lower case L in some font-types (l != I)  and show a warning

It does not fix the issue that message parser is too greedy and adds some characters to domain names. (https://github.com/deltachat/message-parser/issues/91)

Before: opening a markup link with trailing )

image

After:

image

Clicking a link like https://deIta.chat just opened https://deita.chat

Now:

image

#skip-changelog

### DIFF
--- a/packages/frontend/src/components/message/Link.tsx
+++ b/packages/frontend/src/components/message/Link.tsx
@@ -144,7 +144,7 @@ function LabeledLinkConfirmationDialog(
                 .then(() => props.onClose())
             }}
           >
-            {tx('copy')}
+            {tx('global_menu_edit_copy_desktop')}
           </FooterActionButton>
           <FooterActionButton onClick={props.onClose}>
             {tx('cancel')}
--- a/packages/frontend/src/components/message/Link.tsx
+++ b/packages/frontend/src/components/message/Link.tsx
@@ -106,7 +106,7 @@ function LabeledLinkConfirmationDialog(
   const openLinkSafely = useOpenLinkSafely()
   const [isChecked, setIsChecked] = useState(false)
   const toggleIsChecked = () => setIsChecked(checked => !checked)
-
+  const url = new URL(props.target)
   return (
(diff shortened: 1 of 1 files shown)


Simon-Laux (changes requested):
showing the punycode to the user is intentional - it is a security feature to show the real url instead of what it appears to be: https://en.wikipedia.org/wiki/IDN_homograph_attack

the bug is in the message parser.

Simon-Laux:
Since there is also https://github.com/deltachat/message-parser/issues/92 we could ignore puny code completely for this dialog for now. but we should read it once it is fixed in the message parser -> so if we go that route then please add a comment on the issue I just linked to that we don't forget to re-add it

nicodh:
> showing the punycode to the user is intentional - it is a security feature to show the real url instead of what it appears to be

If any Destination coming from message parser which has a punycode property != undefined is an invalid or suspicious Link, then we should add a condition not go to LabeledLinkConfirmationDialog but always show a warning if it is present.

I saw in the source code that punycode is named PunycodeWarning in message parser, so maybe that would be a better name to understand "from outside" what's behind it


> the bug is in the message parser.

yes that's what I wrote in the description

nicodh:
Waiting for another solution Simon said he is planning to implement

Simon-Laux:
> Waiting for another solution Simon said he is planning to implement

comparing lower and uppercase would be too simple, we can decode and encode the url from/to puny code then we can compare the results of these operations, if they don't match we should show the warning.

For this I made a pr to expose the needed functions from the message parser: https://github.com/deltachat/message-parser/pull/95

nicodh:
Probably obsolete with https://github.com/deltachat/message-parser/pull/96

PR #4621, opened 2025-02-10
Update core to 1.155.4

nicodh:
Should be added to the test release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,10 +21,12 @@
 - display past members in the group member list #4531
 
 ### Changed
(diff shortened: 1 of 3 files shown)


r10s:
once merged, please check about that it really shows 1.155.4 - the tag was used two times at different commits, with a force-push or so in between (the difference is only the version shown in log/about, however, having that correct will ease later debugging and talking with users)

nicodh:
Is that also the case with the npm package? We only import that.

r10s:
idk, it is only important the the correct version is reported in about/log :)

nicodh:
It is. I checked that before.

Issue #4624, opened 2025-02-11
Rename copy & paste language keys


nicodh:
Currently we have several places where language keys from the global menu are used for the basic words "copy" & "paste".

We should have keys for that and only use keys like global_menu_edit_copy_desktop if there is a special usage in the global menu.

nicodh:
The question is, how we achieve that without having a time were these words appear untranslated in the UI.

The easiest way would be to rename the keys global_menu_edit_copy_desktop and global_menu_edit_paste_desktop and use the existing translations in all languages

nicodh:
I assume these keys are only used in desktop

r10s:
no, they the keys are also used elsewhere. in genered, we wanted to add the "_desktop" suffix only for strings that _differ_ from the corresponding android strings, but that was not taken into account that time.

i doubt, it is worth going the effort to rename keys, there are looooooots of inconsistences in there, and we would easily be busy with that for quite some time, including having new bugs introduced and overseen sth. note, also ubuntu touch is using the keys.

so, i would not be too buerocratic here, and accept it as it is :)

nicodh:
Ok, I agree on not touching running system here.

But should we also stick to that for new usage? So using using the "wrong" key even in more places? Or at least for new usage be more consistent?

Like introducing a new "copy" key to be used in all other places than general_menu. We have a lot of generics like on, off, yes, no, scan, open, delete etc. but not for cut, copy, paste ?

That would not affect existing stuff

r10s:
> But should we also stick to that for new usage?

yes, just use the old maybe-not-so-nice key also for new occurrences [^1]

otherwise, we would have doubled keys for a while then.

and as said, i fear, someone is unhappy with other keys then. and this is a can of worms ending in lots of drama ... i do not think, this all is worth the effort

[^1]: if we, at some point, decide to go for another key, one anyway need to search for all occurrences, having a handful more does not add much to that

PR #4625, opened 2025-02-11
chore: Rename copy paste language keys

nicodh:
See #4624 

#skip-changelog


r10s:
this will result in lots of untranslated text, or?

unless i've overseen sth, we should not do it that way. in general, i doubt it is worth the effort to renames keys, it is a bit ugly and grown, but well

PR #4533, opened 2025-01-22
tauri: implement `runtime.readClipboardImage` and TempFileFunctions, also fix pasting of (non-image) files into composer

Simon-Laux:
also contains a bugfix for electron -> fix pasting of (non image) files into composer by using web apis instead of `runtime.writeClipboardToTempFile`

todo:
- [x] read clipboard image
- [x] write tmp files 
- [x] delete tmp files

this pr also contains:
- #4528
- #4585
- 16a855da9ced6ab24d5a283dc84ed6dea629745c side effects for desktop settings & first part of content protection
- a46b5fb53319c41948ed1b47aad87b3410103693 fix bug with `LocaleData.locale`
- a49eef53ccc314b1629d357a6ba9e3738c255c30 help window localisation

PR #4626, opened 2025-02-11
fix: tauri: build failing (base64 dependency)

WofWca:
Apparently it started failing after https://github.com/deltachat/deltachat-desktop/pull/4533.

#skip-changelog


nicodh (approved):
Approved by result: fixes the failings checks

PR #4627, opened 2025-02-11
Remove spam for allowed and denied permissions

ralphtheninja:
The code was written in a way to reduce spamming when a permission was denied. But allowed permissions still had the same spammy behavior because Electron wants to call the permission handler a gazillion times.

This PR fixes so no spamming occurs, regardless if the permission is denied or allowed.

### DIFF
--- a/packages/target-electron/src/deltachat/webxdc.ts
+++ b/packages/target-electron/src/deltachat/webxdc.ts
@@ -62,6 +62,18 @@ const CSP =
   media-src 'self' data: blob: ;\
   webrtc 'block'"
 
+/**
+ * Allowed permissions for webxdc applications.
+ * https://www.electronjs.org/docs/latest/api/session#sessetpermissioncheckhandlerhandler
+ * https://www.electronjs.org/docs/latest/api/session#sessetpermissionrequesthandlerhandler
+ */
+const ALLOWED_PERMISSIONS: string[] = [
+  // Games might lock the pointer
+  'pointerLock',
+  // Games might do that too
+  'fullscreen',
+]
+
 const WRAPPER_PATH = 'webxdc-wrapper.45870014933640136498.html'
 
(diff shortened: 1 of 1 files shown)


packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -425,31 +437,21 @@ export default class DCWebxdc extends SplitOut {
         Exclude<Parameters<setPermissionRequestHandler>[0], null>
       >[1]
       const loggedPermissionRequests: { [K in permission_arg]?: true } = {}
```
ralphtheninja, line 439: Is this type really needed? It's just an object with strings as keys right?

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -425,31 +437,25 @@ export default class DCWebxdc extends SplitOut {
...
+        // Prevent webxdcs from spamming the log
+        if (!loggedPermissionRequests[permission]) {
+          loggedPermissionRequests[permission] = true
+          if (isAllowed) {
+            log.info(
+              `ALLOWED permission '${permission}' to webxdc '${webxdcInfo.name}'`
+            )
+          } else {
+            log.info(
+              `DENIED permission '${permission}' to webxdc '${webxdcInfo.name}'. If you think that's a bug and you need that permission, then please open an issue on github.`
```
ralphtheninja, line 453: Note that I'm using this on a single row for a reason. If you add a line break in there, it will mess up the log entry. All rows in the log should start the same, i.e. with the time.

ralphtheninja:
After a chat with @nicodh it looks like this only happens on the media permission (that we know of), so this PR isn't really needed to prevent spamming for the current state of desktop since we don't allow the media permission anyway.

So it only becomes a problem if you add something like this to the permission handler:

Screenshot from 2025-02-12 15-17-11

PR #4629, opened 2025-02-11
fix: "Show in Chat" in gallery not working

WofWca:
The bug was likely introduced in
https://github.com/deltachat/deltachat-desktop/pull/4554
(da4b9167513c76c5bc89e1302aedaf4cde020e40).
FYI the bug has only been released with 1.53.0.


nicodh (approved):
Review by reading & testing

Issue #4630, opened 2025-02-11
mentions should be ON by default


r10s:
this was overseen when targeting https://github.com/deltachat/deltachat-desktop/issues/4461

- apart from group replies, mentions were also ON on desktop before, this should not change
- also on android/iOS, the default is ON, desktop should do the same

PR #4632, opened 2025-02-12
Add new device message

nicodh:
#skip-changelog

### DIFF
--- a/packages/frontend/src/deviceMessages.ts
+++ b/packages/frontend/src/deviceMessages.ts
@@ -26,10 +26,10 @@ export async function updateDeviceChat(
     }
   }
 
-  await addDeviceMessage('changelog-version-1.53.0', {
-    text: `What's new in 1.53.0?
+  await addDeviceMessage('changelog-version-1.54.0', {
+    text: `What's new in 1.54.0?
 
-    🚀 Notifications on mentions in muted chats (configurable)
+    🚀 Notifications on mentions in muted chats (enabled by default)
     💥 Highlight the first unread message on opening chat
     🎹 New shortcuts for search & new chats:
        - Search: Ctrl/Cmd + F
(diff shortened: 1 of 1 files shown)


packages/frontend/src/deviceMessages.ts
```diff
@@ -26,13 +26,14 @@ export async function updateDeviceChat(
...
   }
 
-  await addDeviceMessage('changelog-version-1.52.0', {
-    text: `What's new in 1.52.0?
+  await addDeviceMessage('changelog-version-1.53.0', {
+    text: `What's new in 1.53.0?
 
-    🚀 New App Picker available in attachment menu! Shows apps from https://webxdc.org
-    🎹 Improved keyboard navigation
+    🚀 Notifications on mentions in muted chats (configurable)
```
WofWca, line 32: FYI https://github.com/deltachat/deltachat-desktop/issues/4630

packages/frontend/src/deviceMessages.ts
```diff
@@ -26,13 +26,14 @@ export async function updateDeviceChat(
     }
   }
 
-  await addDeviceMessage('changelog-version-1.52.0', {
-    text: `What's new in 1.52.0?
+  await addDeviceMessage('changelog-version-1.53.0', {
+    text: `What's new in 1.53.0?
```
WofWca, line 30: Why 1.53 and not 1.54?

WofWca (commented):
Otherwise I don't think anything else is worth adding.

packages/frontend/src/deviceMessages.ts
Simon-Laux: still test release I think

Simon-Laux (approved):

Simon-Laux (approved):

PR #4635, opened 2025-02-13
Sanitize filename

nicodh:
#skip-changelog

### DIFF
--- a/packages/target-electron/src/ipc.ts
+++ b/packages/target-electron/src/ipc.ts
@@ -8,7 +8,15 @@ import {
   shell,
   NativeImage,
 } from 'electron'
-import path, { basename, extname, join, posix, sep, dirname } from 'path'
+import path, {
+  basename,
+  extname,
+  join,
+  posix,
+  sep,
+  dirname,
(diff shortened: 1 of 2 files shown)


packages/target-electron/src/ipc.ts
```diff
@@ -374,6 +382,8 @@ export async function copyFileToInternalTmpDir(
 ): Promise<string> {
   const sourceFileName = basename(sourcePath)
   const sourceDir = dirname(sourcePath)
+  // make sure fileName includes only a file name, no path or whatever
+  fileName = basename(normalize(fileName))
```
ralphtheninja, line 386: Why is `normalize` needed? Shouldn't `basename()` be enough?

nicodh: was not sure if basename can be "tricked" if it gets an unexpected argument, while normalize is obviously created for "normalizing" but may be I exxagerate...

ralphtheninja: All good. Better safe than sorry!

PR #4637, opened 2025-02-13
fix source code links in app picker

### LINKED ISSUE #4631: Links in app picker don't work
If you select an app and click on the link to the source code, nothing happens.

macOS 15.3.1
DC-Desktop 1.53

Simon-Laux:
fix #4631


WofWca (commented):
This makes it work.

Though I wonder if we want a punycode warning as in here?:

image

Simon-Laux:
> Though I wonder if we want a punycode warning as in here?:

we can make an issue for that, good idea.

maybe we should also expose a separate api in message parser lib to check links for puny code.

nicodh (approved):

Issue #4468, opened 2025-01-09
Tauri: Meta: Postponed Ideas


Simon-Laux:
- [ ] clean shutdown (maybe copy cancel guard stuff from dc-rpc-server?)

- [ ] even better source code structure in the rust code
- [ ] Use channel for jsonrpc extra performance? I asked in discord, right now I think its only one direction backend to frontend -> could already be useful for that one direction backend to frontend, because that direction has more data anyway

## security

- [ ]  security -> try to find a way to get rid of (or disable or lock down) the global reference `__TAURI_INTERNALS__`, that it can only be called from the runtime interface implementation
- [ ] Find out how/if the Isolation pattern with the hook script can help us to harden the app (I can't think of a way it would help in our use case)

## less priority

- [ ] for blobs scheme - find out whether we can stream big files
	- streaming for videos and larger audio blob files:
		- https://github.com/tauri-apps/tauri/blob/fdaf48fc4a706b968e92b2b44e3830c51a5be1fe/examples/streaming/main.rs#L168
- [ ] try out https://docs.rs/specta/latest/specta/index.html - for typesafe tauri api (`specta` feature)
      - but that's probably not needed in our case?

PR #4641, opened 2025-02-14
[Tauri] Menu bars

Septias:
Add functionality for main menu in tauri backend.

- [x] Dialogs
- [x] Edit view
- [x] Developers
- [x] Show devtools open only if setting is set
- [x] Delete action 
  -  https://github.com/tauri-apps/tauri/issues/12718
- [x] Checkboxes for zoom level
- [x] Mac mainbar
- [x] Float on top

close https://github.com/deltachat/delta-tauri-todo/issues/7
close https://github.com/deltachat/delta-tauri-todo/issues/8

I currently have the problem that `config.json` store is not available during initialization and I can not check the proper value of the zoom level. Can we fix that @Simon-Laux or is it anyways not persisted? I think it would be nicer If we persisted it with actual managed state. I can also not check the mac mainbar.

### DIFF
--- a/packages/target-tauri/src-tauri/src/help_window.rs
+++ b/packages/target-tauri/src-tauri/src/help_window.rs
@@ -1,5 +1,5 @@
 use log::warn;
-use tauri::{Manager, WebviewWindow};
+use tauri::{Manager, Runtime};
 
 #[derive(Debug, thiserror::Error)]
 pub(crate) enum Error {
@@ -17,8 +17,8 @@ impl serde::Serialize for Error {
 }
 
 #[tauri::command]
-pub(crate) fn open_help_window(
-    app: tauri::AppHandle,
+pub(crate) fn open_help_window<A: Runtime>(
+    app: tauri::AppHandle<A>,
     locale: &str,
     anchor: Option<&str>,
 ) -> Result<(), Error> {
@@ -30,7 +30,7 @@ pub(crate) fn open_help_window(
 
     let app_url = tauri::WebviewUrl::App(url.into());
 
-    let mut help_window: WebviewWindow = if let Some(help_window) = app.get_webview_window("help") {
+    let mut help_window = if let Some(help_window) = app.get_webview_window("help") {
         help_window
     } else {
         tauri::WebviewWindowBuilder::new(&app, "help", app_url.clone()).build()?
--- a/packages/target-tauri/src-tauri/src/lib.rs
+++ b/packages/target-tauri/src-tauri/src/lib.rs
@@ -1,8 +1,8 @@
 use std::time::SystemTime;
 
 use clipboard::copy_image_to_clipboard;
-use settings::load_and_apply_desktop_settings_on_startup;
 use menus::main_menu::create_main_menu;
+use settings::load_and_apply_desktop_settings_on_startup;
 use state::{app::AppState, deltachat::DeltaChatAppState};
 use tauri::Manager;
 mod app_path;
--- a/packages/target-tauri/src-tauri/src/lib.rs
+++ b/packages/target-tauri/src-tauri/src/lib.rs
@@ -191,14 +191,6 @@ pub fn run() {
                 webview.set_title("")?;
             }
 
-            let zoom_factor = app
-                .handle()
-                .get_store("config.json")
-                .and_then(|store| store.get(ZOOM_FACTOR_KEY))
-                .and_then(|f| f.as_f64())
-                .unwrap_or(1.0);
-
-            set_zoom(&app.handle(), zoom_factor, "main")?;
             app.set_menu(create_main_menu(app.handle())?)?;
             app.on_menu_event(handle_menu_event);
 
--- a/packages/target-tauri/src-tauri/src/lib.rs
+++ b/packages/target-tauri/src-tauri/src/lib.rs
@@ -71,6 +71,7 @@ pub fn run() {
     let startup_timestamp = SystemTime::now();
 
     let mut builder = tauri::Builder::default()
+        .plugin(tauri_plugin_opener::init())
         .plugin(tauri_plugin_store::Builder::new().build())
         .plugin(tauri_plugin_clipboard_manager::init())
         .plugin(tauri_plugin_dialog::init())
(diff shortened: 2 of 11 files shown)


Simon-Laux:
> Show devtools open only if setting is set

what do you mean by setting is set? you could base it on the cargo feature for now, later it could be controlled by cli argument

packages/target-tauri/src-tauri/src/help_window.rs
```diff
@@ -17,8 +17,8 @@ impl serde::Serialize for Error {
 }
 
 #[tauri::command]
-pub(crate) fn open_help_window(
-    app: tauri::AppHandle,
+pub(crate) fn open_help_window<A: Runtime>(
+    app: tauri::AppHandle<A>,
```
Simon-Laux, line 21: why is this needed?

packages/target-tauri/src-tauri/tauri.conf.json5
```diff
@@ -62,7 +62,9 @@
             // open current logfile & open links
             "shell:allow-open",
             // Badge counter
-            "core:window:allow-set-badge-count"
+            "core:window:allow-set-badge-count",
+            "core:menu:default",
+            "core:menu:allow-new"
```
Simon-Laux, line 67: please add comment for what this is needed

Simon-Laux: is this even needed? as far as I understand it these capabilities limit only what you are able to call from javascript.

packages/target-tauri/src-tauri/src/menus/main_menu.rs
```diff
@@ -0,0 +1,278 @@
...
+                            None::<&str>,
+                        )?,
+                    ],
+                )?,
+            ],
+        )
+    });
+    builder.on_menu_event(|app, event| {
+        match event.id().as_ref() {
+            "settings" => {
```
Simon-Laux, line 189: maybe you could make constants or a string enum for these, to prevent typos between the two instances of the string.
enum would especially make sense, because then we could have a match statement that ensures all cases are handled.

packages/target-tauri/src-tauri/src/lib.rs
```diff
@@ -87,6 +90,8 @@ pub fn run() {
             }));
     }
 
+    builder = create_main_menu(builder);
+
```
Simon-Laux, line 94: you don't need to define it here, you can do it later like this:

```rs
    window.set_menu(menu)
        // for macOS use this and have some switch logic to switch the menu based on focused window
        // (implemented as tauti plugin or probably also would work as state)
        app.set_menu(menu)
```

besides that there are quite some cases where it needs to be updated/rebuild:
- on MacOS on changing window
- when translations change (@nojisan will make port the translation function to rust, so we can also use it in the backend)
- maybe also  when some kind of values change?

Simon-Laux:
> I currently have the problem that config.json store is not available during initialization and I can not check the proper value of the zoom level

you can define it after initialisation, after the store is loaded. makes more sense anyway, because then the same function would be used to rebuild/update it is for example the language changes

packages/target-tauri/src-tauri/src/help_window.rs
Septias: To make typechecker happy. I call this from the menu where Apphandle is more abstract

packages/target-tauri/src-tauri/src/lib.rs
Septias: I still need the zoom level on startup to set the proper check in the zoom settings. When adding the menu later I firstly don't see where a good location for that would be and users could see bad state.

Septias:
Can I add nix in this PR? Nix sadly requires you to have the flake and lock in source control. I can always keep them in the staged area but it would be nice for me to add them. Plus others nix users immediately get a working environment.

packages/target-tauri/src-tauri/src/lib.rs
Simon-Laux: try it out, I bet it is so fast that the other menu is not even noticed by anyone during startup. the location for it should be the place where delta core and so on is loaded: somewhere at the end of the setup function/closure: `.setup(move |app| {`

Septias:
I think they will not add the delete action in Tauri, can we just remove that entry from the menu @Simon-Laux?

Simon-Laux:
> I think they will not add the delete action in Tauri, can we just remove that entry from the menu @Simon-Laux?

I think we could ignore it for now, it is not important. (I never used it and it doesn't even have a keyboard shortcut on macOS)

Though similarly to https://github.com/deltachat/deltachat-desktop/issues/4232 we could make a list or an issue of what we omitted in the tauri version.

Septias:
Okay

Simon-Laux:
turns out we already have an issue for postponed tauri ideas, you may add it there https://github.com/deltachat/deltachat-desktop/issues/4468

Septias:
@Simon-Laux PR is mostly done now, only thing missing is mac support which I can not add as I don't have one.

packages/target-tauri/src-tauri/src/menus/mod.rs
```diff
@@ -0,0 +1,229 @@
...
+
+    let store = app
+        .get_store("config.json")
+        .ok_or(anyhow::anyhow!("store not found"))?;
+    store.set(ZOOM_FACTOR_KEY, zoom);
+
+    app.set_menu(create_main_menu(app)?)?;
+
+    Ok(())
+}
```
Simon-Laux, line 229: you can just set the new value in the store and then call `apply_zoom_factor`

Simon-Laux:
float on top support for help window works, but state is not shown on windows and I can't deactivate it after activating it.

Simon-Laux:
also undo and redo do not work on windows, is this expected?

Septias:
Undo and redo are part of [PredefinedMenuItem](https://docs.rs/tauri/latest/tauri/menu/struct.PredefinedMenuItem.html) which currently do not support windows and linux but I suppose that at some point it will just work. For now, we can also disable it the whole submenu or specific items.

Simon-Laux:
I took this over in https://github.com/deltachat/deltachat-desktop/pull/4755

PR #4642, opened 2025-02-14
Improve invite link copy

### LINKED ISSUE #4530: Bad UX/UI in contact QR screeen

nicodh:
#skip-changelog

resolves #4530

image


packages/frontend/scss/dialogs/_qr-code.scss
```diff
@@ -42,3 +42,11 @@
     min-height: 0;
   }
 }
+
+.copy-link-icon {
+  width: 32px;
+  height: 32px;
+  transform: rotate(135deg);
+  margin-right: 8px;
```
WofWca, line 50:
```suggestion
  margin-inline-end: 8px;
```

Did not test

WofWca (commented):
Where does the icon come from? Isn't it supposed to be attributed?

This looks better.

But it doesn't resolve the "feedback" part of the referenced issue.

nicodh:
> Where does the icon come from? Isn't it supposed to be attributed?

How? I don't see Icon attriubutes for Footer buttons.

nicodh:
> But it doesn't resolve the "feedback" part of the referenced issue.

Right. This is only one part, but makes sense already as it's an improvement imo

WofWca:
> How? I don't see Icon attriubutes for Footer buttons.

I mean a license.

nicodh:
> > How? I don't see Icon attriubutes for Footer buttons.
> 
> I mean a license.

It's from https://fonts.google.com/icons like most of the other icons we are using

WofWca (approved):

Simon-Laux (approved):

PR #4644, opened 2025-02-15
Prepare release 1.54.0

nicodh:


WofWca (approved):
I smoke-tested the preview build. Works alright.
The code is also good.

As discussed, we want to have Android and Desktop core versions synced to fix QR and real-time issues, so let's release.

PR #4648, opened 2025-02-15
fix: "jump down" unread counter not clearing

WofWca:
Reproduction steps:
1. Open a chat.
2. Scroll it up.
3. Receive a message or a few in this chat.
4. Scroll down to read all the new messages.
  The counter will be updating as you read the messages, as expected.
5. Scroll up a bit.

The "jump down" button will appear again,
but with an incorrect count.

The bug was introduced in 1a752fa56918857c8f27d2ea9d52000b995f0b8f
(https://github.com/deltachat/deltachat-desktop/pull/4594).

This commit somewhat reverts it, i.e. ensures that
`useUnreadCount` does not lose its state when we start / stop
displaying the "jump down" button.


nicodh (changes requested):
I tested this and the behaviour is the same as in main, so the bug is not fixed.

Is see that onUnreadMessageInView returns immediately if window is in focus? How should the messages ever be marked as read?

nicodh:
Btw. strange that this never was reported so far as bug. It exists at least since 1.52.0.

It is pretty obvious that the badge in the chatlist still shows the number of unread messages even if you scroll to the buttom

WofWca:
> onUnreadMessageInView returns immediately if window is in focus?

It's the opposite

https://github.com/deltachat/deltachat-desktop/blob/f1873b57e64e40ad39d3cbc7677ea30f28b8ed0b/packages/frontend/src/components/message/MessageList.tsx#L159-L162

> I tested this and the behaviour is the same as in main, so the bug is not fixed.

Does the "Message info" dialog show that the last message is read?
I re-tested this again, and for me this does fix the bug.
Maybe the issue is that the message isn't actually marked as "read"?

nicodh:
> > onUnreadMessageInView returns immediately if window is in focus?
> 
> It's the opposite

Argh - ok since I was in the debugger the chat window has no focus anymore. So now I set a debug break point in MessageList.tsx line 198  
BackendRemote.rpc
          .markseenMsgs(accountId, messageIdsToMarkAsRead)
          .then(debouncedUpdateBadgeCounter)

But that was not triggered when scrolling down and thus the counter doesn't get updated. (even without brek point of course)

Interestingly it gets updated as soon as I click somewhere in the chat window or the composer. So maybe when switching to the Deltachat Window and just scroll it does not really focus the chat window?

nicodh:
https://github.com/user-attachments/assets/c3e33e5a-3739-4ff3-86a1-9e26ae7b731c

The interruption was the debugger at markseenMsgs

WofWca:
Is the window focused in the first half of the video? If not, I'd say it's expected.

nicodh:
> Is the window focused in the first half of the video? If not, I'd say it's expected.

The chat is open and I switched from another app back to deltachat. See there are new messages and scroll down. The jumpDown button disappears but the chat list item still shows the counter and markRead is not triggered. So I scrolled to bottom see all messages but it needs another action to trigger the focus? The scrolling should be enough for triggering. So maybe we need another condition. Is it possible to scroll a window without having focus?
I scroll with topuchpad not by clicking on the scrollbar

WofWca:
> Is it possible to scroll a window without having focus?

Yes, e.g. if you have two windows side-by-side.

> The chat is open and I switched from another app back to deltachat.

In this case yeah, sounds like a bug. It's not reproducible on Windows. There it's enough to just Alt + Tab back to the Delta Chat window, then scrolling with the scrollbar does mark messages as read.

Either way, this MR is not relevant to this.

nicodh (approved):
The observed behaviour was caused by a special case - focus was not triggered in a scrolled window, but the fix solves the bug, that the counter is not correct updated

WofWca:
> scrolling with the scrollbar

I meant to say "with the scroll wheel"

packages/frontend/src/components/message/MessageList.tsx
```diff
@@ -692,14 +692,18 @@ export default function MessageList({ accountId, chat, refComposer }: Props) {
...
-          accountId={accountId}
-          chat={chat}
-          jumpToMessage={jumpToMessage}
-          jumpToMessageStack={jumpToMessageStack}
-        />
-      )}
+      <JumpDownButton
+        // We're using `hidden` prop instead of simply not rendering
+        // the component because it's stateful,
+        // namely, its `useUnreadCount()` is stateful
```
WofWca, line 698:
```suggestion
        // the component because it's stateful.
        // Namely, its `useUnreadCount()` is stateful
```

PR #4650, opened 2025-02-15
Fix app picker caching

ralphtheninja:
An attempt at fixing the caching issues in the app picker.

#skip-changelog

### DIFF
--- a/packages/frontend/src/components/composer/AppPickerWrapper.tsx
+++ b/packages/frontend/src/components/composer/AppPickerWrapper.tsx
@@ -40,6 +40,7 @@ export const AppPickerWrapper = ({ onAppSelected }: Props) => {
           AppStoreUrl + 'xdcget-lock.json'
         )
         const apps = getJsonFromBase64(response.blob) as AppInfo[]
+        if (apps === null) return
         apps.sort((a: AppInfo, b: AppInfo) => {
           const dateA = new Date(a.date)
           const dateB = new Date(b.date)
--- a/packages/frontend/src/components/composer/AppPickerWrapper.tsx
+++ b/packages/frontend/src/components/composer/AppPickerWrapper.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect } from 'react'
+import React, { useEffect, useState } from 'react'
 import { AppPicker, AppInfo, AppStoreUrl } from '../AppPicker'
 import { getLogger } from '../../../../shared/logger'
 
@@ -27,11 +27,11 @@ const getJsonFromBase64 = (base64: string): any => {
 
 type Props = {
   onAppSelected?: (app: AppInfo) => void
-  apps: AppInfo[]
-  setApps: (apps: AppInfo[]) => void
 }
 
-export const AppPickerWrapper = ({ onAppSelected, apps, setApps }: Props) => {
+export const AppPickerWrapper = ({ onAppSelected }: Props) => {
+  const [apps, setApps] = useState<AppInfo[]>([])
+
   useEffect(() => {
     const fetchApps = async () => {
       try {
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/composer/AppPickerWrapper.tsx
```diff
@@ -27,11 +27,11 @@ const getJsonFromBase64 = (base64: string): any => {
 
 type Props = {
   onAppSelected?: (app: AppInfo) => void
-  apps: AppInfo[]
-  setApps: (apps: AppInfo[]) => void
 }
 
-export const AppPickerWrapper = ({ onAppSelected, apps, setApps }: Props) => {
+export const AppPickerWrapper = ({ onAppSelected }: Props) => {
+  const [apps, setApps] = useState<AppInfo[]>([])
```
ralphtheninja, line 33: Might as well move the state into this component. It's a wrapper after all? The parent component is only interested in what `onAppSelected` calls back with anyway.

packages/frontend/src/components/composer/AppPickerWrapper.tsx
```diff
@@ -52,17 +52,13 @@ export const AppPickerWrapper = ({ onAppSelected, apps, setApps }: Props) => {
           app.author = url.pathname.split('/')[1]
           app.date = moment(app.date).format('LL')
         }
-        if (apps) {
-          setApps(apps)
-        }
+        setApps(apps)
```
ralphtheninja, line 55: This if-statement seems redundant.

packages/frontend/src/components/composer/AppPickerWrapper.tsx
```diff
@@ -52,17 +52,13 @@ export const AppPickerWrapper = ({ onAppSelected, apps, setApps }: Props) => {
...
       } catch (error) {
         log.error('Failed to fetch apps:', error)
       }
     }
-    if (!apps?.length) {
-      fetchApps()
-    }
-  }, [apps, setApps])
+    fetchApps()
+  }, [setApps])
```
ralphtheninja, line 61: Don't cache on `apps`, this is the result we want to display. The results are cached in core already.

WofWca:
What issues does this fix?

nicodh:
> What issues does this fix?

The requests are cached too long, so updates in the app store are not reflected.

packages/frontend/src/components/composer/AppPickerWrapper.tsx
nicodh: Initially the idea of the wrapper was, to keep the picker component dumb in the sense it just displays apps. But since we load the icons inside the app picker from the same source, I think now it makes sense to move the fetchApps effect also into the AppPicker and thus get rid of the wrapper component completely

packages/frontend/src/components/composer/AppPickerWrapper.tsx
nicodh: Only if we return an empty array instead of null in the catch clause in getJsonFromBase64. Which makes sense. Just to make sure apps is always an (possibly empty) array.

packages/frontend/src/components/composer/AppPickerWrapper.tsx
nicodh: Well the original idea was not to reload all apps any time you (re)open the AppPicker, since changes are rare. But then we would at least have to set a client-side cache-expire time. If core cares about that (only reloading the apps if they changed) we don't need to care about that on the client. (which I didn't know when implementing this tbh)

nicodh:
Just found a small (pre-existing) issue in the wording. See #4654

packages/frontend/src/components/composer/AppPickerWrapper.tsx
nicodh: > so adding a second caching layer just makes things needlessly complicated

yes that's why I said above we don't need the client cache

nicodh (approved):
Reviewed by reading & testing

For me it would be fine to add this to the current release and improve stuff (like removing the AppPickerWrapper) later

packages/frontend/src/components/composer/AppPickerWrapper.tsx
ralphtheninja: Updated with null check, earlier in the code.

PR #4651, opened 2025-02-16
fix: keep the order of contacts when calling getContactsByIds

link2xt:
Otherwise contacts are sorted in arbitrary order
because getContactsByIds stores them in a Rust HashMap.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Changed
 
+- keep the order of contacts when calling getContactsByIds #4651
+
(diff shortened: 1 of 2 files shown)


WofWca (approved):
Makes sense.

There are other occurrences of `getContactsByIds()` where we don't respect the order, e.g. here:

https://github.com/deltachat/deltachat-desktop/blob/5eb89aca2346a3a1739893075650caf4f75be15d/packages/frontend/src/components/dialogs/ReactionsDialog/index.tsx#L77-L92

Maybe we should change it everywhere.

PR #4660, opened 2025-02-16
fix: a11y: add `aria-posinset` for virtualized

WofWca:
For `react-virtualized` lists.

[From MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-posinset):

> The aria-posinset attribute defines an element's number or position
> in the current set of listitems or treeitems
> when not all items are present in the DOM.


WofWca:
After https://github.com/deltachat/deltachat-desktop/pull/4868, this will also need to be done for the gallery.

WofWca:
Let's convert this into an issue: https://github.com/deltachat/deltachat-desktop/issues/5574.

https://github.com/deltachat/deltachat-desktop/pull/5044 did a part of the job.

PR #4662, opened 2025-02-16
fix: a11y: better aria-label for "clear" search

WofWca:
#skip-changelog


Simon-Laux (approved):

Simon-Laux:
for changelog would probably make sense to add one entry for "improve accessibility" and then list all pr numbers on that one item

PR #4664, opened 2025-02-16
fix: MessageListStore logger name is Store

ralphtheninja:
Before change:

Screenshot from 2025-02-17 00-48-02

After change:

Screenshot from 2025-02-17 00-48-42

### DIFF
--- a/packages/frontend/src/stores/messagelist.ts
+++ b/packages/frontend/src/stores/messagelist.ts
@@ -146,7 +146,7 @@ class MessageListStore extends Store<MessageListState> {
     private readonly accountId: number,
     private readonly chatId: number
   ) {
-    super(defaultState())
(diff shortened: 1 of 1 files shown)


packages/frontend/src/stores/messagelist.ts
```diff
@@ -146,7 +146,7 @@ class MessageListStore extends Store<MessageListState> {
     private readonly accountId: number,
     private readonly chatId: number
   ) {
-    super(defaultState())
+    super(defaultState(), 'MessageListStore')
```
ralphtheninja, line 149: `name` parameter in `Store` base class defaults to `'Store'`

Simon-Laux (approved):

WofWca (commented):
Makes sense, did not test.

PR #4665, opened 2025-02-17
refactor: remove `SplitOut` class for less complexity

ralphtheninja:
I don't know about the story behind the `SplitOut` class but I don't think it really serves any purpose. Yet one more file to read and understand that doesn't add much. Why not use normal class methods in the same class?

I've tested all methods. But maybe we can wait with merge until we have finished the release. I'll rebase it later.

#skip-changelog Only refactoring, reducing code complexity.


packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -78,10 +77,8 @@ const WRAPPER_PATH = 'webxdc-wrapper.45870014933640136498.html'
 
 const BOUNDS_UI_CONFIG_PREFIX = 'ui.desktop.webxdcBounds'
 
-export default class DCWebxdc extends SplitOut {
-  constructor(controller: DeltaChatController) {
-    super(controller)
-
+export default class DCWebxdc {
+  constructor(private controller: DeltaChatController) {
```
ralphtheninja, line 81: Finally understood that you can do this in TypeScript.

Simon-Laux:
> I don't know about the story behind the SplitOut class but I don't think it really serves any purpose

In the past there were multiple files, that implemented the JSON API, before core had jsonrpc. It made it a bit simpler to access commonly used properties of the controller by giving them shorter names, thats the story behind it.

Simon-Laux (approved):

PR #4668, opened 2025-02-17
Prepare release 1.54.1

nicodh:


WofWca (approved):
The changes are pretty minor, so it's probably OK to release without going through a test release.

Issue #4225, opened 2024-10-19
HTML semantics: move navbars inside the "chat list" and "chat" sections


WofWca:
- Delta Chat Version: Desktop 1.46.8
- Expected behavior: Pressing "Tab" inside the search field moves focus to the chat list
- Actual behavior: Pressing "Tab" inside the search field moves focus to the "chat" section of the navbar

This has to do with this code:

https://github.com/deltachat/deltachat-desktop/blob/c53d95cc531b8a96744186b58d3f52fbfe17b99b/packages/frontend/src/components/screens/MainScreen.tsx#L169-L270

Our current way is that there are 3 sections: navbar, chat list, and message list. I think that having 4 sections makes more sense: chat list, chat list navbar, message list, chat's (message list's) navbar.

Another place where this acts in an unintuitive way is the gallery, where you'd expect to go to gallery from the "gallery" tab, and not to the chat list.

Please let me know if you agree. If so, I will proceed with implementing the suggested changes.

Related: https://github.com/deltachat/deltachat-desktop/pull/4224

PR #4672, opened 2025-02-17
fix: a11y: improve tab order (navbar)

WofWca:
Toggle "hide whitespace" for easier review.

This ensures that the list of chats follows the search bar,
and that the messages list follows the chat header.
Closes https://github.com/deltachat/deltachat-desktop/issues/4225.

This mostly moves code around, and simplifies some things
(such as replacing `width: 30%` duplication) along the way.

But it also changes the relative sizes of the "chat list" and the
"chat" (main view) sections, to a ratio of 3:8
(before the "chat list" section was 30% of the window size).

Otherwise, besides, of course, the tab order fix,
this is not supposed to affect the behavior and the look of the app.

I testes a lot of things:
- various screen sizes (small screen mode included)
- gallery view
- global gallery view
- opening "Search in chat"

But there is still a chance that I missed something.


Simon-Laux (approved):
I tested it and code looks good (though I have not read it line by line).

PR #4673, opened 2025-02-17
Merge `AppPickerWrapper.tsx` and `AppPicker/index.tsx` into one

ralphtheninja:
A few notes on review:

* Don't look at the complete diff immediately
* Check the diff on each commit. I've done the refactoring in several steps.
* Each commit is a working app picker, so you can check out each commit and do `pnpm debug:electron` (`pnpm check` is also good for each commit)
* The final commit pulls the plug on `AppPickerWrapper.tsx` which then is just a component returning another component

#skip-changelog Refactoring

I've tested the app picker on each commit.


packages/frontend/src/components/AppPicker/index.tsx
```diff
@@ -235,7 +234,7 @@ export function AppPicker({ className, onSelect, apps = [] }: Props) {
   }
 
   return (
-    <div className={classNames(styles.appPickerContainer, className)}>
```
ralphtheninja, line 238: Note that this was actually a bug, since `styles.appPickerContainer` here doesn't exist.

packages/frontend/src/components/AppPicker/index.tsx
```diff
@@ -62,6 +82,35 @@ export function AppPicker({ onSelect, apps = [] }: Props) {
     AppCategoryEnum.game,
   ]
 
+  useEffect(() => {
```
ralphtheninja, line 85: This commit moves the logic from the wrapper component.

packages/frontend/src/components/AppPicker/index.tsx
```diff
@@ -230,7 +230,7 @@ export function AppPicker({ onSelect }: Props) {
             <FooterActions>
               <FooterActionButton
                 data-testid='add-app-to-chat'
-                onClick={() => onSelect && onSelect(app)}
```
ralphtheninja, line 185: This is always set so no need to check for it.

packages/frontend/src/components/composer/AppPickerWrapper.tsx
```diff
@@ -2,9 +2,9 @@ import React from 'react'
 import { AppPicker, AppInfo } from '../AppPicker'
 
 type Props = {
-  onAppSelected?: (app: AppInfo) => void
+  onAppSelected: (app: AppInfo) => void
 }
 
 export const AppPickerWrapper = ({ onAppSelected }: Props) => {
-  return <AppPicker onSelect={onAppSelected} />
+  return <AppPicker onAppSelected={onAppSelected} />
```
ralphtheninja: This is just to keep the same api with respect to the `Composer` component.

nicodh (approved):
Reviewed by testing & reading

Issue #4596, opened 2025-02-03
adapt to new saved-messages api


r10s:
since PR https://github.com/deltachat/deltachat-core-rust/pull/5606 we have a new saved-messages-api, allowing to save messages similar as known from whatsapp or telegram. see that PR for reasonings and discussions.

to benefit from the new saved-messages-api on desktop, the following steps are needed:

- make [`dc_save_msgs()`](https://c.delta.chat/classdc__context__t.html#ac8075b06088ddbe39cd8cc3802141102), [`dc_msg_get_original_msg_id()`](https://c.delta.chat/classdc__msg__t.html#a3d3fdca5164b38d93e1f42f55ca67f72) and [`dc_msg_get_saved_msg_id()`](https://c.delta.chat/classdc__msg__t.html#ab13b251a52cf6b36627dff46f3c0be3e) available in **jsonrpc** 

- in self-chat, for incoming messages, **show group avatar, sender name and correct left/right side bubbles and colors**, see eg. https://github.com/deltachat/deltachat-android/pull/3073/commits/75e9984959004e632737cd1e0635ff0593789dae for how to change the condition.
already now, the state is not that bad: messages from others saved eg. on android are shown on desktop on the left side using the correct color already today, only avatar and sender name is missing

- ~~when **forwarding a message to "Saved Messages"** _that was not already saved_ use `dc_save_msgs()`~~
  edit: updated in discussion: keep forwarding as it is

- for every message shown in a chat, check if it was saved [^3]. if so, **show a little bookmark icon beside the date** ([screenshot](https://github.com/deltachat/deltachat-ios/pull/2587)). as the bookmark should also be shown in saved messages, you will draw it if original_msg_id() OR saved_msg_id() is true.

- add a button to **go to original** beside each saved message in the self-chat ([screenshot](https://github.com/deltachat/deltachat-ios/pull/2527))

to get the idea, it makes sense to use iOS or Android 1.54

as the new core is already shipped, there is not a huge urgency to target this issue, no messages saved on iOS/android will be lost or so. but of course, it is a very nice to have adaption, and comparable simple for UI.

[^3]: at least for the first iteration, we do not want to show saved info messages and thelike in the "Saved Messages" chat; this turned out to be confusing and might also require adaptions on systems other than android (remember: saved messages are synced), cmp. https://github.com/deltachat/deltachat-ios/issues/2567

r10s:
closing this for a few days until android is done, a few things are still under discussion

[...]

PR #4674, opened 2025-02-18
Saved messages

nicodh:
Relates to #4596 

- [x] enable to save (or unsave) message from context menu of messages
- [x] show sender & avatar in self chat for saved messages
- [x] show bookmark icon for saved messages in self chat
- [x] show bookmark icon for saved messages in any chat
- [x] show button to jump to original message

New item in context menu:

image

New bookmark icons

image

New jump to original message button

image


packages/frontend/scss/message/_message.scss
```diff
@@ -212,6 +220,14 @@
       );
     }
 
+    & > .saved-message-icon {
+      @include color-svg(
+        './images/icons/bookmark-filled.svg',
+        var(--messagePadlockOutgoing),
```
Simon-Laux, line 226: same here, consider renaming the css var

packages/frontend/src/components/message/Message.tsx
```diff
@@ -301,6 +301,12 @@ function buildContextMenu(
       label: tx('react'),
       action: handleReactClick,
     },
+    // Save Message
+    {
```
Simon-Laux, line 305: shouldn't this be conditional, like show some un-save action if it is already saved?

nicodh: We could skip it, if it is already saved. But atm we have no "unsave" action, except deleting the saved message

Simon-Laux: dc iOS has save and un-save

Simon-Laux: @r10s how does un-save work on iOS?

r10s: > @r10s how does un-save work on iOS?

by deleting the message-id returned by [`get_saved_msg_id()`](https://c.delta.chat/classdc__msg__t.html#ab13b251a52cf6b36627dff46f3c0be3e)

nicodh: Hmm, is that neccessary? I would say its sufficient to delete the saved message

_locales/_untranslated_en.json
```diff
@@ -45,6 +45,12 @@
   "multidevice_connection_takes_too_long": {
     "message": "This takes longer than expected 🐌 make sure both devices are on the same Wi-Fi or network and check troubleshooting"
   },
+  "saved_message": {
+    "message": "Saved Message"
+  },
+  "jump_to_original_message": {
+    "message": "Jump to original message"
+  },
```
Simon-Laux, line 53: do these exist upstream already?

packages/frontend/src/components/message/MessageMetaData.tsx
```diff
@@ -54,6 +56,12 @@ export default function MessageMetaData(props: Props) {
           className={'padlock-icon'}
         />
       )}
+      {isSavedMessage && (
+        <div
+          aria-label={tx('saved_message')}
```
r10s, line 61: better use the existing string here. the context of "message" is anyway clear

```suggestion
          aria-label={tx('saved')}
```

packages/frontend/src/components/ShortcutMenu/index.tsx
```diff
@@ -38,6 +51,16 @@ export default function ShortcutMenu(props: Props) {
         onClick={props.showContextMenu}
         tabIndex={props.tabindexForInteractiveContents}
       />
+      {props.message.originalMsgId !== null && (
+        <button
+          className={classNames(styles.originalMessageButton)}
+          aria-label={tx('jump_to_original_message')}
+          onClick={onClick}
+          title={tx('jump_to_original_message')}
```
r10s, line 59: this is what we're using on android / iOS, it is good enough, and also already known from other places.

we can iterate, if we get corresponding feedback at scale

```suggestion
          aria-label={tx('show_in_chat')}
          onClick={onClick}
          title={tx('show_in_chat')}
```

ralphtheninja:
Should one be able to save messages, within saved messages?

image

packages/frontend/src/components/ShortcutMenu/index.tsx
nicodh: Will change it again once you added the new string...

ralphtheninja (approved):
Reviewed by testing. Works well!

r10s:
> Should one be able to save messages, within saved messages?

no, this should not be possible. it is also not possible on android, that does not have a dedicated "save" button yet (but that is requested and [coming](https://github.com/deltachat/deltachat-android/issues/3599)). on iOS we hide the "save" and "unsave" button within saved messages ("unsave" equals "delete" there).

so, we should just hide save/unsave in "saved messages"

nicodh:
> so, we should just hide save/unsave in "saved messages"

Done with commit 146b81a35e

Simon-Laux (approved):

PR #4675, opened 2025-02-18
fix: a11y: 'tablist' role for gallery / chat view

WofWca:
if you listen to the screen-reader, it's not clear
what the buttons do, and how to change the active view.
This should fix it.

### DIFF
--- a/packages/frontend/src/components/Gallery.tsx
+++ b/packages/frontend/src/components/Gallery.tsx
@@ -278,8 +278,18 @@ export default class Gallery extends Component<
     const showDateHeader =
       currentTab !== 'files' && currentTab !== 'webxdc_apps'
 
+    // FYI the role="tablist" element is not rendered when
+    // `props.chatId === 'all'`.
+    // Would be nice to DRY this somehow.
+    const isTabpanel = this.props.chatId !== 'all'
+
     return (
-      <div className='media-view'>
+      <div
+        role={isTabpanel ? 'tabpanel' : undefined}
+        aria-labelledby={isTabpanel ? 'tab-media-view' : undefined}
+        id='media-view'
+        className='media-view'
+      >
         <ul ref={this.tabListRef} className='tab-list .modifier' role='tablist'>
           <RovingTabindexProvider
             wrapperElementRef={this.tabListRef}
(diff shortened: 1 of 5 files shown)


packages/frontend/src/components/Gallery.tsx
```diff
@@ -279,7 +279,14 @@ export default class Gallery extends Component<
       currentTab !== 'files' && currentTab !== 'webxdc_apps'
 
     return (
-      <div className='media-view'>
+      <div
+        // FYI the role="tablist" element is not rendered when
+        // `props.chatId === 'all'`.
+        role={this.props.chatId === 'all' ? undefined : 'tabpanel'}
+        aria-labelledby='tab-media-view'
```
Simon-Laux, line 286: that element also does not exist in all media view, does it?

packages/frontend/src/components/screens/MainScreen.tsx
```diff
@@ -402,24 +402,41 @@ function ChatNavButtons() {
...
+          aria-selected={activeView === ChatView.Media}
+          aria-controls='media-view'
           aria-label={tx('media')}
           className='navbar-button'
         >
           <Icon coloring='navbar' icon='image' size={18} />
         </Button>
         {settingsStore?.desktopSettings.enableOnDemandLocationStreaming && (
+          // Yes, this is not marked as `role='tab'`.
+          // I'm not sure if this is alright.
```
Simon-Laux, line 439: I guess it should tell that it opens a new window with the map. but the map is an experimental feature and not accessible as far as I know.

packages/frontend/src/components/Gallery.tsx
WofWca: Yes. I wanted to keep it in case something changes.

I guess I'll remove this as well.

packages/frontend/src/components/screens/MainScreen.tsx
WofWca: I've been researching `aria-haspopup="dialog"`. It applies to many buttons in our app, not just this one. But I am not sure yet whether it's good. I suggest to wait for the report, and see if they have concrete suggestions for this.

PR #4679, opened 2025-02-19
Close on invite link copy

nicodh:
- close user feedback after 3 seconds

#skip-changelog


nicodh:
@r10s as discussed before: this closes the QR code dialog when having clicked on the copy link & closes the "toast" after 3 seconds

r10s (approved):

PR #4683, opened 2025-02-20
open map in landscape orientation and constrain default webxdc bounds by screen workarea

### LINKED ISSUE #3847: Change map default dimensions to be bigger and landscape instead of portrait

Simon-Laux:
- open map in landscape orientation and with a bigger window
- constrain size of webxdc window by available screen-workarea-space

closes #3847


ralphtheninja:
I just tested this locally but the map but the map looks like a normal webxdc

Opened by

image

image

I'm guessing this happens because I have opened the map before and it has saved it's bounds already. So this fix only has effect for people opening the map for the first time, right?

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -731,13 +756,18 @@ export default class DCWebxdc {
...
+              msgId,
+              {
+                accountId,
+                displayname: '',
+                chatName,
+                webxdcInfo: messageWithMap.webxdcInfo,
+                href: '',
+              },
+              true
+            )
```
ralphtheninja, line 770: To be honest I don't like it when we pass around booleans. Why not just pass the default bounds in here instead of true?

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -106,7 +126,8 @@ export default class DCWebxdc {
     const openWebxdc = async (
       _ev: IpcMainInvokeEvent,
       msg_id: number,
-      p: DcOpenWebxdcParameters
+      p: DcOpenWebxdcParameters,
+      is_maps_xdc = false // special behaviour for the map dc integration (in this case bigger landscape window)
```
ralphtheninja, line 130: Here we could have a `defaultBounds` parameter instead. This method shouldn't care if we are running a map or not.

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -77,6 +87,16 @@ const WRAPPER_PATH = 'webxdc-wrapper.45870014933640136498.html'
...
+const DEFAULT_BOUNDS_WEBXDC = {
+  width: 375,
+  height: 667,
+}
+
+const DEFAULT_BOUNDS_MAP = {
+  width: 1000,
+  height: 800,
+}
+
```
ralphtheninja, line 99: What about doing something like this instead?

image

ralphtheninja:
@Simon-Laux I can make a PR into this branch instead. It's easier to show what I mean that way. See if you like it or not.

Simon-Laux:
> I'm guessing this happens because I have opened the map before and it has saved it's bounds already. So this fix only has effect for people opening the map for the first time, right?

yes it only affects new/fresh accounts

BTW: I don't like the whole way of reusing webxdc for the map and abusing the update mechanism, which only allows one instance of the map currently. IMO we need to think about a good way to do deltachat extensions, as their own thing, separate from webxdc. Like expose a dedicated jsonrpc api with autogenerated bindings that you can easily develop extensions and don't need to invent a new RPC mechanism over sending status updates that doesn't even do error handling and has no typescript types. Also the map xdc itself still lacks features that the old map box map had. But enough ranting, for now it is what it is and we will need some time where we focus on it to make it really nice.

PR #4686, opened 2025-02-20
Pass in default size instead of booleans

ralphtheninja:
A slightly different approach.

* Pass a default size (note, a `Size` that only has `width` and `height`) instead of a boolean
* Compute the default size for webxdc and map at startup, based on the primary display, instead of putting it into `getLastBounds()`
* `getLastBounds()` returns a bounds if there were any, if this wasn't the case, use the `defaultSize` which already is adjusted to to the size of the primary display

Tweak to https://github.com/deltachat/deltachat-desktop/pull/4683

#skip-changelog PR to a PR

### DIFF
--- a/packages/target-electron/src/deltachat/webxdc.ts
+++ b/packages/target-electron/src/deltachat/webxdc.ts
@@ -23,10 +23,7 @@ import { readdir, stat, rmdir, writeFile, readFile } from 'fs/promises'
 import { existsSync } from 'fs'
 import type DeltaChatController from './controller.js'
 import { getLogger } from '../../../shared/logger.js'
-import {
-  getConfigPath,
-  htmlDistDir,
-} from '../application-constants.js'
+import { getConfigPath, htmlDistDir } from '../application-constants.js'
 import { truncateText } from '@deltachat-desktop/shared/util.js'
 import { tx } from '../load-translations.js'
 import { Bounds, DcOpenWebxdcParameters } from '../../../shared/shared-types.js'
@@ -87,17 +84,18 @@ const WRAPPER_PATH = 'webxdc-wrapper.45870014933640136498.html'
 
 const BOUNDS_UI_CONFIG_PREFIX = 'ui.desktop.webxdcBounds'
 
-const DEFAULT_BOUNDS_WEBXDC = {
-  width: 375,
-  height: 667,
-}
-
-const DEFAULT_BOUNDS_MAP = {
-  width: 1000,
-  height: 800,
-}
+type Size = { width: number; height: number }
 
 export default class DCWebxdc {
+  private DEFAULT_SIZE_WEBXDC: Size = defaultSize({
+    width: 375,
+    height: 667,
+  })
+  private DEFAULT_SIZE_MAP: Size = defaultSize({
+    width: 1000,
+    height: 800,
+  })
+
   constructor(private controller: DeltaChatController) {
     // icon protocol
     app.whenReady().then(() => {
@@ -127,7 +125,7 @@ export default class DCWebxdc {
       _ev: IpcMainInvokeEvent,
       msg_id: number,
       p: DcOpenWebxdcParameters,
-      is_maps_xdc = false // special behaviour for the map dc integration (in this case bigger landscape window)
+      defaultSize: Size = this.DEFAULT_SIZE_WEBXDC
     ) => {
       const { webxdcInfo, chatName, displayname, accountId, href } = p
       const addr = webxdcInfo.selfAddr
@@ -284,11 +282,7 @@ export default class DCWebxdc {
 
       const app_icon = icon_blob && nativeImage?.createFromBuffer(icon_blob)
 
-      const lastBounds = await this.getLastBounds(
-        accountId,
-        msg_id,
-        is_maps_xdc ? DEFAULT_BOUNDS_MAP : DEFAULT_BOUNDS_WEBXDC
-      )
+      const lastBounds = await this.getLastBounds(accountId, msg_id)
       const webxdcWindow = new BrowserWindow({
         webPreferences: {
           partition: partitionFromAccountId(accountId),
@@ -308,7 +302,7 @@ export default class DCWebxdc {
       })
       setContentProtection(webxdcWindow)
       // reposition the window to last position (or default)
-      webxdcWindow.setBounds(lastBounds, true)
+      webxdcWindow.setBounds(lastBounds || defaultSize, true)
       // show after repositioning to avoid blinking
       webxdcWindow.show()
       open_apps[`${accountId}.${msg_id}`] = {
@@ -766,7 +760,7 @@ export default class DCWebxdc {
                 webxdcInfo: messageWithMap.webxdcInfo,
                 href: '',
               },
-              true
+              this.DEFAULT_SIZE_MAP
             )
           }
         }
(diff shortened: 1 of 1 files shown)


packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -87,17 +84,18 @@ const WRAPPER_PATH = 'webxdc-wrapper.45870014933640136498.html'
...
-}
-
-const DEFAULT_BOUNDS_MAP = {
-  width: 1000,
-  height: 800,
-}
+type Size = { width: number; height: number }
 
 export default class DCWebxdc {
+  private DEFAULT_SIZE_WEBXDC: Size = defaultSize({
```
ralphtheninja, line 90: Has to be computed here. On file scope isn't possible since we don't know about the primary display yet.

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -308,7 +302,7 @@ export default class DCWebxdc {
       })
       setContentProtection(webxdcWindow)
       // reposition the window to last position (or default)
-      webxdcWindow.setBounds(lastBounds, true)
+      webxdcWindow.setBounds(lastBounds || defaultSize, true)
```
ralphtheninja, line 305: If we didn't have any `lastBounds`,  just fall back to the default size.

packages/target-electron/src/deltachat/webxdc.ts
Simon-Laux: I wouldn't calculate this on delta chat application start, the user might connect an external monitor during the session and change their primary display to it. I think the calculation should be done always, also after retrieving the last used size - that might be from a different session where the screen size has been different.

Simon-Laux (approved):

packages/target-electron/src/deltachat/webxdc.ts
ralphtheninja: I'll tweak it.

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -921,12 +922,16 @@ export async function webxdcStartUpCleanup() {
   }
 }
 
-function defaultSize(default_size: Size): Size {
+/**
+ * Make sure a default size doesn't extend the size
+ * of the primary display work area.
+ */
+function adjustDefaultSize(size: Size): Size {
```
ralphtheninja, line 929: Renamed the parameter, E_TOO_MANY_DEFAULT :grin:

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -302,7 +302,8 @@ export default class DCWebxdc {
       })
       setContentProtection(webxdcWindow)
       // reposition the window to last position (or default)
-      webxdcWindow.setBounds(lastBounds || defaultSize, true)
+      const bounds = lastBounds || adjustDefaultSize(defaultSize)
```
ralphtheninja, line 305: Compute it _here_.

ralphtheninja: But hmm, maybe we should really adjust the complete bound? E.g.

```js
adjustSize(lastBounds || defaultSize)
```

Simon-Laux: I'd also do that for the lastBounds, they might come from a different screen
```suggestion
      const bounds = adjustDefaultSize(lastBounds || defaultSize)
```

maybe rename the function to remove the default from it's name, maybe some name that represents more the task of "ensure that size does not exceed screen bounds"?

ralphtheninja: yeah I wanted to do this also, but I'm not skilled enough and don't have the energy to wrestle for hours with typescript to make a single function call, it's really getting in my way to write the simplest javascript code :/

ralphtheninja: TypeScript doesn't complain about that change, but then you only return width and height, but you're not going to pass x and y.

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -933,6 +922,19 @@ export async function webxdcStartUpCleanup() {
   }
 }
 
+/**
+ * Make sure a default size doesn't extend the size
+ * of the primary display work area.
+ */
+function adjustBounds(size: Size): Size {
```
ralphtheninja, line 929: @Simon-Laux So this doesn't work now, since it will omit x and y, if they were set in `lastBounds`. Hilfe bitte!

ralphtheninja: I guess TypeScript doesn't complain on this since a `Bounds` is a `Size` but a `Size` isn't a `Bounds`. So what I'd like to do is to accept both a `Bounds` and a `Size` and return x and y as well, if they are present.

Simon-Laux: for the main window it calculates the middle of the screen

Simon-Laux: anyways I thin we could do sth like this?
```ts
const last_bounds:Bounds|null = getLastBounds(accountId,msgId)
const size:Size = adjustDefaultSize(lastBounds || defaultSize)
const bounds:Bounds = {...(last_bounds||{}), ...size}
```

Simon-Laux: `Partial<T>` makes all properties optional

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -923,15 +925,20 @@ export async function webxdcStartUpCleanup() {
 }
 
 /**
- * Make sure a default size doesn't extend the size
- * of the primary display work area.
+ * Make sure bounds doesn't extend the primary display work area.
  */
-function adjustBounds(size: Size): Size {
+function adjustBounds(bounds: Bounds): Bounds {
```
ralphtheninja, line 930: I'm more happy with this. Now we're working with `Bounds` everywhere.

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -923,15 +925,20 @@ export async function webxdcStartUpCleanup() {
...
   const { height: screenHeight, width: screenWidth } =
     screen.getPrimaryDisplay().workAreaSize
+  const width = Math.min(bounds.width, screenWidth)
+  const height = Math.min(bounds.height, screenHeight)
+  const x = bounds.x === -1 ? screenWidth / 2 - width : bounds.x
+  const y = bounds.y === -1 ? screenHeight / 2 : bounds.y
```
ralphtheninja, line 936: This sets x and y like before, sort of. We might need some better tweaking here also on x and y. What happens if x is greater than screenWidth and y greater than screenHeight etc?

Simon-Laux: I would ignore that edgecase for now until it happens in the wild

packages/target-electron/src/deltachat/webxdc.ts
ralphtheninja: > anyways for now we could do sth like this?

That looks pretty nice actually. But we don't do anything with x and y. Maybe we can postpone this?

ralphtheninja:
@Simon-Laux Doing it your way! Thanks for holding my hand with TypeScript.

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -87,12 +84,13 @@ const WRAPPER_PATH = 'webxdc-wrapper.45870014933640136498.html'
...
-const DEFAULT_BOUNDS_WEBXDC = {
+type Size = { width: number; height: number }
+
+const DEFAULT_SIZE_WEBXDC: Size = {
   width: 375,
   height: 667,
 }
-
-const DEFAULT_BOUNDS_MAP = {
+const DEFAULT_SIZE_MAP: Size = {
```
ralphtheninja, line 93: Maybe there should be a space here idk, I kind of thought they belong together so I don't mind.

ralphtheninja:
I'm happy with this now. Ready to squash it unless there's something else we should tweak.

Simon-Laux (approved):

PR #4687, opened 2025-02-20
Update core to 1.155.6

ralphtheninja:


nicodh (changes requested):
We usually add a line to the Changelog after Core updates, including possible user relevant changes

It's also agreed with @Simon-Laux  to update core version in /packages/target-tauri/src-tauri/Cargo.toml and run a "smoke test" and cargo clippy to make sure tauri is still building as expected

ralphtheninja:
> It's also agreed with @Simon-Laux to update core version in /packages/target-tauri/src-tauri/Cargo.toml and run a "smoke test" and cargo clippy to make sure tauri is still building as expected

So then maybe @Simon-Laux can update that file, run the smoke test and push to this branch! :man_dancing:

Simon-Laux:
you should be able to run this script to update core in both js and rust projects:
```
./bin/link_core/link_version.js 1.155.6 # mind that it has no "v"-prefix
```

sometimes there are rust dependency conflicts, but not this time. (mostly they can be resolved by reading error message and updating other dependencies as well, for example with `cargo update`)

for the smoke test you can run:
```
pnpm -w dev:tauri
```
the `-w` stands for workspace script, so it doesn't matter which directory you execute this from, it works anywhere across the whole deltachat-desktop folder

Simon-Laux (approved):

PR #4690, opened 2025-02-21
fix link and anchor issue in html email view

### LINKED ISSUE #4017: HTML email view: Some links don't open browser
More specifically, links without a schema, like `academy.qt.io/enrollments`

Expected: Open link in system browser
Actual: Opening opens a blank page in the html viewer.

### Foundation knowledge
The html email window has two parts: the ui and a sandboxed browser view with disable JS in which the html email is loaded.

Email source is loaded from the `email://` scheme that is implemented in https://github.com/deltachat/deltachat-desktop/blob/ba6534b3a9d60ad3e602f770ba630849cc62b87c/src/main/windows/html_email.ts#L412-L420

The http/s handlers are overwritten by handlers that call the http request method of DC core, if `allow_remote_content` is activated.

Then there is link handling:

https://github.com/deltachat/deltachat-desktop/blob/ba6534b3a9d60ad3e602f770ba630849cc62b87c/src/main/windows/html_email.ts#L472-L493

This `openLink` function is called in two cases:

- `sandboxedView.webContents.on( 'will-navigate', ...` [source](https://github.com/deltachat/deltachat-desktop/blob/ba6534b3a9d60ad3e602f770ba630849cc62b87c/src/main/windows/html_email.ts#L496-L503)
- `sandboxedView.webContents.setWindowOpenHandler(details => { ...` [source](https://github.com/deltachat/deltachat-desktop/blob/ba6534b3a9d60ad3e602f770ba630849cc62b87c/src/main/windows/html_email.ts#L507)

seems this is not enough to catch this case.
I suspect it tries to open it via the `email://` scheme, so we might need to check if it is a domain that is tried to be opened instead of index.html.

concerns:
- we might need to expand the function in the future to load embedded images/media from the email in some edge cases? maybe?
- it might be too late when we get the call to the link in the email scheme handler, the user might already get a white/blank screen and we can not reject the request?
- last resort could be to modify the html to fix those `<a href=` links, to add `https`-scheme, but not sure we want that, could lead to more edge cases/bugs.

If you want or need dev tools for the email content add this to the code in `html_email.ts`.
```
sandboxedView.webContents.openDevTools()
```

Simon-Laux:
closes #4017


packages/target-electron/src/windows/html_email.ts
```diff
@@ -516,6 +516,28 @@ function makeBrowserView(
...
+        if (url.slice(prefix.length)[0] == '/') {
+          urlWithoutPrefix = urlWithoutPrefix.slice(1)
+        }
+        if (urlWithoutPrefix.startsWith('#')) {
+          // double fragment issue workaround
+          // the issue is that when clicking on the same anchor multiple times,
+          // only the first time it jumps to it - the times after it just appends it to the url without jumping,
+          // might be an issue in electopn/chromium
+          // like this `email://index.html#anchor#anchor`
+          let lastFragment = urlWithoutPrefix.split('#').reverse()[0]
```
ralphtheninja, line 531:
```suggestion
          const lastFragment = urlWithoutPrefix.split('#').reverse()[0]
```

ralphtheninja (approved):

PR #4691, opened 2025-02-21
Searchfield for help

nicodh:
This PR adds a search field to the help page, using [pagefind](https://pagefind.app) which generates an index which is then used to search and disply results in the static help page.

Needs latest main of https://github.com/deltachat/deltachat-pages/

image

image


Simon-Laux (commented):
Nice. For the translation keys I think it makes more sense to use the existing substitution pattern, you can convert it to the custom one in `bin/help/help-translations.js`.
- reasons: reusability and consistency

_locales/_untranslated_en.json
```diff
@@ -44,5 +44,8 @@
   },
   "multidevice_connection_takes_too_long": {
     "message": "This takes longer than expected 🐌 make sure both devices are on the same Wi-Fi or network and check troubleshooting"
+  },
+  "one_result":  {
```
Simon-Laux, line 48: the key says nothing about search or search term

nicodh: changed the wording according to the existing keys

bin/help/create-local-help.sh
```diff
@@ -0,0 +1,30 @@
...
+  echo ./static/help/$LANGUAGE
+  if [ -f "$FILE" ]; then
+    # remove previous generated files
+    rm -r ./static/help/$LANGUAGE/pagefind
+    npx pagefind --site ./static/help/$LANGUAGE
+    # remove unused files
+    rm ./static/help/$LANGUAGE/pagefind/*.css
+    rm ./static/help/$LANGUAGE/pagefind/pagefind-*.js
+  else
+    echo "Datei $FILE nicht gefunden"
```
Simon-Laux, line 21: english please
```suggestion
    echo "File $FILE not found"
```

static/help/cs/pagefind/pagefind.js
```diff
@@ -0,0 +1,9 @@
...(new Uint8Array(this.searchMeta));if(Object.keys(this.mergeFilter)?.length){let filters2=this.stringifyFilters(this.mergeFilter);let ptr=await this.getPtr();this.raw_ptr=this.backend.add_synthetic_filter(ptr,filters2)}if(this.ranking){await this.set_ranking(this.ranking)}}async loadEntry(){try{let entry_response=await fetch(`${this.basePath}pagefind-entry.json?ts=${Date.now()}`);let entry_json=await entry_response.json();this.languages=entry_json.languages;if(entry_json.version!==this.version){if(this.primary){console.warn(["Pagefind JS version doesn't match the version in your search index.",`Pagefind JS: ${this.version}. Pagefind index: ${entry_json.version}`,"If you upgraded Pagefind recently, you likely have a cached pagefind.js file.","If you encounter any search errors, try clearing your cache."].join("\n"))}else{console.warn(["Merging a Pagefind index from a different version than the main Pagefind instance.",`Main Pagefind JS: ${this.version}. Merged index (${this.basePath}): ${entry_json.version}`,"If you encounter any search errors, make sure that both sites are running the same version of Pagefind."].join("\n"))}}}catch(e){console.error(`Failed to load Pagefind metadata:
```
ralphtheninja, line 1: These `static/help/${language}/pagefind/pagefind.js` are identical?

nicodh: Yes, but the problem is, that there is no way (or I didn't find one) to tell the generated Javascript via config to use another path to access the generated fragments. So we would have to either change the minified Javascript or add someting to the pagefind project to enable our use case without file duplication.

Of course in Unix systems a symlink would be fine, but that causes problems on Windows/Mac I assume

ralphtheninja: I'm fine with it. I was just wondering if it was expected or not. We solve the windows problem with redundancy.

static/help/en/help.html
```diff
@@ -1,14 +1,14 @@
 <!DOCTYPE html>
-<html><head><meta charset="UTF-8" /><meta name="viewport" content="initial-scale=1.0" /><link rel="stylesheet" href="../help.css" /></head><body><ul id="top">
+<html><head><meta charset="UTF-8" /><meta name="viewport" content="initial-scale=1.0" /><link rel="stylesheet" href="../help.css" /><script src="./pagefind/pagefind-ui.js"></script><script src="../start-pagefind.js"></script></head><body><ul id="top">
```
Simon-Laux, line 2: "pagefind-ui.js" is wrong, it is just "pagefind.js"

Simon-Laux (changes requested):
does not work out of the box and I agree with the idea that `pagefind.js` and `wasm.unknown.pagefind` should not be duplicated - maybe they could be imported via import statement from `start-pagefind.js`?

Simon-Laux:
maybe `https://pagefind.app/docs/ui/#bundle-path` could help, Pagefind (the lib not ui) also has a `basePath` option

nicodh:
> does not work out of the box and I agree with the idea that `pagefind.js` and `wasm.unknown.pagefind` should not be duplicated - maybe they could be imported via import statement from `start-pagefind.js`?

Sorry -  the problem was, that I didn't commit the changes needed after the update of deltachat-pages

nicodh:
Concerning the redundant files: pagefind expects the wasm files to be at the same path as pagefind-ui.js & pagefind.js.
All together build the "bundle-path". The paths to the data are obviously relative.

What I did now is: copy the needed files to static/help/ and remove them from /static/help/<locale>/pagefind/ in build process.
Then copy them to the current language dir on runtime when needed.

Simon-Laux:
> Then copy them to the current language dir on runtime when needed. (when language is switched)

bad idea, app files at runtime should be treated immutable. what you can do is making a custom uri schema for the help and match on the name to return the same file in all sub directories, similar to how `webxdc.js` is served.

nicodh:
I found a solution: added a lang attribute to each help page and moved the pagefind indexing one level up, so all data goes into one pagefind directory while pagefind assigns the index per language and shows only appropriate matches

Simon-Laux (approved):

Simon-Laux:
Bildschirmfoto 2025-02-27 um 14 47 39

Icon/char for clear search is not loaded, but that no blocker in my eyes

nicodh:
> Icon/char for clear search is not loaded, but that no blocker in my eyes

You have an idea why? It's just a Unicode 🞫 ?

Simon-Laux:
> You have an idea why? It's just a Unicode 🞫 ?

you need to load a font that supports it

nicodh:
> > You have an idea why? It's just a Unicode 🞫 ?
> 
> you need to load a font that supports it

You mean it depends if your system font supports it? Does it makes sense to look for another sign with better support? Maybe https://en.wikipedia.org/wiki/Multiplication_sign ?

Simon-Laux:
- either add a font that supports it to the css of the help file
- use just the "X" sign
- use a string/button with the text "clear"
- or use an image/icon.

r10s:
this PR has reverted to help to a pretty old state, i'll try and update at https://github.com/deltachat/deltachat-desktop/pull/4725

PR #4692, opened 2025-02-21
fix: a11y: announce new messages of current chat

WofWca:
- **refactor: single return in MessageListInner**
- **fix: a11y: announce new messages of current chat**

### DIFF
--- a/packages/frontend/src/components/message/MessageList.tsx
+++ b/packages/frontend/src/components/message/MessageList.tsx
@@ -854,60 +854,56 @@ export const MessageListInner = React.memo(
       // over the lifetime of this component.
     })
 
-    if (!loaded) {
-      return (
-        <div id='message-list' ref={messageListRef} onScroll={onScroll2}>
-          <ol></ol>
-        </div>
-      )
-    }
-
     return (
       <div id='message-list' ref={messageListRef} onScroll={onScroll2}>
         <ol>
-          <RovingTabindexProvider wrapperElementRef={messageListRef}>
-            {messageListItems.length === 0 && <EmptyChatMessage chat={chat} />}
-            {activeView.map(messageId => {
-              if (messageId.kind === 'dayMarker') {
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/message/MessageList.tsx
```diff
@@ -854,60 +854,81 @@ export const MessageListInner = React.memo(
...
+        // the entire list of messages every time you switch the chat.
+        aria-busy={!loaded}
+        // Do not announce timestamp updates (see `updateTimestamps()`).
+        // We might want to reconsider this line
+        // when message editing gets implemented.
+        aria-relevant='additions'
+        // TODO how do we _not_ announce messages that get loaded
+        // due to scrolling / jumping to messages?
+        // Add `aria-hidden` for just one render cycle to such messages?
+        // Or apply `aria-busy` to this element when inserting such messages?
```
WofWca, line 877: OK, this really needs to get fixed. Because we also `fetchMoreTop` as soon as you open a chat, resulting in the screen readers announcing those newly inserted messages.

I think we might want to add something similar to `viewState`. A variable that would apply `aria-busy` for a single render. We would set it in `messagelist.ts` for effects such as `fetchMoreMessagesTop`.

WofWca: Another idea: apply `aria-live="off"` when the chat is not fully scrolled to the bottom.

WofWca: Also see https://github.com/deltachat/deltachat-desktop/issues/4743

WofWca:
OK, we took another approach, and that is playing a "plop" sound when you get a message in the current chat. It doesn't read out the message, but you know that you received one, thus can go ahead and Shift + Tab to focus the message list to read them.

I don't know if this is the norm for messengers, but it should work. Let's reconsider later.

PR #4693, opened 2025-02-21
fix: a11y: aria-live for draft quote & attachment

WofWca:
This is handy for the Ctrl + Up shortcut,
and also adds confirmation for when using the regular way,
i.e. context meny -> reply.
Same for attachments.

TODO:
- [ ] I'm testing with NVDA and it seems like `aria-atomic` doesn't work for some reason. It only announces the changes. E.g. if the author's name didn't change, it won't announce it.
- [ ] NVDA doesn't announce when the quote gets removed, when using the Ctrl + Down shortcut.
- [ ] We announce the changes, but don't announce that it's a draft quote. Can this MR still be considered an improvement?
- [ ] Consider alternatives, see comments below.

### DIFF
--- a/packages/frontend/src/components/composer/Composer.tsx
+++ b/packages/frontend/src/components/composer/Composer.tsx
@@ -591,7 +591,16 @@ const Composer = forwardRef<
           // (`id='chat-section-heading'`) is probably enough.
         }
       >
-        <div className='upper-bar'>
+        <section
+          // Keep in mind that this element also changes when switching
+          // between chats.
+          // We probably want such changes to be announced still.
+          aria-live='polite'
+          // Announce quote / editing / attachment _removals_
+          // as well as text changes and node insertions.
+          aria-relevant='all'
+          className='upper-bar'
+        >
           {!messageEditing.isEditingModeActive ? (
             <>
               {draftState.quote !== null && (
@@ -599,11 +608,22 @@ const Composer = forwardRef<
                   className='attachment-quote-section is-quote'
                   aria-label={tx('menu_reply')}
                 >
-                  {/* Check that this is a "full" quote.
-                  TODO it would be nice to show a placeholder otherwise. */}
-                  {'text' in draftState.quote && (
-                    <Quote quote={draftState.quote} tabIndex={0} />
-                  )}
+                  <div
+                    // When changing the quoted message, e.g. with the Ctrl + Up
+                    // shortcut, we should read the author's name
+                    // even if it didn't change.
+                    //
+                    // Note that `aria-atomic` doesn't appear to work,
+                    // at least with NVDA, when it's a _descendant_ of the
+                    // `aria-live` element.
+                    aria-atomic='true'
+                  >
+                    {/* Check that this is a "full" quote.
+                    TODO it would be nice to show a placeholder otherwise. */}
+                    {'text' in draftState.quote && (
+                      <Quote quote={draftState.quote} tabIndex={0} />
+                    )}
+                  </div>
                   <CloseButton
                     onClick={removeQuote}
                     aria-label={tx('remove_quote')}
@@ -683,7 +703,7 @@ const Composer = forwardRef<
               />
             </div>
           )}
-        </div>
+        </section>
         <div className='lower-bar'>
           {!messageEditing.isEditingModeActive && !recording && (
             <MenuAttachment
--- a/packages/frontend/src/components/composer/Composer.tsx
+++ b/packages/frontend/src/components/composer/Composer.tsx
@@ -626,7 +626,14 @@ const Composer = forwardRef<
           // (`id='chat-section-heading'`) is probably enough.
         }
       >
-        <div className='upper-bar'>
+        <section
+          aria-live='polite'
+          // Announce quote / editing / attachment _removals_
+          // as well as text changes and node insertions.
+          aria-relevant='all'
+          aria-busy={draftIsLoading}
+          className='upper-bar'
+        >
           {!messageEditing.isEditingModeActive ? (
             <>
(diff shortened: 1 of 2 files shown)


WofWca:
About the politeness level: maybe we want `assertive`?

See https://w3c.github.io/aria/#aria-live

> For example, if assistive technologies can determine that a change occurred in response to a key press or a mouse click, the assistive technologies might present that change immediately even if the value of the [aria-live](https://w3c.github.io/aria/#aria-live) attribute states otherwise.

pvagner:
> About the politeness level: maybe we want `assertive`?

I don't think so.
In practice assertive causes screen reader to interupt what's currently being spoken if any and start speaking the aria-live change. I think in a chat app we definatelly like to have polite level.
I would say assertive should only be used in situations where aria live region is used as a substitution to a real accessibility implementation or when we can really justify the fact we'd like to be assertive.
For example in the past google docs used to have an offscreen aria live region and it worked like a hidden interface between the app and the screen reader. The editor was not really accessible as such and the app has been using assertive live region to force screen reader to speak what was happening within the editor.

packages/frontend/src/components/composer/Composer.tsx
```diff
@@ -535,11 +535,33 @@ const Composer = forwardRef<
           // (`id='chat-section-heading'`) is probably enough.
         }
       >
-        <div className='upper-bar'>
+        <section
+          // Keep in mind that this element also changes when switching
+          // between chats.
+          // We probably want such changes to be announced still.
```
WofWca: Hmm I don't think so. This is confusing and useless when switching chats, switching chats can be considered a form of "navigation", thus we must not announce the changes.

I would like to refactor `useDraft` such that it returns `loadingDraft`, which we can then use to only apply `aria-live` when we have finished loading the draft initially.
But there is another MR pending which moves `useDraft` to a separate file. Let's do this later to avoid conflicts.

pvagner:
Hello @WofWca ,

Today morning I've got a different idea on how to possibly resolve this issue in a different way, and I'd say it's less prone to variability of aria region implementation issues within screen readers and avoids possible timing issues.

When navigating the deltachat desktop UI the chat input entry has placeholder saying something like "Type in a message". This is also being used as a screen reader accessible label for the edit control.

When there is a content above the chat input entry such as a message I am replying to, or a different content, lets change the aria-label for the chat input entry that will notify screen reader users to the fact there is something important in there.

The downside is that this will only be accessible when the chat input is focused however that's the default use case anyways so I think that's a good trade off between reliability and usability.

What do you think?

WofWca:
Hi @pvagner !

Interesting idea. So basically something like `aria-label="Write a message; reply; attachment"`?

https://github.com/deltachat/deltachat-desktop/blob/1b9065b08f2d94d559c61f26fbc8511a91cf609b/packages/frontend/src/components/composer/ComposerMessageInput.tsx#L241-L271

I am afraid that changing `aria-label` alone will not make screen readers re-announce the input though?

Another possible way is to have the label on the wrapper region, i.e. the composer region, which wraps the attachment, reply, and the input:

https://github.com/deltachat/deltachat-desktop/blob/955b6f3bae4a9a9482a5ad4a4d321697b954b31a/packages/frontend/src/components/composer/Composer.tsx#L538-L546

pvagner:
Hello @WofWca 

Thanks for acknowledging my idea.

My assumption goes like this and I need to verify it and look through the code you are suggesting to take a look at.

* I will press ctrl+m to jump into the text input, screen reader will announce "Write a message". This is expected and should remain like this.
* I will press ctrl+up arrow currently message I'm replying to displays above the text entry field. Screen reader reports nothing right now. However as an addition to displaying the message above the text entry I suggest to change the aria label for the text entry to include the same text that is displayed above. Since the text entry is now focused most screen readers are expected to read its name again as it changes.
I will try to verify this and I'll post again.

Greetings

Peter

pvagner:
> Another possible way is to have the label on the wrapper region, i.e. the composer region, which wraps the attachment, reply, and the input:


My guess is that this won't be that usefull as the label of ancestors such as groupings and pannels are only read while navigating e.g. when focus moves from one part of the UI to the other one.

Here is an example that illustrates this feature

* When new chat has focus, pressing the tab key moves focus to the Edit group or to the Apps and media buttons. The screen reader reads the name of that button receiving the focus and before that it reads the chat name which is part of the ancestry.
* When chat input text entry has focus, we can use ctrl+page up and ctrl+page down to move to the previous / next chat. Label of that wrapping UI element changes however screen readers usually don't read it.

packages/frontend/src/components/composer/Composer.tsx
WofWca: Done.

WofWca:
OK I think at the current state this is still an improvement, even though in some ways it doesn't work as I'd like it to.
I think we can address the rest in follow-ups.

I re-tested this with Orca reader.

pvagner:
Huge thanks @WofWca  I'm testing this on linux with orca and it's working fine over here.
Pressing ctrl+up / ctrl+down when message input has focus causes the screen reader to read selected message for replying.

One additional thing I'd prefer for the future is that the label of the message input should change when writing a reply.
Currently it says "Write a message" and when there is a selected message for replying it should say "Write a reply".
It would be usefull indicator for knowing if a message for replying has already been selected when navigating with tab and shift+tab.

Thanks once again for working on it.

WofWca:
Thank you very much for the confirmation!

> Write a reply

Yes, let's keep that for a follow-up. This might need additional consideration.

nicodh (approved):
Reviewed by reading

Didn't test, but there has been some user testing so it should be fine!

PR #4696, opened 2025-02-21
improvement: hide chat name for single chat search

WofWca:
Also hide the "Archived" and "Contact Request" status.
This info is redundant when searching in a particular single chat.

Before / after:
image image


nicodh (approved):
Reviewed by testing

PR #4699, opened 2025-02-22
Tauri: HTML Email View

Simon-Laux:
- use opener plugin instead of shell plugin
- implement view/window for html emails

## Progress

- [x] window / webview
- [x] resize
- [x] content protection
- [x] header
- [x] custom scheme - load email and deny internet access unless allowed over CSP
- [x] disable JS (needs upstream tauri pr)
   - https://github.com/tauri-apps/wry/pull/1496
   - https://github.com/tauri-apps/tauri/pull/12821
- [x] misc: reload webview ~~& non persistent data store for macOS~~ ( the non persistent data store is already used over incognito mode)
    - https://github.com/tauri-apps/wry/pull/1500
    - https://github.com/tauri-apps/tauri/pull/12818
- [x] ~~load http traffic over core to respect proxy settings~~ disable remote resource fetching when proxy is active instead


packages/target-tauri/src-tauri/src/html_window/commands.rs
```diff
@@ -0,0 +1,167 @@
...
+
+    let instance_state = block_on(html_instances_state.get(webview.window().label()))
+        .ok_or(Error::WindowNotFoundInState)?;
+
+    let always_load_remote_images = CheckMenuItem::new(
+        &app,
+        "tx('always_load_remote_images')",
+        true,
+        always_load,
+        None::<&str>,
```
ralphtheninja, line 101: Quick rust question. What's this? Is that passing in a `None` but typed as a `&str`?

packages/target-tauri/src-tauri/src/html_window/commands.rs
```diff
@@ -0,0 +1,167 @@
...
+        }
+        if event.id() == show_warning.id() {
+            desktop_settings.set(HTML_EMAIL_WARNING_KEY, !warning);
+        }
+    });
+    Ok(())
+}
+
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
```
ralphtheninja, line 133: Does this mean all properties/keys of the struct will be camel cased when serialized?

packages/target-tauri/src-tauri/src/html_window/commands.rs
Simon-Laux: a turbofish https://turbo.fish/about
basically this syntax sets the type of the option. it is also sometimes used when collecting an iterator https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.collect

packages/target-tauri/src-tauri/src/html_window/commands.rs
Simon-Laux: yes, https://serde.rs/container-attrs.html#rename_all

Issue #4128, opened 2024-09-12
unexpected "Escape" key behaviours


r10s:
the "Escape" key is working only in some situations, and not always predictable.

some inconsistencies:

- [x] settings -> esc works
- [x] settings -> chats and media -> esc does not work (there is a dedicated issue for that at https://github.com/deltachat/deltachat-desktop/issues/3868)
- [x] settings -> chats and media -> back -> esc does not work
- [x] enlarge image -> esc works
- [ ] enlarge image -> context menu -> esc closes image and keeps context menu
- [x] enlarge image -> context menu -> close context menu -> esc does not work

EDIT: some more:

- [x] chat -> context menu -> esc works
- [ ] chat -> reaction menu -> esc does not work

WofWca:
FYI some of these might get resolved in the Blueprint removal MR.

maxphilippov:
Turns actually none of these are solved by #4006 cause a lot of the dialogs are actually implemented as a switch between different states rather than stacking, but this looks like a pretty straightforward refactor, I'll probably do that after Blueprint removal PR is merged since there's a lot of Dialog-related code

maxphilippov:
> * enlarge image -> context menu -> esc closes image and keeps context menu

Just as an update for someone who's gonna work on this. On current `main` after Blueprint removal:
* enlarge image -> context manu -> esc closes both image and menu
Should close context menu first, I believe, so this is still a bug, just different behaviour

> * enlarge image -> context menu -> close context menu -> esc does not work

Escape works now

> * settings -> esc works
> * settings -> chats and media -> esc does not work (there is a dedicated issue for that at [Settings should be closable in sub-pages with escape key #3868](https://github.com/deltachat/deltachat-desktop/issues/3868))
> * settings -> chats and media -> back -> esc does not work

These are fixed in #4205

WofWca:
As I said in https://github.com/deltachat/deltachat-desktop/pull/5191#pullrequestreview-2899513666, this is also the case in "New Chat" dialog, where contacts have context menu.

A more generic solution is preferable. For example, utilizing Popover and dialog API. For example, see how closing the "add members" dialog that is on top of "Create Group" dialog doesn't close the "create group" dialog.

WofWca:
Let's reopen, for

> this is also the case in "New Chat" dialog

ell1e:
[...]

PR #4702, opened 2025-02-24
Handle esc key in settings

nicodh:
partly solves #4128

### DIFF
--- a/packages/frontend/src/components/Settings/Settings.tsx
+++ b/packages/frontend/src/components/Settings/Settings.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react'
+import React, { useState, useEffect, useRef } from 'react'
 
 import { useSettingsStore } from '../../stores/settings'
 import { SendBackupDialog } from '../dialogs/SetupMultiDevice'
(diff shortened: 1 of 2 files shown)


nicodh:
Simplified solution instead https://github.com/deltachat/deltachat-desktop/pull/4205 which has too many changes that have impact on all dialogs and the behaviour of buttons etc.

packages/frontend/src/components/Settings/Settings.tsx
```diff
@@ -34,6 +34,19 @@ export default function Settings({ onClose }: DialogProps) {
   const tx = useTranslationFunction()
   const [settingsMode, setSettingsMode] = useState<SettingsView>('main')
 
+  useEffect(() => {
+    document.onkeydown = (evt: KeyboardEvent) => {
+      if (
```
nicodh, line 39: Of course it is not ideal to use the generic document.onkeydown here but it works and solves the issue. Once we need another document.onkeydown we have to look for a better solution

Simon-Laux (approved):
code LGTM, did not test

Simon-Laux:
just noticed that the event listeners are not cleaned up again, useEffect functions should return a callback for cleanup.

Simon-Laux:
> As you see I noticed that too and added an AbortController which will remove the eventListener if

ok the point is yours. maybe add a comment about that it removes the event listener, I didn't knew that new web feature and assumed it would just stop the event listener.

Simon-Laux:
but still what happens when you leave the settings, there is still a leak, better call the abort controller in the cleanup function. But then again I wonder why you don't just use the proven pattern that we use elsewhere for event listeners in use effects

Simon-Laux (changes requested):
there are currently some issues and questions that need to be resolved first before we merge it see previous comment and also I wonder why there are two use effects and listeners now

nicodh:
> But then again I wonder why you don't just use the proven pattern that we use elsewhere for event listeners in use effects:

Didn't see that, so of course that is simpler and easier. I adapted the listener in useEffect

Simon-Laux (approved):

Issue #4548, opened 2025-01-24
sticker picker previews not displayed if folder name has %


adbenitez:
DC version: 1.51.0 (git: v1.33.0-1302-gfd2dfa1d)

if you pick one of them they are displayed properly in chat, it is only in the thumbnails where they are not displayed

![Image](https://github.com/user-attachments/assets/4e7a1f44-3447-4339-a214-cd08cd1d6a50)

WofWca:
The relevant code:

https://github.com/deltachat/deltachat-desktop/blob/f336a5ad3085aeb5826e71ffd3c1e4149c7c42a6/packages/frontend/src/components/composer/EmojiAndStickerPicker.tsx#L91

https://github.com/deltachat/deltachat-desktop/blob/f336a5ad3085aeb5826e71ffd3c1e4149c7c42a6/packages/frontend/src/components/composer/EmojiAndStickerPicker.tsx#L177-L185

On 1.52.0 (git: v1.33.0-1328-g9b66ba51c) if you use just `%` instead of `%20`, you get

```
"uncaughtError"	{"message":"URI malformed","stack":"URIError: URI malformed\n    at decodeURIComponent (<anonymous>)\n    at Function.<anonymous> (file:///C:/Users/user/AppData/Local/Programs/DeltaChat/resources/app.asar/bundle_out/index.js:31:11045)"}
```

We probably don't want to use just the filename as `src`. Perhaps use [`pathToFileURL()`](https://nodejs.org/api/url.html#urlpathtofileurlpath-options) (https://stackoverflow.com/questions/20619488/how-to-convert-local-file-path-to-a-file-url-safely-in-node-js).

Other such places need to be considered as well. For example, we had a lot of problems with thumbnails: https://github.com/deltachat/deltachat-desktop/pull/4247.

Simon-Laux:
Stickers are outside of blob folder. maybe decode uri component is enough for electron, but tauri needs a custom scheme and browser needs a custom url

Simon-Laux:
Actually the problem seems to be that we decode the uri. The simplest solution would be that the sticker bot gets updated to not uri escape folder names. Then the zip file and folder names will also look better as a positive side effect.

PR #4707, opened 2025-02-26
tauri: add support for sticker picker

Simon-Laux:
Before this pr the stickers where shown in the picker, but failed to load, this pr adds the `dcsticker://` scheme to fix that.

### DIFF
--- a/packages/target-tauri/runtime-tauri/runtime.ts
+++ b/packages/target-tauri/runtime-tauri/runtime.ts
@@ -331,6 +331,27 @@ class TauriRuntime implements Runtime {
     }
     return ''
   }
+  transformStickerURL(sticker_path: string): string {
+    const matches = sticker_path.match(/.*(:?\\|\/)(.+?)\1stickers\1(.*)/)
+    // this.log.info({ transformStickerURL: sticker_path, matches })
+
+    if (matches) {
+      let filename = matches[3]
+      if (decodeURIComponent(filename) === filename) {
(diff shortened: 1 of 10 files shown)


packages/target-tauri/runtime-tauri/runtime.ts
```diff
@@ -331,6 +331,27 @@ class TauriRuntime implements Runtime {
     }
     return ''
   }
+  transformStickerURL(sticker_path: string): string {
+    const matches = sticker_path.match(/.*(:?\\|\/)(.+?)\1stickers\1(.*)/)
```
nicodh, line 335: would/should this fix https://github.com/deltachat/deltachat-desktop/issues/4548 ?
At least in my test it didn't.
In electron runtime a return sticker_path.replace(/%20/g, ' ') fixed the problem (decodeURI did not)

nicodh (approved):
Reviewed by reading & testing

packages/target-tauri/runtime-tauri/runtime.ts
Simon-Laux: the whole premise of https://github.com/deltachat/deltachat-desktop/issues/4548 is weird IMO, you normally don't do folders that have uri encoded names. I believe adb should fix his bot, there is no real reason for having this.

Simon-Laux: > should this fix

no, otherwise I'd have linked it or mentioned it.

PR #4710, opened 2025-02-27
fix: tauri: qr code scan from clipboard text

Simon-Laux:


nicodh (approved):
Reviewed by testing & reading

PR #4712, opened 2025-02-28
update translations

nicodh:


WofWca (approved):
Did not looks through the changes, but only the translation files are affected with this MR,

Issue #2599, opened 2022-03-10
Voice message improvements


Simon-Laux:
- [ ] ability to record voice messages
- [ ] only play one voice message at a time, if there is already a voice message playing pause it
- [ ] have a player for voice messages, so users can listen to a voice message while reading an other chat and jump to the next / previous one
- [ ] maybe also restyle the voice message player to look better

adbenitez:
a bug related to this: if you play a voice message and then send a message to reply the voice message in the chat while you are listening, the audio is stopped

sysfu:
Also, it's not possible to reply to a message with a voice recording. Maybe a separate issue should be opened for the mobile client.

sysfu:
I would also like to see the voice player automatically advance and play the next voice message, like TeleGRIDS does.

JoshWho:
100% support the need for voice on the desktop clients.

sysfu:
Is there a way to sponsor this feature request directly, like a bug bounty? Session app has ability to record voice messages from their desktop client and this is the biggest pain point of using Delta Chat since I gave up my tardphone a year ago.

Simon-Laux:
you could propose sponsoring it to delta@merlinux.eu, we don't have a bug bounty program yet, maybe it could make sense looking into it, are there some bug bounty services that you can recommend?

sysfu:
> you could propose sponsoring it to [delta@merlinux.eu](mailto:delta@merlinux.eu), we don't have a bug bounty program yet, maybe it could make sense looking into it, are there some bug bounty services that you can recommend?

I don't really have much experience with bounty services tbh. The two main ones I have heard of are issuehunt and bountysource. I know that there has been some controversy with Bountysource over the years. https://blog.elementary.io/goodbye-bountysource-hello-github-sponsors/

farooqkz:
How about noise reduction? I think many browsers already support that. or at least it's part of the spec for gUM.

Related link: https://developer.mozilla.org/en-US/docs/Web/API/MediaTrackConstraints/noiseSuppression
We could further reduce noise using AI/ML maybe?

farooqkz:
> > How about noise reduction? I think many browsers already support that. or at least it's part of the spec for gUM.
> 
> we don't even have voice message recording in desktop and you are already talking about how to improve it eyes 

Think it of a suggestion of an item to add in the roadmap of voice messages ;)

sysfu:
[...]

PR #4715, opened 2025-02-28
feature: add voice message

nicodh:
- uses MediaRecorder which is restricted to audio/webm or audio/mp4 codecs

 based on pr #3456
related to #2599


nicodh:
This implementation can not be merged as it is, since the audio format is audio/webm which is one of the only supported audio formats of Chromes MediaRecorder, but iOS can't play that format.

So to proceed we have to either add some encoder to a compatible format (mp3, or preferable aac) or we wait for iOS to support the format by using some encoder library.

Since there might be plans to use ogg or opus audio format for voice calls anyway, the usage of extended audio formats in iOS will come sooner or later

PR #4721, opened 2025-03-01
Show dialog if backup is too new

### LINKED ISSUE #4709: adapt to BackupTooNew QR codes
core 1.156 returns a new QR code result, `BackupTooNew`, jsponrpc is already updated at #6580

when that QR code result is returned, an alert with the string  `multidevice_receiver_needs_update` should be shown, see https://github.com/deltachat/deltachat-android/pull/3629 and https://github.com/deltachat/deltachat-ios/pull/2619 for screenshot.

this issue is a blocker as we'll probably have incompatible backup/add-second-device in one of the next version, and we want to be prepared :)

nicodh:
resolves #4709

### DIFF
--- a/packages/frontend/src/hooks/useProcessQr.ts
+++ b/packages/frontend/src/hooks/useProcessQr.ts
@@ -149,6 +149,14 @@ export default function useProcessQR() {
 
       const { qr } = parsed
 
+      if (qr.kind === 'backupTooNew') {
+        await openAlertDialog({
+          message: tx('multidevice_receiver_needs_update'),
+        })
+        callback?.()
+        return
+      }
+
       // Some actions can only be executed when the user already has an account
       // and is logged in
       const isLoggedIn = await BackendRemote.rpc.isConfigured(accountId)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased][unreleased]
 
 ### Added
+- show error message if backup version is not compatible #4721
 - show "Edited" in the message's status line (if it's edited) #4697
 - add "learn more"-button to manage-key section that links to local help #4684
 - add a search field to help page #4691


nicodh:
For testing paste a string starting with DCBACKUP3 like "DCBACKUP3:fooo" in the add second device qr scanning dialog

packages/frontend/src/hooks/useProcessQr.ts
```diff
@@ -149,6 +149,14 @@ export default function useProcessQR() {
 
       const { qr } = parsed
 
+      if (qr.kind === 'backupTooNew') {
+        await openAlertDialog({
+          message: tx('multidevice_receiver_needs_update'),
+        })
+        callback?.()
```
ralphtheninja, line 156: Is this some TypeScript magic?

ralphtheninja: Nvm, I looked it up. I'm guessing this is checking if the callback is defined before invoking it.

ralphtheninja: A general comment about this `useProcessQR()` function. It looks like it's only used in one place, which doesn't pass in a callback, so maybe we should just remove all the callback-invoking code?

ralphtheninja (approved):

packages/frontend/src/hooks/useProcessQr.ts
nicodh: It's used multiple times and sometimes witha callback:
https://github.com/deltachat/deltachat-desktop/blob/e5d32b6a6736d144442c8371339247dcf013877b/packages/frontend/src/components/dialogs/SetupMultiDevice/ReceiveBackupDialog.tsx#L52
or
https://github.com/deltachat/deltachat-desktop/blob/e5d32b6a6736d144442c8371339247dcf013877b/packages/frontend/src/components/dialogs/QrCode.tsx#L239
Which is for closing a dialog etc.

PR #4722, opened 2025-03-01
fix: keep white-space in dialogs

nicodh:
Should be carefully checked for unexpected side effects.

AlertDialogs sometimes show an error besides a translated message, which might look bad if it's a long error message with many line breaks.

I also removed the alternate message type ReactNode from AlertDialog since it was not used (according to type checks)

#skip-changelog

### DIFF
--- a/packages/frontend/scss/_global.scss
+++ b/packages/frontend/scss/_global.scss
@@ -253,3 +253,8 @@ audio {
   padding: 0;
   border: 0;
 }
+
+// used in alert & confirmation dialogs
+p.whitespace {
+  white-space: break-spaces;
+}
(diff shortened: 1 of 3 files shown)


nicodh:
Before:

image

After:

image

packages/frontend/scss/_global.scss
```diff
@@ -253,3 +253,8 @@ audio {
   padding: 0;
   border: 0;
 }
+
+// used in alert & confirmation dialogs
+p.whitespace {
```
nicodh, line 258: I'm not sure where would be the best place for this style rule

WofWca (approved):
I didn't check all the dialogs, but `white-space: break-spaces;` should not be that breaking, because it still allows for wrapping.

PR #4727, opened 2025-03-03
Update core to 1.156.2

nicodh:


WofWca (approved):
Just an upgrade. Looks like there is no need to add any more changelog entries, since it only fixes the bugs that were introduced in a version of core that is not in DC Desktop releases.

PR #4730, opened 2025-03-03
Update translations

nicodh:


r10s (approved):
lgtm, still, usually, translations and core updates do not need an approval

PR #4732, opened 2025-03-03
chore: prepare release 1.54.2

nicodh:


WofWca:
Do we not want to do a test release? There was quite a lot of changes. Basically the only reason this is not a minor (1.56) release is that we want to release message editing in 1.56.

nicodh:
I thought we decided to have not separate version numbers anymore for test releases, (only if we expect a looong test period). So I would mark this release as a Prerelease, announce it as test release in the forum and wait some days before converting it to a final release and upload it to the stores and flatpak.
That would result in less release overhead...

r10s (approved):

ralphtheninja (approved):

WofWca (approved):
Alright, as long as we'll have some testing, it's good.
Code-wise looks alright as well.

PR #4736, opened 2025-03-04
fix: urlencode sticker path

nicodh:
this fixes the problem described in #4548 

I think we should at least support valid foldernames and a % character is a valid character, but misinterpreted by the browser if not encoded. https://datatracker.ietf.org/doc/html/rfc3986#section-2.4

I found this https://github.com/sindresorhus/file-url/blob/main/index.js which still covers some more cases.

Alternatively we could provide a wrapper method for [urlpathtofileurlpath](https://nodejs.org/api/url.html#urlpathtofileurlpath-options) in ipc backend and call it in electron runtime


ralphtheninja (approved):

packages/target-electron/runtime-electron/runtime.ts
```diff
@@ -319,8 +319,12 @@ class ElectronRuntime implements Runtime {
     }
   }
   transformStickerURL(sticker_path: string): string {
-    return sticker_path
+    return encodeURI(`file://${sticker_path}`).replace(
+      /[?#]/g,
+      encodeURIComponent
+    )
```
Simon-Laux, line 325: what is this replace for?

nicodh: It encodes ? and # if they are part of a folder name. I tested it, if they are neccessary. For a folder named various%20animals%201.1?id=3#45 they are neccessary

Simon-Laux: ok

PR #4740, opened 2025-03-05
tauri: webxdc

Simon-Laux:
adding the webxdc feature to the tauri edition

## Progress

- [x] webxdc window handling
- [x] webxdc data - datastore / different session for each webxdc and delete them when webxdc message or account is deleted
- [x] serving files from xdc zip file
- [x] webxdc titlebar
- [x] use translation function
- [x] implement webxdc.js/webxdc-api
- [x] make it work on windows
- [x] don't use account id and message id as label because the label is visible in the webxdc app, which means an attacker could potentially get the amount of messages and accounts the user has at least.
- ~~hardening (make sure webrtc is blocked)~~ will be done in follow up pr
- [x] changelog entries

### Added
- tauri: support for webxdc

### Changed
- switch to account the webxdc is from when sending to chat (tauri and electron edition)

### Fixed
- prevent dragging around of webxdc icon


WofWca:
Hey, I made https://github.com/deltachat/deltachat-desktop/pull/4795, which implements white-list approach for commands, so this MR will probably need to be adjusted if we want that one merged.

WofWca:
It'd be nice to make the tests not fail https://github.com/deltachat/deltachat-desktop/actions/runs/14061055371/job/39371665576?pr=4740

PR #4741, opened 2025-03-05
refactor: dialog footer button styles

nicodh:
- no transform: uppercase
- buttons visible as buttons
- min-width
- more primary / secondary hierarchies

Some screens: (the focus depends on the action before the screenshot and is dynamic of course)
new vs. old

secondary - danger
image

secondary - primary
image

tbd: cancel or close ??

image

min width
image

Dark mode

image

### DIFF
--- a/packages/frontend/src/components/Button/style.module.scss
+++ b/packages/frontend/src/components/Button/style.module.scss
@@ -11,7 +11,6 @@
   justify-content: center;
   padding: 12px 15px;
   text-align: center;
-  text-transform: uppercase;
   user-select: none;
 
   &:hover {
@@ -36,10 +35,12 @@
   &.secondary {
     background-color: var(--buttonSecondaryBackground);
     color: var(--buttonSecondaryText);
+    border: 1px solid #ccc;
   }
 
   &.danger {
-    color: var(--buttonDangerText);
+    color: var(--buttonPrimaryText);
+    background-color: var(--buttonDangerText);
   }
 
   &.active {
--- a/packages/frontend/src/components/Button/style.module.scss
+++ b/packages/frontend/src/components/Button/style.module.scss
@@ -39,8 +39,8 @@
   }
 
   &.danger {
-    color: var(--buttonPrimaryText);
-    background-color: var(--buttonDangerText);
+    color: var(--buttonDangerText);
+    background-color: var(--buttonDangerBackground);
   }
 
   &.active {
--- a/packages/frontend/src/components/Dialog/CloseFooterAction.tsx
+++ b/packages/frontend/src/components/Dialog/CloseFooterAction.tsx
@@ -13,7 +13,9 @@ export default function CloseFooterAction({ onClose }: DialogProps) {
   return (
     <DialogFooter>
       <FooterActions>
-        <FooterActionButton onClick={onClose}>{tx('close')}</FooterActionButton>
+        <FooterActionButton styling='primary' onClick={onClose}>
+          {tx('close')}
+        </FooterActionButton>
       </FooterActions>
     </DialogFooter>
   )
--- a/packages/frontend/src/components/dialogs/AlertDialog.tsx
+++ b/packages/frontend/src/components/dialogs/AlertDialog.tsx
@@ -39,7 +39,7 @@ export default function AlertDialog({
       </DialogBody>
       <DialogFooter>
         <FooterActions>
-          <FooterActionButton styling='primary' onClick={onClick}>
+          <FooterActionButton styling='secondary' onClick={onClick}>
             {okBtnLabel || tx('ok')}
           </FooterActionButton>
         </FooterActions>
--- a/packages/frontend/src/components/dialogs/AlertDialog.tsx
+++ b/packages/frontend/src/components/dialogs/AlertDialog.tsx
@@ -39,7 +39,7 @@ export default function AlertDialog({
       </DialogBody>
       <DialogFooter>
         <FooterActions>
-          <FooterActionButton onClick={onClick}>
+          <FooterActionButton styling='primary' onClick={onClick}>
             {okBtnLabel || tx('ok')}
           </FooterActionButton>
         </FooterActions>
--- a/packages/frontend/themes/_themebase.scss
+++ b/packages/frontend/themes/_themebase.scss
@@ -27,7 +27,8 @@ $hover-contrast-change: 2%;
   --buttonPrimaryText: #{$textNavBar};
   --buttonSecondaryBackground: #{$bgSecondary};
   --buttonSecondaryText: #{$textSecondary};
-  --buttonDangerText: #{$colorDanger};
+  --buttonDangerText: #fff;
+  --buttonDangerBackground: #{$colorDanger};
   --buttonDisabledBackground: #{$bgSecondary};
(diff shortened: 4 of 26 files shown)


nicodh:
These changes are mainly based on a recent discussion with @r10s

WofWca:
> tbd: cancel or close ??

If you actually scan the code, then "cancel" doesn't make sense. But not too important I'd say.

packages/frontend/src/components/Dialog/CloseFooterAction.tsx
```diff
@@ -13,7 +13,9 @@ export default function CloseFooterAction({ onClose }: DialogProps) {
   return (
     <DialogFooter>
       <FooterActions>
-        <FooterActionButton onClick={onClose}>{tx('close')}</FooterActionButton>
+        <FooterActionButton styling='primary' onClick={onClose}>
+          {tx('close')}
+        </FooterActionButton>
```
WofWca, line 18: IDK about this. It's not the primary action e.g. in the "new chat" dialog

image

packages/frontend/src/components/Button/style.module.scss
```diff
@@ -36,10 +35,12 @@
...
     background-color: var(--buttonSecondaryBackground);
     color: var(--buttonSecondaryText);
+    border: 1px solid #ccc;
   }
 
   &.danger {
-    color: var(--buttonDangerText);
+    color: var(--buttonPrimaryText);
+    background-color: var(--buttonDangerText);
   }
```
WofWca, line 44: The contrast of the danger and the primary button is low

image

image

nicodh:
> > tbd: cancel or close ??
> 
> If you actually scan the code, then "cancel" doesn't make sense. But not too important I'd say.

Good point. I always pasted and never used the camera (since on desktop it is blocked) so I oversaw the "default" use case.

packages/frontend/src/components/dialogs/AlertDialog.tsx
```diff
@@ -39,7 +39,7 @@ export default function AlertDialog({
       </DialogBody>
       <DialogFooter>
         <FooterActions>
-          <FooterActionButton onClick={onClick}>
+          <FooterActionButton styling='primary' onClick={onClick}>
```
WofWca, line 42: I'm not sure about making those "dismiss" buttons the primary color. It makes it seem like some action is about to be performed.
There are a few more such places.

WofWca (commented):
Overall I like the new buttons more, e.g. in those smaller "yes / no" dialogs, but not in cases where the "Material" theme is prominently used, i.e. in the "QR" dialog and in the "New Group" dialog, maybe there is more. The tabs in the QR dialog are uppercase, and much bigger than the buttons, which looks weird to me.

Edit: code-wise the MR looks good.

image

Simon-Laux:
I don't quite like the design yet, but it is already a huge improvement 👍

packages/frontend/src/components/Dialog/CloseFooterAction.tsx
nicodh: So the question is: is a single button always the "primary" action or is there no idea of "primary" - "secondary" at all.

packages/frontend/src/components/Button/style.module.scss
nicodh: Can you suggest a color theme here, which has better contrast?

packages/frontend/src/components/Dialog/CloseFooterAction.tsx
WofWca: I'm sure it's specified somewhere in the Material guidelines (not to say that this is the law of design, but If everyone follows it, we should be too probably)

packages/frontend/src/components/dialogs/AlertDialog.tsx
nicodh: Same topic as above. Are single buttons "primary" actions or not?

packages/frontend/src/components/Button/style.module.scss
WofWca: In this particular case, I think a close button at the top would make more sense.
But perhaps this is for a separate merge request.

packages/frontend/src/components/Dialog/CloseFooterAction.tsx
WofWca: In this particular case, I think a close button at the top would make more sense.
But perhaps this is for a separate merge request.

r10s:
in general, "primary" primary button draw attention, making it obvious what the user should do next, what should be tapped in the "most probably, happy path"

that said, in case of "create group", it is clear that the button should be primary.

in case of "close", a normal secondary button seems to be fine as well (the action is something else)

packages/frontend/src/components/Button/style.module.scss
r10s: going up to white, seems reasonable. the color of the button, however should stay. these colors are used widely in a lot of themes, including apple, they cannot be all wrong :)

nicodh:
Applied 2 requested changes from the discussion:
- white color for buttonDangerText for better contrast
- secondary action for AlertDialog Ok

r10s (commented):
literally lgtm :)
cannot say much to the code

packages/frontend/src/components/Button/style.module.scss
WofWca: The contrast is still not satisfactory.

image

It used to be better, on dark theme at least:

image

packages/frontend/themes/_themebase.scss
```diff
@@ -27,7 +27,8 @@ $hover-contrast-change: 2%;
   --buttonPrimaryText: #{$textNavBar};
```
WofWca, line 27:
```suggestion
  --buttonPrimaryText: #{$textPrimary};
```

The same adjustment as for the danger button. This is `#fff`.

WofWca:
```suggestion
  --buttonPrimaryText: #fff;
```

same as for "danger"

Simon-Laux: nitpick: I'd appreciate the use of theming vars for all colours if possible

nicodh: I agree but this should be solved in a  separate issue. The naming is inconsistent:

 --buttonText: #{$colorPrimary};

 --buttonPrimaryBackground: #{$colorPrimary};
 --buttonPrimaryText: #{$textNavBar};
 
  --buttonSecondaryBackground: #{$bgSecondary};
  --buttonSecondaryText: #{$textSecondary};

I introduced new btn colors as theme variables now as a first step.

Looking at the themes:
what's the reason for having these "prototype" themes?

r10s:
thinking it over: for the primary button left in the "show qr code": that is a bit off. i'd make both buttons secondary. primary action is scanning

nicodh:
> thinking it over: for the primary button left in the "show qr code": that is a bit off. i'd make both buttons secondary. primary action is scanning

You mean the "Copy Link" button?

nicodh:
> thinking it over: for the primary button left in the "show qr code": that is a bit off. i'd make both buttons secondary. primary action is scanning

Done

r10s (approved):

PR #4744, opened 2025-03-05
tauri: fix build and csp on windows

### LINKED ISSUE #4742: mail_view_builder compiler error
value used after move

```
error[E0382]: use of moved value: `mail_view_builder`
   --> packages/target-tauri/src-tauri/src/html_window/mod.rs:204:29
    |
130 |     let mut mail_view_builder = tauri::webview::WebviewBuilder::new(
    |         --------------------- move occurs because `mail_view_builder` has type `tauri::WebviewBuilder<tauri_runtime_wry::Wry<tauri::EventLoopMessage>>`, which does not implement the `Copy` trait
...
200 |         let data_dir = mail_view_builder.data_directory(tmp_data_dir.clone());
    |                        ----------------- value moved here
...
204 |         mail_view_builder = mail_view_builder.incognito(true);
    |                             ^^^^^^^^^^^^^^^^^ value used here after move

```

Simon-Laux:
- **fix csp for windows**
- **fix compile issue on windows and linux**

closes #4742

#skip-changelog

### DIFF
--- a/packages/target-tauri/src-tauri/tauri.conf.json5
+++ b/packages/target-tauri/src-tauri/tauri.conf.json5
@@ -26,10 +26,10 @@
         "style-src": "'self' 'unsafe-inline'",
         "font-src": "'self'",
         "script-src": "'self' 'wasm-unsafe-eval'",
-        "worker-src": "blob: ",
-        "child-src": "blob: ",
-        "img-src": "'self' data: blob: dcblob: webxdc-icon: ",
-        "media-src": "'self' dcblob:"
+        "worker-src": "blob: http://blob.localhost",
+        "child-src": "blob: http://blob.localhost",
+        "img-src": "'self' data: blob: http://blob.localhost dcblob: http://dcblob.localhost webxdc-icon: http://webxdc-icon.localhost",
+        "media-src": "'self' dcblob: http://dcblob.localhost"
       },
       // TODO postponed to later
       // "pattern": {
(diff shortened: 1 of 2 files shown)


packages/target-tauri/src-tauri/tauri.conf.json5
```diff
@@ -26,10 +26,10 @@
...
         "font-src": "'self'",
         "script-src": "'self' 'wasm-unsafe-eval'",
-        "worker-src": "blob: ",
-        "child-src": "blob: ",
-        "img-src": "'self' data: blob: dcblob: webxdc-icon: ",
-        "media-src": "'self' dcblob:"
+        "worker-src": "blob: http://blob.localhost",
+        "child-src": "blob: http://blob.localhost",
+        "img-src": "'self' data: blob: http://blob.localhost dcblob: http://dcblob.localhost webxdc-icon: http://webxdc-icon.localhost",
+        "media-src": "'self' dcblob: http://dcblob.localhost"
```
WofWca, line 32: Is there no other way? This practically enables access to all localhost HTTP servers. Try `python3 -m http.server 8080`, then go to http://blob.localhost:8080 : you'll get to the server.

Simon-Laux: maybe there is a way to only enable it for windows & android

Simon-Laux: I'll add this to the my todo list, we will fix it later

PR #4751, opened 2025-03-08
Refactor webxdc lookup

ralphtheninja:
Move out the lookup code by `event.sender` into a separate function for better readability.

#skip-changelog Refactor

### DIFF
--- a/packages/target-electron/src/deltachat/webxdc.ts
+++ b/packages/target-electron/src/deltachat/webxdc.ts
@@ -715,6 +715,7 @@ export default class DCWebxdc {
         }
       }
     )
+
     ipcMain.handle(
       'webxdc:instance-deleted',
       (_ev, accountId: number, instanceId: number) => {
@@ -732,6 +733,7 @@ export default class DCWebxdc {
         s.clearCache()
       }
     )
+
     ipcMain.handle(
       'open-maps-webxdc',
       async (evt, accountId: number, chatId?: number) => {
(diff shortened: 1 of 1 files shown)


packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -796,7 +798,7 @@ export default class DCWebxdc {
         }
       }
     )
-  }
+  } // end of DeltaChatController constructor
```
ralphtheninja, line 786: The constructor is very long and it's hard to know which scope you're in while reading this code.

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -522,52 +523,46 @@ export default class DCWebxdc {
...
     })
 
     ipcMain.handle('webxdc.exit', async event => {
-      const key = Object.keys(open_apps).find(
-        key => open_apps[key].win.webContents === event.sender
-      )
-      if (key) {
-        open_apps[key].win.loadURL('about:blank')
-        open_apps[key].win.close()
+      const app = lookupAppFromEvent(event)
```
ralphtheninja, line 531: Lets work directly with `app` instead of `key`.

packages/target-electron/src/deltachat/webxdc.ts
Simon-Laux: there is also `// #region` `// #endregion` folding support in editors like vscode/ium

Simon-Laux (approved):

packages/target-electron/src/deltachat/webxdc.ts
ralphtheninja: I think I'd like to break this constructor apart a bit later anyway.

Issue #2518, opened 2022-01-14
Option to setup autostart for Deltachat


Simon-Laux:
Option to set up auto start for Deltachat on system start. If we cannot support a certain system/os we can link to a tutorial/guide how the user can do it manually.

lets collect the approaches how we could do it on the different OS/platforms/distribution-formats:

### Windows (normally installed)
For windows (non store/appx) we could create a link/shortcut file in the StartMenu/Autostart folder, that is set to start Deltachat with the minimised argument. There is probably also a way to do this via the windows registry.

### Windows (portable)
Either don't offer option at all (and display a manual guide), or do the same method as for the installed version but put a warning before it `"Does not work with portable versions on a path that changes"`

### Windows (apps, Store version)
maybe there is a special UWP api for it or sth similar?

### MacOS
?, probably not the biggest deal to not have it, because macOS normally restarts all previous applications after a reboot.

### Linux (normally installed)
?

### Linux (AppImage)
?

### Linux (flatpak)
?

PS: we might need to modify electron builder for this so that it allows us to write what distributable format it is before packing the app into that format.

jose1711:
> ### Linux (normally installed)
> 
> ?

copy `deltachat.desktop` file to `~/.config/autostart` for a particular user (as per [XDG Autostart](https://wiki.archlinux.org/title/XDG_Autostart)).

debilin:
We can take a look at [ZapZap](https://github.com/rafatosta/zapzap) build for Flatpak. I think they did a good job for Flatpak to autostart.

the-Arioch:
> because macOS normally restarts all previous applications after a reboot.

So does Windows BTW, if DeltaChat would only want it.

* https://agsci.psu.edu/it/how-to/topics/hardware/enabling-restart-apps-on-windows-computers
* https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-registerapplicationrestart?redirectedfrom=MSDN

nicodh:
Did some research on this: as a first step it should be enough to offer auto start for Win & Mac using electrons built in functionality.
https://www.electronjs.org/docs/latest/api/app#appsetloginitemsettingssettings-macos-windows

For linux it seems harder to set it, (i.e. Signal does offer it only for Win & Mac)
But linux users will anyway be more used to handle auto start manually

Simon-Laux:
very nice that electron has an option for that, that saves us the annoying editing of the registry.

> For linux it seems harder to set it

[...]

PR #4754, opened 2025-03-08
Autostart on Tauri

Simon-Laux:
- **fix: wait for setting to be applied before calling callback**
- **tauri: add autostart**

related to https://github.com/deltachat/deltachat-desktop/issues/2518, now spearheading on tauri, because tauri already has a plugin.
The option is hidden for now on electron and browser, because a disabled autostart option would only create confusion as long as tauri is not the default.

Also the option on tauri is for now only shown in release builds.

Bildschirmfoto 2025-03-08 um 20 05 26

Bildschirmfoto 2025-03-08 um 20 05 32

Bildschirmfoto 2025-03-08 um 20 29 43

### Progress:

- [x] Test it and disable the option for packaging options that don't work
    - [x] Linux package
    - [x] Linux appimage
    - ~~Linux flatpak~~ postponed
    - [x] Windows
    - [x] MacOS

### DIFF
--- a/packages/frontend/src/components/Settings/Advanced.tsx
+++ b/packages/frontend/src/components/Settings/Advanced.tsx
@@ -85,22 +85,22 @@ function SettingsAutoStart() {
   return (
     <>
       <DesktopSettingsSwitch
+        key={String(!autostartState) /* force react to rerender element, so that the switch does not animate */}
         settingsKey='autostart'
         label={tx('pref_autostart')}
         description={
           autostartState
-            ? autostartState.isRegistered
+            ? autostartState.isSupported?
+            autostartState.isRegistered
               ? tx('pref_autostart_registered')
-              : tx('pref_autostart_not_registered')
+              : tx('pref_autostart_not_registered'):
+              tx('pref_autostart_not_supported')
             : undefined // don't show description while it is loading
         }
         disabled={!autostartState?.isSupported}
         disabledValue={false}
         callback={update}
       />
-      {autostartState && !autostartState.isSupported && (
-        <Callout>{tx('pref_autostart_not_supported')}</Callout>
-      )}
     </>
   )
 }
--- a/packages/frontend/src/components/Settings/Advanced.tsx
+++ b/packages/frontend/src/components/Settings/Advanced.tsx
@@ -14,7 +14,6 @@ import { runtime } from '@deltachat-desktop/runtime-interface'
 import CoreSettingsSwitch from './CoreSettingsSwitch'
 import DesktopSettingsSwitch from './DesktopSettingsSwitch'
 import { AutostartState } from '@deltachat-desktop/shared/shared-types'
-import Callout from '../Callout'
 
 type Props = {
   settingsStore: SettingsStoreState
@@ -85,16 +84,17 @@ function SettingsAutoStart() {
   return (
     <>
       <DesktopSettingsSwitch
-        key={String(!autostartState) /* force react to rerender element, so that the switch does not animate */}
+        // force react to rerender element, so that the switch does not animate
+        key={String(!autostartState)}
         settingsKey='autostart'
         label={tx('pref_autostart')}
         description={
           autostartState
-            ? autostartState.isSupported?
-            autostartState.isRegistered
-              ? tx('pref_autostart_registered')
-              : tx('pref_autostart_not_registered'):
-              tx('pref_autostart_not_supported')
+            ? autostartState.isSupported
+              ? autostartState.isRegistered
+                ? tx('pref_autostart_registered')
+                : tx('pref_autostart_not_registered')
+              : tx('pref_autostart_not_supported')
             : undefined // don't show description while it is loading
         }
         disabled={!autostartState?.isSupported}
--- a/packages/frontend/src/components/Settings/Advanced.tsx
+++ b/packages/frontend/src/components/Settings/Advanced.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useEffect, useState, useCallback } from 'react'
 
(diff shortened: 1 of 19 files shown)


Simon-Laux:
works on windows. but I turned it off and restarted windows and then It crashed when I started it from the main menu (this needs to be investigated.

WofWca (commented):
Besides the small nit picks: indeed, for me it also crashes if I launch it after disabling autostart (no reboot required).
I tried to make a debug build with `pnpm tauri build --debug`, but then it says "Autostart is not supported on this system".

The error I get from the preview from this MR is as follows:

```
thread 'main' panicked at C:\Users\runneradmin\.cargo\git\checkouts\tauri-69fbbe4d0942e697\2d029a9\crates\tauri\src\app.rs:1078:11:
Failed to setup app: error encountered during setup hook: The system cannot find the file specified. (os error 2)
```

Workaround to prevent it from crashing: edit `config.json` in the Delta Chat data folder to set autostart to `true`.

packages/target-tauri/src-tauri/src/settings.rs
```diff
@@ -86,3 +92,24 @@ pub(crate) fn get_setting_bool_or(
...
+    use tauri_plugin_autostart::ManagerExt;
+    let store = app.store(CONFIG_FILE)?;
+    let enabled = get_setting_bool_or(store.get(AUTOSTART_KEY), AUTOSTART_DEFAULT);
+
+    let autostart_manager = app.autolaunch();
+
+    if enabled {
+        autostart_manager.enable()?;
+    } else {
+        autostart_manager.disable()?;
```
WofWca, line 112: I think I narrowed down the problem, and it's on this line. I'll try something.

packages/target-tauri/src-tauri/src/settings.rs
```diff
@@ -86,3 +92,24 @@ pub(crate) fn get_setting_bool_or(
...
+    }
```
WofWca, line 113:
```suggestion
    let enable = get_setting_bool_or(store.get(AUTOSTART_KEY), AUTOSTART_DEFAULT);

    let autostart_manager = app.autolaunch();

    let is_enabled = autostart_manager.is_enabled()
        .context("failed to check whether autostart is enabled")?;
    if enable == is_enabled {
        // If we don't return here, `autostart_manager.disable()` below
        // will return an error, at least on Windows.
        return Ok(())
    }

    if enable {
        autostart_manager.enable().context("failed to enable autostart")?;
    } else {
        autostart_manager.disable().context("failed to disable autostart")?;
    }
```

This fixes the crash for me.

packages/frontend/src/components/Settings/Advanced.tsx
```diff
@@ -53,3 +60,40 @@ export default function Advanced({ settingsStore }: Props) {
...
+  return (
+    <>
+      <DesktopSettingsSwitch
+        settingsKey='autostart'
+        label={tx('pref_autostart')}
+        description={
+          autostartState
+            ? autostartState.isRegistered
+              ? tx('pref_autostart_registered')
+              : tx('pref_autostart_not_registered')
```
WofWca, line 87: Also perhaps there is no need to display these, unless we failed to register / unregister autostart. The value of the toggle should be enough.
But this is also not too important.

packages/target-tauri/src-tauri/src/settings.rs
```diff
@@ -30,13 +35,14 @@ pub(crate) fn load_and_apply_desktop_settings_on_startup(app: &AppHandle) -> any
     apply_zoom_factor(app)?;
     // TODO: activate tray icon based on `minimizeToTray`
     apply_content_protection(app)?;
+    apply_autostart(&app)?;
```
WofWca, line 38:
```suggestion
    if let Err(err) = apply_autostart(&app)
        .context("failed to apply autostart") {
        // Not too critical, let's just log.
        error!("{err}")
    };
```

Just a suggestion, I'm not sure if it's best to handle the error here.

This will also require importing `log::error`.

packages/frontend/src/components/Settings/Advanced.tsx
Simon-Laux: > unless we failed to register / unregister autostart

too much complexity for now, we can always improve

packages/frontend/src/components/Settings/Advanced.tsx
```diff
@@ -53,3 +60,40 @@ export default function Advanced({ settingsStore }: Props) {
...
+  const [autostartState, setAutostartState] = useState<AutostartState | null>(
+    null
+  )
+  const update = useCallback(() => {
+    runtime.getAutostartState().then(setAutostartState)
+  }, [])
+
+  useEffect(() => {
+    update()
+  }, [update])
```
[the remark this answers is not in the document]
Simon-Laux: solved it with a react key

packages/frontend/src/components/Settings/Advanced.tsx
```diff
@@ -53,3 +60,40 @@ export default function Advanced({ settingsStore }: Props) {
...
+              : tx('pref_autostart_not_registered')
+            : undefined // don't show description while it is loading
+        }
+        disabled={!autostartState?.isSupported}
+        disabledValue={false}
+        callback={update}
+      />
+      {autostartState && !autostartState.isSupported && (
+        <Callout>{tx('pref_autostart_not_supported')}</Callout>
+      )}
```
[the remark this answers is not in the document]
Simon-Laux: done

WofWca (approved):
I have checked again that there is no crash on Windows.
I did not re-test though if it auto-starts, but it did in the first version of this MR,

Issue #4745, opened 2025-03-05
HTML email view broken on windows and breaking app


Simon-Laux:
opening the window freezes at the first (header)`window.add_child` stage.

This looks like a general tauri issue, we need to make a min reproducable example to report it.

Simon-Laux:
> This looks like a general tauri issue, we need to make a min reproducable example to report it.

The examples from tauri work, so likely a problem in our code. note that the help window is also blank and not closable (even before the titlebar menu branch).

WofWca:
Could you share a link for the examples?

WofWca:
I have tried commenting out the first `add_child`. It still freezes.

https://github.com/deltachat/deltachat-desktop/blob/baed35ddc30bf8ea0b59d0187f9810baaf6b588b/packages/target-tauri/src-tauri/src/html_window/mod.rs#L104

`WebBuilder::new()` docstring references this issue https://github.com/tauri-apps/wry/issues/583.

Simon-Laux:
> Could you share the link for the examples?

https://github.com/tauri-apps/tauri/tree/dev/examples/multiwebview

WofWca:
Once the MRs that reference this issue are merged, this is closed.

PR #4759, opened 2025-03-10
tauri: fix: HTML viewer and help hanging on Win

WofWca:
Partially addresses
https://github.com/deltachat/deltachat-desktop/issues/4745.

This implements the recommendadtion stated in `WebBuilder::new()` docs:
> On Windows, this function deadlocks when used
> in a synchronous command, see
> [the Webview2 issue](https://github.com/tauri-apps/wry/issues/583).
> You should use async commands when creating windows.

The windows don't hang the app anymore, but they are empty on Windows.
Opening the HTML message viewer also opens
"http://email.localhost/index.html" in the default system browser, which doesn't seem intentional.

Tokio complained about `block_on`, so I replaced it with `await`.

### DIFF
--- a/packages/target-tauri/src-tauri/src/html_window/mod.rs
+++ b/packages/target-tauri/src-tauri/src/html_window/mod.rs
@@ -4,10 +4,8 @@ use anyhow::anyhow;
 use log::{error, info, trace, warn};
 
 use tauri::{
-    async_runtime::{block_on, handle},
-    webview::WebviewBuilder,
-    LogicalPosition, LogicalSize, Manager, Url, Webview, WebviewUrl, Window, WindowBuilder,
-    WindowEvent,
+    async_runtime::block_on, webview::WebviewBuilder, LogicalPosition, LogicalSize, Manager, Url,
+    Webview, WebviewUrl, Window, WindowBuilder, WindowEvent,
 };
 use tauri_plugin_dialog::DialogExt;
 use tauri_plugin_opener::OpenerExt;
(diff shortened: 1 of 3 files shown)


packages/target-tauri/src-tauri/src/html_window/mod.rs
```diff
@@ -38,10 +38,10 @@ mod punycode;
 
 #[allow(clippy::too_many_arguments)]
 #[tauri::command]
-pub(crate) fn open_html_window(
+pub(crate) async fn open_html_window(
     app: tauri::AppHandle,
-    html_instances_state: tauri::State<HtmlEmailInstancesState>,
-    dc: tauri::State<DeltaChatAppState>,
+    html_instances_state: tauri::State<'_, HtmlEmailInstancesState>,
+    dc: tauri::State<'_, DeltaChatAppState>,
```
WofWca, line 44: Here I just did what the compiler suggested. I don't really understand what it means.

Simon-Laux: this is a lifetime

Simon-Laux:
> Opening the HTML message viewer also opens
"http://email.localhost/index.html" in the default system browser, which doesn't seem intentional.

there is an on navigation listener that does not account for how the url scheme looks on windows yet, that will be easy to fix.

Thanks for finding the issue!

WofWca:
About the help being empty: I'm working on it now. If I comment out these lines, it works.

https://github.com/deltachat/deltachat-desktop/blob/2238d9497ad65d6c0701206544775806f243d9b4/packages/target-tauri/src-tauri/src/help_window.rs#L46-L48

I'll try to fix.

packages/target-tauri/src-tauri/src/html_window/mod.rs
WofWca: I know that those are lifetimes, but I don't get lifetimes XD

PR #4770, opened 2025-03-12
style: fix app picker clear search button hidden

WofWca:
The original navbar buttons are not affected by this, I tested.

#skip-changelog


nicodh (approved):
Although I don't like that we use the variable --navBarText in several non navBar contexts/components, that probably should be addressed in another theme refactoring PR

Issue #4639, opened 2025-02-13
On PinePhone, scanning a qr code from an image will hang for a really long time like the app crashed


ell1e:
On PinePhone, scanning a qr code from an image will hang for a really long time like the app crashed. The problem seems to be that whatever the scanner does is really slow on this device, and there is zero indicator what's going on and the window also gets stuck when switching between portrait and landscape. Ideally, the window should remain responsive and show some scanning indicator that includes some sort of progress bar.

- Operating System (Linux/Mac/Windows/iOS/Android): Linux ARM64
- Delta Chat Version: desktop 1.50.1 (flathub) core 1.152.1
- Expected behavior: Progress indicator is shown and user understands what's going on
- Actual behavior: Window plays dead
- Steps to reproduce the problem: 1. Use a very slow device. 2. Get a QR code image, preferably blurry and challenging. 3. Scan by opening the image from a file, 5. After file choice was confirmed in the file open dialog, watch what happens right after
- Screenshots:
- Logs:

WofWca:
@r10s Could you please check if this is the recent "QR code issue" or something different?

PR #4777, opened 2025-03-13
perf: improve QR scanner performance

WofWca:
~~This (hopefully) closes (update: wrong issue) https://github.com/deltachat/deltachat-desktop/issues/4639.~~

See individual commits for details.

I tested with dev tools with x6 CPU slowdown. It works fine.

I also tried to play around with constraints, but adding e.g. `width` causes an error for me.

We probably should switch to a library that would do all the stuff in a worker, but these are small changes that I got so far.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 
 ### Fixed
 - fix some webxdc apps showing the "Close app?" prompt unintentionally #4737
+- improve QR scanner performance
 - webxdc: fix menu bar hiding when pressing Escape #4753
 - tauri: fix blobs and webxdc-icon scheme under windows #4705
 - tauri: fix app picker not working for some apps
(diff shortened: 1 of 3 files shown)


nicodh (approved):
In the referred bug it says in "how to reproduce": _Scan by opening the image from a file_

This PR addresses the scanning process via video stream from camera.

 The problem, that the UI freezes while the processing of an uploaded image is in progress is still there.

When loading an image that contains no QR code at all, depending on the image size, it takes quite some time (even without throttling CPU)  and it is not possible to cancel the process via "Cancel"

I made a separate PR for that: #4781

But the changes of this PR are useful so approved

WofWca:
> In the referred bug it says in "how to reproduce": Scan by opening the image from a file

Ohhh. Yeah, you're right, sorry. I'll update the commit message.

PR #4780, opened 2025-03-14
Update core to 1.157.2

nicodh:


WofWca (approved):
Looks like that this also upgrades esbuild to 0.25.1.
Otherwise it's a regular core upgrade.

Simon-Laux (approved):

CHANGELOG.md
```diff
@@ -12,6 +12,12 @@
 ### Changed
 - tauri: replace `tauri-plugin-shell` with `tauri-plugin-opener` #4699
 - new button styles #4741
+- Update `@deltachat/stdio-rpc-server` and `deltachat/jsonrpc-client` to `1.157.1`
+  - Ignore encryption preferences
```
Simon-Laux, line 16: this point might need clarification

nicodh: See https://github.com/chatmail/core/pull/6612

nicodh: @link2xt  Is that a "breaking change" in some scenarios?

nicodh: > We should talk to them about our plans for PGP-contacts and multitransport anyway. For most users it does not matter.

Does this means, we should wait with the core update and the next release with core 1.157.x ?

nicodh: If it's only on Android so desktop is not a use case for them?

Simon-Laux: @nicodh please clarify it in a way that people can not think that this means less encryption.

r10s:
ftr, there is core 1.157.2  now, containing the critical IMAP COMPRESS fix leading eg. to https://github.com/chatmail/core/issues/6656

CHANGELOG.md
Simon-Laux: maybe just remove this misleading line when merging #4785, the description there should be enough

Simon-Laux (approved):

PR #4781, opened 2025-03-15
Use worker for processing qr images

### LINKED ISSUE #4639: On PinePhone, scanning a qr code from an image will hang for a really long time like the app crashed

nicodh:
Resolves #4639

- show the spinner while processing imported qr code images or clipboard
- enable to cancel the processing if it takes too long
- call QrCode.handlePasteFromClipboard from parent component
- use second worker for clipBoard processing to avoid race conditions

### DIFF
--- a/packages/frontend/src/components/QrReader/index.tsx
+++ b/packages/frontend/src/components/QrReader/index.tsx
@@ -15,6 +15,8 @@ import useTranslationFunction from '../../hooks/useTranslationFunction'
 import { ContextMenuContext } from '../../contexts/ContextMenuContext'
 import { runtime } from '@deltachat-desktop/runtime-interface'
 
+import useAlertDialog from '../../hooks/dialog/useAlertDialog'
+
 import { fileToBase64, base64ToImageData } from './helper'
 
 // @ts-ignore:next-line: We're importing a worker here with the help of the
@@ -59,6 +61,7 @@ export const QrReader = forwardRef<QrCodeScanRef, Props>(
   ({ onError, onScanSuccess }: Props, ref) => {
     const tx = useTranslationFunction()
     const { openContextMenu } = useContext(ContextMenuContext)
+    const openAlertDialog = useAlertDialog()
 
     const videoRef = useRef<HTMLVideoElement>(null)
     const canvasRef = useRef<OffscreenCanvas>(new OffscreenCanvas(640, 480))
@@ -158,6 +161,13 @@ export const QrReader = forwardRef<QrCodeScanRef, Props>(
 
     const handlePasteFromClipboard = useCallback(async () => {
       try {
+        if (processingFile) {
+          openAlertDialog({
+            message:
+              'Already processing a file. Please wait for the result or cancel it',
+          })
+          return
+        }
         // Try interpreting the clipboard data as an image
         let base64: string | null = null
         setProcessingFile(true)
@@ -170,6 +180,7 @@ export const QrReader = forwardRef<QrCodeScanRef, Props>(
           const imageData = await base64ToImageData(base64)
           processQrCodeWithWorker(imageData, onScanSuccess, handleError)
         } else {
+          setProcessingFile(false)
           // .. otherwise return non-image data from clipboard directly
           const data = await runtime.readClipboardText()
           if (!data) {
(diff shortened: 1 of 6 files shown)


packages/frontend/src/components/QrReader/index.tsx
```diff
@@ -13,360 +15,475 @@ import useTranslationFunction from '../../hooks/useTranslationFunction'
...
+    useEffect(() => {
+      worker = new QrWorker() as Worker
+      workerClipBoard = new QrWorker() as Worker
+      return () => {
+        // terminate worker if component is unmounted,
+        // otherwise we might get a delayed error message
+        // if scan fails, even after closing the QR dialog
+        worker.terminate()
+        workerClipBoard.terminate()
       }
```
WofWca, line 82: If we go this route, then we should have the workers scoped to the component instance, and not globally.

WofWca (commented):
This should be better.

Though, to note, `fileToBase64` is still performed on the main thread.

packages/frontend/src/components/QrReader/index.tsx
```diff
@@ -13,360 +15,475 @@ import useTranslationFunction from '../../hooks/useTranslationFunction'
...
+          if (!data) {
+            throw new Error('no data in clipboard')
+          }
+          // trim whitespaces because user might copy them by accident when sending over other messengers
+          // see https://github.com/deltachat/deltachat-desktop/issues/4161#issuecomment-2390428338
+          onScanSuccess(data.trim())
         }
-      } catch (error: any) {
+      } catch (error) {
+        setProcessingFile(false)
```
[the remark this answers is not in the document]
nicodh: > setProcessingFile usage seems more convoluted than it has to be. Is there no simpler way?

setProcessingFile is set to true in handleFileInputChange or handlePasteFromClipboard to show the user that "something" is processing. It is either reset to false in processQrCodeWithWorker if that is reached, otherwise set to false in the "else" condition


> there is a bug where after clicking "paste" with plain text in your clipboard, processingFile never gets reset to false

is fixed

nicodh:
Code check fails with:

`/home/runner/work/deltachat-desktop/deltachat-desktop/packages/frontend/src/components/dialogs/QrCode.tsx
Error:   22:10  error  'runtime' is defined but never used. Allowed unused vars must match /^_/u  @typescript-eslint/no-unused-vars`

which is wrong since runtime IS used in line  85

packages/frontend/src/components/QrReader/index.tsx
```diff
@@ -34,352 +41,482 @@ type Props = {
...
+    const inputRef = useRef<HTMLInputElement>(null)
+
+    const [ready, setReady] = useState(false)
+    const [cameraAccessError, setCameraAccessError] = useState(false)
+    const [videoDevices, setVideoDevices] = useState<MediaDeviceInfo[]>([])
+    const [deviceId, setDeviceId] = useState<string | undefined>(undefined)
+    const [processingFile, setProcessingFile] = useState(false)
+
+    const worker = useRef<Worker>(new QrWorker() as Worker)
+    const workerClipBoard = useRef<Worker>(new QrWorker() as Worker)
```
WofWca, line 73: This unconditionally instantiates a new worker on every render, resulting in a memory leak.

Maybe see https://react.dev/reference/react/useRef#avoiding-recreating-the-ref-contents on how to resolve.

WofWca (changes requested):
Aside from the comment, this seems to work OK now.

WofWca:
And about the CI error: it's weird, not reproducible locally. Let's ignore, and resolve separately if it reappears on main.

WofWca (approved):
I have confirmed that the memory leak is gone.
And this seems to work alright.

Issue #4782, opened 2025-03-15
Remove "Prefer e2e encryption" from settings


nicodh:
With https://github.com/chatmail/core/pull/6612 encryption is always preferred if it is available. So we should not show these settings any more

nicodh:
Done in #4785

Issue #4312, opened 2024-11-03
Add UI to reset encryption


link2xt:
There is a core PR to adding JSON-RPC API to reset encryption:
https://github.com/deltachat/deltachat-core-rust/pull/6170

It is needed to close https://github.com/deltachat/deltachat-core-rust/issues/4465

"Reset Encryption" could be in the contact profile menu after "Encryption Info" or inside the Encryption Info dialog:
![reset_encryption](https://github.com/user-attachments/assets/938aa7ba-e114-4103-948d-53410f05f6a6)

Should probably be red as it is not a reversible action.

For chatmail profiles can be completely hidden as resetting encryption there in most cases makes it impossible to send messages. Could also be hidden if developer options are not enabled.

r10s:
let's put it as a button inside "encryption info", it is pretty advanced and should not be present in the primary ui.

should be a normal, unexcited button below the fingerprints. not sure if it needs to be "red", it is reversible when a new encrypted message comes in, this happens also automatically all the time without the user being asked

iequidoo:
Technically it's easy to make it revertible, but then an extra translation is needed at least. I think if it's not a dev option, it should be red, otherwise i'm almost sure there will be complains why it's not possible to set encription back. But maybe just make it a dev option, do they require translation?

Simon-Laux:
"reset encryption" is not a good name, took me a while to figure out what is meant by that. If this is user visible it should have a clearer name and maybe also a link to an FAQ entry next to it that explains what it means

PR #4785, opened 2025-03-15
Remove e2e encryption preferences from settings

nicodh:
solves #4782


Simon-Laux (approved):

Simon-Laux:
- does this mean we now also need https://github.com/deltachat/deltachat-desktop/issues/4312?

PR #4787, opened 2025-03-15
tauri: menu for html email window

Simon-Laux:
- **ignore more logs for iroh**
- **conditionaly compile `get_menu_for_window` method that it does not produce an unused warning and always set the menu on mac when creating a window**
- **html window menu**


WofWca (approved):
The menu works (except for fullscreen on Windows, which, I assume, is a Tauri issue).

The code changes are a little hard to review since the changes are mixed, but they look sane.

Issue #4667, opened 2025-02-17
Show additional dialog with hints about invite link


nicodh:
When copying the invite link there should be an additional dialog explaining the invite link url and its handling (like in Android).

<image src="https://github.com/user-attachments/assets/1f53efcd-d952-405e-bd5e-9e3206f42415" width="400" />

See https://github.com/deltachat/deltachat-android/pull/3367#issuecomment-2420551758

Followup to #4530

WofWca:
Closed by https://github.com/deltachat/deltachat-desktop/issues/4667

PR #4789, opened 2025-03-15
fix: Show additional dialog with hints about invite link #4667

friedmag:
solves #4667

### DIFF
--- a/packages/frontend/src/components/dialogs/QrCode.tsx
+++ b/packages/frontend/src/components/dialogs/QrCode.tsx
@@ -27,8 +27,10 @@ import useProcessQr from '../../hooks/useProcessQr'
 import useTranslationFunction from '../../hooks/useTranslationFunction'
 import { selectedAccountId } from '../../ScreenController'
 
+import useDialog from '../../hooks/dialog/useDialog'
 import type { DialogProps } from '../../contexts/DialogContext'
 import useAlertDialog from '../../hooks/dialog/useAlertDialog'
+import QrCodeCopyConfirmationDialog from './QrCodeCopyConfirmationDialog'
 
 const log = getLogger('renderer/dialogs/QrCode')
 
@@ -105,14 +107,21 @@ export function QrCodeShowQrInner({
 }) {
   const { userFeedback } = useContext(ScreenContext)
   const tx = useTranslationFunction()
+  const { openDialog } = useDialog()
 
   const onCopy = () => {
-    runtime.writeClipboardText(qrCode).then(_ => {
-      userFeedback({
-        type: 'success',
-        text: tx('copy_qr_data_success'),
-      })
-      onClose()
+    // Pop up confirmation dialog when clicked instead of copying the link directly
+    openDialog(QrCodeCopyConfirmationDialog, {
+      message: tx('share_invite_link_explain'),
+      content: qrCode,
+      copyCb: () => {
+        userFeedback({
+          type: 'success',
+          text: tx('copy_qr_data_success'),
+        })
+        onClose()
+      },
+      // no cancelCb; skip closing the window, maybe the user wants to use the QR code after all
     })
   }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - tauri: replace `tauri-plugin-shell` with `tauri-plugin-opener` #4699
 - new button styles #4741
 - remove end-to-end encryption preferences from settings (e2e encryption is always preferred when available) #4782
+- add dialog with hints about invite link #4667
 - Update `@deltachat/stdio-rpc-server` and `deltachat/jsonrpc-client` to `1.157.2`
   - Delete messages on IMAP when deleting chat
   - Allow doubled avatar resolution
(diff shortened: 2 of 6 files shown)


packages/frontend/src/components/dialogs/QrCode.tsx
```diff
@@ -105,14 +107,22 @@ export function QrCodeShowQrInner({
...
 
   const onCopy = () => {
-    runtime.writeClipboardText(qrCode).then(_ => {
-      userFeedback({
-        type: 'success',
-        text: tx('copy_qr_data_success'),
-      })
-      onClose()
+    // https://github.com/deltachat/deltachat-desktop/issues/4667
+    // Pop up confirmation dialog when clicked instead of copying the link directly
```
Simon-Laux, line 114: I'd say issue does not need to be mentioned unless you mention an issue that still exists or need the reference to refer to a specific bug, like "this is why we did it this way". In this case it is just an enchantment in my opinion, so not necessary to mention.

packages/frontend/src/components/dialogs/QrCode.tsx
```diff
@@ -105,14 +107,22 @@ export function QrCodeShowQrInner({
...
+    openDialog(QrCodeCopyConfirmationDialog, {
```
Simon-Laux, line 115: it is better to use the existing confirmation dialog, unless you have a different design for this specific confirmation dialog.

you can change the labels for the buttons of the existing confirmation dialog.

Though the image in https://github.com/deltachat/deltachat-desktop/issues/4667 shows a custom dialog that shows a preview of the link, so in this case it should be custom I guess?

Simon-Laux (commented):
Thanks for your contribution.
Code LGTM, I did not test it yet. compared to the image in the issue the preview of the link is still missing, but besides that it looks fine to me

packages/frontend/src/components/dialogs/QrCode.tsx
friedmag: Sure, will fix.

packages/frontend/src/components/dialogs/QrCode.tsx
friedmag: It didn't look quite right with the existing dialog I was looking at, at least... but I can add the URL preview to justify it better, sure.

BTW, I'm having issues running the e2e - I get `[ERROR]: Data dir does not exist, make sure you follow the steps in the Readme file` but I'm not seeing anything about it in the `E2E-TESTING.md`. Any suggestions?

nicodh: Do you have a "data" dir in /packages/target-browser/ ?
You can test it by runnng pnpm start:browser and open localhost:3000

nicodh: The e2e tests start the browser target so you have to follow the steps described in packages/target-browser/Readme.md

btw. there is a missing step now in the e2e tests, since they use the copy link dialog

friedmag: Ah, I didn't notice the other Readme.md in there. Now that they're running, I'll get them passing again.

friedmag: I take it back. I thought they were running, but just got further - it's failing right after loading localhost:3000, not finding `.main-container`. It seems like something is still wrong with the data - it's expecting to log straight into a profile, but I'm just getting a `Create New Profile/I Already Have a Profile`?

I must have missed another step somewhere. It seems to have generated some account files, but is not using them. Ideas?

nicodh: So first point is: can you run desktop in the browser and create an account there?
If so, everything should be prepared to run the e2e tests.
Are you on Linux/Mac or Windows?

nicodh: If you look at the failing e2e tests you can see that 4 profiles were created. The failing tests is the "start chat with user " test, which is expected since to start a chat one user copies the invite linkt to clipboard https://github.com/deltachat/deltachat-desktop/blob/1c2cbdd5d95f72dd20a63278814521640fdc444d/packages/e2e-tests/tests/basic-functionality.spec.ts#L90

It's there were the new dialog is shown so an additional click is needed.

nicodh: This is what is needed to fix the e2e tests:
https://github.com/deltachat/deltachat-desktop/commit/e412f1e2b4964db856db233b4ccd7d612ff54b66

friedmag: Yep, I was just having trouble running the tests locally (Windows), but I finally figured it out. I was not setting `NODE_ENV=test` which is needed to bypass the `/authenticate`. I noticed it in the `env.example` so tried it, and after that of course it was obvious that was in the GitHub Action. Oops. Maybe a mention in the E2E doc?

I should be able to make the preview update & all the tests pass later tonight.

friedmag: Preview is added, and tests pass

packages/frontend/src/components/dialogs/QrCode.tsx
friedmag: Removed

nicodh (changes requested):
Looks good now!

If you would add a padding to .copy-content-preview like in the screenshot in #4699 and attributes styling={'secondary'} to the buttons it would be ready to merge!

friedmag:
> Looks good now!
> 
> If you would add a padding to .copy-content-preview like in the screenshot in #4699 and attributes styling={'secondary'} to the buttons it would be ready to merge!

Done

nicodh (approved):
Thanks for contributing and fast response time!

PR #4795, opened 2025-03-16
tauri: hardening: disable all commands by default

WofWca:
[The Tauri's API example app](https://github.com/tauri-apps/tauri/blob/cde0ff7798a46712f69b34aab952209f45500fe9/examples/api/src-tauri/build.rs#L9-L21)
was used as a reference for the implementation of this commit.

This https://github.com/tauri-apps/tauri/pull/9008 says:
> By default all app commands are still allowed,
> so if you do not use `tauri_build::Attributes::app_manifest`
> you're not affected by this change.

Tauri docs don't seem to say much about how to disable this
"allow all commands by default" thing.

I have tested this change and verified that commands
are not available in the HTML email viewer contents webview,
and that only the required ones are available in its header webview.

Note that the upcoming webxdc MR
(https://github.com/deltachat/deltachat-desktop/pull/4740)
will need to be adjusted
to specify the allowed commands for the webxdc app windows.


packages/target-tauri/src-tauri/build.rs
```diff
@@ -17,7 +17,43 @@ fn main() {
     println!("cargo:rustc-env=BUILD_TIME_STAMP={build_time_stamp}");
     println!("cargo:rustc-env=BUILD_INFO_GIT={}", get_git_ref());
 
-    tauri_build::build()
+    tauri_build::try_build(tauri_build::Attributes::default().app_manifest(
+        tauri_build::AppManifest::new().commands(&[
```
Simon-Laux, line 21: a comment about why and how this works would be nice here
```suggestion
        // This adds capabilities for the commands so we can specify which window has access to what in the tauri configuration.
        // see also https://docs.rs/tauri-build/latest/tauri_build/struct.AppManifest.html
        // and https://v2.tauri.app/security/permissions/
        tauri_build::AppManifest::new().commands(&[
```
But you can word this better than I can...

Simon-Laux (approved):

packages/target-tauri/src-tauri/build.rs
WofWca: You're right. I thought of adding this, but forgot.

WofWca:
So, shall I merge? Or do you want the finish up https://github.com/deltachat/deltachat-desktop/pull/4740 first?

Simon-Laux:
you can go ahead and merge this pr, I need to rebase the webxdc pr anyway and adjust it to include translations

PR #4797, opened 2025-03-16
add option to reset encryption state for contact in encryption info dialog

### LINKED ISSUE #4312: Add UI to reset encryption

Simon-Laux:
- add option to reset encryption state for contact in encryption info dialog
- update styling of footer button to match new style

closes #4312

Bildschirmfoto 2025-03-16 um 20 41 33

Bildschirmfoto 2025-03-16 um 20 41 39


r10s:
is this part of multi-transport and its change to bind identity to keys instead of adresses?

otherwise, not sure how useful it is to add additional constraints just now. it might be better to reconsider when multi-transport is done

Simon-Laux:
> We are probably going to remove chatmail/non-chatmail distincution for multi-transport

We'll probably have some successor that we can use to tell if a user should be able to add contacts by email address?

Anyways since we are removing the encryption preference I thought this might be needed, because the only use case I could imagine for the global encryption preference was to send unencrypted messages.

nicodh (approved):

Issue #4695, opened 2025-02-21
allow editing messages


r10s:
after https://github.com/deltachat/deltachat-desktop/issues/4694 is done:

- add jsonrpc for `dc_send_edit_request()`

- add the menu entry "Edit" (string `global_menu_edit_desktop`) to the context menu of outgoing messages that are _not_ info/videochat  _and_ have a text _and_ does not have html _and_ where the chat allows sending in general (check `chat_can_send()`, there were discussions about having a core-api, but were discarded for $reasons) 

- when "Edit" is tapped:
    - create a draft with title set to "Edit Message" (string `edit_message`) and subtitle the text to edit (use the same layout as for quotes)
    - right of title/subtitles, as usual, show a "X" that cancels editing and reverts to normal sending mode
    - hide "attachment" button, leave "send" button as is
    - it is okay to discard existing draft, if that is easier than keeping them
    - in case the whole text is removed, disable the "send" button as usual

- when "Send" is tapped from "Edit":
    - call `dc_send_edit_request()` with the new text
    - it is fine to call the function even if the text is not changed (will do nothing then)
    - go back to normal "enter message to send" mode

this UI is known from many other messengers

refer to https://github.com/deltachat/deltachat-ios/pull/2611 and https://github.com/deltachat/deltachat-android/pull/3601 for example code and **screenshots**

r10s:
`global_menu_edit_desktop` which is a simple "Edit" in all languages (it is used like that also elsewhere) is fine. 

it is used in a context menu, the context is clear.  so it is better to add less noise, this is also matches internal and external consistency.

WofWca:
I can speak for Russian, the "Edit" menu option is often "Правка", which is a little vague. Not unbearably vague, but still, why not `edit_message`?

r10s:
on topic: in contrast to https://github.com/deltachat/deltachat-desktop/issues/4694, this issue should not go to the next release. 

to avoid at least some confusion, before actually allow editing, there should be a release out that can show edits accordingly.

WofWca:
And what about the composer, displaying the message that is being edited? Do we also use `global_menu_edit_desktop` there?

WofWca:
Hey, what about forwarded messages? Do we want them to be editable? Because core allows it, I just tested.

r10s:
> Hey, what about forwarded messages? Do we want them to be editable? Because core allows it, I just tested.

[...]

PR #4799, opened 2025-03-17
Delete msgs for all

nicodh:
Needs core PR https://github.com/chatmail/core/pull/6672 to be released

### DIFF
--- a/packages/frontend/src/components/Dialog/styles.module.scss
+++ b/packages/frontend/src/components/Dialog/styles.module.scss
@@ -154,6 +154,7 @@ $paddingVertical: 20px;
 
 .footerActionButton {
   min-width: 120px;
+  white-space: nowrap;
   padding: 8px 12px !important;
   & + & {
     margin-left: 20px;
--- a/packages/frontend/src/components/dialogs/ConfirmDeleteMessage.tsx
+++ b/packages/frontend/src/components/dialogs/ConfirmDeleteMessage.tsx
@@ -1,0 +1,75 @@
+import React from 'react'
+
+import Dialog, {
+  DialogBody,
+  DialogContent,
+  DialogFooter,
+  FooterActions,
+} from '../Dialog'
+import FooterActionButton from '../Dialog/FooterActionButton'
+import useTranslationFunction from '../../hooks/useTranslationFunction'
+
+import type { DialogProps } from '../../contexts/DialogContext'
+import { BackendRemote, Type } from '../../backend-com'
+import { C } from '@deltachat/jsonrpc-client'
+
+export type Props = {
+  accountId: number
+  msg: Type.Message
+  canSend: boolean
+} & DialogProps
+
+export default function ConfirmDeleteMessageDialog(props: Props) {
+  const tx = useTranslationFunction()
+
+  const { accountId, msg, onClose } = props
+
+  const deleteForAllPossible =
+    msg.sender.id === C.DC_CONTACT_ID_SELF && props.canSend
+
+  const deleteMessage = (deleteForEveryone: boolean) => {
+    if (deleteForEveryone) {
+      BackendRemote.rpc.deleteMessagesForAll(accountId, [msg.id])
+    } else {
+      BackendRemote.rpc.deleteMessages(accountId, [msg.id])
+    }
+    onClose()
+  }
+
+  return (
+    <Dialog onClose={onClose}>
+      <DialogBody>
+        <DialogContent>
+          <p className='whitespace'>{tx('ask_delete_message')}</p>
+        </DialogContent>
+      </DialogBody>
+      <DialogFooter>
+        <FooterActions>
+          <FooterActionButton
+            styling='secondary'
+            onClick={() => onClose()}
+            data-testid='cancel'
+          >
+            {tx('cancel')}
+          </FooterActionButton>
+          <FooterActionButton
+            styling={'danger'}
+            onClick={() => deleteMessage(false)}
+            data-testid='delete_for_me'
+          >
+            {tx('delete_for_me')}
+          </FooterActionButton>
+          {deleteForAllPossible && (
+            <FooterActionButton
+              styling={'danger'}
+              onClick={() => deleteMessage(true)}
+              data-testid='delete_for_everyone'
+            >
+              {tx('delete_for_everyone')}
+            </FooterActionButton>
+          )}
+        </FooterActions>
+      </DialogFooter>
+    </Dialog>
+  )
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - tauri: implement runtime.copyFileToInternalTmpDir and allow opening files in temp dir #4778
 - tauri: implement backend translation fn and use it in tauri files as needed #4790
 - add withdraw qr code context menu option #4798
+- delete message for all chat members #4716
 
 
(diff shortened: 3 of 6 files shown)


nicodh:
@r10s we don't ask for "deleteForAll" when clearing all messages in a chat. Is that different on Android/iOS?

packages/frontend/src/components/dialogs/ConfirmDeleteMessage.tsx
```diff
@@ -0,0 +1,75 @@
...
+  msg: Type.Message
+  canSend: boolean
+} & DialogProps
+
+export default function ConfirmDeleteMessageDialog(props: Props) {
+  const tx = useTranslationFunction()
+
+  const { accountId, msg, onClose } = props
+
+  const deleteForAllPossible =
```
WofWca, line 27: I noticed that this is also available in "saved messages". I think it doesn't make sense there.

Android checks for this https://github.com/deltachat/deltachat-android/pull/3633/files#diff-6046f8bf3371c6d149fcd32fc2b3ac0e1536a66e34b13e0096208086136529a5R360

ralphtheninja (approved):
Reviewed by testing and reading code

r10s:
> @r10s we don't ask for "deleteForAll" when clearing all messages in a chat. Is that different on Android/iOS?

"Delete for all" is available only for one's own messages. when clearing a chat, usually, others messages are cleared as well.

moreover, "Delete for all" is more a decision done explicitly for a message, not for batches

packages/frontend/src/components/dialogs/ConfirmDeleteMessage.tsx
```diff
@@ -0,0 +1,75 @@
...
+    msg.sender.id === C.DC_CONTACT_ID_SELF && props.canSend
+
+  const deleteMessage = (deleteForEveryone: boolean) => {
+    if (deleteForEveryone) {
+      BackendRemote.rpc.deleteMessagesForAll(accountId, [msg.id])
```
WofWca, line 32: This seems like a place where it is important to report an error should one happen.

Though I know that the presence of the error is otherwise visible to the user via the fact that the message didn't disappear from the chat.

packages/frontend/src/components/Dialog/styles.module.scss
```diff
@@ -154,6 +154,7 @@ $paddingVertical: 20px;
 
 .footerActionButton {
   min-width: 120px;
+  white-space: nowrap;
```
WofWca, line 157: I know we don't support narrow screens really well, and that it's the case for most other dialogs, but it's an important feature, so maybe we should make the dialog work on narrow screens?

image

packages/frontend/src/components/dialogs/ConfirmDeleteMessage.tsx
WofWca, line 27: Also, are we supposed to be able to delete info messages for all?

Android doesn't allow this: https://github.com/deltachat/deltachat-android/pull/3633/files#diff-200a4a1e945e1f2cf60ed07d8b3cf56c974ced107f47cf4ff60e0ebe993abc4cR69

WofWca (commented):
I tested this, though without the new core API. On a surface it works alright.

The issue says

> similar to https://github.com/deltachat/deltachat-desktop/issues/4695, this should not go to the release from these days, but to the next release.

Which release is that gonna be? Do we want this merged now?

packages/frontend/src/components/Dialog/styles.module.scss
WofWca, line 157: Another thing: the dialog does not look good on some languages

image

r10s: can't we have the buttons have a minimal width and otherwise fit the text? and if the buttons do not fit into one line, let them flow in multiple lines?

nicodh: That's a different issue imo, since to achieve this we have to define a different dialog width for smaller screens. Currently it's 500px if you don't define another width.

nicodh: Can be solved with media queries

nicodh: With a higher minimal-width all buttons get that width and then even the buttons with english text don't fit anymore. We would need a larger dialog width then depending on the text length

So as discussed in the issue this might be easier to solve in a column oriented button layout.
While shorter texts still fit in the buttons with whitespace: nowrap, longer texts don't.


And the problem mentioned in the comment above for small screens would also be solved.

packages/frontend/src/components/dialogs/ConfirmDeleteMessage.tsx
nicodh: I will filter by isInfo

nicodh:
> Which release is that gonna be? Do we want this merged now?

I think it makes sense to have a "soon" release in the next days without this feature and after some more testing (and e2e tests added) we can have a minor release some days later

packages/frontend/src/components/dialogs/ConfirmDeleteMessage.tsx
nicodh: Added filter in with 585369ff27324

packages/frontend/src/components/dialogs/ConfirmDeleteMessage.tsx
nicodh: I added a userFeedback for errors here.

@r10s 
I think it would make sense to add at least a confirmation for "delete for all" but then we need a new translation for that

packages/frontend/src/components/Dialog/styles.module.scss
Simon-Laux: another option would be to have a sub context menu on delete, like signal has on mobile or better visible on telegram. that way the message in the confirmation can also be less confusing (currently delete for me could also be understood as only delete on one device maybe?)

r10s:
yes, this issue should not block a release fixing the IMAP COMPRESS bug. but it should come asap still :)

packages/frontend/src/components/Dialog/styles.module.scss
WofWca: This should be resolved by https://github.com/deltachat/deltachat-desktop/pull/4806.

WofWca: Same, this should be resolved by https://github.com/deltachat/deltachat-desktop/pull/4806.

packages/frontend/src/components/dialogs/ConfirmDeleteMessage.tsx
```diff
@@ -0,0 +1,94 @@
...
+            data-testid='cancel'
+          >
+            {tx('cancel')}
+          </FooterActionButton>
+          <FooterActionButton
+            styling={'danger'}
+            onClick={() => deleteMessage(false)}
+            data-testid='delete_for_me'
+          >
+            {tx('delete_for_me')}
```
WofWca, line 79: Same as in https://github.com/deltachat/deltachat-android/pull/3633/files#diff-6046f8bf3371c6d149fcd32fc2b3ac0e1536a66e34b13e0096208086136529a5R348

```suggestion
            {chat.isSelfTalk ? tx('delete') : tx('delete_for_me')}
```

PR #4802, opened 2025-03-17
no second device experiments

r10s:
add-second-device is arguably no longer experimental, we're pointing to that options officially and often.

#skip-changelog as rewordings usually do not make it to the changelog (not sure if that fits on desktop, someone, please adapt to real policies :)

counterpart of https://github.com/deltachat/deltachat-android/pull/3685 and https://github.com/deltachat/deltachat-ios/pull/2646


nicodh (approved):

WofWca (approved):
I don't think we need a changelog entry. Code looks good.

PR #4805, opened 2025-03-18
refactor: simplify QR worker code

WofWca:
#skip-changelog


nicodh:
This is just for (maybe) better readability?

WofWca:
Well I guess you could put it that way.

It's better typed IMO. You don't have to check `.current?.` every time. You know that it's the same worker.

nicodh (approved):
I agree that readability is better in the functions but not on initialization :-)
But I'm fine with the refactoring

WofWca:
IDK, the initialization didn't change much IMO, I only added two `const`s.

PR #4806, opened 2025-03-18
fix: adapt dialogs to small screens

nicodh:
As a first start:
- set dialog width to 100vw if width < 500px
- set footer buttons flex-direction to column if width < 500px
- adapt button margins to vertical layout

There is probably space for more improvements, but for now we should focus on any unwanted side effects and improve in separate PRs.

Example:

Invite QR code with width 300px

Before:

image

After:
qr-code-screen-width-300-after

Second Device 350px

Before (Cancel not visible!)

image

After

second-device-screen-width 350-after


packages/frontend/src/components/Dialog/styles.module.scss
```diff
@@ -4,6 +4,10 @@ $paddingVertical: 20px;
 
 .dialog {
   padding: 0;
+  @media (width < 500px){
+    width: 100vw !important;
+    min-width: 100vw !important;
```
WofWca, line 9:
```suggestion
    max-width: 100vw !important;
    min-width: 0 !important;
```

This makes more sense semantically, and also doesn't make the dialogs that are narrower than 500px initially stretch unnecessarily.

WofWca, line 9: `calc(100vw - 1rem)`

image

WofWca (approved):
So much better!

I tested it with a bunch (but not all) of dialogs. They look alright.

And this shouldn't be breaking, since the majority of the added styles only apply to narrower screens (which we

Simon-Laux:
would be nice it would take the full screen without a border (for the big dialogs, for the small ones it makes sense to leave some space that it still looks like a dialog.

PR #4807, opened 2025-03-18
chore: fix false eslint error

nicodh:
#skip-changelog


WofWca (changes requested):
Actually the warning is not a false-positive anymore on current `main`, so the line actually needs to get removed.

I don't know how CI got it. Maybe it rebased the MR onto main or something?

WofWca:
> Maybe it rebased the MR onto main or something?

Indeed that's what happened. See https://github.com/deltachat/deltachat-desktop/actions/runs/13902349138/job/38903977762 :

> HEAD is now at 899166f8 Merge 6a7af56a556dc4d52b039ba1c3d1416701013727 into 479084e6b9240a197da13c0a6fa281030e4e1537

Issue #4763, opened 2025-03-11
Trying to send empty text should show a warning/error


nicodh:
When sending an empty message (no text or just empty spaces) "nothing" happens. The console shows an error but the user sees no hint.

An error message could be shown "Please enter a message"

Alternatively we could hide the send button (like Android does it)

nicodh:
Closed with #4817

PR #4817, opened 2025-03-18
Hide send button if nothing to send

nicodh:
This change provides a solution for #4763  with only minimal code changes by just hiding the send button if there is nothing to send.
~~Only downside so far: it does not work when a message is edited, but we have a warning there if the send button is clicked without a message text~~

~~Maybe it would be possible to adapt the editMessageInput to provide a function to detect if the textarea of an edited message is empty?~~

#skip-changelog

### DIFF
--- a/packages/frontend/scss/composer/_composer.scss
+++ b/packages/frontend/scss/composer/_composer.scss
@@ -79,7 +79,8 @@
 
     .emoji-button {
       height: 40px;
-      padding: 0px 8px 0px 3px;
+      padding: 0px 3px;
+      margin-inline-end: 5px;
       &:focus:not(:focus-visible) {
         outline: none;
       }
--- a/packages/frontend/scss/composer/_composer.scss
+++ b/packages/frontend/scss/composer/_composer.scss
@@ -79,7 +79,7 @@
 
     .emoji-button {
       height: 40px;
-      padding: 0px 3px;
+      padding: 0px 8px 0px 3px;
       &:focus:not(:focus-visible) {
         outline: none;
       }
--- a/packages/frontend/src/components/composer/Composer.tsx
+++ b/packages/frontend/src/components/composer/Composer.tsx
@@ -91,7 +91,6 @@ const Composer = forwardRef<
   const chatId = selectedChat.id
   const [showEmojiPicker, setShowEmojiPicker] = useState(false)
   const [showAppPicker, setShowAppPicker] = useState(false)
-  const [showSendButton, setShowSendButton] = useState(false)
   const [currentEditText, setCurrentEditText] = useState('')
 
   const emojiAndStickerRef = useRef<HTMLDivElement>(null)
@@ -139,13 +138,7 @@ const Composer = forwardRef<
     }
   }, [selectedChat.canSend])
 
-  useEffect(() => {
-    if (currentEditText === '' && !draftState.file) {
-      setShowSendButton(false)
-    } else {
-      setShowSendButton(true)
-    }
-  }, [draftState, currentEditText])
+  const showSendButton = currentEditText !== '' || !!draftState.file
 
   useEffect(() => {
     return onDCEvent(accountId, 'SecurejoinJoinerProgress', ({ progress }) => {
--- a/packages/frontend/src/components/composer/Composer.tsx
+++ b/packages/frontend/src/components/composer/Composer.tsx
@@ -91,7 +91,7 @@ const Composer = forwardRef<
   const chatId = selectedChat.id
   const [showEmojiPicker, setShowEmojiPicker] = useState(false)
   const [showAppPicker, setShowAppPicker] = useState(false)
-  const [canSendMessage, setCanSendMessage] = useState(false)
+  const [showSendButton, setShowSendButton] = useState(false)
 
   const emojiAndStickerRef = useRef<HTMLDivElement>(null)
   const pickerButtonRef = useRef<HTMLButtonElement>(null)
@@ -143,9 +143,9 @@ const Composer = forwardRef<
       currentComposerMessageInputRef.current?.getText() === '' &&
       !draftState.file
     ) {
-      setCanSendMessage(false)
+      setShowSendButton(false)
     } else {
-      setCanSendMessage(true)
+      setShowSendButton(true)
     }
   }, [currentComposerMessageInputRef, draftState])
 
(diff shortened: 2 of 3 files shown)


r10s:
my first reaction was, that the button should be disabled, and not hidden [^1]

then i checked whatsapp, signal, telegram - all of them hide instead disable. i still do not think, it is the way to go, but ux is also what users are used to, so, probably fine, at least we can give it a try.

looking closer, however, all of them have _less or none_ moving elements on showing; the send button _replaces_ eg. the "voice message" button, we should keep that in mind, when we go for that. but it is probably also okay for now

but we should adapt the padding right of the emoji button, that just does not look good (it is fine to add some general padding also right of the sending button)

current state:

Screenshot 2025-03-20 at 15 36 00

suggestion (roughly, i'd use the same spacing as the paper clip):

Screenshot 2025-03-20 at 15 36 00 copy

[^1]: ui/ux rule of thumb is that if a control is available in general, but not just now, and an easily be enabled by the user, it is better to hide, and not to disable

nicodh:
> my first reaction was, that the button should be disabled, and not hidden

I just looked at the App how it is handled there

packages/frontend/src/components/composer/Composer.tsx
```diff
@@ -91,6 +91,7 @@ const Composer = forwardRef<
   const chatId = selectedChat.id
   const [showEmojiPicker, setShowEmojiPicker] = useState(false)
   const [showAppPicker, setShowAppPicker] = useState(false)
+  const [canSendMessage, setCanSendMessage] = useState(false)
```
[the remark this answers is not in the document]
nicodh: Renamed it to showSendButton

r10s (approved):
cannot say much to the code, but tests were fine :)

packages/frontend/scss/composer/_composer.scss
```diff
@@ -79,7 +79,7 @@
 
     .emoji-button {
       height: 40px;
-      padding: 0px 3px;
+      padding: 0px 8px 0px 3px;
```
WofWca, line 82:
```suggestion
      padding: 0px 3px;
      margin-inline-end: 5px;
```

This makes it look like this

image


Otherwise it's like this 

image

packages/frontend/src/components/composer/Composer.tsx
```diff
@@ -137,6 +138,17 @@ const Composer = forwardRef<
...
+  useEffect(() => {
+    if (
+      currentComposerMessageInputRef.current?.getText() === '' &&
+      !draftState.file
+    ) {
+      setShowSendButton(false)
+    } else {
+      setShowSendButton(true)
+    }
+  }, [currentComposerMessageInputRef, draftState])
```
WofWca, line 150: This looks fragile to me. The effect only runs when `draftState` changes, but it changes only when we save the draft to the back-end.
Additionally, `showSendButton` also affects the button when it's an "edit message" button.

Does it have to be an effect? `showSendButton` is basically a computed value, and not a state of its own.
Can't we just

```typescript
const showSendButton =
  currentComposerMessageInputRef.current?.getText() !== '' ||
  draftState.file != null`?
```

WofWca (commented):
After a bit of testing this appears to work alright.

packages/frontend/src/components/composer/Composer.tsx
WofWca: This would also solve the

>  Only downside so far: it does not work when a message is edited, but we have a warning there if the send button is clicked without a message text

WofWca: Hmm, I just realized that `currentComposerMessageInputRef.current?.getText()` will still only run on the re-renders of this component, because the state of the text input is not exposed to it. Maybe it's time to lift the state of the text input?

Either way what I suggested above should be an improvement, unless it somehow doesn't work.

nicodh: >  Can't we just
> 
> ```ts
> const showSendButton =
>   currentComposerMessageInputRef.current?.getText() !== '' ||
>   draftState.file != null
> ```

As you mentioned already that does not hide the button after sending the message since there is no rerender in the Composer component.

> The effect only runs when draftState changes, but it changes only when we save the draft to the back-end.

Yes, but that happens after each change so it is a reliable trigger I would say

But anyway I will investigate for more possibilities. As said above this is a minimistic approach so far

WofWca: Even if rewriting this without an effect doesn't improve behavior, I think it's better because it's simpler code-wise.

WofWca:
I rebased it, but this MR doesn't seem to fix the E2E test flakiness. Now for me it fails at

https://github.com/deltachat/deltachat-desktop/blob/95e8348f6a2a92b01f301ace0b6da0cd27556d3c/packages/e2e-tests/tests/basic-functionality.spec.ts#L143

and the input is empty, which I think happens when the draft gets cleared.

packages/frontend/src/components/composer/Composer.tsx
nicodh: I kept the useEffect since we need the changes from ComposerInput to trigger a render as said above. I included now also the edit text scenario.

This is as said an approach to avoid much refactoring since we will do that "later" with the Composer refactoring.

But of course if you have a better idea you are welcome to propose a MR :-)

WofWca: With the latest commit the "send" button doesn't disappear after you send a message.

If we really want this behavior, I suggest to revert that last commit and merge. I tested the MR prior to that commit and it seemed to work alright.

nicodh: > With the latest commit the "send" button doesn't disappear after you send a message.

So removing the text from the textfield does not trigger the onChange event. It would need another onChange call  
immediately after sending the message:
```
this.props.sendMessageOrEditRequest()
this.props.onChange('')
```

WofWca: It has to do with the fact that `textareaRef.setText()` is used outside of the component, I think it's async after the message gets sent, in `useDraft`.

WofWca: Perhaps as another temporary hack one could add `onChange` to inside of the `setText()` method of the `ComposerTextarea` component.

nicodh:
> Now the button also persists its state when switching between chats. E.g. if you switch to a chat that already has a draft, the button will not be displayed.

So another call of onChange was needed. Now we have 4 places in ComposerMessageInout where we call it, which is not nice. But that's one of the reasons why we want to refactor the Composer I assume

Tested many use cases and all were fine so far

WofWca (approved):
Based on a bit of testing, this works fine.

But it's very fragile.

Issue #4200, opened 2024-10-10
Investigate whether SettingsStore can be removed


Simon-Laux:
**What?**

Investigate whether the whole SettingsStore can be removed, or at least the account / core-config (`settings`-key) specific part of it.

**Background:**

`SettingsStore` was made at time where core calls to get config was still expensive.
Meanwhile circumstances changed:
- core configuration is cached inside of the core so they are not time-expensive sql requests anymore
- calls to core are not blocking anymore thanks to jsonrpc and async.

I think it still makes sense to batch getConfig over `BackendRemote.rpc.batchGetConfig` in the config screens that load multiple keys at once, but only when/before opening those screens.

**Why now?**

Recent discussions/questions in https://github.com/deltachat/deltachat-desktop/pull/4192 brought this topic to my attention again.

If we remove the account specific stuff from `SettingsStore`, then we don't need to reload it anymore on account switching. And thus reduce potential race-condition bugs caused by it.

r10s:
closing stale issue; we usually do not keep refactorings and code-level things open.

if it turns out to be advantageous to do some refactoring, it should just be done it on adding/improving a feature; it is usually not a task on its own and, as here, tends to get stale

WofWca:
IMO at least the `desktopSettings` part should be kept. While Core configs can be changed by core itself, invalidating the cache, we don't expect the same for `desktopSettings`. So it's nice not to re-load them every time.

PR #4819, opened 2025-03-19
Delete all with updated core

nicodh:
#skip-changelog

### DIFF
--- a/packages/frontend/src/components/AccountListSidebar/AccountItem.tsx
+++ b/packages/frontend/src/components/AccountListSidebar/AccountItem.tsx
@@ -139,7 +139,7 @@ export default function AccountItem({
         // set Timeout forces it to be run after react update
         setTimeout(() => {
           ActionEmitter.emitAction(KeybindAction.Settings_Open)
-        }, 0)
+        }, 100)
       },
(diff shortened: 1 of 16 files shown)


packages/frontend/src/components/AccountListSidebar/AccountItem.tsx
```diff
@@ -139,7 +139,7 @@ export default function AccountItem({
         // set Timeout forces it to be run after react update
         setTimeout(() => {
           ActionEmitter.emitAction(KeybindAction.Settings_Open)
-        }, 0)
+        }, 100)
```
Simon-Laux, line 142: I don't like adding delays to fix issues.

Simon-Laux: settings should either use react suspense to wait for the settings store, or probably even better we should explore to remove the settings store https://github.com/deltachat/deltachat-desktop/issues/4200

Simon-Laux: maybe too much work for the moment, to get the release out I'd approve it when there is also a comment here explaining the delay.

nicodh:
Tests are now in #4827

Issue #4783, opened 2025-03-15
Remove "Import keys" from settings


nicodh:
see discussion here: https://github.com/deltachat/interface/issues/80

Needs updates in the help pages:
- [x] https://github.com/deltachat/deltachat-pages/blob/main/en/help.md#can-i-reuse-my-existing-private-key-importkey
- [x] https://github.com/deltachat/deltachat-pages/blob/main/en/help.md#i-cant-import-my-existing-pgp-key-into-delta-chat

See https://github.com/deltachat/deltachat-pages/issues/1069

r10s:
i ticked the missing faq entries, which are updated meanwhile.

reopening the issue, however, as the import via an autocrypt setup message is still possible, but should not according to https://github.com/deltachat/interface/issues/80 - the corresponding android and iOS PR have that removed as well. reason is the same described at https://delta.chat/en/help#importkey

PR #4822, opened 2025-03-20
fix: remove handling for receiving autocrypt message

nicodh:
solves #4783


r10s (approved):
tested, and code lgtm (lots of removed things, aways a joy to see code being remved :)

PR #4826, opened 2025-03-20
tauri: hardening: more validation for commands

WofWca:
See individual commits for descriptions.

### DIFF
--- a/packages/target-tauri/src-tauri/src/temp_file.rs
+++ b/packages/target-tauri/src-tauri/src/temp_file.rs
@@ -61,10 +61,8 @@ async fn create_tmp_file(app: &AppHandle, name: &str) -> Result<(File, PathBuf),
 
     let file_name = file_as_path
         .as_ref()
-        .components()
-        .last()
-        .ok_or(Error::InvalidFileName)?
-        .as_os_str();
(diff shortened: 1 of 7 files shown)


packages/target-tauri/src-tauri/src/temp_file.rs
```diff
@@ -163,23 +176,41 @@ pub(crate) async fn remove_temp_file(app: AppHandle, path: SafePathBuf) -> Resul
...
+
     {
-        return Err(Error::SourcePathOutsideOfBlobsDir);
+        let maybe_blobs_dir = source_path
+            .parent()
+            .ok_or(Error::SourcePathOutsideOfBlobsDir)?;
+        if maybe_blobs_dir
+            .file_name()
+            .ok_or(Error::SourcePathOutsideOfBlobsDir)?
+            != "dc.db-blobs"
```
Simon-Laux, line 197: I choose the ends with intentional as I thought someone might possible have an db with a different name, though thinking about it that edge-case will probably not happen, so checking for exact name is also fine

Simon-Laux (approved):
clippy has 2 issues, but besides that looks good, didn't test

PR #4829, opened 2025-03-21
Prepare release 1.56.0

nicodh:


WofWca:
Shouldn't we add a device message?

CHANGELOG.md
```diff
@@ -3,23 +3,19 @@
 ## [Unreleased][unreleased]
 
 ### Added
-- delete message for all chat members #4716
 
 ### Changed
 - Update local help (2025-03-20)
```
WofWca, line 8: Why is this part left out?

WofWca (commented):
I like this changelog approach more. Though I want to point out that it is different from what we had before.

Code-wise looks alright, apart from my other comments. But I didn't check the translation files.

CHANGELOG.md
nicodh: I moved it to the changed section

nicodh:
> Shouldn't we add a device message?

I added one now. Mentioned also the removal of e2e encryption preferences & import keys since that are kind of "breaking changes"

packages/frontend/src/deviceMessages.ts
```diff
@@ -27,18 +27,16 @@ export async function updateDeviceChat(
...
-    💥 Highlight the first unread message on opening chat
-    🎹 New shortcuts for search & new chats:
-       - Search: Ctrl/Cmd + F
-       - Search in Chat: Ctrl/Cmd + Shift + F
-       - New Chat: Ctrl/Cmd + N
-    You can find all Shortcuts in the Help menu or with Ctrl + /
+    📝 Edit messages: already sent messages can be edited now
+    ❌ Delete messages for all chat members as a new option
+    🔒 E2e encryption is always preferred when available (and preferences are removed from setting)
+    ⚠️ Import of encryption keys is removed (to avoid various related [limitations](https://github.com/deltachat/interface/issues/80) )
```
WofWca, line 36:
```suggestion
📝 Edit messages: already sent messages can be edited now
❌ Delete messages for all chat members as a new option
🔒 E2E encryption is always preferred when available (and preferences are removed from setting)
⚠️ Import of encryption keys is removed (to avoid various related [limitations](https://github.com/deltachat/interface/issues/80))
```

Removed this weird indentation, and added a few one-character changes

image

WofWca (approved):
Less goooo

packages/frontend/src/deviceMessages.ts
nicodh: Hmm, I consider the indented list better readable... ?

nicodh: image

WofWca: IDK, if the sentences didn't wrap then maybe yes.

As you wish.

PR #4834, opened 2025-03-22
tauri: created tx lib & added pluralisations to tx function

nojisan:
- **added fallback to en when a key doesn't exist in a particular language**
- **fixed formatting and added changelog message**
- **changelog message corrected**
- **Pluralisation in translation function**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,13 +3,13 @@
 ## [Unreleased][unreleased]
 
 ### Added
-- tauri: added fallback to en when a key doesn't exist in a particular language #4818
 
 ### Changed
 
 ### Fixed
 - tauri: improve security a little
 - tauri: improve performance a little #4810
+- tauri: translation fn: added fallback to en when a key doesn't exist in a particular language #4818
 
 <a id="1_55_0"></a>
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased][unreleased]
 
 ### Added
+- tauri: added fallback to en when a key doesn't exist in a particular language #4818
 
 ### Changed
 
--- a/packages/target-tauri/src-tauri/src/html_window/mod.rs
+++ b/packages/target-tauri/src-tauri/src/html_window/mod.rs
@@ -10,6 +10,7 @@ use tauri::{
 use tauri_plugin_dialog::DialogExt;
 use tauri_plugin_opener::OpenerExt;
 use tauri_plugin_store::StoreExt;
+use translationfn::Substitution;
 
 use crate::{
     html_window::{
@@ -173,10 +174,14 @@ pub(crate) async fn open_html_window(
                 let url2 = url.clone();
                 app_arc
                     .dialog()
-                    .message(
-                        format!("Punycode detected: {}", tx.sync_translate("puny_code_warning_question", /*"puny_code_decode_host(orginal_host_name)"*/)),
-                    )
-                    .title(tx.sync_translate("puny_code_warning_header"))
+                    .message(format!(
+                        "Punycode detected: {}",
+                        tx.sync_translate(
+                            "puny_code_warning_question",
+                            Substitution::String(vec![&puny_code_encode_host(orginal_host_name)])
+                        )
+                    ))
+                    .title(tx.sync_translate("puny_code_warning_header", Substitution::None))
                     .kind(tauri_plugin_dialog::MessageDialogKind::Warning)
                     .buttons(tauri_plugin_dialog::MessageDialogButtons::OkCancelCustom(
                         // Todo: use tx function here (Problems with Lifetime)
--- a/packages/target-tauri/translationfn/src/lib.rs
+++ b/packages/target-tauri/translationfn/src/lib.rs
@@ -1,0 +1,278 @@
+use std::collections::HashMap;
+
+use intl_pluralrules::{PluralCategory, PluralRuleType, PluralRules};
+use log::error;
+use regex::{self, Regex};
+use unic_langid::{LanguageIdentifier, LanguageIdentifierError};
+
+pub enum Substitution<'a> {
+    None,
+    String(Vec<&'a str>),
+    Quantity(usize),
+    QuantityFloat(f32),
+}
+
+pub struct TranslationEngine {
+    messages: HashMap<String, HashMap<String, String>>,
+    plural_rules: PluralRules,
+    var_finder: Regex,
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("language id malformed")]
+    LangIdMalformed,
+    #[error("parsing of lang id failed {0:?}")]
+    LangIdParsingFailed(LanguageIdentifierError),
+    #[error("Plural rule not found {0}")]
+    PluralRuleNotFound(&'static str),
+    #[error("Invalid RegEx Pattern")]
+    RegEx(#[from] regex::Error),
+}
+
+impl TranslationEngine {
+    pub fn new(
+        messages: HashMap<String, HashMap<String, String>>,
+        langid: &str,
+    ) -> Result<Self, Error> {
+        let langid: LanguageIdentifier = langid
+            .split("_")
+            .next()
+            .ok_or(Error::LangIdMalformed)?
+            .parse()
+            .map_err(Error::LangIdParsingFailed)?;
+        let pr = PluralRules::create(langid, PluralRuleType::CARDINAL)
+            .map_err(Error::PluralRuleNotFound)?;
+        let var_finder = Regex::new(r"%\d\$[\w\d]")?;
+
+        Ok(TranslationEngine {
+            messages,
+            plural_rules: pr,
+            var_finder,
+        })
+    }
+    pub fn translate(&self, key: &str, substitution: Substitution) -> String {
+        if let Some(entry) = self.messages.get(key) {
+            match substitution {
+                Substitution::None => {
+                    if let Some(message) = entry.get("message") {
+                        message.to_owned()
+                    } else {
+                        error!("Message not existing for {key}");
+                        key.to_owned()
+                    }
+                }
+                Substitution::String(items) => {
+                    //TODO: remove unwrap, use log instead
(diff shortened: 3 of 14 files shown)


packages/target-tauri/translationfn/src/lib.rs
```diff
@@ -0,0 +1,278 @@
...
+            .unwrap();
+        let txen = TranslationEngine::new(msgs, "en").unwrap();
+        let result = txen.translate("n_hours", Substitution::Quantity(1));
+        assert_eq!(result, "1 hour");
+
+        let result = txen.translate("n_hours", Substitution::Quantity(2));
+        assert_eq!(result, "2 hours");
+    }
+
+    //TODO: test for dead errors to not crash and are returned
```
WofWca, line 277: This comment is not clear to me, but maybe it's just me.

packages/target-tauri/translationfn/src/lib.rs
```diff
@@ -0,0 +1,278 @@
...
+                        Ok(PluralCategory::FEW) => "few",
+                        Ok(PluralCategory::MANY) => "many",
+                        Ok(PluralCategory::OTHER) => "other",
+                        Err(_) => {
+                            error!("Failed to  {quantity}");
+                            return key.to_owned();
+                        }
+                    };
+                    if let Some(message) = entry.get(quantity_key) {
+                        message.replace("%d", &quantity.to_string())
```
WofWca, line 94: So this won't work for strings with multiple things to substitute, and also with some strings with a single parameter, e.g. `Export all %1$d profiles`.

TBH I'm not sure about having an implementation that neither is used nor does work properly.
Maybe mentioning it in the comments would be alright.

packages/target-tauri/translationfn/src/lib.rs
```diff
@@ -0,0 +1,278 @@
...
+                    };
+                    if let Some(message) = entry.get(quantity_key) {
+                        message.replace("%d", &quantity_float.to_string())
+                    } else {
+                        error!(
+                            "Message for key was found {key}, but variant {quantity_key} to represent quantity of {quantity_float} is missing"
+                        );
+                        key.to_owned()
+                    }
+                }
```
WofWca, line 123: I believe this could be rewritten by making the function generic.
But this is also manageable as is.

packages/target-tauri/translationfn/src/lib.rs
```diff
@@ -0,0 +1,278 @@
...
+                Substitution::None => {
+                    if let Some(message) = entry.get("message") {
+                        message.to_owned()
+                    } else {
+                        error!("Message not existing for {key}");
+                        key.to_owned()
+                    }
+                }
+                Substitution::String(items) => {
+                    //TODO: remove unwrap, use log instead
```
WofWca, line 66: Shouldn't this be addressed prior to the merge?
We wouldn't want to panic in a translation function.

WofWca (changes requested):
I left a couple of comments, the important one being a panic.

packages/target-tauri/translationfn/src/lib.rs
Simon-Laux:
```suggestion
    // TODO: test for that errors to not crash and are returned as errors instead (creation of translation engine)
    // TODO: test that errors of translate function do not crash and just return the key (not sure if we can check if logging worked, might be too difficult/complex, so just checking for not panicking is enough)
```

packages/target-tauri/translationfn/src/lib.rs
Simon-Laux: afaik we don't have quantity translations with multiple fields to substitute, or do we?

Simon-Laux: the "%d" should probably also handle the "%1$d" case, so we should adjust it, maybe:
```suggestion
                        let quantity_string = quantity.to_string();
                        message.replace("%1$d", & quantity_string).replace("%d", & quantity_string)
```

packages/target-tauri/translationfn/src/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+                            "Message for key was found {key}, but variant {quantity_key} to represent quantity of {quantity_float} is missing"
+                        );
+                        key.to_owned()
+                    }
+                }
+            }
+        } else {
+            error!("Translation for key {key} missing");
+            key.to_owned()
+        }
```
WofWca, line 156: A nitpick to reduce nesting greatly. Feel free to ignore.

```suggestion
        let Some(entry) = self.messages.get(key) else {
            error!("Translation for key {key} missing");
            return key.to_owned();
        };
        match substitution {
            Substitution::None => {
                if let Some(message) = entry.get("message") {
                    message.to_owned()
                } else {
                    error!("Message not existing for {key}");
                    key.to_owned()
                }
            }
            Substitution::String(items) => {
                let Some(message) = entry.get("message") else {
                    error!("Message not existing for {key}");
                    return key.to_owned();
                };
                let mut msg = message.clone();
                for matches in self.var_finder.find_iter(message) {
                    let from = matches.as_str();
                    let Some(position) = from.chars().nth(1).and_then(|from| from.to_digit(10))
                    else {
                        error!("Invalid format of replacement pattern");
                        return key.to_owned();
                    };
                    if let Some(item) = items.get(position as usize - 1) {
                        msg = msg.replace(from, item);
                    } else {
                        error!("Invalid item position {position}");
                        return key.to_owned();
                    }
                }
                msg
            }
            Substitution::Quantity(quantity) => {
                let quantity_key = match self.plural_rules.select(quantity) {
                    Ok(PluralCategory::ZERO) => "zero",
                    Ok(PluralCategory::ONE) => "one",
                    Ok(PluralCategory::TWO) => "two",
                    Ok(PluralCategory::FEW) => "few",
                    Ok(PluralCategory::MANY) => "many",
                    Ok(PluralCategory::OTHER) => "other",
                    Err(_) => {
                        error!("Failed to  {quantity}");
                        return key.to_owned();
                    }
                };
                if let Some(message) = entry.get(quantity_key) {
                    let quantity_string = quantity.to_string();
                    message
                        .replace("%1$d", &quantity_string)
                        .replace("%d", &quantity_string)
                } else {
                    error!(
                        "Message for key was found {key}, but variant {quantity_key} to represent quantity of {quantity} is missing"
                    );
                    key.to_owned()
                }
            }
            Substitution::QuantityFloat(quantity_float) => {
                let quantity_key = match self.plural_rules.select(quantity_float) {
                    Ok(PluralCategory::ZERO) => "zero",
                    Ok(PluralCategory::ONE) => "one",
                    Ok(PluralCategory::TWO) => "two",
                    Ok(PluralCategory::FEW) => "few",
                    Ok(PluralCategory::MANY) => "many",
                    Ok(PluralCategory::OTHER) => "other",
                    Err(_) => {
                        error!("Failed to  {quantity_float}");
                        return key.to_owned();
                    }
                };
                if let Some(message) = entry.get(quantity_key) {
                    message.replace("%d", &quantity_float.to_string())
                } else {
                    error!(
                        "Message for key was found {key}, but variant {quantity_key} to represent quantity of {quantity_float} is missing"
                    );
                    key.to_owned()
                }
            }
        }
```

packages/target-tauri/translationfn/src/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+    clippy::correctness,
+    clippy::all,
+    clippy::wildcard_imports,
+    clippy::needless_borrow,
+    clippy::cast_lossless,
+    clippy::unused_async,
+    clippy::explicit_iter_loop,
+    clippy::explicit_into_iter_loop,
+    clippy::cloned_instead_of_copied
+)]
```
WofWca, line 13: Is this here by mistake? If not, IMU this turns `correctness` from error into a warning, right? I don't think we want that.

WofWca (commented):
OK, the `unwrap()`s are gone. This needs a rebase and one final test, then can be merged IMO.

packages/target-tauri/translationfn/src/lib.rs
WofWca: Let's mark as resolved since it's not important

nicodh:
Noone seems to work on this. Can be reopened if needed

nicodh:
Sorry didn't see that the last commit is from November...

Simon-Laux (approved):

.github/workflows/rust-tauri-lint.yml
```diff
@@ -63,3 +63,32 @@ jobs:
...
+        with:
+          toolchain: 1.88
+          components: clippy
+          override: true
+      - name: fake html dist dir to make it build
+        run: |
+          mkdir packages/target-tauri/html-dist
+          touch packages/target-tauri/html-dist/tauri_main.html
+      - name: Unit tests
+        run: cargo test --all
```
Simon-Laux, line 96:
```suggestion
        env:
            VERSION_INFO_GIT_REF: "testing"
        run: cargo test --all
```

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 ## [2.35.0] - 2025-12-21
 
 ### Added
+- tauri: created translationfn library & added pluralisation functionality #4834
```
```suggestion
```

We now auto-generate CHANGELOG.

packages/target-tauri/src-tauri/src/html_window/mod.rs
```diff
@@ -199,16 +200,16 @@ pub(crate) async fn open_html_window(
                 .message(format!(
                     "{}\n\n{}",
                     tx.sync_translate(
-                        "puny_code_warning_question", /*"puny_code_decode_host(orginal_host_name)"*/
+                        "puny_code_warning_question",
+                        Substitution::String(vec![&puny_code_encode_host(orginal_host_name)])
                     ),
-                    // TODO substitution
-                    tx.sync_translate("puny_code_warning_description"),
+                    tx.sync_translate("puny_code_warning_description", Substitution::None),
```
WofWca:
```suggestion
                    // TODO substitution
                    tx.sync_translate("puny_code_warning_description", Substitution::None),
```

The TODO has not been addressed.

PR #4839, opened 2025-03-23
tauri: improvement: clear tmp and stop IO on exit

WofWca:
For reference, here is our Electron exit function:

https://github.com/deltachat/deltachat-desktop/blob/8d1b7c8663339e125df4f760f612c89d7b4cc64d/packages/target-electron/src/index.ts#L242-L288

FYI we don't need to `DesktopSettings.saveImmediate()`, because https://tauri.app/plugin/store/ states:

> // You can manually save the store after making changes.
> // Otherwise, it will save upon graceful exit

- [ ] Please verify that `stop_io()` is used properly. Do we need something apart from `stop_io().await`?

I have tested the changes, and it doesn't appear to error out or hang. But, again, I'm not sure whether this is a proper shutdown.


WofWca:
OK, I suggest to address the TODOs in separate MRs. This should already be an improvement.

Simon-Laux (approved):

PR #4851, opened 2025-03-25
tauri: webxdc hardening: disable webrtc by overwriting the JS API

Simon-Laux:
depends on https://github.com/tauri-apps/tauri/pull/13076

Try it on:
- [x] MacOS: works ✅ 
- [x] Linux: works ✅ 
- [x] Windows

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,7 @@
 - fix sticker folder not resolved on windows #4939
 - tauri: improve performance a little #4812
 - settings: fix: wait for setting to be applied before calling callback #4754
+- tauri: prevent webrtc from being accessed in webxdc apps #4851
 
 ### Removed
 - remove experimental option to disable IMAP IDLE #4991
--- a/packages/target-tauri/src-tauri/src/webxdc/commands_main_window.rs
+++ b/packages/target-tauri/src-tauri/src/webxdc/commands_main_window.rs
@@ -47,13 +47,15 @@ console.log("hello from INIT_SCRIPT")
 
 // remove peer connection by overwriting api
 try {
-window.RTCPeerConnection = ()=>{};
-RTCPeerConnection = ()=>{};
-} catch (e){console.error("failed to overwrite RTCPeerConnection apis",e)}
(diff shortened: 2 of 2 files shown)


r10s:
this means macOS does not need FILL500, right?

WofWca:
> this means delta-tauri on macOS does not need FILL500, right?

Yes. On macOS, Tauri uses WebKit, and the 500 hack only applies to Chromium.

packages/target-tauri/src-tauri/capabilities/main-window.toml
```diff
@@ -10,7 +10,7 @@ permissions = [
     # but they aren't, so we have to add this line to be able to use channels.
     # See https://github.com/tauri-apps/tauri/pull/13032.
     # TODO remove this after Tauri upgrade.
-    "core:__TAURI_CHANNEL__:allow-fetch",
+    # "core:__TAURI_CHANNEL__:allow-fetch",
```
WofWca, line 13: Would be nice to have this change in a separate commit at least.

WofWca (commented):
Damn, this seems to be working on Windows!

I tried creating a webxdc with a top-level inline `<script`, as such

```html
<html>
  <head>
    <script>console.log('App script ran!', RTCPeerConnection);</script>
```

I also added an iframe with the same top-level script. This is what it prints!

image

I, however, did not try to bypass it otherwise (and I'm not sure how one would do it).

Perhaps prior to merging this MR we'd want to see the upstream MR merged. It's getting feedback that it's breaking, so we'll probably have to update this MR then as well.

WofWca (changes requested):
This comment will need to be updated

https://github.com/deltachat/deltachat-desktop/blob/11125e0c581c4c00b36313fa461cccbf4ae4c5d6/packages/target-tauri/src-tauri/src/webxdc/commands_main_window.rs#L39-L54

WofWca (approved):
Yep, works on Windows!

image

WofWca:
The commit message / MR title need to be updated perhaps.

PR #4855, opened 2025-03-26
Add style for saved message icon

### LINKED ISSUE #4854: "saved messages" icon not visible on images
using the new "save" function, the "bookmark" icon is not really visible on images. seems it is fully transparent or so:

Image

nicodh:
resolve #4854 

#skip-changelog

image


WofWca (approved):
Did not test, but the code is alright. The same color as for the padlock.

PR #4864, opened 2025-03-27
Update message parser

### LINKED ISSUE #4559: message-parser bug: labeled links are parsed incorrectly

nicodh:
resolves #4559

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - improve fatal error dialog readability by removing color from deltachat-rpc-server errors
 - prevent dragging around of webxdc icon #4740
 - tauri: clear temp folder on exit #4839
+- fix wrong punycode warnings in links #4864
 
 <a id="1_56_0"></a>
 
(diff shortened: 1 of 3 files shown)


WofWca (approved):
Assuming that the new version of the parser is better (I didn't check), this is good, just an upgrade.

PR #4866, opened 2025-03-28
show explicit action on 'start backup' button

r10s:
according to interface guidelines,
it is recommended to show the action on the button, where possible and reasonable wrt effort.

we do not always have the verb/action,
but in this case we do :)

the string comes from iOS,
thanks for the hint in the forum.

before / after:

Screenshot 2025-03-28 at 12 31 32
Screenshot 2025-03-28 at 12 25 58


WofWca (approved):
Yep, this is better.

Simon-Laux (approved):

PR #4869, opened 2025-03-28
deny/disable link preview on long-press on macOS

Simon-Laux:
depends on https://github.com/tauri-apps/tauri/pull/13090


Simon-Laux:
converting to draft because it needs to wait for tauri

WofWca (approved):
Didn't test, but this should do it.

It would be nice to fix the Clippy warnings though.

PR #4871, opened 2025-03-29
make tauri compile for android

Simon-Laux:
- **make it compile again for android and attempt to make locales work**
- **logs on android**
- **hardcode locales on android**
- **switch to esm modules for runtime and bundle.js**

| |  |
|----|----|
|Bildschirmfoto 2025-03-29 um 18 36 35 | Bildschirmfoto 2025-03-29 um 18 36 23 |

---

- [x] make locale loading work
- [x] disable menu
- [x] make it compile


r10s:
interesting technical aspect - are apps developed using tauri and run on android a thing?

Simon-Laux:
>  are apps developed using tauri and run on android a thing?

It is relatively new and seems to be more for simpler apps, like multiple web views is not really useable yet out of the box. I don't know out of my head if there are any production grade mobile apps yet. Maybe you can find examples in some tauri awesome list or on the showcase channel of their discord server.

For tauri in general there is a Minecraft launcher and modmanager https://github.com/modrinth/code/tree/main/apps/app, but thats the only tauri app that I know of so far I have really used. But that's only my personal experience.

r10s:
thanks for the background information!

packages/frontend/bin/build-frontend-ts.mjs
```diff
@@ -26,7 +26,7 @@ function config(options) {
   return {
     entryPoints: ['src/main.tsx'],
     bundle: true,
-    minify: isMinify,
+    minify: false,
```
WofWca, line 29: Why? `isMinify` is unused now.

CHANGELOG.md
```diff
@@ -4,10 +4,12 @@
 
 ### Added
 - tauri: support for webxdc #4740
+- tauri: experimental: make it compile for android #4871
 
 ### Changed
 - switch to account the webxdc is from when sending to chat (tauri and electron edition) #4740
 - change the Reply button for messages to be a verb rather than a noun #4853
+- change script format and imports to esm/module #4871
```
WofWca, line 12: This is not a user-visible change, right?

packages/target-tauri/src-tauri/src/lib.rs
```diff
@@ -128,11 +130,15 @@ pub fn run() {
             temp_file::write_temp_file,
             temp_file::remove_temp_file,
             temp_file::copy_blob_file_to_internal_tmp_dir,
+            #[cfg(desktop)]
             webxdc::commands_main_window::on_webxdc_message_changed,
+            #[cfg(desktop)]
```
WofWca, line 135: It'd be nice to have it documented somewhere why these are desktop-only. Maybe here, or in a separate `.md` file.

packages/target-tauri/src-tauri/src/state/app.rs
```diff
@@ -41,8 +44,11 @@ impl AppState {
 
         let current_log_file_path = get_current_log_file_task.await??;
 
-        create_tmp_folder(app.handle()).await?;
-        clear_tmp_folder(app.handle()).await?;
+        #[cfg(not(target_os = "android"))]
```
WofWca, line 47: Why is it sometimes `cfg(desktop)` and sometimes `cfg(not(android))`?

WofWca (commented):
I haven't checked how well it works on Android, but I did read through the code to see how it affects Desktop, and it pretty much doesn't. Code-wise it's also OK, maintainability of the main targets (desktop) doesn't suffer much.

But my question is: how committed are we to supporting Android? Is it fine to break it from time to time?
Should we add CI for it to monitor whether it breaks?

packages/frontend/bin/build-frontend-ts.mjs
Simon-Laux: good catch a debugging hack that was left in

CHANGELOG.md
Simon-Laux: ideally not, it would be user visible if were any problems with it.

we can prefix it with sth. to mark it as technical detail.

packages/target-tauri/src-tauri/src/lib.rs
Simon-Laux: added a few comments

CHANGELOG.md
```diff
@@ -3,12 +3,14 @@
 ## [Unreleased][unreleased]
 
 ### Added
+<<<<<<< HEAD
```
WofWca, line 6:
```suggestion
```

WofWca (approved):

PR #4872, opened 2025-03-29
Image improvements

### LINKED ISSUE #4410: Find an appropriate solution for displaying images
We need a solution to meet following requirements:

- [ ] the height of an image must be set even before the image is loaded (core provides dimensions in the message object)
- [ ] image-only messages should not show (extreme) padding except for very narrow images
- [ ] define a range to avoid too big and too small images
- [ ] narrow image-only messages should display a footer wide enough for timestamp and reactions

Related to:
- #4404
- PRs #4407 & #4315

nicodh:
Precalculate the height for images. See comment in  https://github.com/deltachat/deltachat-desktop/pull/4872/files#diff-271e25946bbd8fe9873cc1b850876b8f5b24a5ab8fad20802d458a42e895a011R62

resolves #4410

- [x] the height of an image must be set even before the image is loaded (core provides dimensions in the message object)
- [x] image-only messages should not show (extreme) padding except for very narrow images
- [x] define a range to avoid too big and too small images
- [x] narrow image-only messages should display a footer wide enough for timestamp and reactions
- [x] avoid linebreaks in footer

Before
image-improvements-before

After

image-improvements-after

Small screen with landscape ratio

very-small-screen1


r10s (approved):
as discussed in av, lgtm

Issue #4218, opened 2024-10-16
unexpected zoom behaviour in image cropper


r10s:
in the profile image selector, the zoom gestures are different from the [image viewer](https://github.com/deltachat/deltachat-desktop/issues/2632), but not correct.

in the profile image selector the zoom gestures are exactly swapped - the gesture that should make the image smaller (pinch in) makes it actually bigger:

https://github.com/user-attachments/assets/bded386f-6778-454f-973e-0930b027cc79

this is unintuitive, imagine you have a sponge between the fingers, that would also get smaller :)

and again, this is not a personal preference, but standard :)

in the past, the swapped behaviour also gave the impression that zooming is entirely impossible, btw, which led to adding the +/- buttons

maxphilippov:
Oh, so this is a duplicate of #3893 then? Just the pinch zoom is incorrect?

r10s:
i think, #3893 can be closed then, there is nothing left than this issue, which is more on point

r10s:
pinch/unpinch is no longer flipped, but does not work at all, instead, the image is moved a little bit left/right, it looks weird to the user.

r10s:
i have to correct myself: pinch/unpinch is working on current main, but it is just pretty slow, so the impact of one pinch/unpinch scales the image only a _little bit_, compared to what it does on other apps.

but in general, it is there and working as expected.

closing the issue

PR #4873, opened 2025-03-29
Don't upscale images in full screen view

nicodh:
Is there a reason why we upscale images in full screen view?

Before: (image has size 120 x 120 px)

image

After:

image

#skip-changelog


r10s:
hm, not sure. 

obviously, not-zooming is better for icons and small png - but eg. for for photos sent in low quality, on common screen sizes, zooming seems to be a better default. photos are arguably the more common case

also, [as zooming in is not intuitive on many platforms](https://github.com/deltachat/deltachat-desktop/issues/4218), the old approach may be better in many cases - even though i know about the bug, i often forget how to zoom :)

what are signal/whatsapp/telegram doing here?

if not-zooming is a common thing, and #4218 is targeted, i would be fine with giving this PR a try, but for now, i would postpone it for given reasons

nicodh:
> i find it reasonable to show the image _larger_ than in the preview, maybe not full screen, but eg. twice the size or so. otherwise, the user just gets the same information as without full image view

well - images smaller than min-width are stretched (to have the minimal width at least) so the full view would show the image as it is "in real".

> were the complains about the current behaviour?

No. So no priority. To me it just looks "buggy" since it's unusual (as you say, other platforms don't do that)

> also, https://github.com/deltachat/deltachat-desktop/issues/4218, the old approach may be better in many cases - even though i know about the bug, i often forget how to zoom :)

> but all (WhatsApp, Signal) have intuitive zooming in place

Signal doesn't react on gestures in my tests. Neither on Linux nor on Mac. Even Ctrl + scroll does not work in Signal full screen view, which I would consider the most common way to zoom on desktop. (You even don't need a touchpad!) So no zooming at all

While Ctrl + scroll works in Deltachat for me on Linux and Mac for zooming.

Maybe discuss that occasionally in a call

r10s:
> Signal doesn't react on gestures in my tests. 

yes, signal does not use standard pinch/unpinch as whatsapp/telegram does. but doing pinch/unpinch gestures on signal does not result in weird behaviour, instead they default to a cursor that defaults to "zoom in". but i am nut suggesting that for deltachat, we should support standard  pinch/unpinch

> While Ctrl + scroll works in Deltachat for me on Linux and Mac for zooming

yes, also two finger slide work for zooming for me, it is not that there is nothing in place.

issue is that the _standard_ since iphone1, _pinch/unpinch_ does not work as expected

ftr, to the initial question what the reason for "fullscreen" is: it was introduced at https://github.com/deltachat/deltachat-desktop/pull/2516 before zooming was added. but, as said, before we have standard zoom gestures, i would not change the behaviour

> Maybe discuss that occasionally in a call

+1, let's discuss remaining open questions in call

r10s (approved):
as discussed, the pinch/unpinch is there on mac, but needs some acceleration, so it is not that bad.

idea is to scale all/some images to twice the size on opening, that should fit for must usecases, should not need much scaling, and is fine also for png etc.

otherwise lgtm

WofWca (approved):
This seems to work, but I didn't check the docs of the library that we use, as to whether this is what we're supposed to do.

We have some size-related CSS in `.attachment-view img` also FYI.

PR #4874, opened 2025-03-29
tauri: hardening: webxdc: disable all web permissions

WofWca:
#skip-changelog because Tauri webxdc has not been released yet.

I verified that putting `navigator.getUserMedia({ video: true }, console.log, console.error)` in the console no longer causes a permission prompt on Windows.

- [x] I just realized that the new build script might not be able to run on Windows, because it's a .sh script.
  OK, it runs in CI, but still I guess I need to rewrite it.

### DIFF
--- a/packages/target-tauri/src-tauri/src/webxdc/webxdc_scheme.rs
+++ b/packages/target-tauri/src-tauri/src/webxdc/webxdc_scheme.rs
@@ -116,6 +116,24 @@ pub(crate) fn webxdc_protocol<R: tauri::Runtime>(
                 // Ensure that the browser doesn't try to interpret the file
                 // as a PDF file. See below about PDF, `CONTENT_TYPE`.
                 .header(http::header::X_CONTENT_TYPE_OPTIONS, "nosniff")
+                // Make sure that the webxdc app cannot even request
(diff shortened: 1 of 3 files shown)


packages/target-tauri/src-tauri/src/webxdc/webxdc_scheme.rs
```diff
@@ -227,3 +245,87 @@ pub(crate) fn webxdc_protocol<R: tauri::Runtime>(
...
 }
+
+// Yep, there is no way to "disable all by default", see
+// https://github.com/w3c/webappsec-permissions-policy/issues/189.
+//
+// The list is taken from
+// https://github.com/w3c/webappsec-permissions-policy/blob/main/features.md.
+//
+// We make sure that this list is up to date using the
+// `/bin/webxdc-check-permissions-policy-count.sh` script.
```
Simon-Laux, line 256: still has old name?

WofWca:
```suggestion
// `/bin/webxdc-check-permissions-policy-count.js` script.
```

Simon-Laux (approved):

PR #4875, opened 2025-03-29
Do not render Markdown links if not enabled

nicodh:
Related to #4559

#skip-changelog since it's an advanced feature anyway

### DIFF
--- a/packages/frontend/src/components/message/MessageMarkdown.tsx
+++ b/packages/frontend/src/components/message/MessageMarkdown.tsx
@@ -23,11 +23,9 @@ let parseMessage: (message: string) => ParsedElement[] = m =>
   parse_desktop_set(m)
 
 SettingsStoreInstance.subscribe(newState => {
-  if (newState?.desktopSettings.experimentalEnableMarkdownInMessages) {
-    parseMessage = m => parse_text(m, true)
-  } else {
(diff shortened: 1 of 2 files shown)


Simon-Laux:
advanced features also need changelog entries

packages/frontend/src/components/message/MessageMarkdown.tsx
```diff
@@ -23,11 +23,9 @@ let parseMessage: (message: string) => ParsedElement[] = m =>
...
 
 SettingsStoreInstance.subscribe(newState => {
-  if (newState?.desktopSettings.experimentalEnableMarkdownInMessages) {
-    parseMessage = m => parse_text(m, true)
-  } else {
-    parseMessage = m => parse_desktop_set(m)
-  }
+  const markDownEnabled =
+    !!newState?.desktopSettings.experimentalEnableMarkdownInMessages
+  parseMessage = m => parse_text(m, markDownEnabled)
```
r10s, line 28: `parse_desktop_set()` can be removed alltogether, or? that would also show if there are more usages

r10s (approved):
makes sense, thanks a lot for taking care!

Simon-Laux (approved):

PR #4876, opened 2025-03-29
Expose open dialog IDs from DialogContext

### LINKED ISSUE #4865: "Esc" sometimes closes the wrong dialog
when going for a backup, and cancel the action with the "Esc" key, not the topmost dialog is closed, but the one that lies under the topmost.

only the next "Esc" then closes the topmost:

https://github.com/user-attachments/assets/4d8e9af1-726b-4c3e-8bd6-e55ed71d1c2e

to reproduce:

- open "Settings / Chats and Media / Export Backup"
- now there are three dialogs stacked on the screen: 1. Settings, 2. Chats and Media and 3. Export Backup
- hit "Esc" - and the 2. is closed - where it should close 3.

the issue may be present for other dialog stacks as well

(came over that when looking at https://support.delta.chat/t/how-do-i-delete-user-account-mailbox-on-self-installed-chatmail-server/3654/6 - and yes, the "OK" button should better read "Start Backup" as e.g. on iOS, i created a PR for that at https://github.com/deltachat/deltachat-desktop/pull/4866)

nicodh:
resolves #4865

Closing the "latest" dialog by calling closeDialog(latestDialogID) could be a generic solution to "always" close the last opened dialog. Just a consideration in case we have longer stacks of dialogs some day

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - prevent dragging around of webxdc icon #4740
 - tauri: clear temp folder on exit #4839
 - fix wrong punycode warnings in links #4864
+- fix esc key closing wrong dialog in settings #4865
 
 <a id="1_56_0"></a>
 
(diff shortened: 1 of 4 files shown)


Simon-Laux (approved):
hacky... but looks like it will work.

PR #4877, opened 2025-03-29
slight gradients for avatars for a more modern look

Simon-Laux:
- **gradient for color avatars**
- **add gradient to largeProfileImageArea and add type for it so ts doesn't complain**
- **add gradient to avatars in message view (author avatars)**

Basically makes the colors a bit less flat or harsh. Gives a slight impression of 3D.
Telegram also does this, though they have a much stronger gradient.

| Before | After |
|----|----|
| Bildschirmfoto 2025-03-29 um 20 44 11|Bildschirmfoto 2025-03-29 um 20 44 39 | 

Why did I do this? This is inspired by feedback from older people that complained about harsh colors. (they also asked for a different/softer nav bar color, but thats a different topic for a different pr and discussion)

## TODO:

- [x] flip the gradient.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - switch to account the webxdc is from when sending to chat (tauri and electron edition) #4740
 - change the Reply button for messages to be a verb rather than a noun #4853
 - Update message-parser to v0.13.0
+- slight gradients for avatars for a more modern look #4877
 
 ### Fixed
 - tauri: improve security #4826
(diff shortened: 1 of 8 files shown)


r10s:
while being sceptical first, to no introduce different avatars on different platforms (occupational illness :), the change is so subtile, that it is more a variation of the same avatar, so, yeah, we can go for it. although i never heard about complains, looks good, good job 👍

technically: the light source assumed for "3D" effects on "2D" UIs is assumed to be somewhere  the upper left corner - that means, the gradient should be flipped by 180°

nicodh (approved):
Code looks good to me

PR #4880, opened 2025-03-30
remove addr from create chat and add view profile to context menu

### LINKED ISSUE #4878: remove ADDR from contact lists, allow to open profile from "new chat"
the contact lists for "new chat" and the member list still show the email address, which we want to tune down for known reasons (multi transport, non-telling addresses etc)

instead, the should be able to open the profile easily, which has more appropriate information.

this is missing, however for "New Chat". so, to close this issue, the already existing context menu should get the option "View Profile" (string `menu_view_profile`)

on iOS we removed them at these places since half a year without issues. we left them, however, in contact has no green checkmark, this might be a good compromise for classic email usage for now

corresponding android issue: https://github.com/deltachat/deltachat-android/issues/3701

Simon-Laux:
- **remove addr from contact list items unless they are not verified.**
- **create chat: add context menu option to view profile**

closes #4878

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,14 @@
 
 ### Added
 - tauri: support for webxdc #4740, #4852
+- create chat: add context menu option to view profile #4880
 
 ### Changed
 - switch to account the webxdc is from when sending to chat (tauri and electron edition) #4740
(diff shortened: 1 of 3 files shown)


r10s (approved):
nice! code lgtm, and also tested by building :)

PR #4882, opened 2025-03-30
change usage of `nameAndAddr` to `displayName`

### LINKED ISSUE #4879: prefer `NAME` over `NAME (ADDR)`
replace the usage of getNameAndAddr() by "getDisplayname() unconditionally

as a rule of thumb getnameAndAddr() should no longer be used and he getAddr() only within the profile; we're not full there yet, but on the way :)

corresponding android PR with more information: https://github.com/deltachat/deltachat-android/pull/3699

Simon-Laux:
closes #4879

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - change the Reply button for messages to be a verb rather than a noun #4853
 - Update message-parser to v0.13.0
 - slight gradients for avatars for a more modern look #4877
+- change usage of `nameAndAddr` to `displayName` #4882
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


r10s (approved):
lgtm, code base, not sure if it is exhaustive.

to be sure, i think, we can also strike nameAndAddr from the bindings, we might deprecate that anyways

Simon-Laux:
> lgtm, code base, not sure if it is exhaustive.

did a search for the string

nicodh:
This PR breaks some of the e2e tests!

PR #4883, opened 2025-03-30
Add separators to the context menu

### LINKED ISSUE #4803: add menu separator
by the added functionality, it becomes more pressing to have sth. as a "separator" for the menus.

eg. the message context menu may have 10+ items, which is too much, ppl can recognize the position of 6~7 items, then things become cognitive overload.

the message's context menu could then look eg. like

```
Reply
Reply Privately
Forward
React
Edit
-----------------
Save
Save As
Copy Text
Copy Image
Resend
Info
-----------------
Delete
```

of course, non-matching items should stay hidden.

also the other menus can be adapted, but that can be done in subsequent PR once we have a separator ready. this issue is about having the option at all, and adapt message's context menu

Simon-Laux:
- **start with adding separators to context menu and fix some biome lints**
- **Add seperators to the context menu. styling and skipping of separators when using keybord navigation**

closes #4803

image


packages/frontend/scss/misc/_context_menu.scss
```diff
@@ -41,6 +41,15 @@
     0 2px 4px rgba(16, 22, 26, 0.2),
     0 8px 24px rgba(16, 22, 26, 0.2);
 
+  hr.seperator {
```
r10s, line 44: typo: separator

packages/frontend/scss/misc/_context_menu.scss
```diff
@@ -41,6 +41,15 @@
...
+    background-color: #8080804a;
+    height: 2px;
+    border: 0;
+    margin: 2px;
```
r10s, line 48: let's make the things a bit more subtle, as in many other apps and os - thinner line, more whitespace:

 ```suggestion
     height: 1px;
     border: 0;
     margin: 6px;
```

r10s (approved):
very cool! thanks a looooot, this makes things much more accessible!

wrt accessibilty: the separator is just ignored for screen readers, right?

only a small style suggestion and a type, otherwise that can be merged.

nicodh (approved):
Reviewed by reading & testing

Simon-Laux:
> wrt accessibilty: the separator is just ignored for screen readers, right?

I didn't test, but used hr to follow the semantic html rules that should make it accessible by default

Simon-Laux:
I discussed the ordering and separator positions with @r10s in a chat.

PR #4884, opened 2025-03-30
scroll to top when search query changes

### LINKED ISSUE #4776: search results should scroll to top on search query changes

Simon-Laux:
closes #4776


nicodh:
Bug #4776 is not clear:
which of the possible 3 result lists should scroll up? I see at least 2 (contacts & messages) not scrolling up with this PR

Simon-Laux:
the issue that @missytake raised in chat was about not seeing chat results, so this pr is only about chat results

nicodh (approved):
Reviewed by reading & testing

PR #4885, opened 2025-03-30
Avoid wrapping of reactions

### LINKED ISSUE #4322: Hide reactions instead of wrapping
Avoid overlay of reactions if there are too many to display them in one row

nicodh:
resolves #4322

based on #4872

https://github.com/user-attachments/assets/2d337085-55af-40c6-a4d3-12a28bb4ebc8


r10s (approved):
i'd use same font size and the +counter should show the number of _other ppl who have reacted_ (not: emoji), otherwise lgtm

PR #4886, opened 2025-03-30
Share contact from profile dialog

### LINKED ISSUE #4833: Make it possible to Share a contact from the contact profile
Delta Chat on Android has the ability to directly Share the contact from the three-dot menu of the contact profile, but on desktop there is no such option so users copy-paste the email address instead:
https://support.delta.chat/t/design-study-in-sharing-dc-contacts/3643

Simon-Laux:
- **lint fixes**
- **add button to share contact from profile view dialog**

closes #4833

Bildschirmfoto 2025-03-30 um 22 10 56

And on macOS is uses the share icon apple users are more familiar with.
Bildschirmfoto 2025-03-30 um 22 06 29


nicodh:
Took me quite some time to "find" the share icon.
Maybe place it right besides the Avatar/name? Would be more intuitive imo

share

Simon-Laux:
The suggestion in the issue was in the 3 dot menu, I put it as extra icon for more visibility.
I think even more "intuitive" would be a share button next to the send message button.

nicodh (approved):
Since there will be some more changes in the profile view it's fine to merge this now as it is and the UI redesign will follow very soon

PR #4887, opened 2025-03-30
Use user.name instead of user.address

nicodh:
#skip-changelog


packages/e2e-tests/tests/basic-functionality.spec.ts
```diff
@@ -330,7 +330,7 @@ test('create group', async ({ page, context, browserName }) => {
   await page.getByTestId('confirm-join-group').getByTestId('confirm').click()
   // userA invited you to group message
   await expect(page.locator('#message-list li').nth(1)).toContainText(
-    userA.address
```
Simon-Laux, line 333: this seems to fail. does it make sense to check for the info message here that is supplied by core?

nicodh: I think the whole issue avoid emails "anywhere" except on a special transports view, needs some more work on the UI and in the e2e tests

Simon-Laux (approved):

PR #4888, opened 2025-03-30
migrate account mute state to new is_muted config option

### LINKED ISSUE #4024: migrate account mute state to new is_muted config option
- [x] save state in `is_muted` instead of `ui.desktop.muted`
- [x] migration: on startup loop though all accounts and set `is_muted` to  `ui.desktop.muted` if set. then unset/delete `ui.desktop.muted`
- [x] update documentation https://github.com/deltachat/interface/blob/main/implementation/known_ui_config.md (probably say  that it was previously saved in ui.desktop.muted, but desktop does only use that until version (current released version in changelog at time of addressing this)) -> https://github.com/deltachat/interface/pull/81

Simon-Laux:
closes #4024

### DIFF
--- a/packages/frontend/src/stores/accountNotifications.ts
+++ b/packages/frontend/src/stores/accountNotifications.ts
@@ -10,7 +10,8 @@ import { Store, useStore } from './store'
  *
  * `"1"` means muted, anything else is interpreted as not muted
  */
-const UI_CONFIG_DESKTOP_MUTED = 'ui.desktop.muted'
+const UI_CONFIG_MUTED = 'ui.id_muted'
+const UI_CONFIG_DESKTOP_MUTED_OLD = 'ui.desktop.muted'
 
 interface AccountNotificationState {
   muted: boolean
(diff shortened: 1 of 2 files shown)


nicodh (approved):
Reviewed by reading & testing

packages/frontend/src/stores/accountNotifications.ts
```diff
@@ -10,7 +10,8 @@ import { Store, useStore } from './store'
  *
  * `"1"` means muted, anything else is interpreted as not muted
  */
-const UI_CONFIG_DESKTOP_MUTED = 'ui.desktop.muted'
+const UI_CONFIG_MUTED = 'ui.id_muted'
```
Simon-Laux, line 13: found an issue

Simon-Laux: made a fix: https://github.com/deltachat/deltachat-desktop/pull/4924

PR #4889, opened 2025-03-31
i18n: fix wrong order of substitutions for some strings

### LINKED ISSUE #3897: i18n: wrong order of substitutions for some strings
- Delta Chat Version: 1.45
- Expected behavior: a string [`%2$s가 %1$s를 멤버로 추가함`](https://github.com/deltachat/deltachat-desktop/blob/3e8f73c0bd44945fd0dce8341d55b2a1c400e568/_locales/ko.xml#L688) is substituted as `THE ADDED 가 THE ADDER를 멤버로 추가함`
- Actual behavior: it is substituted as `THE ADDER 가 THE ADDED를 멤버로 추가함`, which is the wrong way around.
    1. Switch to Korean language (ko).
    2. Be in the group where someone adds someone else.
    3. See the message that is printed when that happens.
- Screenshots:
- Logs:

How to find (some) strings that are wrong: search for `%2.*%1`. There are 94 occurrences for 16 languages currently.

The responsible code:

https://github.com/deltachat/deltachat-desktop/blob/b342a1d47b505e68caaec71f79c381c3f304405a/src/shared/localize.ts#L73-L83

It assumes that the substitutions are always in the same order for all strings for all languages.
And if you feel like "it's just a matter of looking at the digits after `%`, keep in mind that some strings only have one substitution, and some strings have several substitutions, but they are not numbered (such as [`ask_send_following_n_files_to`](https://github.com/deltachat/deltachat-desktop/blob/0049dc4bfb0f13a8848e613e4d96026bf5381354/_locales/en.xml#L312-L315)

It feels like it's about time to use a library for this. Or write one, and cover it with tests. Or convert the original xml files to a better-supported format with a tried-and-tested converter. We already have [code that converts XML to JSON](https://github.com/deltachat/deltachat-desktop/blob/bf11419376a1263b62699c770f6ed7b6c12a059c/bin/build-shared-convert-translations.mjs#L71).

Also as a part of this I guess this

https://github.com/deltachat/deltachat-desktop/blob/dd8361e65e35529056f5bb2a343a593a1a2f5586/src/renderer/components/dialogs/ConfirmSendingFiles.tsx#L48

needs to be fixed to support non-English plural rules.

Simon-Laux:
closes #3897

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@
 - fix wrong punycode warnings in links #4864
 - scroll to top when search query changes
 - fix esc key closing wrong dialog in settings #4865
-- i18n: fix wrong order of substitutions for some strings
+- i18n: fix wrong order of substitutions for some strings #4889
 
 <a id="1_56_0"></a>
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 - fix wrong punycode warnings in links #4864
 - scroll to top when search query changes
 - fix esc key closing wrong dialog in settings #4865
+- i18n: fix wrong order of substitutions for some strings
 
 <a id="1_56_0"></a>
 
--- a/packages/shared/localize.ts
+++ b/packages/shared/localize.ts
@@ -116,7 +116,9 @@ export function translate(
         substitutions = [substitutions]
       }
 
+      let counter = -1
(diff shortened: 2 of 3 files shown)


nicodh (approved):
Reviewed by reading!

Nice that you also added an appropriate test!

PR #4890, opened 2025-03-31
fix double escape bypasses dialog attribute `canEscapeKeyClose={false}`

### LINKED ISSUE #4397: Check where Dialog closed by double escape should be handled
Followup to #4365 

It turns out that you can close any Dialog in Chrome by firing excape key twice. We should make sure that this causes no unexpected behaviour. Probably only Dialogs with canEscapeKeyClose={false} are affected

Simon-Laux:
closes #4397


nicodh (approved):
Reviewed by testing & reading

PR #4892, opened 2025-03-31
focuses first visible item on arrow down key on input in create chat dialog

### LINKED ISSUE #1450: New chat dialog
- Platform (linux/mac/windows): macOS 10.15.3
- Delta Chat Desktop Version: 0.999.1

In the "new chat" dialog the user can enter a few characters as live search, and the hit list is shortened to the entries containing these characters. That’s good. But on macOS the expectation of users would be, that after entering a few characters, the user could select the first entry of the list by pressing "Cursor down", and subsequently select others with "Cursor up" / "Cursor down"
Bildschirmfoto 2020-02-12 um 23 03 23

Simon-Laux:
closes #1450


packages/frontend/src/components/dialogs/CreateChat/index.tsx
```diff
@@ -258,6 +258,17 @@ function CreateChatMain(props: CreateChatMainProps) {
 
   const fixedSizeListOuterRef = useRef<HTMLElement>(null)
 
+  const onKeyDown = (ev: React.KeyboardEvent) => {
+    if (ev.code === 'ArrowDown') {
+      ;(
+        fixedSizeListOuterRef.current?.firstElementChild?.firstElementChild
+          ?.firstElementChild?.firstElementChild as HTMLElement
```
nicodh, line 265: Not sure if that is a reliable path. Maybe @WofWca have a look at it, since he introduced the fixedSizeListOuterRef

Simon-Laux: selects the first visible/loaded item, we could probably scroll up before to make this select the first item?

WofWca: Yeah, this doesn't look too good. `querySelector('button')` or alike would be better I'd say, similar to here

https://github.com/deltachat/deltachat-desktop/blob/bbd353d68b83ea537e6f6735a5abb6e54c4a6b0c/packages/frontend/src/components/chat/ChatList.tsx#L294-L303

WofWca (commented):
This should work.
However, it's fragile. Adding a test, in addition to writing a more resilient selector, would help greatly.

Also we would probably want to add this behavior to the rest of such dialogs (e.g. "add members", etc).

packages/e2e-tests/tests/basic-functionality.spec.ts
```diff
@@ -379,6 +379,22 @@ test('add app from picker to chat', async ({ page }) => {
   await expect(webxdcMessage).toContainText(appName)
 })
 
+/**
+ * focuses first visible item on arrow down key on input in create chat dialog
+ */
```
WofWca, line 384:
```suggestion
```

IDK, it's just duplicated.

packages/e2e-tests/tests/basic-functionality.spec.ts
```diff
@@ -379,6 +379,22 @@ test('add app from picker to chat', async ({ page }) => {
...
+  page,
+}) => {
+  const userA = getUser(0)
+  await switchToProfile(page, userA.id)
+  await page.locator('#new-chat-button').click()
+  await page.locator('dialog *:focus').waitFor({ state: 'visible' })
+  await page.locator('*:focus').press('ArrowDown')
+
+  // check if moved the focus down
+  expect(page.locator('*:focus')).toContainText('New Contact')
```
WofWca, line 395:
```suggestion
  await expect(page.locator('*:focus')).toContainText('New Contact')
```

Hmmm, didn't your IDE show the warning about this?

WofWca (approved):

PR #4893, opened 2025-03-31
add support for invite links to create chat dialog

### LINKED ISSUE #4158: [UX] Entering some sort of join link or room id into the search after pressing the "+" button doesn't offer to let me join

Simon-Laux:
closes #4158


nicodh (changes requested):
Works as expected but the Create Chat dialog stays open so it's hard to see that the new chat was created

PR #4895, opened 2025-03-31
fix order when sending multiple files at once

### LINKED ISSUE #4729: enqueue multiple files/images for sending sequentially to preserve order
When sending more than 2 files or images at once, they are displayed in a different order to the recipient than to the sender. If a user sends multiple files from DC-Android to a user with DC-Desktop, the displayed order is the same for both. Therefore, I suspect that this is not a core bug.

macOS 15.3.1
DC-Desktop 1.54

Simon-Laux:
fix #4729


nicodh (approved):
Reviewed by reading

PR #4897, opened 2025-04-01
Redesign profile view

nicodh:
New profile:
image

Dark mode blocked contact:

image

Group
image


packages/frontend/src/components/dialogs/ViewProfile/styles.module.scss
```diff
@@ -15,3 +15,17 @@
...
+  margin: 20px 0px;
+  justify-content: center;
+  gap: 30px;
+  button {
+    min-width: 150px;
+  }
+}
+
+.addressLine {
+  color: #666;
```
WofWca, line 30: Maybe it's just me, but I think people will be wondering why the address is darkened.

r10s (approved):
very, very excellent!

reviewed by testing:

- the email address should get the same font and icon color as the other items; this results in a cleaner look, and does not hilite the address more (even though it is grey, it is hilited as looking different)

- the avatar focus is a little bit too spacy:
Screenshot 2025-04-01 at 15 44 04

i did not thought much about changing the group profile at the same time, but probably okay, at least other messengers have similar  styles for group and contacts, so let's stay with that as well

r10s:
ah: the "unblock contact" button should not be primary as well. it is probably not the most common thing one wants to do when viewing your stalker's profile

WofWca (commented):
I like the avatar and profile name change (perhaps more space is needed between the avatar and the name), but the "share" button seems really oddly placed. I liked its original place more.

The code looks fine otherwise.

packages/frontend/src/components/dialogs/ViewProfile/styles.module.scss
r10s: yip, it should just have the same style as the other two lines

r10s:
> but the "share" button seems really oddly placed. I liked its original place more.

it is fine, one will get used to it :)

it looks more sorted that way, having the three dot menu in the middle of sth is also weird (but other positions would have been even weirder)

however, reason for a text button is, that the "share" icon is not really known on desktops. it is better to use the term "share" here, this is even what eg. macOS is doing on many places.

Simon-Laux:
I'm fine with the text share button and would also have said the thing about whitespace.

nicodh:
> * the avatar focus is a little bit too spacy:

Fixed!

PR #4899, opened 2025-04-01
Update to core 1.158.0

nicodh:


WofWca (approved):
FYI this also upgrades esbuild.

Otherwise: just an upgrade.

Simon-Laux (approved):

PR #4908, opened 2025-04-01
tauri: add cli interface `--help`, `--version`, and developer options (like `--dev-mode`)

Simon-Laux:
- [x] measure size increase
- [x] fix git version: see https://github.com/deltachat/deltachat-desktop/pull/4909

### DIFF
--- a/packages/target-tauri/src-tauri/src/lib.rs
+++ b/packages/target-tauri/src-tauri/src/lib.rs
@@ -13,10 +13,12 @@ use state::{
     translations::TranslationState, webxdc_instances::WebxdcInstancesState,
 };
 use tauri::Manager;
-use util::csp::add_custom_schemes_to_csp_for_window_and_android;
+use tauri_plugin_log::{Target, TargetKind};
 
 mod app_path;
 mod blobs;
+#[cfg(desktop)]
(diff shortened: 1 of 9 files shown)


packages/target-tauri/src-tauri/src/lib.rs
```diff
@@ -86,6 +89,11 @@ fn get_current_logfile(state: tauri::State<AppState>) -> String {
 pub fn run() {
     let startup_timestamp = SystemTime::now();
 
+    #[cfg(desktop)]
+    let run_config = cli::parse_cli_options();
```
WofWca, line 93: Do we want this in production? This pulls in `chrono` and `clap` dependencies. Did you check the build size change?

WofWca (commented):
The code looks good. Didn't test.
Though I wonder if we need this in prod.
Also `clippy` is mad.

packages/target-tauri/src-tauri/src/lib.rs
Simon-Laux: size increase is not even a mb, maybe ~170kb

WofWca (approved):

PR #4909, opened 2025-04-02
Add necessary second parameter to git describe

nicodh:
see #4649

#skip-changelog


WofWca (approved):
Did not test, looks fine.

PR #4915, opened 2025-04-02
tauri: perf: `Arc::clone` instead of `window.clone()`

WofWca:
I did not measure the performance impact,
but it should be a tiny bit better.

#skip-changelog because this is minor.

This is mostly just me practicing / learning Rust, so feel free to close if you don't want to bother.

### DIFF
--- a/packages/target-tauri/src-tauri/src/help_window.rs
+++ b/packages/target-tauri/src-tauri/src/help_window.rs
@@ -1,3 +1,5 @@
+use std::sync::Arc;
+
 use log::{error, warn};
 use tauri::{Manager, State, WebviewWindow};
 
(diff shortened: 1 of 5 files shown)


Simon-Laux (changes requested):
contains a comment and clippy has some complaints.
Otherwise I'm neither for nor against this change. Though I think this has no measurable benefit for us as this is not about IO and not about code paths that are triggered very often.

packages/target-tauri/src-tauri/src/help_window.rs
```diff
@@ -93,12 +101,13 @@ pub(crate) async fn open_help_window(
...
-        let help_window_clone = help_window.clone();
+        let help_window_arc = Arc::new(help_window);
+        let help_window_clone = Arc::clone(&help_window_arc);
         menu_manager
             .register_window(
                 &app,
-                &help_window,
-                Box::new(move |app| create_help_menu(app, &help_window_clone)),
+                &*help_window_arc,
+                Box::new(move |app| create_help_menu(app, &*help_window_clone)),
```
WofWca, line 110: OK, obeying the clippy warning `warning: deref which would be done by auto-deref` causes a compile error because our `WindowAbstraction` doesn't implement for `Arc<...`. I'm not sure what's the best way to fix this, besides adding `impl ... for Arc`.

WofWca:
OK, let's close then.
Thanks for taking a look!

PR #4916, opened 2025-04-02
tauri: perf: webxdc: `spawn(set_title)`

WofWca:
This is based on another MR, so only look at the last commit.

`set_title` needs to perform DB calls, so let's move it
into a separate task.

However, based on my testing, this did not produce a significant
webxdc startup speed increase.  
Thus, feel free to reject this.

#skip-changelog because this is insignificant.

### DIFF
--- a/packages/target-tauri/src-tauri/src/help_window.rs
+++ b/packages/target-tauri/src-tauri/src/help_window.rs
@@ -1,3 +1,5 @@
+use std::sync::Arc;
+
 use log::{error, warn};
 use tauri::{Manager, State, WebviewWindow};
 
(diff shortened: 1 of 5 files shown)


Simon-Laux:
> This is based on another MR, so only look at the last commit.

I would prefer if this pr would take that PR as base then, so it's easier to review.

WofWca:
> I would prefer if this pr would take that PR as base then, so it's easier to review.

Hm I was thinking that this is not great because it would be "a request to merge into another MR", but it's easy to switch the branch back I guess.

> But don't do two closures, just do one inside of

Hmm I did this because I wanted the `?` operator. I'll see if it's possible to use it in a task next time.

OK, let's close then, since it complicates the code a little.

Simon-Laux:
> Hmm I did this because I wanted the ? operator. I'll see if it's possible to use it in a task next time.

I have a working workaround in the blobs schema, you can look there for reference.

WofWca:
That is what I used as a reference, except I made it a closure instead of an async block, and I moved its declaration outside of `spawn`.

PR #4920, opened 2025-04-03
Feature voice message as mp3

nicodh:
This is a workaround as long as iOS does not support the weba codec and as long as we don't have some audio converter in core.


packages/frontend/src/components/composer/Composer.tsx
```diff
@@ -129,6 +132,25 @@ const Composer = forwardRef<
...
+    reader.readAsDataURL(voiceData)
+    reader.onloadend = async () => {
+      if (!reader.result) {
+        log.error('Cannot convert blob to base64. reader.result is null')
+        return
+      }
+      const b64 = reader.result.toString().split(',')[1]
+      const filename = b64.slice(3, 12).toString() + '.mp3'
+      const path = await runtime.writeTempFileFromBase64(filename, b64)
+      currentComposerMessageInputRef.current?.setState({ loadingDraft: false })
```
nicodh, line 146: This should not be necessary but at the time saveVoiceAsDraft is called the currentComposerMessageInputRef.state is still on "loadingDraft: true" and addFileToDraft fails ?

packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
```diff
@@ -0,0 +1,96 @@
...
+}: {
+  recording: boolean
+  setRecording: (f: boolean) => void
+  saveVoiceAsDraft: (blob: Blob) => void
+}) => {
+  const recorder = useRef<MicRecorderType>(
+    new MicRecorder({
+      bitRate: 128,
+    })
+  )
```
WofWca, line 24: Again, this will re-instantiate the `MicRecorder` on every re-render. I don't know how expensive it is.

packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
```diff
@@ -0,0 +1,96 @@
...
+    } else {
+      setRecordingTime(0)
+    }
+
+    return () => {
+      if (timer) {
+        window.clearInterval(timer)
+      }
+    }
+  }, [recording])
```
WofWca, line 49: I think this can be simplified by writing down the start time of recording, and then just forcing a re-render every second and recomputing the duration.
But this is also fine I guess.

WofWca (commented):
Haven't looked thoroughly yet.

packages/frontend/src/components/AudioRecorder/Encoder.ts
```diff
@@ -0,0 +1,106 @@
+import { Mp3Encoder } from '@breezystack/lamejs'
+
+interface EncoderConfig {
+  sampleRate?: number
+  bitRate?: number
+}
+
+class Encoder {
```
WofWca, line 8: You mentioned that this is a forked version of https://github.com/closeio/mic-recorder-to-mp3 ? We need to comply with its license. Putting a top-level JSDoc comment in each file with `@license` and the text of the original license notice should work.  
And in general I think this should be made clear that this is just copy-pasted, so it should be treated as a dependency, until we actually separate it out into a dependency.

packages/frontend/src/components/composer/Composer.tsx
```diff
@@ -259,6 +285,15 @@ const Composer = forwardRef<
...
     }
   }, [shiftPressed])
+
+  useEffect(() => {
+    if (draftState.file || draftState.text) {
+      setVoiceMessageDisabled(true)
+    } else {
+      setVoiceMessageDisabled(false)
+    }
+  }, [draftState])
```
WofWca, line 295: This doesn't look like it has to be an effect.

packages/frontend/src/components/composer/ComposerMessageInput.tsx
```diff
@@ -118,6 +118,10 @@ export default class ComposerMessageInput extends React.Component<
     return this.state.text
   }
 
+  hasText(): boolean {
+    return !this.getText().match(/^\s*$/)
```
Simon-Laux, line 122: https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/String/trim and then checking length might be easier to understand code?

packages/frontend/src/components/AudioRecorder/Encoder.ts
nicodh: This should be sufficient?

packages/frontend/src/components/composer/Composer.tsx
```diff
@@ -129,6 +132,25 @@ const Composer = forwardRef<
...
+  const saveVoiceAsDraft = (voiceData: Blob) => {
+    const reader = new FileReader()
+    reader.readAsDataURL(voiceData)
+    reader.onloadend = async () => {
+      if (!reader.result) {
+        log.error('Cannot convert blob to base64. reader.result is null')
+        return
+      }
+      const b64 = reader.result.toString().split(',')[1]
+      const filename = b64.slice(3, 12).toString() + '.mp3'
```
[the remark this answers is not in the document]
nicodh: It was 0, 12 before, which was just a kind of a hash. But I found it starting with 2 slashes so I removed them.
This line was still there from Farooqs PR. 
Maybe we can just generate a real hash or a timestamp...

packages/frontend/src/components/composer/ComposerMessageInput.tsx
nicodh: I just moved the existing regex, so the regex was not introduced with this PR

packages/frontend/src/components/composer/Composer.tsx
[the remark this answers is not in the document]
nicodh: Uses a different simple hash now

r10s:
very nice to push that forward! tested a bit - voice recording and sending works already nicely, also for iOS - thanks a lot to give us more time to adapt iOS to ogg, and to remove the pressure 🙏

some UI things that come to my mind:

- i like the idea of the pulsating record icon

- the time display during recording should use the normal font (as used in message input) and there should not be spaces before/after colon

- it is not so nice that on normal message input, on entering the first letter, that shifts to the left now (as the record button disappears). that appears restless and unstable.
all of whatsapp/telegram/signal have the record button right of the input field - i think, we should go for that as well.
this also fits nicely with the send button only shown when text is entered:
we can have less moving parts now, by just replacing record-button by send-button - which both should have the exact same size

- the pulsating audio and the time display during recording can stay where they are, on the left

- for the "stop" button is a bit unclear what happens - it is a "stages" things, but that is a bit hard to explain. maybe just call it "ok" - and add a "cancel" button at the same time

nicodh:
Added ok & cancel buttons

image

r10s:
very nice!

- i would swap ok / cancel (at similar places, "ok" is rightmost)

- ok should not have the destructive "red", let's just use the existing primary and secondary buttons and not invent sth new

a bit nitpicky maybe, but it gives a bad impression when _not_ using voice messages but enter some text:

- there is a nervous flickering when the record-voice-button is swapped with the second-button are swapped, one even sees all three partly

https://github.com/user-attachments/assets/89813dcd-df95-41dd-96e7-436eadba11c7

as this is a very often used UI element, i think, it makes sense to check if that part can be improved easily, to avoid, again, a "restless and unstable" first impression

_locales/_untranslated_en.json
```diff
@@ -59,5 +59,11 @@
   },
   "reset_encryption_confirm": {
     "message": "Forget encryption key for this contact."
+  },
+  "voice_send": {
+    "message": "Send a voice message"
+  },
+  "voice_send_cannot": {
```
Simon-Laux, line 66: this seems unused?

Simon-Laux (changes requested):
does not work on macOS (neither in electron nor tauri target work, browser version in firefox works after accepting audio permissions)

r10s:
> does not work on macOS (neither electron nor tauri target works, browser version in firefox works after accepting audio permissions)

works for me on macOS, just double checked. i did nothing special, however, when i was asked for permission, i hit "ok" (cannot remember details, i was playing around  with 5 different apps, all quering permissions :))

Simon-Laux:
Ok works with a built/packaged version of the app, there it asks for permission. Still we need to tell the user somehow when the permission is not granted.

WofWca:
Will it show the permission prompt again if you don't grant it the first time?

Simon-Laux:
> Will it show the permission prompt again if you don't grant it the first time?

not on MacOS

packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  const onRecordingStart = () => {
+    setRecording(true)
+    recorder.current
+      ?.start()
+      .then(() => {
+        // something else
+      })
+      .catch((err: any) => {
+        console.error(err)
```
Simon-Laux, line 57: The error should be shown to the user and the recording should not look like it works

nicodh:
I added error handling to show errors to the user. And show an error if no audio signal is received (like it can happen if you mute the mic in the system settings etc.).
Also added a kind of volume meter to show the level of the incoming sound (tbd)

image


https://github.com/user-attachments/assets/894898a0-aba9-4a9f-bc80-621773d9829e

packages/frontend/src/components/AudioRecorder/MicRecorder.ts
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  /**
+   * Return the volume of the microphone
+   * @return {number}
+   */
+  calculateVolume(input: any) {
+    let i
+    let sum = 0.0
+    for (i = 0; i < input.length; ++i) {
```
WofWca, line 170:
```suggestion
    let sum = 0.0
    for (let i = 0; i < input.length; ++i) {
```

packages/frontend/src/components/AudioRecorder/styles.module.scss
```diff
@@ -0,0 +1,132 @@
+.microphoneButton {
+  height: 40px;
+  width: 40px;
+  padding-left: 2px;
```
WofWca, line 4:
```suggestion
  padding-right: 3px;
```

image
image

This centers the mic icon a little better relative to the "send" button.

Compared to "before":

image

WofWca: Sorry, I updated the comment. Turns out padding-right was not it.

packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
```diff
@@ -0,0 +1,161 @@
...
+      .then(([buffer, _blob]) => {
+        saveVoiceAsDraft(new Blob(buffer, { type: 'audio/mp3' }))
+      })
+      .catch((err: any) => {
+        console.error(err)
+        onError(new AudioRecorderError(err.message))
+      })
+  }
+
+  const onRecordingCancel = () => {
```
WofWca, line 124: If you cancel recording, the composer textarea becomes disabled.

packages/frontend/src/components/AudioRecorder/styles.module.scss
```diff
@@ -0,0 +1,132 @@
...
+  padding-bottom: 8px;
+  background-color: var(--composerBg);
+  border: none;
+
+  &:focus {
+    outline: none;
+  }
```
WofWca, line 11:
```suggestion
```

This is bad for accessibility. Why is it needed? We already have `:focus-visible` selector for all buttons

https://github.com/deltachat/deltachat-desktop/blob/e5d32b6a6736d144442c8371339247dcf013877b/packages/frontend/scss/_global.scss#L84-L86

packages/frontend/src/components/AudioRecorder/MicRecorder.ts
```diff
@@ -0,0 +1,180 @@
+/**
+ * This file is based on https://github.com/closeio/mic-recorder-to-mp3/blob/master/src/mic-recorder.js
+ *
+ * @license MIT Copyright (c) 2017 Elastic Inc. (Close.io) see ./LICENSE
```
WofWca, line 4: The LICENSE file is empty, I assume it's not intentional?

packages/frontend/src/components/AudioRecorder/styles.module.scss
```diff
@@ -0,0 +1,132 @@
...
+    height: 12px;
+    position: absolute;
+    left: 0;
+
+    .step {
+      display: flex;
+      width: 10%;
+      height: 10px;
+      border: 1px solid #666;
+    }
```
r10s, line 107: the line between the segments have the double thickness, which does not look good.

also, the lines better pick up a shading that exist and is more decent, to fit into the rest of the bar.

```suggestion
      border-left: 1px solid #bbb;
      border-top: 1px solid #bbb;
      border-bottom: 1px solid #bbb;
    }

    .step:last-child {
      border-right: 1px solid #bbb;
    }
```

^^ maybe that can be done in a smarter way, but this does the job, is easy to understand, and keeps things together

r10s (commented):
very nice!

let's keep "audio controls" and "message sending controls" together, that means, move the buttons right, and the level meter to time and pulsating mic:

Screenshot 2025-04-07 at 13 28 52

packages/frontend/src/components/AudioRecorder/MicRecorder.ts
nicodh: Right, it wasn't

r10s:
ah, another things that needs to be targeted before it can be released:

you can switch to another chat while recording, and the bar will stay as is. the message then is sent to whatever chat was selected last, this is quite unexpected ...

i suggest to stop recording and stage the recording when switching chats (this is what eg. signal  is doing)

packages/frontend/src/components/AudioRecorder/Encoder.ts
```diff
@@ -0,0 +1,112 @@
...
+  private config: Required<EncoderConfig>
+  private mp3Encoder: Mp3Encoder
+  private maxSamples: number
+  private samplesMono: Int16Array | null
+  private dataBuffer: Int8Array[] = []
+
+  constructor(config: EncoderConfig = {}) {
+    this.config = {
+      sampleRate: 44100,
+      bitRate: 128,
```
r10s, line 24: this results in unexpectedly high data usage. until we can switch to opus, i suggest to half both parameters, for voice that is usually good enough

```suggestion
      sampleRate: 22050,
      bitRate: 64,
```

this should result in files being only 1/4 of size ([older considerations](https://github.com/deltachat/deltachat-android/issues/535#issuecomment-451418674))

moreover, we should respect the "media_quality" setting, but that can be done in another PR to not complicate things here further

WofWca: If we're into downsampling, for speech 8000 would be enough to make it understandable, but yeah, that is gonna be phone-level quality.

https://en.wikipedia.org/wiki/Sampling_(signal_processing)#Speech_sampling

For reference, in Telegram a one-minute voice message takes up ~230KB.

packages/frontend/src/components/AudioRecorder/styles.module.scss
nicodh: Done

packages/frontend/src/components/AudioRecorder/Encoder.ts
nicodh: I played around with various values but it didn't had any impact on the size of the recorded audio. It's always ~484kB for 30 seconds.

_locales/_untranslated_en.json
nicodh: It's used (now) as aria-label for the microphone button

packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
nicodh: It's related again to the loadingDraft in state. I fixed that with commit [48a17](https://github.com/deltachat/deltachat-desktop/pull/4920/commits/48a1709936cfbad47379441a9e2f3a859409485f)

packages/frontend/src/components/AudioRecorder/Encoder.ts
r10s: hm, then there is a bug somewhere, maybe in the used library, ~1 mb/minute is the standard rate mp3 has as 128 kbit, 44.1 kHz 

i think, it should not block merging this PR, but we might want to file an issue for that

Simon-Laux:
About colors in the volume meter: many apps such as discord, mumble (in first setup), audacity and professional Digital Audio Workstations have different colors for ranges, like green, yellow and red (red is if you are too loud so that it clips/overloads).

https://github.com/user-attachments/assets/043e0fd6-2351-4b75-8351-01d39f59b2cc
https://github.com/user-attachments/assets/1928827d-054c-4ecd-bc5a-9c897b599765

Just an idea, I'm not saying that we need it, especially not now, I think it is fine as is (with button placement that r10s posted). maybe we find a more intuitive solution in the future like a waveform.

nicodh:
> About colors in the volume meter: many apps such as discord, mumble (in first setup), audacity and professional Digital Audio Workstations have different colors for ranges, like green, yellow and red (red is if you are too loud so that it clips/overloads).

We have that too:
image

You can see it, if you put the audio input to a maximum in your system settings

Only question is: it's hard to detect the "real" limit where clipping starts...

I set the values here: https://github.com/deltachat/deltachat-desktop/pull/4920/files#diff-769f75b81cc687f60e9af48e627fdb8efe8608b32f185df559644c374358998dR131

nicodh:
> ah, another things that needs to be targeted before it can be released:
> 
> you can switch to another chat while recording, and the bar will stay as is. the message then is sent to whatever chat was selected last, this is quite unexpected ...
> 
> i suggest to stop recording and stage the recording when switching chats (this is what eg. signal is doing)

I fixed it now lilke this: when you switch chats without stopping the recording it will be lost. If you stop recording before switching your record will be saved as draft. That should be sufficient for now.

packages/frontend/src/components/AudioRecorder/Encoder.ts
nicodh: I will open a new issue when this PR is merged

packages/frontend/src/components/composer/Composer.tsx
```diff
@@ -129,6 +138,41 @@ const Composer = forwardRef<
...
+      const path = await runtime.writeTempFileFromBase64(filename, b64)
+      addFileToDraft(path, basename(path), 'Voice').catch((reason: any) => {
+        log.error('Cannot send message:', reason)
+        // show some error dialogue
+      })
+    }
+  }
+
+  const onAudioError = (err: AudioRecorderError) => {
+    console.error('onAudioError', err)
```
WofWca, line 165: Do we want to violate our "no console log" rule in this MR?

packages/frontend/src/components/composer/Composer.tsx
```diff
@@ -129,6 +138,41 @@ const Composer = forwardRef<
...
+    reader.onloadend = async () => {
+      if (!reader.result) {
+        log.error('Cannot convert blob to base64. reader.result is null')
+        return
+      }
+      const b64 = reader.result.toString().split(',')[1]
+      // random filename
+      const filename = Math.random().toString(36).substring(2, 10) + '.mp3'
+      const path = await runtime.writeTempFileFromBase64(filename, b64)
+      addFileToDraft(path, basename(path), 'Voice').catch((reason: any) => {
```
WofWca, line 157: `basename` appears to do nothing on Windows. It uses Browserify polyfill. Why don't we use `filename` here?

packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
```diff
@@ -0,0 +1,164 @@
...
+    )
+  } else {
+    return (
+      <div className={styles.audioRecorder}>
+        <button
+          className={styles.microphoneButton}
+          onClick={() => onRecordingStop()}
+        >
+          <span />
+        </button>
```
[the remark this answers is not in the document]
nicodh: Well, since it's a button and before had the "start recording" action I thought it maks sense to add an action here too. But most people will probably click on the ok button to stop the recording

packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
```diff
@@ -0,0 +1,164 @@
...
+        <span />
+      </button>
+    )
+  } else {
+    return (
+      <div className={styles.audioRecorder}>
+        <button
+          className={styles.microphoneButton}
+          onClick={() => onRecordingStop()}
+        >
```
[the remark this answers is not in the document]
nicodh: we need a new label then. Maybe:
aria-label={tx('stop_recording')}

Will add that to the Ok button too

packages/frontend/src/components/composer/Composer.tsx
```diff
@@ -745,7 +804,10 @@ export function useDraft(
     if (chatId === null || !canSend) {
       return
     }
-    if (inputRef.current?.textareaRef.current?.disabled) {
+    if (
+      inputRef.current?.textareaRef.current?.disabled &&
+      draftRef.current.viewType !== 'Voice'
```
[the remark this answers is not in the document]
nicodh: Not needed anymore after the draftState fix

packages/frontend/scss/composer/_composer.scss
```diff
@@ -57,14 +57,17 @@
...
-    .emoji-button {
+    .emoji-button,
+    .microphone-button {
       background: none;
       border: none;
     }
 
     .attachment-button {
-      width: 40px;
+      width: 30px;
```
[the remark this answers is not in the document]
nicodh: Thats a left over when the microphone button was left near the attachment button. Will change it

packages/frontend/src/components/composer/Composer.tsx
```diff
@@ -157,10 +201,13 @@ const Composer = forwardRef<
         if (chatId === null) {
           throw new Error('chat id is undefined')
         }
-        if (!regularMessageInputRef.current) {
+        if (
+          !regularMessageInputRef.current &&
+          !(draftState.file && draftState.viewType === 'Voice')
```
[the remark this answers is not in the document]
nicodh: Not needed anymore

WofWca (approved):
This feels pretty solid overall.
I left a couple more non-critical comments, but I won't insist on addressing them in this MR. I'll also create an issue with further improvement ideas.

IMO this can be merged at the current state.

Thanks for finally bringing such a demanded feature to life!

packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
```diff
@@ -0,0 +1,174 @@
...
+ * 3. a coloured background with a color gradient from green to red
+ */
+const VolumeMeter = (prop: { volume: number }) => {
+  const steps = 10
+  const volumeBarRef = useRef<HTMLDivElement>(null)
+  const totalWidth = Number(volumeBarRef.current?.clientWidth) || 0
+  // doubling the volume shows a more realistic volume level
+  const level = Math.min(totalWidth * prop.volume * 2, totalWidth)
+  const levelWidth = totalWidth > 0 ? `${totalWidth - level}px` : '100%'
+  console.debug(`volume: ${(prop.volume * 100).toFixed(2)}%`)
```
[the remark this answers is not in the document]
nicodh: Removed

PR #4924, opened 2025-04-03
fix typo in CONFIG_MUTED

Simon-Laux:
I did not add extra migrations, because the change wasn't released yet.


r10s (approved):

r10s (changes requested):
wait, why is it `ui.is_muted`? 

unless i missed sth, i think it should be [`is_muted`](https://c.delta.chat/classdc__context__t.html#aff3b894f6cfca46cab5248fdffdf083d)

WofWca (approved):
I don't know what's the true core key, but the code does replace all occurrences of it.

PR #4925, opened 2025-04-04
tauri: add `--watch-translations` cli flag

Simon-Laux:
- **tauri: add `--translation-watch` also load from the right folder during development**
- **rename option to --watch-translations**

enabled by default in development. So that you can edit translations and they are automatically reloaded like in the electron edition.
I renamed the option, because I thought the old name was confusing:

Bildschirmfoto 2025-04-04 um 01 56 22

---

> [!TIP]
>
>  Idea: Together with the env var `DELTACHAT_LOCALE_DIR` you could make a visual editor for translations that could be used by translators without a dev setup, possibly even with direct integration into transifex (or weblate or whatever we eill be using then).

### DIFF
--- a/packages/target-tauri/src-tauri/src/i18n/mod.rs
+++ b/packages/target-tauri/src-tauri/src/i18n/mod.rs
@@ -68,3 +68,41 @@ pub async fn get_all_languages(app: &AppHandle) -> Result<Vec<(String, String)>,
     vec.sort_unstable_by(|a, b| a.0.cmp(&b.0));
     Ok(vec)
 }
+
+#[cfg(desktop)]
+pub fn watch_translations(app: AppHandle) {
+    use std::time::Duration;
+    use tauri::Manager;
+
+    use crate::{
+        state::main_window_channels::MainWindowEvents, util::fs_watcher::async_watch_debounced,
+        MainWindowChannels,
+    };
+
+    let app_clone = app.clone();
+    let callback = Box::new(move || {
+        let app_clone = app_clone.clone(); // Shadows arc
+        async move {
+            if let Err(err) = app_clone
+                .state::<MainWindowChannels>()
+                .emit_event(MainWindowEvents::LocaleReloaded(None))
+                .await
+            {
+                log::error!("watch_translations: failed notify frontend: {err}:?");
+            }
+        }
+    });
+
+    tauri::async_runtime::spawn(async move {
+        if let Ok(watch_dir) = get_locales_dir(&app).await {
+            log::info!("watch_translations: watching for changes in {watch_dir:?}");
+            if let Err(err) =
+                async_watch_debounced(watch_dir, callback, Duration::from_millis(400)).await
+            {
+                log::error!("watch_translations: failed to watch locales dir: {err:?}");
+            }
+        } else {
+            log::error!("watch_translations: failed to get locales dir");
+        }
+    });
+}
--- a/packages/target-tauri/src-tauri/src/i18n/mod.rs
+++ b/packages/target-tauri/src-tauri/src/i18n/mod.rs
@@ -81,7 +81,7 @@ pub fn watch_translations(app: AppHandle) {
 
     let app_clone = app.clone();
     let callback = Box::new(move || {
-        let app_clone = app_clone.clone(); // Shadows arc
+        let app_clone = app_clone.clone();
         async move {
             if let Err(err) = app_clone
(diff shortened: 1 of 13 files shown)


packages/target-tauri/src-tauri/src/util/fs_watcher.rs
```diff
@@ -0,0 +1,66 @@
...
+    watcher.watch(path.as_ref(), RecursiveMode::Recursive)?;
+
+    let mut last_action = SystemTime::now();
+
+    while let Some(res) = rx.recv().await {
+        match res {
+            Ok(event) => {
+                log::trace!("changed: {:?}", event);
+                if event.kind.is_create() || event.kind.is_modify() || event.kind.is_remove() {
+                    let now = SystemTime::now();
```
Simon-Laux, line 47: maybe we should replace it with `Instant::now()` because `SystemTime` can run backwards in theory.

packages/target-tauri/src-tauri/src/util/fs_watcher.rs
```diff
@@ -0,0 +1,66 @@
...
+                        callback().await;
+                        last_action = now;
+                    } else {
+                        log::error!("debounce: system time error: duration_since");
+                        return Ok(());
+                    }
+                }
+            }
+            Err(e) => log::error!("watch error: {:?}", e),
+        }
```
WofWca, line 62: I'd like to remind about the early return pattern (just remind and not force it).

packages/target-tauri/src-tauri/src/i18n/mod.rs
```diff
@@ -68,3 +68,41 @@ pub async fn get_all_languages(app: &AppHandle) -> Result<Vec<(String, String)>,
...
+    use tauri::Manager;
+
+    use crate::{
+        state::main_window_channels::MainWindowEvents, util::fs_watcher::async_watch_debounced,
+        MainWindowChannels,
+    };
+
+    let app_clone = app.clone();
+    let callback = Box::new(move || {
+        let app_clone = app_clone.clone(); // Shadows arc
```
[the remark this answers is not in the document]
Simon-Laux:
```suggestion
        let app_clone = app_clone.clone();
```

packages/target-tauri/src-tauri/src/util/fs_watcher.rs
Simon-Laux: how would you do that here?

packages/target-tauri/src-tauri/src/util/fs_watcher.rs
Simon-Laux: Though on the other hand instant is a bit counter intuitive as time for it is not linear.. We gotta try it to be sure.

packages/target-tauri/src-tauri/src/util/fs_watcher.rs
WofWca:
```rust
let event = match res {
    Err(e) => {
        log...
        contunue;
    },
    Ok(event) => event,
};
```

Also `if` can be un-nested.

```rust
if !(,..) {
    continue;
}
```

packages/target-tauri/src-tauri/src/i18n/mod.rs
```diff
@@ -68,3 +68,41 @@ pub async fn get_all_languages(app: &AppHandle) -> Result<Vec<(String, String)>,
...
+                log::error!("watch_translations: failed notify frontend: {err}:?");
+            }
+        }
+    });
+
+    tauri::async_runtime::spawn(async move {
+        if let Ok(watch_dir) = get_locales_dir(&app).await {
+            log::info!("watch_translations: watching for changes in {watch_dir:?}");
+            if let Err(err) =
+                async_watch_debounced(watch_dir, callback, Duration::from_millis(400)).await
```
WofWca, line 100: It'd be nice to have an explanation about why debounce is needed. At a glance, we're just emitting an event, which shouldn't be expensive.

WofWca (approved):
The code looks alright, didn't test. I still don't think that we need these advanced options in production build, but it's off by default, so it's fine.

packages/target-tauri/src-tauri/src/i18n/mod.rs
Simon-Laux: there are extra crates for debouncing: https://lib.rs/search?q=debounce+notify

PR #4926, opened 2025-04-04
tauri: resume from sleep

Simon-Laux:
choose elapsed time method over native apis, because native apis are too much work for now, and not clear whether it would be worth the extra complexity. (a tiny amount of less cpu usage for a ton of more complexity to maintain)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 - tauri: experimental: make it compile for android #4871
 - `Cmd + N` shortcut to open new chat on macOS #4901
 - tauri: add cli interface: `--help`, `--version`, and developer options (like `--dev-mode`) #4908
-- tauri: resume from sleep #4926
+- tauri: handle resume from sleep #4926
 
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - tauri: experimental: make it compile for android #4871
 - `Cmd + N` shortcut to open new chat on macOS #4901
 - tauri: add cli interface: `--help`, `--version`, and developer options (like `--dev-mode`) #4908
+- tauri: resume from sleep #4926
 
 
 ### Changed
--- a/packages/target-tauri/src-tauri/src/resume_from_sleep.rs
+++ b/packages/target-tauri/src-tauri/src/resume_from_sleep.rs
@@ -27,11 +27,11 @@ pub(crate) fn start_resume_after_sleep_detector(app: &AppHandle) {
         loop {
             sleep(PROBE_INTERVAL).await;
             let now = SystemTime::now();
-            log::debug!(
-                "{:?}",
-                now.duration_since(last_time)
-                    .unwrap_or(Duration::from_secs(1))
-            );
+            // log::debug!(
+            //     "{:?}",
+            //     now.duration_since(last_time)
+            //         .unwrap_or(Duration::from_secs(1))
+            // );
             if now
                 .duration_since(last_time)
                 .unwrap_or(Duration::from_secs(1))
--- a/packages/target-tauri/src-tauri/src/resume_from_sleep.rs
+++ b/packages/target-tauri/src-tauri/src/resume_from_sleep.rs
@@ -1,12 +1,12 @@
-// Detector to find out whether computer resumed from sleep
-//
-// Why?
-// internet connection and imap connection gets broken by sleep
-// and needs to be reconnected
-//
-// How?
-// While we could have looked into talking with system apis (like https://github.com/pewsheen/psp),
-// For now we choose a much simpler solution: check for jumps in system time
+//! Detector to find out whether computer resumed from sleep
+//!
+//! Why?
+//! internet connection and imap connection gets broken by sleep
+//! and needs to be reconnected
+//!
+//! How?
+//! While we could have looked into talking with system apis (like https://github.com/pewsheen/psp),
+//! For now we choose a much simpler solution: check for jumps in system time
 
 use std::time::{Duration, SystemTime};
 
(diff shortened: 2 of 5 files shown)


packages/target-tauri/src-tauri/src/resume_from_sleep.rs
```diff
@@ -0,0 +1,52 @@
...
+
+use crate::{state::main_window_channels::MainWindowEvents, MainWindowChannels};
+
+const PROBE_INTERVAL: Duration = Duration::from_secs(60);
+// smallest sleep duration we can detect with this
+const JUMP_DETECTION_THRESHOLD: Duration = Duration::from_secs(3);
+
+pub(crate) fn start_resume_after_sleep_detector(app: &AppHandle) {
+    let cloned_app = app.clone();
+    let _ = spawn(async move {
```
WofWca, line 24:
```suggestion
    spawn(async move {
```

This will get rid of the clippy warning I think.

packages/target-tauri/src-tauri/src/resume_from_sleep.rs
```diff
@@ -0,0 +1,52 @@
+// Detector to find out whether computer resumed from sleep
+//
+// Why?
+// internet connection and imap connection gets broken by sleep
+// and needs to be reconnected
+//
+// How?
+// While we could have looked into talking with system apis (like https://github.com/pewsheen/psp),
+// For now we choose a much simpler solution: check for jumps in system time
```
WofWca, line 9:
```suggestion
//! Detector to find out whether computer resumed from sleep
//!
//! Why?
//! internet connection and imap connection gets broken by sleep
//! and needs to be reconnected
//!
//! How?
//! While we could have looked into talking with system apis (like https://github.com/pewsheen/psp),
//! For now we choose a much simpler solution: check for jumps in system time
```

packages/target-tauri/src-tauri/src/resume_from_sleep.rs
```diff
@@ -0,0 +1,52 @@
...
+                .unwrap_or(Duration::from_secs(1))
+                > threshold
+            {
+                log::info!("Potential resume after sleep detected");
+                if let Err(err) = cloned_app
+                    .state::<MainWindowChannels>()
+                    .emit_event(MainWindowEvents::ResumeFromSleep)
+                    .await
+                {
+                    log::error!("failed to call inform frontend about resume from sleep: {err:?}");
```
WofWca, line 46:
```suggestion
                    log::error!("failed to inform frontend about resume from sleep: {err:?}");
```

packages/target-tauri/src-tauri/src/resume_from_sleep.rs
```diff
@@ -0,0 +1,52 @@
...
+            sleep(PROBE_INTERVAL).await;
+            let now = SystemTime::now();
+            // log::debug!(
+            //     "{:?}",
+            //     now.duration_since(last_time)
+            //         .unwrap_or(Duration::from_secs(1))
+            // );
+            if now
+                .duration_since(last_time)
+                .unwrap_or(Duration::from_secs(1))
```
WofWca, line 37: Shouldn't we use `Instant` here as well?

packages/target-tauri/src-tauri/src/resume_from_sleep.rs
```diff
@@ -0,0 +1,52 @@
...
+                log::info!("Potential resume after sleep detected");
+                if let Err(err) = cloned_app
+                    .state::<MainWindowChannels>()
+                    .emit_event(MainWindowEvents::ResumeFromSleep)
+                    .await
+                {
+                    log::error!("failed to call inform frontend about resume from sleep: {err:?}");
+                }
+            }
+            last_time = now;
```
WofWca, line 49:
```suggestion
            last_time = SystemTime::now();
```

This should be more reliable, because of the `.await` on `emit_event`.

WofWca (approved):
Ha! Clever.

Did you actually test this though?

Simon-Laux:
> Did you actually test this though?

yes I did on windows, linux and macOS.

packages/target-tauri/src-tauri/src/resume_from_sleep.rs
Simon-Laux: nope instant does not work here, I tried it. instant will just pause when you suspend, so there is no time difference detected.

packages/target-tauri/src-tauri/src/resume_from_sleep.rs
```diff
@@ -0,0 +1,52 @@
...
+
+const PROBE_INTERVAL: Duration = Duration::from_secs(60);
+// smallest sleep duration we can detect with this
+const JUMP_DETECTION_THRESHOLD: Duration = Duration::from_secs(3);
+
+pub(crate) fn start_resume_after_sleep_detector(app: &AppHandle) {
+    let cloned_app = app.clone();
+    spawn(async move {
+        let threshold = PROBE_INTERVAL + JUMP_DETECTION_THRESHOLD;
+        let mut last_time = SystemTime::now();
```
Simon-Laux, line 26:
```suggestion
        // Using `SystemTime` instead of `Instant` because instant pauses on sleep in my tests,
        // it is also documented in `Instant` that it doesn't measure time linearly.
        let mut last_time = SystemTime::now();
```

packages/target-tauri/src-tauri/src/resume_from_sleep.rs
Simon-Laux: I made https://github.com/deltachat/deltachat-desktop/pull/4926/files#r2029880930 to clarify/document the reasoning

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 - tauri: experimental: make it compile for android #4871
 - `Cmd + N` shortcut to open new chat on macOS #4901
 - tauri: add cli interface: `--help`, `--version`, and developer options (like `--dev-mode`) #4908
+- tauri: resume from sleep #4926
```
WofWca, line 15:
```suggestion
- tauri: handle resume from sleep #4926
```

WofWca (approved):
The code is pretty short and simple. I didn't test, but you say you did, so let's go!

PR #4929, opened 2025-04-04
tauri notifications 2

Simon-Laux:
start own notifications crate that does notifications for all 3 platforms seperately, because I found no crate that does this with all apis we need

## Progress
- [x] MacOS (implementation and tested)
   - [x] notifications (sending and clicking on them
   - [x] notifications with images
   - [x] clicking on notifications from previous sessions works
   - [x] bonus: reply from the notification
- [x] Windows (implementation and tested)
   - [x] Basic toasts (with image support for imaage attachments)
   - [x] Clicking on toast opens DC
   - [x] Optional (because electron has this bug too): clicking on notifications in the notification centre should work
- [x] Linux (implementation and tested)
- [x] Flatpak (tested)
- [x] rebase (after deeplink is merged)
- [x] changelog entry: `tauri: added notifications`

### Bugs
- [x] macOS, don't crash in dev mode when there is no bundle id set
      - `*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'bundleProxyForCurrentProcess is nil: mainBundle.bundleURL '`

### Testing

Keep in mind that MacOS and Windows versions only work when bundled & installed, so `pnpm tauri dev` will likely not have working notifications, though you can see in the log what functions would be called.

### What works:

|                                                        | kde                                                       | gnome                    | windows                                             | macos                               |
| ------------------------------------------------------ | --------------------------------------------------------- | ------------------------ | --------------------------------------------------- | ----------------------------------- |
| display toast notifications                            | works                                                     | works                    | works                                               | works                               |
| display images                                         | works                                                     | works                    | works, avatars are even round on newer win versions | works                               |
| removing existing notifications while in toast display | no                                                        | no                       | works                                               | works                               |
| removing existing notifications in center              | maybe works, but they don't land in the center            | no                       | ?                                                   | works                               |
| click on notifiacation                                 | works (but window is not moved to the front just focused) | works [^1]               | ?                                                   | works                               |
| click on notification from center                      | - does not appear in center                               | works [^1]               | TODO with the deeplink                              | works                               |
| bonus                                                  | -                                                         | -                        | -                                                   | buttons and reply from notification, also threading/grouping works |
| notifications from previous sessions work              | not possible in xdg api?  [^2]                                | not possible in xdg api? [^2]  | TODO with deeplink probably yes                     | yes                                 |

[^1]: but you sometimes get a "window ready" notification instead of directly opening the window

[^2]: maybe possible via dbus activation and parameters, may be only possible in the new portal notification api?

while linux notifications work in their basic function, there as still some quirks, but we don't have enough resources  to address them right now (even if my perfectionism would like otherwise). it works good enough for normal usage for now. there are also questions with the api or better which api to use because linux has multiple different apis, atleast `org.freedesktop.Notifications` and `org.freedesktop.portal.Notification`.


Simon-Laux:
About linux notifications:

I plan to replace notify_rust with [ashpd](https://docs.rs/ashpd/latest/ashpd) because:
 - it is more specialized on only Linux
 - we will probably use it for other things like drag and drop portals - so good to get some experience with it
 - it's API looks more rusty
 - it is mentioned in the gnome blog post about improving notifications: https://blogs.gnome.org/shell-dev/2024/04/23/notifications-46-and-beyond/
 - on KDE notifications behave as expected: toast time runs out, then it is found in the notification center

PR #4930, opened 2025-04-05
tauri: feat: hide window on tray click if visible

WofWca:
#skip-changelog


Simon-Laux:
Note:This change is only about windows and linux on some desktops, macOS does not focus on a single click on tray icon, instead it shows the menu on right and on left click, because on macOS there is always a dock icon for the app that you can click to open the app's window.

PR #4933, opened 2025-04-05
move `notifications` to `frontendAndTauri`

Simon-Laux:
and remove outdated comments

#skip-changelog not user visible


WofWca (approved):
I don't know what's the functional benefit of this, but it makes sense, since we now have the `notification` setting in Tauri as well, since the tray MR,

Simon-Laux:
> I don't know what's the functional benefit of this

there is not any functional change, this is just cosmetic to give the reader an impression of which settings are also used by the runtime.

PR #4934, opened 2025-04-05
tauri: fix: start io for all accounts

Simon-Laux:
maybe we should think about whether we move this to the frontend on all 3 platforms

#skip-changelog tauri is not released yet and this is a small bugfix, but feel free to disagree and add a changelog entry


WofWca (approved):
I don't know how settings are supposed to be organized, but this looks like it should work.

PR #4935, opened 2025-04-05
tauri: feat: add taskbar "unread" badge on Windows

WofWca:
As I said in the comments, Tauri's `setBadgeCount` is not supported
on Windows.

This will not show the count, but only the fact
that there is a new message.
This is not ideal, but it's better than nothing.

image

TODO:
- [x] Verify that this doesn't break the app for non-Windows targets (try the preview builds).
   - [x] macOS
   - [x] linux


Simon-Laux:
I'd rather have us contribute to tauri to fix that it is not supported... But this already is an improvement.

packages/target-tauri/runtime-tauri/runtime.ts
```diff
@@ -506,7 +506,16 @@ class TauriRuntime implements Runtime {
...
-    getCurrentWindow().setBadgeCount(value === 0 ? undefined : value)
+    const window = getCurrentWindow()
+
+    // According to the docs, `setBadgeCount` is unsupported on Windows,
+    // and we should use `setOverlayIcon` instead.
+    window.setBadgeCount(value === 0 ? undefined : value)
+    // Yes, this won't show the count.
+    window.setOverlayIcon?.(
+      value === 0 ? undefined : 'images/tray/unread-badge.png'
+    )
```
Simon-Laux, line 517: would make sense to only run this on windows. you could pass that info to here via RuntimeInfo

WofWca:
Yes, I considered that.

I think we should implement the proper counter in Tauri though, and not just a binary badge.
The former takes more effort though.

packages/target-tauri/runtime-tauri/runtime.ts
WofWca: I would also prefer that. But this function is Windows-only as of now, So I suggest to keep it as is until the badge counter is supported by Tauri, or until we are otherwise forced to do it.

Simon-Laux:
> I think we should implement the proper counter in Tauri though, and not just a binary badge.
The former takes more effort though.

I don't think so, it is a windows api: https://learn.microsoft.com/en-us/uwp/api/windows.ui.notifications.badgenotification?view=winrt-26100#definition

Though for older windows, the overlay icon method might still make sense, though I don't think we will spend resources to post to windows 7.

WofWca:
> I don't think so, it is a windows api: https://learn.microsoft.com/en-us/uwp/api/windows.ui.notifications.badgenotification?view=winrt-26100#definition

Hmmm. I assumed that it's not supported because Windows only exposes the "overlay icon" thing.

PR #4936, opened 2025-04-06
tauri: hardening: check webview label in custom protos

WofWca:
There is not a known explot that this fixes, but let's do it to be extra safe.

### DIFF
--- a/packages/target-tauri/src-tauri/src/html_window/email_scheme.rs
+++ b/packages/target-tauri/src-tauri/src/html_window/email_scheme.rs
@@ -34,6 +34,15 @@ pub(crate) fn email_protocol<R: tauri::Runtime>(
     // URI format is email://<anything>
     // (file path doesn't matter, because the html content is always returned)
 
+    let webview_label = ctx.webview_label();
+    if !webview_label.starts_with("html-window:") || !webview_label.ends_with("-mail") {
+        error!(
+            "prevented other window from accessing email:// scheme (webview label: {})",
+            webview_label
+        );
+        return;
+    }
+
     let webview_label = ctx.webview_label().to_owned().replace("-mail", "");
     let app_handle = ctx.app_handle().clone();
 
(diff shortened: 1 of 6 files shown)


packages/target-tauri/src-tauri/src/html_window/email_scheme.rs
```diff
@@ -34,6 +34,15 @@ pub(crate) fn email_protocol<R: tauri::Runtime>(
...
 
+    let webview_label = ctx.webview_label();
+    if !webview_label.starts_with("html-window:") || !webview_label.ends_with("-mail") {
+        error!(
+            "prevented other window from accessing email:// scheme (webview label: {})",
+            webview_label
+        );
+        return;
+    }
+
```
Simon-Laux, line 45: technically this unnecessary, because this scheme loads the email based on the window/webview it was accessed by, so in the case of another window accessing it it will return an error already.

packages/target-tauri/src-tauri/src/webxdc/webxdc_scheme.rs
```diff
@@ -99,6 +99,14 @@ pub(crate) fn webxdc_protocol<R: tauri::Runtime>(
     // windows/android:   webxdc://webxdc.localhost/<path>
 
     let webview_label = ctx.webview_label().to_owned();
+    if !webview_label.starts_with("webxdc:") {
+        error!(
+            "prevented other window from accessing webxdc:// scheme (webview label: {})",
+            webview_label
+        );
+        return;
+    }
```
Simon-Laux, line 108: same as above, this scheme only works inside of a webxdc. so technically this wouldn't be necessary. however we could also have it for extra clarity.

Simon-Laux (approved):

packages/target-tauri/src-tauri/src/html_window/email_scheme.rs
WofWca: This is the case right now, but hypothetically the behavior could change.
So I think it's a safe bet to just have a simple guard at the very start of the function.
This also helps with readability. It's media to clear that this protocol is only supposed to be used by windows with such label.

PR #4705, opened 2025-02-26
tauri: fix blobs and webxdc-icon scheme under windows & fix sticker picker

Simon-Laux:
fixes that blobs were not displayed on windows

PR #4941, opened 2025-04-06
tauri: fix: sticker picker previews not working, again

WofWca:
This is a follow-up to 99e55226940224b0a853a588c27d434ee2cc9390
(https://github.com/deltachat/deltachat-desktop/pull/4705).

There are several problems that this fixes:
1. On Windows, the path separator is `\`, so the URL would not
   be formatted according to the specified format:
   we'd put `\` instead of `/` between the sticker pack folder
   and the sticker name.
   However, this did not cause problems, because the browser
   automatically replaced it with a forward slash.
2. The result of `encodeURIComponent` was never used.
3. If the sticker pack folder name or the sticker name
   was not a validly percent-encoded string,
   the front-end would crash, because of `decodeURIComponent()`.

Now the stickers are displayed for path suchs as
"Utya%20Duck" and "Utya Duck%".

Closes https://github.com/deltachat/deltachat-desktop/issues/4548
for Tauri.


packages/target-tauri/runtime-tauri/runtime.ts
```diff
@@ -407,16 +407,17 @@ class TauriRuntime implements Runtime {
...
 
     if (matches) {
-      let filename = matches[3]
-      if (decodeURIComponent(filename) === filename) {
-        // if it is not already encoded then encode it.
-        filename = encodeURIComponent(filename)
-      }
-      return `${this.runtime_info?.tauriSpecific?.scheme.stickers}${matches[2]}/${matches[3]}`
+      // Keep in mind that the sticker pack folder and sticker name
+      // are basically arbitrary.
```
WofWca, line 417:
```suggestion
      // Keep in mind that the sticker pack folder and sticker name
      // can include arbitrary characters.
```

Simon-Laux (approved):

PR #4949, opened 2025-04-07
tauri: fix: respect `enableWebxdcDevTools` setting

WofWca:
I have tested this on Windows. It works.

I haven't re-checked whether the attack that `enableWebxdcDevTools` was supposed to mitigate still works.


Simon-Laux:
also works on macOS

packages/target-tauri/src-tauri/src/webxdc/commands_main_window.rs
```diff
@@ -313,6 +315,24 @@ pub(crate) async fn open_webxdc<'a>(
...
         // have no effect (see below).
         .proxy_url(dummy_localhost_proxy_url.clone())
+        .devtools({
+            // Dev tools might not work on macOS in production,
+            // see comments around `enableWebxdcDevTools`.
+            //
+            // TODO check whether opening dev tools is an exfiltration risk
+            // on WebKit (see comments about `enableWebxdcDevTools`),
+            // otherwise we need no special treatment
+            // for webxdc windows' dev tools.
```
Simon-Laux, line 325: it's not only exfiltration, that you can paste stuff you don't know about can also be seen as a risk. anyway in macOS appstore releases this will probably be deactivated anyway.

Simon-Laux (approved):

packages/target-tauri/src-tauri/src/webxdc/commands_main_window.rs
WofWca: Perhaps I haven't phrased it super well. I mentioned "special treatment" because we have a separate switch for the main window's dev tools, and if there is no exfiltration, then we probably should have the same switch for both webxdc and the main window.

WofWca:
The second commit is https://github.com/deltachat/deltachat-desktop/pull/4950

PR #4951, opened 2025-04-07
Feat(Tauri): Add drag & drop support

Septias:
Drag and drop support for runtime tauri. This PR adds a new function `setDragListener` that can handle tauri emitted drag events. We need this because browser drag events are not reliably supported across operating systems (see https://github.com/tauri-apps/tauri/discussions/4736). Sadly tauri does not provide a browser native `FileList` but only a list of strings that contain the paths. That is why the custom tauri drag handler copies the behaviour of the native drag handler. 

close https://github.com/deltachat/delta-tauri-todo/issues/15

- [x] drop in a single file
- [x] drop in multiple files
- [x] drop out a single file

#skip-changelog

### DIFF
--- a/packages/frontend/src/components/message/MessageListAndComposer.tsx
+++ b/packages/frontend/src/components/message/MessageListAndComposer.tsx
@@ -111,19 +111,26 @@ export default function MessageListAndComposer({ accountId, chat }: Props) {
     regularMessageInputRef
   )
 
-  const onDrop = async (e: React.DragEvent<any>) => {
+  runtime.setDragListener((e) => {
+    if (e.payload.type == "drop") {
+      handleDrop(e.payload.paths)
+    }
+  });
+
+  const onDrop = (e: React.DragEvent<any> ) => {
+    e.preventDefault()
+    e.stopPropagation()
+    handleDrop(e.dataTransfer.files)
+  }
+  const handleDrop = async (fileList: FileList) => {
     if (chat === null) {
       log.warn('dropped something, but no chat is selected')
       return
     }
 
-    e.preventDefault()
-    e.stopPropagation()
 
     const sanitizedFileList: File[] = []
     {
-      const fileList: FileList =
-        /* (e.target as any).files */ e.dataTransfer.files
       for (let i = 0; i < fileList.length; i++) {
         const file = fileList[i]
         if (runtime.isDroppedFileFromOutside(file)) {
--- a/packages/frontend/src/components/message/MessageListAndComposer.tsx
+++ b/packages/frontend/src/components/message/MessageListAndComposer.tsx
@@ -1,5 +1,5 @@
 import React, { useRef, useEffect, useCallback } from 'react'
-import { basename, join, parse } from 'path'
+import { basename, join, parse, ParsedPath } from 'path'
 import { T } from '@deltachat/jsonrpc-client'
 
 import Composer, { useDraft } from '../composer/Composer'
@@ -14,6 +14,8 @@ import ConfirmSendingFiles from '../dialogs/ConfirmSendingFiles'
 import { ReactionsBarProvider } from '../ReactionsBar'
 import useDialog from '../../hooks/dialog/useDialog'
 import useMessage from '../../hooks/chat/useMessage'
+import { BackendRemote } from '../../backend-com'
+import { Viewtype } from '@deltachat/jsonrpc-client/dist/generated/types'
 
 const log = getLogger('renderer/MessageListAndComposer')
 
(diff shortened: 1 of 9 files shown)


packages/target-tauri/src-tauri/capabilities/main-window.toml
```diff
@@ -19,6 +19,9 @@ permissions = [
     "store:allow-save",
     "store:allow-set",
     "store:allow-delete",
+    # events
```
Simon-Laux, line 22: let's say it is only for drag and drop events, because we explicitly don't use events for other things anymore. we migrated to using channels for security (events can be seen by all windows).

```suggestion
    # events - only used for drag and drop
```

Septias:
Is there a way to check at runtime if the runtime is tauri? I want to disable the native drag handler in this case so that drags are not handled twice.

Septias:
I didn't touch the browser packages, why is CI failing me there?

Simon-Laux (changes requested):
### Testing

my test on macOS:

> drop in a single file

file extension is missing, I drag in an image, but dc desktop doesn't display it because it needs the file extension to detect the type.

> drop in multiple files

opens a ton of dialogs, looks like it wants to send it to all kinds of chats of all kinds of accounts. (maybe all chats that I had previously selected? maybe there is a event listener leak where event listeners are never cleaned up?).

> dragging out of a single file

works fine, but filename is a uuid, could we use the original filename?

packages/target-tauri/runtime-tauri/runtime.ts
```diff
@@ -102,6 +104,13 @@ class TauriRuntime implements Runtime {
   emitUIFullyReady(): void {
     invoke('ui_frontend_ready')
   }
+  async setDragListener(
+    fn: (event: Event<DragDropEvent>) => void
+  ): Promise<UnlistenFn> {
+    return await getCurrentWebview().onDragDropEvent(event => {
+      fn(event)
+    })
+  }
```
Simon-Laux, line 113: this creates a new listener every time that is never cleaned up.

Simon-Laux:
I can only drag out images, not files, is this a limitation of the tauri api?

Septias:
I fixed image name + extension and multiple listeners for drop events. I'm sure not being able to drag out files is a tauri limitation. 
For me, dragging out an image now gives `Pasted Image` in Files menu. Can you recheck if you still get the hash?

packages/target-tauri/src-tauri/capabilities/main-window.toml
```diff
@@ -20,6 +20,9 @@ permissions = [
     "store:allow-get",
     "store:allow-set",
     "store:allow-delete",
+    # events - only used for drag and drop
+    "core:event:allow-listen",
+    "core:event:default",
```
WofWca, line 25: `emit` is not used? Can we remove this?

packages/frontend/src/components/message/MessageListAndComposer.tsx
```diff
@@ -95,26 +105,96 @@ export default function MessageListAndComposer({ accountId, chat }: Props) {
...
+        return
+      }
+      if (chat === null) {
+        log.warn('dropped something, but no chat is selected')
+        return
+      }
+
+      // sanitize files
+      const paths = e.payload.paths
+      const forbiddenPathRegEx = /DeltaChat\/.+?\.sqlite-blobs\//gi
```
WofWca, line 120: Can we avoid duplication? And BTW I think `sqlite-blobs` is no longer used.

https://github.com/deltachat/deltachat-desktop/blob/9e7a6a1e815ca66a6cfe0df5787107cda428973d/packages/target-electron/runtime-electron/runtime.ts#L139

packages/runtime/runtime.ts
```diff
@@ -10,6 +10,8 @@ import {
 } from '@deltachat-desktop/shared/shared-types.js'
 import { LocaleData } from '@deltachat-desktop/shared/localize.js'
 import { BaseDeltaChat } from '@deltachat/jsonrpc-client'
+import { DragDropEvent } from '@tauri-apps/api/webview'
+import { Event, UnlistenFn } from '@tauri-apps/api/event'
```
WofWca, line 14:
```suggestion
import type { DragDropEvent } from '@tauri-apps/api/webview'
import type { Event, UnlistenFn } from '@tauri-apps/api/event'
```

packages/frontend/src/components/message/MessageListAndComposer.tsx
```diff
@@ -95,26 +105,96 @@ export default function MessageListAndComposer({ accountId, chat }: Props) {
...
+              })
+            }
+          },
+        })
+      }
+    })
+    return () => {
+      unset.then(u => {
+        log.info('dragListenerUnset')
+        u()
```
WofWca, line 174: Isn't this gonna be an error on Electron? Would be nice to get rid of the `any` types.

packages/frontend/src/components/message/MessageListAndComposer.tsx
```diff
@@ -69,6 +71,14 @@ type Props = {
   accountId: number
 }
 
+function fullPath(file: ParsedPath) {
+  return file.dir + '/' + file.name + file.ext
```
WofWca, line 75: This won't work on Windows.

packages/frontend/src/components/message/MessageListAndComposer.tsx
```diff
@@ -69,6 +71,14 @@ type Props = {
...
+}
+function isImage(file: ParsedPath) {
+  const imageExtensions = ['.jpg', '.jpeg', '.png', '.gif']
```
WofWca, line 78: IIRC GIF has its own Viewtype?

https://github.com/deltachat/deltachat-desktop/blob/9047e97f479e8ddf6abeed35e140ea63a64c6fd7/packages/frontend/src/components/attachment/mediaAttachment.tsx#L40

packages/frontend/src/components/message/MessageListAndComposer.tsx
```diff
@@ -95,26 +105,96 @@ export default function MessageListAndComposer({ accountId, chat }: Props) {
...
+    })
+    return () => {
+      unset.then(u => {
+        log.info('dragListenerUnset')
+        u()
+      })
     }
+  })
 
+  const onDrop = (e: React.DragEvent<any>) => {
```
WofWca, line 179: Isn't this function going to be invoked as well on Tauri?

WofWca (changes requested):

packages/frontend/src/components/message/MessageListAndComposer.tsx
Septias: Yeah this still needs to be adapted

packages/frontend/src/components/message/MessageListAndComposer.tsx
Septias: Actually, it should work because Rust uses an OS-agnostic Path type, which parses fine in core and works from that point onwards on all OS.

packages/runtime/package.json
```diff
@@ -8,6 +8,7 @@
   },
   "dependencies": {
     "@deltachat-desktop/shared": "link:../shared",
-    "@deltachat/jsonrpc-client": "catalog:"
+    "@deltachat/jsonrpc-client": "catalog:",
+    "@tauri-apps/api": "^2.2.0"
```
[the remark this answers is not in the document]
Septias: This is only used for a type. I can also just copy the type over if that is preferred.

Simon-Laux:
superceded by https://github.com/deltachat/deltachat-desktop/pull/5109

PR #4954, opened 2025-04-07
remove now unused stock translations

r10s:
the translations were removed at
https://github.com/chatmail/core/pull/6777

#skip-changelog


Simon-Laux (approved):
strictly speaking is for a core update, but not technically required since not set stock strings will still compile, so merging already.

Issue #3657, opened 2024-02-03
Register to handle i.delta.chat links


Simon-Laux:
I'm not completely sure if that is possible on desktop, but it is definitely possible on mobile (android and iOS).

https://github.com/deltachat/invite/issues/19

DC is registered as a handler for the custom schemes at https://github.com/deltachat/deltachat-desktop/blob/master/build/gen-electron-builder-config.js#L21, not sure if it is possible to handle deep-linking URLs as well, so this would require diving into electron and electron-builder documentation.

Simon-Laux:
for macOS there is https://developer.apple.com/documentation/bundleresources/applinks, not sure yet how to set the associated domains with electron builder, but it should be possible.

Simon-Laux:
we need to add this to the entitlement file:
```xml
<key>com.apple.developer.associated-domains</key>
<array>
	<string>applinks:i.delta.chat</string>
</array>
```

as the link file on the server (https://i.delta.chat/.well-known/apple-app-site-association) is cached on Apple CDN we can use `applinks:i.delta.chat?mode=developer` instead to skip the CDN for development (when we need to change the file on i.delta.chat for example for updating the app id of desktop, because it might need some code in front of it like the iOS one did, currently I'm not sure what kind of code this is, might be the apple team id.).

r10s:
this seems not possible easily. closing stale feature proposal

Simon-Laux:
FTR: it is implemented and working in the tauri edition.

It could be easy, but it could also be harder (adding to entitlement and associated IDs is easy, handling the event in electron could be easy or hard). 
Anyway it seems like would only work for macOS and all this is "just" for minimizing data when having DC installed and clicking on invite links. This data is only a metadata risk when invite page code changes or the browser is compromised, both highly unlikely. So this issue has not that much reward for the cost.

PR #4956, opened 2025-04-08
start with deeplinking and xdc filehandling on tauri

Simon-Laux:
Additionally to deep linking and xdc file opening support this also adds microphone and camera usage strings and entitlements on macOS (and a few more entitlements that will become relevant when we attempt to publish it to to Mac AppStore).

Progress:
- [x] accept links from cli
- [x] accept xdc file (via local file:// links) from cli
- webxdc file icons
    - [x] Mac
    - [x] linux (well, default document icon instead of zip file icon, like the electron version already does)
    - [x] windows
- [X] MacOS: works when deltachat is already active
- [x] MacOS: works when deltachat is not already active -> crashes at the moment (issue in my handler code most likely)
- [x] MacOS: debug the program to find out why it doesn't work with real invite links, but works for seemingly all other links (invite links were app links already, but I didn't use my patched tao version, so it of course didn't work)
- [x] MacOS: `setDefaultApplicationAtURL` for the custom schemes
- [x] Linux deb
- [x] Linux rpm (untested, but should work)
- [x] Windows
     - [x] nsis (`setup.exe`)
     - ~~wix (`setup.msi`)~~ moved to later, because I can't get it to work easily
     - ~~msstore, appx~~ tauri has no support for building UWP / APPX packages yet and thus no way to append to the manifest

### Optional sidequest - get i.delta.chat links to work (#3657) on MacOS:

- [x] implement `application:continueUserActivity:restorationHandler:` https://github.com/tauri-apps/tao/pull/1108
- [x] find a better solution than prefixing app id with apple dev team id. (either script that builds with extra config file, or better yet, get it to work without the team id, which I prepared for in https://github.com/deltachat/invite/commit/7db4cd02bb9a181a3012b4071f7b476b332f515c)
   - don't have the prefix anymore, just in the app id in the entitlement and it still seems to work, but maybe only because both variants (with and without prefix) are present in the file https://github.com/deltachat/invite/commit/7db4cd02bb9a181a3012b4071f7b476b332f515c

We could probably also implement this for the electron target/runtime, as electron has the handler already https://www.electronjs.org/docs/latest/api/app#event-continue-activity-macos.
Though would probably require an updated app-id (the thing that is registered on the apple homepage) and provisioning profile. (~~and maybe some magic to not have the change the apps bundle id, as this could have some side effects for existing users that we don't want~~ seems like it is possible without changing the bundle id.)

### Testing

Keep in mind that MacOS and Windows versions only work when bundled & installed, macOS version also needs to be signed additionally for the `i.delta.chat` links to work, because that requires an entitlement from apple.

---

this pr depends on
- https://github.com/tauri-apps/tao/pull/1108


Simon-Laux:
> wix (setup.msi) moved to later, because I can't get it to work easily

we want a custom icon for the webxdc file type association. This is not yet supported by the fileAssociations key in the config file, because the app icon is hardcoded as icon there.

There is some way to extend it (https://v2.tauri.app/distribute/windows-installer/#extending-the-installer-with-wix-fragments), but I can't figure out how to use it. I always get this error:

```
   Compiling deltachat-tauri v1.56.1 (C:\Users\User\test\deltachat-desktop\packages\target-tauri\src-tauri)
    Finished `release` profile [optimized] target(s) in 8m 04s
       Built application at: C:\Users\User\test\deltachat-desktop\target\release\deltachat-tauri.exe
        Info Target: x64
     Running candle for "main.wxs"
     Running candle for "C:\\Users\\User\\test\\deltachat-desktop\\packages\\target-tauri\\src-tauri\\./windows/fragment
s/file-handler.wxs"
     Running light to produce C:\Users\User\test\deltachat-desktop\target\release\bundle\msi\deltachat-tauri_1.56.1_x64_
en-US.msi
failed to bundle project: error running light.exe: `failed to run C:\Users\User\AppData\Local\tauri\WixTools314\light.ex
e`
       Error failed to bundle project: error running light.exe: `failed to run C:\Users\User\AppData\Local\tauri\WixTool
s314\light.exe`
```

I also tried it with the example from the website, but seems like I had no luck. If someone else wants to try, this is what I have got so far, but I would just postpone it for now as the NSIS installer works fine, this is only about the `.msi`/wix installer:

tauri config:
```json
"bundle": {
  "windows": {
      "wix": {
        "fragmentPaths": ["./windows/fragments/file-handler.wxs"],
        "componentRefs": ["DCTauriFileHandlers"]
      }
    }
}
```

`src-tauri/windows/fragments/file-handler.wxs`:
```xml
<?xml version="1.0" encoding="utf-8"?>
<Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
  <Fragment>
    <DirectoryRef Id="TARGETDIR">
      <Component Id="DCTauriFileHandlers" Guid="77208bf2-9dca-41d4-a14f-983c109df115">
        <!-- https://docs.firegiant.com/wix3/xsd/wix/component/  -->

        <File Id="Path" Source="{{main_binary_path}}" KeyPath="yes" Checksum="yes"/>
        <ProgId Id="{{../../product_name}}.xdc" Advertise="yes" Description="Webxdc Mini App">
            <Extension Id="xdc" Advertise="yes">
                <Verb Id="open" Command="Share App with {{../../product_name}}" Argument="&quot;%1&quot;" />
            </Extension>
        </ProgId>

      </Component>
    </DirectoryRef>
  </Fragment>
</Wix>
```

WofWca (commented):
I tested this a little on Windows. I built with `--debug` and installed the nsis version. Opening an `.xdc` file works nicely, even when DC is initially closed.

image

but the icon on .xdc files is not there:

image

However, i.delta.chat links, mailto: and the custom protocols didn't seem to have registered:

image

I also tried installing a production build, and nothing appeared to change.

As per code, right now I can't say about the [`hooks.nsi`](https://github.com/deltachat/deltachat-desktop/pull/4956/files#diff-768988089bd4d8448361cb8493b53cbe38842a0c8581a959b589d059d0019b88) file much now, that's not something I've ever worked with. I also can't say about the macOS-specific files (entitlements, etc)  
The Rust part of the code seems OK at a glance (note the `todo!`s, however), but I didn't dig deep.

WofWca:
I tried the release NSIS installed in a Windows 10 Sandbox again. Again, it only registered as a handler for .xdc files, but not links or a mail agent.

packages/target-tauri/src-tauri/windows/hooks.nsi
```diff
@@ -0,0 +1,42 @@
+!macro URI_SCHEME protocol
```
WofWca, line 1: We should probably link to some docs that help understand what this file is about.
Firstly, https://tauri.app/distribute/windows-installer/#extending-the-installer, then maybe an article on "how to register app as default for custom schemes on Windows".

PR #4961, opened 2025-04-08
Transport api followup

nicodh:
This PR contains only a reorganization of code and moves the ConfigureProgressDialog from LoginForm to a separate file

It is based on PR #4945 but separated to make the review of the other PR easier

#skip-changelog


Simon-Laux (approved):
code lgtm, looks like the code was mostly just moved?

nicodh:
> code lgtm, looks like the code was mostly just moved?

Yes as described - but needs #4945 to be merged first

PR #4968, opened 2025-04-10
tauri: chore: upgrade core & dependencies, rust-version

WofWca:
Run `cargo update --recursize deltachat deltachat-jsonrpc`;
Bump `rust-version` from 1.84 to 1.85 because `shadowsocks@1.23.0`
requires the `edition2024` feature.

I checked that it works.

#skip-changelog because the "upgrade core" entry is already present.

### DIFF
--- a/.github/workflows/rust-tauri-lint.yml
+++ b/.github/workflows/rust-tauri-lint.yml
@@ -23,7 +23,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: 1.84
+          toolchain: 1.85
           override: true
       - run: rustup component add rustfmt
       - uses: actions-rs/cargo@v1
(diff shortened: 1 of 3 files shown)


WofWca:
CI error:

```
Caused by:
    feature `edition2024` is required
  
    The package requires the Cargo feature called `edition2024`, but that feature is not stabilized in this version of Cargo (1.84.1 (66221abde 2024-11-19)).
    Consider trying a newer version of Cargo (this may require the nightly release).
    See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#edition-2024 for more information about the status of this feature.
```

Can we just bump the Rust version in Cargo.toml and in

https://github.com/deltachat/deltachat-desktop/blob/aa07f3902e7f642935b9513908028a441fa8ec84/.github/workflows/rust-tauri-lint.yml#L26

?

Simon-Laux (approved):

PR #4970, opened 2025-04-10
remove address from primary settings UI

r10s:
showing the address here prominently results in bad UX as ppl try to pass the address around,
which will not result in e2ee and will often not work.

see https://github.com/deltachat/deltachat-ios/pull/2664 for more resoning and discussion, eg. why remove it unconditionally

cmp https://github.com/deltachat/deltachat-android/pull/3730

#skip-changelog


nicodh (approved):

nicodh (changes requested):
We need to adapt the e2e tests which read the address from existing profiles from thge settings. I will do that

nicodh (approved):

PR #4971, opened 2025-04-11
Min height for Bio field

### LINKED ISSUE #4967: The "Bio" field in the profile view has a too small minimum height
The "Bio" field in the profile view has a too small minimum height. This means on a small screen and when there's lots of other info in the profile view, the "Bio" field is so small that a single mouse wheel scroll almost advances you more than what's visible in the tiny area, making it challenging to read it if it's longer than 4-5 words:

Image

- Delta Chat Version: 1.56.0 (latest on flathub), core v1.157.3
- Expected behavior: "Bio" field has larger minimum height, with the profile view as a whole instead adding a scroll bar if needed
- Actual behavior: "Bio" field minimum height seems to be too small
- Steps to reproduce the problem: 1. Resize delta chat window to be small, 2. Enable related chats, 3. Look at somebody's profile with related chats and a "Bio" specified that is longer than a few words
- Screenshots: see above
- Logs:

nicodh:
resolves #4967

#skip-changelog

### DIFF
--- a/packages/frontend/src/components/dialogs/ViewProfile/styles.module.scss
+++ b/packages/frontend/src/components/dialogs/ViewProfile/styles.module.scss
@@ -10,10 +10,11 @@
 }
 
 .statusText {
-  padding: 15px 15px;
+  padding: 10px 15px;
   color: var(--globalText);
   white-space: pre-wrap;
   overflow-y: scroll;
+  min-height: 42px;
 }
 
 .buttonWrap {
--- a/packages/frontend/src/components/dialogs/ViewProfile/styles.module.scss
+++ b/packages/frontend/src/components/dialogs/ViewProfile/styles.module.scss
@@ -10,11 +10,9 @@
 }
 
 .statusText {
-  padding: 10px 15px;
+  padding: 15px 15px;
   color: var(--globalText);
   white-space: pre-wrap;
-  overflow-y: scroll;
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/dialogs/ViewProfile/styles.module.scss
```diff
@@ -10,10 +10,11 @@
 }
 
 .statusText {
-  padding: 15px 15px;
+  padding: 10px 15px;
   color: var(--globalText);
   white-space: pre-wrap;
   overflow-y: scroll;
+  min-height: 42px;
```
WofWca, line 17:
```suggestion
  padding: 15px 15px;
  color: var(--globalText);
  white-space: pre-wrap;
```

WofWca (commented):
This would be an improvement, but I don't think this is the way to close the issue. With longer bios, 42px is still not enough of a scope to read through them.

Instead I suggest to make it non-scrollable, and let the dialog be scrollable instead

image

packages/frontend/src/components/dialogs/ViewProfile/styles.module.scss
nicodh: But that then it's always full expanded. But maybe fine here...

Simon-Laux (approved):
code LGTM

PR #4974, opened 2025-04-11
Fix sticker folder on windows

nicodh:
resolves #4943

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,7 @@
 - tauri: fix emoji picker being super ugly
 - tauri: use current locale in "Help" window when opening it through menu
 - tauri: fix launching a second instance of Delta Chat not focusing the main window if it's closed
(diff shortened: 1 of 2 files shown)


WofWca (commented):
I don't know what the code does, but this fixes it on my Windows machine, at least my particular case. Also it keeps working on my Linux machine

nicodh:
The point is that entities that are valid path parts but would be decoded by the browser (like %, #, ? ) etc. need to be encoded before passed to the browser:
a path like "folder%20name/file" would be decoded to  "folder name/file" which does not exist so it is passed to the browser encoded: "folder%2520name/file"
Since Window paths have backslashes these will be encoded to %5C and since the browser on windows does not decode them back it fails

WofWca (approved):
It seems weird that we need an in-house URL parser/serializer, but, since we don't support certain kinds of paths anyway (such as `Abc%abc`), this should be an improvement.

Issue #4976, opened 2025-04-11
Improve voice message recording


WofWca:
Follow-up to https://github.com/deltachat/deltachat-desktop/pull/4920.

- [x] If you click "OK" within 1 second after starting recording, you'll get the NO_INPUT error, regardless of whether your mic is working
   https://github.com/deltachat/deltachat-desktop/blob/b998471f09a1f0b124f95647183fad14411a3b49/packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx#L100-L103

   https://github.com/user-attachments/assets/c941ace7-fb54-409b-8b29-0306f9440e8a

- [x] It doesn't look good on narrow screens. Although still usable

   https://github.com/deltachat/deltachat-desktop/pull/4986
   
   ![Image](https://github.com/user-attachments/assets/a3468adf-affb-44e9-b0b4-1baa2fa5b5f6)
- [x] Need to ensure that it properly works when the user doesn't grant the permission the first time
   https://github.com/deltachat/deltachat-desktop/pull/4920#issuecomment-2779868680
   https://github.com/deltachat/deltachat-desktop/issues/4988

nicodh:
>   [ ]  If you click "OK" within 1 second after starting recording, you'll get the NO_INPUT error, regardless of whether your mic is working

That could only happen if you click cancel before any sound was passed to the Recorder which happens so fast that I could not reproduce your case. But an issue is, that you probably don't want to see an error message after canceling the recording.

nicodh:
> It doesn't look good on narrow screens. Although still usable

This should be better now since with PR #4986 the vol meter is hidden for small screens via media query 

https://github.com/deltachat/deltachat-desktop/pull/4986/files?diff=unified&w=1#diff-769f75b81cc687f60e9af48e627fdb8efe8608b32f185df559644c374358998d

WofWca:
OK, looks like we can close this now. The Tauri permissions issue is tracked separately.

PR #4983, opened 2025-04-12
Dont show audio error dialog after stopping immediately

nicodh:
related to #4976

#skip-changelog


r10s (approved):
makes sense code-wise, did not test

PR #4984, opened 2025-04-12
show signature/bio in settings

### LINKED ISSUE #4978: show bio summary in settings screen
the place [freed up by the removal of the email address](https://github.com/deltachat/deltachat-desktop/pull/4970) can now be used by a summary of the bio (summary means, replace lineends by spaces and show one line)

if no bio is set, the string `pref_default_status_label` should be shown.

idea is that the profile becomes more weight by multi-transport, therefore it is good to let ppl know already from the setting page that sth like a bio exist (forcing it on onboarding, however, is too much). one set, it is more visible, so that it might be update more often etc.

moreover, it also looks better, the current name only is a bit lost beside the big avatar. 

font-size-wise, i would pick up the same as the old email address. avatar size etc. can stay as is.

reference implementation, including **screenshot** at https://github.com/deltachat/deltachat-ios/pull/2665

thanks @adbenitez for the hint at https://github.com/deltachat/deltachat-android/pull/3730#issuecomment-2797112301

Simon-Laux:
closes #4978

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,7 @@
 - hide 'show classic email' for chatmail, move down otherwise #4902
 - profile view redesign #4897
 - update translations (2025-04-09)
-- show signature/bio in settings
+- show signature/bio in settings #4984
 
 ### Fixed
 - tauri: improve security #4826, #4936, #4937, #4944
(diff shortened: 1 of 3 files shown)


r10s (approved):
nice! thanks a lot <3

PR #4985, opened 2025-04-12
Privacy Page in Settings

Simon-Laux:
- **add `autodel_server_enabled_hint` hint to autodeletion settings**
- **add Privacy Settings screen**
- **move context protection from experimental settings to privacy settings**

As discussed with adb, this pr adds a new privacy settings page.
He already did that in arcane chat and it makes sense to find the privacy settings more easily.

We might want to wait with merging this until there are corresponding prs on android and ios, so we can sync this change and also update FAQ where nessesary.

Bildschirmfoto 2025-04-12 um 20 53 23

---

Bildschirmfoto 2025-04-12 um 20 57 47
Bildschirmfoto 2025-04-12 um 20 58 02


Simon-Laux:
r10s told me that things are still unclear and we should postpone this after multi transport is done

r10s:
thanks for postponing.

thing is that it is not _that_ clear - eg. "delete from device" is not privacy on telegram, and may also be considered  together with "auto download", which is not privacy. and then there is "advanced", so one might need "advanced / privacy" or "privacy / advaced" sooner or later - signal went that way, and i have seen ppl struggling to find things more than one time.

i am not totally against that, however, i am not up to think about that now. letting it sneak in, with potentially making things different, but not much better.
there are so many ongoings, we do not need another moving target :)

PR #4986, opened 2025-04-13
Ask for microphone permission on Mac

### LINKED ISSUE #4988: Mac does not allow microphone access
when running a local build Mac asks for microphone permission. (shown as "Terminal wants access")

But when downloading a prebuilt dmg (even the one of #4986 ) it does not ask.

But it also does not show an error (no record input)

This may be related to the fact that the Dev-Build has less permissions per default (it could only be opened by twice ignoring the Security warning for a not certified app)

nicodh:
New runtime methods to ask for media access

resolves #4988

### DIFF
--- a/packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
+++ b/packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
@@ -93,25 +93,27 @@ export const AudioRecorder = ({
   const tx = useTranslationFunction()
   const [volume, setVolume] = useState(0)
   const recorder = useRef<MicRecorder | null>(null)
-  const  { openDialog } = useDialog()
+  const { openDialog } = useDialog()
+
+  const onAccessDenied = useCallback(() => {
+    openDialog(AlertDialog, {
+      message: tx('perm_explain_access_to_mic_denied'),
+    })
+  }, [openDialog, tx])
 
   const onRecordingStart = async () => {
     const access = await runtime.checkMediaAccess('microphone')
     if (access === 'not-determined') {
       const accessAllowed = await runtime.askForMediaAccess('microphone')
       if (!accessAllowed) {
-        openDialog(AlertDialog, {
-          message: 'Permission needed for recording voice messages'
-        })
+        onAccessDenied()
         return
       }
     } else if (access === 'denied') {
-      openDialog(AlertDialog, {
-        message: 'You need to set permission in system settings!'
-      })
+      onAccessDenied()
       return
     }
-    
+
     if (!recorder.current) {
       recorder.current = new MicRecorder(setVolume, {
         bitRate: 128,
--- a/packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
+++ b/packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
@@ -2,6 +2,7 @@ import React, { useCallback, useEffect, useRef, useState } from 'react'
 import MicRecorder from './MicRecorder'
 import styles from './styles.module.scss'
 import useTranslationFunction from '../../hooks/useTranslationFunction'
+import { runtime } from '@deltachat-desktop/runtime-interface'
 
 export enum AudioErrorType {
   'NO_INPUT',
@@ -91,7 +92,11 @@ export const AudioRecorder = ({
   const [volume, setVolume] = useState(0)
   const recorder = useRef<MicRecorder | null>(null)
 
-  const onRecordingStart = () => {
+  const onRecordingStart = async () => {
+    const access = await runtime.checkMediaAccess('microphone')
+    console.log('microphone access', access)
+    const t = await runtime.askForMediaAccess('microphone')
+    console.log(t)
     if (!recorder.current) {
       recorder.current = new MicRecorder(setVolume, {
         bitRate: 128,
--- a/packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
+++ b/packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
@@ -102,16 +102,24 @@ export const AudioRecorder = ({
   }, [openDialog, tx])
 
   const onRecordingStart = async () => {
-    const access = await runtime.checkMediaAccess('microphone')
-    if (access === 'not-determined') {
-      const accessAllowed = await runtime.askForMediaAccess('microphone')
-      if (!accessAllowed) {
+    let access = 'unknown'
+    try {
+      access = await runtime.checkMediaAccess('microphone')
+      if (access === 'denied') {
         onAccessDenied()
         return
       }
-    } else if (access === 'denied') {
-      onAccessDenied()
-      return
+      if (access === 'not-determined') {
+        // try to ask for access
+        const accessAllowed = await runtime.askForMediaAccess('microphone')
+        if (!accessAllowed) {
+          onAccessDenied()
+          return
+        }
+      }
+    } catch (e) {
+      // ignore errors thrown if platform does not
+      // support askForMediaAccess or checkMediaAccess
     }
 
     if (!recorder.current) {
--- a/packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
+++ b/packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
@@ -3,6 +3,8 @@ import MicRecorder from './MicRecorder'
 import styles from './styles.module.scss'
 import useTranslationFunction from '../../hooks/useTranslationFunction'
 import { runtime } from '@deltachat-desktop/runtime-interface'
+import useDialog from '../../hooks/dialog/useDialog'
+import AlertDialog from '../dialogs/AlertDialog'
 
 export enum AudioErrorType {
   'NO_INPUT',
@@ -91,12 +93,25 @@ export const AudioRecorder = ({
   const tx = useTranslationFunction()
   const [volume, setVolume] = useState(0)
   const recorder = useRef<MicRecorder | null>(null)
+  const  { openDialog } = useDialog()
 
   const onRecordingStart = async () => {
     const access = await runtime.checkMediaAccess('microphone')
-    console.log('microphone access', access)
-    const t = await runtime.askForMediaAccess('microphone')
-    console.log(t)
+    if (access === 'not-determined') {
+      const accessAllowed = await runtime.askForMediaAccess('microphone')
+      if (!accessAllowed) {
+        openDialog(AlertDialog, {
+          message: 'Permission needed for recording voice messages'
+        })
+        return
+      }
+    } else if (access === 'denied') {
+      openDialog(AlertDialog, {
+        message: 'You need to set permission in system settings!'
+      })
+      return
+    }
+    
     if (!recorder.current) {
       recorder.current = new MicRecorder(setVolume, {
         bitRate: 128,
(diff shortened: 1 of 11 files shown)


packages/target-electron/build/gen-electron-builder-config.js
```diff
@@ -100,7 +100,7 @@ build['mac'] = {
   entitlementsInherit: 'build/entitlements.mac.plist',
   extendInfo: {
     NSCameraUsageDescription: 'For scanning qr codes.',
-    // NSMicrophoneUsageDescription: "For voice messages",
+    NSMicrophoneUsageDescription: 'For recoring voice messages',
```
Simon-Laux, line 103:
```suggestion
    NSMicrophoneUsageDescription: 'For recording voice messages',
```

Simon-Laux:
converted to draft because of the testing commit: 7d2fa65b5a0df5ede773800ff4d877ba8144d7ca

nicodh:
I didn't change the camera handling for now but tested various scenarios on Mac and all worked fine so far.

Windows tests are still missing

packages/target-electron/src/ipc.ts
```diff
@@ -173,6 +174,34 @@ export async function init(cwd: string, logHandler: LogHandler) {
...
+      return systemPreferences.getMediaAccessStatus('camera')
+    } else if (mediaType === 'microphone') {
+      return systemPreferences.getMediaAccessStatus('microphone')
+    } else {
+      log.warn('checkMediaAccess: unknown media type', mediaType)
+      return 'unknown'
+    }
+  })
+
+  ipcMain.handle('askForMediaAccess', (_ev, mediaType: string) => {
```
WofWca, line 191: Isn't this supposed to return `boolean`, according to `runtime.ts`?

packages/target-electron/src/ipc.ts
```diff
@@ -173,6 +174,34 @@ export async function init(cwd: string, logHandler: LogHandler) {
     ev.returnValue = app.getPath(arg)
   })
 
+  ipcMain.handle('checkMediaAccess', (_ev, mediaType: string) => {
```
WofWca, line 177: A type annotation here would be nice as well.

WofWca (commented):
It works on Windows on Electron. The code looks sensible, also adds the Mac-specific manifest (or whatever it's called) keys.

Simon-Laux:
FTR: I have a commit in the deep linking pr that also adds the manifest and entitlement permissions for  macOS to the tauri target https://github.com/deltachat/deltachat-desktop/pull/4956/files#diff-20b627c7d30466aab2f5ee10c5ea2a7a138702d5d414c1faee1047bac0a61dca

packages/target-electron/src/ipc.ts
nicodh: I changed the implementation. askForMediaAccess throws an error now instead return "unknown", when not supported on the current platform

packages/target-electron/src/ipc.ts
nicodh: This is not really possible, since the return type of systemPreferences.getMediaAccessStatus is String and not MediaAccesStatus as expected
> Returns string - Can be not-determined, granted, denied, restricted or unknown.

I couldn't find a documentation about when it returns "restricted" so I only covered the "denied" case

nicodh:
For testing on Mac, download preview dmg and reset permissions with these commands:

`tccutil reset All chat.delta.desktop.electron.devbuild`

`codesign --force --deep --sign - /Applications/DeltaChat-DevBuild.app/Contents/MacOS/DeltaChat-DevBuild`

packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
```diff
@@ -90,8 +93,35 @@ export const AudioRecorder = ({
...
+      }
+      if (access === 'not-determined') {
+        // try to ask for access
+        const accessAllowed = await runtime.askForMediaAccess('microphone')
+        if (!accessAllowed) {
+          onAccessDenied()
+          return
+        }
+      }
+    } catch (e) {
```
WofWca, line 120: This is ignoring too many possible errors. For example, this suppresses "method not implemented" on Tauri. I'd prefer if it only wrapped the `await` part.

packages/target-electron/src/ipc.ts
WofWca: In that case the `runtime.ts` declaration of `checkMediaAccess` should also return a string.

packages/target-electron/src/ipc.ts
```diff
@@ -173,6 +175,41 @@ export async function init(cwd: string, logHandler: LogHandler) {
...
+   * https://www.electronjs.org/docs/latest/api/system-preferences#systempreferencesaskformediaaccessmediatype-macos
+   */
+  ipcMain.handle(
+    'askForMediaAccess',
+    (_ev, mediaType: MediaType): Promise<boolean> => {
+      if (!systemPreferences.askForMediaAccess) {
+        throw new Error('askForMediaAccess is not supported on this platform')
+      }
+      if (mediaType === 'camera') {
+        return systemPreferences.askForMediaAccess('camera')
```
WofWca, line 204: This is unused though, right? It'd be nice to remove it then, for security.

packages/target-electron/src/windows/main.ts
```diff
@@ -164,21 +159,11 @@ export function init(options: { hidden: boolean }) {
...
     return permission_handler(permission as any)
   })
   window.webContents.session.setPermissionRequestHandler(
     (_wc, permission, callback) => {
-      if (systemPreferences.askForMediaAccess && permission === 'media') {
-        systemPreferences.askForMediaAccess('camera').then(callback)
-      } else {
-        callback(permission_handler(permission))
-      }
+      callback(permission_handler(permission))
```
WofWca, line 166: I'm not sure what this change is about, but does the QR scanner still work on a fresh instance of DC on iOS? Now we don't seem to ever request camera permission.

WofWca:
This still works on Windows BTW.

packages/target-electron/src/windows/main.ts
nicodh: The camera is requested on the call
navigator.mediaDevices.getUserMedia({
       video: videoConstraints,
       audio: false,
 })
 
The mernissionHandler can not distinguish between camera access & microphone access so if you block camera access you also block microphone access - so it is not usable

packages/target-electron/src/ipc.ts
nicodh: I would prefer to keep it, otherwise we would have to change the runtime method signature to only support microphone which might not be sufficient for other use cases

packages/target-electron/src/ipc.ts
nicodh: I don't agree. The documentation says it can have one of these values but the type is string which means it is not correctly typed. Why should we keep that missing type definition.
Right now we can only handle " not-determined" and "denied" - maybe later more, but just having a string does not help in any ways I would say

packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
nicodh: If it wraps the await part if has to catch the error also. Do distinguish the errors we would have to define a custom notSupportedByPlatform error and only catch that silently. But I'm not sure if it is worth the effort. The error thrown in taui runtime is just a developer hint anyway. We won't go public before having implemented all missing runtime methods...

We could add a logger and log the effective error messages for debugging

packages/target-electron/src/windows/main.ts
WofWca: Huh, but microphone access does _not_ get requested on

https://github.com/deltachat/deltachat-desktop/blob/d887c6f078158fd91c7fc8386a38e1d511a83678/packages/frontend/src/components/AudioRecorder/MicRecorder.ts#L131

?

nicodh: There the user would be asked to access the microphone on first start if we didn't call askForMediaAccess('microphone') before. If he denies there he will be shown a message "Permission denied" and so any next time he tries to use the microphone.
Our permissionHandler would not help, since it can't distinguish between mic or camera access, it's just the "media" permission.

If the camera access is denied in QrReader we show always the explanation about the permission to access the camera instead of the video so the user is informed there. And we did ignore the case that the media permission could be for audio since it was only needed for camera so far.

Another question would be, how to disable the mic and/or camera access on Linux. Maybe we could implement an own handling for that?

packages/target-electron/src/ipc.ts
Simon-Laux: probably these values come from https://developer.apple.com/documentation/avfoundation/avauthorizationstatus#Status-Values

so restricted means "A status that indicates the app isn’t permitted to use media capture devices."

packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
Simon-Laux: I would prefer if it would just open an error dialog displaying the error in this case, of course additional to logging it.

BTW this is a perfect example use case for https://github.com/deltachat/deltachat-desktop/issues/3883

packages/target-electron/src/ipc.ts
nicodh: Well then the system would just show the message "Permission denied" which makes sense since the user can't change the permission

packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
nicodh: > I would prefer if it would just open an error dialog displaying the error in this case, of course additional to logging it.

What case you mean? "Not implemented"? "Not supported"? What should the user do about it?

Please don't forget the whole part starting in the try clause is to precheck if mic is accessible on Mac or Win and if not AND on Mac ask For it before the system does it anyway. Just do show a better undestandable error. So implementing custom errors here would only catch errors around systemPreferences immanent functions.

The actual mic access happens later and errors that happen there are shown to the user. 

Thats why I said " But I'm not sure if it is worth the effort" - especially speculating about possible future errors occuring in that part.

Of course we can do it but on that level there is also lot other stuff to do...

nicodh: Don't forget that Linux always throw on checkForMediaAccess and Linux & Win on askForMediaAccess

Simon-Laux: > Don't forget that Linux always throw on checkForMediaAccess and Linux & Win on askForMediaAccess

didn't knew that, then it makes probably sense to only log it for now.

nicodh: TO bring this discussion to an end I changed the return type one more time: now "Unknown" is returned if checkMediaAcess is not supported and "undefined" if askMediaAccess is not supported.

The only downside is now that we have to do a strict compare to distinguish the return value false from undefined ...

nicodh:
> i tested on a mac, it works as expected, thanks a lot!
> 
> however, the message for the missing permission is misleading, at least on sequoia 15.4.1, i cannot find the "app settings" or "permissions":
> 
> instead, i had to go to "System Settings / Privacy & Security / Microphone / Delta Chat"

That's a translation problem. If we have only one messages for all different desktop OS and the apps it should not be that specific

r10s:
> That's a translation problem. If we have only one messages for all different desktop OS and the apps it should not be that specific

agree, i edited to post above, sth. as 

> To send audio messages, enable "Microphone" in the system settings at "Permissions" or "Privacy & Security".

should fit many today and future systems

nicodh:
> what is the click path for windows / linux?

On windows you can either Allow/Deny Microphone access for all Applications or specific for each app. Privacy -> Windows permissions -> app permissions -> microphone 

On Linux I didn't find a way to block it so far!!
Maybe we should introduce an own flag for that in the settings?
But would be hard to make sure a "Deny" will be applied to all attempts to access the microphone...

PR #4987, opened 2025-04-13
Fix replace on null error

nicodh:
#skip-changelog

This failing e2e test should not have been ignored:

https://github.com/deltachat/deltachat-desktop/actions/runs/14422080442/job/40445918481


Simon-Laux (approved):

WofWca:
Why is it `null` though? The types say it can only be a string.

PR #4990, opened 2025-04-13
disable migration function, don't run it automatically on pnpm test anymore

Simon-Laux:
#skip-changelog


nicodh (approved):
LGTM
Although I would say versions since 2022 would be sufficient (reducing test data by ~10MB)

PR #4992, opened 2025-04-14
Fix wrong tx key

nicodh:
There is not key 'login'

#skip-changelog


WofWca (approved):
Indeed. I thought our type definitions would catch this. Turns out no.

PR #4995, opened 2025-04-14
fix: some untranslated strings, add tx type safety

WofWca:
This is a follow-up to https://github.com/deltachat/deltachat-desktop/commit/1c989ce0866b590e7e83b095c567ae652b4a64f1
(https://github.com/deltachat/deltachat-desktop/pull/4287).

Now it's not only about auto-completion, but also enforcing
that the key is present.
We have `./bin/test_for_missing_translations.sh`,
but we don't run it on every build, so sometimes we end up
specifying a wrong key.

One still can pass an arbitrary string to `tx` as such
`tx(myStr as any)`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,7 @@
 - fix member list is not refreshed if it changes while you look at the group profile #4894
 - remove unexpected horizontal scroll in gallery #4891
 - i18n: fix wrong order of substitutions for some strings #4889
+- i18n: translate some more strings
 - accessibility: don't announce "padlock" on messages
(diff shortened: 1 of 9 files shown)


Simon-Laux (approved):

Simon-Laux:
for me it says 
```
src/components/screens/WelcomeScreen/InstantOnboardingScreen.tsx(171,22): error TS2345: Argument of type '"login"' is not assignable to parameter of type 'TranslationKey'.
```
why is this not shown on the ci?

WofWca:
> why is this not shown on the ci?

Because it merged `main` into this MR's branch prior to running. The "login" string has been fixed on main recently.

WofWca:
I rebased

PR #4998, opened 2025-04-14
remove experimental option to disable IMAP IDLE

### LINKED ISSUE #4991: remove "disable idle"
While debugging a user reported issue it turned out the "disable IMAP idle" advanced settings was enabled leading to delays.  Android and iOS don't have this option I think so we can just remove it (and use IMAP idle unconditionally).

Simon-Laux:
closes #4991

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,7 +70,7 @@
 - fix sticker folder not resolved on windows #4939
 
 ### Removed
-- remove experimental option to disable IMAP IDLE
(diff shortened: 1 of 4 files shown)


nicodh (changes requested):
- disable_idle is still referenced in /packages/frontend/src/stores/settings.ts
- and the CoreSettingsSwitch import is obsolete in ExperimentalFeatures.tsx

nicodh (approved):

Issue #4989, opened 2025-04-13
Remove the dialog about AEAP


nicodh:
Remove the dialog about AEAP (which shows the aeap_explanation translated string), because AEAP will be removed. 

See: https://github.com/deltachat/deltachat-android/pull/3698

Simon-Laux:
shouldn't it be replaced by a message telling you that "AEAP is currently not available, but will come back better soon" or sth like that. Or have I missed some even newer discussions?

CC @Hocuri @r10s

Hocuri:
When you try to do AEAP, configure() will return an error message "Changing your email address is not supported right now. Check back in a few months!". When configure() returns an error message, the UI should show a dialog with the error message, anyway. So, there is no need for extra code in the UI for this.

r10s:
hm, but that affects new transport or jsonrpc only - cffi configure() has no return value.

anyways, i will update iOS soon to new api as well (just now released 1.58 uses the old configure() with core 1.159.1), apart from AEAP, which just times out, things seem to be fine

Simon-Laux:
> cffi configure() has no return value.

doesn't it have the error event on cffi? I think you use `dc_get_last_error` on cffi to get the configuration error, so the error message might be in there already?

r10s:
yip, there is the error event and `dc_get_last_error()`, this is what is done. still, could be that `configure()` internally has a return value now, that is not exposed to cffi, that was my question

r10s:
answering myself after looking at the code: indeed `configure()` has a return value internally, that is set when AEAP is failing; however, the error is only logged but _not_ exposed to to cffi via event, there seens to be a `progress!(self, 0, Some(error_msg.clone()));` missing in `add_transport_inner()`, or the return value needs to be set in cffi.

but not a big deal, nothing to do here. it is only about aeap, and iOS will switch to new transport-api anyway soon

nicodh:
Since #4999 changes of the mail address will show the message "Changing your email address is not supported right now. Check back in a few months!" and returns immediately without changing the address

PR #4999, opened 2025-04-14
Refactoring configure & instant onboarding

nicodh:
This PR cleans up some outdated comments and namings and moves the account creation call to ConfigureProgressDialog to avoid inconsistencies (like passing default Credentials in createInstantAccount after account creation). Also prepares to pass the proxy configuration before account creation 

 - move configurationQR to ConfigureProgressDialog
 - move defaultCredentials to ConfigureProgressDialog
 - move addTransportFromQr to ConfigureProgressDialog
 - add inline documentation
 - show new AEAP message on addr change #4989

#skip-changelog

### DIFF
--- a/packages/frontend/src/components/LoginForm.tsx
+++ b/packages/frontend/src/components/LoginForm.tsx
@@ -40,28 +40,6 @@ const CertificateChecks = {
   acceptInvalidCertificates: 'acceptInvalidCertificates',
 } as const satisfies { [P in T.EnteredCertificateChecks]: P }
 
-export function defaultCredentials(credentials?: Credentials): Credentials {
-  const defaultCredentials: Credentials = {
-    addr: '',
-    imapUser: null,
-    password: '',
-    imapServer: null,
-    imapPort: null,
-    imapSecurity: null,
-    certificateChecks: null,
-    smtpUser: null,
-    smtpPassword: null,
-    smtpServer: null,
-    smtpPort: null,
-    smtpSecurity: null,
-    oauth2: null,
-    proxyEnabled: false,
-    proxyUrl: null,
-  }
-
-  return { ...defaultCredentials, ...credentials }
-}
-
 type LoginProps = React.PropsWithChildren<{
   credentials: Credentials
   setCredentials: (credentials: Credentials) => void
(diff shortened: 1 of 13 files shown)


packages/frontend/src/components/LoginForm.tsx
```diff
@@ -40,28 +40,6 @@ const CertificateChecks = {
   acceptInvalidCertificates: 'acceptInvalidCertificates',
 } as const satisfies { [P in T.EnteredCertificateChecks]: P }
 
-export function defaultCredentials(credentials?: Credentials): Credentials {
```
Simon-Laux, line 43: why move this to the progress dialog? from what I understand it is only used for the login form

Simon-Laux: like core does not need them to be set to null. can we remove it if it is not used?

Simon-Laux (changes requested):
classical email login does not work with this.

fails with message "Multi transport is not supported right now. Check back in a few months!"

packages/frontend/src/components/LoginForm.tsx
Simon-Laux: let's better make a new file for it

packages/frontend/src/components/dialogs/ConfigureProgressDialog.tsx
```diff
@@ -71,27 +100,52 @@ export function ConfigureProgressDialog({
...
+          if (qrCode) {
+            // create a new transport for accountId based on the QR code
+            await BackendRemote.rpc.addTransportFromQr(accountId, qrCode)
+            // Select 'Device Messages' chat as the initial one. This will serve
+            // as a first introduction to the app after they've entered
+            const deviceChatId = await getDeviceChatId(accountId)
+            if (deviceChatId) {
+              await saveLastChatId(accountId, deviceChatId)
+              // SettingsStoreInstance is reloaded the first time the main screen is shown
+            }
```
Simon-Laux, line 129: if I understand the code flow correctly this won't be called for classic email login anymore?

nicodh: Is fixed

nicodh:
> classical email login does not work with this.
is fixed. There is always an empty default transport, what I didn't expect

> Might be time for an e2e test that checks the classical email login, as we don't test it ourselves so often anymore
I'm on it already

Simon-Laux (approved):
my comments were addressed, code looks good. didn't test

PR #5001, opened 2025-04-15
Onboarding e2e tests

nicodh:


WofWca:
Isn't this just because it can't click it because it's not clickable because a dialog is open?

nicodh:
> Isn't this just because it can't click it because it's not clickable because a dialog is open?

The item is searched in the open dialog. But the search is for the name Bob and so is not found

nicodh:
The question is: why is the user listed with mail address instead user name and why are there members with different avatars than the visible profiles?

WofWca:
Looking at the video, the contact only appears in the list of chats _after_ the dialog gets opened, I'd assume this is the reason, but I don't know if the name should be known at that point regardless

[92cb86ae1e13da45521cef4ad40622a9bc1ceab9.webm](https://github.com/user-attachments/assets/e268c14f-a8a3-4ee1-bdef-94357b131e3c)

nicodh:
And if a user has no image the avatar in the Group dialog shows the first character of "Me" which explains the "M" Avatar

nicodh:
Seems to be a timing issue. I local UI tests the tests is fine.
Maybe we have to send a message first

nicodh:
> Seems to be a timing issue. I local UI tests the tests is fine. Maybe we have to send a message first

That fixed the problem

docs/E2E-TESTING.md
```diff
@@ -10,6 +10,8 @@ and run `npx playwright install --with-deps`
 
 A convenient alternative to install and use playwright is the VSCode [plugin](https://playwright.dev/docs/getting-started-vscode). It also provides some functionality for running or recording tests.
 
+Copy packages/e2e-tests/\_env to packages/e2e-tests/.env
```
WofWca, line 13: It's a bit annoying that you have to do this manually, but alright.

WofWca (approved):
It's nice that the tests are now separated somewhat!

I haven't looked through every line, but this adds more tests, which is good.

I also looked for potential flakiness points, and this MR does not appear to introduce more of them.

PR #5003, opened 2025-04-16
Show EditProfile on empty displayname

### LINKED ISSUE #5002: Classic Email Onboarding can result in empty displayname
If you don't enter a displayname in InstantOnboardingScreen but click directly to "Use other server" => "Classic Email Login" and enter your credentials there, the "Login" button forwards to the Main screen in the new created account, although no displayname is set

nicodh:
resolves #5002

### DIFF
--- a/packages/frontend/src/components/screens/AccountSetupScreen.tsx
+++ b/packages/frontend/src/components/screens/AccountSetupScreen.tsx
@@ -18,6 +18,8 @@ import { DialogId } from '../../contexts/DialogContext'
 import AlertDialog from '../dialogs/AlertDialog'
 
 import type ScreenController from '../../ScreenController'
+import EditProfileDialog from '../dialogs/EditProfileDialog'
+import { useSettingsStore } from '../../stores/settings'
 
 export default function AccountSetupScreen({
   selectAccount,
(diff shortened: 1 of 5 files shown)


packages/frontend/src/components/screens/AccountSetupScreen.tsx
```diff
@@ -32,25 +34,31 @@ export default function AccountSetupScreen({
...
-          ),
-        proxyUpdated: credentials.proxyEnabled,
-      }),
-    [accountId, openDialog, selectAccount, credentials]
-  )
+  const onClickLogin = useCallback(() => {
+    openDialog(ConfigureProgressDialog, {
+      credentials,
+      onSuccess: async () => {
+        const profileComplete = await selectAccount(accountId)
```
nicodh, line 44: The same code could also go to https://github.com/deltachat/deltachat-desktop/blob/beebd9e4e4b8a43555368ef0fe810110bdbe7e77/packages/frontend/src/components/AccountListSidebar/AccountListSidebar.tsx#L65
Then we would show the EditProfile dialog on any account change if the selected account has no displayname 
🤔

nicodh:
As discussed in the issue, we won't force a displayname

PR #5004, opened 2025-04-16
go back to released tauri version

Simon-Laux:
they meanwhile have merged all our contributions.

Also update other rust dependencies.

note that this commit doesn't work yet - it looks like channels don't really work by default:

Bildschirmfoto 2025-04-16 um 19 21 25


WofWca:
I pushed an upgrade to JS Tauri dependencies, this fixes it.

PR #5008, opened 2025-04-17
fix: error message not being shown on RPC calls

WofWca:
The error thrown by `await BackendRemote.rpc...()` is not
`instanceof Error`, so we handled it incorrectly often times.

FYI there are a lot more `catch`es in our code base,
and I didn't re-check all of them.
This only fixes those that used `instanceof Error`.

Follow-up to https://github.com/deltachat/deltachat-desktop/pull/4981.

Related: https://github.com/deltachat/deltachat-desktop/issues/3883.


Simon-Laux (approved):
LGTM

PR #5015, opened 2025-04-21
Add Profile explain hint

### LINKED ISSUE #5014: show hint below editing self-profile
the string "Your profile image, name and bio will be sent together with your messages when communicating with other users." (`pref_who_can_see_profile_explain`) should be show below the input fields of the self-profile, cmp. android and iOS to get an idea:

the style should be the same as comparable hints at other places in desktop.

this issue is not about the wording, only to get the string in.

nicodh:
resolves #5014

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,6 +100,7 @@
 - add withdraw qr code context menu option #4798
 - adapt dialogs for small screens
 - mute chat for 8 hours
+- show hint how the profile is transferred #5014
 
 ### Changed
 - tauri: replace `tauri-plugin-shell` with `tauri-plugin-opener` #4699
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,7 +100,7 @@
 - add withdraw qr code context menu option #4798
 - adapt dialogs for small screens
 - mute chat for 8 hours
-- show hint how the profile is transferred #5014
+- add a hint to "Edit profile" about how the profile name, image and bio is transferred #5014
 
 ### Changed
 - tauri: replace `tauri-plugin-shell` with `tauri-plugin-opener` #4699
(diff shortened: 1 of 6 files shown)


packages/frontend/themes/_themebase.scss
```diff
@@ -184,6 +184,7 @@ $hover-contrast-change: 2%;
   --globalLinkColor: #2090ea; // Manual value: This will not get generated in the future
   --globalBackground: #{$bgPrimary};
   --globalText: #{$textPrimary};
+  --textLessContrast: #{$colorLessContrast}; // less contrast to background than global text
```
nicodh, line 187: Introducing a new generic text color with a meaningful name should avoid to have too many similar colors. So anytime you want to have less contrast for a text you can use this variable. We could also rename it to hintTextColor to be more specific (?)
TBD: this is in contrast to the suggestion in https://github.com/deltachat/deltachat-desktop/issues/3018
Which says all css variable should be per component only.
That would end up in having too many variables imho

WofWca (commented):
The contrast is indeed low

image

Otherwise looks OK.

packages/frontend/themes/dark.scss
```diff
@@ -29,6 +29,7 @@ $textSecondary: #ddd;
 $btnPrimaryText: #fff;
 $btnDangerText: #fff;
 $btnDangerBackground: $colorDanger;
+$colorLessContrast: ##898989;
```
WofWca, line 32:
```suggestion
$colorLessContrast: #898989;
```

CHANGELOG.md
```diff
@@ -100,6 +100,7 @@
 - add withdraw qr code context menu option #4798
 - adapt dialogs for small screens
 - mute chat for 8 hours
+- show hint how the profile is transferred #5014
```
WofWca, line 103:
```suggestion
- add a hint to "Edit profile" about how the profile name, image and bio is transferred #5014
```

WofWca (commented):
image

Apart from the syntax error, it's good, the contrast is acceptable now.

WofWca (approved):

PR #5016, opened 2025-04-21
Avoid react warning and speed up rendering

nicodh:
Avoid RadioGroup.tsx:21 Warning: Each child in a list should have a unique "key" prop.

#skip-changelog


WofWca (approved):
Makes sense

Issue #4743, opened 2025-03-05
Issues from Accessibility audit


Simon-Laux:
We recently got an accessibility quick scan from HAN. They took some time to discuss/test the app with us in a call and gave us a report. treefit and wofwca also made notes during the call, this issues combines points from our notes and the report they gave us. The order is roughly chronological based on the call. (might make sense to reorder and group it into areas of concern, later).

- [x] Hierarchy predictability issue. So user can move logically through the chats. They say we have 3 sections. "accounts", "chats" "messages". From search we need to focus the list of chats
	- [2.4.3 Focus Order](https://www.w3.org/WAI/WCAG22/Understanding/focus-order)
	- Visual sections in the top bar (https://github.com/deltachat/deltachat-desktop/pull/4672)
	- 2025-05-04: Further improvements, turn the sections into landmarks: https://github.com/deltachat/deltachat-desktop/pull/5067

- [x] unread badge says "1", should say "1 new message" (https://github.com/deltachat/deltachat-desktop/pull/4655)

- [x] Text in text avatar should not be read
	- https://github.com/deltachat/deltachat-desktop/pull/4661

- [x] Chat nav header (button that opens chat profile) should announce a bit more what it is (like "chat/group with name" instead of just chatname)??
	- 2025-08-04 https://github.com/deltachat/deltachat-desktop/pull/5067 should address this. It will announce "Chat with Bob section: view profile button"

- [ ] QR code button should have a link purpose
	- (announce what it does/what it is for, that it opens a dialog)
	- [2.4.4 link purpose in the standard](https://www.w3.org/WAI/WCAG22/Understanding/link-purpose-in-context)
	- Should say what you'd do with the QR code. The link purpose should be clear to the user.

- [ ] QR Code should have a useful alt/label that is not just the filename
	- [x] QR invite https://github.com/deltachat/deltachat-desktop/pull/5038
	- [x] Own secure join group qr code (qr code button)
	- [x] group join code
	- [ ] add second device code

- [ ] QR dialog: "scan to chat with" should be the header probably

- [x] Gallery: Tabs are not selectable via keyboard (https://github.com/deltachat/deltachat-desktop/pull/4675)
	They use "roving tabindex", which was the case when the audit was performed. Maybe they expected "Tab" to work there. Should we remove roving tabindex?

- [x] Gallery not sure how to escape it. -> wofwca made a tab
   Edit: also see https://github.com/deltachat/deltachat-desktop/issues/5074

[...]

PR #5021, opened 2025-04-22
improvement: a11y: `aria-live` for message status

WofWca:
This should somewhat address this point in https://github.com/deltachat/deltachat-desktop/issues/4743

> Screenreader does not announce that it is successfully sent


nicodh (approved):
Review by testing

PR #5023, opened 2025-04-22
improvement: a11y: announce msg "edited" change

WofWca:
This will only announce the "edited" status,
and not the new text of the message.

This also includes a refactor, see individual commits for clarity.

I have checked that the layout didn't change, at least in basic cases.

### DIFF
--- a/packages/frontend/src/components/message/MessageMetaData.tsx
+++ b/packages/frontend/src/components/message/MessageMetaData.tsx
@@ -50,6 +50,9 @@ export default function MessageMetaData(props: Props) {
         ),
       })}
     >
+      {/* FYI the padlock doesn't need `aria-live`
+      as we don't expect it to get removed. See
+      https://github.com/deltachat/deltachat-desktop/pull/5023#discussion_r2059382983 */}
       {padlock && (
         <div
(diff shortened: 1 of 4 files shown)


packages/frontend/src/components/message/MessageMetaData.tsx
```diff
@@ -50,6 +50,8 @@ export default function MessageMetaData(props: Props) {
         ),
       })}
     >
+      {/* TODO a11y: add `aria-live` in case a padlock gets removed?
+      e.g. when a message gets edited, but the edit is not encrypted. */}
```
Simon-Laux, line 54: not sure if you can even send an unencrypted edit to an encrypted message. I don't think this todo is necessary.

CC @r10s

r10s: yip, it is not possible to send unencrypted edit to encrypted message. so, padlock cannot be removed by edits

WofWca: Thanks for checking this out!

Simon-Laux (approved):

PR #5025, opened 2025-04-22
fix: a11y: `aria-setsize="-1"` for message list

WofWca:
It would be nice to also apply this to our virtualized lists,
but this requires modifying the library:
https://github.com/bvaughn/react-window/issues/808.

FYI NVDA still announces the number of messages, which, I suppose, is an NVDA bug.


WofWca:
Waaait, we're supposed to apply this attribute on individual items, and not the container.

WofWca:
Hmmm, maybe just setting a different role on the `<ol>` would be better.

PR #5028, opened 2025-04-23
improvement: a11y: search: separate tabindex divs

WofWca:
Previously, all the search results, i.e. contacts, chats, messages,
were a part of the same "roving tabindex" widget.
This is perhaps not intuitive, because they're in different sections.
Let's fix that, and make each section separately tabbable to.

For easier review, turn off "show whitespace".

I have checked that it works.


nicodh (approved):
Review by testing

PR #5029, opened 2025-04-23
Migrate frontend/scss files

### LINKED ISSUE #4942: SCSS deprecation: replace @import with @use
This will remove global variables, which we rely on in multiple places, so we need to do:
- [ ] ensure we use (runtime) css variables (`var(--variables)`) more and avoid scss $variables - atleast in the frontend
- [ ] convert everything in the project to use `@use` or `@forward`
- [ ] convert `_themebase.scss` to a mixin / css function

To get rid of the deprecation warnings and to future proof the code base to be ready for the scss updated that removes `@import`.

I discovered this when I updated sass in #4940 

whoever does this could also try to do #3018 afterwards, because it is also about variables.

nicodh:
resolves #4942 

This PR just refactors the deprecated @import statements. The proposed restructuring of theming in #3018 should go to a separate PR. It's hard anyway to make sure to avoid unintended side effects. 

#skip-changelog

### DIFF
--- a/packages/frontend/scss/_connectivity-toast.scss
+++ b/packages/frontend/scss/_connectivity-toast.scss
@@ -1,3 +1,5 @@
+@use 'mixins';
+
 .ConnectivityToast {
   position: fixed;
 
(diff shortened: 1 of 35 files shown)


nicodh:
Some values in the both _dev themes are now hard coded instead referencing variables. Do we need these themes at all anymore?

nicodh:
Not to be merged before release 1.58.0

WofWca (approved):

PR #5030, opened 2025-04-23
improvement: add `aria-label(ledby)` for lists

WofWca:
So that screen readers announce "messages list" or "chats list"
instead of just "list" when you focus an item inside of it.

For easier review, turn on "hide whitespace".

TODO:
- [x] Merge the MR this is based on.
- [x] Add a changelog entry.

### DIFF
--- a/packages/frontend/src/components/chat/ChatList.tsx
+++ b/packages/frontend/src/components/chat/ChatList.tsx
@@ -5,6 +5,8 @@ import React, {
   useCallback,
   ComponentType,
   useMemo,
+  HTMLAttributes,
+  useLayoutEffect,
 } from 'react'
 import {
   FixedSizeList as List,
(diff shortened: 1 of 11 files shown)


packages/frontend/src/components/chat/ChatList.tsx
```diff
@@ -419,6 +456,18 @@ export default function ChatList(props: {
             >
               <div ref={tabindexWrapperElementChats}>
                 <ChatListPart
+                  olElementAttrs={{
+                    // TODO perhaps `pref_` is not nice,
+                    // we might need a separate string.
+                    // The same goes for other occurrences
+                    // of `tx('pref_chats')`.
```
Simon-Laux, line 463: @r10s may have an answer for this question.

Simon-Laux (approved):

PR #5031, opened 2025-04-23
Prepare release 1.58.0

nicodh:


WofWca (approved):
It's about time for a release! 
Note that we again go straight to 1.58 instead of 1.57.

PR #5032, opened 2025-04-23
Fallback for displayName to address

### LINKED ISSUE #5022: what to show as `webxdc.selfName` show if `displayname` is not set
in the happy path, we enforce setting the displayname, but there still may be situations where it is unset, eg. when onboarding with a classic email address, we do not enforce setting for [$reasons](https://github.com/deltachat/deltachat-desktop/issues/5002) - but even if that is different, it still may happen. also, we should do the same to have a pattern for other webxdc hosts.

currently all three delta chat's do different things 😅

- android shows the email address 

- desktop shows selfAddr - which breaks layout of most webxdc as it is far too long and really looks bad

- iOS shows text "unknown"

i think, we should do the same on all OS, and i think, the android approach is reasonable. also at other places, we use the address if the name is unknown. this was also the situation before we changed selfAddr calculation and no one has really complained. as said, it is anyway a cornercase, webxdc cannot send the address somewhere nor can corellate reasonably, and if the user sets a name, things are fine

if we anyway do changes in that area, we could also provide [self_name in webxdc-info](https://c.delta.chat/classdc__msg__t.html#aee526949983962ed25adf33a2f5e6878) (tbh, i thought it is like that :)

nicodh:
resolves #5022 

#skip-changelog

### DIFF
--- a/packages/frontend/src/system-integration/webxdc.ts
+++ b/packages/frontend/src/system-integration/webxdc.ts
@@ -6,18 +6,30 @@ import { BackendRemote, Type } from '../backend-com'
 import { runtime } from '@deltachat-desktop/runtime-interface'
 
 export function initWebxdc() {
-  BackendRemote.on('WebxdcStatusUpdate', (accountId: number, { msgId }: { msgId: number}) => {
-    runtime.notifyWebxdcStatusUpdate(accountId, msgId)
-  })
-  BackendRemote.on('WebxdcRealtimeData', (accountId: number, { msgId, data }: { msgId: number, data: any}) => {
-    runtime.notifyWebxdcRealtimeData(accountId, msgId, data)
-  })
-  BackendRemote.on('MsgsChanged', (accountId: number, { msgId }: { msgId: number}) => {
-    runtime.notifyWebxdcMessageChanged(accountId, msgId)
-  })
-  BackendRemote.on('WebxdcInstanceDeleted', (accountId: number, { msgId }: { msgId: number}) => {
-    runtime.notifyWebxdcInstanceDeleted(accountId, msgId)
-  })
+  BackendRemote.on(
+    'WebxdcStatusUpdate',
+    (accountId: number, { msgId }: { msgId: number }) => {
+      runtime.notifyWebxdcStatusUpdate(accountId, msgId)
+    }
+  )
+  BackendRemote.on(
+    'WebxdcRealtimeData',
+    (accountId: number, { msgId, data }: { msgId: number; data: any }) => {
+      runtime.notifyWebxdcRealtimeData(accountId, msgId, data)
+    }
+  )
+  BackendRemote.on(
+    'MsgsChanged',
+    (accountId: number, { msgId }: { msgId: number }) => {
+      runtime.notifyWebxdcMessageChanged(accountId, msgId)
+    }
+  )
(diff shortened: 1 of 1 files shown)


packages/frontend/src/system-integration/webxdc.ts
```diff
@@ -6,18 +6,30 @@ import { BackendRemote, Type } from '../backend-com'
...
+    (accountId: number, { msgId }: { msgId: number }) => {
+      runtime.notifyWebxdcMessageChanged(accountId, msgId)
+    }
+  )
+  BackendRemote.on(
+    'WebxdcInstanceDeleted',
+    (accountId: number, { msgId }: { msgId: number }) => {
+      runtime.notifyWebxdcInstanceDeleted(accountId, msgId)
+    }
+  )
```
WofWca, line 32: What is this about? Why do we have to convert `data` to `any`?

packages/frontend/src/system-integration/webxdc.ts
```diff
@@ -44,10 +56,10 @@ export async function internalOpenWebxdc(
...
     await BackendRemote.rpc.getBasicChatInfo(accountId, message.chatId)
   ).name
-  const displayname = await BackendRemote.rpc.getConfig(
-    accountId,
-    'displayname'
-  )
+  const account: Type.Account =
+    await BackendRemote.rpc.getAccountInfo(accountId)
+  const displayname =
+    account.kind === 'Configured' ? account.displayName || account.addr : ''
```
WofWca, line 62:
```suggestion
    account.kind === 'Configured' ? account.displayName || account.addr : null
```

packages/frontend/src/system-integration/webxdc.ts
```diff
@@ -44,10 +56,10 @@ export async function internalOpenWebxdc(
   const chatName = (
     await BackendRemote.rpc.getBasicChatInfo(accountId, message.chatId)
   ).name
-  const displayname = await BackendRemote.rpc.getConfig(
-    accountId,
-    'displayname'
-  )
+  const account: Type.Account =
+    await BackendRemote.rpc.getAccountInfo(accountId)
```
WofWca, line 60: It would be nice to only do the backend call only if the `displayname` is empty.

WofWca (approved):
Looks like this does close the issue, although I haven't tested this with an empty `displayname`.

packages/frontend/src/system-integration/webxdc.ts
nicodh: Hmm, looks as if linting in my VSCode was broken yesterday night and tricked me to add types. Just removed it again.

packages/frontend/src/system-integration/webxdc.ts
nicodh: But that would need 2 calls, the getConfig call and a second if displayName is empty

packages/frontend/src/system-integration/webxdc.ts
nicodh: Changed it

packages/frontend/src/system-integration/webxdc.ts
WofWca: Ohh, yes, you're right. I missed the fact that `getConfig` is async.

PR #5034, opened 2025-04-24
refactor: fix log error: `undefined.getStackTrace`

WofWca:
#skip-changelog

I have not tested this, but it should work.


Simon-Laux:
may make sense to do this in the constructor of the logger class? so that you don't have to think about doing it everywhere manually

WofWca:
Umm, I don't think you can do that? AFAIK `bind` specifies the value of `this`, which is an _instance_ of a class.

Another way is just `(...args) => log.error(...args)`.

nicodh (approved):
Did not test but LGTM

PR #5039, opened 2025-04-24
refactor: remove unused `hover` var

WofWca:
Hover state is now managed by CSS.

#skip-changelog


Simon-Laux (changes requested):
wasn't this for the case that you do open a context menu that the hover effect stays so you still see what chat the context menu is for?

WofWca:
I'm looking at the MR that introduced the variable (https://github.com/deltachat/deltachat-desktop/pull/3318), and it turns out it has been unused from the very start.

WofWca:
And I checked that the "hover" state still works on this MR

image

Simon-Laux (approved):

PR #5040, opened 2025-04-24
fix: a11y: `role='tablist'` for accounts list

WofWca:
This should make it more clear that those are not just some
"buttons". UX-wise these are very close to "tabs",
so let's mark them as such.

This also adds `aria-selected`.

This, however, doesn't add the corresponding `role='tabpanel'`
element (see the comment).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Changed
 
 ### Fixed
+- accessibility: add `role='tablist'` for accounts list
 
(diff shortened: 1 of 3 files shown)


nicodh (approved):
Can't say much about the effect of this change but it seems reasonable

WofWca:
Have you tested this with a screen reader? NVDA now announces the number of accounts and says "tab control" instead of just "button".

PR #5041, opened 2025-04-24
improvement: a11y: mark chat list items as tabs

WofWca:
Conceptually they are tabs, and not just abstract buttons,
because they control what's displayed
in the "chat" ("messages") section.

TODO:
- [ ] merge the MRs that this depends on (all except the last commit)


WofWca:
Note about making this work with multiselect: I think that in the default, non-multiselect mode we should stick to `aria-selected`, [as W3 recommends](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/), but when the multiselect mode gets activated, we should also apply `aria-current` to the currently active chat.
I have not seen documentation oh how this should be handled specifically.

Alternative: adjust multiselect code such that the active chat gets unselected when we enter multiselect mode.

Simon-Laux (approved):
LGTM

Issue #5046, opened 2025-04-26
Sender partially hidden


Raiden-GH:
After saving an image/video to "Saved Messages," the sender's name is partially hidden.

macOS 15.4.1
DC-Desktop 1.57

![Image](https://github.com/user-attachments/assets/46733af4-cd2a-476d-a8bd-b2b04029e7a4)

WofWca:
Similar to https://github.com/deltachat/deltachat-desktop/issues/3914

PR #5047, opened 2025-04-27
fix: ugly margins on some attachments

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/5046.

This has to do with the `withContentAbove` variable
(i.e. the `.content-above` class) not being in sync
with reality, i.e. `showAuthor`, etc.
So, let's nuke the variable.

This also fixes or makes other cases more consistent,
namely with forwarded messages, e.g. forwarded image messages,
where we wouldn't apply `content-above` to the attachment element,
relying instead on the `.forwarded` selector in `_message.scss`.

Note that I haven't checked all the possible combinations
of the variables that affect the presence of "content above".
Since the bug that this fixes is not a regression,
this probably shouldn't go into 1.58.

Related:
- https://github.com/deltachat/deltachat-desktop/issues/3914.


packages/frontend/scss/message/_message-attachment.scss
```diff
@@ -79,7 +79,18 @@
     border-bottom-right-radius: 0px;
   }
 
-  &.content-above {
+  // "Has content above" selector
+  // (i.e. author title, "Forwarded" indicator, quote).
+  // FYI there are duplicates of this code.
```
nicodh, line 84: Wouldn't that be a reason to use a mixin here?

```
// Mixin to apply styles when an element has visible preceding content
// (i.e. author title, "Forwarded" indicator, quote).
//
// Note that `.can-hide` means `display: none;`
// only on non-experimental themes.
// But TBH we don't care about the other themes too much.
@mixin has-content-above {
  // Has a preceding element
  *:not(.can-hide) + &,
  // Has a parent that has a preceding element
  .msg-container *:not(.can-hide) + * & {
    @content;
  }
}
```
and then just

```
@include has-content-above {
    margin-top: 4px;
    border-top-left-radius: 0px;
    border-top-right-radius: 0px;
  }
  ```

**Review by nicodh — APPROVED** [review]
Review by testing & reading

Added somes suggestions

File: CHANGELOG.md [hunk 2]
```diff
@@ -8,6 +8,7 @@
 
 ### Fixed
 - accessibility: announce when a message gets edited
+- ugly margins / paddings on some saved messages with attachments
```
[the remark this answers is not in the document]
WofWca: Adding MR requires another `git commit` (too much hassle). About issue number - I guess if we're OK with this approach, I'll start adding it. But then still, some MRs don't resolve any issue.

So IDK, do we want to keep this tradition? These numbers are not linkified anyway, and you can find the related commits with `git blame`.

packages/frontend/scss/message/_message-attachment.scss
WofWca: Very nice, I didn't know of this syntax! I'll rewrite it.

WofWca:
Waiting until 1.58 release

Issue #4590, opened 2025-01-31
a11y: stop constantly re-focusing composer


WofWca:
- Operating System (Linux/Mac/Windows/iOS/Android):
- Delta Chat Version: 1.52.1
- Expected behavior: `document.activeElement` remains where the user wants / expects it
- Actual behavior: The composer gets forcefully `.focus()`ed in a lot of cases
- Steps to reproduce the problem:
  1. Tab-navigate to the "QR" button in search.
  2. Press "Space" to open the QR dialog.
  3. Press Escape to close the dialog. This full re-focus the composer
- Screenshots:
- Logs:

At the very least we want to start with an option that would stop this behavior.

A possible in-between solution would be to make pressing Escape focus the composer. This is intuitive IMO. Unless we want Escape to close the current chat, like in Telegram (but then we can use Ctrl + W for this, like in WhatsApp).

WofWca:
I found another problem with this. After you send a message, the composer also gets re-focused (after getting disabled and then un-disabled), and the screen reader re-announces it.
This is gonna be really noticeable when we start announcing new incoming messages (see https://github.com/deltachat/deltachat-desktop/issues/4743). We would want to only announce the conversation itself so it feels like a conversation, and this issue stands in the way.

WofWca:
This is really annoying when working in keyboard-only mode and interacting with dialogs or context menus, e.g. in the chat list or the message list.

fcr--:
The problem of the current state of delta.chat is now that whenever I select the window, unless I click in a few very specific places of the window, the composer will lose focus.  And even though it would be intuitive, typing text on a keyboard does not currently refocus the composer.

WofWca:
Indeed, that is unfortunate. Do you have a suggestion on how this can be fixed without sacrificing accessibility?

nicodh:
Closing for now. If there is a strong demand and we decide to work on it we can reopen it

PR #5049, opened 2025-04-27
fix: a11y: dont disable composer while sending

WofWca:
The primary point of this change is to remove
`textarea.focus()`, which makes screen readers re-announce
the composer, which is disorienting and pointless.

Partially addresses
https://github.com/deltachat/deltachat-desktop/issues/4590#issuecomment-2821985528.

This change has side-effects:
- Now if the user starts typing right after sending a message,
  we will not ignore it, as we would if the composer was disabled,
  which is good.
- However, as a consequence, now it's possible to type
  while a message is being sent,
  which might have unexpected consequences itself,
  i.e. race conditions, i.e. bugs.

This also ~~adds an~~ improves the alert dialog that pops up if sending fails.

"hide whitespace" should make this easier to review.

I have tested this for a bit, and it appears to work OK. I have tried causing some race conditions, but I haven't seen any that seem too important.

TODO:
- [x] Merge https://github.com/deltachat/deltachat-desktop/pull/5048, which this MR is based on.

### DIFF
--- a/packages/frontend/src/components/composer/Composer.tsx
+++ b/packages/frontend/src/components/composer/Composer.tsx
@@ -76,6 +76,8 @@ const Composer = forwardRef<
     ) => Promise<void>
     removeFile: () => void
     clearDraftStateButKeepTextareaValue: () => void
+    clearDraftStateAndUpdateTextareaValue: () => void
+    setDraftStateAndUpdateTextareaValue: (newValue: DraftObject) => void
   }
 >((props, ref) => {
   const {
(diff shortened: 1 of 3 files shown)


WofWca:
This would also solve https://github.com/deltachat/deltachat-desktop/issues/5101.

nicodh:
Can you please rebase this? Then I will review it asap

WofWca:
Rebased

packages/frontend/src/components/composer/Composer.tsx
```diff
@@ -210,49 +212,68 @@ const Composer = forwardRef<
         if (textareaRef) {
           if (textareaRef.disabled) {
             throw new Error(
-              'text area is disabled, this means it is either already sending or loading the draft'
+              'text area is disabled, this means it is loading the draft'
```
nicodh, line 215: I just tested: these errors are not handled so the user has no clue what's going on, he just can't send (that was already the case before this PR but here another scenario is added to throw an error in a maybe not too unexpected use case )

So maybe wrap more of this function in a try catch instead only catching errors that occur in sendMessagePromise

WofWca: > but here another scenario is added to throw an error in a maybe not too unexpected use case

Umm, what's that other scenario?

nicodh: Ah ok, it's just a wording update. So not part of this PR but should be handled in another one imo

WofWca:
Sorry about the CHANGELOG mess... Let me smuggle the fix in this MR.

PR #5052, opened 2025-04-28
Configure proxy dialog

### LINKED ISSUE #4994: Add proxy settings before onboarding
It must be possible to add proxy settings before onboarding, in case the proxy xsettings are needed for the initial onboarding.

The dialog for entering proxy configuration should be accessible from a 3 dot menu on Login screen:

Image

The dialog should look similar to Android:

new-proxy-settings

nicodh:
resolves #4994 

 - new dialog fory multi proxy configuration
 - reachable via three dot menu in InstantOnboarding Screen or in Advanced Settings
 - new three-dot-menu header button  for dialog headers
 - remove proxy configuration from settings dialog
 - enable proxy configuration before onboarding

image


packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+      activeProxy,
+    })
+  }
+
+  const addProxy = async (proxyUrl: string) => {
+    if (proxyState.proxies.includes(proxyUrl)) {
+      log.warn('skip already existing proxy', proxyUrl)
+      // proxy alread exists, should we show a message here?
+      return
+    }
```
WofWca, line 142:
```suggestion
    if (proxyState.proxies.includes(proxyUrl)) {
      log.warn('skip already existing proxy', proxyUrl)
      setShowNewProxyForm(false)
      setNewProxyUrl('')
      // proxy alread exists, should we show a message here?
      return
    }
```

packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+            </Button>
+          </>
+        )}
+        {!showNewProxyForm && (
+          <Button
+            className={styles.addProxyButton}
+            onClick={() => setShowNewProxyForm(true)}
+            styling='secondary'
+          >
+            ＋
```
WofWca, line 369: What is this plus? It's not announced by NVDA. Why not use the regular ASCII one?

packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+              value={newProxyUrl}
+              onChange={e => setNewProxyUrl(e.target.value)}
+            />
+            <Button
+              className='save-proxy'
+              onClick={() => addProxy(newProxyUrl)}
+            >
+              {tx('proxy_add')}
+            </Button>
+          </>
```
WofWca, line 361: It would be nice to make this an actual `<form>` so that you can submit with "Enter". But this is fine I guess.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## [Unreleased][unreleased]
 
 ### Added
+  - new proxy configuration dialog #5052
```
WofWca, line 6:
```suggestion
- new proxy configuration dialog #5052
```

packages/frontend/src/components/dialogs/ProxyConfiguration/styles.module.scss
```diff
@@ -0,0 +1,74 @@
+.proxyDialogBody {
+  display: flex;
+}
+
+.proxyList {
+  flex-grow: 1;
+  overflow-y: scroll;
```
WofWca, line 7: You can't see saved proxies on not-so-tall screens

image

I suggest to make the entire dialog body scrollable instead. And perhaps also move the "add proxy" section to the dialog body.

packages/frontend/src/components/dialogs/ProxyConfiguration/ProxyItemRow.tsx
```diff
@@ -0,0 +1,125 @@
...
+function getConnectivityString(status: number) {
+  if (status >= C.DC_CONNECTIVITY_CONNECTED) {
+    return 'connectivity_connected'
+  }
+  if (status >= C.DC_CONNECTIVITY_CONNECTING) {
+    return 'connectivity_connecting'
+  }
+  if (status >= C.DC_CONNECTIVITY_WORKING) {
+    return 'connectivity_updating'
+  }
```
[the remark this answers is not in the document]
nicodh: > condition is unreachable

Good catch!

> On a side note, it seems like we have too many places with similar code.

Yes, thought the same - maybe move this to Connectivity Toast in another PR

packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
nicodh: Not sure if that is expected behaviour. It would look like as if it worked. I think it would be better to disable the button in that case. Just tested and seems we need some new disabled colors for primary & secondary buttons...

WofWca: Well, arguably it did work.

Otherwise I'd vote for showing an error message.

packages/frontend/src/components/dialogs/ProxyConfiguration/styles.module.scss
nicodh: If the whole body is scrollable it might result in a hidden add button or a hidden enabe/disable button if you have multiple proxies. That seems more problematic to me than the edge case of screen height < 400px

packages/frontend/src/components/dialogs/ProxyConfiguration/ProxyItemRow.tsx
```diff
@@ -0,0 +1,125 @@
...
+        role='radio'
+        aria-checked={isActive}
+        onKeyDown={e => {
+          if (e.key === ' ' || e.key === 'Enter') {
+            e.preventDefault()
+            changeActiveProxy(proxyUrl)
+          }
+        }}
+      >
+        <input
```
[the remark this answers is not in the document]
nicodh: Changed it to activate Use Proxy

packages/frontend/src/components/dialogs/ProxyConfiguration/ProxyItemRow.tsx
```diff
@@ -0,0 +1,125 @@
...
+          id={`proxy-${index}`}
+          name='proxy-selection'
+          type='radio'
+          onChange={() => changeActiveProxy(proxyUrl)}
+          value={proxyUrl}
+          checked={isActive}
+          className={styles.radioButton}
+          aria-labelledby={`proxy-label-${index}`}
```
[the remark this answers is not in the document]
nicodh: Added an id to the div holding the label

packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+
+  const addProxy = async (proxyUrl: string) => {
+    if (proxyState.proxies.includes(proxyUrl)) {
+      log.warn('skip already existing proxy', proxyUrl)
+      // proxy alread exists, should we show a message here?
+      return
+    }
+    let proxyValid = proxyUrl.length > 0
+    if (proxyValid) {
+      const parsedUrl = await BackendRemote.rpc.checkQr(accountId, proxyUrl)
```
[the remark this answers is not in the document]
nicodh: Now it chates the error and shows the error message in the dialog

packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+            ...prev,
+            enabled,
+            proxies,
+            activeProxy,
+          }))
+        }
+      } catch (error: any) {
+        log.error('failed to load proxy settings', error)
+        openAlertDialog({
+          message: `${tx('error')} ${error.message || error.toString()}`,
```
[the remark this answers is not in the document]
nicodh: Uses unknownErrorToString now

packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
nicodh: Button is disabled now

packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
nicodh: The ASCII plus looks bad here. I added an aria-label attribute

packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
nicodh: I changed it by just using a form tag which has no additional attributes. Not sure what the impact on accessibility is here...

WofWca: Yes, now "Enter" works. Although it'd be nice to add `type="submit"` on the button for good measure.

packages/frontend/src/components/dialogs/ProxyConfiguration/styles.module.scss
WofWca: Can one call it "hidden" though? You can scroll to it. And if you already have proxies, you already know that there is an "add" button.
The way it is right now, it's just not possible to see the proxies on short window size, no matter how you scroll.

packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
```diff
@@ -0,0 +1,403 @@
...
+  // and is not already in the list of proxies
+  const maybeValidProxyUrl = (url: string): boolean => {
+    const parts = url.split('://')
+    return (
+      parts.length === 2 &&
+      parts[0].length >= 2 && // shortest protocol is ss://
+      parts[1].length >= 1 && // host
+      !proxyState.proxies.includes(url)
+    )
+  }
```
WofWca, line 324: IDK about adding this frontend validation. This should be a job for the back end. What if we add support for a format that does not conform to this pattern?

packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
```diff
@@ -0,0 +1,403 @@
...
+            <DeltaInput
+              label={tx('proxy_add_url_hint')}
+              value={newProxyUrl}
+              onChange={e => setNewProxyUrl(e.target.value)}
+            />
+            <Button
+              className='save-proxy'
+              onClick={() => addProxy(newProxyUrl)}
+              styling='primary'
+              disabled={!maybeValidProxyUrl(newProxyUrl)}
```
WofWca, line 378: I think this should be a validation error on the proxy input (`setCustomValidity`) as opposed to the button being disabled with no indication of what's wrong.

WofWca (approved):
OK, the remaining issues are non-blockers IMO. I might make an MR with my vision.

Great work!

packages/frontend/src/components/Dialog/DialogHeader.tsx
```diff
@@ -9,24 +9,54 @@ import EditButton from './EditButton'
...
+          <Button
+            className={classNames(styles.headerThreeDotButton)}
+            aria-label={tx('menu_more_options')}
+            data-testid={dataTestid + '-context-menu'}
+            onClick={onContextMenuClick}
+          >
+            <Icon coloring='contextMenu' icon='more' rotation={90} size={24} />
+          </Button>
+        </span>
+      )}
```
WofWca, line 59: One cannot set up a proxy when importing a backup and adding a second device though.

packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
```diff
@@ -0,0 +1,403 @@
...
+        onClose={closeDialog}
+        dataTestid='proxy-settings'
+      />
+      <DialogBody className={styles.proxyDialogBody}>
+        <div className={styles.container}>
+          <SettingsSwitch
+            label={tx('proxy_use_proxy')}
+            value={proxyState.enabled}
+            onChange={changeProxyEnable}
+          />
```
WofWca, line 344: When you open the dialog for the first time, you have no proxies, and when you toggle this on, you get an error message right away. Not a nice experience.
FYI Android opens the "add proxy" dialog in such a case.

WofWca (commented):
I have also tested the flow of adding a proxy on onboarding. It works.

packages/frontend/src/components/dialogs/ProxyConfiguration/styles.module.scss
nicodh: Maybe we add a media query for that edge case?

packages/frontend/src/components/screens/WelcomeScreen/InstantOnboardingScreen.tsx
```diff
@@ -18,11 +18,33 @@ import useDialog from '../../../hooks/dialog/useDialog'
...
+import ProxyConfiguration from '../../dialogs/ProxyConfiguration'
+import { selectedAccountId } from '../../../ScreenController'
+
 type Props = {
   onCancel: () => void
   selectedAccountId: number
 }
 
+function buildContextMenu(openDialog: OpenDialog) {
+  const tx = window.static_translate // don't use the i18n context here for now as this component is inefficient (rendered one menu for every message)
```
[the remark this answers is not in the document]
nicodh: It's just copied  from another buildContextMenu call. Will clean it...

packages/frontend/src/components/dialogs/ProxyConfiguration/styles.module.scss
nicodh: Added one for screen height <= 500px

packages/frontend/src/components/dialogs/ProxyConfiguration/styles.module.scss
```diff
@@ -0,0 +1,74 @@
...
+  border-bottom: 1px solid var(--separatorColor);
+  padding: 10px 0 10px 10px;
+  &:hover {
+    background-color: var(--globalHoverBg);
+  }
+}
+
+.explain {
+  margin-right: 20px;
+  margin-top: 10px;
```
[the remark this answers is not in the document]
nicodh: Adapted

packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
nicodh: Well you could also argue that you can't enable proxy if you have not yet entered a proxy URL. We could hide the enable button and only show it after you added a proxy URL.

WofWca: Well sure, logically you can't, but the UX is not the best IMO. (I don't think this issue is a reason not to merge the MR BTW).

packages/frontend/src/components/Dialog/DialogHeader.tsx
nicodh: Since the settings are also imported from backup it would not make sense to set it before importing a backup.

Adding as second device might work if you add the proxy settings and then go back to the initial screen top open the Add as second device dialog.

The behaviour is just copied from Android atm

WofWca: > Since the settings are also imported from backup it would not make sense to set it before importing a backup.

Hmm, I didn't know that, interesting.

OK, let's mark as resolved.

PR #5057, opened 2025-04-30
Add missing disabled styles

nicodh:
#skip-changelog

before: disabled secondary and primary button

image

after

image


nicodh:
Needed for #5052

WofWca (approved):
Appearance-wise looks alright (I guess a good contrast is not required for disabled buttons), but code-wise, the `:disabled` style is already defined a few lines above, I think this should be just a matter of moving that style to below `.primary` and `.secondary` so that it takes precedence.

PR #5058, opened 2025-04-30
Reduce bitRate to 32kb/s

### LINKED ISSUE #4977: Reduce filesize of voice messages
It should be possible to tweak the Encoder to use a stronger compression and thus a smaller file size.

Just passing smaller bitRate & sampleRate has no effect atm

nicodh:
resolves #4977

As discussed before, 32 bitrate seems to be a reasonable value

Filesize is now ~240 kB per minute


r10s (approved):

WofWca (approved):
The quality seems fine to me, I tested it.

Simon-Laux (commented):
code LGTM, didn't test

PR #5060, opened 2025-04-30
Avoid stickers to be cropped in height

nicodh:
- no min-width for stickers (to avoid height crop)
 - no border radius for stickers
 - no max-width for images in portrait mode

#skip-changelog

### DIFF
--- a/packages/frontend/scss/message/_message-attachment.scss
+++ b/packages/frontend/scss/message/_message-attachment.scss
@@ -31,6 +31,9 @@
     object-position: center;
     max-width: 450px;
     min-width: 200px;
+    &.sticker {
+      min-width: auto;
+    }
 
     // The padding on the bottom of the bubble produces 4 extra pixels of space at the
     //   bottom, so this doesn't match up with the padding numbers above.
@@ -208,4 +211,9 @@
       margin-top: 3px;
     }
   }
+}
+
+// no border radius for stickers
+.type-sticker .message-attachment-media {
+  border-radius: 0px;
 }
--- a/packages/frontend/scss/message/_message-attachment.scss
+++ b/packages/frontend/scss/message/_message-attachment.scss
@@ -209,8 +209,3 @@
     }
   }
 }
-
-// no border radius for stickers
-.type-sticker .message-attachment-media {
-  border-radius: 0px;
-}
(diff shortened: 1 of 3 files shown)


adbenitez (approved):

packages/frontend/scss/message/_message-attachment.scss
```diff
@@ -209,3 +209,8 @@
     }
   }
 }
+
+// no border radius for stickers
+.type-sticker .message-attachment-media {
+  border-radius: 0px;
+}
```
WofWca, line 216: We already have a place with this selector

https://github.com/deltachat/deltachat-desktop/blob/ff6b346ed393c28b3d73a562bf7e3ba9807ea72a/packages/frontend/scss/message/_message.scss#L299

Would be nice not to spread related code over multiple files.

packages/frontend/scss/message/_message-attachment.scss
```diff
@@ -31,6 +31,9 @@
     object-position: center;
     max-width: 450px;
     min-width: 200px;
+    &.sticker {
+      min-width: auto;
+    }
```
WofWca, line 36: I think this can also be moved to here, making the `message.viewType === 'Sticker' ? 'sticker' : null` unnecessary.

https://github.com/deltachat/deltachat-desktop/blob/ff6b346ed393c28b3d73a562bf7e3ba9807ea72a/packages/frontend/scss/message/_message.scss#L299-L309

WofWca (commented):
I haven't checked all possible image sizes, but this fixes the reported cases of stickers being weird.

Note that images still do get cropped at the sides on narrower window sizes, which IMO is unfortunate for Desktop.

packages/frontend/scss/message/_message-attachment.scss
nicodh: Turned out there is still an edge case: on screens smaller 900px the message width is smaller than the image min-width so we need to set the width to "auto" but not for stickers, since they are limited to 200px height

packages/frontend/scss/message/_message.scss
```diff
@@ -409,6 +411,13 @@
     &.outgoing {
       margin-left: auto;
     }
+    /**
+     * on small screens the limiting factor is the
+     * width not the height (except for stickers)
+     */
+    img.portrait:not(.sticker) {
```
WofWca, line 418: Again, `.message:not(.type-sticker)` can be used here it seems. But alright.

WofWca (approved):

packages/frontend/scss/message/_message.scss
nicodh: .type-sticker is a class on the higher order message item, not on the image

PR #5061, opened 2025-05-01
tauri: refactor: reduce code nesting with let else

WofWca:
Try to shut up about early returns challenge failed.

#skip-changelog

Marked as draft only in order to not steal attention,
but it's actually ready for review, if you do have time for stuff like this.


WofWca:
A note about conflicts: it's OK to just "accept current" in each case and abandon my changes.

Simon-Laux (approved):
Makes sense. I like that it reduces nesting.

Issue #3321, opened 2023-07-20
webxdc sendUpdate does not work in onclose/visibility changed event handler on closing the app


Simon-Laux:
webxdc sendUpdate does not work in onclose/visibility changed even handler on closing the app.

```
window.on("visibilitychange", (ev)  => {
 window.webxdc.sendUpdate(update, descr);
})
```
In this case desktop does not manage to send the update. I suspect it is because sendUpdate uses the async electron ipc `invoke`-function.

I tried using a the synchronous ipc functions (without changing that sendUpdate returns a promise, it uses jsonrpc under the hood in the back-end and JS has no `::block_on(future/promise)` function, so the promise is necessary), but I was not successful in my first try (but I was also very tired and had a hard time concentrating, so maybe I could still make it work, but I don't see very high chances.)

At the moment I would recommend the following workaround, but I haven't tested if closing the window works with standard browser apis in our case:
```
window.on("visibilitychange", (ev)  => {
 ev.preventDefault(); // prevent the closing of the window
 await window.webxdc.sendUpdate(update, descr);
 // somehow close window via JS //maybe with window.close() ?
})
```

Maybe it could also make sense to add some close callback api to the webxdc spec, where you can set a custom close handler that is run on closing and returns a promise, when it is resolved or after a timeout the app is then closed by the host app (in our case DC).

WofWca:
Also see
* https://codeberg.org/webxdc/editor/commit/9d56ab4167ad9d72f79baecde50401dc036bc0dd
* https://github.com/webxdc/webxdc-test/commit/4add6f73af6b526028b20c3a9ed607210068510e

WofWca:
Something notable. I'm trying https://github.com/webxdc/webxdc-test/commit/4add6f73af6b526028b20c3a9ed607210068510e, and it appears that `sendUpdate` works inside of `beforeunload` even without `event.preventDefault()`.

![image](https://github.com/deltachat/deltachat-desktop/assets/39462442/4443dfd9-44ac-4088-919a-ac9f0cc6bb68)

Simon-Laux:
> So the workaround would be to add a beforeunload listener besides the visibilitychange listener?

Maybe would need a way to prevent sending duplicate updates?

r10s:
> So the workaround would be to add a beforeunload listener besides the visibilitychange listener?

[...]

PR #5065, opened 2025-05-01
tauri: fix: webxdc: sendUpdate() in visibilitychange

WofWca:
...and `beforeunload` and `pagehide`.

Related:
- https://github.com/deltachat/deltachat-desktop/issues/3321.

TODO:
- [ ] Test on macOS
- [x] Test on Linux
- [x] Merge https://github.com/deltachat/deltachat-desktop/pull/5062


WofWca:
According to @Simon-Laux, this doesn't work on macOS.
I might investigate later, but I guess this can be merged to fix things on Windows, unless we're afraid of negative side-effects.

Simon-Laux (approved):
works on linux. so only does not work on macOS. this is already an improvement so we can merge it and mention that it does not work yet on Mac in #4468

PR #5066, opened 2025-05-01
tauri: fix: don't use `process::exit`, return instead

WofWca:
This is to make sure that all the `drop()` that need to be called
do get called.

I don't know whether this has a practical effect,
but let's do it to be sure.

#skip-changelog because it's not clear whether there is a user-visible change.

### DIFF
--- a/packages/target-tauri/src-tauri/src/main.rs
+++ b/packages/target-tauri/src-tauri/src/main.rs
@@ -1,6 +1,10 @@
 // Prevents additional console window on Windows in release, DO NOT REMOVE!!
 #![cfg_attr(not(debug_assertions), windows_subsystem = "windows")]
 
-fn main() {
-    deltachat_tauri_lib::run()
+fn main() -> std::process::ExitCode {
+    let exit_code_i32 = deltachat_tauri_lib::run();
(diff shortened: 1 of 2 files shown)


packages/target-tauri/src-tauri/src/main.rs
```diff
@@ -2,5 +2,11 @@
 #![cfg_attr(not(debug_assertions), windows_subsystem = "windows")]
 
 fn main() {
```
Simon-Laux, line 4: I'd guess that this function can just return the exit code?

packages/target-tauri/src-tauri/src/main.rs
```diff
@@ -2,5 +2,11 @@
...
-    deltachat_tauri_lib::run()
+    let exit_code = deltachat_tauri_lib::run();
+    if exit_code != 0 {
+        // We could probably just run this regardless of the code,
+        // but let's be on the safe side and ensure
+        // that all `drop()`s are run on normal exit.
+        std::process::exit(exit_code);
+    }
```
Simon-Laux, line 11: as far as I understand rust drop was already run at this point.

we could also return an exit code: https://doc.rust-lang.org/std/process/struct.ExitCode.html

Simon-Laux (approved):

PR #5067, opened 2025-05-02
fix: a11y: add landmarks

WofWca:
As per https://www.w3.org/WAI/ARIA/apg/practices/landmark-regions/
> Including all perceivable content on a page
> in one of its landmark regions and giving each landmark region
> a semantically meaningful role is one of the most effective ways
> of ensuring assistive technology users
> will not overlook information that is relevant to their needs.

This should make the general layout of the app much easier to navigate. The accessible structure is now much more close to the actual one (related: https://github.com/deltachat/deltachat-desktop/pull/4672).

See individual commits for more details.
If this is too much to review, the commits can be merged one-by-one in sequence. However, later commits depend on previous commits, they change the same areas.

Turn on "hide whitespace" for easier review.

TODO:
- [x] Merge https://github.com/deltachat/deltachat-desktop/pull/5030.
- [ ] (can be done in a separate MR) Look for other elements and dedicate them to appropriate landmarks.
   - [ ] Message list, with its header.


WofWca:
I have reviewed the changes since the rebase. They still work and make sense (at least for the most part).

nicodh:
Same for this PR. Please rebase then I will review it asap

WofWca:
Rebased.

Edit: let me fix the error.

WofWca:
OK, I fixed the type error and adjusted the E2E tests for the changes.

nicodh (approved):
Reviewed by testing

I also read the changes but there are too many parts that need more expertise in accessibility handling, like when it's appropriate to use section instead div, remove a nav or a wrap etc.
Thanks to WofWca that you care for that!

So on the accessibility level I can't really evaluate but I can confirm that the PR doesn't break things as far as I can see.

WofWca:
> like when it's appropriate to use section instead div

`<section>` implies `role='region'`, and `role='region'` with `aria-label` will make screen readers announce the label of the region when focus enters the region. This is the basic idea of landmarks.
For more, see the link in the MR description.

Basically the way to test this is to use a screen-reader and just keep pressing tab on the main screen of the app.

Issue #4993, opened 2025-04-14
`selfAddr` is the same for all webxdcs


Septias:
- Operating System (Linux/Mac/Windows/iOS/Android): Nixos
- Delta Chat Version: 34926b2b9
- Expected behavior: The webxdc's `selfAddr` should be different for all webxdcs because it depends on the unique rfc724mid
- Actual behavior: The webxdc's `selfAddr` is the same for all webxdcs
- Steps to reproduce the problem: Send two instances of the testing xdc [webxdc-test-1.1.1.zip](https://github.com/user-attachments/files/19735091/webxdc-test-1.1.1.zip) into saved messages. Then, restart Delta Chat and compare the selfAddr in both instances. They are the same.

WofWca:
Isn't this a core issue? There is not a lot of logic in Desktop around `selfAddr`, we're just taking the value of `webxdcInfo.selfAddr`:

https://github.com/deltachat/deltachat-desktop/blob/f6b780d24e1d455a0fda93ad819fd7793fafcaa6/packages/target-electron/src/deltachat/webxdc.ts#L132

Septias:
It might be a core issue but the tests do not reproduce it. Maybe I have to create one where core actually restarts or something alike.

WofWca:
Indeed I can reproduce this. I sent two instances of "Test Webxdc" to saved messages and `selfAddr` inside of them is shown to be the same.

WofWca:
Hm, however, `(await exp.rpc.getMessage(acc, msg)).webxdcInfo.selfAddr` returns different IDs. So this seems desktop-specific.

WofWca:
This looks like a duplicate of https://github.com/deltachat/deltachat-desktop/issues/4401.

Issue #4401, opened 2024-12-10
1.49 sometimes reports wrong `webxdc.self_addr` 


r10s:
in a recent poll, i encountered, that i could vote _two_ times from the same profile but different devices.

i adapted the webxdc-test to show the info at https://github.com/webxdc/webxdc-test/pull/31 - and yes, my iOS reports different selfAddr than desktop for my r10s-nine-account:

<img width=400 src=https://github.com/user-attachments/assets/cd2df1bb-990e-42f8-9556-c75248488a96> <img width=290 src=https://github.com/user-attachments/assets/fe0bb3f5-cefc-4812-97c0-653bd81cdb5e><br><i>(left: desktop, right: iOS)</i>

corresponding code and discussion is at https://github.com/deltachat/deltachat-core-rust/pull/6237

for the r10s-nine-account, this is reproducible. on the same devices, i have a nine-test-account, there things are in sync.

so, wondering if we missed sth. on calulating selfAddr as sha256(fingerprint+rfc724_mid)  maybe fingerprint is returned in different casing or so  - or if that is only me

r10s:
i checked on current deltachat-desktop main using core 1.151.5, there i am getting on desktop the same selfAddr as on iOS - and _not_ always the same.

so, we can close this issue here, still i do not get how that could happen, feels a bit fragile without knowing what was going on (maybe somehow rfc-message-id was set wrong or empty in the message-object only in jsonrpc, idk)

PR #5068, opened 2025-05-03
fix: webxdc: same `selfAddr` on different webxdcs

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/4993,
and likely https://github.com/deltachat/deltachat-desktop/issues/4401.

The issue is that we register the `webxdc:` scheme protocol handler
once per account, when you open the first webxdc, and when we do,
we incorrectly use the captured `webxdcInfo` var inside the handler,
resulting in us using `selfAddr` of the first opened webxdc app
for all the subsequently opened apps.

The fix is to get `selfAddr` from the respective webxdcInfo object
for each app instance.
This commit also does the same for other `webxdcInfo` properties,
such as `sendUpdateInterval`.

`selfAddr` used to be the email address, so this used to be fine,
but then `selfAddr` started to be different between messages
(https://github.com/deltachat/deltachat-desktop/issues/4401).

I have confirmed that this fixes the bug.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Changed
 
 ### Fixed
+- webxdc apps sometimes having wrong `selfAddr`, resulting in apps treating the same user as a new one (e.g. the "Poll" app would allow you to vote twice) #5068
 - accessibility: add accessible labels and descriptions to more items #5050, #5055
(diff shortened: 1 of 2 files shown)


Simon-Laux (approved):

WofWca:
Do we want to merge now? That would mean that it goes into 1.58.0? I'd vote "yes".

PR #5082, opened 2025-05-06
Always use a fix height

nicodh:
Height for images should never be "auto" since the message height might be wrong calculated. So the fix to avoid width cropped images is to reduce the height for small screens to 300px instead 450px. The cropping then only happens on screen width <= 380px which is fine I would say

#skip-changelog

### DIFF
--- a/packages/frontend/scss/message/_message.scss
+++ b/packages/frontend/scss/message/_message.scss
@@ -416,7 +416,7 @@
      * width not the height (except for stickers)
      */
     img.portrait:not(.sticker) {
-      height: auto;
+      height: 300px;
     }
   }
 }


packages/frontend/scss/message/_message.scss
```diff
@@ -416,7 +416,7 @@
      * width not the height (except for stickers)
```
WofWca, line 416: Does the comment still make sense?

WofWca (approved):
I don't know how the new image sizes work, but this fixes the bug with the chat being scrolled up right after you opened it (see https://github.com/deltachat/deltachat-desktop/issues/4404).

Also I'm not sure whether this is the size that we want, some might say 300px is too small. for a portrait image.

WofWca:
> The cropping then only happens on screen width <= 380px which is fine I would say

Is this about some specific case, or all cropping in general? Because this doesn't seem to be true. This image starts getting cropped at 780px window size and below

image

nicodh:
> > The cropping then only happens on screen width <= 380px which is fine I would say
> 
> Is this about some specific case, or all cropping in general? Because this doesn't seem to be true. This image starts getting cropped at 780px window size and below

The rule is only for Portrait images. In smallScreenMode  (width < BREAKPOINT_FOR_SMALLSCREEN_MODE 720px) the chatlist is hidden and the message list is wider again so there might be a gap between 780px and 720px where cropping starts.

I think it's hard to ge full responsiveness with images when at the same time using only fix css heights. I think we cover most cases "roughly" now, Of course more media queries could provide a more granular rendering...

PR #5083, opened 2025-05-06
Small css cleanups

nicodh:
Best to be reviewed commit by commit

#skip-changelog

### DIFF
--- a/packages/frontend/scss/message/_message.scss
+++ b/packages/frontend/scss/message/_message.scss
@@ -388,10 +388,6 @@
   .message {
     max-width: 100%;
 
-    /**
-     * on small screens the limiting factor is the
-     * width not the height (except for stickers)
-     */
     img.portrait:not(.sticker) {
       height: 300px;
     }
--- a/packages/frontend/scss/message/_message.scss
+++ b/packages/frontend/scss/message/_message.scss
@@ -173,7 +173,6 @@
 
 .message.incoming {
   margin-left: 0;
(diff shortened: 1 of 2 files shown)


packages/frontend/scss/message/_message.scss
```diff
@@ -385,36 +384,10 @@
   }
 }
 
-// Spec: container > 593px
-@media (min-width: 926px) {
```
nicodh, line 389: It seems we had 2 styles in media queries that were applied to all screens < 925px and also for screens > 926px
I'm not sure if that was for purpose.

@Simon-Laux seems it was introduced quite some time ago:
https://github.com/deltachat/deltachat-desktop/commit/5952a2058999f90303152718120d56c13316361f

Simon-Laux: can't remember exactly. also it is older than the linked commit.

WofWca:
This is not just a refactor, right?

nicodh:
> This is not just a refactor, right?

Well it does not introduce new styles but just consolidates existing ones.
The only real "change" is: margin-right/margin-left was never applied (since overriden by both media queries) - so margin:auto is not needed anymore and was also removed

WofWca (approved):
Makes sense. Based on a bit of testing, this does not affect behavior. The code also doesn't seem to suggest so. So this is just a refactor, and a good one.

BTW, it seems like margins are in fact completely unused.

PR #5085, opened 2025-05-06
Make secondary button style default

### LINKED ISSUE #5045: some alerts do not use the new primary/secondary buttons
while alerts eg. for "delete confirmation" use the new primary/secondary style, the alert used to confirm attaching multiple files/images do not:

Image

(the send button should be blue-ish and the cancel button grey and a border, just like at comparable places)

maybe there are more alerts that do have the "old" buttons, i just came over did, i did not search exhaustively

(v1.57.0 on mac)

nicodh:
resolves #5045

Just 2 usages of the Button component needed an extra style, since they were not "real" buttons:
- navbar buttons on main
- password preview button on login

#skip-changelog

### DIFF
--- a/packages/frontend/src/components/Button/style.module.scss
+++ b/packages/frontend/src/components/Button/style.module.scss
@@ -48,9 +48,6 @@
   &.borderless {
     background-color: transparent;
     border: none;
-    &:hover {
-      background-color: transparent;
-    }
   }
 
   &.active {
--- a/packages/frontend/src/components/Button/style.module.scss
+++ b/packages/frontend/src/components/Button/style.module.scss
@@ -30,6 +30,7 @@
   &.primary {
     background-color: var(--buttonPrimaryBackground);
     color: var(--buttonPrimaryText);
+    border: var(--buttonPrimaryBackground);
     &:hover {
       background-color: var(--buttonPrimaryHover);
     }
@@ -38,6 +39,7 @@
   &.danger {
     color: var(--buttonDangerText);
     background-color: var(--buttonDangerBackground);
+    border: var(--buttonPrimaryBackground);
     &:hover {
       background-color: var(--buttonDangerHover);
     }
@@ -58,5 +60,8 @@
   &:disabled {
     background-color: var(--buttonDisabledBackground);
     color: var(--buttonDisabledText);
+    &:hover {
+      background-color: var(--buttonDisabledBackground);
+    }
   }
 }
--- a/packages/frontend/src/components/Button/style.module.scss
+++ b/packages/frontend/src/components/Button/style.module.scss
@@ -1,9 +1,9 @@
 .button {
   align-items: center;
-  background-color: transparent;
+  background-color: var(--buttonSecondaryBackground);
   border-radius: 10px;
-  border: 0;
-  color: var(--buttonText);
+  color: var(--buttonSecondaryText);
+  border: 1px solid #ccc;
   cursor: pointer;
   display: flex;
   font-size: 16px;
@@ -14,7 +14,7 @@
   user-select: none;
 
   &:hover {
-    background-color: var(--buttonHover);
+    background-color: var(--buttonSecondaryHover);
   }
 
   &:disabled {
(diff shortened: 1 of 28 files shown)


packages/frontend/src/components/Button/style.module.scss
```diff
@@ -1,9 +1,9 @@
 .button {
   align-items: center;
-  background-color: transparent;
+  background-color: var(--buttonSecondaryBackground);
   border-radius: 10px;
-  border: 0;
-  color: var(--buttonText);
+  color: var(--buttonSecondaryText);
+  border: 1px solid #ccc;
```
WofWca, line 6: This applies the border to primary buttons as well, which IMO looks a bit weird. But not a big deal I guess.

image

packages/frontend/src/components/Button/style.module.scss
```diff
@@ -35,15 +35,6 @@
     }
   }
 
-  &.secondary {
-    background-color: var(--buttonSecondaryBackground);
-    color: var(--buttonSecondaryText);
-    border: 1px solid #ccc;
-    &:hover {
-      background-color: var(--buttonSecondaryHover);
```
WofWca, line 43: Just FYI, hover disabled primary buttons look a little weird, but this is not about this MR.

image

packages/frontend/src/components/Button/style.module.scss
WofWca, line 6: Three-dot button in "create new profile" has a border, which seems unintentional

image

WofWca, line 6: Also take a look a proxy. I think we need a margin there.

image

WofWca, line 6: And the same from "go back from Archive".

image

I recommend going through all the references of the Button component and seeing how they look.

WofWca (changes requested):
The idea of the MR is probably good.
But some buttons got quite ugly IMO.

I am not sure though if this is good to just override the styles of the button in CSS, instead of passing a prop to it to remove the border and stuff.
In those cases it perhaps makes more sense to use a raw `<button>`. But for now I guess a prop to revert to the behavior that we used to have prior to this MR will do, if we don't want to rewrite this much stuff right now.

Also wouldn't it make sense to remove the 'secondary' option altogether then?

https://github.com/deltachat/deltachat-desktop/blob/13355b05753878cd20e63b7d30f71faa6fdba661/packages/frontend/src/components/Button/index.tsx#L8

Or make it clear that it's the default here

https://github.com/deltachat/deltachat-desktop/blob/13355b05753878cd20e63b7d30f71faa6fdba661/packages/frontend/src/components/Button/index.tsx#L23

nicodh:
> I am not sure though if this is good to just override the styles of the button in CSS, instead of passing a prop to it to remove the border and stuff.

I introduced a new styling "borderless" for the cases where a button should not look like a button


> 
> Also wouldn't it make sense to remove the 'secondary' option altogether then?

Is removed now

packages/frontend/src/components/Button/style.module.scss
nicodh: Since the proxy component was introduced later than this PR it was not adapted. I applied the new borlderless styling now where needed

packages/frontend/src/components/Button/style.module.scss
nicodh: is fixed

packages/frontend/src/components/Button/style.module.scss
nicodh: Is fixed

nicodh: is fixed

packages/frontend/src/components/Button/style.module.scss
```diff
@@ -30,34 +30,38 @@
...
       background-color: var(--buttonDangerHover);
     }
   }
 
+  &.borderless {
+    background-color: transparent;
+    border: none;
+    &:hover {
+      background-color: transparent;
+    }
```
WofWca, line 53:
```suggestion
```

This removes hover color from the buttons, e.g. the archive button. I don't think we want that?

WofWca (approved):
OK, I have checked the buttons that I have checked before, and they look as before (apart from hover). But I haven't checked all of them.
The code also looks good.

PR #5088, opened 2025-05-07
Add migration-tests to packageignore_list

nicodh:
New builds were 35mb bigger than 1.56.0 since the migration-tests were not excluded

#skip-changelog


nicodh:
> Makes sense.
> 
> Though I suspect we have some other unintentionally included files there.

Yes, maybe we should have a look at this list soon

PR #5091, opened 2025-05-08
Add electron to flake

Septias:
Add electron to flake, this is needed because the npm electron package can not be run natively on nixos.
To make this work, electron bin has to be deleted `rm ./packages/target-electron/node_modules/.bin/electron`

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -46,6 +46,7 @@
           cargo
           cargo-tauri
           nodejs
+          electron
         ];
 
         buildInputs = with pkgs; [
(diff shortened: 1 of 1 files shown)


Simon-Laux:
> To make this work, electron bin has to be deleted rm ./packages/target-electron/node_modules/.bin/electron

can't this be also done in the flake?

Or alternatively could the flake construct the FHS environment that electron needs?

Septias:
Best this would be to build packages with nix instead of npm or pnpm, but that is a bit overkill IMO. This is just supposed to be a quick fix, everything more elaborate is probably a waste of time.

Simon-Laux:
> To make this work, electron bin has to be deleted rm ./packages/target-electron/node_modules/.bin/electron

PR #5098, opened 2025-05-12
Profile tweaks

### LINKED ISSUE #5093: profile tweaks
as discussed in some other channels, here are some tweaks for the new profile:

# move "share" to three-dot-menu

the "share" button is not really needed to be always available that prominently:

- it happens often that one taps the wrong button
- the new profile is about "put profile data used for recognising ppl to focus" - cleaning up not always needed things, resulting in less cluttering, helps on that
- also on ios/android it is in the "three dot menu"

in the three dot menu, it should be the second entry, as we have now dividers, we can also make use of them:

```
Edit Name
Share 
-----------
Encryption Info
Block Contact
```

note, that "Block Contact" should become last.

the width of the remaining button can stay as is, it would look as follows then:

# use another icon than @ for the address

the line showing the email address reads as `@ mbpdqec77@nine.testrun.org` - which is quite misleading and looks as a fedivers address.

let's better use the "server" icon as in https://github.com/deltachat/deltachat-ios/pull/2690 - it can be regarded as a server, but as well as a spreadsheet - whatever the user sees, sth buerocratic seems fine (using an icon similar from some icon repo desktop is using is fine)

# move "bio" up, strike headline

the idea is to have avatar/name/bio together - these three things are also set together in the settings. the "Bio" headline is not needed when showing it, it may be easily confusing if it is only a link eg.

the following is a suggestion, it could also be layouted differently, but we should do that with as few UI elements and cluttering as possible:

important thing is that there is a clear "end mark" of the bio (the start is implicitly there because of the title)

nicodh:
resolves #5093

aligns all items to have the same margin

image


r10s (commented):
nice!

- the icons is are bit different in stroke width, but i is good enough for now - it is not totally settled which icons we want to have at the end.

- the "verified by" layout ,however broke with this PR - it looks like the following now ...
<br>... where it should be just text as before:<br>

- i would use the same padding above/below the "Send Message" button

- i think, a layout with the "Send Message" button being left-aligned as the other items is more coherent (only title is centered then, also at other places, we do not center buttons)

nicodh:
> * the icons is are bit different in stroke width, but i is good enough for now - it is not totally settled which icons we want to have at the end.

I fixed the verifiied margin at least

> * the "verified by" layout ,however broke with this PR - it looks like the following now ...

is fixed now

> * i would use the same padding above/below the "Send Message" button. or even slightly more below

done

> * i think, a layout with the "Send Message" button being left-aligned as the other items is more coherent (only title is centered then, also at other places, we do not center buttons)

done

r10s (approved):
thanks a lot!

reviewed by testing, did not check the code

PR #5099, opened 2025-05-12
Show AlertDialog on error

### LINKED ISSUE #5092: Missing error handling when sending message
When calling BackendRemote.rpc.sendMsg an appropriate error handling is missing. Possible errors are silently catched and logged. https://github.com/deltachat/deltachat-desktop/blob/379d77f6d617cfaaf0385603d45db11b755d380b/packages/frontend/src/components/composer/Composer.tsx#L249

We should at least show an error dialog.

Atm. since we don't show an error but the draft stays as it is, it will look like sending doesn't do anything (which will typically result in hitting the send button again and again)

For testing see current bug on core: see https://github.com/chatmail/core/issues/6847 

Tbd: should we remove the draft? Are there possible use cases where it might make sense to try (later) again to send the message?

See

nicodh:
resolves #5092

This is an example. The draft will be kept. https://github.com/chatmail/core/issues/6847 si silved now and part of the pgp-contacts PR

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 ### Changed
 
 ### Fixed
-
+- show error when sendMsg fails #5092
 
 <a id="1_58_1"></a>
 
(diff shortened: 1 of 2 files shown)


nicodh:
Also see https://github.com/chatmail/core/issues/6849

Simon-Laux (approved):

PR #5102, opened 2025-05-13
fix: better webxdc realtime channel send error msg

WofWca:
#skip-changelog because it's insignificant

### DIFF
--- a/packages/target-electron/static/webxdc-preload.js
+++ b/packages/target-electron/static/webxdc-preload.js
@@ -28,7 +28,7 @@ class RealtimeListener {
   }
 
   send(data) {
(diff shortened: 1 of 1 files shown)


packages/target-electron/static/webxdc-preload.js
```diff
@@ -28,7 +28,7 @@ class RealtimeListener {
   }
 
   send(data) {
-    if ((!data) instanceof Uint8Array) {
+    if ((!data instanceof Uint8Array)) {
```
WofWca, line 31:
```suggestion
    if (!(data instanceof Uint8Array)) {
```

nicodh (approved):

PR #5104, opened 2025-05-13
Update to core 1.159.4

nicodh:


WofWca (approved):
Just an upgrade. FYI this also upgrades ESbuild.
And also fixes the formatting issue in Tauri.

PR #5105, opened 2025-05-14
Improve group profile

nicodh:
- adapt margins like in profile view
- hide related chat separator if no related chats exist

#skip-changelog (minor changes)

### DIFF
--- a/packages/frontend/scss/dialogs/_group-styles.scss
+++ b/packages/frontend/scss/dialogs/_group-styles.scss
@@ -72,3 +72,13 @@ input.group-name-input {
     }
   }
 }
+
+.group-related-chats-list-wrapper {
+  margin-left: 20px;
+  margin-right: 10px;
+}
+
+.group-member-contact-list-wrapper {
+  margin-left: 10px;
+  margin-right: 5px;
+}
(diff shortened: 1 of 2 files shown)


packages/frontend/scss/dialogs/_group-styles.scss
```diff
@@ -72,3 +72,13 @@ input.group-name-input {
...
 }
+
+.group-related-chats-list-wrapper {
+  margin-left: 20px;
+  margin-right: 10px;
+}
+
+.group-member-contact-list-wrapper {
+  margin-left: 10px;
+  margin-right: 5px;
```
WofWca, line 83: I think it would be nice to add comments about how these numbers are picked, and that they are also linked with the ones in the ViewProfile dialog.

WofWca (approved):
TBH I find this extra margin really weird, but this MR in particular just unifies things, as a follow-up to https://github.com/deltachat/deltachat-desktop/pull/5098, so it's good.

image

WofWca:
> TBH I find this extra margin really weird

(Perhaps it should have been padding on individual items instead).

packages/frontend/scss/dialogs/_group-styles.scss
nicodh: I think these numbers are a result that there was never a real desing concept behind this dialog, so that's why it had different indentations for the dialog header, for the separator and for the items. (and before this and the other PR even for different item categories) But since the lists are also used in other places I didn't want to adapt the list items. That's why we have this inconsistent mix of margins to get a consistent result.

PR #5107, opened 2025-05-14
Prepare release 1.58.2

nicodh:


r10s (approved):

WofWca (commented):
Just an upgrade, looks good.

BTW, don't we want core 1.159.5?

nicodh:
> Just an upgrade, looks good.
> 
> BTW, don't we want core 1.159.5?

Not neccessarily. The only change there is not relevant for desktop (according to Hocuri)

PR #5109, opened 2025-05-14
Tauri Drag and Drop 2

Simon-Laux:
add drag in and drag out to tauri edition.

- accept drop one file in
- accept drop multiple files in
- drag out single file

original pr by @Septias: https://github.com/deltachat/deltachat-desktop/pull/4951

tested on macos, linux and windows.

on flatpak only dragging out seems to work on kde. dragging files in don't work.
I failed to understand how to utilize the file transfer portal - It should give us a file with a special mime type, but I can't figure out how to enable/trigger this behavior, in my test it still transferred the absolute path to the file, to which the flatpak app has no access to.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 ### Changed
 
 ### Fixed
-
+- show error when sendMsg fails #5092
 
 <a id="1_58_1"></a>
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
 - tauri: added notifications
 
 ### Changed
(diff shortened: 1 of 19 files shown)


nicodh (commented):
Review by testing on Linux (Ubuntu Budgie)
1. Single file IN works
2. Multiple file IN works
3. Single file OUT creates a string 'Verknüpfung mit dcblob://4b1c3b89-1e10-44a2-8226-385ff8b8eaff/1e99a9c5bdd95ca998a7a662447ac30.jpg' (yes in german since my OS is german) instead of a copy of the file

Septias (approved):

Issue #5111, opened 2025-05-15
Desktop crashes when Android adds a new profile to a chat


hpk42:
When i added with gplay-android 1.58.3 a profile to a chat that was also open on desktop, i got the following crash

```
TypeError: Cannot read properties of undefined (reading 'id')
    at file:///home/hpk/p/delta/desktop/packages/target-electron/src/components/contact/ContactList.tsx:50:26
    at Array.map (<anonymous>)
    at Component (file:///home/hpk/p/delta/desktop/packages/target-electron/src/components/contact/ContactList.tsx:38:17)
    at renderWithHooks (file:///home/hpk/p/delta/desktop/node_modules/.pnpm/react-dom@18.3.1_react@18.3.1/node_modules/react-dom/cjs/react-dom.development.js:15486:18)
    at updateFunctionComponent (file:///home/hpk/p/delta/desktop/node_modules/.pnpm/react-dom@18.3.1_react@18.3.1/node_modules/react-dom/cjs/react-dom.development.js:19617:20)
    at beginWork (file:///home/hpk/p/delta/desktop/node_modules/.pnpm/react-dom@18.3.1_react@18.3.1/node_modules/react-dom/cjs/react-dom.development.js:21640:16)
    at beginWork$1 (file:///home/hpk/p/delta/desktop/node_modules/.pnpm/react-dom@18.3.1_react@18.3.1/node_modules/react-dom/cjs/react-dom.development.js:27465:14)
    at performUnitOfWork (file:///home/hpk/p/delta/desktop/node_modules/.pnpm/react-dom@18.3.1_react@18.3.1/node_modules/react-dom/cjs/react-dom.development.js:26599:12)
    at workLoopSync (file:///home/hpk/p/delta/desktop/node_modules/.pnpm/react-dom@18.3.1_react@18.3.1/node_modules/react-dom/cjs/react-dom.development.js:26505:5)
    at renderRootSync (file:///home/hpk/p/delta/desktop/node_modules/.pnpm/react-dom@18.3.1_react@18.3.1/node_modules/react-dom/cjs/react-dom.development.js:26473:7)
```

[desktop-crash-log.txt](https://github.com/user-attachments/files/20226199/desktop-crash-log.txt)

WofWca:
The crash occurs here

[...]

Issue #4646, opened 2025-02-15
Member list is not refreshed if it changes while you look at the group profile


link2xt:
Apparently this code does not trigger a re-render:
https://github.com/deltachat/deltachat-desktop/blob/0c1fe678f13b2c73828694e4afcb6a1d5f138499/packages/frontend/src/components/dialogs/ViewGroup.tsx#L149-L155

I noticed that core function `getContactsByIds` puts all contacts into a hash and will re-order them "randomly". This is what happens to `pastContacts` and I want to fix this in the core so `getContactsByIds` preserves the order so I can sort them by the timestamp of leaving. EDIT: made a fix on desktop side https://github.com/deltachat/deltachat-desktop/pull/4651

But for non-past contacts it turns out `getContactsByIds` is never actually used.

To reproduce, open a group profile, then from the phone add/remove members. Profile member list should re-render, but it doesn't.

WofWca:
The problem is that we directly mutate the state, instead of using React.

We get the group object from `openDialog` as a prop. So, to fix this properly, some refactoring is needed.

This works well enough when we _do_ get a re-render, i.e. when we go to a profile, modify it, and then go back.

Maybe adding a hack to force a re-render would work also.

PR #5113, opened 2025-05-15
fix: crash in "View Group" dialog on adding member

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/5111.

Related issues and MRs that have to do with updating members list:
- https://github.com/deltachat/deltachat-desktop/pull/4894.
- https://github.com/deltachat/deltachat-desktop/issues/4646.
- https://github.com/deltachat/deltachat-desktop/pull/4531.
- https://github.com/deltachat/deltachat-desktop/pull/4353.

I have tested this for a bit. Adding, removing, blocking, unblocking members from the current and another device. I have not encountered any issues.
Although I don't claim that this is fool-proof. This could potentially surface another bug.
My biggest concern is that this commit makes it no longer re-fetch the entire list of contacts, so it might be that we're sometimes not updating when we should.

### DIFF
--- a/packages/frontend/src/components/dialogs/ViewGroup.tsx
+++ b/packages/frontend/src/components/dialogs/ViewGroup.tsx
@@ -81,6 +81,16 @@ export const useGroup = (accountId: number, chat: T.FullChat) => {
   }
 
   type GroupContacts = typeof group.contacts
+  /**
+   * Why setGroupContacts & setGroupMembers?
+   * setGroupMembers triggers a modifiyGroup in the backend and is called after
+   * changes triggered by the user while setGroupContacts is only called after
+   * changes from "outside" (not on this client) after DCEvent "ContactsChanged"
+   * to update the group in local state
+   *
+   * It takes a "setter" argument to make sure, the update always happens
+   * on the latest group state
+   */
   const setGroupContacts = useCallback(
     (setter: (oldContacts: GroupContacts) => GroupContacts) => {
       setGroup(group => {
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/dialogs/ViewGroup.tsx
```diff
@@ -80,18 +80,16 @@ export const useGroup = (accountId: number, chat: T.FullChat) => {
...
-      .getContactsByIds(accountId, group.contactIds)
-      .then(contacts => {
-        // update contacts in case a contact changed
-        // while this dialog is open (e.g. contact got blocked)
-        setGroup(group => ({
-          ...group,
-          contacts: group.contactIds.map(id => contacts[id]),
-        }))
+  type GroupContacts = typeof group.contacts
+  const setGroupContacts = useCallback(
```
nicodh, line 84:
```suggestion
  /**
   * Why setGroupContacts & setGroupMembers?
   * setGroupMembers triggers a modifiyGroup in the backend and is called after
   * changes triggered by the user while setGroupContacts is only called after
   * changes from "outside" (not on this client) after DCEvent "ContactsChanged"
   * to update the group in local state
   *
   * It takes a "setter" argument to make sure, the update always happens
   * on the latest group state
   */
  const setGroupContacts = useCallback(
```

nicodh (approved):
Review by testing & reading

I proposed a comment to make it easier to understand these changes

I found another bug related to GroupView (but not introduced by this PR) which I will fix  in a separate PR

packages/frontend/src/components/dialogs/ViewGroup.tsx
WofWca: We should just get rid of `setGroupMembers`, it's buggy.
https://github.com/deltachat/deltachat-desktop/issues/5111#issuecomment-2883371449

But yeah, let's add the comment until then.

PR #5115, opened 2025-05-16
Pass current state to AddMember dialog

### LINKED ISSUE #5114: Removed contact appears as group member in AddMemberDialog
When a group member is removed while the ViewGroup dialog is opened, the view reflects the change by showing the contact as "past member". But when opening the AddMember dialog the contact is still shown as group member (i.e. checked)

nicodh:
resolves #5114

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Fixed
 - crash when a member gets added to a group and "View Group" dialog is open #5111
+- show appropriate state in AddMemberDialog #5114
 
 <a id="1_58_2"></a>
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ### Fixed
 - crash when a member gets added to a group and "View Group" dialog is open #5111
+- show appropriate state in AddMemberDialog #5114
```
WofWca, line 11: This doesn't make sense to users.

WofWca (approved):
The change is good, but it does not resolve the issue fully, because if group membership updates happen while the `AddMemberDialog` is open, the same issue can happen.

PR #5119, opened 2025-05-17
Show error if jumpToMessage failed

### LINKED ISSUE #5118: Show error dialog if jump to message failed
If jumpToMessage fails for any reason, the user gets no feedback at all. So it just seems that the application is not responding.

We should at least show an error dialog.

Followup to #5100

nicodh:
resolves #5118

Just to show the user what happened instead "no reaction" at all

#skip-changelog

### DIFF
--- a/packages/frontend/src/stores/messagelist.ts
+++ b/packages/frontend/src/stores/messagelist.ts
@@ -981,6 +981,10 @@ class MessageListStore extends Store<MessageListState> {
             `even belong to chat ${chatId}? Or did the message get deleted?\n` +
             `Anyways, falling back to jumping to the last message.`
         )
+        window.__userFeedback({
+          type: 'error',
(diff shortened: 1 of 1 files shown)


r10s:
in theory, that should anyway no longer happen, right?

r10s (approved):

packages/frontend/src/stores/messagelist.ts
```diff
@@ -981,6 +981,10 @@ class MessageListStore extends Store<MessageListState> {
             `even belong to chat ${chatId}? Or did the message get deleted?\n` +
             `Anyways, falling back to jumping to the last message.`
         )
+        window.__userFeedback({
+          type: 'error',
+          text: `${window.static_translate('error')}: message not found`,
```
WofWca, line 986: For the next time: we have the `error_x` string, which should be more appropriate.

PR #5123, opened 2025-05-19
Ignore dynamic created data in prettier

nicodh:
Avoid superfluous warnings in prettier check

#skip-changelog


WofWca (approved):
Makes sense.
This shouldn't break anything.

PR #5124, opened 2025-05-19
Upgrade to react 18.3

nicodh:
As preparation for react 19.x


WofWca (approved):
We are already using React 18.3.1

https://github.com/deltachat/deltachat-desktop/blob/c570967058588f8ff36a07ca84df9cf8c2a35462/pnpm-lock.yaml#L143-L145

This change does not affect behavior, no hashes are changed in the lock file.

Issue #5097, opened 2025-05-11
After opening a chat chat input text entry field is stealing the focus


pvagner:
- Operating System (Linux/Mac/Windows/iOS/Android): Arch Linux
- Delta Chat Version: 1.58.1 379d77f6d
- Expected behavior: I'd like to use tab and shift+tab to navigate the UI e.g. search field, profile list, toolbars, chat log, chat input and so on.
- Actual behavior: When a chat has been opened at least once chat input text entry field is being focused all the time instead.
- Steps to reproduce the problem:

  * Launch deltachat-tauri
  * Notice the last active chat conversation opens automatically. The chat input text entry field has the focus so you can start to type your message.
  * Now press ctrl+f keyboard shortcut to move focus to the search entry field.
  * Now try to navigate with tab or shift+tab and notice how you are unable to focus other focusable elements besides chat input.

Please note I have built both tauri and electron editions of deltachat-desktop on my own. I can reliably reproduce this with deltachat-tauri and I can't reproduce this with deltachat-desktop thus I assume this might be tauri specific.

WofWca:
Thanks for the report! I would say this is probably a sub-issue of https://github.com/deltachat/deltachat-desktop/issues/4590.

So, what you're saying is that the focus never leaves the message input or the search input at all? What happens when you press Tab? The input gets focused again?
This is not the case on Windows at least.

pvagner:
@WofWca  Thanks for acknowledging. I apologize, I've missed #4590.

My precise steps are like this:

* Launch delta-tauri
* The window opens and chat input receives the focus. Screen reader reports "Write a message - my contact partner name - Deltachat".
* I press tab to move focus forwards or press shift+tab to move focus backwards and nothing happens. I can press shift+tab multiple times and nothing happens. After pressing shift+tab like 10 times I can begin typing to verify chat input is still in focus.
* In order to better showcase the issue I can press ctrl+f which will correctly move focus to the search input.
* With search input focused pressing tab or shift+tab moves focus to the chat input. Thus other than using advanced screen reader features like object navigation or flat review I have no other way on how to navigate the tauri edition of the app.
*

PR #5125, opened 2025-05-19
fix: a11y: focus being locked on message input

WofWca:
On Linux and macOS.
Such that you'd be unable to focus anything else
by using the Tab key.

Closes https://github.com/deltachat/deltachat-desktop/issues/5097.
Partially addresses
https://github.com/deltachat/deltachat-desktop/issues/4590.

This also affects Electron on all platforms and Tauri for Windows:
now we will not re-focus the composer
after you un-select some text in a message.
This can be considered a downside,
but I was not able to come up with a simple soltuion
that I am sure is also good accessibility-wise.

On WebKit, the `selectionchange` event fires pretty much every time
something is clicked or when focus changes.
Namely, it fires when the focus exits the composer,
which is what caused the bug.

## Motivation

Although this is gonna be a bit annoying for mouse users
that they now have to click the composer manually
after certain actions, I am afraid we can't keep
the approach of always keeping it in focus by default.
Instead we should only focus it only in some concrete cases,
where we are sure it is appropriate.

The majority of the "always keep composer in focus" code
has been introduced in 199ea232539e77ea9b4c903451b90a56dc880554
(https://github.com/deltachat/deltachat-desktop/pull/2007).
The idea was to keep it in focus _by default_,
except some specific cases:
> Maybe we should rather listen for the blur event
> on the composer text area and check there
> if we want to give up the focus.

Some other issues related to composer re-focusing:
- 5dfc265c98563c87bd6f774d47c131bd839d90c9.
- 24510e95c5b865ce9b3372200e8b6d7692e1beb0.
- 5f54cb71829cd0f70070d1cccb5e848fba154dbb.
  (https://github.com/deltachat/deltachat-desktop/pull/3313,
  https://github.com/deltachat/deltachat-desktop/issues/3286).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - accessibility: don't re-announce message input (composer) after sending every message #5049
 - accessibility: correct `aria-posinset` for chat list #5044
 - don't close context menues on window resize #5418
+- tauri: accessibility: fix focus always being locked on the message input #5125
 
 <a id="2_11_1"></a>
 
--- a/packages/frontend/src/components/message/MessageListAndComposer.tsx
+++ b/packages/frontend/src/components/message/MessageListAndComposer.tsx
@@ -251,20 +251,6 @@ export default function MessageListAndComposer({ accountId, chat }: Props) {
     [hasOpenDialogs]
   )
 
-  const onSelectionChange = useCallback(() => {
-    const selection = window.getSelection()
-
-    if (
-      selection?.type === 'Caret' ||
-      (selection?.type === 'Range' && selection.rangeCount > 0)
-    )
-      return
-
-    // Only one of these is actually rendered at any given moment.
-    regularMessageInputRef.current?.focus()
-    editMessageInputRef.current?.focus()
-  }, [])
-
   const onEscapeKeyUp = useCallback((ev: KeyboardEvent) => {
     if (ev.code === 'Escape') {
       // Only one of these is actually rendered at any given moment.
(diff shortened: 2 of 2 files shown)


WofWca:
CC @Simon-Laux because you reviewed https://github.com/deltachat/deltachat-desktop/pull/2007, just let me know whether you're strongly opposed to this, or want to block until you have time to figure out the impact of this change.

Simon-Laux:
I have no real opinion on this. better ask @r10s, @hpk42 and @nicodh

nicodh:
> that they now have to click the composer manually after certain actions,

which are these actions? Can you give some example?  In my tests I didn't find a difference to the existing behaviour (on Linux/Electron)

> On WebKit, the selectionchange event fires pretty much every time

Do I understand right that these changes are needed to fix a problem that only occurs on Webkit?

WofWca:
> > that they now have to click the composer manually after certain actions,

> which are these actions?

See this:

> now we will not re-focus the composer after you un-select some text in a message.

***

> Do I understand right that these changes are needed to fix a problem that only occurs on Webkit?

Yes, at least as of now.

pvagner:
Hello,
Excuse me for the bump. With this not applied, the tauri edition is very very difficult to use without a mouse or with a screen reader exclusivelly. My gues is that majority of screen reader users would refuse to use the app in such a state eventhough you are doing your best to address all other accessibility related issues and indeed it's really a great work I must say.

nicodh:
Here also a rebase is needed

WofWca:
Rebased.

nicodh (approved):
As Tauri is no officially supported client yet, the problem with webkit on Tauri is not enough reason to change a behaviour in the official client imo, but If this also fixes partly #4590 I would say we should go for it. 

We still have Ctrl + M to jump into the composer input at any time from anywhere.

Let's see if there are many people complaining which I would not expect...

PR #5126, opened 2025-05-19
Import chokidar only when needed

nicodh:
#skip-changelog

It seems the default import of chokidar (which is only used in watch mode) causes trouble in the build process sometimes.

`
file:///build/deltachat-desktop/src/deltachat-desktop/node_modules/.pnpm/[chokidar@3.6.0](mailto:chokidar@3.6.0)/node_modules/chokidar/esm/index.js:6
import { readdirp } from 'readdirp';
         ^^^^^^^^
SyntaxError: Named export 'readdirp' not found. The requested module 'readdirp' is a CommonJS module, which may not support all module.exports as named exports.
`

Since the build process doesn't need the watch parameter we can avoid the

### DIFF
--- a/bin/copy.js
+++ b/bin/copy.js
@@ -2,7 +2,6 @@
 //@ts-check
 import { copyFile, mkdir, readdir, stat } from 'fs/promises'
 import { join } from 'path'
-import { watch as _watch } from 'chokidar'
 
 async function copyRecursive(source, destination) {
   if ((await stat(source)).isDirectory()) {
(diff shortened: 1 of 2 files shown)


bin/copy.js
```diff
@@ -47,6 +46,8 @@ async function copy(source, destination, watch = false) {
   let scheduled = undefined
 
   if (watch === true) {
+    // Dynamically import chokidar only when watch is true
```
WofWca, line 49: The comment is kinda weird without context. I suggest either removing it or adding info about why we shouldn't use static imports.

WofWca (approved):
It doesn't look like this will break things, unless chokidar has side-effects.

PR #5129, opened 2025-05-20
Upgrade react 19.1 & type fixes

nicodh:
This change needs some torough testing I also had to apply some changes in types and/or default values that should also be checked.

Once we have this in production we can proceed with using the new features like getting rid of forwardRef or using userTransition where appropriate

#skip-changelog or would it make sense to report the update to users? Will be only interesting for techies?


packages/frontend/src/contexts/RovingTabindex.tsx
```diff
@@ -98,7 +98,7 @@ type ProviderProps = PropsWithChildren<{
    * An element that wraps all the elements that we need the roving tabindex
    * behavior applied to them within this UI widget.
    */
-  wrapperElementRef: RefObject<HTMLElement>
+  wrapperElementRef: RefObject<HTMLDivElement | HTMLElement | null>
```
WofWca, line 101: Do we have to add `HTMLDivElement` here? Isn't it implied by `HTMLElement`?

WofWca (approved):
Works alright based on a bit of testing. The code is also fine, mostly types, but also non-breaking `useRef` updates.

Something that I guess needs checking is whether our crash screens still work.

PR #5140, opened 2025-05-22
Update group members

### LINKED ISSUE #5132: External changes can be overriden with add/remove member in groups
If 2 users change members at the same time it can cause unexpected results, since the users passed to modifyGroup may not include the fresh updated members and thus override changes  from the other user.

nicodh:
resolves #5132


WofWca (commented):
I like the "diff" approach, but can't we write this without `useState` and `useEffect`, and use events / callbacks instead? I mean those two `addMembers` and `removeMember` only get set when the dialog form gets submitted, and then immediately get reset when the "event" gets handled. This sounds a lot like a callback.

nicodh:
Right - that simplifies it even further...

packages/frontend/src/components/dialogs/ViewGroup.tsx
```diff
@@ -51,49 +55,90 @@ export default function ViewGroup(
...
+          .catch(error => {
+            openDialog(AlertDialog, {
+              title: tx('error'),
+              message: tx(
+                'error_x',
+                `Failed to modify group members: ${unknownErrorToString(error)}`
+              ),
+            })
+          })
+      }
```
WofWca, line 102: A bit too much nesting for my liking, but alright.

WofWca (approved):
Much better!
One more state is gone 🦀🦀🦀
The unrelated changes are also alright.

Works based on a bit of testing, and E2E report.

Issue #5141, opened 2025-05-22
Task: rework media view


nicodh:
The gallery should be refactored in 3 steps:

- [x] show app icon instead image in navbar and open media view with webxdc apps as first (and selected) tab
- [x] open media view in own dialog
- [x] show last opened apps icons in navbar on main screen

WofWca:
> open media view in own dialog

This is related, but seemingly different: https://github.com/deltachat/deltachat-desktop/issues/5074.

r10s:
> This is related, but seemingly different: https://github.com/deltachat/deltachat-desktop/issues/5074.

i think, this is pretty much the same than point 2 of the list above

Simon-Laux:
> open media view in own dialog

IMO it should use the available space on desktop and not be a super narrow dialog

r10s:
> IMO it should use the available space on desktop and not be a super narrow dialog

the idea is not to have a small dialog, but to use the window size, minus some padding, so that the background stays visible in a dimmed state. similar to settings, however, much wider.

PR #5142, opened 2025-05-22
Refactor gallery step1

nicodh:
Show app icon in navbar and open media view with app tab as default tab

related to #5141

#skip-changelog (will be added in step 2 or 3)

### DIFF
--- a/packages/frontend/src/components/Gallery.tsx
+++ b/packages/frontend/src/components/Gallery.tsx
@@ -27,7 +27,7 @@ import {
 
 const log = getLogger('renderer/Gallery')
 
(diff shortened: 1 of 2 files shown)


r10s:
the maps icon is rendered weirdly. not sure if it is worth to fix here, as two further steps are missing :)

Screenshot 2025-05-22 at 19 04 57

r10s (approved):
nice! step by step! thanks a lot for pushing that <3

nicodh:
> the maps icon is rendered weirdly. not sure if it is worth to fix here, as two further steps are missing :)

Hmm, it doesn't look like that for me. But there was a style missing to avoid borders on the wrapping button

Issue #5144, opened 2025-05-23
When uploading image/video with perfect size, the send button gets hidden


miicat:
## Prerequirements
* Have image/video file with "perfect" size (about 540 × 960) (you can use attached video

https://github.com/user-attachments/assets/122a8646-dfe0-473c-a719-f11644d446d1

)

## Steps to reproduce:
1. Try to upload the image/video file to any chat
2. When Delta chat is in windowed more
  --> the send button gets hidden, pressing enter doesn't send the file

<img width="1112" alt="Image" src="https://github.com/user-attachments/assets/b6077e2e-0814-4c20-a39a-62e573348b03" />

3. When Delta chat is in fullscreen, or big enough 
  --> the send button stays visible and can be used

<img width="1710" alt="Image" src="https://github.com/user-attachments/assets/a5bd7012-7b06-4a72-a5e9-ac325019ff52" />

- Operating System Mac and Windows
- Delta Chat Version: 1.58.2 (git: v1.58.2)
- Expected behavior: the send button stays visible and can be used
- Actual behavior: the send button gets hidden, pressing enter doesn't send the file
- Logs: I can provide if needed

WofWca:
Thanks for the report!
This is a duplicate of https://github.com/deltachat/deltachat-desktop/issues/5128.

PR #5145, opened 2025-05-23
Set max height in drafts

### LINKED ISSUE #5128: Regression: attached media file thumbnail is often shown too large, making it impossible to send without physical keyboard
Since a few weeks, the flatpak version of delta chat desktop will show attached media file such that the thumbnail is often shown too large, making it impossible to send without physical keyboard:

Image

This happens on big screens as well, where if you click anywhere else instead of sending right away you can't easily reach the input bar and send anymore either.

- Operating System (Linux/Mac/Windows/iOS/Android): Linux using Wayland session, seen also on MacOS & Windows (see #5144 )
- Delta Chat Version: 1.58.2 (git: flathub), core v1.159.4
- Expected behavior: thumbnail is small enough that it fits the window without moving vital elements like the input bar out of bounds
- Actual behavior: preview thumbnail when sending is sometimes way too large, even on big screens
- Steps to reproduce the problem: 1. Press the attach button, 2. Choose "File", 3. Pick a .webm video in the file dialog
- Screenshots: see above
- Logs:

nicodh:
resolves #5128 & #5144

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - crash when a member gets added to a group and "View Group" dialog is open #5111
 - show appropriate state in AddMemberDialog #5114
 - tauri: fix: ignore `dcnotification:` deep-link when the app is already running
+- define max height for video in drafts #5128
 
 <a id="1_58_2"></a>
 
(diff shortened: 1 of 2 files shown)


WofWca (approved):
I am not sure about a blind cursor unset on videos, but this should work.

PR #5146, opened 2025-05-23
improvement: notify on msg in all non-curr chats

WofWca:
Instead of only showing notifications for messages
of other accounts, when the window is focused.

This is mainly for accessibility, but is also in line with
other messengers.

Together with https://github.com/deltachat/deltachat-desktop/pull/5143,
this concludes fixing accessibility issues related to
screen reader users not being notified of new messages.

There has been a bit of a discussion regarding when to show and when not to show notifications in https://github.com/deltachat/deltachat-desktop/pull/3680.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased][unreleased]
 
 ### Added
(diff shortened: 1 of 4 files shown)


nicodh (approved):
Review by reading

PR #5149, opened 2025-05-25
Open media view (gallery) in dialog

### LINKED ISSUE #5074: Move gallery to a dialog, instead of it being a tab
- Operating System (Linux/Mac/Windows/iOS/Android): All Desktop
- Delta Chat Version: 1.57.0
- Expected behavior: Accessing the gallery and escaping it is simple and intuitive
- Actual behavior:
   The gallery is a tabpanel, alongside with the "chat" tabpanel. This, IMO, is quite weird, more so for screen reader and keyboard users.
   - It's a little hard to just escape it, as you would usually want, with the Escape key. Instead you have to click on the "chat" tab.
   - When you enter the "chat" tab, the screen reader (NVDA in my case), quite redundantly IMO, announces "Chat property page". This is the tab that you most commonly use, and it doesn't need to be specified every time. Of course one can hack around this, but IMO this would be just ignoring the unfortunate design choice.
   - The Global gallery thing is arguably even more hard to escape. To do it
   1. Open a chat
   2. Try to enter the gallery and go back from it using keyboard.
- Screenshots:

   Image

- Logs:

https://github.com/deltachat/deltachat-desktop/blob/bd29e81429275949103fd46220575104499b098d/packages/frontend/src/components/screens/MainScreen/MainScreen.tsx#L461-L492

Proposed solution: for a start, turn the "gallery" tab into a button that opens a dialog.
If the concern is that it's too easy to escape it, we can disable "Escape to close dialog" and require explicit "Close" click.

For comparison, Signal also replaces the "Chat" section with the gallery, but it's dialog-like rather than tab-like, with a "back" button.
In Telegram, you access the gallery by first clicking on the profile header, and then the Gallery button in the opened dialog.

nicodh:
resolves #5074 and step2 of #5141 

- exposes "CloseAllDialogs" globally and closes all dialogs in jumpToMessage
- removes AlternateView from MessageList

TBD:
- so far the galleryImageKeepAspectRatio is not implemented (needs a 3dot menu or an icon): do we really need it?
- we could show some more information (like total number of items) (can also be done in #4868 )

We should show the chat name in this dialog, since it is not visible anymore. 

Maybe add it to the headline?

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased][unreleased]
 
 ### Added
+- open all media view (gallery) in an own dialog #5141 #5074
 
 ### Changed
 - update `@deltachat/stdio-rpc-server` and `deltachat/jsonrpc-client` to `1.159.5`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 - open all media view (gallery) in an own dialog #5141 #5074
+- show last apps in chat navbar
 
 ### Changed
 - update `@deltachat/stdio-rpc-server` and `deltachat/jsonrpc-client` to `1.159.5`
--- a/_locales/ar.xml
+++ b/_locales/ar.xml
@@ -187,8 +187,6 @@
     <string name="pref_language">اللغة</string>
     <string name="pref_use_system_emoji">استخدام الرموز التعبيرية بالنظام</string>
     <string name="pref_use_system_emoji_explain"> Delta Chat تعطيل الرموز التعبيرية المدمجة في </string>
-    <!-- deprecated -->
-    <string name="pref_app_access">دخول البرنامج</string>
     <string name="pref_chats">المحادثات</string>
     <string name="pref_in_chat_sounds">أصوات خلال الدردش</string>
     <string name="pref_other">أخرى</string>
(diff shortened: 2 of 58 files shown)


r10s:
great stuff!

> so far the galleryImageKeepAspectRatio is not implemented (needs a 3dot menu or an icon): do we really need it?

i should at least not block merging this PR that comes with far more important things.

if we consider it important, we can re-add that at any point.

> we could show some more information

in the tabs? not sure, that would clutter UI and was also not done before. it should not be part of this PR and is another discussion.

ftr, on iOS, we open the first tab with not-empty content, but even that i would not regard as part of this PR

> We should show the chat name in this dialog, since it is not visible anymore.

no, this is not needed: 

- the name is naturally shown still in the disabled background, where the user already knows its position
- the context is much clearer as in the old implementation - you should not be able to select another chat or anything else before closing the existing one
- in the apps, the only one with active elements where it is more important, the title already shows the chat context
- adding the title in brackets or else where clutters UI and distracts from other parts

so, i would say, thing are just fine :) thanks a lot!

nicodh:
> a minor:
> 
> there is some padding missing end of the search field:

is fixed

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## [Unreleased][unreleased]
 
 ### Added
```
Simon-Laux, line 5: changelog item for translation update is missing

PR #5151, opened 2025-05-26
Show last app icons on NavBar

nicodh:
Up to 3 icons are shown:

image

### DIFF
--- a/packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
+++ b/packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
@@ -16,7 +16,7 @@ import MailingListProfile from '../../dialogs/MessageListProfile'
 import SettingsStoreInstance, {
   useSettingsStore,
 } from '../../../stores/settings'
-import { Type } from '../../../backend-com'
+import { BackendRemote, Type } from '../../../backend-com'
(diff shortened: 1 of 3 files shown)


Simon-Laux:
borders should be round for consistency and icons should also have a white background because there should be no transparency in app icons. also on small screens there should be less icons visible.

Simon-Laux (approved):

packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
```diff
@@ -402,3 +436,33 @@ function ChatNavButtons({ chat }: { chat: T.FullChat }) {
...
+    return null
+  }
+  return (
+    <div className={styles.webxdcIcons}>
+      {apps.map(app => (
+        <img
+          key={app.id}
+          className={styles.webxdcIcon}
+          src={runtime.getWebxdcIconURL(accountId, app.id)}
+          alt=''
```
Simon-Laux, line 457: for accessibility we should also put the app name here?
```suggestion
          alt={app.webxdcInfo?.name}
```

nicodh:
> * on macOS, for me, tapping the icons does not open the app

Is fixed, data-no-drag-region was missing

PR #5152, opened 2025-05-26
tauri bundle set min mac version and fix double tao macros

Simon-Laux:
- **set tauri bundle min macos version to 14**
- **dedup tao macros**


WofWca (approved):
Looks alright. Nothing should break.

PR #5156, opened 2025-05-28
Load all chats initially in contact profile

### LINKED ISSUE #4982: weird double scrolling makes it hard to get through all shared contacts
the new profile has two scrolling regions that:

- the whole dialog
- the list of shared chats

this has some disadvantages:

- if the windows is at lower height, only very few of the "shared chats" are visible, you can move the mouse pointer there and scroll, but this results in all scrolling happen in that height:
<br><i>scrolling only takes place in the red-marked area<Ii>

- if you move  the cursor up, outside the "shared chats" and scroll down, the whole page scrolls. but scrolling stops fast, one gets the impression, there are only eg.  4 shared chats. only when scrolling a _second_ time, more shared chats appear:
<br><i>scrolling stops like that - but there are far more chats that appear only on scrolling again</I>

both is unexpected and unusual. having a scrolling area in a scrolling area also seems more an accident than on purpose.

it seems better to have only one scrolling area that let the whole dialog scroll (without static title, however)

for groups, otoh, things are working like expected, there is only one scrolling area already, no matter how many members there are

nicodh:
resolves #4982

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### Fixed
 - improve performance a little
+- show all shared chats at once in contact profile #4982
 
 <a id="1_59_0"></a>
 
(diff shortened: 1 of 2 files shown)


r10s (approved):
much more predictable, thanks! and i agree, having several hundreds of shared chats is the exception, so it should be fast enough usually - if speed becomes an issue, we can iterate.

PR #5172, opened 2025-05-31
Enable zoom keybindings

nicodh:
resolves #890

PR #5175, opened 2025-06-01
Use built in zoom keyhandler

### LINKED ISSUE #890: Adjust global font size

nicodh:
resolves #890

Alternate solution to #5172 
Advantage:
- less code, uses the built in Electron features
- no additional runtime methods needed

Disadvantage:
- the menu shows the wrong shortcut for non-english keyboards ( Ctrl + = instead of Ctrl + + )
- if we don't save the zoomFactor in desktop settings, we can't apply this zoomFactor to html_email views
- no possibility to listen to zoom changes in main context

There seems to be no perfect solution. When looking at other Electron based software VSCode, Signal etc. they also have workarounds. Signal uses the same approach as this PR.

Should be tested with all OS and many languages.

### DIFF
--- a/packages/target-electron/src/windows/main.ts
+++ b/packages/target-electron/src/windows/main.ts
@@ -113,11 +113,11 @@ export function init(options: { hidden: boolean }) {
   window.on('resize', saveBounds)
 
   window.once('show', () => {
-    if (DesktopSettings.state.zoomFactor !== null) {
+    if (DesktopSettings.state.zoomFactor !== undefined) {
       // apply existing legacy zoomFactor once
       mainWindow.webContents.setZoomFactor(DesktopSettings.state.zoomFactor)
       // we don't save or read zoomFactor from settings any more
-      DesktopSettings.update({ zoomFactor: null })
+      DesktopSettings.update({ zoomFactor: undefined })
     }
   })
   window.on('close', () => {})
--- a/packages/shared/shared-types.d.ts
+++ b/packages/shared/shared-types.d.ts
@@ -44,7 +44,7 @@ export interface DesktopSettingsType {
   /** @deprecated we don't store/read the zoomFactor from settings since PR #5175
    *  chromium saves it per same-origin
    */
-  zoomFactor: number | null
+  zoomFactor: number | undefined
   /** address to the active theme file scheme: "custom:name" or "dc:name" */
   activeTheme: string
   minimizeToTray: boolean


nicodh:
Tested (in german) on Linux, Windows & Mac

packages/target-electron/src/windows/main.ts
```diff
@@ -113,7 +113,11 @@ export function init(options: { hidden: boolean }) {
   window.on('resize', saveBounds)
 
   window.once('show', () => {
-    mainWindow.webContents.setZoomFactor(DesktopSettings.state.zoomFactor)
+    if (DesktopSettings.state.zoomFactor > 0) {
+      // apply existing legacy zoomFactor once before setting it to 0
+      mainWindow.webContents.setZoomFactor(DesktopSettings.state.zoomFactor)
```
WofWca, line 118: I think it'd make sense to leave a comment on the declaration of `zoomFactor` to say that it's basically unused.

WofWca, line 118: But, looking at the usage of the variable, it doesn't looks like it's OK to just reset it to 0.

https://github.com/deltachat/deltachat-desktop/blob/efb874dc608ee7cd716308fd730327fd9e8773fc/packages/target-electron/src/windows/html_email.ts#L222-L225

packages/target-electron/src/menu.ts
```diff
@@ -340,8 +300,19 @@ function getMenuTemplate(
           click: () => mainWindow.toggleAlwaysOnTop(),
         },
         {
-          label: tx('zoom'),
-          submenu: getZoomFactors(),
+          accelerator: 'CmdOrCtrl+=',
```
WofWca, line 303: We should be consistent across all windows I think. HTML email viewer, webxdc. There it's `Ctrl++`

This goes for menu items placement, labels, shortcuts, the presence of separators.

packages/frontend/src/keybindings.ts
```diff
@@ -152,10 +152,7 @@ export function keyDownEvent2Action(
       if ((ev.target as HTMLElement)?.id === 'composer-textarea') {
         return KeybindAction.MessageList_PageDown
       }
-    } else if (
-      (ev.metaKey || ev.ctrlKey) &&
-      (ev.key === '/' || ev.code === 'Slash')
-    ) {
+    } else if ((ev.metaKey || ev.ctrlKey) && ev.key === '/') {
```
WofWca, line 155: What's the point of this change?

nicodh: As I mentioned in the active dev chat: on a german keyboard the "-" key has the code 'Slash' so the key just opens the Shortcut Hints instead zooming out.

packages/target-electron/src/windows/main.ts
nicodh: I did some tests: it seems that window.webContents.getZoomLevel() or window.webContents.getZoomFactor() in html_email.ts always returns the same numbers, no matter if I zoomed in or out and request the level after that change.

It seems that Electron remembers the zoom level of the main window, but not on the added windows. At the same time, the zoom-changed event is not triggered when zoom changes so we can't listen to that to save the zoom level afte changes.

@WofWca do you remember what was the idea behind this:

https://github.com/deltachat/deltachat-desktop/blob/efb874dc608ee7cd716308fd730327fd9e8773fc/packages/target-electron/src/windows/html_email.ts#L278-L300


Added in PR https://github.com/deltachat/deltachat-desktop/pull/4061

nicodh: The comment says "legacy" I thought that is enough.

But it seems we get some problems with the zoom factor in html_email view, if we can't save the desktop zoom level in the settings. I added a new "disadvantage" in the PR description...

WofWca: I mean the declaration of `zoomFactor`, and not just this place.

WofWca: I think that is about making sure that there is no blank space between the header and the content.
Also this keeps the two webviews in sync zoom-wise, apparently. And saves bounds.

WofWca: > It seems that Electron remembers the zoom level of the main window, but not on the added windows.

It remembers the zoom level of webxdc windows at least. Here it might have to do with the fact that two webviews are used?

WofWca: > It remembers the zoom level of webxdc windows at least.

By this I mean that it's persistent between app relaunches.

nicodh:
It seems that the fact that html_email view needs the zoomFactor of the main Window adds some addtional requirements to this approach.

Since html_email uses DesktopSettings.state.zoomFactor in several places (on start, in sandbox view and in 'html-view:resize-content' event handler) a refactoring of that has quite some complexity (which we want to avoid with this approach). Especially since there is no documentation at all, in that file.

I found no way to listen to the built-in zoom handler (zoom-changed is only emitted on resize by pinch) So what we could do, add some manual zoom change handler on the main frontend window and a runtime onResize  handler to save the zoomLevel in the desktop settings as before

WofWca:
I don't get it, why we can't just replace the `zoomFactor` variable with `mainWindow.webContents.getZoomFactor()`?

nicodh:
> I don't get it, why we can't just replace the `zoomFactor` variable with `mainWindow.webContents.getZoomFactor()`?

Ah that's what I was looking for. Then we don't need an event handler, just fetch the value on the fly.

nicodh:
Somehow the html view to me seems to be bigger than before. Can you please test too...

WofWca:
I will within a few hours.

WofWca:
Yes indeed, it looks like the initial zoom does not get set on the contents of the HTML email. So, if you make the main window zoom really small, open an HTML window, and then press "Zoom In", it would actually zoom out.

nicodh:
> Yes indeed, it looks like the initial zoom does not get set on the contents of the HTML email. So, if you make the main window zoom really small, open an HTML window, and then press "Zoom In", it would actually zoom out.

Can't reproduce that.

WofWca:
OK, I can't reproduce now either XD

Anyways, the zoom of the HTML viewer also seems to be remembered, at least when you zoom in-out the header webview.
Is there an issue with it being "bigger"?

nicodh:
> OK, I can't reproduce now either XD
>  Is there an issue with it being "bigger"?

Hmm, same here: now it looks good, although I didn't change anything. Maybe some weird Electron caching?

packages/target-electron/src/windows/html_email.ts
```diff
@@ -68,6 +68,8 @@ export function openHtmlEmailWindow(
     y: undefined,
   }
 
+  const zoomFactor = mainWindow.window?.webContents.getZoomFactor() || 1.0
```
WofWca:
```suggestion
  const mainWindowZoomFactor = mainWindow.window?.webContents.getZoomFactor() || 1.0
```

packages/target-electron/src/windows/main.ts
```diff
@@ -113,7 +113,12 @@ export function init(options: { hidden: boolean }) {
   window.on('resize', saveBounds)
 
   window.once('show', () => {
-    mainWindow.webContents.setZoomFactor(DesktopSettings.state.zoomFactor)
+    if (DesktopSettings.state.zoomFactor !== 1) {
+      // apply existing legacy zoomFactor once
+      mainWindow.webContents.setZoomFactor(DesktopSettings.state.zoomFactor)
```
WofWca, line 118: Looks like we can make it `null` now.

packages/target-electron/src/windows/help.ts
```diff
@@ -157,9 +157,21 @@ export async function openHelpWindow(locale: string, anchor?: string) {
...
+            role: 'zoomIn',
+          },
+          {
+            accelerator: 'CmdOrCtrl+-',
+            label: tx('menu_zoom_out'),
+            role: 'zoomOut',
+          },
+          {
+            accelerator: 'CmdOrCtrl+0',
+            label: `${tx('reset')}`,
```
WofWca, line 172: It's probably not the best string. Maybe keep the default one? Or introduce another string ourselves?

WofWca (approved):
> OK, I can't reproduce now either XD

Actually, I can. I keep reopening the same HTML email, and the zoom differs from one open to another.
But whatever that error is, I don't think it's related to this MR, or that it got worse with it.

This MR simplifies the code, unifies the zooming stuff among windows, and makes zooming more flexible and intuitive.

packages/target-electron/src/windows/main.ts
nicodh: I marked it as deprecated and made it nullable

packages/target-electron/src/windows/help.ts
nicodh: > Maybe keep the default one?

Which one you mean? The built in from Electron is english only afaik

I think it's self explaining since we have:
Zoom In
Zoom Out
Reset

But we can also introduce 'menu_zoom_reset'

packages/target-electron/src/windows/main.ts
```diff
@@ -113,7 +113,12 @@ export function init(options: { hidden: boolean }) {
   window.on('resize', saveBounds)
 
   window.once('show', () => {
-    mainWindow.webContents.setZoomFactor(DesktopSettings.state.zoomFactor)
+    if (DesktopSettings.state.zoomFactor !== null) {
+      // apply existing legacy zoomFactor once
+      mainWindow.webContents.setZoomFactor(DesktopSettings.state.zoomFactor)
+      // we don't save or read zoomFactor from settings any more
+      DesktopSettings.update({ zoomFactor: null })
```
Simon-Laux: you could ideally reset to `undefined`, so that the key is removed from the file. (json stringify ignores undefined keys). but for that, you would also need to update the if statements and the default.

packages/target-electron/src/windows/help.ts
WofWca: I mean "reset". You can probably infer the context, but maybe not, and "reset" is scary.

nicodh:
I tested the failing e2e tests locally and they were fine!

WofWca:
> but for that, you would also need to update the if statements and the default.

The default has not been updated, which results in the zoom level not persisting between launches, as reported in https://github.com/deltachat/deltachat-desktop/issues/890#issuecomment-2934212924

PR #5181, opened 2025-06-02
tauri: fix: remember webxdc window position & size

WofWca:
TODO:
- [x] Merge https://github.com/deltachat/deltachat-desktop/pull/5171.

### DIFF
--- a/packages/target-tauri/src-tauri/src/webxdc/commands_main_window.rs
+++ b/packages/target-tauri/src-tauri/src/webxdc/commands_main_window.rs
@@ -14,12 +14,13 @@ use deltachat::{
 };
 use log::{error, trace, warn};
 
+use rand::distr::SampleString;
+use sha2::Digest;
 use tauri::{
     async_runtime::block_on, image::Image, AppHandle, Manager, State, Url, WebviewUrl,
     WebviewWindowBuilder, WindowEvent,
 };
 use tauri_plugin_store::StoreExt;
-use uuid::Uuid;
 
 #[cfg(desktop)]
 use crate::{menus::webxdc_menu::create_webxdc_window_menu, settings::get_content_protection};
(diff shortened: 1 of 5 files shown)


packages/target-tauri/src-tauri/src/webxdc/commands_main_window.rs
```diff
@@ -231,8 +232,48 @@ pub(crate) async fn open_webxdc<'a>(
...
+        // using `tauri_plugin_window_state`,
+        // but at the same time we don't want to expose
+        // the account_id and message_id to the webxdc app
+        // (through `globalThis.__TAURI_INTERNALS__.metadata.currentWindow.label`)
+        // because account_id and message_id can give away how many accounts
+        // and messages the user has.
+        // So the app could
+        //
+        // TODO we probably need to consider a more secure way
+        // to store this secret. But it's not a super important one.
```
Simon-Laux, line 251: The primary attack we worry about here is that the webxdc gets it's window label/id and thus an attacker in the same chat could infer how many messages and accounts you have.

Simon-Laux: But the other thing here could be missing housekeeping, if you delete a webxdc, it's id and position still stay in the window size config file, which is wasting some space, not notable amounts, but still information is retained that is not needed anymore.

WofWca: > The primary attack we worry about here is that the webxdc gets it's window label/id and thus an attacker in the same chat could infer how many messages and accounts you have.

Well yes, that is solved, but still, it's a secret, and I don't know much about storing them securely. E.g. see the "local database encryption" conversation.

WofWca: > missing housekeeping

Good point, though IMO this should be a job for the plugin, e.g. delete window state if it's older than N days.

PR #5182, opened 2025-06-02
tauri: prevent moving around of the whole app with the touchpad gestures on windows

### LINKED ISSUE #5075: tauri windows touchpad bounce scrolling
you can bounce scroll the app around on windows with as touch pad and two finger scrolling gesture. I thought only apple had this bounce scrolling, but apparently chrome/edge has it too. I have already disabled it for macOS.

scrolling the ui around

Simon-Laux:
closes #5075


WofWca (approved):
It seems weird that there is overscroll for an unscrollable element.

But I guess this works, and it shouldn't break stuff.

I haven't tested this.

PR #5184, opened 2025-06-03
No image context menu on app icons

nicodh:
Avoid "default" electron context menu for images on App Icons:

image

#skip-changelog

(reported by @hpk42 )


WofWca (approved):
Idk, what does this achieve? Why not keep it, if they really want to right-click?

But OK, not a big deal I guess

PR #5185, opened 2025-06-03
Handle click event on context menu layer

### LINKED ISSUE #5169: right clicking my QR code sometimes gives me only "copy image"; other times gives me "copy image" and "reset code"
- Operating System Linux
- Delta Chat Version: 1.58.2
- Expected behavior: right-click on qr code should always show menu with both "copy image" and "reset code"
- Actual behavior: sometimes (i can't figure out when), it only shows "copy image" -- and other times it shows both.
- Steps to reproduce the problem:  click on the QR code button.  right-click on the QR code a couple different times to see what context menu comes up.

nicodh:
resolves #5169

This PR at least avoids that a second right click (with opened context menu) on a QR code opens the default electron context menu. It probably fixes also the issue reported in #5169 

#skip-changelog

### DIFF
--- a/packages/frontend/src/components/ContextMenu.tsx
+++ b/packages/frontend/src/components/ContextMenu.tsx
@@ -131,7 +131,8 @@ export function ContextMenuLayer({
     setPosition({ top, left })
   }, [])
 
-  const cancel = useCallback(() => {
+  const cancel = useCallback((evt: React.MouseEvent) => {
+    evt.preventDefault()
     window.__setContextMenuActive(false)
     setCurrentItems([])
     layerRef.current?.close()


Simon-Laux (approved):
LGTM, didn't test

Simon-Laux:
> I am not completely sure though whether the second context menu has been completely useless, but it's probably the right thing to get rid of this unintentional behavior.

The native menu is needed at-least for the composer for the copy/cut/paste actions there. it is a fallback (when no other context menu is available), so calling prevent default when we handled the righ click is a fine solution in my eyes.

WofWca:
I was talking about the fact that this removes the context menu for items where we do have `onContextMenu` listener already, maybe there are cases where it's still useful. For example, in Tauri on Windows you can "Save Image As", on the QR image.

Simon-Laux:
if there are useful functions, then we should replicate them in our context menu.

nicodh:
> I was talking about the fact that this removes the context menu for items where we do have `onContextMenu` listener already, maybe there are cases where it's still useful. For example, in Tauri on Windows you can "Save Image As", on the QR image.

The ContextMenuLayer is only shown when a ContextMenu is opened. Even before this changes you could not immediately open another ContextMenu when one was open. (except for the electron_context_menu)

Right click on any context menu (like a chat item), the immediately right click on another one, the existing one will be closed but the new one is not opened.

So with the change of this PR we have the same behaviour even for electron_context_menus

PR #5188, opened 2025-06-03
improvement: reverse recent webxdcs order

WofWca:
See https://github.com/deltachat/deltachat-desktop/pull/5151#issuecomment-2909576653.
The argument for the original order (with `mediaIds.reverse()`)
was that the order matches the order in the gallery dialog.
However, I don't think that this matching is important.
What's more important is that it's more intuitive that
the items that is the last one (the rightmost) corresponds to the
message that was sent the last.
Otherwise I personally need to think really really hard
which app I'm about to open (I have several instances
of the same app).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Zoom In/Out with Ctrl +/- #890
 
 ### Changed
+- reverse order of recent webxdc apps in chat header (last item means last message)
 
 ### Fixed
 - fix `runtime.isDroppedFileFromOutside` is not working as indended #5165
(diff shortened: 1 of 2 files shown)


Simon-Laux:
> What's more important is that it's more intuitive that
the items that is the last one (the rightmost) corresponds to the
message that was sent the last.

why? because of the ordering in the message list of newest last? - but the chatlist is newest first.

I would argue the opposite, that the current solution is the intuitive one.

WofWca:
> but the chatlist is newest first.

IDK, I don't think of the chat list when I'm about to open an app.

nicodh:
If I "know" that the 3 Apps are the 3 recent apps, I also would expect the first being the newest

WofWca:
Alright

PR #5190, opened 2025-06-03
Fix horizontal scroll

### LINKED ISSUE #5162: Horizontal scrollbar appears on a single chat when window width is ~900px
- Operating System (Linux/Mac/Windows/iOS/Android): Linux/Wayland
- Delta Chat Version: 1.58.1 (git: v1.58.1-makepkg)
- Expected behavior: It shouldn't appear
- Actual behavior: It appears

### Reproduction

It happens for only two groups. For one of them(webxdc testing ground), it appears but then disappears quickly. For the other group(DC Desktop small UI bugs), it appears and doesn't disappear at all. You only need to switch to one of these groups.

In the second group, an element is having too much width. I found it by `MessageList.querySelectorAll("*").filter(elm  => elm.scrollWidth > MessageList.clientWidth)`. The element has an image, if I `display: none` it or decrease `max-width` to `420px`, it's then gone.

In the first group, however, I'm not sure why this happens or what causes this. It's possible that these are two different bugs.

### Related

https://github.com/deltachat/deltachat-desktop/issues/5161

They seem to be different bugs.

nicodh:
resolves #5162

- use breakpoint variable

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - tauri: fix webxdc apps not receiving `visibilitychange`, `beforeunload` and `pagehide` when the window gets closed (except on macOS) #5065
 - tauri: save zoom level between webxdc app launches #5163
 - tauri: prevent moving around of the whole app with the touchpad gestures on windows #5182
+- fix horizontal scroll in message list #5162
 
 <a id="1_59_1"></a>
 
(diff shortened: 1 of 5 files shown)


WofWca (approved):
Looks like this should fix it, while improving code quality.
I tried to reproduce the bug and indeed it happens between 900 and 925px.

However, the referenced issue is supposed to be #5162

PR #5191, opened 2025-06-04
Hide reactionsbar on escape

nicodh:
- enlarge image -> context menu -> esc closes context menu
- chat -> reaction menu -> esc does now work

#skip-changelog

### DIFF
--- a/packages/frontend/scss/gallery/_media.scss
+++ b/packages/frontend/scss/gallery/_media.scss
@@ -82,3 +82,7 @@ $tab-height: 46px;
     }
   }
 }
+
+.image-context-menu-container:focus-visible {
+  outline: none;
+}
--- a/packages/frontend/src/components/dialogs/FullscreenMedia.tsx
+++ b/packages/frontend/src/components/dialogs/FullscreenMedia.tsx
@@ -60,6 +60,7 @@ export default function FullscreenMedia(props: Props & DialogProps) {
     previous: false,
     next: false,
   })
+  const [contextMenuOpen, setContextMenuOpen] = useState(false)
 
   const [mediaMessageList, setMediaMessageList] = useState<number[]>([])
   useEffect(() => {
@@ -127,6 +128,14 @@ export default function FullscreenMedia(props: Props & DialogProps) {
     },
   ])
 
+  const openContextmenu = useCallback(
+    (evt: React.MouseEvent) => {
+      openMenu(evt)
+      setContextMenuOpen(true)
+    },
+    [openMenu]
+  )
+
   let elm = null
 
   if (isImage(fileMime)) {
@@ -143,7 +152,7 @@ export default function FullscreenMedia(props: Props & DialogProps) {
               <TransformComponent>
                 <div
                   className='image-context-menu-container'
-                  onContextMenu={openMenu}
+                  onContextMenu={openContextmenu}
                   tabIndex={0}
                 >
                   <img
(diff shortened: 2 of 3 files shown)


packages/frontend/src/components/dialogs/FullscreenMedia.tsx
```diff
@@ -250,6 +259,16 @@ export default function FullscreenMedia(props: Props & DialogProps) {
...
+      if (ev.code === 'Escape') {
+        if (contextMenuOpen) {
+          setContextMenuOpen(false)
+          ev.preventDefault()
+          ev.stopPropagation()
+          return
+        }
+        onClose()
+        return
+      }
```
WofWca:
```suggestion
      if (ev.code === 'Escape' && contextMenuOpen) {
        // Only close the context menu, instead of closing both
        // the context menu and the whole FullscreenMedia dialog.
        setContextMenuOpen(false)
        ev.preventDefault()
        ev.stopPropagation()
        return
      }
```

I would avoid manual handling of `onClose()`. But I guess this is a matter of preference.

Accepting this suggestion would require also removing `onClose` from the hook dependencies, and removing `canEscapeKeyClose={false}`.

packages/frontend/src/components/dialogs/FullscreenMedia.tsx
```diff
@@ -60,6 +60,7 @@ export default function FullscreenMedia(props: Props & DialogProps) {
     previous: false,
     next: false,
   })
+  const [contextMenuOpen, setContextMenuOpen] = useState(false)
```
WofWca: This never gets reset to `false`, resulting in `contextMenuOpen` being incorrectly `true` when the context menu is actually closed. Although currently the only downside to this is that you have to press "Escape" twice to close the full screen dialog.

I think this hook should be used instead:

https://github.com/deltachat/deltachat-desktop/blob/cc4fd19cd64a32ca8edca2151ef1b4272cf231e1/packages/frontend/src/components/ContextMenu.tsx#L455-L469

packages/frontend/scss/gallery/_media.scss
```diff
@@ -82,3 +82,7 @@ $tab-height: 46px;
     }
   }
 }
+
+.image-context-menu-container:focus-visible {
+  outline: none;
+}
```
WofWca:
```suggestion
.image-context-menu-container:focus-visible {
  // This element gets focused by default when you open
  // the fullscreen media. Outlining it is unnecessary.
  outline: none;
}
```

packages/frontend/src/components/dialogs/FullscreenMedia.tsx
nicodh: Changed it!

packages/frontend/src/components/dialogs/FullscreenMedia.tsx
```diff
@@ -250,6 +250,10 @@ export default function FullscreenMedia(props: Props & DialogProps) {
       if (ev.repeat) {
         return
       }
+      if (ev.code === 'Escape' && isContextMenuActive) {
+        ev.preventDefault()
```
WofWca:
```suggestion
      if (ev.code === 'Escape' && isContextMenuActive) {
        // Only close the context menu, instead of closing both
        // the context menu and the whole FullscreenMedia dialog.
        ev.preventDefault()
```

PR #5198, opened 2025-06-07
refactor: add comment about `getPathForFile`, drag

WofWca:
This commit does not affect behavior,
`isDroppedFileFromOutside()` returns `true` for an empty string.


nicodh (approved):
Just a comment

PR #5199, opened 2025-06-08
tauri: fix: "Connectivity" dialog style CSP

WofWca:
This fixes the dialog content having dark text on dark background.

Add a hash for the core's style,
and don't try to inject styles dynamically,
instead always fall back to light theme.

This does not affect the Electron version.

This is ugly, but better than nothing:

image

image


Simon-Laux (approved):
code looks good

PR #4964, opened 2025-04-09
docs: add "webxdc network isolation" to RELEASE.md

WofWca:
This depends on https://github.com/webxdc/webxdc-test/pull/40.

#skip-changelog

PR #5202, opened 2025-06-10
Lower minimum macOS version by removing dummy proxy for webxdc on macos. (CSP security layer is still in place)

### LINKED ISSUE #5201: tauri: dependency on macos 14
currently the min macOS version our tauri edition runs on is version 14, this is because we enable the `"macos-proxy"` tauri feature which uses proxy apis. Which we use as an additional security layer next to CSP for webxdc. But it requires linking to new apis which are not available on older macOS versions.

https://github.com/deltachat/deltachat-desktop/blob/4ba8caadc9951d195d163b41fd293ce04b80318f/packages/target-tauri/src-tauri/src/webxdc/commands_main_window.rs#L337

Would be cool if we would also support older macOS versions.

Our options are:
- explore the usage of the availability macro: https://docs.rs/objc2/latest/objc2/macro.available.html
  - while generally interesting to learn how that works, I believe that this is out of scope and not worth the work for this particular problem
- though probably easiest would be just to remove the proxy on macOS, it is just an security layer that is not strictly needed.

Simon-Laux:
closes #5201

### DIFF
--- a/packages/target-tauri/src-tauri/src/webxdc/commands_main_window.rs
+++ b/packages/target-tauri/src-tauri/src/webxdc/commands_main_window.rs
@@ -14,12 +14,13 @@ use deltachat::{
 };
 use log::{error, trace, warn};
 
+use rand::distr::SampleString;
+use sha2::Digest;
 use tauri::{
     async_runtime::block_on, image::Image, AppHandle, Manager, State, Url, WebviewUrl,
     WebviewWindowBuilder, WindowEvent,
 };
 use tauri_plugin_store::StoreExt;
-use uuid::Uuid;
 
 #[cfg(desktop)]
 use crate::{menus::webxdc_menu::create_webxdc_window_menu, settings::get_content_protection};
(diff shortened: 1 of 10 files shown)


WofWca:
Hmmm I wonder I'd we should have a separate build with proxy disabled, Instead of also disabling it for macOS 14? Would it be a pain? I don't know much about the Apple infrastructure.

Simon-Laux:
why have two different versions? I think the proxy thing as reason alone is too hard to explain to users.

WofWca:
> I think the proxy thing as reason alone is too hard to explain to users.

No need to explain. Just have a "main" version for macOS 14 and above, and a "legacy" version for 13 and below.

Simon-Laux:
> Would it be a pain?

extra releases are always annoying as they add on build time.

Also I don't think the gain here is high enough to outweigh the cost and confusion factor of having multiple different releases.
If we want it conditional then we should look into the availability macros. - like I mentioned in https://github.com/deltachat/deltachat-desktop/issues/5201

packages/target-tauri/src-tauri/src/webxdc/commands_main_window.rs
```diff
@@ -356,6 +347,20 @@ pub(crate) async fn open_webxdc<'a>(
...
 
+    // This is only for non-macOS platforms.
+    // Use a non-working proxy to almost(!) isolate the app
+    // from the internet.
+    // "Almost" because there are still cases where the webview
+    // will bypass the proxy, such as with WebRTC.
+    // To disable WebRTC, we take separate measures.
+    //
+    // Note that `additional_browser_args` might make `proxy_url`
+    // have no effect (see below).
```
Simon-Laux, line 399:
```suggestion
    // have no effect (see below).
    //
    // IDEA: we disabled it on macOS, because it increased the minimum version to 14.
    // But there may be a possibility to enable it conditionally when the api is available, see
    // https://github.com/deltachat/deltachat-desktop/issues/5201
```

WofWca (approved):
OK, let's merge and keep an eye on isolation. Hopefull https://github.com/deltachat/deltachat-desktop/pull/4964 will help in this.

The changelog entry, again, will need to be moved, prior to the merge.

PR #5203, opened 2025-06-13
fix: minor fixes for `ProxyConfiguration`

WofWca:
Follow-up to https://github.com/deltachat/deltachat-desktop/pull/5052.

#skip-changelog

### DIFF
--- a/packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
+++ b/packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
@@ -18,7 +18,7 @@ import Button from '../../Button'
 
 import styles from './styles.module.scss'
 import { Proxy } from '../../Settings/DefaultCredentials'
-import { debounceWithInit } from '../../chat/ChatListHelpers'
+import { debounce } from 'debounce'
 
 import { getLogger } from '@deltachat-desktop/shared/logger'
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
```diff
@@ -209,37 +208,33 @@ export default function ProxyConfiguration(
...
+      setConnectivityStatus(connectivity)
     }
-    checkConnectivity()
-    return () => {
-      removeConnectivityListener()
-    }
+
+    const debouncedUpdate = debounce(update, 300)
+    debouncedUpdate()
+    debouncedUpdate.flush()
```
nicodh, line 230: Is there any advantage to call flush here immediately instead of passing immediate: true as 3rd parameter to the debounce function?

nicodh (approved):

packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
WofWca: Yes. `immediate` would make it so that if two `ConnectivityChanged` events fire rapidly, the underlying function would only get invoked after the first one, but not after the second one.

```
debouncedUpdate()
debouncedUpdate.flush()
```

is practically equivalent to just

```
update()
```

except that it also starts the debounce period.

Issue #3072, opened 2023-01-08
clear unread count on jump down button when jumping to bottom


Simon-Laux:
jump down button also has a mode to jump to previous message after jumping to a message from a quote.
So only if it is in jump to bottom mode it should call "mark noticed chat" to clear the unread counter.

This is mostly annoying because of #2975, but even when that is fixed, there is still the case of skipping a large amount of messages when marking messages as read, because you just jumped to the bottom and did not have them in view.
(maybe fix #2975 first actually, because without this issue the bug is harder to see)

WofWca:
The relevant code:

https://github.com/deltachat/deltachat-desktop/blob/fb05580f9c8962b0257c4c0107ab4dd25f157910/packages/frontend/src/stores/messagelist.ts#L425-L454

https://github.com/deltachat/deltachat-desktop/blob/bdca2728906da437fc7569cbfa745dc798ef8ba6/packages/frontend/src/components/message/MessageList.tsx#L811

WofWca:
To be clear, here's what the issue looks like:

https://github.com/user-attachments/assets/c8bc20d7-42ec-4336-8911-3e17591728a1

PR #5204, opened 2025-06-14
fix: set "unread" count to 0 when jump to bottom

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/3072.

I have verified that this works.


nicodh (approved):
Reviewed by testing & reading

PR #5219, opened 2025-06-22
upload to artifact instead of releases in tauri release ci

Simon-Laux:


.github/workflows/build-tauri-release.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     tags:
       - 'v*'
+  pull_request: # TODO remove after testing
```
Simon-Laux:
```suggestion
```

PR #5222, opened 2025-06-23
fix: some RPC fetch races (add `useRpcFetch`)

WofWca:
Partially addresses https://github.com/deltachat/deltachat-desktop/issues/4501.

#skip-changelog because I don't know any particular bugs that this fixes.

This does not fix the majority of places that need to be fixed, but at least this introduces the hook that we can use from now on, and migrate to gradually.

However, note that the hook is not universal. That is, at its current state it cannot be used to address _all_ the potential async races that we have in our app. Namely:
- It doesn't support debouncing / throttling. The places where we use throttle / debounce are probably the ones that need race condition protection the most.
  Edit: actually, I think it would be easy to do, by wrapping the RPC method into [`async-throttle`](https://www.npmjs.com/package/@jcoreio/async-throttle) / debounce.
  Edit 2: I pushed a commit that adds more generic functions that can be utilized for this.
- Passing arguments to the RPC call is a little ugly, code completion doesn't show the doc strings or the names of the arguments.
- It doesn't support setting the initial value, e.g. in case we already have it, e.g. [here](https://github.com/deltachat/deltachat-desktop/blob/efb874dc608ee7cd716308fd730327fd9e8773fc/packages/frontend/src/components/message/MessageList.tsx#L78-L115).

Nonetheless, I think it does not _have to be_ universal. Later one we could introduce a new hook for those cases or update this one.

The commits can be merged one-by-one, e.g. in case you want to only merge the new function, but not change the existing code.

I have tested the affected code, and it seems to work alright.

### DIFF
--- a/packages/frontend/src/hooks/useFetch.ts
+++ b/packages/frontend/src/hooks/useFetch.ts
@@ -1,13 +1,8 @@
 import { useCallback, useEffect, useMemo, useState } from 'react'
 import { BackendRemote } from '../backend-com'
 
-type MethodsOnly<T> = {
-  [P in keyof T]: T[P] extends (...args: any) => any ? T[P] : never
-}
-type MethodOf<T> = MethodsOnly<T>[keyof MethodsOnly<T>]
 type Result<T> = { ok: true; value: T } | { ok: false; err: unknown }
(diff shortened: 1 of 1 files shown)


nicodh (approved):
Looks good to me.
I added some (optional) comments in a separate PR #5233

PR #5223, opened 2025-06-24
refactor: simplify ChatContext `selectChat`

WofWca:
Use the more declarative `useRpcFetch` instead of manual
race condition mitigation.

I have tested the change for a while.


Simon-Laux (approved):
tested it, fixes the issue of "not jumping to first unread message in chat" for me

WofWca:
> fixes the issue of "not jumping to first unread message in chat" for me

Ohhh. Interesting. So that's not just a _potential_ bug.

WofWca:
Hmm. But I don't understand how it fixes it still. Without the fix we're supposed to _eventually_ jump to the last message still, even though we initially load the chat with the last message shown.

PR #5224, opened 2025-06-25
perf: replace some `debounce`s with `throttle`

WofWca:
This makes the app much snappier,
at a cost of sometimes performing the debounced / throttled
operations more frequently.

TODO:
- [ ] Merge the MR that this is based on.


packages/frontend/src/hooks/useRpcFetch.ts
Simon-Laux, line 1: needs to be rebased? doesn't the file already exist on main branch?

Simon-Laux (approved):

PR #5225, opened 2025-06-25
fix: potential RPC races in accounts list, perf

WofWca:
Partially addresses

Follow-up to https://github.com/deltachat/deltachat-desktop/pull/5222.

Also improve performance, do not refresh `syncAllAccounts`
unnecessarily,
and do not refresh accounts when switching to another account
(changing `selectedAccountId`).

Also add error handling when can't load accounts.

TODO:
- [x] Merge https://github.com/deltachat/deltachat-desktop/pull/5222.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,12 +3,14 @@
 ## [Unreleased][unreleased]
 
 ### Fixed
+- fix outdated info being shown sometimes in some places #5222, #5225
 - accessibility: add accessible labels for lists (messages list, chat list, profiles list) #5030
 - improve performance: remove message context menu open delay
 - improve performace in contact search #5230
 - improve performance: don't mark messages as seen unnecessarily when focusing window #5243
 - improve performance in "Edit Group" a little #5237
 - tauri: accessibility: fix outline being barely visible on Windows, and adjust some other colors #5217
+- improve performance a little in some other places #5225
 
 <a id="1_60_0"></a>
 
--- a/packages/frontend/src/components/AccountListSidebar/AccountListSidebar.tsx
+++ b/packages/frontend/src/components/AccountListSidebar/AccountListSidebar.tsx
@@ -2,7 +2,6 @@ import React, {
   useCallback,
   useContext,
   useEffect,
-  useMemo,
   useRef,
   useState,
 } from 'react'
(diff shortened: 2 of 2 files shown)


packages/frontend/src/hooks/useRpcFetch.ts
Simon-Laux, line 1: may need a rebase, the file is called differently on `main` branch

Simon-Laux (approved):

packages/frontend/src/components/AccountListSidebar/AccountItem.tsx
```diff
@@ -46,26 +47,31 @@ export default function AccountItem({
...
-        .catch(log.error.bind(log))
-    }, 200)
+  const accountFetch = useRpcFetch(BackendRemote.rpc.getAccountInfo, [
+    accountId,
+  ])
+  if (accountFetch.result?.ok === false) {
+    log.error('Failed to fetch account', accountFetch.result.err)
+  }
+  const account = accountFetch.lingeringResult?.ok
+    ? accountFetch.lingeringResult.value
```
nicodh, line 59: When looking at this I would like to challenge the naming of "lingeringResult".
"Lingering" means previous (and possibly outdated) result. So here it looks as if we only set the lingeringResult. It works, since the lingeringResult points to the final result once the result is loaded but that is not obvious from the name.

Possible names that would indicate that:
currentResult
resultWithFallback
previousOrFinalResult

This not related to this PR but should be discussed imho

WofWca: I agree. I like `previousOrFinalResult`, maybe `previousOrCurrentResult` more. Not so much longer, but much clearer.
Edit: maybe even `prevOrCurrResult`.

nicodh: but "current" still could be previous. Final expresses more what's the difference to previous

WofWca: > but "current" still could be previous.

`current` is when `lingeringResult === result`. "Final" sounds like "will never change".

nicodh: Ok, I would be fine with previousOrCurrentResult too!

Issue #4698, opened 2025-02-21
Chat scrolls up when a message gets edited (such that it's taller)


WofWca:
- Operating System (Linux/Mac/Windows/iOS/Android):
- Delta Chat Version: No version yet, message editing is not supported yet
- Expected behavior: The message list remains scrolled to the bottom
- Actual behavior: The message list scrolls up
- Steps to reproduce the problem:
  1. Open a chat.
  2. From another device, edit one of the last messages. Add more text to the message, such that it becomes taller
- Screenshots:
- Logs:

There is not yet a version of DC Desktop where you can see this in action right now, but it will start happening after we upgrade core to a version that includes https://github.com/deltachat/deltachat-core-rust/pull/6550.
The issue is similar to
- https://github.com/deltachat/deltachat-desktop/issues/4404
- https://github.com/deltachat/deltachat-desktop/issues/3763
- https://github.com/deltachat/deltachat-desktop/issues/3753

Editing a message triggers a `MsgsChanged` event, which we handle here:

https://github.com/deltachat/deltachat-desktop/blob/89c6c9dad2b39aa935067bb9acffd09f4db5bc54/packages/frontend/src/stores/messagelist.ts#L82-L88

Maybe we could set `viewState` to `scrollToBottom` `ifClose` when the new text is different from the current text of the message to solve this. Need to keep in mind that the "edit message" feature might expand to editing images, etc.

WofWca:
Hmmm, this is not the case when you are the one who makes an edit.

nicodh:
I tested this: the "Scroll down" button is shown, so the user sees that something has changed. I won't consider that a bug.

It "could" scroll down immediately but I would not say it "must". At least I would not expect it.

We could discuss if it would help to get a "Message edited" notification in general when a message is edited, so if a MsgsChanged event is related to an edited message we show a notification. That would avoid that you miss the case if  someone edited an already read message

r10s:
ftr, there are no plans atm to notify about edits or deletions. or let edits become more formal otherwise (thumbstones, history)

edits are mainly for typo fixing etc, and this is also communicated that way in the help [gg](https://delta.chat/en/help#edit)

nicodh:
> ftr, there are no plans atm to notify about edits or deletions. or let edits become more formal otherwise (thumbstones, history)
> 
> edits are mainly for typo fixing etc, this is bot unusual and also communicated that way in the help https://delta.chat/en/help#edit

Good to know!

[...]

PR #5227, opened 2025-06-25
Fetch webxdc info seperately with `rpc.getWebxdcInfo`, because new core will no longer provide it in the MessageObject

### LINKED ISSUE #5218: Call get_webxdc_info() to get webxdc info
Since core 1.160.0 Message.webxdc_info is removed as loading it may fail loading the whole message: https://github.com/chatmail/core/pull/6920

Now desktop needs to call `get_webxdc_info` as a separate call and process errors, e.g. show the Webxdc message as a file or something like this if .xdc file turns out not to be a proper Webxdc file or if there is a temporary error.

nicodh:
Resolves #5218 

I already tested various stuff, only thing I'm not really sure about is if broken media handling in Gallery is still the same as before, as I didn't had the various use cases

### DIFF
--- a/packages/frontend/src/components/attachment/mediaAttachment.tsx
+++ b/packages/frontend/src/components/attachment/mediaAttachment.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useRef, useState, useEffect } from 'react'
+import React, { useContext, useRef } from 'react'
 import { filesize } from 'filesize'
 
 import {
@@ -33,6 +33,7 @@ import type { JumpToMessage, DeleteMessage } from '../../hooks/chat/useMessage'
 import { useRovingTabindex } from '../../contexts/RovingTabindex'
 import ConfirmDeleteMessageDialog from '../dialogs/ConfirmDeleteMessage'
 import { BackendRemote } from '../../backend-com'
+import { useRpcFetch } from '../../hooks/useFetch'
 
 const log = getLogger('mediaAttachment')
 
@@ -61,7 +62,7 @@ const contextMenuFactory = (
     },
     viewType === 'Webxdc' && {
       label: tx('start_app'),
-      action: openWebxdc.bind(null, message),
+      action: openWebxdc.bind(null, message, undefined),
     },
     {
       label: tx('menu_export_attachment'),
@@ -599,8 +600,6 @@ export function WebxdcAttachment({
   const contextMenu = useContext(ContextMenuContext)
   const { jumpToMessage, deleteMessage } = useMessage()
   const accountId = selectedAccountId()
-  const [webxdcInfo, setWebxdcInfo] = useState<T.WebxdcMessageInfo | null>(null)
-  const [isLoadingWebxdcInfo, setIsLoadingWebxdcInfo] = useState(false)
 
   const interactiveElRef = useRef<any>(null)
   const rovingTabindex = useRovingTabindex(interactiveElRef)
@@ -610,34 +609,21 @@ export function WebxdcAttachment({
     tabIndex: rovingTabindex.tabIndex,
   } as const
 
-  useEffect(() => {
-    if (loadResult.kind === 'message') {
-      setIsLoadingWebxdcInfo(true)
-      BackendRemote.rpc
-        .getWebxdcInfo(accountId, messageId)
-        .then(info => {
-          if (info) {
-            setWebxdcInfo(info)
-          } else {
-            setWebxdcInfo(null)
-          }
-        })
-        .catch(error => {
-          log.error(
-            'Failed to load webxdc info for message:',
-            messageId,
-            'accountId:',
-            accountId,
-            'error:',
-            error
-          )
-          setWebxdcInfo(null)
-        })
-        .finally(() => {
-          setIsLoadingWebxdcInfo(false)
-        })
-    }
-  }, [accountId, messageId, loadResult.kind])
+  const webxdcInfoFetch = useRpcFetch(
+    BackendRemote.rpc.getWebxdcInfo,
+    loadResult.kind === 'message' ? [accountId, messageId] : null
+  )
+  if (webxdcInfoFetch?.result?.ok === false) {
+    log.error(
+      `Failed to load webxdc info for message:
+      messageId: ${messageId},
+      accountId: ${accountId},
+      error: ${webxdcInfoFetch.result.err}`
+    )
+  }
+  const webxdcInfo = webxdcInfoFetch?.result?.ok
+    ? webxdcInfoFetch.result.value
+    : null
 
   // If the message is not yet loaded by GridGallery, show loading state
   if (!loadResult || loadResult.kind !== 'message') {
@@ -662,7 +648,7 @@ export function WebxdcAttachment({
   // At this point, loadResult.kind is guaranteed to be 'message'
   const message = loadResult
 
-  if (isLoadingWebxdcInfo) {
+  if (webxdcInfoFetch?.loading) {
     const onContextMenu = getBrokenMediaContextMenu(
       contextMenu.openContextMenu,
       openDialog,
@@ -730,7 +716,7 @@ export function WebxdcAttachment({
         ref={interactiveElRef}
         className={'media-attachment-webxdc ' + rovingTabindex.className}
         onContextMenu={openContextMenu}
-        onClick={openWebxdc.bind(null, message)}
+        onClick={openWebxdc.bind(null, message, webxdcInfo ?? undefined)}
         {...rovingTabindexProps}
       >
         <img
--- a/packages/frontend/src/components/attachment/mediaAttachment.tsx
+++ b/packages/frontend/src/components/attachment/mediaAttachment.tsx
@@ -625,29 +625,6 @@ export function WebxdcAttachment({
     ? webxdcInfoFetch.result.value
     : null
 
-  // If the message is not yet loaded by GridGallery, show loading state
-  if (!loadResult || loadResult.kind !== 'message') {
-    return (
-      <div
-        ref={interactiveElRef}
-        className={'media-attachment-webxdc ' + rovingTabindex.className}
-        {...rovingTabindexProps}
-      >
-        <img
-          className='icon'
-          src={runtime.getWebxdcIconURL(selectedAccountId(), messageId)}
-        />
-        <div className='text-part'>
-          <div className='name'>Loading...</div>
-          <div className='summary'></div>
-        </div>
-      </div>
-    )
-  }
-
-  // At this point, loadResult.kind is guaranteed to be 'message'
-  const message = loadResult
-
   if (webxdcInfoFetch?.loading) {
     const onContextMenu = getBrokenMediaContextMenu(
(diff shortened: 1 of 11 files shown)


WofWca:
https://github.com/deltachat/deltachat-desktop/pull/5222 would be great and cutting down the amount of lines here.

WofWca:
@nicodh do you want to rewrite this with the new `useRpcFetch` or shall I review as is?

Simon-Laux:
Does this really require a new core? from what I can tell the direct api is already available in the core we use.
(it just removes the (premature) performance optimisation of directly returning the webxdc info with the message object, that caused some issues somewhere)
So I think we can already merge this?

nicodh:
Ok, just checked it: it's the other way round: 1.160.0 core will not run without this changes. So I think it makes sense to merge this first and then do the upgrade to core 1.160.

nicodh:
> @nicodh do you want to rewrite this with the new `useRpcFetch` or shall I review as is?

Please review as it is. I will adapt in a separate PR to new useRpcFetch.

packages/frontend/src/components/attachment/messageAttachment.tsx
```diff
@@ -234,6 +235,32 @@ export function DraftAttachment({
...
+            attachment.id,
+            error
+          )
+          setWebxdcInfo(null)
+        })
+        .finally(() => {
+          setIsLoadingWebxdcInfo(false)
+        })
+    }
+  }, [accountId, attachment, attachment.id, attachment.viewType])
```
WofWca: Having `attachment` (draft message) as dependency makes it re-run the effect every time the draft changes, leading to jumpiness

packages/frontend/src/system-integration/webxdc.ts
```diff
@@ -37,7 +37,8 @@ export async function internalOpenWebxdc(
     messageId = message.parentId
     message = await BackendRemote.rpc.getMessage(accountId, messageId)
   }
-  if (!message.webxdcInfo) {
+  const webxdcInfo = await BackendRemote.rpc.getWebxdcInfo(accountId, messageId)
```
WofWca: For better performance, `internalOpenWebxdc` should accept `webxdcInfo`, at least optionally, because it's already available most (if not all) of the time.

packages/frontend/src/components/message/Message.tsx
```diff
@@ -1114,13 +1114,40 @@ function WebxdcMessageContent({
   tabindexForInteractiveContents: -1 | 0
 }) {
   const tx = useTranslationFunction()
+  const [webxdcInfo, setWebxdcInfo] = useState<T.WebxdcMessageInfo | null>(null)
+  const [isLoadingWebxdcInfo, setIsLoadingWebxdcInfo] = useState(false)
+  const accountId = selectedAccountId()
+
+  useEffect(() => {
```
WofWca: Looks like we need to listen on `MsgsChanged` around here, because otherwise the message info doesn't update live, e.g. with the Editor app. Try sending one, and then typing something. After that, the updates will only appear when you switch the chat.

And maybe the same goes for all other places where apps are displayed, e.g. gallery?

image

packages/frontend/src/components/message/Message.tsx
```diff
@@ -1114,13 +1114,40 @@ function WebxdcMessageContent({
...
+  }, [accountId, message.id, message.viewType])
+
   if (message.viewType !== 'Webxdc') {
     return null
   }
-  const info = message.webxdcInfo || {
-    name: 'INFO MISSING!',
+
+  const info = webxdcInfo || {
+    name: isLoadingWebxdcInfo ? 'Loading...' : 'INFO MISSING!',
```
WofWca, line 1176: At certain screen widths, the "loading..." text is shorter (height-wise) than the text after the eventual load of the app, which results in the infamous "chat is not scrolled all the way to the bottom when you open it" bug. See
- https://github.com/deltachat/deltachat-desktop/issues/4404
- https://github.com/deltachat/deltachat-desktop/issues/3763
- https://github.com/deltachat/deltachat-desktop/issues/3753
- https://github.com/deltachat/deltachat-desktop/issues/4698

image

I think we either have to wait for message info to load in `messagelist.ts`, basically reverting the core change that caused this MR, or ensure that the webxdc message never gets taller when `webxdcInfo` eventually loads. This could be achieved by either hard-coding the message height for the top portion of a webxdc message (i.e. everything besides the message text), as we did for images not so long ago, or by making the app title and the "summary" never wrap. But keep in mind that the summary could be missing (and it is missing by default for a freshly sent app).
This will also solve the problem with height changes that occur later on, after the message list has been fully loaded, upon `webxdc.sendUpdate({ document, summary})` calls.

packages/frontend/src/components/attachment/mediaAttachment.tsx
```diff
@@ -609,44 +610,97 @@ export function WebxdcAttachment({
...
     const onContextMenu = getBrokenMediaContextMenu(
       contextMenu.openContextMenu,
       openDialog,
       deleteMessage,
       messageId,
       accountId
     )
-    // webxdc info is not set, show different error
-    log.error('message.webxdcInfo is undefined, msgid:', messageId)
+    log.error('webxdcInfo is not available, msgid:', messageId)
```
WofWca, line 661: Hm actually I forgot to check out this file.

This error gets printed for each message upon the initial render. Looks like `isLoadingWebxdcInfo` should be `true` initially, and the same goes for all other places I think.

packages/frontend/src/components/attachment/mediaAttachment.tsx
```diff
@@ -609,44 +610,97 @@ export function WebxdcAttachment({
...
+      >
+        <img
+          className='icon'
+          src={runtime.getWebxdcIconURL(selectedAccountId(), messageId)}
+        />
+        <div className='text-part'>
+          <div className='name'>Loading...</div>
+          <div className='summary'></div>
+        </div>
+      </div>
```
WofWca: It would be nice not to copy-paste this.

Simon-Laux:
I changed pr title because it was confusing me again - the `rpc.getWebxdcInfo` api is not new.

packages/frontend/src/components/attachment/messageAttachment.tsx
nicodh: Addressed in 0fbe7dc09745a33a08d45f306b42a3c7b27bf829

packages/frontend/src/system-integration/webxdc.ts
nicodh: Adressed in 6e441010042ecf17343c43fa008f2bdc97be8b18

packages/frontend/src/components/message/Message.tsx
nicodh: I added a listener for event WebxdcStatusUpdate 554c6bea556830990a3e23e12f6a244af312d562

packages/frontend/src/components/attachment/mediaAttachment.tsx
nicodh: mediaAttachment was refactored to use the new rpcFetch...
Other use cases are  in DraftAttachment and WebxdcMessageContent where isLoadingWebxdcInfo is set anyway

packages/frontend/src/components/attachment/mediaAttachment.tsx
nicodh: I agree but that should go into another PR - this PR is already pretty overloaded anyway

packages/frontend/src/components/attachment/messageAttachment.tsx
WofWca: Oh, right. We're making a new message on almost every key stroke...

In that case I guess not showing "Loading..." would solve the jumpiness problem, until we get around to implementing proper draft saving.

> Addressed in https://github.com/deltachat/deltachat-desktop/commit/0fbe7dc09745a33a08d45f306b42a3c7b27bf829

Relying on filename now seems wrong, but I guess it's not that big of a potential bug.

packages/frontend/src/components/message/Message.tsx
nicodh: I think this shopuld be solved by applying the right media query breakpoints.
Like with images it seems there are some rules applied earlier than others where both should use the same breakpoint.
I will have a look but that could also go into a separate PR imho

packages/frontend/src/components/message/Message.tsx
nicodh: The same applies to mediaAttachement in Gallery where the content is also not updated when WebxdcStatusUpdate events happen, but that could maybe adresses in a followup since the use case to have the app and gallery open at the same time is probably more rare

packages/frontend/src/components/attachment/mediaAttachment.tsx
WofWca: > where isLoadingWebxdcInfo is set anyway

Is it? It's initialized to `false` in those places, only until the `useEffect` runs. But fine I guess.

packages/frontend/src/components/attachment/mediaAttachment.tsx
```diff
@@ -609,44 +609,84 @@ export function WebxdcAttachment({
...
+        ref={interactiveElRef}
+        className={'media-attachment-webxdc ' + rovingTabindex.className}
+        {...rovingTabindexProps}
+      >
+        <img
+          className='icon'
+          src={runtime.getWebxdcIconURL(selectedAccountId(), messageId)}
+        />
+        <div className='text-part'>
+          <div className='name'>Loading...</div>
```
WofWca: This is wrong. `loadResult` is never falsy, and `loadResult.kind !== 'message'` only leaves `kind === "loadingError"`. So it's not a "loading" state.

nicodh: You are right! I removed it in 7b44c5aca56f068db15a15f67059b9d87465ed85

packages/frontend/src/components/message/Message.tsx
WofWca: Do you think `WebxdcStatusUpdate` is better than `MsgsChanged` for this?

packages/frontend/src/components/attachment/mediaAttachment.tsx
```diff
@@ -609,44 +609,61 @@ export function WebxdcAttachment({
...
       messageId,
       accountId
     )
-
     return (
       <div
         ref={interactiveElRef}
-        className={'media-attachment-webxdc broken ' + rovingTabindex.className}
-        title={loadResult.error}
+        className={'media-attachment-webxdc ' + rovingTabindex.className}
```
WofWca, line 639: > only thing I'm not really sure about is if broken media handling in Gallery is still the same as before, as I didn't had the various use cases

Indeed it's not the same.

- `getBrokenMediaContextMenu` is used for the "loading" state.
- `loadResult.error` is not used anymore.
- `tx('attachment_failed_to_load')` is not used, intead 'webxdcInfo is not available' is shown in both cases.

But it's not super important I guess.

PR #5229, opened 2025-06-25
chore: upgrade Electron from 34 to 37

dotlambda:
fails to build for me with
```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'esbuild' imported from /build/source/packages/target-electron/bin/build.js
```

### DIFF
--- a/packages/target-electron/package.json
+++ b/packages/target-electron/package.json
@@ -77,7 +77,7 @@
     "chai": "^5.1.1",
     "chokidar": "^3.6.0",
     "debounce": "^1.2.0",
-    "electron": "^34.0.1",
+    "electron": "^37.1.0",
     "electron-builder": "^24.13.3",
     "esbuild": "^0.25.0",
     "mocha": "^10.7.0",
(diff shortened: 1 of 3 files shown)


WofWca:
Maybe `pnpm approve-builds` needs to be run for Esbuild?

dotlambda:
I think this is just some problem in Nixpkgs that shouldn't block this change.

WofWca:
Thanks! This seems to be working.

Wave you gone through release notes to see if there is anything breaking that affects us?
- https://www.electronjs.org/blog/electron-35-0
- https://www.electronjs.org/blog/electron-36-0

I have glanced over, and nothing caught my eye. Well, except for `commandLine`, which still works fine.

I have also tested webxdc isolation, and it still works OK.

dotlambda:
We might want to test on GNOME (e.g. vanilla Ubuntu) because of the switch to Gtk4.

dotlambda:
We can also go to version 35 instead. Which one is preferred?

WofWca (approved):
Electron 37's changelog also looks fine. There is some child process stuff, but I didn't see how this can affect us, it's probably fine.

I re-tested webxdc again, and clicked around for a bit. It seems to work OK.

nicodh (approved):

PR #5230, opened 2025-06-26
perf: reduce debouncing in `useContactIds`

WofWca:
Note that this affects many places:
- Chat list search.
- "New Chat" dialog search.
- "Attach Contact" dialog search.

***
- replace `debounce` with `asyncThrottle`,
  which allows for much lower delay thanks to the fact
  that it awaits the previous call.
- fix potential race conditions, utilize `useFetch`.
  This partially addresses

Also note that this adds a new NPM dependency.

The esbuild upgrade has been automatically performed by PNPM.

I have tested the change.

TODO:
- [ ] Merge https://github.com/deltachat/deltachat-desktop/pull/5222.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixed
 - accessibility: add accessible labels for lists (messages list, chat list, profiles list) #5030
 - improve performance: remove message context menu open delay
+- improve performace in contact search #5230
 - improve performance: don't mark messages as seen unnecessarily when focusing window #5243
 - improve performance in "Edit Group" a little #5237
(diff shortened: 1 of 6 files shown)


Simon-Laux (approved):

packages/frontend/src/hooks/useFetch.ts
Simon-Laux, line 1: may need to be rebased? I wonder why this shows up here as new, because it already exists on main branch?

WofWca: I guess GitHub sucks at fast-forward merges.
Edit: or maybe it's because I force-push all the time.

PR #5231, opened 2025-06-27
Update AppIcons immediately when sending an app to a chat

nicodh:
Followup to https://github.com/deltachat/deltachat-desktop/pull/5151

Introduces a new MessageSent event that is triggered (and subscribed to) in frontend only

DcEvent MsgsChanged was not sufficient for 2 reasons:

- it is triggered on any draft change so even when typing a text it is permanently triggered
- it is not triggered when sending a message to self chat (at least it misses the msgId then)

### DIFF
--- a/packages/frontend/src/hooks/chat/useMessage.ts
+++ b/packages/frontend/src/hooks/chat/useMessage.ts
@@ -4,7 +4,7 @@ import useChat from './useChat'
 import { BackendRemote } from '../../backend-com'
 import { ChatView } from '../../contexts/ChatContext'
 import { getLogger } from '../../../../shared/logger'
-import { notifyMessageSent } from '../useMessageSentNotifier'
+import { notifyWebxdcMessageSent } from '../useWebxdcMessageSent'
 
 import type { T } from '@deltachat/jsonrpc-client'
 
@@ -141,7 +141,7 @@ export default function useMessage() {
       })
 
       // Notify about the sent message (listeners can filter by message type if needed)
-      notifyMessageSent(accountId, chatId, message)
+      notifyWebxdcMessageSent(accountId, chatId, message)
 
       // Jump down on sending
       jumpToMessage({
--- a/packages/frontend/src/hooks/useMessageSentNotifier.ts
+++ b/packages/frontend/src/hooks/useMessageSentNotifier.ts
@@ -15,7 +15,7 @@ export function notifyMessageSent(
 }
 
 /**
- * A notifier that allows components to subscribe to message sent events.
+ * A notifier that allows components to subscribe to webxdc message sent events.
  * For cases where DC Event 'MsgsChanged' is not sufficient
  * (like on MainScreen where we want to update AppIcons immediately when a
  * Webxdc message is sent).
--- a/packages/frontend/src/hooks/useWebxdcMessageSent.ts
+++ b/packages/frontend/src/hooks/useWebxdcMessageSent.ts
@@ -1,0 +1,66 @@
+import { useCallback, useEffect } from 'react'
+import type { T } from '@deltachat/jsonrpc-client'
+
+// Simple global state for Webxdc message notifications
+const webxdcMessageListeners: Set<(accountId: number, chatId: number) => void> =
+  new Set()
+
+export function notifyWebxdcMessageSent(
+  accountId: number,
+  chatId: number,
+  message: Partial<T.MessageData>
+) {
+  // Only notify if the message is a Webxdc message
+  if (
+    message.viewtype === 'Webxdc' ||
+    (message.filename && message.filename.endsWith('.xdc'))
+  ) {
+    webxdcMessageListeners.forEach(callback => {
+      try {
+        callback(accountId, chatId)
+      } catch (_error: unknown) {
+        // nothing to do here
+      }
+    })
+  }
+}
+
+/**
+ * A notifier that allows components to subscribe to Webxdc message sent events.
+ * For cases where DC Event 'MsgsChanged' is not sufficient
+ * (like on MainScreen where we want to update AppIcons immediately when a
+ * Webxdc message is sent).
+ */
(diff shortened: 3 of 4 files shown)


packages/frontend/src/hooks/chat/useMessage.ts
```diff
@@ -139,6 +140,9 @@ export default function useMessage() {
         ...message,
       })
 
+      // Notify about the sent message (listeners can filter by message type if needed)
+      notifyMessageSent(accountId, chatId, message)
```
WofWca: Introducing this front-end only event system doesn't seem like a good idea to me. I think it's better to modify core. There are a couple of problems with it:
- `notifyMessageSent` is supposed to be alongside `sendMsg` everywhere. But it's not called when sending a sticker.
- It doesn't work when you send a message from another device.
- It will require some hacks if / when we start using `sendDraft` (which BTW we shouldn't go for IMO).

But, since the work has already been done and it works, maybe it would be fine to fall back to pointing out the flaws of this system in the comments, and urging the devs not to use it.

packages/frontend/src/hooks/useMessageSentNotifier.ts
```diff
@@ -0,0 +1,63 @@
...
+const messageListeners: Set<
+  (accountId: number, chatId: number, message: Partial<T.MessageData>) => void
+> = new Set()
+
+export function notifyMessageSent(
+  accountId: number,
+  chatId: number,
+  message: Partial<T.MessageData>
+) {
+  messageListeners.forEach(callback => callback(accountId, chatId, message))
```
WofWca, line 14: This should not throw if any of the callbacks throw.

packages/frontend/src/hooks/useMessageSentNotifier.ts
```diff
@@ -0,0 +1,63 @@
...
+}
+
+/**
+ * A notifier that allows components to subscribe to message sent events.
+ * For cases where DC Event 'MsgsChanged' is not sufficient
+ * (like on MainScreen where we want to update AppIcons immediately when a
+ * Webxdc message is sent).
+ */
+export function useMessageSentNotifier() {
+  const [listeners] = useState(() => messageListeners)
```
WofWca, line 24: Why this weird `useState` thing? Why not `useRef` or, better yet, using the `messageListeners` var directly?

WofWca (commented):
Apart from the comments, the code looks good.

packages/frontend/src/hooks/useMessageSentNotifier.ts
WofWca: Oh, and if one listener throws, others need to be invoked still. So I guess `try { callback(...) } catch(e) {}` would work.

packages/frontend/src/hooks/chat/useMessage.ts
nicodh: I started with limiting this to a webxdc message sent event but then extended to a more general event. The limited version would at least avoid problem 1 you mention.
Problem 2 is very theoretical since if you send a message from one device you will look at that device and should see the app icon being updated there. Having the same chat open in another device and "missing" the icon there is very edgy

In general I agree that we should not introduce additional frontend events but passing this just to core seems also wrong since the AppIcons are only displayed in Desktop (due to having more space)

So the root question is, if we need this event only in Desktop or if it might be used in other apps as well.

I will can rename the hook and the functionality to make clear it is limited to webxdc messages to avoid "misuse"

packages/frontend/src/hooks/useMessageSentNotifier.ts
nicodh: Added a try - catch

packages/frontend/src/hooks/useMessageSentNotifier.ts
nicodh: Is removed. Was a left over...

packages/frontend/src/hooks/chat/useMessage.ts
WofWca: > but passing this just to core seems also wrong since the AppIcons are only displayed in Desktop (due to having more space)

What's wrong with that? A simple event that fires every time you sent a message sounds perfectly reasonable to me.

packages/frontend/src/hooks/useWebxdcMessageSent.ts
```diff
@@ -0,0 +1,66 @@
...
+  // Only notify if the message is a Webxdc message
+  if (
+    message.viewtype === 'Webxdc' ||
+    (message.filename && message.filename.endsWith('.xdc'))
+  ) {
+    webxdcMessageListeners.forEach(callback => {
+      try {
+        callback(accountId, chatId)
+      } catch (_error: unknown) {
+        // nothing to do here
```
WofWca, line 22: The error should at least be logged I think, otherwise it's just lost.

packages/frontend/src/hooks/chat/useMessage.ts
```diff
@@ -139,6 +140,9 @@ export default function useMessage() {
         ...message,
       })
 
+      // Notify about the sent message (listeners can filter by message type if needed)
+      notifyWebxdcMessageSent(accountId, chatId, message)
```
WofWca, line 144: This doesn't really make sense semantically, because the message is not necessarily a message with a webxdc attachment.

WofWca, line 144: TBH I'd prefer the previous "generic" approach, but with only a comment warning against using these functions.
But alright.

WofWca (approved):

packages/frontend/src/hooks/chat/useMessage.ts
nicodh: I mean if a API called is only needed by one client it's questionable to implement it in core if there are other solutions. But I will discuss that with core, maybe they agree with your opinion.

PR #5233, opened 2025-06-27
Extend some comments

nicodh:
These comments intend to help understand what useFetch does (without reading the mentioned sources first)

At least for me they would have helped on first read

#skip-changelog

### DIFF
--- a/packages/frontend/src/hooks/useFetch.ts
+++ b/packages/frontend/src/hooks/useFetch.ts
@@ -285,9 +285,11 @@ export function useRpcFetch<
  * A (possibly) more convenient version of {@link useFetch},
  * when dealing with `BackendRemote.rpc` calls.
  *
- * This hook provides better TypeScript support when working with
- * the backend RPC methods, as it can infer the correct argument types
- * and return types from the method signature.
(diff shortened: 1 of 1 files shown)


packages/frontend/src/hooks/useFetch.ts
```diff
@@ -223,9 +280,14 @@ export function useRpcFetch<
...
   Args extends null | Parameters<Method>,
 >(method: Method, args: Args): null | Ret<Method>
+
 /**
  * A (possibly) more convenient version of {@link useFetch},
  * when dealing with `BackendRemote.rpc` calls.
+ *
+ * This hook provides better TypeScript support when working with
+ * the backend RPC methods, as it can infer the correct argument types
+ * and return types from the method signature.
```
WofWca:
```suggestion
 * Namely, it allows you to pass `BackendRemote.rpc` methods directly
 * as the `method` arg, without having to worry about the `this` value.
 *
 * @example
 * useRpcFetch(BackendRemote.rpc.getMessage, [accId, msgId])
```

They all provide type support.

WofWca (approved):
Thanks! Good comments.
Although one might argue that some are a bit redundant (e.g. `Result`).
One suggestion.

PR #5234, opened 2025-06-27
Add text field in member group listing to filter by name and email

trufae:


packages/frontend/scss/dialogs/_group-styles.scss
```diff
@@ -81,3 +81,10 @@ input.group-name-input {
     padding-inline-start: 20px;
   }
 }
+
+.group-member-filter {
```
WofWca, line 85: It looks a little ugly to me
Edit: I mean the result, not the code.

packages/frontend/src/components/dialogs/ViewGroup.tsx
```diff
@@ -406,6 +409,14 @@ function ViewGroupInner(
                     quantity: group.contactIds.length,
                   })}
             </div>
+            <div className='group-member-filter'>
+              <SearchInput
```
WofWca: This component cannot be used, because it's specific to the main screen.

https://github.com/deltachat/deltachat-desktop/blob/8ca44bd1bb2ac7212787980654785013be9a2220/packages/frontend/src/components/SearchInput/index.tsx#L48-L57

packages/frontend/src/components/dialogs/ViewGroup.tsx
```diff
@@ -406,6 +409,14 @@ function ViewGroupInner(
...
+                id='group-member-filter'
+                onChange={e => setMemberFilter(e.target.value)}
+                value={memberFilter}
+                onClear={() => setMemberFilter('')}
```
WofWca, line 417: It's quite laggy for big groups. Have you considered [`useDeferredValue`](https://react.dev/reference/react/useDeferredValue) or a debounce?

WofWca (commented):
image

Overall I think it's a nice feature.
Though I guess someone else with authority to change the UI should say if it's good.

A changelog entry is missing also.

nicodh:
I think this is a very useful feature 👍 but still needs some rework:
- a separate searchfield input (can be similar to that in the search component)
- a debounce on keydown to avoid laggy behaviour when searching in big groups

Styling can be discussed and improved in a last step 

If you need help or support feel free to ask here or get in contact with us!

nicodh (changes requested):
I think this is a very useful feature 👍 but still needs some rework:

a separate searchfield input (can be similar to that in the search component)
a debounce on keydown to avoid laggy behaviour when searching in big groups
Styling can be discussed and improved in a last step

If you need help or support feel free to ask here or get in contact with us!

trufae:
I never touched react native before but i’ll do my best, if anyone here is more fluent in this workflow i’m happy to hand over too :)

packages/frontend/src/components/dialogs/ViewGroup.tsx
trufae: Solved

packages/frontend/src/components/dialogs/ViewGroup.tsx
trufae: Solved

packages/frontend/src/components/dialogs/ViewGroup/GroupSearchInput.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  const hasValue = value.length > 0
+
+  return (
+    <div className={styles.searchInputContainer}>
+      <input
+        id={id}
+        placeholder={tx('search')}
+        aria-label={tx('search')}
+        aria-description={tx('search_explain')}
```
WofWca, line 50: This string is not applicable.

packages/frontend/src/components/dialogs/ViewGroup/GroupSearchInput.module.scss
```diff
@@ -0,0 +1,52 @@
+$borderRadius: 10px;
+$closeButtonSize: 30px;
+
+.searchInputContainer {
```
WofWca, line 4: So are these two files copy-pasted? I'd add a comment about that,

WofWca:
@trufae would you like to address the comments or should someone take over the MR?

trufae:
I have very little time and i maintain too many projects, i would love to get this feature in, but i understand the effort required to get it right and port it to the android and ios platforms too. So probably the best path would be if you can take over and finish yourself :)

thanks for all the reviews! It’s been a fun journey and i learned a lot about all these technologies i never touched before. Hopefully i’ll do better next time i contribute.

ell1e:
This would have come in very handy today to undo some potentially malicious disruptive action. Sorry that this is just basically a "would love this" message, but this feature would help a lot.

r10s:
though there may be large, sort-of-public groups where search-members is more useful, search-members may sometimes by useful for smaller groups Delta Chat is designed for.

still, we should take care not to clutter UI for the major use case.

this can be done by hiding the search field by default and adding a small magnifier icon right of the "123 members" title:

Screenshot 2025-11-01 at 20 27 42

nicodh:
Closed in favour of #5697

PR #5238, opened 2025-06-28
perf: open message context menu instantly

WofWca:
By removing an unnecessary call to `rpc.getFullChatById()`:
a `T.FullChat` object is already provided in props.


Simon-Laux (approved):
LGTM

PR #5243, opened 2025-06-30
perf: clean up unread message observing, improve perf

WofWca:
Do some refactoring, and reduce the amount of unnecessary RPC calls.

Prior to this, when the window gets focused,
we'd try to mark _all_ visible messages as seen,
even if they're already seen or outgoing.
This would cause an unnecessary `debouncedUpdateBadgeCounter`,
which is quite expensive.

See individual commits for better understanding.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased][unreleased]
 
+### Fixed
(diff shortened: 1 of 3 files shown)


Simon-Laux (approved):
:star_struck:

PR #5244, opened 2025-06-30
fix: not marking last info message as read

WofWca:
This also reduces the margin between info messages
to (somewhat) match Android.

Note that there are many kinds of `info-message`s, including daymarkers, webxdc info messages, "no chat selected" message. This affects all of them.

Reproduction steps:
1. Create a group.
2. Open the group chat.
   Do not unfocus the window for the rest of the steps.
3. From another device, trigger an info message in this group,
   for example, by setting "disappearing messages".
4. Make sure to scroll to the bottom of the messsage list.

You will notice that the "unread" counters for this chat remain at 1.

This had to do with the `.info-message` margin pushing
the `.message-observer-bottom` to the very bottom
of the message list, such that when you scroll to the bottom,
it doesn't register as `isIntersecting` in
`unreadMessageInViewIntersectionObserver`.
The `.info-message` margin is greater than the common margin
of `.message-list-and-composer__message-list` items,
so they two collapse, thus `.message-observer-bottom`
is pushed to the bottom.
Regular message bubbles do not have such margin, so this bug
does not apply to them.

image
image

The look, before / after:

 image image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Update last used app icons immediately after sending a new app
 
 ### Changed
+- adjust distance between info messages to match Delta Chat for Android #5244
 - tauri: macOS: webxdc: Remove the nowhere-proxy to support pre-14 macOS. #5202
 - reword 'Save As' to 'Export Attachment' to have a clearer cut to 'Save' #5245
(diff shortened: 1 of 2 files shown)


Simon-Laux (approved):
LGTM

PR #2160, opened 2021-03-03
Change "Download attachment" to "Export attachment"

Jikstra:
Fixes https://github.com/deltachat/deltachat-desktop/issues/2066

PR #5245, opened 2025-06-30
reword 'Save As' to 'Export Attachment'

r10s:
in fact, it is 'Export Attachment' on android since forever, and it was changed on desktop to 'Save As' for forgotten or today minor reasons at https://github.com/deltachat/deltachat-desktop/pull/2546

now, that we have 'Save' used for 'Saving Messages' on all platforms, it makes sense to go back to 'Export Attachment' - we've seen that coming, that 'Save' and 'Save As' is a too close wording for different things, but now we get real user feedback and it is time to iterate :)

that way, we use the same term again for the same functionality.

ftr, 'Download Attachment' was bad as well, see
https://github.com/deltachat/deltachat-desktop/pull/2160

ftr, could be shortened to just "Export", but that is another consideration

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ### Changed
 - tauri: macOS: webxdc: Remove the nowhere-proxy to support pre-14 macOS. #5202
+- reword 'Save As' to 'Export Attachment' to have a clearer cut to 'Save' #5245
 
 ### Fixed
 - tauri: remember webxdc app windows' position and size between app re-launches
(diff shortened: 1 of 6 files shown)


r10s:
> Forward to saved messages

that make sense, but is too long - and is then shortcutted to "Save" :)

> bookmark

that was considered as well (see paragraph "previous considerations" at https://github.com/chatmail/core/pull/5606), however, it does not really fit, as actually things are "saved" and "copied" (even the chat is called "Saved Messages").

apart from that, "export" somehow underlines the privacy, as it is not just a document flowing around somewhere. "export" also underlines that it is not the same type of "as used internally" - there is no "open", which usually belongs to "save as" and also not fitting "save".  also, "export" was used for years before without much issues afaik, the change to "Save As" did not go back to some real user feedback we got, afaik.

all minors, sure, but all in all, i think "export" is just fine

PR #5247, opened 2025-07-01
refactor: add a docstring for `throttle`

WofWca:


packages/shared/util.ts
```diff
@@ -15,6 +15,10 @@ export function isInviteLink(url: string) {
   return url.startsWith('https://i.delta.chat/') && url.includes('#')
 }
 
+/**
+ * Like [Lodash `throttle`](https://lodash.com/docs/4.17.15#throttle)
+ * with default options (`{ leading: true, trailing: true }`).
```
Simon-Laux, line 20: could be a bit more descriptive, like i never really used `lodash` (despite knowing what it is), others (especially younger[^1] devs) may also not be familiar with this reference

[^1]: younger as in js experience not age

WofWca: Well IDK, following the link and reading Lodash docs should work IMO.

PR #5255, opened 2025-07-03
Upgrade to core 1.160.0

nicodh:
I had to add 

```
[patch.crates-io]
parking_lot = "=0.12.4"
```
to Cargo.toml to avoid:

```
error: failed to select a version for `parking_lot`.
    ... required by package `hickory-resolver v0.25.2`
    ... which satisfies dependency `hickory-resolver = "^0.25.2"` (locked to 0.25.2) of package `deltachat v1.160.0 (https://github.com/chatmail/core?tag=v1.160.0#a40337f4)`
    ... which satisfies git dependency `deltachat` of package `deltachat-tauri v1.59.2 (/home/ndh/projekte/deltachat/deltachat-desktop/packages/target-tauri/src-tauri)`
versions that meet the requirements `^0.12` (locked to 0.12.3) are: 0.12.3
```
Not sure if there is a better way to fix that.

I also added a missing entry for the electron upgrade

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@
 - tauri: remember HTML email viewer window position / size for all HTML messages together, instead of separately for each individual message #5171
 - tauri: fix fullscreen media view zoom, pan, pinch not working quite right #5200
 - tauri: fix fullscreen avatar for selfavatar #5240
+- upgrade `@deltachat/stdio-rpc-server` and `deltachat/jsonrpc-client` to `1.160.0`
+- upgrade electron from 34 to 37
 
 <a id="1_59_2"></a>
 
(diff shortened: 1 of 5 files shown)


WofWca:
> Not sure if there is a better way to fix that.

Editing Cargo.toml followed by `cargo update deltachat deltachat-jsonrpc` (or just `cargo update`) fixes it.

nicodh:
> > Not sure if there is a better way to fix that.
> 
> Editing Cargo.toml followed by `cargo update deltachat deltachat-jsonrpc` (or just `cargo update`) fixes it.

Ok, I'm pretty sure I tried that (but is already some time ago...)
So I removed the entry [patch.crates-io]

nicodh:
> Have you checked the "recent apps" acts as intended, since [chatmail/core#6875](https://github.com/chatmail/core/pull/6875) ?

Our implementation still shows the messages ordered by IDs - so that's a followup to this PR. I already have it on my list

WofWca (approved):

PR #5256, opened 2025-07-03
Show 0 B instead ? if file size is 0, undefined or null

### LINKED ISSUE #5253: Missing file size of an empty file, just show a question mark
I noticed this while sharing some files, one of it was empty, after this I noticed that on the empty file it does not show like on the other ones a size as example 0 B instead it was just showing a "?". On phone, it was missing completely, just an empty buggy message (For this I opened an [issue](https://github.com/deltachat/deltachat-android/issues/3796) too).

- Operating System (Linux/Mac/Windows/iOS/Android): Manjaro
- Delta Chat Version: 1.58.2
- Expected behavior: Showing the size of the file, as on other files
- Actual behavior: It just shows '?' instead of 0 B
- Steps to reproduce the problem: Sharing an empty File
- Screenshots:
Image

nicodh:
Since the question marks provides no additional ifo for the user I think we do not need to distinguish if the size is 0, null or undefined in the UI

resolves #5253

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - tauri: fix fullscreen media view zoom, pan, pinch not working quite right #5200
 - tauri: fix fullscreen avatar for selfavatar #5240
 - fix: showing 0 instead ? as size for empty files #5253
+- show avatar for deleted saved messages #5221
 
 <a id="1_59_2"></a>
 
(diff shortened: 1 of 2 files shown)


nicodh:
> IDK IDK To me it seems like we rather need to return `null` instead of `0` if there is an error in getting file size here:

> So I'd at least remove `?? 0` since fileBytes is always a number.

So if fileBytes is always a number (which is expected since the type is number) we don't need any fallback. But as we know often things are not as they should and for that case I'd prefer showing a 0 instead an exception.

WofWca:
> the question marks provides no additional ifo for the user

I don't see how that's the case. 0 means "0", "?" means "we don't know". Why show the incorrect info.

PR #5260, opened 2025-07-06
Order app icons by last update

nicodh:
Preserve the order core provides

#skip-changelog

### DIFF
--- a/packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
+++ b/packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
@@ -161,15 +161,23 @@ export default function MainScreen({ accountId }: Props) {
       null,
       null
     )
-    mediaIds.reverse() // newest first
     const mediaLoadResult = await BackendRemote.rpc.getMessages(
       accountId,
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
```diff
@@ -161,15 +161,23 @@ export default function MainScreen({ accountId }: Props) {
...
     const mediaLoadResult = await BackendRemote.rpc.getMessages(
       accountId,
       mediaIds.slice(0, maxIcons)
     )
-    const lastMessages = Object.values(mediaLoadResult).filter(
-      result => result.kind === 'message'
-    )
-    setLastWebxdcApps(lastMessages.reverse()) // show newest first
+    const lastUpdatedApps = mediaIds
+      .reverse()
```
WofWca, line 169: I'd prefer the original way, of simply mutating the array after getting it, so that it's harder to accidentally mutate the array another time in the future.

WofWca (approved):
Didn't test, but looks alright, and should do what we want.

PR #5262, opened 2025-07-06
Update truncated text width on resize

nicodh:
Since webxdcInfo is loaded after message is loaded we have to care about height changes (i.e. avoid them to avoid wrong height calculation in message list)

followup for #5227

#skip-changelog

### DIFF
--- a/packages/frontend/src/components/message/Message.tsx
+++ b/packages/frontend/src/components/message/Message.tsx
@@ -62,6 +62,7 @@ import type { PrivateReply } from '../../hooks/chat/usePrivateReply'
 import type { JumpToMessage } from '../../hooks/chat/useMessage'
 import { mouseEventToPosition } from '../../utils/mouseEventToPosition'
 import { useRovingTabindex } from '../../contexts/RovingTabindex'
+import { useRpcFetch } from '../../hooks/useFetch'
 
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/message/Message.tsx
```diff
@@ -1161,22 +1161,54 @@ function WebxdcMessageContent({
...
+    let infoText = info.document + ' - ' + info.name
+
+    if (infoText.length > maxLength) {
+      infoText =
+        truncateText(info.document, maxLength - info.name.length) +
+        ' - ' +
+        info.name
+    }
+
+    return infoText
```
WofWca, line 1211: Is this impossible to solve with CSS? Isn't this just "no linebreaks with ellipsis"?

nicodh: Well, the summary could get a fix width with a number of media queries, but the document/name string should be treated with priority for the name, so the width of the document depends on the length of the name

WofWca: > but the document/name string should be treated with priority for the name

Would `flex-grow` / `flex-shrink` work for this? I haven't reviewed what the code does yet just so as you know.

PR #5263, opened 2025-07-07
Update translations

nicodh:


WofWca (approved):
I didn't go through all strings, but this only affects translation files.

This also adds a new language. Our build script will pick it up, right?

PR #5264, opened 2025-07-07
Avoid line breaks in webxdc-info

nicodh:
#skip-changelog

### DIFF
--- a/packages/frontend/scss/message/_message.scss
+++ b/packages/frontend/scss/message/_message.scss
@@ -1,6 +1,9 @@
 @use '../mixins';
 @use '../variables';
 
+// limit for text width, needed to trigger ellipsis
+$info-text-max-width: 400px;
+
 .message-wrapper {
   --shortcut-menu-visibility: hidden;
   --shortcut-menu-multiplier: 0;
@@ -121,10 +124,15 @@
         // firefox does not support this, but chromium and webkit do https://caniuse.com/?search=user-drag
         -webkit-user-drag: none;
       }
+      // nowrap and min-height address the problem that
+      // the height of a message bubble should not change
+      // after it's first load, even if the content changes
+      // after loading webxdcInfo async
+      // see https://github.com/deltachat/deltachat-desktop/issues/3753
       div.info-text {
         display: flex;
         flex-direction: row;
-        max-width: 400px;
+        max-width: $info-text-max-width;
         min-width: 0;
         white-space: nowrap;
         font-size: medium;
@@ -132,15 +140,13 @@
         gap: 3px;
       }
       div.name {
-        flex-shrink: 0;
-        flex-grow: 1;
         min-width: 0;
         overflow: hidden;
         text-overflow: ellipsis;
         white-space: nowrap;
       }
       div.document {
-        flex-shrink: 1;
+        flex-shrink: 999999999999999;
         min-width: 0;
         overflow: hidden;
         text-overflow: ellipsis;
@@ -157,7 +163,8 @@
         white-space: nowrap;
         overflow: hidden;
         text-overflow: ellipsis;
-        max-width: 400px;
+        max-width: $info-text-max-width;
+        min-height: 18px;
       }
       div.experimental {
         color: var(--colorDanger);
--- a/packages/frontend/scss/message/_message.scss
+++ b/packages/frontend/scss/message/_message.scss
@@ -164,7 +164,19 @@ $info-text-max-width: 400px;
         overflow: hidden;
         text-overflow: ellipsis;
         max-width: $info-text-max-width;
-        min-height: 18px;
+
+        $font-size: 0.875rem;
+        $line-height: 1.15;
+        font-size: $font-size;
+        line-height: $line-height;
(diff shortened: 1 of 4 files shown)


packages/frontend/scss/message/_message.scss
```diff
@@ -121,9 +121,43 @@
...
+      div.name {
+        flex-shrink: 0;
+        flex-grow: 1;
+        min-width: 0;
+        overflow: hidden;
+        text-overflow: ellipsis;
+        white-space: nowrap;
+      }
+      div.document {
+        flex-shrink: 1;
```
WofWca:
```suggestion
      div.name {
        min-width: 0;
        overflow: hidden;
        text-overflow: ellipsis;
        white-space: nowrap;
      }
      div.document {
        flex-shrink: 999999999999999;
```

This removes `flex-shrink: 0;` from name, such that it never overflows the message bubble (e.g. if the name is super long), while making sure that "document" shrinks before the name.

packages/frontend/src/components/message/Message.tsx
```diff
@@ -1191,13 +1191,15 @@ function WebxdcMessageContent({
...
       />
       <div
-        className='name'
+        className='info-text'
         title={`${info.document ? info.document + ' \n' : ''}${info.name}`}
       >
-        {info.document && truncateText(info.document, 24) + ' - '}
-        {truncateText(info.name, 42)}
+        <div className='document'>
+          {info.document && truncateText(info.document, 32)}
```
WofWca: `truncateText` seems to be unnecessary now, right? I'm not suggesting to remove it, but just to understand what's going on. If so, maybe it's worth adding a comment?

packages/frontend/scss/message/_message.scss
```diff
@@ -121,9 +121,43 @@
...
+        }
+        &:empty {
+          display: none;
+        }
+      }
+      div.summary {
+        white-space: nowrap;
+        overflow: hidden;
+        text-overflow: ellipsis;
+        max-width: 400px;
```
WofWca: Let's declare a variable for the two `400px` values?

WofWca (commented):
So this supersedes https://github.com/deltachat/deltachat-desktop/pull/5262.

This solves the line break issue!

But note that IMO this only incidentally solves the "chat scrolls up" problem, https://github.com/deltachat/deltachat-desktop/pull/5227#discussion_r2173774006. See what happens on reload:

https://github.com/user-attachments/assets/b91049ca-4c29-4450-bfef-1c0040312cad

Note the brief "Info missing!".
This is due to the fact that, as I mentioned in https://github.com/deltachat/deltachat-desktop/pull/5227#discussion_r2173780954, `isLoadingWebxdcInfo` is incorrectly set to `false` on initial render, which causes the message bubble to be taller than when `isLoadingWebxdcInfo` gets set to `true`. When this issue is solved, then the "scroll up" issue will come back. In turn, I suggest to solve this by making sure that the "loading" state is never shorter (in height) than the final width, that is by making sure `summary` is not empty on the "loading" state.

I can make an MR for this.

packages/frontend/scss/message/_message.scss
WofWca: Done.

packages/frontend/src/components/message/Message.tsx
```diff
@@ -1191,13 +1190,13 @@ function WebxdcMessageContent({
         // because there is a button below that does the same
       />
       <div
-        className='name'
+        className='info-text'
         title={`${info.document ? info.document + ' \n' : ''}${info.name}`}
       >
-        {info.document && truncateText(info.document, 24) + ' - '}
-        {truncateText(info.name, 42)}
+        <div className='document'>{info.document && info.document}</div>
```
WofWca:
```suggestion
        <div className='document'>{info.document}</div>
```

packages/frontend/scss/message/_message.scss
WofWca: Done.

packages/frontend/scss/message/_message.scss
```diff
@@ -121,9 +124,47 @@
...
+        &:empty {
+          display: none;
+        }
+      }
+      div.summary {
+        white-space: nowrap;
+        overflow: hidden;
+        text-overflow: ellipsis;
+        max-width: $info-text-max-width;
+        min-height: 18px;
```
WofWca:
```suggestion

        $font-size: 0.875rem;
        $line-height: 1.15;
        font-size: $font-size;
        line-height: $line-height;
        // Ensure that even if the element is empty
        // (e.g. while we're loading `webxdcInfo`),
        // it still takes height, so that when the info finally loads
        // and `summary` is present, the message bubble doesn't get taller,
        // which would result in the message list getting scrolled.
        // See
        // https://github.com/deltachat/deltachat-desktop/pull/5227#discussion_r2173774006.
        min-height: calc($font-size * $line-height);
```

WofWca:
I re-tested it. Works well! Both the "loading" state and further changes to `summary` and `document` post-load.

PR #5265, opened 2025-07-07
fix: races in "Recent apps", improve perf

WofWca:
~~# skip-changelog because I have not seen the bug appear~~.
Edit:
Actually, I've seen it a few times. Maybe need to add an entry after all.

TODO:
- [x] Merge #5230


Simon-Laux:
needs to be rebased, before I can really review it.

WofWca:
Rebased.

nicodh (approved):

Changelog entry not needed, so far there is no bug report for that...

WofWca:
> Changelog entry not needed, so far there is no bug report for that...

Well IDK if this means that we don't need an entry.

Simon-Laux:
> Changelog entry not needed, so far there is no bug report for that...

if there is a change, then it should be in the changelog unless it is a fix for a bug that was not released yet. (also if it is a change only relevant to developers, then it could also be omitted if it is small)

PR #5266, opened 2025-07-07
fix: "recent apps" section showing least recent

WofWca:
Looks like https://github.com/deltachat/deltachat-desktop/pull/5260
didn't work out as intended, I should have tested.

I have verified that this works, it shows the most recent app first.
Make sure to `pnpm install` before testing yourself.

#skip-changelog


nicodh (approved):
You are totally right! I moved the reverse() as a last small refactoring before pushing, not being aware that the slice then happens on the wrong order.

PR #5269, opened 2025-07-09
New condition for preview builds

nicodh:
To reduce the size of Downloads on https://download.delta.chat, till we have a better solution

Now a line 
#public-preview
is needed in the PR description to trigger the upload of a preview on download.delta.chat

### DIFF
--- a/.github/workflows/build-preview.yml
+++ b/.github/workflows/build-preview.yml
@@ -115,8 +115,8 @@ jobs:
        PR_ID: ${{ steps.prepare.outputs.prid }}
        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     - name: Upload result to artifacts
+      if: contains(github.event.pull_request.body, '#public-preview') == false
       id: upload-github
-      if: steps.upload.outcome == 'failure'
       uses: actions/upload-artifact@main
       with:
         name: ${{ matrix.os }} output
--- a/.github/workflows/build-preview.yml
+++ b/.github/workflows/build-preview.yml
@@ -98,9 +98,6 @@ jobs:
         move Delta*Chat*.exe preview/deltachat-desktop-${{ steps.prepare.outputs.prid }}.portable.exe
         dir preview
         cd ..
-    # Debugging
-    - name: Debug PR body
-      run: echo "${{ github.event.pull_request.body }}"
     # Upload Step
     - name: upload folder
       id: upload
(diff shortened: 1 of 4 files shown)


WofWca:
How about simply uploading them to GutHub artifacts instead?

nicodh:
> How about simply uploading them to GutHub artifacts instead?

Yes of course. That can happen anyway. I can add it.

nicodh:
Ok, just tested it. Works now. Artifact previews are generated if no previews are uploaded to download.delta.chat

.github/workflows/build-preview.yml
```diff
@@ -113,8 +115,8 @@ jobs:
        PR_ID: ${{ steps.prepare.outputs.prid }}
        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     - name: Upload result to artifacts
+      if: contains(github.event.pull_request.body, '#public-preview') == false
```
WofWca:
```suggestion
      if: contains(github.event.pull_request.body, '#public-preview') == false || steps.upload.outcome == 'failure'
```

Don't we want to keep the "failure" condition? (my syntax might be wrong)

.github/workflows/build-preview.yml
```diff
@@ -3,6 +3,7 @@ name: Preview Builds
 
 on: 
   pull_request:
+    types: [opened, synchronize, reopened, edited]
```
WofWca, line 6: A comment about why we'd want to re-run the workflow on edit would be nice IMO.

nicodh: I wrote a line in DEVELOPMENT.md already

This also works if the line is added later, since the upload step is triggered on "edit" also.

docs/DEVELOPMENT.md
```diff
@@ -107,9 +107,10 @@ If you're unsure it's always safe to run `pnpm -w fix` to fix everything. If you
 
 ### CI github actions
 
-We have several [github actions](../.github/workflows/) configured to be executed for each PR. These include code validation, tests and preview builds which are downloadable from the artifacts and from https://download.delta.chat/desktop/preview/
+We have several [github actions](../.github/workflows/) configured to be executed for each PR. These include code validation, tests and preview builds which are downloadable from the artifacts.
+Previews can be uploaded to https://download.delta.chat/desktop/preview/ by adding a line #public-preview to the PR description. This also works if the line is added later, since the upload step is triggered on "edit" also.
```
Simon-Laux:
```suggestion
Previews can be uploaded to https://download.delta.chat/desktop/preview/ by adding a line "#public-preview" to the PR description. This also works if the line is added later, since the upload step is triggered on "edit" also.
```

nicodh:
Otherwise we woulld need to add a commit for every change to trigger a new check. At least comments don't trigger a new check.

PR #5273, opened 2025-07-09
Slightly increase contrast for unread badge on pinned muted chats. 

Simon-Laux:
This was especially need the in the dark theme.

| Before | After |
|------------|--------|
| image| image|

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 - tauri: fix fullscreen avatar for selfavatar #5240
 - fix: showing 0 instead ? as size for empty files #5253
 - show avatar for deleted saved messages #5221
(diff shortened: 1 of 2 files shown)


r10s:
this _decreases_ the contrast between pinned and not pinned chats. 

resulting in pinned chats being noticed worse. also hovering is off now.

also, the badge has still bad contrast to background in dark mode.

to solve the issue on point, and not open unrelated issues and discussions, better make the badge lighter in dark mode - light mode is fine, as pointed out above

r10s:
to keep it simple, the dark-muted-unread-badge can be light unconditionally, so also for not-pinned chats.

let's go for the same grey as the dark-muted-unread-badge in the account switcher.

PR #5277, opened 2025-07-10
Prepare release 1.60.1

nicodh:
Although there no known serious bugs so far with core 1.160.0, we don't want to risk anything and will do the next core update when core version 2.0.0 is out

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## [Unreleased][unreleased]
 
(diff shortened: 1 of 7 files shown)


WofWca (approved):
Since 1.60.0 has not been posted to the stores yet, it's OK to downgrade.

WofWca:
Also it's correct that this is branched off of 1.60.0 and not current `main`.

PR #5283, opened 2025-07-11
refactor: fix potential ChatListItem `memo()` bugs

WofWca:
We do not re-render `ChatListItem`
even when `onContextMenu` prop changes,
due to the fact that we don't check if this prop changed
in our `areEqual` function.

This commit makes it harder to introduce a bug
when introducing new props for `ChatListItem`,
and potentially fixes some existing bugs.

I have verified that the component still doesn't re-render unnecessarily,
e.g. when switching to another chat.

TODO:
- [x] merge the base branch #5275

### DIFF
--- a/packages/frontend/src/components/chat/ChatListItemRow.tsx
+++ b/packages/frontend/src/components/chat/ChatListItemRow.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useCallback, useRef } from 'react'
 import { areEqual } from 'react-window'
 
 import ChatListItem, {
@@ -69,20 +69,36 @@ export const ChatListItemRowChat = React.memo<{
     activeContextMenuChatId,
   } = data
   const chatId = chatListIds[index]
+  const chat = chatCache[chatId]
+
+  // Using refs to avoid re-renders, because `ChatListItem` is memoized.
+  // TODO `useRef` docs recomment against updating refs during rendering.
+  // Is it bad in this case?
+  const onChatClickRef = useRef(onChatClick)
+  onChatClickRef.current = onChatClick
+
+  const onContextMenu = useCallback(
+    (event: React.MouseEvent) => {
+      if (chat?.kind === 'ChatListItem') {
+        openContextMenu(event, chat, selectedChatId)
+      }
+    },
+    [chat, openContextMenu, selectedChatId]
+  )
+  const onContextMenuRef = useRef(onContextMenu)
+  onContextMenuRef.current = onContextMenu
 
   return (
     <li style={style}>
       <ChatListItem
         roleTab={roleTabs}
         isSelected={selectedChatId === chatId}
-        chatListItem={chatCache[chatId] || undefined}
-        onClick={onChatClick.bind(null, chatId)}
-        onContextMenu={event => {
-          const chat = chatCache[chatId]
-          if (chat?.kind === 'ChatListItem') {
-            openContextMenu(event, chat, selectedChatId)
-          }
-        }}
+        chatListItem={chat}
+        onClick={useCallback(() => onChatClickRef.current(chatId), [chatId])}
+        onContextMenu={useCallback(
+          (event: React.MouseEvent) => onContextMenuRef.current(event),
+          []
+        )}
         isContextMenuActive={activeContextMenuChatId === chatId}
       />
     </li>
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/chat/ChatListItemRow.tsx
```diff
@@ -59,19 +59,35 @@ export const ChatListItemRowChat = React.memo<{
     activeContextMenuChatId,
   } = data
   const chatId = chatListIds[index]
+  const chat = chatCache[chatId]
+
+  // Using refs to avoid re-renders, because `ChatListItem` is memoized.
+  // TODO `useRef` docs recomment against updating refs during rendering.
+  // Is it bad in this case?
+  const onChatClickRef = useRef(onChatClick)
+  onChatClickRef.current = onChatClick
```
Simon-Laux, line 78: why is this also necessary for `onChatClick`?

packages/frontend/src/components/chat/ChatListItemRow.tsx
Simon-Laux, line 1: why `useRef` instead of `useMemo`/`useCallback`?

packages/frontend/src/components/chat/ChatListItemRow.tsx
WofWca: Because it also changes on chat changes, resulting in all chat list items getting re-rendered

packages/frontend/src/components/chat/ChatListItemRow.tsx
WofWca: Because those do change when dependencies change, resulting in re-renders.

packages/frontend/src/components/chat/ChatListItemRow.tsx
Simon-Laux: wouldn't it make more sense to have the ref inside of the MainScreen Component where the changing is introduced? (because select function chat changes - we could wrap select chat into a ref and then use the ref in the onChatClick function in the main screen)

that would also create less react references in total

WofWca: This is a hack, that's only needed inside of this component, so I wanted to keep it limited to just this component. It's easier to reason about it when it's limited, I think.

WofWca: But not that I'm too strongly opposed to your idea. Do you think your idea is better?

Simon-Laux: > Do you think your idea is better?

yes. reasoning:

I think it's better to have more pure/dumb components without much own state. 
And in this case I think we should address the "changing state, even though it is nothing that needs to trigger a rerender" problem close to the root where it is introduced once instead of executing a similar workaround many times later of the component tree.

WofWca: Well I would agree, unless this was a hack.

I kind of don't get why this isn't a common pattern.

WofWca:
Rebased.

packages/frontend/src/components/chat/ChatListItemRow.tsx
WofWca: > I kind of don't get why this isn't a common pattern.

OK, I think I've come up with some differences. One big example being `async` callbacks. `async` callbacks would capture all their dependencies, and the dependencies would not change while the callback is being executed. Whereas `ref`s _can_ change while a callback is being executed.
In this regard I think using the `ref` hack is fine here.

WofWca: So, @Simon-Laux do you think it's acceptable at the current state?

Simon-Laux: I still would prefer to have such thing/hacks in the parent component, so that the chatlistitem row doesn't get changing props in those cases.
Like currently you create many refs when there could just be one in the parent, or am I missing something here?

But I won't stand in the way of merging this. It's a somewhat theoretical discussion at this point and we can always change it and simplify it in the future.

WofWca: > create many refs when there could just be one in the parent

I don't get it, what would be the TypeScript type of this ref?

> I still would prefer to have such thing/hacks in the parent component

I'll re-state my main point just in case: since this is a hack, I want to keep its scope as narrow as possible, i.e. keep it in `ChatListItemRowChat` and not `ChatList`.

PR #5286, opened 2025-07-14
fix copying of links into richtext editors

Simon-Laux:
we set it to empty anchor/fragment as a theoretical security precaution, but it harms usability and seems to serve no purpose in practice

### DIFF
--- a/packages/frontend/src/components/message/Link.tsx
+++ b/packages/frontend/src/components/message/Link.tsx
@@ -86,7 +86,7 @@ export const LabeledLink = ({
   }
   return (
     <a
-      href={'#' + target}
+      href={target}
       x-target-url={target}
       title={realUrl}
       onClick={onClick}
@@ -210,7 +210,7 @@ export const Link = ({
 
   return (
     <a
-      href='#'
+      href={asciiUrl}
       x-target-url={asciiUrl}
       title={asciiUrl}
       onClick={onClick}
--- a/packages/frontend/src/components/message/MessageMarkdown.tsx
+++ b/packages/frontend/src/components/message/MessageMarkdown.tsx
@@ -224,7 +224,7 @@ function EmailLink({
 
   return (
     <a
-      href={'#'}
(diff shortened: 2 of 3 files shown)


nicodh (approved):
Reviewed by reading

packages/frontend/src/components/message/MessageMarkdown.tsx
```diff
@@ -224,7 +224,7 @@ function EmailLink({
 
   return (
     <a
-      href={'#'}
+      href={`mailto:${email}`}
```
WofWca, line 227: Shouldn't we ensure to `preventDefault` here?

packages/frontend/src/components/message/Link.tsx
```diff
@@ -210,7 +210,7 @@ export const Link = ({
 
   return (
     <a
-      href='#'
+      href={asciiUrl}
```
WofWca, line 213: Why not `target`? (I don't know much about Punycode)

Simon-Laux: I think it doesn't matter that much. sure you may copy punycode encoded url if it contains punycode, but I don't think that is an issue.
You could say it is a feature that you can see who the host actually is, I don't know of many legitimate domains that use Unicode.

the reason I did it like this here was that I just copy pasted the variable from `title` and `x-target-url`

packages/frontend/src/components/message/MessageMarkdown.tsx
Simon-Laux: we could do that for the browser version, but in electron it is alread blocked by the code that blocks navigation.

Simon-Laux: on tauri it is also a problem:
image

Simon-Laux: made a pr https://github.com/deltachat/deltachat-desktop/pull/5292

PR #5287, opened 2025-07-14
Api change chat type

### LINKED ISSUE #5258: Adapt to new Broadcast Channel chat types
Core release 2.0.0 includes https://github.com/chatmail/core/pull/6901, which includes a new API for broadcasts, which are now called "Channel" in the user-facing UI.

These steps are necessary now (also see https://github.com/deltachat/deltachat-android/pull/3783/):
- [x] Rename "broadcast list" to "channel" in the UI. There are new stock strings in which `broadcast_list` is replaced with `channel`
- [x] Adapt to the new chat types `InBroadcast` and `OutBroadcast` for incoming / outgoing broadcast channels, where the former is similar to a `Mailinglist` and the latter is similar to a `Broadcast` (which was removed).
- [x] Use the new `create_broadcast()` API, which replaced the deprecated `create_broadcast_list()` API
- [x] Make it possible to leave channels
    - In a chat, if the chat is an InBroadcast, and it's not a contact request, then the `Leave` menu option is shown with the translated stock string `menu_leave_channel` as its label.
    - If the user clicks on it, the confirmation dialog has `menu_leave_channel` (rather than `menu_leave_group`) as its positive option.
- [x] Test whether everything works fine using the "UI Testing" list at https://github.com/chatmail/core/issues/6884
- [x] Wait for a core release with https://github.com/chatmail/core/pull/6901

Broadcast lists are experimental, so, it would be fine if you postpone some of these things (e.g. the renaming of strings) and the UX isn't super great. But you should at least make sure that the app doesn't crash when core returns a `{In|Out}Broadcast` chat type.

Note: In the code (except for strings), we use the name "Broadcast" rather than "Channel", because the word "channel" already appears a lot in the code already, which would make it hard to grep for it.

nicodh:
resolves #5258

- [x] Rename "broadcast list" to "channel" in the UI. There are new stock strings in which broadcast_list is replaced with channel
- [x] Adapt to the new chat types InBroadcast and OutBroadcast for incoming / outgoing broadcast channels, where the former is similar to a Mailinglist and the latter is similar to a Broadcast (which was removed).
- [x] Use the new create_broadcast() API, which replaced the deprecated create_broadcast_list() API
- [x] Make it possible to leave channels
- [x] Test whether everything works fine using the "UI Testing" list at https://github.com/chatmail/core/issues/6884
- [x] as author of a channel change avatar of channel
- [x] block channel contact request (as recipient)
- [x] receivers can't edit the mailinglist profile see #5299


packages/frontend/src/hooks/chat/useChatDialog.ts
```diff
@@ -124,11 +124,13 @@ export default function useChatDialog() {
     [openDialog]
   )
 
-  const openLeaveChatDialog = useCallback(
-    async (accountId: number, chatId: number) => {
+  const openLeaveGroupOrChannelDialog = useCallback(
+    async (accountId: number, chatId: number, isGroup: boolean) => {
```
WofWca, line 128: Wouldn't `chatType` be better as the argument?
Not a big deal though.

packages/frontend/src/components/chat/ChatListContextMenu.tsx
```diff
@@ -265,14 +272,20 @@ export function useChatListContextMenu(): {
                 action: onEncrInfo,
               },
             { type: 'separator' },
+            // Leave channel
+            chatListItem.chatType === C.DC_CHAT_TYPE_IN_BROADCAST &&
+              !chatListItem.isContactRequest && {
```
WofWca, line 280: Why the `isContactRequest` condition?
Isn't it better to merge this into `isGroup &&`, with only `label` being different?

packages/frontend/src/hooks/chat/useChatDialog.ts
r10s: note, that soon there will be a `createEmail` or in addition, see https://github.com/deltachat/deltachat-desktop/issues/5294 - so maybe iterate when targeting that.

on iOS, we finally created an enum for that, https://github.com/deltachat/deltachat-ios/pull/2770/files#diff-a1c19380608c8dd2b017e3fb0dee7a2e818d7c79c37c10ce6dc25f76f1997bcdR14

packages/frontend/src/components/chat/ChatListContextMenu.tsx
nicodh: A contact request of type IN_BROADCAST has the properties isSelfInGroup = true and isGroup = false. So the only way to distinguish that from a channel you already joined is the isContactRequest.

isGroup is used also for other conditions so it is not possible to merge both imo

WofWca: OK, then about my first question. Still, why do we only allow to leave the channel if it's not contact request? We don't have this condition for groups, so why only for channels?

nicodh: Is it even possible to invite someone to a group, you don't have a one-to-one chat already? We never had that condition for "leave group"

For a channel you get a ContactRequest even if you already have a one-to-one chat

WofWca: > For a channel you get a ContactRequest even if you already have a one-to-one chat

So what's the point of checking `isContactRequest` then?

iOS doesn't check `isContactRequest` when deciding whether to show "leave channel":

https://github.com/deltachat/deltachat-ios/blob/b7e8b941715cb8c05cf1f6aa7082754b45511f59/deltachat-ios/Controller/ProfileViewController.swift#L351-L356

nicodh: > so what's the point of checking isContactRequest then?

Just not to to show a "Leave chat" option if you have not yet accepted the request.
I talked to Hocuri before starting and he said the condition should be is InBroadcast and is not a contactRequest.
Makes sense to me.

nicodh: Although we still show "Leave chat" after you left since atm there is no flag to check if you already left

WofWca: Alright then.

WofWca (approved):
OK, this seems alright, but it's hard to review it in separation from https://github.com/deltachat/deltachat-desktop/pull/5288.
I suggest to merge, and re-evaluate everything in that MR.

WofWca:
And note that this does _not_ fully resolve https://github.com/deltachat/deltachat-desktop/issues/5258, e.g. because the new incoming channels show up as "Mailing list"s in the UI, apparently (unless I'm testing incorrectly):

image

nicodh:
And the subtitle is also wrong.

WofWca:
> And the subtitle is also wrong.

(not sure what you mean exactly but it was me who named it "new channel 07-19")

PR #5288, opened 2025-07-14
Upgrade to core 2.x

nicodh:
#public-preview

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,6 @@
 ### Changed
 - update translations (17-20-2025)
 - upgrade `@deltachat/stdio-rpc-server` and `deltachat/jsonrpc-client` to `2.3.0`
-  - API-Changes
-    - jsonrpc: Add CommandApi::create_group_chat_unencrypted() (#6927).
-    - [breaking] In ChatListItem, replace is_group and is_(out_)broadcast with chat_type property (#7003).
   - Features / Changes
     - Donation request device message (#6913).
 - remove resetEncryptionInfo


Simon-Laux:
core version in tauri is not updated yet

nicodh:
> core version in tauri is not updated yet

Does it need any adaptions? Just did a smoke test

Simon-Laux:
> Does it need any adaptions?

I don't think so, as long as it still compiles and there also were no changes to the electron backend/main-process code, then it should be fine.

_locales/_untranslated_en.json
```diff
@@ -79,5 +79,32 @@
...
+    "message": "Remove %1$s from channel?"
+  },
+  "chat_new_channel_hint": {
+    "message": "In a channel, recipients don't see who else is part of the channel."
+  },
+  "please_enter_channel_name": {
+    "message": "Please enter channel name"
+  },
+  "edit_channel": {
+    "message": "Edit Channel"
```
WofWca: We're supposed to simply pull existing translations, aren't we?

nicodh: Yes these will be obsolete after #5300 except please_enter_channel_name and "edit_channel"

packages/frontend/src/components/message/MessageList.tsx
```diff
@@ -37,8 +37,8 @@ import { markChatAsSeen } from '../../backend/chat'
 type ChatTypes =
   | C.DC_CHAT_TYPE_SINGLE
   | C.DC_CHAT_TYPE_GROUP
-  | C.DC_CHAT_TYPE_BROADCAST
-  | C.DC_CHAT_TYPE_MAILINGLIST
+  | C.DC_CHAT_TYPE_IN_BROADCAST
+  | C.DC_CHAT_TYPE_OUT_BROADCAST
```
WofWca, line 41: So, `DC_CHAT_TYPE_BROADCAST` was seemingly just renamed to `DC_CHAT_TYPE_OUT_BROADCAST`, but what about `DC_CHAT_TYPE_MAILINGLIST`? It seems to be still a thing in core, it doesn't seem like it's just a one-to-one replacement. `DC_CHAT_TYPE_IN_BROADCAST` has a different ID.

https://github.com/chatmail/core/blob/2cf979de53d0de045bc547071668592989891f53/deltachat-ffi/deltachat.h#L5749-L5781

https://github.com/chatmail/core/blob/22258f72693c3e184e5e0c7df3254690f8e8748f/src/constants.rs#L140-L169

Do you know a good resource to learn about this?

_locales/_untranslated_en.json
nicodh: Translations are updated

packages/frontend/src/components/message/MessageList.tsx
nicodh: Some comments in this PR are helpful: https://github.com/chatmail/core/pull/6901/files

I readded Mailinglist in some conditions

r10s: DC_CHAT_TYPE_MAILINGLIST and DC_CHAT_IN_BROADCAST are similar, but both exist. eg. we want to have different subtitles.

DC_CHAT_TYPE_MAILINGLIST should probably just not be changed, but DC_CHAT_IN_BROADCAST should be added.

for DC_CHAT_OUT_BROADCAST it is different: DC_CHAT_OUT_BROADCAST has replaced DC_CHAT_BROADCAST

nicodh: The main difference is that IN_BROADCAST is encrypted and MAILINGLIST is not

WofWca: So if a chat is `DC_CHAT_TYPE_MAILINGLIST`, then we're supposed to e.g. show "Mailing list" in the "View Chat" dialog, and not "Channel"? Am I reading this correctly?

https://github.com/deltachat/deltachat-desktop/issues/5258 says:

> Rename "broadcast list" to "channel" in the UI. There are new stock strings in which broadcast_list is replaced with channel

It's not entirely correct then?
Edit: OK, in my last question I mixed up "broadcast list" with "mailing list".

r10s: > So if a chat is DC_CHAT_TYPE_MAILINGLIST, then we're supposed to e.g. show "Mailing list" in the "View Chat" dialog, and not "Channel"? Am I reading this correctly?

yes. the string "Mailing list" should be shown for DC_CHAT_TYPE_MAILINGLIST where it was shown before, no changes required there.

nicodh: Is is still shown in the chat list header as before the change

nicodh: So to summarize: on a non chatmail account, receiving a "normal" mail sent like by some random mail client will be shown as "Mailinglist" - probable other recipients (multiple recipients or CC recipients) will be shown as "Group members"

Maybe we can add that to FAQ now we have a more clear distinction in the UI

r10s: > on a non chatmail account, receiving a "normal" mail sent like by some random mail client will be shown as "Mailinglist"

no, this is wrong. only "real, classic mailinglists" will become DC_CHAT_TYPE_MAILINGLIST. usually these are newsletters one has subscribed to, or spam.

"mail sent like by some random mail client" will become DC_CHAT_TYPE_GROUP or DC_CHAT_TYPE_SINGLE with is_encrypted() == false

maybe it was wrong, to bring up "mailinglist" wrt channels alltogether :) (tho https://github.com/deltachat/deltachat-desktop/issues/5258 only says that IN_BROADCAST is similar to MAILINGLIST, not to replace or change MAILINGLIST :)

nicodh: > no, this is wrong. only "real, classic mailinglists" will become DC_CHAT_TYPE_MAILINGLIST. usually these are newsletters one has subscribed to, or spam.

Ah - ok. So I had some newsletter which was shown as mailinglist that's why I misinterpreted them
We really need a thorough documentation for these different types

r10s: > We really need a thorough documentation for these different types

https://c.delta.chat/group__DC__CHAT__TYPE.html

nicodh: Yes I saw that, but I was not aware that Mailinglist in DC context is similar to a "usual" mailinglist.

And the documentation should cover all the different behaviours of the different chat types, like why can't I edit the group name here, but not there etc. Especially when we have another chat type channel.

Would be easier for developers and users :-)

But this is not a complaint, just a remark - I will also contribute to docs on that...

nicodh: @r10s  Is that comment correct?

https://github.com/deltachat/deltachat-desktop/pull/5288/files#diff-8360bd603e5389f39fc1f833ed81642507861965ac651d05d33a5a04183de159R332

r10s: which comment precisely are you referring to? the link opens the while diff of this PR for me :)

nicodh: Maybe you have to wait a bit until the browser processed the deep link?

It's about this comment:

 // We don't allow editing of non encryped groups (email groups)
  // i.e. changing name, avatar or recipients
  // since it cannot be guaranteed that the recipients will adapt
  // these changes (image is not shown at all in MTAs, group name is
  // just the subject and recipients are basically just an email
  // distribution list)

CHANGELOG.md
```diff
@@ -2,8 +2,25 @@
...
-- update translations (17-07-2025)
+### Breaking
+- after upgrade to `@deltachat/stdio-rpc-server` version `2.x` you can not go back to 1.x versions any more since account data is not backward compatible!
+
+### Changed
+- update translations (17-20-2025)
+- upgrade `@deltachat/stdio-rpc-server` and `deltachat/jsonrpc-client` to `2.3.0`
+  - API-Changes
+    - jsonrpc: Add CommandApi::create_group_chat_unencrypted() (#6927).
+    - [breaking] In ChatListItem, replace is_group and is_(out_)broadcast with chat_type property (#7003).
```
WofWca: This does not affect end users. I personally think that we therefore should not include this in the changelog.

WofWca (approved):
OK, I looked through the code again. Looks good. Addresses the breaking changes from core, as far as I can tell.

Playing around with this also showed no significant issues. The insignificant issues, such as "delete for all" being displayed in unencrypted groups are, as discussed in private chats, to be addressed later. Also there are some minor UI issues, like a not entirely correct title in "edit channel", but they can be addressed in separate MRs.

Let's move forward!

packages/frontend/src/components/message/MessageList.tsx
r10s: yes, the comment makes sense

PR #5289, opened 2025-07-14
feat: allow to select contacts in the forward dialog

### LINKED ISSUE #3030: show contacts in forward message dialog, too

Simon-Laux:
close #3030

### TODO

- [ ] style contact divider
- [ ] style contact and chat items so they have the same margin
- [ ] fix rowing tab index
- [ ] react to events - (but is this even necessary for a chat selector dialog?)


r10s:
closing and putting to resurrection for reasons mentioned in #3030, it is not a priority

PR #5291, opened 2025-07-15
Log catched useFetch errors

Simon-Laux:
Because we don't always handle them in the UI code and we need to be
aware of them when they happen regardless.

#skip-changelog


Simon-Laux:
> IDK about withoutStackTrace

the stack trace is location of the log statement, so not particularly interesting in our case, though If I should be wrong about this, it's easy to change

PR #5295, opened 2025-07-16
donation device message

r10s:
the message will be added with recent cores after 30 days if there are at least 100 messages - if not, 30 days later we check again. the message is not added twice

see https://github.com/chatmail/core/pull/6914 for more discussions and https://github.com/deltachat/deltachat-android/pull/3813#issuecomment-3068630592 for a **screenshot**


Simon-Laux (approved):

WofWca (approved):
Looks OK, didn't test.

Note that this will need a translations update to work.
Edit: and a core upgrade.
So this cannot be merged now.

nicodh:
> Note that this will need a translations update to work. Edit: and a core upgrade. So this cannot be merged now.

Translations are in #5300 
Core is in #5288

So once these are merged this should be fine

PR #5297, opened 2025-07-16
feat: multiple selection in chat list

WofWca:
image

To use it, simply hold Shift or Ctrl and click on the chat list items.
Keyboard is also supported.
Accessibility is taken into account
(`aria-selected`, `aria-multiselectable`).

This has minimal effect on non-multiselect behavior,
i.e. if you don't hold Shift or Ctrl, you won't see a big difference.
The most significant change in this regard is that
if you right-click a chat that is not the active chat,
it will get highlighted, and the active chat will get un-highlighted,
until the context menu gets closed.

Note that the active chat and the selected chats have the same look
(i.e. darker background).
I have tried to minimize confusion between them.
But in the end we might need to style them differently.
Although at the current state it also works quite well IMO.

Summary of the changes:
- Make `ChatListContextMenu` support multiple items
  (without changing behavior for single selection).
  The behavior if things such as "pin / unpin", "mute / unmute"
  is the same as in Android.
  The "Delete Chat" dialog also supports multiple items,
  and displays the to be deleted chat names.
- Rename `selectedChatId` to `activeChatId` in `ChatList.tsx`
- Introduce the `useMultiselect` hook. It basically takes `click`
  and `focus` events, and outputs the set of selected items.
  It basically depends on the existing `useRovingTabindex` hook
  for arrow key focus management.
  I have tried searching for existing such libraries,
  but didn't find one that's sufficiently generic and simple.
- Introduce `useChatListMultiselect` hook that wraps
  `useMultiselect`, and more concretely implements multiselect
  specifically for chat list. For example,
  resets selection when the active chat changes.
- Pass down `onClick` and `onFocus` handlers
  to chat list item components.
- Adjust `onContextMenu` to handle multiselect.

The more useful feature would be to multiselect messages,
but I decided to start with a simpler case,
as I once did with `useRovingTabindex`.

TODO:
- [x] Address TODOs in the code, clean up comments.
- [x] Get feedback.
- [x] Consider putting this behind a feature flag.
  Edit: We have decided that this is not needed.
- [x] Consider adding more UI, namely a "n chats selected" indicator.
  Edit: we have decided that this can be done later.
- [x] Add E2E tests, maybe a lot of them. But probably after getting feedback.
  https://github.com/deltachat/deltachat-desktop/pull/5431 - not a lot of them, but covers the basic stuff.
- [x] Write better MR description
  - [x] Maybe something about existing libraries.
  - [x] Summary of the changes: context menu to support multiple items, but otherwise unchanged; introduced the generic `useMultiselect` hook that basically relies on `useRovingTabindex`.
- [x] Merge the MR that this is based on: https://github.com/deltachat/deltachat-desktop/pull/5283
- [x] Add CHANGELOG entry.
- [x] Apparently it's planned to make a release with just https://github.com/deltachat/deltachat-desktop/pull/5423. If so, let's merge this MR after that release.

Otherwise this works quite well IMO.

#public-preview

### DIFF
--- a/packages/frontend/src/components/chat/ChatList.tsx
+++ b/packages/frontend/src/components/chat/ChatList.tsx
@@ -51,6 +51,11 @@ import { isInviteLink } from '../../../../shared/util'
 import { RovingTabindexProvider } from '../../contexts/RovingTabindex'
 import { useRpcFetch } from '../../hooks/useFetch'
 import { useSettingsStore } from '../../stores/settings'
+import { useMultiselect } from '../../hooks/useMultiselect'
+import { getLogger } from '@deltachat-desktop/shared/logger'
+import { useHasChanged2 } from '../../hooks/useHasChanged'
+
+const useMultiselectLog = getLogger('ChatListMultiselect')
 
 const enum LoadStatus {
   FETCHING = 1,
@@ -198,7 +203,7 @@ export default function ChatList(props: {
     showArchivedChats
   )
 
-  const { openContextMenu, activeContextMenuChatId } = useChatListContextMenu()
+  const { openContextMenu, activeContextMenuChatIds } = useChatListContextMenu()
   const createChatByContactId = useCreateChatByContactId()
   const { selectChat } = useChat()
 
@@ -346,25 +351,33 @@ export default function ChatList(props: {
   //   selectFirstChat()
   // )
 
+  const multiselect = useChatListMultiselect(
+    chatListIds,
+    activeChatId,
+    accountId
+  )
+
   const chatlistData: ChatListItemData = useMemo(() => {
     return {
       // This should be in sync with `olElementAttrs` of `ChatListPart`.
       roleTabs: true,
 
       activeChatId,
+      multiselect,
       chatListIds,
       chatCache,
       onChatClick,
       openContextMenu,
-      activeContextMenuChatId,
+      activeContextMenuChatIds,
     }
   }, [
     activeChatId,
+    multiselect,
     chatListIds,
     chatCache,
     onChatClick,
     openContextMenu,
-    activeContextMenuChatId,
+    activeContextMenuChatIds,
   ])
 
   const contactlistData: ContactChatListItemData = useMemo(() => {
@@ -480,6 +493,7 @@ export default function ChatList(props: {
                     // This should be in sync with `chatlistData.roleTabs`.
                     role: 'tablist',
                     'aria-orientation': 'vertical',
+                    'aria-multiselectable': multiselect != undefined,
 
                     // TODO perhaps `pref_` is not nice,
                     // we might need a separate string.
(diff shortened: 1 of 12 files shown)


r10s:
very nice!

and super useful - up to now, i do the multi-select things on macOS via a multi-device setup on an iphone :)

played a bit around on macOS, works mostly as expected, i also think, the highlighting is fine. adding even more colors, not sure. one also sees by title the context of the message list.

> To use it, simply hold Shift or Ctrl and click on the chat list items.

the Ctrl-click should probably select/deselect a single item; this does not work on macOS; here it is anyways the ⌘-key, which, however, does not work as well

WofWca:
Oh, macOS, right...
I'll handle it.

Thanks for testing!

WofWca:
I pushed a commit that should fix it.

packages/frontend/src/components/chat/ChatList.tsx
```diff
@@ -810,6 +824,173 @@ function useContactAndMessageLogic(
...
+  // we _don't_ want it to get added back to selection,
+  // to avoid, let's say, accidentally deleting a chat.
+  //
+  // For example, let's say the user had two chats selected, but then
+  // one of the selected chats got archived (e.g. from another device).
+  // Then they forgot about selecting it and went to delete some other chats:
+  // they selected two other chats for deletion, but then a message arrives
+  // to the archived chat, unarchiving it.
+  // This would suddenly make 4 chats selected, instead of 3,
+  // and they could unintentionally delete the one they archived.
```
nicodh, line 932: This seems to me a very rare use case: someone selects chats on one device and while they are selected he archives chats on another device?

packages/frontend/src/components/chat/ChatList.tsx
```diff
@@ -810,6 +824,173 @@ function useContactAndMessageLogic(
...
+      // is re-created when accountId changes.
+      // But let's do it for future-proofing.
+      accountId,
+    ]
+  )
+  // Put behind a ref to avoid re-renders.
+  const selectionId = useRef(selectionId_)
+  selectionId.current = selectionId_
+
+  const [multiselectState_, setMultiselectState_] = useState({
```
nicodh, line 885: What is the idea of the appended _ in the name? (same for  selectionId_)

Is that the same as a trailing _ which is much more common (even in our code base) to mark it as "non public"?

packages/frontend/src/components/chat/ChatList.tsx
```diff
@@ -810,6 +824,173 @@ function useContactAndMessageLogic(
...
+    // If you are going to change the initial selection,
+    // make sure to also check what happens when you invoke the context menu
+    // for a chat that is not selected.
```
nicodh: If the context menu is invoked the selection should reset anyway? (and did so in my tests)

packages/frontend/src/components/chat/ChatList.tsx
```diff
@@ -810,6 +824,173 @@ function useContactAndMessageLogic(
...
+    : // Make sure to have the active chat selected by default,
+      // e.g. for the case when the user wants to Ctrl + Click
+      // another chat.
+      // This is important, because the chat list items
+      // are only styled as "selected" based on `multiselect`,
+      // and not `activeChatId` (see `isSelected` in `ChatListItemRowChat`).
+      //
+      // TODO or should we do this though? Why complicate things?
+      // Just mark "selected" and "active" chat distinctly
+      // and we're good?
```
nicodh, line 916: I think it's fine and expected: active chat is always selected initially and it is possible to deselect it with Ctrl + Click

packages/frontend/src/components/chat/ChatListItemRow.tsx
```diff
@@ -60,30 +72,176 @@ export const ChatListItemRowChat = React.memo<{
...
+          multiselect
+            ? // When `multiselect` is provided, we don't need
+              // to also check `activeChatId === chatId`, because multiselect
+              // also follows it during "normal" usage,
+              // including switching the active chat with shortcuts
+              // and through notifications.
+              //
+              // TODO however, for completeness and to avoid confusion,
+              // we should probably add a distinct `isActive` prop,
+              // and a separate style for `isSelected`.
```
nicodh, line 228: a separate style would be nice and makes sense imo

packages/frontend/src/components/chat/ChatList.tsx
```diff
@@ -810,6 +824,173 @@ function useContactAndMessageLogic(
...
+  //
+  // An alternative would be to simply reset (invalidate) selection
+  // on _any_ change to `chatListIds`,
+  // but that would not be nice, e.g. for example if only the order
+  // of chat list items changed, e.g. when receiving a new message.
+  //
+  // This also handles the more common case of simply resetting selection
+  // after you have deleted / archived selected chats.
+  //
+  // TODO should this be part of `useMultiselect`?
```
nicodh, line 942: Would be nice if easy to achieve, since Chatlist is huge already...

packages/frontend/src/hooks/useMultiselect.ts
```diff
@@ -0,0 +1,320 @@
...
+       *
+       * It does not matter whether the provided `item` is a member of
+       * {@link availableItems}.
+       *
+       * @returns whether the caller should `preventDefault()`
+       * the original event and not execute the regular click action.
+       */
+      onClick,
+      /**
+       * This mist be invoked when an item is focused,
```
nicodh:
```suggestion
       * This must be invoked when an item is focused,
```

nicodh (approved):
Reviewed by testing & reading

Thanks for the thorough comments, makes it much easier to understand!

Added some comments, no blocker though.

Open tasks: Only the CHANGELOG needs to be updated and the base MR merged. 
The other tasks are not mandatory imo
A counter would be nice but can be done later.

I would not mark it as experimental, since it's anyway only visible after a shift/ctrl click.

Not to be merged now, but if the 2.x has matured some more days.

packages/frontend/src/components/chat/ChatList.tsx
WofWca: Yes, it's to mark it as "you probably don't want to use it".
I'd prefer a leading underscore, but I was thrown off by the fact that we allow for such variables to be unused.

packages/frontend/src/components/chat/ChatList.tsx
WofWca: What could be different is that we wouldn't change selection on such an event, and wouldn't restore it to the currently active chat.

OK, I think this comment is not of much use, let's remove it.

packages/frontend/src/components/chat/ChatList.tsx
WofWca: > it is possible to deselect it with Ctrl + Click

But then you could end up in a state where none of the chats are selected, so no chat is marked as "active".

packages/frontend/src/components/chat/ChatList.tsx
WofWca: Actually, with the current implementation, with the code to which this comment applies present, it's possible to reproduce this from just one device: select 2 chats, then archive them, then get a message in one of the chats.

With the second device being used in my example in the code, I want to emphasize that "don't try to solve this by simply reset selection after performing a context menu action".

packages/frontend/src/components/chat/ChatList.tsx
WofWca: Let's see about that when we get to multiselecting messages.

packages/frontend/src/components/chat/ChatListItemRow.tsx
WofWca: BTW Firefox uses a different-colored border for multiselect, with a thicker border for the active tab.

WofWca:
Rebased.

packages/frontend/src/components/chat/ChatList.tsx
WofWca: Removed.

WofWca:
I removed a comment and fixed a typo. Nothing new to review.

pvagner:
Hello,
I have downloaded the latest preview testing app image for linux and I think I have found an issue which is not with this PR, but with chat list in general or with an orca screen reader on linux or with chromium.
Currently the list of conversations has role tablist and individual conversations have role=tab.
I assumed that tablist is not expected to support selecting multiple items. But I have found it's not the case by reading mdm at this url: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-multiselectable
Thus on linux orca reports when the conversation is selected for the first time but it then does not report if it's unselected or selected again for the second time.
I don't know how to verify this other than with a screen reader and my guess is that for multi select using roles such as list and listitem might result in better support accros browser engines and screen readers.

WofWca:
Hey!

> Thus on linux orca reports when the conversation is selected for the first time but it then does not report if it's unselected or selected again for the second time.

I'm failing to understand this part. How would a chat list item get unselected, without this MR? How would it get selected again?
Are you talking about "Live Regions" or...?

pvagner:
Okay, my steps are like this.

I have my main chatmail account.
I have grabbed the test version, started a new profile
and started a conversation.
My test profile is named testing.
My main profile is named pvagner.
When I am using up and down arrow keys to navigate in the list of conversations on the testing profile running the deltachat build from this pull request I have these two items in the list:
* pvagner
* device messages

So if conversation pvagner is in the focus, pressing down arrow key moves focus to the conversation device messages.
When the conversation device messages has focus pressing up arrow key moves focus to the conversation pvagner.

When navigating with arrow keys alone or when holding down ctrl key and navigating with arrow keys and no multi selection was attempted everything should get deselected and the active item should become selected i.e. aria-selected should be set to true. So the initial item when pressing down ctrl key should get selected.
When pressing space key and active item does not change I should be able to cycle aria-selected for the active item like a toggle.

So now back to my list of conversations as described.
Conversation pvagner is active.
I am pressing down arrow key making conversation device messages active ensuring nothing else is selected in the process.
I am pressing down ctrl key, keeping it held down, pressing up arrow key. From the screen reader perspective nothing is reportedly changing selection and conversation pvagner becomes active.
I am still keeping ctrl key held down and I'm pressing space bar to toggle the selection. Screen reader is reporting the conversation pvagner becomes selected.
I am still keeping ctrl key held down and pressing the spacebar to deselect the conversation pvagner. However screen reader is not reporting this.
Still holding down the ctrl key I am pressing space key multiple times and nothing is changing. Am I actually changing selection at this point or not?
With the ctrl key still held down I am pressing down arrow key. The conversation device messages becomes active. Pressing the space key with ctrl key still held down makes it selected.

I know my steps are very complicated but I would expect the multi selection to work in a similar way on how the most file managers are operating when browsing files. For example when selecting multiple files using any file manager on linux and windows.

Or are you envisioning it differently and I am missunderstanding?

Thanks and greetings

Peter

pvagner:
> I'm failing to understand this part. How would a chat list item get unselected, without this MR? How would it get selected again?

Oh, I am sorry I haven't answered this at all.
Without this MR we are not dealing with selections from the screen reader perspective, we are only looking at focus changes. So selecting and unselecting individual conversation items is not possible without this MR.
When not considering multi selection focused item is the only selected item.

WofWca:
Let's continue in https://github.com/deltachat/deltachat-desktop/issues/5386.

PR #5299, opened 2025-07-17
Mailinglist header should not be editable for recipients

nicodh:
commit ce7bbe5cd7fee shows the changes before renaming the file

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - after upgrade to `@deltachat/stdio-rpc-server` version `2.x` you can not go back to 1.x versions any more since account data is not backward compatible!
 
 ### Changed
+- update translations (17-07-2025)
 - upgrade `@deltachat/stdio-rpc-server` and `deltachat/jsonrpc-client` to `2.2.0`
   - API-Changes
     - jsonrpc: Add CommandApi::create_group_chat_unencrypted() (#6927).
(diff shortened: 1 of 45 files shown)


WofWca:
Especially that core supports it.

nicodh:
Well in  https://github.com/chatmail/core/issues/6884 which is much newer it says:
On the receiving side of a channel, check that it's not possible to change the name/avatar - neither by clicking on "Edit" in the menu, nor by enlarging the avatar and then clicking "Edit")

I discussed it with @Hocuri today and decided it's not the time now to open the UI discussion about the new channel feature. The edit feature can be easily brought back (that's why I made a separate PR for that)

packages/frontend/src/components/dialogs/MailingListProfile.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export default function MailingListProfile(
+  props: {
+    chat: T.BasicChat
+  } & DialogProps
+) {
+  const { onClose, chat } = props
+
+  const tx = useTranslationFunction()
+  const [groupImage] = useGroupImage(chat.profileImage)
```
WofWca: The hook can be removed, I'm pretty sure?

WofWca (approved):
OK, this MR seems to do the job then, but I didn't test it.

Issue #5059, opened 2025-04-30
Migrate to new PGP contacts API


adbenitez:
with https://github.com/chatmail/core/pull/6796 there will be a separation between PGP contacts (encrypted) and classic email-address-only contacts (no encryption key known)
 
- [x] show email icon for unencrypted messages, don't show padlock for encrypted (i.e. show an email icon iff `Message.show_padlock` returns false)
- [x] hide "add member" and "show QR invite" buttons for non-pgp group chats and disable/hidde option/button to remove members (available API: `FullChat.is_encrypted`)
- [x] hide "leave" for for unencryted group chats, EDIT: "clone" can stay as we get  "New E-Mail" (thanks @Hocuri )
- [x] do not allow to edit unencryted group chats names and avatars
- [x] do not allow to edit mailinglist names and avatars (might be already the case)
- [x] open "e2ee explained" dialog also on info message DC_INFO_CHAT_E2EE (in addition to DC_INFO_PROTECTION_ENABLED, cmp [android/#3822](https://github.com/deltachat/deltachat-android/pull/3822) ) - otherwise, DC_INFO_CHAT_E2EE does not require any special rendering or icon
- ~~display icon for non-pgp contacts in places where contact lists are displayed (available API: `ContactObject.is_pgp_contact`)~~
  - ~~main contact list / address book~~
  - ~~blocked contacts list~~
  - ~~group member list~~
  - ~~broadcast list / channel member list~~
  - ~~search results~~
- ~~display icon for non-pgp chats in chatlist (available API: `ChatListItem.is_encrypted`,~~ ~~not added yet at the time of writing this issue, check https://github.com/chatmail/core/pull/6796/files#r2068530190~~)
  - ~~also do it for chat items in the search results~~
- ~~display icon for non-pgp chat in opened conversation/chat title bar (available API: `FullChat.is_encrypted`)~~
- ~~display icon for non-pgp chat in the group/chat profile dialog (available API: `FullChat.is_encrypted`)~~
- ~~display icon for non-pgp contact in the contact profile dialog (available API: `ContactObject.is_pgp_contact`)~~

EDIT: updated the list

Hocuri:
[...]

PR #5302, opened 2025-07-18
Show email icon for unencrypted messages

nicodh:
task 1 in #5059


packages/frontend/src/components/message/MessageMetaData.tsx
```diff
@@ -50,20 +50,20 @@ export default function MessageMetaData(props: Props) {
...
           // We should not announce this for _every_ message.
           // This is available in the "Message info" dialog.
-          // In addition, if the message is not encerypted,
-          // we simply don't display the padlock,
-          // but arguably "not encrypted" is more important of a status
+          // In addition, if the message is encrypted,
+          // we simply don't display the email icon,
+          // since "not encrypted" is more important of a status
           // than "encrypted".
           aria-hidden={true}
```
WofWca, line 61: So, isn't this time to unmark it as hidden then? Or is this supposed to be implied by something else? iOS seems to not show it to screen readers?

_locales/_untranslated_en.json
```diff
@@ -106,5 +106,8 @@
   },
   "edit_channel": {
     "message": "Edit Channel"
+  },
+  "email": {
+    "message": "Email"
```
WofWca, line 111: How about adding "unencrypted"? I'm not sure though if this always means "unencrypted", e.g. if there is some external PGP stuff?

WofWca (commented):
I confirm that the word "padlock" is no longer present in the code base.
I also tested this. Seems to work.
Code-wise this is also good.

For reference: https://github.com/deltachat/deltachat-ios/pull/2749

So, the padlock icon is now unused, shall we remove it?

BTW I noticed that this icon is also shown in saved messages from unencrypted groups. Is it something that we want? Shall we ping core devs?

nicodh:
> So, the padlock icon is now unused, shall we remove it?

Yes, let's remove it.

Also added an issue for updating the help pages: https://github.com/deltachat/deltachat-pages/issues/1143

_locales/_untranslated_en.json
nicodh: Discussed that with @r10s and decided to stay compliant with the other apps

nicodh:
Added @r10s as reviewer since your other questions also affect the other apps

WofWca (approved):
OK, none of my points are blockers then, but let's not forget about them.

packages/frontend/src/components/message/MessageMetaData.tsx
r10s: did the combination of `aria-hidden` and `aria-label` ever made sense?
"hide the element from screenreader - and here is a label" :)

anyways, what else do we read out and what is the order? if we read out date and status well, we can also unconditionally read out "E-Mail".

if we want to have another wording, that should be applied on all OS, this is not a decision for desktop.

for the comment itself - not sure if that is helpful, i did not really got it, and these anyway turn out to be fast outdated, maybe just remove.

all no blockers, however.

WofWca: > did the combination of aria-hidden and aria-label ever made sense?

Well, behavior-wise it doesn't, but I think it's more "proper" to write it this way.

> anyways, what else do we read out and what is the order?

We read everything that appears on the screen in order (message text, then date, then "delivery status: delivered"), except the padlock (now envelope).

Issue #5294, opened 2025-07-16
add "New E-Mail" option to create and send e-mail


r10s:
this feature shall go into version 2.0 and will allow users to create email within Delta Chat, resulting in the same chat types as incoming ad-hoc-group creation. it is mainly about using some different strings:

- [x] add "New E-Mail" (string `new_email`) option to the "New Chat" dialog, below "New Channel"

- [x] when tapped, open the "create group dialog" with the following small tweaks:
    - [x] set title to  "New E-Mail" (string `new_email`) 
    - [x] set group name to "Subject (string `subject`)
    - [x] hide option to set avatar
    - [x] change to "recipient" wording (as for channels, strings  `n_recipients`, `add_recipients`)
    - [x] rename ok-button to "Continue" (string `perm_continue`)

- [x] when "Continue" is pressed, create the chat with jsonrpc's `create_group_chat_unencrypted(accountId, subject)` and open the returned chat id, where the user can proceed as usual

- [ ] remove "Add Contact Manually" from the "New Chat" dialog. double check, the option exist at "New Chat / New Contact", "QR Code" and "New Chat / New E-Mail / Add Recipients"

the chat itself picks up "group wording" again, we may want to change that, see discussions linked below, but that is out of scope of this issue. moreover, there are lots of discussions about wording etc. in other channels, i would skip that here :)

to get an idea, **Screenshots** are available at https://github.com/deltachat/deltachat-ios/pull/2770, other related PR are https://github.com/deltachat/deltachat-android/pull/3815 and https://github.com/chatmail/core/pull/6932

WofWca:
How is "Clone Chat" supposed to behave? Are we supposed to look at `chatType` to see if it's a `MAILINGLIST` or `GROUP`?

WofWca:
Do we hide this option if `isChatmail`, as in https://github.com/deltachat/deltachat-ios/pull/2770?

WofWca:
> remove "Add Contact Manually" from the "New Chat" dialog. double check, the option exist at "New Chat / New Contact", "QR Code" and "New Chat / New E-Mail / Add Recipients"

Desktop does this differently as of now. It only shows the "Add contact manually" option at the bottom, only when the search query is not empty.

<img width="412" height="204" alt="Image" src="https://github.com/user-attachments/assets/ba1fecba-eb28-47d8-a36d-ddb72da62db7" />

[...]

PR #5303, opened 2025-07-18
feat: add "New E-Mail" dialog (unencrypted email)

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/5294.


packages/frontend/src/components/dialogs/ViewGroup.tsx
```diff
@@ -428,7 +435,7 @@ function ViewGroupInner(
                   <>
                     <PseudoListItemAddMember
                       onClick={() => showAddMemberDialog()}
-                      isBroadcast={isBroadcast}
+                      labelMembersOrRecipients={membersOrRecipients}
                     />
                     {!isBroadcast && (
                       <PseudoListItemShowQrCode
```
nicodh, line 441: QrCode should not be shown here. Based on the condition !chat.isEncrypted
If you click on the QR code you get an error from core:
"Can't generate SecureJoin QR code for ad-hoc group "

packages/frontend/src/components/dialogs/CreateChat/index.tsx
```diff
@@ -158,6 +195,12 @@ function CreateChatMain(props: CreateChatMainProps) {
     queryStr.length === 0 &&
     (settingsStore?.desktopSettings.enableBroadcastLists ?? false)
   const needToRenderAddContactQRScan = queryStr.length === 0
+
+  // Chatmail accounts can't send unencrypted emails. See
+  // - https://github.com/deltachat/deltachat-desktop/issues/5294#issuecomment-3089552788
+  // - https://github.com/deltachat/deltachat-ios/blob/a0043be425d9c14f4039561957adb82ef1ab2adb/deltachat-ios/Controller/NewChatViewController.swift#L76-L78
+  const needToRenderNewEmail = !isChatmail && queryStr.length === 0
+
   const needToRenderAddContact = !(
```
nicodh: I would prefer the wording showAddContact or showNewContact instead "needToRender" (although the wording is not invented in this PR)

And it would be consistent with
showPseudoListItemAddContactFromInviteLink

packages/frontend/src/components/dialogs/ViewGroup.tsx
```diff
@@ -309,14 +309,21 @@ function ViewGroupInner(
...
   }
 
+  // Note that this might also need `C.DC_GCL_ADDRESS` for unencrypted groups,
+  // but we're not supposed to display this component for those.
   const listFlags = C.DC_GCL_ADD_SELF
 
+  // Note that we are not checking `chat.isEncrypted`,
+  // unlike in "New E-Mail" dialog.
+  // See https://github.com/deltachat/deltachat-desktop/issues/5294
+  // > the chat itself picks up "group wording"
```
nicodh, line 319: Are you sure this is meant by "group wording"? I couldn't find the "discussion" about that?
It's a bit strange to handle the wording for adding members different on creating and on editing the unencrypted group. Maybe to somehow differentiate to the read only recipients in a mailinglist?

nicodh (changes requested):
I think it would make sense to add the related tasks from #5059 also to this PR. 

Maybe that's why you still kept the AddContactQRCode ?

packages/frontend/src/components/dialogs/ViewGroup.tsx
WofWca: That's how I interpreted it at least, I'm not 100% sure.

packages/frontend/src/components/dialogs/CreateChat/index.tsx
WofWca: OK, I'll add a rename commit.

WofWca (commented):
> Maybe that's why you still kept the AddContactQRCode ?

Well yes, I only addressed the points from the issue, i.e. add the "New E-Mail" dialog, and preserved everything else.

The same goes for the comment https://github.com/deltachat/deltachat-desktop/pull/5303/files#r2217174958.

packages/frontend/src/components/dialogs/CreateChat/index.tsx
WofWca: Done.

nicodh:
> I think it would make sense to add the related tasks from #5059 also to this PR.

I can also add the tasks (they were in "my" issue finally) :-)

WofWca:
Well if you have other things to do, I can reassign the task to myself.

PR #5305, opened 2025-07-19
Remove unused images

nicodh:
Needs some more testing maybe.  After removing the padlock icon I started looking for more unused images/icons.
Many of these were added long ago...

#skip-changelog


nicodh:
Not necessary needed for next release. There are enough changes already...

WofWca (commented):
Some of these were made in Inkscape, according to `git show origin/remove-unused-images -S"inksc"`, which might mean that they are the "original" images, and not just copy pasted ones. Maybe they shouldn't be removed?

nicodh:
> Some of these were made in Inkscape, according to `git show origin/remove-unused-images -S"inksc"`, which might mean that they are the "original" images, and not just copy pasted ones. Maybe they shouldn't be removed?

I don't think so - at least they should not be shipped with the app if they are not used.

WofWca (approved):
OK then.
I verified that all of these are unused, by searching for file names from `git show stat`.

nicodh:
I found 3 icons that were used dynamically and removed unused icon names.

Will wait a bit more before merging

Simon-Laux (approved):

PR #5311, opened 2025-07-20
test: add some E2E tests for unencrypted groups

WofWca:
#skip-changelog

### DIFF
--- a/packages/frontend/src/components/dialogs/CreateChat/index.tsx
+++ b/packages/frontend/src/components/dialogs/CreateChat/index.tsx
@@ -90,7 +90,7 @@ export default function CreateChat(props: DialogProps) {
   const [viewMode, setViewMode] = useState<ViewMode>('main_')
 
   return (
-    <Dialog width={400} onClose={onClose} fixed>
+    <Dialog width={400} onClose={onClose} fixed dataTestid='create-chat-dialog'>
       {viewMode == 'main_' && <CreateChatMain {...{ setViewMode, onClose }} />}
       {viewMode == GroupType.REGULAR_GROUP && (
         <>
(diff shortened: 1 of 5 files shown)


WofWca:
The tests are failing because of https://github.com/deltachat/deltachat-desktop/pull/5312. This will need a rebase on top of it.
The tests are green locally with that commit cherry-picked, so this can be reviewed.

packages/frontend/src/components/dialogs/CreateChat/index.tsx
```diff
@@ -199,7 +199,16 @@ function CreateChatMain(props: CreateChatMainProps) {
   // Chatmail accounts can't send unencrypted emails. See
   // - https://github.com/deltachat/deltachat-desktop/issues/5294#issuecomment-3089552788
   // - https://github.com/deltachat/deltachat-ios/blob/a0043be425d9c14f4039561957adb82ef1ab2adb/deltachat-ios/Controller/NewChatViewController.swift#L76-L78
-  const showNewEmail = !isChatmail && queryStr.length === 0
+  const [forceShowNewEmail_, setForceShowNewEmail_] = useState(false)
```
nicodh, line 202: I think we should not add functions in the production code just for testing. Only data-test attributes if possible.

We have now secrets for creating real non chatmail accounts, so maybe better use these

nicodh (changes requested):

packages/frontend/src/components/dialogs/CreateChat/index.tsx
WofWca: It's just a simple toggle to hide or show a button, which is clearly defined.
I mean, in a perfect world we wouldn't need it, but don't think it's that bad.
Otherwise if the server goes down, or if you don't know the secret (e.g. run tests locally), the test will not work.

I suggest to keep this test and add another one, maybe reuse the code.

WofWca:
It would be nice to test https://github.com/deltachat/deltachat-desktop/pull/5371 and https://github.com/deltachat/deltachat-desktop/pull/5373 as well.

nicodh (approved):
Since we decided to remove the workaround later when it is not needed anymore let's merge this now and improve later.

PR #5318, opened 2025-07-21
Remove green checkmarks

nicodh:
see https://github.com/deltachat/deltachat-ios/pull/2781


packages/frontend/src/components/message/Message.tsx
```diff
@@ -663,21 +663,12 @@ export default function Message(props: {
...
           'info-message',
           isWebxdcInfo && 'webxdc-info',
-          isInteractive && 'interactive'
+          isInteractive && 'interactive',
+          isProtectionEnabledMsg && 'e2ee-info' // used in e2e-tests
         )}
         id={String(message.id)}
         onContextMenu={showContextMenu}
       >
-        {(isProtectionBrokenMsg || isProtectionEnabledMsg) && (
```
WofWca, line 671: Are we also supposed to remove `isProtectionBrokenMsg` icon?

If so, then the CSS and the icon file need to be removed as well I think.

packages/frontend/src/components/ProfileInfoHeader/index.tsx
```diff
@@ -35,9 +33,6 @@ export default function ProfileInfoHeader({
       <div className={styles.infoContainer}>
         <p className={styles.displayName} data-test-id='profile-display-name'>
           {displayName}
-          <span className={styles.verifiedIconWrapper}>
```
WofWca, line 38: The respective class needs to be removed as well I suppose.

packages/frontend/src/components/ProfileInfoHeader/index.tsx
```diff
@@ -18,7 +17,6 @@ export default function ProfileInfoHeader({
   avatarPath,
   color,
   displayName,
-  isVerified,
```
WofWca, line 21: I suggest to also remove it from the `Props` type.

packages/frontend/src/components/message/Message.tsx
r10s: `isProtectionBrokenMsg` is never true for new messages, so icon, css, whatnot can be removed.

(for historic broken-messages, the text stays as is and is sufficient, no need to make them nice or outstanding any longer)

r10s:
what's missing is to change opening the local help at `#e2ee`  instead of `#e2eeguarantee`  (the anchor will be there after local help is updated)

nicodh:
> what's missing is to change opening the local help at `#e2ee` instead of `#e2eeguarantee` (the anchor will be there after local help is updated)

Yes - that will come together with the help pages update

r10s:
i did the help update at https://github.com/deltachat/deltachat-desktop/pull/5324 - feel free to modify the PR (or close it if you want to have it somewhere else)

WofWca (approved):

PR #5320, opened 2025-07-21
Upgrade to core 2.4.0

nicodh:


WofWca (approved):
Just an upgrade

I see nothing in the changelog that requires us to make updates https://github.com/chatmail/core/blob/main/CHANGELOG.md.

PR #5324, opened 2025-07-22
update local help

r10s:
ran `./bin/help/create-local-help.sh`, assuming this is still the way to go, cc @nicodh 

an help update is required for the new anchor `#e2ee` needed for https://github.com/deltachat/deltachat-desktop/pull/5318 and for the other huge differences wrt encryption

### DIFF
--- a/packages/frontend/src/components/dialogs/ProtectionStatusDialog.tsx
+++ b/packages/frontend/src/components/dialogs/ProtectionStatusDialog.tsx
@@ -17,7 +17,7 @@ import useDialog from '../../hooks/dialog/useDialog'
 import type { DialogProps } from '../../contexts/DialogContext'
 
 const VERIFICATION_BROKEN_ANCHOR = 'nocryptanymore'
-const VERIFICATION_ENABLED_ANCHOR = 'e2eeguarantee'
+const VERIFICATION_ENABLED_ANCHOR = 'e2ee'
 const VERIFICATION_REQUIRED_ANCHOR = 'howtoe2ee'
 
 export function ProtectionBrokenDialog({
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/dialogs/ProtectionStatusDialog.tsx
```diff
@@ -17,7 +17,7 @@ import useDialog from '../../hooks/dialog/useDialog'
 import type { DialogProps } from '../../contexts/DialogContext'
 
 const VERIFICATION_BROKEN_ANCHOR = 'nocryptanymore'
-const VERIFICATION_ENABLED_ANCHOR = 'e2eeguarantee'
+const VERIFICATION_ENABLED_ANCHOR = 'e2ee'
```
r10s, line 20: for maintenance, i'd also adapt the name of the variable to eg. E2EE_ANCHOR, it is not only about verification (the anchor is opened when [DC_INFO_CHAT_E2EE or DC_INFO_PROTECTION_ENABLED](https://c.delta.chat/classdc__msg__t.html#ab3eeb59143e67fda31d3648bd775f77b) are tapped - the former is added when the chat is e2ee but _not_ verified (whatever that means in detail :)

VERIFICATION_BROKEN_ANCHOR can be removed alltogether, but that cleanup is probably done better in another PR

WofWca (approved):
I didn't write that generator script, and didn't look through the help files but the MR looks sane.

BTW it would be nice to stop indexing the "search" files. As well as help and translation files, maybe with the help of git submodules.

r10s:
@nicodh i leave merging up to you :)

PR #5328, opened 2025-07-23
Change verified by handling

### LINKED ISSUE #5327: Handle contact.isVerified if contact.verifiedBy = null
If a contact isVerified but verifiedBy = null we should show "Introduced" (verified_by_unknown) in the Profile

see https://github.com/chatmail/core/pull/7027/files

nicodh:
If verifierId === null and isVerified === true
display "Introduced" prefixed by a green checkmark.

resolves #5327
see https://github.com/chatmail/core/pull/7027/files#diff-7d5e9c113766adfcca5e5b842258ba90cab966d995b7df97eb0e1cafbd18781cR5294


packages/frontend/src/components/dialogs/ViewProfile/index.tsx
```diff
@@ -181,28 +181,33 @@ export function ViewProfileInner({
...
+              label: tx('verified_by', displayName),
+              action: () => openViewProfileDialog(accountId, verifierContactId),
+            })
+          }
         } catch (error) {
           log.error('failed to load verifier contact', error)
           setVerifier({
             label:
               'verified by: failed to load verifier contact, please report this issue',
-            action: () => openViewProfileDialog(accountId, verifierContactId),
```
WofWca, line 205: It doesn't seem like we must remove the action?

packages/frontend/src/components/dialogs/ViewProfile/index.tsx
```diff
@@ -181,28 +181,33 @@ export function ViewProfileInner({
...
-      if (contact.verifierId === null) {
-        setVerifier(null)
-      } else if (contact.verifierId === C.DC_CONTACT_ID_SELF) {
-        setVerifier({ label: tx('verified_by_you') })
+      if (contact.isVerified) {
+        // it might happen that a verified contact has no verifiedBy ID
+        setVerifier({ label: tx('verified_by_unknown') })
       } else {
-        setVerifier(null) // make sure it rather shows nothing than wrong values
+        setVerifier(null) // will be overridden if verifiedBy ID is available
```
WofWca, line 188:
```suggestion
        // will be overridden if verifiedBy ID is available,
        // although that should never happen as of writing.
        setVerifier(null)
```

packages/frontend/src/components/dialogs/ViewProfile/index.tsx
nicodh: But what should haoppen if you open the profile for an ID that backend couldn't load?

packages/frontend/src/components/dialogs/ViewProfile/index.tsx
nicodh: I would say the default is, that there is an verifiedBy ID

packages/frontend/src/components/dialogs/ViewProfile/index.tsx
WofWca: IDK, maybe it didn't load just this time. Why not try to load it another time.

packages/frontend/src/components/dialogs/ViewProfile/index.tsx
WofWca: ~~I mean it's impossible that `isVerified === true && verifierId == null`.~~

https://github.com/chatmail/core/blob/0359481ba487d7eea7747a4c401dfdd97423e68a/src/contact.rs#L1587-L1595

WofWca: > I mean it's impossible that isVerified === true && verifierId == null.

Whoops, I meant `isVerified === false && verifierId != null`.

nicodh: True - so yes we could add that comment line

PR #5329, opened 2025-07-23
Use git url for stylus

nicodh:
Hopefully just as temporaray workaround

#skip-changelog


nicodh:
I didn't really investigate if we need stylus, just see that everything compiles as expected. Seems enough to me, since it's a dependency of typescript-css-modules so it should only affect the build process

PR #5330, opened 2025-07-23
upgrade core to 2.6.0

nicodh:


WofWca (approved):
Just an upgrade.
Fixes a crash, as far as I understand.

PR #5333, opened 2025-07-23
Preview release 2.6.0

nicodh:
Preview for the next release

#skip-changelog
#public-preview

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1892,7 +1892,7 @@ dependencies = [
 
 [[package]]
 name = "deltachat-tauri"
-version = "1.60.1"
+version = "2.6.0"
 dependencies = [
  "anyhow",
  "ashpd",
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "private": "true",
   "name": "deltachat-desktop",
   "type": "module",
-  "version": "1.60.1",
+  "version": "2.6.0",
   "scripts": {
     "preinstall": "node ./bin/check-nodejs-version.js",
     "check": "pnpm check:types && pnpm check:lint && pnpm check:format && pnpm check:target-versions && pnpm check:log-conventions",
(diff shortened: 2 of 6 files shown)


.github/workflows/build-preview.yml
```diff
@@ -112,7 +112,7 @@ jobs:
       if: steps.upload.outcome == 'success'
       run: node ./bin/github-actions/postLinksToDetails.js
       env:
-       PR_ID: ${{ steps.prepare.outputs.prid }}
+       PR_ID: ${{ steps.prepare.outputs.branch }}
```
WofWca: The variable name also needs to be updated, it seems.

.github/workflows/build-preview.yml
```diff
@@ -112,7 +112,7 @@ jobs:
       if: steps.upload.outcome == 'success'
       run: node ./bin/github-actions/postLinksToDetails.js
       env:
```
WofWca: This script now adds links that are 404:
https://download.delta.chat/desktop/preview/deltachat-desktop-preview-release-2.6.0.AppImage

WofWca (changes requested):

.github/workflows/build-preview.yml
WofWca: That's due to the changes from dashes to dots.

nicodh: Hmm, and those are coming from actions/upload-artifact@main which we don't control...

if the script uses base_url it should be fixed now

WofWca: It takes either `FULL_ARTIFACT_URL` or `PR_BRANCH`. Works now.

nicodh:
> Let's release. 

As long as we stay in the testing phase, this PR should not be merged (since that would remove the preview builds).

So this PR will stay till we do the final release or a new version is needed. (which is probably the case when core releases  2.7.0 before our final release )

PR #5342, opened 2025-07-26
Test: Create non chatmail account

nicodh:
#skip-changelog

read mail server urls from secrets

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -83,6 +83,8 @@ jobs:
       NODE_ENV: test
       VERSION_INFO_GIT_REF: ${{ github.ref }}
       DC_ACCOUNTS_DIR: ../../e2e-tests/data/accounts
+      DC_CHATMAIL_SERVER: ${{ secrets.CI_CHATMAIL_SERVER }}
+      DC_MAIL_SERVER_URL: ${{ secrets.CI_MAIL_SERVER }}
     timeout-minutes: 60
     runs-on: ${{ matrix.os }}
     strategy:
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -84,7 +84,8 @@ jobs:
       VERSION_INFO_GIT_REF: ${{ github.ref }}
       DC_ACCOUNTS_DIR: ../../e2e-tests/data/accounts
       DC_CHATMAIL_SERVER: ${{ secrets.CI_CHATMAIL_SERVER }}
-      DC_MAIL_SERVER_URL: ${{ secrets.CI_MAIL_SERVER }}
+      DC_MAIL_SERVER: ${{ secrets.CI_MAIL_SERVER_URL }}
+      DC_MAIL_SERVER_TOKEN: ${{ secrets.CI_MAIL_SERVER_TOKEN }}
     timeout-minutes: 60
     runs-on: ${{ matrix.os }}
     strategy:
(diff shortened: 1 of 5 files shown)


.github/workflows/e2e.yml
```diff
@@ -83,6 +83,8 @@ jobs:
       NODE_ENV: test
       VERSION_INFO_GIT_REF: ${{ github.ref }}
       DC_ACCOUNTS_DIR: ../../e2e-tests/data/accounts
+      DC_CHATMAIL_SERVER: ${{ secrets.CI_CHATMAIL_SERVER }}
+      DC_MAIL_SERVER_URL: ${{ secrets.CI_MAIL_SERVER }}
```
WofWca: I think it's best to have just the token itself as the secret, and then assemble the complete URL by string formatting. Better ensures that it won't be leaked. See https://docs.github.com/en/actions/reference/security/secure-use#writing-workflows

> Structured data can cause secret redaction within logs to fail, because redaction largely relies on finding an exact match for the specific secret value.

> If a secret is used to generate another sensitive value within a workflow, that generated value should be formally [registered as a secret](https://github.com/actions/toolkit/tree/main/packages/core#setting-a-secret), so that it will be redacted if it ever appears in the logs. For example, if using a private key to generate a signed JWT to access a web API, be sure to register that JWT as a secret or else it won’t be redacted if it ever enters the log output.

nicodh: I agree. Changed it. The inconsistent naming of the secrets will be fixed when this is merged. MAIL_SERVER, MAIL_SERVER_URL etc.

.github/workflows/e2e.yml
```diff
@@ -83,6 +83,9 @@ jobs:
       NODE_ENV: test
       VERSION_INFO_GIT_REF: ${{ github.ref }}
       DC_ACCOUNTS_DIR: ../../e2e-tests/data/accounts
+      DC_CHATMAIL_SERVER: ${{ secrets.CI_CHATMAIL_SERVER }}
```
WofWca: Sooo, are we supposed to hide the CI Chatmail server from others? They can't run tests locally then.

Also, I believe we can't really hide this (and the non-Chatmail server), because it's exposed in the UI, namely in "account and password" settings anyway as tests run. If a test fails, the Playwright report might contain a video where this can be seen.
And the email also exposed in the test artifact output. And in the test logs (I am not sure why it didn't get auto-redacted).

packages/e2e-tests/tests/plain-email-tests.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+  const userB = existingProfiles[1]
+  const subject = 'Test Chat Subject'
+  const emailUserB = userB.address
+  // prepare last open chat for receiving user
+  await switchToProfile(page, userA.id)
+  await page.locator('#new-chat-button').click()
+
+  await expect(page.getByRole('button', { name: 'New Group' })).toBeVisible()
+
+  // Since we're on a Chatmail server, this button is supposed to be shown.
```
WofWca:
```suggestion
  // Since we're on a non-Chatmail server, this button is supposed to be shown.
```

WofWca (commented):
The test itself looks OK. But it doesn't actually send emails.

I left a comment about secrets.

.github/workflows/e2e.yml
nicodh: Well for the server URLs the idea is more to make them configurable without touching the code/tests. Only the token is a "real" secret.

But you are right, the server names could just be variables then.

nicodh:
> The test itself looks OK. But it doesn't actually send emails.

Yes, that will be done in follow ups. This is mainly about having a non chatmail server in the tests and create accounts which are not keyAccounts

nicodh:
> Moving from "secrets" to "variables" can be done later.

Already done

PR #5344, opened 2025-07-27
Improve outline visibility (a bit)

nicodh:
When navigating with the keyboard in light mode it seems sometimes the focus is "lost" since the contrast of the outline is too small. This PR improves that for:
- selected chat item
- create chat button
- Info messages

Before:

https://github.com/user-attachments/assets/04054f97-8465-48dc-b12f-2188ad3305c3

After:

https://github.com/user-attachments/assets/d9eefe8a-83ad-44fa-b8d0-180a3151d59f

#skip-changelog

### DIFF
--- a/packages/frontend/scss/chat/_chat-list-item.scss
+++ b/packages/frontend/scss/chat/_chat-list-item.scss
@@ -16,6 +16,7 @@
   flex-direction: row;
   padding: 0px 10px;
   align-items: center;
+  outline-offset: -1px;
 
   .status-icon {
     flex-shrink: 0;
@@ -360,4 +361,9 @@
 #new-chat-button {
   right: 15px;
   position: absolute;
+  outline-offset: 2px;
+  &:focus {
+    height: 52px;
+    width: 52px;
+  }
 }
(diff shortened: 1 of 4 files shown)


packages/frontend/scss/chat/_chat-list-item.scss
```diff
@@ -360,4 +361,9 @@
 #new-chat-button {
   right: 15px;
   position: absolute;
+  outline-offset: 2px;
+  &:focus {
+    height: 52px;
+    width: 52px;
```
WofWca: Ummm IDK about this. This shifts the button to the left a bit.

Is this supposed to be about visibility or just a fun little thing?

WofWca (approved):
Yep, the chat list items needed that.

packages/frontend/scss/chat/_chat-list-item.scss
nicodh: For visibility. The small move should make it more aware...

WofWca: Well, outline should be enough I think.

nicodh: Well the focus jumps quite some distance here and the movement should make that visible.

At least for me when using keyboard navigation (what I do mostly) there are two jumps that are hard to follow:
From seletected chat to create chat button and from first message to the attachment in composer input.

WofWca: Alright then. Also consider `scale`.

PR #5347, opened 2025-07-28
improvement: a11y: add `aria-label` for some menus

WofWca:
https://www.w3.org/WAI/ARIA/apg/patterns/menubar/ :
> An element with role menu either has:
> - aria-labelledby set to a value that refers to the menuitem
>   or button that controls its display.
> - A label provided by aria-label.


nicodh (approved):
Reviewed by reading & testing

Issue #5337, opened 2025-07-25
unreliable UX when replying to info messages


r10s:
in contrast to android and iOS, desktop allows to reply to info or system messages (webxdc, member added, e2ee-enabled etc)

as these messages are partly generated locally, "jumping up" on quoting often does not work. also, the message may not exist for the other users or devices, leading to questions. might even be a privacy issue eg. if a locally set name is quoted - or an info message contains private data otherwise 

also, it seems not to be a super-important thing and is comparable rarely used, so that our resources are probably spend better elsewhere.

moreover, the different behaviours on android/iOS leads to support effort (i just got a request again privately, this is not the first time this pops up)

as the gist of "locally generated" is not about to change, this is not really fixable. we should just not allow to reply to these messages.

adbenitez:
+1 at least not until quoting properly works otherwise it feels more like a half-cooked buggy feature, and good point about the privacy concerns!!! this could leak local set custom names!?

WofWca:
We have at least two places that do quoting: Ctrl + Up shortcut and the message context menu.

r10s:
> Ctrl + Up shortcut and the message context menu.

message context menu is easy, the options should just be removed there.

for the shortcut: either skip the info-messages or allow marking them, but do not set a quote - whatever is easier

WofWca:
> might even be a privacy issue eg. if a locally set name is quoted

That might be something for core as well then.

***

https://github.com/deltachat/deltachat-desktop/pull/5348 addresses this for menu items, but the Ctrl + ArrowUp is still remaining. That might be solved larger refactor, i.e. making the message list available to the shortcut handling code, which would also improve performance.

https://github.com/deltachat/deltachat-desktop/blob/b352801fb8d2e816e364aadced44235299a6baa0/packages/frontend/src/components/composer/Composer.tsx#L928-L997

saper:
Where can I learn about those `locally generated` things? Maybe they should be treated as different animals altogether and not regular messages?

WofWca:
> Where can I learn about those locally generated things?

Most if not all of those are messages with `is_info` set to `true`. See the https://github.com/chatmail/core repo.

r10s:
> Maybe they should be treated as different animals altogether and not regular messages?

this is the case, these messages are shown totally different; not even as bubbles

nicodh:
Resolved with #5471

PR #5348, opened 2025-07-28
fix: hide "Reply" and "Save" for info and call msgs

WofWca:
Partially addresses
https://github.com/deltachat/deltachat-desktop/issues/5337.

It is still possible to reply to info messages
with the Ctrl + ArrowUp shortcut.

### DIFF
--- a/packages/e2e-tests/tests/basic-tests.spec.ts
+++ b/packages/e2e-tests/tests/basic-tests.spec.ts
@@ -151,6 +151,47 @@ test('send message', async ({ page }) => {
   await expect(receivedMessageText).toHaveText(messageText)
 })
 
+test('message menu items presence', async ({ page }) => {
+  const userA = existingProfiles[0]
+  const userB = existingProfiles[1]
+  await switchToProfile(page, userA.id)
+  await page
+    .locator('.chat-list .chat-list-item')
+    .filter({ hasText: userB.name })
+    .click()
+
+  const someRegularOutgoingMessage = page
+    .getByLabel('Messages')
+    .getByText('Hello')
+    .first()
+  await someRegularOutgoingMessage.click({ button: 'right' })
+  await expect(page.getByRole('menu').getByRole('menuitem')).toHaveText([
+    'Reply',
+    'Forward',
+    'React',
+    'Edit',
+    'Save',
+    'Copy Text',
+    'Resend',
+    'Info',
+    'Delete Message',
+  ])
+  await page.keyboard.press('Escape')
+
+  const someInfoMessage = page
+    .getByLabel('Messages')
+    .getByText('Messages are end-to-end encrypted')
+    .first()
+  await someInfoMessage.click({ button: 'right' })
+  await expect(page.getByRole('menu').getByRole('menuitem')).toHaveText([
+    'Forward',
+    'Copy Text',
+    'Info',
+    'Delete Message',
+  ])
+  await page.keyboard.press('Escape')
+})
+
 /**
  * user A deletes one message for himself
  */
--- a/packages/frontend/src/components/message/Message.tsx
+++ b/packages/frontend/src/components/message/Message.tsx
@@ -274,8 +274,13 @@ function buildContextMenu(
   const showCopyImage = !!message.file && message.viewType === 'Image'
   const showResend = message.sender.id === C.DC_CONTACT_ID_SELF
 
-  // Do not show "reply" in read-only chats
-  const showReply = chat.canSend
+  const isInfoOrCallInvitation =
+    message.isInfo || message.viewType === 'VideochatInvitation'
+  // Do not show "reply" in read-only chats, and for info messages.
+  // See
+  // - https://github.com/deltachat/deltachat-desktop/issues/5337
+  // - https://github.com/deltachat/deltachat-android/blob/52c01976821803fa2d8a177f93576fa4082ef5bd/src/main/java/org/thoughtcrime/securesms/ConversationFragment.java#L332-L332
+  const showReply = chat.canSend && !isInfoOrCallInvitation
 
   // See
   // - https://github.com/deltachat/deltachat-desktop/issues/4695.
@@ -286,8 +291,7 @@ function buildContextMenu(
     chat.isEncrypted &&
     message.text !== '' &&
     chat.canSend &&
-    !message.isInfo &&
-    message.viewType !== 'VideochatInvitation' &&
+    !isInfoOrCallInvitation &&
     !message.hasHtml
 
   // Do not show "react" for system messages
@@ -297,6 +301,7 @@ function buildContextMenu(
   const showReplyPrivately =
     (conversationType.chatType === C.DC_CHAT_TYPE_GROUP ||
       conversationType.chatType === C.DC_CHAT_TYPE_IN_BROADCAST) &&
+    !isInfoOrCallInvitation &&
     message.fromId > C.DC_CONTACT_ID_LAST_SPECIAL
 
   return [
@@ -330,8 +335,11 @@ function buildContextMenu(
     },
     { type: 'separator' },
     // Save Message
+    // For reference, the conditions when it's shown:
+    // https://github.com/deltachat/deltachat-android/blob/52c01976821803fa2d8a177f93576fa4082ef5bd/src/main/java/org/thoughtcrime/securesms/ConversationFragment.java#L342
     !chat.isSelfTalk &&
-      !isSavedMessage && {
+      !isSavedMessage &&
+      !isInfoOrCallInvitation && {
         label: tx('save'),
         action: () =>
           BackendRemote.rpc.saveMsgs(selectedAccountId(), [message.id]),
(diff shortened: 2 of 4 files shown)


WofWca:
I'll fix the test, it's a minor thing.
Edit: fixed.

packages/e2e-tests/tests/basic-tests.spec.ts
```diff
@@ -151,6 +151,47 @@ test('send message', async ({ page }) => {
...
+  ])
+  await page.keyboard.press('Escape')
+
+  const someInfoMessage = page
+    .getByLabel('Messages')
+    .getByText('Messages are end-to-end encrypted')
+    .first()
+  await someInfoMessage.click({ button: 'right' })
+  await expect(page.getByRole('menu').getByRole('menuitem')).toHaveText([
+    'Forward',
```
Simon-Laux, line 187: what about forwarding info messages @r10s, maybe we should also hide that option?

Simon-Laux: @WofWca shouldn't this test also check that the unwanted items are not there, or that the menu looks as expected?

WofWca: > shouldn't this test also check that the unwanted items are not there

It does:

> If you pass an array as an expected value, the expectations are:
>
> 1. Locator resolves to a list of elements.
> 2. The number of elements equals the number of expected values in the array.
> 3. Elements from the list have text matching expected array values, one by one, in order.

> or that the menu looks as expected?

You mean screenshot tests? I haven't gotten around to checking out how to implement them. We probably want some in the end I'd say, but `toHaveText` is pretty reliable anyway IMO.

r10s: > what about forwarding info messages @r10s, maybe we should also hide that option?

changing that seems out of scope of the issue.

 forwarding results in new message-id it does not the issues initially discussed. the text is forwarded as is, this seems expected as well.

also, forwarding is allowed on android/iOS as well, for consistency, let's stay with that on desktop as well.

r10s:
thanks a lot for removing "save" as well, which has the same issues as discussed with "reply".

packages/frontend/src/components/message/Message.tsx
```diff
@@ -277,8 +277,12 @@ function buildContextMenu(
   const showCopyImage = !!message.file && message.viewType === 'Image'
   const showResend = message.sender.id === C.DC_CONTACT_ID_SELF
 
-  // Do not show "reply" in read-only chats
-  const showReply = chat.canSend
+  // Do not show "reply" in read-only chats, and for info messages.
+  // See
+  // - https://github.com/deltachat/deltachat-desktop/issues/5337
+  // - https://github.com/deltachat/deltachat-android/blob/52c01976821803fa2d8a177f93576fa4082ef5bd/src/main/java/org/thoughtcrime/securesms/ConversationFragment.java#L332C21-L332C30
```
nicodh: I don't think it makes sense to add all these links to the source code, the implementation may change over time, but these links stay until someone has the idea of removing or updating them. They are very helpful for reviewing though, so better add them as comment in the PR?

Wen looking at the implementation which is refrenced there I see that Android does not allow to reply to VideoChat Invitations but desktop does. So it would make sense (at least for consistency) to add message.viewType != VideochatInvitation to the showReply conditions

packages/frontend/src/components/message/Message.tsx
```diff
@@ -296,6 +300,7 @@ function buildContextMenu(
   const showReplyPrivately =
     (conversationType.chatType === C.DC_CHAT_TYPE_GROUP ||
```
nicodh, line 302: Same here

nicodh (changes requested):

packages/frontend/src/components/message/Message.tsx
WofWca: > the implementation may change over time

It's not hard to switch to latest version of the file:

image

WofWca: > Wen looking at the implementation which is refrenced there I see that Android does not allow to reply to VideoChat Invitations but desktop does.

Oh. Yep, video chats suffer from the same "locally-generated" problem. Although jumping to message seems to work on another device, unlike with info messages.

nicodh: > > the implementation may change over time
> 
> It's not hard to switch to latest version of the file:
> 
Shure but do we want these links in our source code ? It is easy to jump to this PR  and there one would find the reference in the comment

How should be decided which reference links should make it in in our source code and which not? It could kind of "pollute" our code sooner or later.

WofWca: > It is easy to jump to this PR and there one would find the reference in the comment

Why jump to this PR when you can jump right to the reference implementation, with no intermediaries.
One can easily follow the link and re-check with the latest version of the Android implementation.
I would make it a non-perma link, but the only reason not to do that is because the referenced code could move somewhere else.

> It could kind of "pollute" our code sooner or later.

The link is relevant to the code itself, and not just this MR, therefore it's in the code and not in the MR.

packages/frontend/src/components/message/Message.tsx
WofWca: I addressed it, see the force-push.

packages/frontend/src/components/message/Message.tsx
nicodh: Ok, then can we maybe agree that one link is enough? The one that points to the function usage OR the to the function itself

WofWca: In general I think the latest version of the code should be self-sufficient. git commit history should not be a thing that is necessary to understand the code.

WofWca: > Ok, then can we maybe agree that one link is enough?

OK, I'll remove one.

packages/e2e-tests/tests/basic-tests.spec.ts
Simon-Laux: > You mean screenshot tests?

no. as far as I understand the test code it checks if the expected entries are there, but it does not check that the unexpected items are not there, does it?

packages/frontend/src/components/message/Message.tsx
Simon-Laux: > Oh. Yep, video chats suffer from the same "locally-generated" problem

really? I thought they are normal messages and just look like info messages?

WofWca: Well what I mean is that if you reply to a video chat invitation message, the quote text depends on the locale.
If you send an invitation and reply to it from one device, on another device with another locale it will be displayed differently.
Edit: I mean on the second device the quote text will be displayed in the locale of the first message.

https://github.com/deltachat/deltachat-desktop/blob/dd857469f343053271f39f0f0befbb010398be15/packages/frontend/src/stockStrings.ts#L47-L48

I guess it also has to do with what ROS has found.

WofWca: Oh. And I just found out that the same applies to sticker messages. Oh well...

https://github.com/deltachat/deltachat-desktop/blob/dd857469f343053271f39f0f0befbb010398be15/packages/frontend/src/stockStrings.ts#L54-L55

Edit: aaaaaand for images and videos

https://github.com/deltachat/deltachat-desktop/blob/dd857469f343053271f39f0f0befbb010398be15/packages/frontend/src/stockStrings.ts#L18-L22

packages/e2e-tests/tests/basic-tests.spec.ts
WofWca: It does, see my comment above.

packages/frontend/src/components/message/Message.tsx
Simon-Laux: The point of reply in my view is that you can jump to the referenced message, the text of it is not that important.
So I personally would allow to reply to video chat messages.

WofWca: I guess the solution would be to eliminate the `setStockStrings` API altogether.

Simon-Laux: localising info messages in the UI instead of the core by type would make sense (info messages are not translated after they are received when the user changes their language).
Maybe the way it is currently is because extra properties are not trivial in the cffi? we'd need extra properties to transfer the substitution data to the UI (for example: name of the person that was added removed and who did it).

I wouldn't propose removing translation in core altogether, because there may very well be use cases where it is the best/only solution.

r10s: _all_ message types have strings used eg. in summaries, copy-to-clipboard, reactions to messages etc

let's leave locale discussion out of this PR, this is another discussion or issue

it is fine for now, if some translated strings are quoted, this is the case since years :)

Simon-Laux: > So I personally would allow to reply to video chat messages.

FTR: those video chat messages will be removed soon, then this point makes no sense anymore:
https://github.com/chatmail/core/issues/7137

PR #5351, opened 2025-07-28
chore: target-electron: upgrade `application-config`

saper:
application-config@3.0.0 understands FreeBSD

This should fix the problem reported in the forum:

https://support.delta.chat/t/delta-chat-on-freebsd/2948/16?u=saper


WofWca (commented):
Thanks!
The title seems wrong, this is about `target-electron`, and not `target-browser`.

I didn't verify this, but I don't mind a simple upgrade anyway.

WofWca (commented):
Hm, actually, doesn't this MR have a potential to break existing installations?

WofWca:
The https://github.com/LinusU/node-application-config/releases section includes migration guides, but they don't say anything special, besides "upgrade node and use ESM imports".

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## [Unreleased][unreleased]
 
 ### Changed
+- upgrade `application-config` to `^3.0.0`
```
WofWca: This should probably be in the "Fixed" section, and should probably say "Support FreeBSD" or "improve support for FreeBSD".

But it's not super important I guess.

WofWca (approved):
OK, about breaking: here is the diff:

https://github.com/LinusU/node-application-config/compare/v1.0.1...v3.0.0

Looking at the commit messages, there is nothing breaking in the package itself, but it upgrades `application-config-path
TypeScript icon, indicating that this package has built-in type declarations`: https://github.com/LinusU/node-application-config-path/compare/v0.1.0...v1.0.0.
And the culprit is this commit https://github.com/LinusU/node-application-config-path/commit/4ca1cc5cd3521a3d30ff3267c5821c10b680df60, which only changes behavior for `freebsd` and `openbsd`.

Thus I think this can be merged.

saper:
There is a breaking change: they switched to returning promises. Can you double-check if `appConfig.write(copy)` gets resolved correctly somewhere else and the error is reported?

WofWca:
Well the code looks pretty much equivalent to me, Promise-wise.

saper:
> Well the code looks pretty much equivalent to me, Promise-wise.

Apparently no caller cares about the failure of the save process, so we should at least spit out the message somewhere.

saper:
Now it gets logged like this 
````
47s [i]main/state: Saving state to /tmp/.brokenconfig/DeltaChat/config.json
47s [E]main/state: State save failed Error: EISDIR: illegal operation on a directory, read
    at async readFileHandle (node:internal/fs/promises:549:24)
    at async main (/usr/home/saper/src/deltachat-desktop/node_modules/.pnpm/write-json-file@5.0.0/node_modules/write-json-file/index.js:37:16)
    at async writeJsonFile (/usr/home/saper/src/deltachat-desktop/node_modules/.pnpm/write-json-file@5.0.0/node_modules/write-json-file/index.js:81:2)
    at async ApplicationConfig.write (/usr/home/saper/src/deltachat-desktop/node_modules/.pnpm/application-config@3.0.0/node_modules/application-config/index.js:27:5) {
  errno: -21,
  code: 'EISDIR',
  syscall: 'read'
} 
<anonymous> (/usr/home/saper/src/deltachat-desktop/packages/target-electron/src/desktop_settings.ts:58:11)
````

File: packages/target-electron/src/desktop_settings.ts [hunk 2]
```diff
@@ -57,14 +54,9 @@ class PersistentState extends EventEmitter {
...
-      appConfig.write(copy, (err: any) => {
-        if (err) {
-          log.error('State save failed', err)
-          rej(err)
-        }
-        res(err)
-      })
+    return appConfig.write(copy).catch(err => {
+      log.error('State save failed', err)
+      return err
```
WofWca:
```suggestion
      throw err
```

Otherwise the promise returned from `saveImmediate` never rejects.

saper:
```
29s [i]main/state: Saving state to /tmp/.brokenconfig/DeltaChat/config.json
29s [E]main/state: State save failed Error: EISDIR: illegal operation on a directory, read
    at async readFileHandle (node:internal/fs/promises:549:24)
    at async main (/usr/home/saper/src/deltachat-desktop/node_modules/.pnpm/write-json-file@5.0.0/node_modules/write-json-file/index.js:37:16)
    at async writeJsonFile (/usr/home/saper/src/deltachat-desktop/node_modules/.pnpm/write-json-file@5.0.0/node_modules/write-json-file/index.js:81:2)
    at async ApplicationConfig.write (/usr/home/saper/src/deltachat-desktop/node_modules/.pnpm/application-config@3.0.0/node_modules/application-config/index.js:27:5) {
  errno: -21,
  code: 'EISDIR',
  syscall: 'read'
} 
<anonymous> (/usr/home/saper/src/deltachat-desktop/packages/target-electron/src/desktop_settings.ts:58:11)
(node:61669) UnhandledPromiseRejectionWarning: Error: State save failed
    at <anonymous> (/usr/home/saper/src/deltachat-desktop/packages/target-electron/src/desktop_settings.ts:59:13)
(node:61669) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 6)

```

saper:
I think this is now ready...

packages/target-electron/src/desktop_settings.ts
```diff
@@ -57,14 +54,9 @@ class PersistentState extends EventEmitter {
...
-      appConfig.write(copy, (err: any) => {
-        if (err) {
-          log.error('State save failed', err)
-          rej(err)
-        }
-        res(err)
-      })
+    return appConfig.write(copy).catch(err => {
+      log.error('State save failed', err)
+      throw new Error('State save failed', { cause: err })
```
WofWca, line 59: Nice, I didn't know about the `cause` option.

saper: > Nice, I didn't know about the `cause` option.

This is relatively new. [It appeared in Node 18 I think](https://nodejs.org/docs/latest-v18.x/api/errors.html#new-errormessage-options). If we need to support older ones, we should not use it, but then we would have to pack a whole exception as a string or something.

PR #5357, opened 2025-07-29
Archived Chats title should have user-select: none

ralphtheninja:
Fixes:

Screenshot from 2025-07-30 00-49-37


WofWca (approved):
A minor change, can't break much.
Looks good.

Issue #3833, opened 2024-05-19
adapt to last-minute instant-onboarding changes


r10s:
https://github.com/deltachat/deltachat-android/pull/3060, https://github.com/deltachat/deltachat-android/pull/3064, https://github.com/deltachat/deltachat-android/pull/3068 are the outcome of one-to-one discussions during the current gathering; see there fore reasoning and discussion.

in summary, the following changes are needed in onboarding:

- [x] pull translations
- [x] create a regular secondary button with the string **Use Other Server** (`instant_onboarding_show_more_instances`) below "Agree & Create Profile"
- [x] on tapping **Use Other Server**, an dialog (same as for "I already have a login") with the following options should open:
    - **List Chatmail Servers** (`instant_onboarding_other_server`) - do the same as "Explore Other Options" before
    - **Classic E-Mail Login** (`manual_account_setup_option`) - do the same as "I Already Have A Login / Manual Login" before
    - **Scan Invitation Code** (`scan_invitation_code`) - do the same a "Scan QR Code" before
- [x] remove the lower two buttons "Explore Other Options" and "Scan QR Code"
- [x] remove  "I Already Have A Login / Manual Login" (options left are "Add as Second Device" and "Restore From Backup")

the following changes are needed in "New Chat":

- [x] rename "Scan QR Code" to **New Contact** (`menu_new_contact`) and move it up
- [x] if `isChatmail()` is `true`, the placeholder must no longer read "Enter name or e-mail address" but just **Search** (`search`)
- [x] if an email address is entered, rename "New Contact" to **Add Contact Manually** (`menu_new_classic_contact`) 

(showing it unconditional to `isChatmail()` seems to be the easiest option, gist is that when `isChatmail()` is `true`, there is no easy path to enter an email address (alternatives would be https://github.com/deltachat/deltachat-android/pull/3064 , but that would require a menu in the "QR Code" dialog and may be harder)

in manual contact adding activity:

- [ ] use the title **Add Contact Manually** (`menu_new_classic_contact`)
- [ ] add the hint `new_classic_contact_explain` below the input fields

PR #5358, opened 2025-07-30
Hide add contact manually if on chatmail

### LINKED ISSUE #5336: Can add address-contacts to encrypted group (but it rightfully fails)
- Delta Chat Version: 2.6.0
- Expected behavior: Can't add address-contacts to encrypted group in the "Add Members" dialog
- Actual behavior: Can
  1. "New Chat"
  2. "New Group"
  3. "Add Members"
  4. Type in an email address and click on the checkbox.
  5. Finalize group creation
  6. Take a look at the list of members. You'll see that the address-contacts are not present there.
- Screenshots:
  
  Image

- Logs:

This is related to https://github.com/deltachat/deltachat-desktop/issues/5294 :

> remove "Add Contact Manually" from the "New Chat" dialog. double check, the option exist at "New Chat / New Contact", "QR Code" and "New Chat / New E-Mail / Add Recipients"

Maybe we after all need a separate "Add contact manually" button, because this whole thing with typing the address in search is weird. It's a search UX-wise, and you don't expect to "find" an address that you never added to Delta Chat there.

nicodh:
This is the first step to resolve _"remove "Add Contact Manually"_ in #5294

"Add contact manually" should never be shown when on a Chatmail account since it is not possible to add a contact just by email.

It does only partial resolve #5336 though, since on non chatmail accounts it is still possible to add contacts manually to Groups. That willl be addressed in another PR

### DIFF
--- a/packages/frontend/src/components/helpers/PseudoListItem.tsx
+++ b/packages/frontend/src/components/helpers/PseudoListItem.tsx
@@ -117,13 +117,13 @@ export const PseudoListItemAddContact = ({
   const settingsStore = useSettingsStore()[0]
   const isChatmail = settingsStore?.settings.is_chatmail === '1'
 
+  if (isChatmail) return null
+
   return (
     <PseudoListItem
       id='newcontact'
       cutoff='+'
-      text={
-        isChatmail ? tx('menu_new_classic_contact') : tx('menu_new_contact')
-      }
+      text={tx('menu_new_contact')}
       subText={
         queryStrIsEmail ? queryStr + ' ...' : tx('contacts_type_email_above')
       }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased][unreleased]
 
 ### Changed
+- removed "Add contact manually" when on chatmail account #5358
 
 ### Added
 


packages/frontend/src/components/helpers/PseudoListItem.tsx
```diff
@@ -117,13 +117,13 @@ export const PseudoListItemAddContact = ({
   const settingsStore = useSettingsStore()[0]
   const isChatmail = settingsStore?.settings.is_chatmail === '1'
 
+  if (isChatmail) return null
```
WofWca, line 120: This approach is IMO not the best semantically, might cause bugs in the future. Namely, `contactsAndExtraItems.length` is calculated to account for the presence of this element, but it's not present.

https://github.com/deltachat/deltachat-desktop/blob/a36f56ff9c7600ef380f90db33dcae54cb48f08b/packages/frontend/src/components/dialogs/CreateChat/index.tsx#L345

https://github.com/deltachat/deltachat-desktop/blob/a36f56ff9c7600ef380f90db33dcae54cb48f08b/packages/frontend/src/components/dialogs/CreateChat/index.tsx#L212-L226

It's better to modify `contactsAndExtraItems`, in case of the `CreateChat` component. For other components where `PseudoListItemAddContact` is used, it's probably the same.

WofWca (approved):
If the plan is to completely remove the possibility of adding address-contacts on Chatmail accounts, including `echo@nine.testrun.org`, `admin@nine.testrun.org` ([`passthrough_recipients`](https://github.com/chatmail/relay/blob/d880937d448a1fcdd94006659e54b144d0b912f6/chatmaild/src/chatmaild/ini/chatmail.ini.f#L41-L46)), and for the possible whitelisted senders (`passthrough_senders`), then this does the job (apart from the "add members" part that you mentioned in the MR description).

Related:
- https://github.com/deltachat/deltachat-android/issues/3163
- https://github.com/deltachat/deltachat-android/pull/3166
- https://github.com/deltachat/deltachat-android/pull/3064
- https://github.com/deltachat/deltachat-desktop/issues/3833

packages/frontend/src/components/helpers/PseudoListItem.tsx
nicodh: Yes I will address this in the follow up (that's why the branch is named hide-add-manually-step1

But this change makes sure the PseudoListItemAddContact is never shown on chatmail accounts.

We need to pass a parameter to the AddMemberDialog to decide wether the add contact manually should be shown etc.

WofWca: > But this change makes sure the PseudoListItemAddContact is never shown on chatmail accounts.

Yes, I think keeping the condition in two places, inside and outside, is ok.

PR #5363, opened 2025-07-31
Avoid adding plain address contacts to groups

### LINKED ISSUE #5336: Can add address-contacts to encrypted group (but it rightfully fails)

nicodh:
resolves #5336

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 ### Added
 
 ### Fixed
-- disable adding address contacts to groups #5336
+- don't show "add contact manually" (by email address) for groups #5336
 - Make `Archived Chats` title non selectable
 
 <a id="2_8_0"></a>


packages/frontend/src/components/dialogs/CreateChat/index.tsx
```diff
@@ -246,13 +247,20 @@ function CreateChatMain(props: CreateChatMainProps) {
...
     if (!queryStrIsValidEmail) return
 
-    const contactId = await BackendRemote.rpc.createContact(
-      accountId,
-      queryStr.trim(),
-      null
-    )
-    await createChatByContactId(accountId, contactId)
-    onClose()
+    try {
```
WofWca, line 250: Are these `try` wrappers unrelated or are they somehow relevant?

packages/frontend/src/components/dialogs/AddMember/AddMemberDialog.tsx
```diff
@@ -32,6 +32,10 @@ export function AddMemberDialog({
     queryStrIsValidEmail,
     refreshContacts,
   } = useLazyLoadedContacts(listFlags, queryStr)
+
+  // compare bitwise if address flag is set
+  const allowAddManually = (listFlags & C.DC_GCL_ADDRESS) !== 0
```
WofWca, line 37: Hmmm, I'd prefer a more direct way to determine this.

CHANGELOG.md
```diff
@@ -8,7 +8,7 @@
 ### Added
 
 ### Fixed
-
+- disable adding address contacts to groups #5336
```
```suggestion
- don't show "add contact manually" (by email address) on Chatmail accounts, because it's useless most of the time #5336
```

WofWca (approved):
Looks good, though I didn't test.

This also addresses https://github.com/deltachat/deltachat-desktop/pull/5358#discussion_r2242710963

packages/frontend/src/components/dialogs/CreateChat/index.tsx
nicodh: well I came on this when testing what happens if createGroup throws an error. It turned out nothing. So I wrapped it in try catch and found this other case where a thrown error would just vanish. So yes it is unrelated

packages/frontend/src/components/dialogs/AddMember/AddMemberDialog.tsx
nicodh: We could an additional parameter. I heard that there are anyway changes planned in that area so we will soon refactor here

CHANGELOG.md
nicodh: Adapted it, but _"removed "Add contact manually" when on chatmail account"_ is already in the changed section

Issue #5365, opened 2025-07-31
classic mail: hide 'enlarge avatar' and 'disappearing messages'


r10s:
'enlarge avatar' and 'disappearing messages' do not make sense for classic email and should be hidden.

cmp https://github.com/deltachat/deltachat-android/pull/3847 and https://github.com/deltachat/deltachat-ios/pull/2797

WofWca:
And "edit message".

WofWca:
After https://github.com/deltachat/deltachat-desktop/pull/5371, what remains are avatars. But I'm not sure yet how to check whether to full-screen the avatar for address-contacts. For chats we have `isEncrypted`, but not for contacts.

r10s:
there is isKeyContact() for contacts, see iOS implementation is using that, android uses a different approach (see links atop for both PR)

PR #5371, opened 2025-08-01
fix: hide "Disappearing Messages" and "Edit" for unencrypted chats

WofWca:
- **fix: hide "Edit Message" for unencrypted chats**
- **fix: hide "Disappearing Messages" for unencrypted**

Partially addresses
https://github.com/deltachat/deltachat-desktop/issues/5365.

What's left is avatars.

PR #5373, opened 2025-08-02
fix: hide fullscreen avatar view for unencrypted

WofWca:
A side effect is that one can't enlarge the "Device Messages"
chat avatar. Which is probably good.

Closes https://github.com/deltachat/deltachat-desktop/issues/5365,
together with
https://github.com/deltachat/deltachat-desktop/pull/5371.

### DIFF
--- a/packages/frontend/src/components/Avatar/index.tsx
+++ b/packages/frontend/src/components/Avatar/index.tsx
@@ -138,8 +138,7 @@ export function ClickForFullscreenAvatarWrapper(
 ) {
   const { openDialog } = useDialog()
 
-  const { children, filename, disableFullscreen, ...buttonProps } =
-    props
+  const { children, filename, disableFullscreen, ...buttonProps } = props
 
   return filename && !disableFullscreen ? (
     <button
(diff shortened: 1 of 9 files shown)


r10s:
> A side effect is that one can't enlarge the "Device Messages"
chat avatar. Which is probably good.

yes, that's great as well!

when on that, maybe also do not allow enlarging "Saved Messages" avatar

(also filed an issue for android and will do a PR for iOS)

WofWca:
I added a code comment.

packages/frontend/src/components/Avatar/index.tsx
```diff
@@ -132,25 +133,56 @@ export function AvatarFromContact(
 export function ClickForFullscreenAvatarWrapper(
   props: React.ButtonHTMLAttributes<HTMLButtonElement> & {
     filename?: string
+    disableClickForFullscreen: boolean
```
nicodh: Naming proposal: disableFullscreen would be sufficient imo

packages/frontend/src/components/Avatar/index.tsx
```diff
@@ -132,25 +133,56 @@ export function AvatarFromContact(
...
+    ? [chatOrContact, undefined]
+    : [undefined, chatOrContact]
+
+  // It's just an "envelope" icon, there is no need to view it in full screen.
+  // See https://github.com/deltachat/deltachat-desktop/issues/5365.
+  const isAvatarADummyImage =
+    chat != undefined ? !chat.isEncrypted : !contact.isKeyContact
+
+  return isAvatarADummyImage
+}
```
nicodh, line 185: Why do we need to declare 4 variables here?
What is the advantage compared to

```
export function shouldDisableClickForFullscreen(
  chatOrContact: { isEncrypted?: boolean } | { isKeyContact?: boolean }
): boolean {
  // It's just an "envelope" icon, there is no need to view it in full screen.
  // See https://github.com/deltachat/deltachat-desktop/issues/5365.
  if ('isEncrypted' in chatOrContact) {
    return !chatOrContact.isEncrypted
  } else if ('isKeyContact' in chatOrContact) {
    return !chatOrContact.isKeyContact
  }
  return false
}
```

Would be better to read imo

WofWca: OK, I removed some vars.

I wanted to ensure type safety, e.g. handle the case where both `BasicChat` and `Contact` have the same fields.

PR #5376, opened 2025-08-04
fix: a11y: focus "React" menu on open

WofWca:


nicodh (approved):
Reviewed by reading & testing

Too sad that we can't jump into the emoji list in Picker provided by emoji-mart since all the buttons have tabindex="-1"

Just had a look but seems there is no way to change that easily
https://github.com/missive/emoji-mart/blob/main/packages/emoji-mart/src/components/Picker/Picker.tsx#L770

WofWca:
> Too sad that we can't jump into the emoji list in Picker

Why would we want to though? Are you talking about to jumping there when the emoji picker opens, or about focusing the emojis when it is open already?

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Fixed
 - don't show "Edit Message", "Disappearing Messages" and fullscreen avatar view in classic E-Mail chats #5365
+- accessibility: improve keyboard and screen reader accessibility of the "Add Reaction" menu
```
```suggestion
- accessibility: improve keyboard and screen reader accessibility of the "Add Reaction" menu #5376
```

PR #5382, opened 2025-08-06
Truncate app title and description

### LINKED ISSUE #5381: app selector: do not wrap title or subtitle
title and subtitle of apps are meant to be short and to fit into one line, this is also documented that way.

if apps do not follow these rules text shall just be truncated - and not broken into multiple lines, cluttering UI for the whole selector, see eg. the first app here:

Image

the layout is broken, things are no longer aligned, and it looks bad. webxdc.org shows ellipsis instead:

Image

there is no need to count characters or so, just not use more than one line.

for the concrete example, i also contacted the author to remove at least "Webxdc" from the title, but it is good to have a rendering that "always" fits.

when on that, maybe also show the "author" line greyish, eg. 0x808080 (that way dark and light mode both are fine)

ralphtheninja:
Closes #5381 

Before:

image

After:

image

### DIFF
--- a/packages/frontend/src/components/AppPicker/styles.module.scss
+++ b/packages/frontend/src/components/AppPicker/styles.module.scss
@@ -94,8 +94,33 @@
   display: flex;
   flex-direction: column;
   justify-content: center;
+  width: 208px;
   p {
     margin: 2px 0;
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+  }
+  button {
+    border: none;
+    background: none;
+    color: var(--globalLinkColor);
+    text-align: left;
+    margin: 2px 0;
+    padding: 0;
+  }
+}
+
+.appInfoFullWidth {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  width: 100%;
+  p {
+    margin: 2px 0;
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
   }
   button {
     border: none;
@@ -124,6 +149,13 @@
 .appName {
   font-size: 16px;
   font-weight: bold;
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
+
+.appAuthor {
+  color: #808080;
 }
 
 .tabBar {
--- a/packages/frontend/src/components/AppPicker/styles.module.scss
+++ b/packages/frontend/src/components/AppPicker/styles.module.scss
@@ -94,7 +94,7 @@
   display: flex;
   flex-direction: column;
   justify-content: center;
-  width: 208px;
+  width: 100%;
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/AppPicker/styles.module.scss
```diff
@@ -94,8 +94,33 @@
...
+.appInfoFullWidth {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  width: 100%;
+  p {
+    margin: 2px 0;
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
```
ralphtheninja: This is a bit ugly but solves the problem. I'm open to suggestions on how to make this cleaner. So basically what I want is to force the parent div's width to 208 pixels **when in the app picker**, but in the dialog it can be 100% width.

ralphtheninja: What I'd really like to do is to be able to use `.appInfo` in the dialog, but also set the width to 100%. **OR** in the app picker, specifically set the width to 208 pixels.

Why this strange value you say? Well the parent container is set to 300px, the app icon is 64px with 16px margin, which leaves 208px. Then the child `<p>` items will adjust to this width and will not expand the parents width.

ralphtheninja:
@r10s Are you fine with the ellipsis or should we just truncate hard? The app title here will be "fixed" once they update the app name.

packages/frontend/src/components/AppPicker/styles.module.scss
nicodh: You can remove the redundant .appInfoFullWidth completely, set the width in .appInfo to 100% and add these lines.

```
/** limited width in app picker */
.appListItem .appInfo {
  width: 208px;
}
```

and then remove the condition in index.tsx of course

`className={setSearchQuery ? styles.appInfoFullWidth : styles.appInfo} `

nicodh (changes requested):

r10s:
> @r10s Are you fine with the ellipsis or should we just truncate hard?

an ellipsis is totally fine, and even the nicer solution.

thanks a lot for taking care - i tested, and from that point of view, things improved a lot and are fine!

packages/frontend/src/components/AppPicker/styles.module.scss
ralphtheninja: @nicodh Beautiful! Thanks! :hugs:

ralphtheninja: Just to clarify, the math above doesn't check out because I forgot to mention there are multiple divs involved. Ultimately there's **a** div with hard width set to 300px, which also has some padding:

image

So it's really 288 - 64 - 16 = 208.

nicodh (approved):

PR #5384, opened 2025-08-06
Add comments to ChatList and related components

nicodh:
Wording is hard: we have a ChatList that can also be a list of search results which means a list of messages or contacts. So a ChatListItem can be a ChatListChatItem or a ChatListMessageItem etc...

Just a first draft here, besides commenting a rewording: ChatListItemNormal => DefaultChatListItem

#skip-changelog

### DIFF
--- a/packages/frontend/src/components/chat/ChatList.tsx
+++ b/packages/frontend/src/components/chat/ChatList.tsx
@@ -62,6 +62,24 @@ const enum LoadStatus {
   LOADED = 2,
 }
 
+/**
+ * This component holds either a list of chats OR the result
+ * of a search query including chats, contacts and messages.
+ *
+ * <ChatList>
+ *   <ChatListPart> // virtual list (one for each type of search result)
+ *     <ChatListItemRow>
+ *       <ChatListItem(Default (Chat) | Message | Contact) />
+ *     </ChatListItemRow>
+ *   </ChatListPart>
+ * </ChatList>
+ */
+
+/**
+ * wrapper for a virtual list that handles scrolling and loading items
+ *
+ * up to 3 ChatListParts can be present in the chat list
+ */
 export function ChatListPart<
   T extends
     | ChatListItemData
--- a/packages/frontend/src/components/chat/ChatList.tsx
+++ b/packages/frontend/src/components/chat/ChatList.tsx
@@ -44,8 +44,8 @@ import useTranslationFunction, {
 
 import type {
   ChatListItemData,
-  ContactChatListItemData,
-  MessageChatListItemData,
+  ChatListContactItemData,
+  ChatListMessageItemData,
 } from './ChatListItemRow'
 import { isInviteLink } from '../../../../shared/util'
 import { RovingTabindexProvider } from '../../contexts/RovingTabindex'
@@ -78,13 +78,14 @@ const enum LoadStatus {
 /**
  * wrapper for a virtual list that handles scrolling and loading items
  *
- * up to 3 ChatListParts can be present in the chat list
+ * ChatList has two modes: regular and search mode.
+ * In search mode there are 3 ChatListParts: for chats, contacts, and messages
  */
 export function ChatListPart<
   T extends
     | ChatListItemData
-    | ContactChatListItemData
-    | MessageChatListItemData,
+    | ChatListContactItemData
+    | ChatListMessageItemData,
 >({
   isRowLoaded,
   loadMoreRows,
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/chat/ChatListItem.tsx
```diff
@@ -302,8 +302,11 @@ function ChatListItemError({
   )
 }
 
-function ChatListItemNormal({
-  chatListItem,
+/**
+ * Default is Chat Item
+ */
+function DefaultChatListItem({
```
WofWca: "Default" sounds weird, seems to imply that there is some kind of logic, that if you don't specify the item type, this one will be assumed.

packages/frontend/src/components/chat/ChatListItem.tsx
```diff
@@ -312,7 +315,7 @@ function ChatListItemNormal({
   isContextMenuActive,
   ...rest
 }: {
-  chatListItem: Type.ChatListItemFetchResult & {
+  chatItem: Type.ChatListItemFetchResult & {
```
WofWca: IDK IDK. `chatItem` reads as "an item of a chat", which is a message.

packages/frontend/src/components/chat/ChatListItemRow.tsx
```diff
@@ -59,6 +73,9 @@ export type MessageChatListItemData = {
   isSingleChatSearch: boolean
 }
 
+/**
+ * Data for a contact item
```
WofWca, line 77: These comments seem redundant, the variable name is sufficient IMO.

packages/frontend/src/components/chat/ChatList.tsx
```diff
@@ -62,6 +62,24 @@ const enum LoadStatus {
...
+ *       <ChatListItem(Default (Chat) | Message | Contact) />
+ *     </ChatListItemRow>
+ *   </ChatListPart>
+ * </ChatList>
+ */
+
+/**
+ * wrapper for a virtual list that handles scrolling and loading items
+ *
+ * up to 3 ChatListParts can be present in the chat list
```
WofWca: This sounds really weird. I'd either say something along the lines of "ChatList has two modes: regular and search mode. In the search mode there are 3 ChatListParts: for chats, contacts, and messages".

Or simply removed this comment, because one can just look at the references to `ChatListPart`.

packages/frontend/src/components/chat/ChatListItem.tsx
nicodh: You used the term regular yourself in the first comment. That's what I want to express: the "default" is that the chatlist is a list of chats and in search mode there can be other items which are not of type chat

packages/frontend/src/components/chat/ChatListItem.tsx
nicodh: As I sad: wording is hard. I think an item in a chat list is a chat. But since we have different types of items we have chatItems, messageItems and contactItems.

But that's why I said it's a first draft. The idea of rewording is that it is more intuitiv to understand. If that is not the case for you it's probably the wrong wording. But the existing wording was not at all "self explaining" for me. ChatListItemNormal - hard to say what is "normal" 
And ChatListItem is just an item in a chat list which in our case can be of 3 different types so the name doesn't help to distinguish the item from others

packages/frontend/src/components/chat/ChatList.tsx
nicodh: Your description is fine, even more verbose.
But I would not remove it, the references are
ChatListItemData
ContactChatListItemData
MessageChatListItemData

not really self explaining, what is ContactChatListItemData ?
the data of the items of a Contacts Chat List?
But what is a MessageChatList? Doesn't make sense. Only if you know what is meant...

packages/frontend/src/components/chat/ChatListItem.tsx
WofWca: I think that "Default" is different from "Regular". "Regular" fits better here, see

> seems to imply that there is some kind of logic, that if you don't specify the item type, this one will be assumed.

packages/frontend/src/components/chat/ChatListItem.tsx
WofWca: I'd say it should either stay as `chatListItem` or just `chat`.

farooqkz:
nooooooo someone else is again stealing me contributions idea :(

PR #5389, opened 2025-08-08
Share email by address not VCard

### LINKED ISSUE #5364: Task: share email contacts by email not by VCard
Related to https://github.com/deltachat/deltachat-android/issues/3850

Sharing an email contact should just copy the email address to the chat it should be shared with. To avoid the illusion that it's a key contact that comes with a key

nicodh:
resolves #5364

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 ### Changed
 
 ### Fixed
-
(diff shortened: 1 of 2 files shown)


r10s (approved):
makes sense, and this is what we're doing on android as well.

also tested a bit, cannot say much to the code, but lgtm

PR #5401, opened 2025-08-14
No email icon for not downloaded messages

### LINKED ISSUE #5399: Not fully downloaded messages display an ✉️ icon
Version is 2.10.

Not fully downloaded messages look like this:
Image

Previously the message had "no padlock", but now that "no padlock" is replaced with envelope, we should not display envelope in cases where we don't know whether the message is encrypted or not.

nicodh:
resolves #5399

### DIFF
--- a/packages/frontend/src/components/message/MessageMetaData.tsx
+++ b/packages/frontend/src/components/message/MessageMetaData.tsx
@@ -53,7 +53,8 @@ export default function MessageMetaData(props: Props) {
       {/* FYI the email doesn't need `aria-live`
       as we don't expect it to get removed. See
       https://github.com/deltachat/deltachat-desktop/pull/5023#discussion_r2059382983 */}
-      {!encrypted && (
+      {!encrypted && status !== 'in_noticed' && (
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/message/MessageMetaData.tsx
```diff
@@ -53,7 +53,8 @@ export default function MessageMetaData(props: Props) {
       {/* FYI the email doesn't need `aria-live`
       as we don't expect it to get removed. See
       https://github.com/deltachat/deltachat-desktop/pull/5023#discussion_r2059382983 */}
-      {!encrypted && (
+      {!encrypted && status !== 'in_noticed' && (
+        // if a message is in status in_notice we don't know if it is encrypted or not
```
r10s, line 57: this seems wrong, also the comment.

you can see this also when opening a classic mail chat with fresh messages, the last message initially does not have an envelope as it is not noticed yet (but fresh).

i think, what is meant is `if !encrypted && downloadState == DONE { show envelope }`

PR #5404, opened 2025-08-16
Reload main window on language change

### LINKED ISSUE #5403: Not all translations get updated when changing the language
- Operating System (Linux/Mac/Windows/iOS/Android): N/A
- Delta Chat Version: 1.58.2
- Expected behavior: All strings should be updated according to the new locale selected
- Actual behavior: Not all of them are updated
- Steps to reproduce the problem: Launch DC with Language A, change it to lang B and see some strings are not updated

### Notes
Restarting DeltaChat is a workaround. 
I think the problem is not using dynamic translation hook and instead using `window.static_translate`.

nicodh:
resolves #5403 

#skip-changelog

We have to reload the main window (not restart the app!) to make sure all caches are reset. Otherwise we would always have to care explicit that caches are reset on language change.

Since we "remember" last open chat and draft etc. there is no impact for the user except a short reload of the window.

FYI Signal has to restart the whole app on each language change

image

@r10s  TBD: do we need to show a confirm dialog before reloading?

### DIFF
--- a/packages/target-electron/runtime-electron/runtime.ts
+++ b/packages/target-electron/runtime-electron/runtime.ts
@@ -420,6 +420,7 @@ class ElectronRuntime implements Runtime {
     )
     ipcBackend.on('chooseLanguage', (_ev, locale) => {
       this.onChooseLanguage?.(locale)
+      ipcBackend.send('reload-main-window')
     })
     ipcBackend.on('theme-update', () => this.onThemeUpdate?.())
     ipcBackend.on('showAboutDialog', () => this.onShowDialog?.('about'))


r10s:
did not try yet, but a confirmation seems to be not needed. 

(in case of signal, the restarting probably takes longer, and and the window is gone during that time)

farooqkz:
Why reload the window? Why not just use `useTranslationFunction()` everywhere instead of `window.static_translate`?

nicodh:
> Why reload the window? Why not just use `useTranslationFunction()` everywhere instead of `window.static_translate`?

As i wrote: it's not the static_translate but the cache. ChatList caches all chat items, so the locale that was set when the cache was filled still remains (for summaryText etc.), if the whole chache is not invalidated. The same might apply to contacts that are cached and so forth.

farooqkz:
Ah yeah you are right.

Simon-Laux:
> if the whole chache is not invalidated.

Emitting events to inform UI that cache needs to be invalidated is core responsibility.
I made a PR that emits the required events when language changes: https://github.com/chatmail/core/pull/7719

---

why?:

This PR is a workaround which I don't like for several reasons:
- it breaks reloading when you change translation strings during development (when it reloads it forgets react state, so the dialog you changed the text in is not open anymore)
- it forgets react state when you change language during development, so comparing dialogs is more annoying because they close and you have to re-open them
- it bypasses all the effort we put into the `useTranslationFunction()` hook system, which works really well outside these edge cases
- it is only implemented for electron and there it just skips the rest of the code. (tauri is skipped and hot reloading logic is not removed)

I think it is a decent hotfix/temporary, but we can do better, would be sad to throw away a nice feature which works to 95%. If core emits the events to invalidate the cache, then it would work to nearly[^1] 100%.

[^1]: "nearly" because there may still be a UI bug with translation of static timestamps, but that could be likely fixed by changing them to use the `useTranslationFunction()` - hook

PR #5408, opened 2025-08-16
Check downloadState in Metadata

### LINKED ISSUE #5407: double check envelope state
this issue is just a reminder the comment at https://github.com/deltachat/deltachat-desktop/pull/5401 does not get lost.

no version with this PR was released yet.

nicodh:
resolves #5407 

#skip-changelog (just a followup)


r10s (approved):
lgtm, as far as i can tell

not sure about the internal logic of forwarding downloadState and if it can be not set or so

PR #5423, opened 2025-08-23
Flip fuses

nicodh:
See https://www.electronjs.org/docs/latest/tutorial/fuses

Result is:

  RunAsNode is Disabled
  EnableCookieEncryption is Disabled (default) // not needed
  EnableNodeOptionsEnvironmentVariable is Disabled
  EnableNodeCliInspectArguments is Disabled
  EnableEmbeddedAsarIntegrityValidation is Disabled (default) 
  OnlyLoadAppFromAsar is Disabled  (default)
  LoadBrowserProcessSpecificV8Snapshot is Disabled  (default) // causes problems on AppImage (and more?)
  GrantFileProtocolExtraPrivileges  is Disabled (default)

#public-preview


Simon-Laux:
> EnableCookieEncryption

probably not useful since we store no cookies, unless it also applies to localstorage.
Theoretical (not confirmed) problem: Could potentially show a keychain dialog on linux (that have auto login to desktop), which could be confusing to users.

> onlyLoadAppFromAsar

this will break flatpak, should be conditional: only active when `NO_ASAR` is not set

> GrantFileProtocolExtraPrivileges

maybe we can explore to disable this in the future. we could definitely serve the app files on a custom scheme (I guess it currently uses file scheme?), as long as the file protocol without the extra privileges still allows loading media from the blob dir it should work.

nicodh:
> > EnableCookieEncryption
> 
> probably not useful since we store no cookies, unless it also applies to localstorage. Theoretical (not confirmed) problem: Could potentially show a keychain dialog on linux (that have auto login to desktop), which could be confusing to users.

I think it's no risk to disable it, I just started with the full  maximal settings

> 
> > onlyLoadAppFromAsar
> 
> this will break flatpak, should be conditional: only active when `NO_ASAR` is not set

Ok, thanks for the hint I will adapt it

> 
> > GrantFileProtocolExtraPrivileges
> 
> maybe we can explore to disable this in the future. we could definitely serve the app files on a custom scheme (I guess it currently uses file scheme?), as long as the file protocol without the extra privileges still allows loading media from the blob dir it should work.

Yes I agree. But for now we still need it.

Simon-Laux:
> GrantFileProtocolExtraPrivileges

for reference: the previous attempt at removing file scheme entirely had some problems loading media https://github.com/deltachat/deltachat-desktop/pull/2171, if media still works on restricted file protocol, then we could move just the app code to an internal scheme.

tauri show that a custom protocol for blobs is possible, but even there I am considering adding http streaming for videos:

> - [ ] streaming for videos and larger audio blob files
> ~ https://github.com/deltachat/deltachat-tauri/issues/1

packages/target-electron/build/afterPackHook.mjs
```diff
@@ -186,3 +185,49 @@ async function deleteNotNeededPrebuildsFromUnpackedASAR(
...
+      console.log(file);
+    });
+    throw new Error('Could not apply electron fuses since target not exists: ' + appPath)
+  }
+
+  console.log('Applying electron fuses to:', appPath)
+  await flipFuses(appPath, {
+    version: FuseVersion.V1,
+    [FuseV1Options.RunAsNode]: false, // Disables ELECTRON_RUN_AS_NODE
+    [FuseV1Options.EnableCookieEncryption]: false, // not needed
```
WofWca: > > EnableCookieEncryption
>
> probably not useful since we store no cookies, unless it also applies to localstorage.

How about WebXDC?

Also, why even mess with this fuse if it's not related to that supposed security issue and is disabled by default anyway? To me it seems better just to use the default value, i.e. not specify this here at all.

The same goes for `GrantFileProtocolExtraPrivileges`, and maybe other fuses.

WofWca (commented):
I don't seem to be able to start the portable .exe version. The portable version from another MR works fine.

Code-wise apart from my comment, looks fine.
Including the unrelated changes, including the addition of `@parcel/watcher` to `onlyBuiltDependencies`, although I don't know if it's needed.

packages/target-electron/build/afterPackHook.mjs
nicodh: Of course we could just skip the fuses, if we don't override the defaults but to me it seems more transparent to write all settings explicit here

nicodh:
> I don't seem to be able to start the portable .exe version. The portable version from another MR works fine.

Do you see any error message in the logs?
Maybe related to OnlyLoadAppFromAsar ?

WofWca:
The log doesn't get created. I think it manages to unpack the files to the temp dir, but then the process stops. There is also nothing in the command line if you launch DC through the command line.

nicodh:
> The log doesn't get created. I think it manages to unpack the files to the temp dir, but then the process stops. There is also nothing in the command line if you launch DC through the command line.

So maybe we should generate a preview with 
EnableEmbeddedAsarIntegrityValidation disabled
OnlyLoadAppFromAsar disabled
and check again

dotlambda:
On NixOS this results in
```
Applying electron fuses to: /build/source/packages/target-electron/dist/linux-unpacked/deltachat-desktop
  ⨯ EACCES: permission denied, open '/build/source/packages/target-electron/dist/linux-unpacked/deltachat-desktop'  failedTask=build stackTrace=Error: EACCES: permission denied, open '/build/source/packages/target-electron/dist/linux-unpacked/deltachat-desktop'
```
Any idea why that could be?

nicodh:
> On NixOS this results in
> Applying electron fuses to: /build/source/packages/target-electron/dist/linux-unpacked/deltachat-desktop
>   ⨯ EACCES: permission denied, open '/build/source/packages/target-electron/dist/linux-unpacked/deltachat-desktop'  failedTask=build stackTrace=Error: EACCES: permission denied, open '/build/source/packages/target-electron/dist/linux-unpacked/deltachat-desktop'
> ```
> 
> Any idea why that could be?

Well the command looks correct, and since the build linux-unpacked/deltachat-desktop is created by desktop-builder it should have write access to that file. Anything special on the build system? Do you have access and can test the command:
npx @electron/fuses write --app /build/source/packages/target-electron/dist/linux-unpacked/deltachat-desktop RunAsNode=off

Simon-Laux:
nix seems to link to a system version of electron, maybe that could be a reason why it can not be modified?

could mean that we need to flip the fuse in the nix build script and not in our building process.
maybe we should add another option to disable the fuse setting next to the NO_ASAR option?

nicodh:
@dotlambda should we provide a environment variable to disable/enable the flip fuses step?
SKIP_FUSES

if true we don't run that script.

So far there is no known issue with fuses on NixOS so it just could be skipped. No need to apply the script after build process...

Simon-Laux:
As far as I understand nix currently: we could do a new nix package that wraps the electron package and allows us to set fuses on it:
- should be a new package, because the electron packages is build and cached on the nixos cache server, if we modify compiler option there, then it will be built locally instead, which takes much more time. so a new package that takes the result from the electron package, copies it and flips the fuses on it would not have the problem of increased build times.
- disadvantage is that the end user would have another copy of electron in the end – but with the separated "fused electron" package it could be just a size increase per fuse configuration and not per application, like if signal and other electron apps would set the same fuses, there would be a shared electron.

Ideally there were some runtime options that could be set in the nix wrapper script that starts the electron from the electron package, to avoid this electron binary duplication that is caused by modifying the electron binary.

But I don't know enough about nixos, maybe there is some way to flip the fuses in somekind of overlayfs that the electron binary is not really modified, just that it gets read with the modifications applied in RAM. Although that would only be an option if there is no performance impact.

I believe we should add a `SKIP_FUSES` option. This would make fix the nixos build. Ideally we find a smart way to flip the fuses in the nix package/build script. But nix on mac is not that common, so could also be fine to not set the fuses on the nix package, if it is too hard to do.

Update: I just looked into the signal package: it sets fuses, but on nixos the modified electron binary is not included in the installed package, it also uses the system electron package like we do, so it seems like fuses are generally not set on nixos yet.

dotlambda:
> @dotlambda should we provide a environment variable to disable/enable the flip fuses step?

That sounds like a good idea

PR #5427, opened 2025-08-29
Update packages

nicodh:
#skip-changelog

Just for maintenance to avoid "huge" updates. All updates happened in devDependencies

### DIFF
--- a/packages/target-tauri/runtime-tauri/tsconfig.json
+++ b/packages/target-tauri/runtime-tauri/tsconfig.json
@@ -8,7 +8,6 @@
     "allowJs": false,
     "strictNullChecks": true,
     "strict": true,
-    "moduleResolution": "bundler",
-    "skipLibCheck": true
+    "moduleResolution": "bundler"
   }
 }
--- a/packages/target-tauri/runtime-tauri/tsconfig.json
+++ b/packages/target-tauri/runtime-tauri/tsconfig.json
@@ -9,6 +9,6 @@
     "strictNullChecks": true,
     "strict": true,
     "moduleResolution": "bundler",
-    "skipLibCheck": true,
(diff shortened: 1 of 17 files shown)


packages/target-tauri/runtime-tauri/tsconfig.json
```diff
@@ -8,6 +8,7 @@
     "allowJs": false,
     "strictNullChecks": true,
     "strict": true,
-    "moduleResolution": "bundler"
+    "moduleResolution": "bundler",
+    "skipLibCheck": true
```
WofWca: Is `skipLibCheck` required? In all files?
https://www.typescriptlang.org/tsconfig/#skipLibCheck

nicodh: Yes. All packages had "duplicate identifier" issues after the update

Simon-Laux: > All packages had "duplicate identifier" issues after the update

not for me:
- electron, browser and e2e tests did not have this issue
- frontend and the other packages seem to have issues

the difference is that the packages with issues don't explicitly specify `@types/node` in their dependencies, after doing that it works.

Here is the patch file on top of 5801dc2c4ca3551790110cb9ebe2d439e3628c1d (first commit of this pr): [fix.patch](https://github.com/user-attachments/files/22049077/fix.patch) - I used "catalog:" to specify the node types as workspace dependency so its version stays in sync across all our packages in the future.

So `skipLibCheck` is not needed after all.

nicodh: Ah cool. Thanks for the fix!

WofWca (approved):

PR #5431, opened 2025-08-31
test: add tests for chat list multiselect

WofWca:
Follow-up to
https://github.com/deltachat/deltachat-desktop/pull/5297.

This doesn't cover all the intricate behaviors,
but it does cover the basic ones.

It's a separate MR because the base one is already quite big, and is approved, so I don't want to make noise in it. But it doesn't matter if this one's merged before or after that one.

#skip-changelog

### DIFF
--- a/packages/e2e-tests/tests/chat-list-multiselect.spec.ts
+++ b/packages/e2e-tests/tests/chat-list-multiselect.spec.ts
@@ -244,6 +244,138 @@ test.describe('context menu', () => {
     await page.keyboard.press('Escape')
     await expect(page.getByRole('menu')).not.toBeVisible()
   })
+
+  test('when mixed items are selected, shows appropriate bulk actions', async () => {
+    await getChat(7).click()
+    await expectSelectedChats([7])
+    await getChat(5).click({
+      modifiers: ['ControlOrMeta'],
+    })
+    await expectSelectedChats([7, 5])
+
+    await getChat(5).click({
+      button: 'right',
+    })
+    await page.getByRole('menuitem', { name: 'Pin Chat' }).click()
+
+    // Unfortunately we have to wait for core to respond to the "Pin" action,
+    // otherwise the menu would think that the chats are still not pinned.
+    await expect(
+      chatList.getByRole('tab', { name: 'Some chat ' })
+    ).toContainText([
+      'Some chat 7',
+      'Some chat 5',
+
+      'Some chat 9',
+      'Some chat 8',
+      'Some chat 6',
+      'Some chat 4',
+      'Some chat 3',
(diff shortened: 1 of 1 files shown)


packages/e2e-tests/tests/chat-list-multiselect.spec.ts
```diff
@@ -0,0 +1,481 @@
...
+async function createNDummyChats(n: number) {
+  for (let i = 0; i < n; i++) {
+    await page.getByRole('button', { name: 'New Chat' }).click()
+    await page.getByRole('button', { name: 'New Group' }).click()
+    await page
+      .getByRole('textbox', { name: 'Group Name' })
+      .fill(`Some chat ${(i + 1).toString()}`)
+    await page.getByTestId('group-create-button').click()
+  }
+  await expect(chatList.getByRole('tab', { name: 'Some chat ' })).toContainText(
```
nicodh, line 81: Maybe comment here that the order of the strings in the array matters, since https://elaichenkov.github.io/playwright-expect/modules/tocontaintext.html says the parameter is of type string not string array?

packages/e2e-tests/tests/chat-list-multiselect.spec.ts
```diff
@@ -0,0 +1,481 @@
...
+  })
+  await expectSelectedChats([7, 6, 5, 2])
+
+  await getChat(6).click({
+    button: 'right',
+  })
+  await page.getByRole('menuitem', { name: 'Delete Chat' }).click()
+
+  await expect(page.getByRole('dialog')).toContainText(
+    'Delete 4 chats on all your devices?'
```
nicodh, line 459: This will fail if the translation changes, but I agrree on the decision better have a failing test after a translation update then missing if the dialog does not show the expected text

packages/e2e-tests/tests/chat-list-multiselect.spec.ts
WofWca: IDK what that doc that you linked is. In the official docs it's documented: https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-have-text

I do not think that we want to comment this on every such usage of `toContainText` / `toHaveText`.

packages/e2e-tests/tests/chat-list-multiselect.spec.ts
WofWca: > This will fail if the translation changes

Well yes, but how would we get around this in a perfect world. I doubt that we want to pull the strings dynamically from the translations file.
These are E2E tests after all.

WofWca:
I force-pushed a minor cleanup, not much to review.

PR #5433, opened 2025-08-31
Add one year option for disappearing messages

### LINKED ISSUE #5421: update disappearing messages options
- after some discussion in the forum and other channels, we decided to add the option "After 1 Year" (string key `autodel_after_1_year`) with the value 31536000 (60*60*24*365); see https://github.com/deltachat/deltachat-android/pull/3467 for reference implementation

- the stock strings `DC_STR_EPHEMERAL_TIMER_1_YEAR_BY_YOU` and `DC_STR_EPHEMERAL_TIMER_1_YEAR_BY_OTHER` needs to be set to `ephemeral_timer_1_year_by_you` resp. `ephemeral_timer_1_year_by_other`

- the "1 minute" and "30 minutes" options shall be removed, see https://github.com/deltachat/deltachat-android/pull/3897 for reasoning

counterpart of https://github.com/deltachat/deltachat-ios/issues/2816

nicodh:
resolves #5421

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Added
 -flip electron fuses to avoid security debates #5423
-- add one year for disappearing messages #5421
+- add "one year" option for disappearing messages #5421
 
 ### Fixed
 - don't show "Reply" and "Save" for info messages (e.g. "user A removed user B") and video call invitations #5337
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 -flip electron fuses to avoid security debates #5423
+- add one year for disappearing messages #5421
 
 ### Fixed
 - don't show "Reply" and "Save" for info messages (e.g. "user A removed user B") and video call invitations #5337
--- a/packages/frontend/src/components/dialogs/DisappearingMessages.tsx
+++ b/packages/frontend/src/components/dialogs/DisappearingMessages.tsx
@@ -27,6 +27,7 @@ enum DisappearingMessageDuration {
   ONE_DAY = Timespans.ONE_DAY_IN_SECONDS,
   ONE_WEEK = Timespans.ONE_WEEK_IN_SECONDS,
   FIVE_WEEKS = Timespans.ONE_WEEK_IN_SECONDS * 5,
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Added
 -flip electron fuses to avoid security debates #5423
+- add one year for disappearing messages #5421
```
```suggestion
- add "one year" option for disappearing messages #5421
```

packages/frontend/src/stockStrings.ts
```diff
@@ -118,9 +118,11 @@ export async function updateCoreStrings() {
   strings[C.DC_STR_EPHEMERAL_TIMER_SECONDS_BY_OTHER] = tx(
     'ephemeral_timer_seconds_by_other'
   )
+  // deprecated only used in recent versions
   strings[C.DC_STR_EPHEMERAL_TIMER_1_MINUTE_BY_YOU] = tx(
     'ephemeral_timer_1_minute_by_you'
   )
+  // deprecated only used in recent versions
```
WofWca:
```suggestion
  // deprecated only used in older versions
  strings[C.DC_STR_EPHEMERAL_TIMER_1_MINUTE_BY_YOU] = tx(
    'ephemeral_timer_1_minute_by_you'
  )
  // deprecated only used in older versions
```

"Older" maybe?

packages/frontend/src/components/dialogs/DisappearingMessages.tsx
```diff
@@ -57,39 +58,34 @@ function SelectDisappearingMessageDuration({
       />
       <Radio
         key={'eph-1'}
-        label={tx('after_1_minute')}
```
WofWca, line 60: So what if the value is already set to "1 minute" in an older version? If they upgrade, none of the radio buttons will be shown as selected. Not ideal IMO. (But also not too bad).

WofWca (commented):
Looks good, besides the comments, didn't test.

packages/frontend/src/components/dialogs/DisappearingMessages.tsx
r10s: it is fine - one needs to select a new option. the message in that chat and the selected timing will stay as is. we also do not expect these values to be used often.

r10s (approved):

WofWca (approved):

PR #5438, opened 2025-09-02
Dont close menues on resize

### LINKED ISSUE #5418: Regression: "Reply" and "React" partially broken on touch due to menus closing on window resize
There was a somewhat major regression in deltachat desktop on any touch device or hybrid device, since now most menus and popups close when the window resizes. This should as far as I know never be done on any platform, because a window resize can be automatic (rather than user directed) and for example be caused by the touch keyboard opening or closing. Window resize should ideally never cause any UI action ever.

- Operating System (Linux/Mac/Windows/iOS/Android): Linux Wayland
- Delta Chat Version: 	2.10.0  Delta Chat Core	v2.10.0
- Expected behavior: window resize doesn't close or change anything ever, unless it is guaranteed to be a user-directed window resize caused by touch drag or mouse drag. differentiating safely like this may however not be possible, so ideally resize should just never trigger any UI actions.
- Actual behavior: window resize closes items.

  1. Use touch device with touch keyboard enabled
  2. Long press message to open the menu
  3. The menu opening will likely close the touch keyboard if it was open. This triggers a window resize. This causes the menu to immediately close again.

- Screenshots:
- Logs:

nicodh:
resolves #5418 

While this PR solves the related issue it will introduce other maybe unexpected behaviours, when context menues are open and the window is resized. But since resizing is mostly triggered by the user (except in the use case the issue describes) there is no high urgency to fix or adapt the positioning.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased][unreleased]
 
+### Fixed
+- don't close context menues on window resize #5418
+
+
 <a id="2_11_1"></a>
 
 ## [2.11.1] - 2025-09-01
--- a/packages/frontend/src/components/ContextMenu.tsx
+++ b/packages/frontend/src/components/ContextMenu.tsx
@@ -374,13 +374,9 @@ export function ContextMenu(props: {
       }
     }
 
-    const onResize = () => closeCallback()
-
     document.addEventListener('keydown', onKeyDown)
-    window.addEventListener('resize', onResize)
     return () => {
       document.removeEventListener('keydown', onKeyDown)
-      window.removeEventListener('resize', onResize)
     }
   }, [openSublevels, closeCallback, expandMenu])
 


WofWca (approved):
This is a good change, i.e. as per discussions in the issue.
But I'm not sure if we're supposed to close it, because the "react" thing still closes, but it closes on scroll, and not resize (resize also causes 'scroll' event):

https://github.com/deltachat/deltachat-desktop/blob/54676b5a2da01bf007a6723be8b4e5a5558cc96b/packages/frontend/src/components/message/MessageList.tsx#L293-L296

nicodh:
> This is a good change, i.e. as per discussions in the issue. But I'm not sure if we're supposed to close it, because the "react" thing still closes, but it closes on scroll, and not resize (resize also causes 'scroll' event):

Good point. But what do you mean with " We might call `onScroll` manually with `null` argument." in the comment in the file?

WofWca:
> But what do you mean with " We might call onScroll manually with null argument." in the comment in the file?

If you look at the references to `onScroll` you'll see

https://github.com/deltachat/deltachat-desktop/blob/54676b5a2da01bf007a6723be8b4e5a5558cc96b/packages/frontend/src/components/message/MessageList.tsx#L587

The `if (ev)` is what handles such a call.

PR #5442, opened 2025-09-03
fix: a11y: delivery status being announced twice

WofWca:
The issue was that apparently NVDA first announces the label
(`aria-label`) of the changed element, and then announces
the change itself, which, in this case, is a change to
`aria-label`. This results in it simply announcing the label twice.

Let's mitigate this by removing `aria-label`.
And along the way, let's split the button
into a `role='status'` element and a "Message Info" button.

I have verified that the look of the app is the same, with WebXDC messages, info messages, and videos.

TODO:
- [x] Merge the base, up to the last commit.
  It's easier to test it after the rebase, therefore it's draft.


nicodh (approved):
Reviewed by readin g & testing

Issue #5448, opened 2025-09-06
Improve documentations for Desktop


farooqkz:
Hello.
Back when I was working on Desktop, I had trouble understanding how things work together. And what does each component or function does. This is a big blockade for first time contributors.

I'm creating this issue to improve docs for Desktop, most importantly frontend package. Scope of this issue is docstrings, comments and changes to readibility as long it doesn't change semantics(like change of variable names).

Few months ago, when I talked with Nico about this, he was positive. But then I couldn't contribute due to you-know-what issues.

For starters, I would suggest these:

 - [ ] Write docstrings for all functions
 - [ ] Write docstrings for all of our hooks
 - [ ] Use some tool or package to create a graph of the frontend. So a first timer can realize how components are arranged together.
 - [ ] Change location of ts files. Including components, utils and hooks. For instance, in some cases, it might make sense to give a group of related components a new home, under a new directory.
 - [ ] See how we can document usage and behavior of different components. Haven't researched about this. Maybe we should use docstrings for React components, too?

I'm assigned this issue to meself. As I want to take the lead. I also don't make promises about size of contributions and also delivery time.

As a rule, there is a single commit per component, hook or function. So if for whatever reason we didn't want to add docs for that entity, we could simply revert the commit and still merge the PR.

nicodh:
Her some feedback:

>  Write docstrings for all functions
> Write docstrings for all of our hooks

we don't need docstrings since we are in typescript which allows to assign types inline
types defintions can be improved in some places but not with the usage of doc strings

> Use some tool or package to create a graph of the frontend. So a first timer can realize how components are arranged together.

Can you suggest one?

> Change location of ts files. Including components, utils and hooks. For instance, in some cases, it might make sense to give a group of related components a new home, under a new directory.

That is more a refactoring than documentation. Where that makes sense and where not, needs to be discussed in detail

> See how we can document usage and behavior of different components. Haven't researched about this. Maybe we should use docstrings for React components, too?

Again doc strings are not needed. Comments should go to all components or hooks that are quite complex or use some un-intuitive approach

farooqkz:
> Her some feedback:
> 
> > Write docstrings for all functions
> > Write docstrings for all of our hooks
> 
> we don't need docstrings since we are in typescript which allows to assign types inline types defintions can be improved in some places but not with the usage of doc strings
> 
> > Use some tool or package to create a graph of the frontend. So a first timer can realize how components are arranged together.
> 
[...]

PR #5449, opened 2025-09-06
Add some dosctrings and comments

farooqkz:
Hello!
So I have added some docstrings for some functions and hooks.

As I said in #5448, there is one commit per entity(hook or function).

### DIFF
--- a/packages/frontend/src/stores/locations.ts
+++ b/packages/frontend/src/stores/locations.ts
@@ -4,18 +4,30 @@ import { onReady } from '../onready'
 import { selectedAccountId } from '../ScreenController'
 import { Store } from './store'
 
-export class state {
+interface MapSettings {
+  timestampFrom: number
+  timestampTo: number
+}
+
+export class State {
   selectedChat: Type.FullChat | null = null
-  mapSettings = {
+  mapSettings: MapSettings = {
     timestampFrom: 0,
     timestampTo: 0,
   }
   locations: T.Location[] = []
 }
 
-export const locationStore = new Store(new state(), 'location')
+export const locationStore = new Store(new State(), 'location')
 
-const getLocations = async (chatId: number, mapSettings: todo) => {
+/*
+ * Fetch all sent locations(in live location sharing) for a given chat during a given timeframe.
+ * Then store them all in the location store.
+ *
+ * @param {number} chatId A chat Id
+ * @param {MapSettings} mapSettings Specification of the timeframe start and end
+ */
+const getLocations = async (chatId: number, mapSettings: MapSettings) => {
   const { timestampFrom, timestampTo } = mapSettings
   const locations: T.Location[] = await BackendRemote.rpc.getLocations(
     selectedAccountId(),
@@ -29,6 +41,11 @@ const getLocations = async (chatId: number, mapSettings: todo) => {
   }, 'getLocations')
 }
 
+/*
+* Add callbacks for fetching locations and storing them in LocationStore
+* There are callbacks for when a location changes, a messages changes
+* or when a new message arrives.
+*/
 onReady(() => {
   BackendRemote.on('LocationChanged', (accountId, { contactId }) => {
     if (accountId !== window.__selectedAccountId) {
--- a/packages/frontend/src/stores/messagelist.ts
+++ b/packages/frontend/src/stores/messagelist.ts
@@ -52,7 +52,31 @@ const defaultState = () =>
     loaded: false,
   }) as MessageListState
 
-export function useMessageList(accountId: number, chatId: number) {
+
+// TODO: find a better name for this interface
+interface MessageListR {
+  state: MessageListState,
+  store: MessageListStore,
+  fetchMoreTop: () => void,
+  fetchMoreBottom: () => void,
+}
+
+
+/*
+ * A hook to read a portion of messages(a view) for a given chat. It creates a store(MessageListStore)
+ * for the given chat and account and loads messages on it. It always has a maximum specified number
+ * of messages as per PAGE_SIZE constant. The hook returns theocurrent message list state, the store
+ * and two callbacks to fetch more messages from either top or bottom, if any.
+ *
+ * The hook depends on SettingsStore, however, just volume is read from it. It also depends on a number of DC events:
+ *  Delivery of a message,
+ *  A new incoming message,
+ *  Change of a reaction,
+ *  Change of a message,
+ *  Change of read status of a Message
+ *  If sending a message has failed
+ */
+export function useMessageList(accountId: number, chatId: number): MessageListR {
   const store = useMemo(() => {
     const store = new MessageListStore(accountId, chatId)
     store.effect.loadChat()
@@ -160,6 +184,11 @@ export function useMessageList(accountId: number, chatId: number) {
   return { store, state, fetchMoreTop, fetchMoreBottom }
 }
 
+/*
+ * Simply returns a subarray of items, from start to end
+ *
+ * @return {T[]} The subarray from start to end
+ */
 function getView<T>(items: T[], start: number, end: number) {
   return items.slice(start, end + 1)
 }
@@ -324,8 +353,8 @@ class MessageListStore extends Store<MessageListState> {
         // Those are actual messages, but we don't render them
         this.log.warn(
           `setMessageState called for message ${messageId}, ` +
-            `state ${messageState}, but it's not loaded. ` +
-            "Ignoring, in hopes that we'll automatically load it later."
+          `state ${messageState}, but it's not loaded. ` +
+          "Ignoring, in hopes that we'll automatically load it later."
         )
         return
       }
@@ -383,7 +412,7 @@ class MessageListStore extends Store<MessageListState> {
         // FYI there is similar code in `MessageList.tsx`.
         if (
           window.__internal_jump_to_message_asap?.accountId ===
-            this.accountId &&
+          this.accountId &&
           window.__internal_jump_to_message_asap.chatId === this.chatId
         ) {
           const jumpArgs =
@@ -699,16 +728,16 @@ class MessageListStore extends Store<MessageListState> {
         last_item === undefined
           ? -1
           : messageListItems.findIndex(item => {
-              if (last_item.kind !== item.kind) {
-                return false
+            if (last_item.kind !== item.kind) {
+              return false
+            } else {
+              if (item.kind === 'message') {
+                return item.msg_id === (last_item as any).msg_id
               } else {
-                if (item.kind === 'message') {
-                  return item.msg_id === (last_item as any).msg_id
-                } else {
-                  return item.timestamp === (last_item as any).timestamp
-                }
+                return item.timestamp === (last_item as any).timestamp
               }
-            })
+            }
+          })
 
       // check if there is an intersection
       if (indexStart !== -1 && messageListItems[indexStart + 1]) {
@@ -929,10 +958,10 @@ class MessageListStore extends Store<MessageListState> {
     if (!isMessageInCurrentChat) {
       this.log.error(
         'Tried to show messages from a different chat.\n' +
-          `this.accountId === ${this.accountId}, ` +
-          `this.chatId === ${this.chatId}, ` +
-          `target IDs: ${accountId}, ${chatId}. ` +
-          `jumpToMessageId === ${jumpToMessageId}`
+        `this.accountId === ${this.accountId}, ` +
+        `this.chatId === ${this.chatId}, ` +
+        `target IDs: ${accountId}, ${chatId}. ` +
+        `jumpToMessageId === ${jumpToMessageId}`
       )
     }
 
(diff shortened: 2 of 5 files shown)


packages/frontend/src/stores/locations.ts
```diff
@@ -29,6 +41,11 @@ const getLocations = async (chatId: number, mapSettings: todo) => {
   }, 'getLocations')
 }
 
+/*
+ * Add callbacks for fetching locations and storing them in LocationStore
+ * There are callbacks for when a location changes, a messages changes
+ * or when a new message arrives.
```
WofWca: The last two lines seem redundant.

packages/frontend/src/stores/locations.ts
```diff
@@ -4,18 +4,30 @@ import { onReady } from '../onready'
...
-export const locationStore = new Store(new state(), 'location')
+export const locationStore = new Store(new State(), 'location')
 
-const getLocations = async (chatId: number, mapSettings: todo) => {
+/*
+ * Fetch all sent locations(in live location sharing) for a given chat during a given timeframe.
+ * Then store them all in the location store.
+ *
+ * @param {number} chatId A chat Id
+ * @param {MapSettings} mapSettings Specification of the timeframe start and end
```
WofWca: The type annotations don't need to be duplicated. The same goes for all other changes in this MR.

packages/frontend/src/stores/locations.ts
```diff
@@ -4,18 +4,30 @@ import { onReady } from '../onready'
...
 
-export const locationStore = new Store(new state(), 'location')
+export const locationStore = new Store(new State(), 'location')
 
-const getLocations = async (chatId: number, mapSettings: todo) => {
+/*
+ * Fetch all sent locations(in live location sharing) for a given chat during a given timeframe.
+ * Then store them all in the location store.
+ *
+ * @param {number} chatId A chat Id
```
WofWca: IMO documenting that "chatId is a chat Id" is pointless.

packages/frontend/src/stores/messagelist.ts
```diff
@@ -52,7 +52,32 @@ const defaultState = () =>
...
+  store: MessageListStore
+  fetchMoreTop: () => void
+  fetchMoreBottom: () => void
+}
+
+/*
+ * A hook to read a portion of messages(a view) for a given chat. It creates a store(MessageListStore)
+ * for the given chat and account and loads messages on it. It always has a maximum specified number
+ * of messages as per PAGE_SIZE constant. The hook returns theocurrent message list state, the store
+ * and two callbacks to fetch more messages from either top or bottom, if any.
```
WofWca: > The hook returns theocurrent message list state, the store and two callbacks to fetch more messages from either top or bottom, if any.

This information is clear from the return type of the hook.

packages/frontend/src/stores/messagelist.ts
```diff
@@ -52,7 +52,32 @@ const defaultState = () =>
     loaded: false,
   }) as MessageListState
 
-export function useMessageList(accountId: number, chatId: number) {
+// TODO: find a better name for this interface
+interface MessageListR {
```
WofWca: This type can be inlined IMO.

packages/frontend/src/stores/messagelist.ts
```diff
@@ -52,7 +52,32 @@ const defaultState = () =>
...
+ * of messages as per PAGE_SIZE constant. The hook returns theocurrent message list state, the store
+ * and two callbacks to fetch more messages from either top or bottom, if any.
+ *
+ * The hook depends on SettingsStore, however, just volume is read from it. It also depends on a number of DC events:
+ *  Delivery of a message,
+ *  A new incoming message,
+ *  Change of a reaction,
+ *  Change of a message,
+ *  Change of read status of a Message
+ *  If sending a message has failed
```
WofWca: Honestly I don't see how reading this is better than just reading the body of the function. _and_ it would also need to be updated every time you change the body of the function.

packages/frontend/src/stores/messagelist.ts
```diff
@@ -160,6 +185,11 @@ export function useMessageList(accountId: number, chatId: number) {
   return { store, state, fetchMoreTop, fetchMoreBottom }
 }
 
+/*
+ * Simply returns a subarray of items, from start to end
+ *
+ * @return {T[]} The subarray from start to end
+ */
 function getView<T>(items: T[], start: number, end: number) {
```
```suggestion
 * Simply returns a subarray of items, from start to end
 */
function getView<T>(items: T[], start: number, end: number): T[] {
```

packages/frontend/src/utils/getDirection.ts
```diff
@@ -1,6 +1,16 @@
 import { C } from '@deltachat/jsonrpc-client'
 
 // TODO: move to frontend
-export function getDirection({ fromId }: { fromId: number }) {
+/*
+ * For a given contact ID, is the message to this account or to another account?
+ * In other words, for a given contact ID, is it id of self or another contact?
+ *
+ * @return "outgoing" | "incoming"
```
WofWca:
```suggestion
 * Is this a message that we sent out, or someone sent to us?
```

packages/frontend/src/stores/locations.ts
farooqkz: I agree. But docstrings require this. Do you have suggestion for an alternative?

packages/frontend/src/stores/locations.ts
farooqkz: I agree

farooqkz:
> Honestly this just seems like documenting for the sake of documenting. Some of the additions seem useful, but most are not

You might be right. I have fallen in the trap :)

BTW, you guys now work on weekend? :)

packages/frontend/src/stores/locations.ts
WofWca: Umm where did you get the info that they require this?

packages/frontend/src/stores/locations.ts
farooqkz: oh man how did I do this? I completely agree!

WofWca:
> BTW, you guys now work on weekend? :)

Always has been

farooqkz:
> I agree with WofWcas comments. Explaining comments should only be applied if there is an "un-intuitive" or unexpected specialty/hack or workaround or if it is helpful to understand some complexity
> 
> We almost don't need doc strings since we are in typescript.

My goal is that when I hover on a component, function or hook in me IDE, I get a description what does this thing do. I think we need docstrings for it unless it is obvious from the entity name.

> My 👍 was for the general idea to improve docs & comments which is appreciated but we should always take care for a good balance between too much and too few (which of course can be discussed in reviews)

I agree about the balance.

packages/frontend/src/stores/messagelist.ts
farooqkz: Hmm I didn't know I can omit types in docstrings. Will commit it

packages/frontend/src/stores/locations.ts
farooqkz: My imagination xD

packages/frontend/src/stores/messagelist.ts
farooqkz: I think it is helpful if you have a docstring or something which explains what the hook depends on. But I highly agree that manually writing this is very bad. I think an automated tool would help? If you agree about their utility in dosctring, however you think adding them manually will cause more harm in future than good, I can remove them. Actually I kinda have the same opinion. But I'm not very sure.

packages/frontend/src/stores/messagelist.ts
farooqkz: Let's ask for another vote. I'm not really sure. Maybe yes, maybe no.

farooqkz:
I think I learned a valuable lesson here. I did a very good job by submitting a small PR/patch, instead of considering that Nico has agreed and make an extensive PR. I will try to document things we need in #5448 and try to move slow to clearly see the way ahead. Call it Wisedom (TM)!

packages/frontend/src/stores/messagelist.ts
nicodh: In general it can help to have a readable return type. But if there is no need to reuse an interface (like when we export it to another file) there is no reason to define it as separate type imo. I consider the inline variant very readable.
Btw: that is what I see in vscode when hovering over the function name, even when there is no return type defined explicit. It's just derived from the code 

```typescript
export function useMessageList(accountId: number, chatId: number): {
  store: MessageListStore,
  state: MessageListState,
  fetchMoreTop: () => void,
  fetchMoreBottom: () => void
} {

```

Simon-Laux:
some people say that code should be so self explanatory that you need no comments.
IMO that is just an ideal, sometimes unreachable in a reasonable time, so I agree with what nico says.
I think there is an exception though, where it always makes sense to add doc comments, and that are library and utility functions, especially those ones that are exported, because there are sometimes cases, where you can not jump to the definition of a method. (for example you have a typescript package that got compiled to ts and when you jump to symbols, then you land in the compiled `.d.ts` type definitions that don't include the source code)

farooqkz:
> some people say that code should be so self explanatory that you need no comments. IMO that is just an ideal, sometimes unreachable in a reasonable time, so I agree with what nico says. I think there is an exception though, where it always makes sense to add doc comments, and that are library and utility functions, especially those ones that are exported, because there are sometimes cases, where you can not jump to the definition of a method. (for example you have a typescript package that got compiled to ts and when you jump to symbols, then you land in the compiled `.d.ts` type definitions that don't include the source code)

So you think we shouldn't add docstrings to hooks and components?

A good idea might be adding docs only to reusable components such as `Button`

Simon-Laux:
> So you think we shouldn't add docstrings to hooks and components?

if they require explanation then they need a doc-string, otherwise they do not. Writing and keeping doc strings up-to-date is extra work, so it needs to provide enough extra value.

farooqkz:
> > So you think we shouldn't add docstrings to hooks and components?
> 
> if they require explanation then they need a doc-string, otherwise they do not. Writing and keeping doc strings up-to-date is extra work, so it needs to provide enough extra value.

Of course. I meant adding docstrings for those which need it.

packages/frontend/src/stores/messagelist.ts
farooqkz: Can we decide on this? The more I'm thinking to it, the more I think we should remove it.

packages/frontend/src/stores/locations.ts
```diff
@@ -4,18 +4,27 @@ import { onReady } from '../onready'
 import { selectedAccountId } from '../ScreenController'
 import { Store } from './store'
 
-export class state {
+interface MapSettings {
+  timestampFrom: number
+  timestampTo: number
+}
+
```
nicodh, line 11: No need to define an interface here. Inline type defintion is sufficient

packages/frontend/src/utils/copyToBlobDir.ts
```diff
@@ -1,6 +1,12 @@
 import { BackendRemote } from '../backend-com'
 
-// Copy a file to the blob directory
+/*
+ * Copy the file from `filepath` to the blob directory and create a blob.
+ * Throws an error if no account is selected.
+ *
+ * @param {string} Path to the file
+ * @return {Promise<string>} A promise which resolves to the blob path
```
nicodh: we don't need jsdoc strings since we have the types already defined inline

packages/frontend/src/utils/lastUsedPaths.ts
```diff
@@ -22,6 +22,13 @@ const defaultLocations: {
...
 }
 
+/*
+ * What path last has been used for this data? Return a default path if
+ * no path has been used.
+ *
+ * @param {string} key The data slot you are looking for.
+ * @return {Promise<{ defaultPath: string, setLastPath: (lastPath: string) : void}>} The default path as well as a callback to set the last path.
+ */
 export async function rememberLastUsedPath(key: LastUsedSlot) {
```
nicodh, line 28: again jsdoc strings are not needed. You should see the return type when hovering over the function name

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 
 ### Changed
 - update `@deltachat/stdio-rpc-server` and `deltachat/jsonrpc-client` to `2.12.0`
+- Add some docstrings for `useMessageList` hook, `getDirection` function, `copyToBlobDir` function and `locations.ts` all in frontend package
```
nicodh: I would not mention that in the changelog. It is only for devs and even they not "neccessarily" need to know what has changed

use #skip-changelog instead

nicodh (changes requested):
Please be aware of a good resource management. This PR has 26 comments and 16 commits for roughly 67 lines of added comments.

I would appreciate to focus on "really missing" docs - like stuff where you remember it was really hard to understand just by reading the code.
:-)

farooqkz:
Let's get this PR merged. I will make sure to not make PR to this repository ever again, since DeltaChat resources are so scarce and outside contributors have to learn proper resource management.

packages/frontend/src/stores/locations.ts
farooqkz: you sure? it is also used in line 14

nicodh: Ok, that could also be typed by state['mapSettings'] but it's maybe better to define an interface then!

packages/frontend/src/stores/messagelist.ts
farooqkz: done

farooqkz:
hints on why is it failing?

nicodh:
Not related to the changes of this PR. We have a CI server which is a bit flaky currently

PR #5450, opened 2025-09-06
Add SKIP_FUSES env var

nicodh:
If SKIP_FUSES = true we skip the flip fuses script.

This helps building nixOS packages. See comments in #5423 

#skip-changelog


WofWca (approved):
This looks good.
But it's not actually used in any of our build scripts, further work is to be done?

nicodh:
> This looks good. But it's not actually used in any of our build scripts, further work is to be done?

I assumed that will be set in the nixOS build script environment to avoid the reported error in https://github.com/deltachat/deltachat-desktop/pull/5423#issuecomment-3243039424

PR #5451, opened 2025-09-06
fix: remove RTCPeerConnection exhaustion (FILL500)

WofWca:
This partially reverts 750721ca5b8e4e9c1c376c95cec470eefefd474c
(https://github.com/deltachat/deltachat-desktop/pull/3098).
Not completely though: the `<iframe>` thing is still there,
and I have not removed it yet for the sake of being conservative.

The most important point of this change is to prepare ourselves
for the removal of the `IsolateSandboxedIframes` Chromium feature.
It has been enabled by default, and is to be removed soon-ish:
https://issues.chromium.org/issues/362246896.

Earlier we simply un-enabled it in
50bcd88a80cfb6a86e30efae4f723ab46f3d08aa
(https://github.com/deltachat/deltachat-desktop/pull/4351),
but when it's removed, we will not be able to.
We could hold off on upgrading Electron then, but let's not.

Another thing is, of course, the fact that this speeds up
the process of launching an app.

And in general this approach seems more thorough to me.
It also provides another exfiltration security layer (a dummy proxy).

This uses the same approach that we took in the Tauri version.

On top of that, the 500 limit itself is not something
that is set in stone and should be relied on.

To check that this still works, follow the instructions in
https://github.com/webxdc/webxdc-test/pull/40/files.

### DIFF
--- a/packages/target-electron/src/deltachat/webxdc.ts
+++ b/packages/target-electron/src/deltachat/webxdc.ts
@@ -39,6 +39,7 @@ import {
 import { T } from '@deltachat/jsonrpc-client'
 import type * as Jsonrpc from '@deltachat/jsonrpc-client'
 import { setContentProtection } from '../content-protection.js'
+import { Server } from 'net'
 
 const __dirname = dirname(fileURLToPath(import.meta.url))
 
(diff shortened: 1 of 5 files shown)


WofWca:
Rebased.
This should be good to merge when https://github.com/deltachat/deltachat-desktop/pull/5455 is merged.

WofWca:
Rebased, and adjusted comments a little.

I also re-tested this with the instructions above. All works well.

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -232,6 +311,18 @@ export default class DCWebxdc {
         alwaysOnTop: main_window?.isAlwaysOnTop(),
         show: false,
       })
+      // Settings this should make WebRTC always use the proxy.
+      // However, since the proxy won't work, this should, in theory,
+      // effectively disable WebRTC.
```
WofWca, line 316:
```suggestion
      // Settings this should make WebRTC always use the proxy.
      // However, since the proxy won't work, this should, in theory,
      // effectively disable WebRTC.
      // See https://www.rfc-editor.org/rfc/rfc8828#name-modes-and-recommendations
      // > If the proxy does not support UDP
      // > (as is the case for all HTTP and most SOCKS [RFC1928] proxies),
      // > or the WebRTC implementation does not support UDP proxying,
      // > the use of UDP will be disabled,
      // > and TCP will be used to send and receive media through the proxy.
```

PR #5454, opened 2025-09-08
Add new maps:// protocol

nicodh:
- add new maps.xdc (for testing)

This is a proposal to allow multiple services to fetch tiles and layers in the maps.xdc

Since all requests are piped through the core there is no need anymore to add EXCLUDE to the hostRules.

BUT: now any webxdc that has internet_access can use the maps:// protocol to access any URL in the internet. Before it was only openstreetmap.org that was accessible.

TBD:
1. do we need additional restrictions?
2. should we inform the user that any webxdc app that has request_integration='map' replaces the original maps.xdc and can also access any url in the internet?
3. should we maybe ask the user per URL to confirm the internet access the first time the app connects to a new url?

Comments:
For Android/iOS it is already now the case, that an integrated app (which is only maps for now) can access any url in the internet. On the other hand we communicate that webxdc apps can't access the internet. And an app with request_integration='map' still needs to be saved in the self chat to really have access.

Nevertheless I would vote to implement 3. which is more explicit and only happens rarely.

#skip-changelog (as this is just an experimental proposal)


packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -177,7 +175,10 @@ export default class DCWebxdc {
         const ses = sessionFromAccountId(accountId)
         accounts_sessions.push(accountId)
         ses.protocol.handle('webxdc', (...args) =>
-          webxdcProtocolHandler(this.rpc, ...args)
+          webxdcProtocolHandler(this.rpc, appId, ...args)
+        )
+        ses.protocol.handle('maps', (...args) =>
+          mapProtocolHandler(this.rpc, appId, accountId, ...args)
```
WofWca: This reintroduces the bug with all apps using the same `appId`. See git blame.
Maybe I should have written it in a more future-proof way.

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -840,36 +900,12 @@ async function webxdcProtocolHandler(
...
-  ) => {
-    const headers = new Headers()
-    if (!open_apps[id].internet_access) {
-      headers.append('Content-Security-Policy', CSP)
-    }
-    // Ensure that the client doesn't try to interpret a file as
-    // one with 'application/pdf' mime type and therefore open it
-    // in the PDF viewer, see
-    // "XDC-01-005 WP1: Full CSP bypass via desktop app PDF embed"
-    // https://public.opentech.fund/documents/XDC-01-report_2_1.pdf
```
WofWca, line 856: Any reason to remove this comment?

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -826,8 +827,67 @@ export default class DCWebxdc {
...
+  try {
+    const response = await rpc.getHttpResponse(
+      accountId,
+      request.url.replace('maps://', 'https://')
+    )
+    const blob = Buffer.from(response.blob, 'base64')
+    return makeResponse(appId, blob, {}, mimeType)
+  } catch (error) {
+    log.error('map: load blob:', error)
+    return makeResponse(appId, '', { status: 404 })
```
WofWca: Nitpick: 404 doesn't seem appropriate, at least not always.

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -177,7 +175,10 @@ export default class DCWebxdc {
         const ses = sessionFromAccountId(accountId)
         accounts_sessions.push(accountId)
         ses.protocol.handle('webxdc', (...args) =>
-          webxdcProtocolHandler(this.rpc, ...args)
+          webxdcProtocolHandler(this.rpc, appId, ...args)
+        )
+        ses.protocol.handle('maps', (...args) =>
```
WofWca: So, can't we ensure that only the maps .xdc can access this protocol, by making a separate `ses`?
Or there should be another way.

Otherwise this is a pretty concerning attack vector.

WofWca:
Overall it's nice that requests are passed through core, i.e. they can use the same proxy.

packages/target-electron/src/deltachat/webxdc.ts
nicodh: Ok I see, there is no app specific session but only an account specific. Yes then it makes sense to have a separate session for the maps.xcd

packages/target-electron/src/deltachat/webxdc.ts
nicodh: Maybe add some comment here: the first created sesssion is reused for all other  apps that will be opened by the same accountId so don't rely on any app or webxdcInfo specific stuff here.

I wonder why we need at all have one session per account and not per app.

WofWca: > I wonder why we need at all have one session per account and not per app.

Probably because of `delete_webxdc_account_data`

https://github.com/deltachat/deltachat-desktop/blob/b352801fb8d2e816e364aadced44235299a6baa0/packages/target-electron/src/deltachat/webxdc.ts#L1064-L1079

Simon-Laux: tauri edition uses a different session for each app.

see packages/target-tauri/src-tauri/src/webxdc/data_storage.rs

I woudn't change it in electron for normal webxdc apps, unless we also want to add a migration step, but that sounds like too much extra complexity for unknown gain. Note that mobile probably doesn't even use any separate sessions/data files.

maybe it could make sense to add it to iOS at some point, now that we know how to use the api for it: https://developer.apple.com/documentation/webkit/wkwebsitedatastore (but again I don't know how big the risk of not doing that really is, and migration steps could also introduce bugs)

CC @r10s

packages/target-electron/src/deltachat/webxdc.ts
Simon-Laux: we could add a comment for now, if we want something better we'd need to change the core api to also return the real response code.

packages/target-electron/src/deltachat/webxdc.ts
nicodh: Readded it

packages/target-electron/src/deltachat/webxdc.ts
nicodh: I choosed a different approach: the maps:// protocol is now restricted by CSP: only apps with internet access are now allowed to add image sources with maps://* as target.

I tried the separate sessions approach but somehow the partition persist:webxdc_<accountId>.<appId> resulted in errors. Maybe there is some magic relying on the format persist:webxdc_<accountId> ??
Tried it here: #5455

nicodh:
Closed in favour of #5455

PR #5455, opened 2025-09-10
Maps with own session (WIP)

nicodh:
#skip-changelog


packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -959,10 +1030,33 @@ function sessionFromAccountId(accountId: number) {
...
 }
 
+/**
+ * create a separate session for each app that has internet access
+ */
+function sessionFromAppId(appId: string) {
+  // const partitionKey = `persist:webxdc_${appId.split('.')[0]}`
+  // this results in an error, it only works
+  // with the accountId (can be tested with the line above)
+  const partitionKey = `persist:webxdc_${appId}`
```
nicodh: This results in a somehow different session (type?):
persist: webxdc_2.123 doesn't work
persist:webxdc_2 works

The maps.xdc doesn't even load and the last log entry is a  permission denied for "openExternal" , which is a permission taht is not requested if the pure account number is used

WofWca: You forgot to use `sessionFromAppId` instead of `partitionFromAccountId` when initializing `BrowserWindow`.

nicodh: Ah thanks! That was the missing part.

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -58,9 +55,13 @@ const open_apps: {
   [instanceId: string]: AppInstance
 } = {}
 
-// holds all accounts which have a session with webxdc scheme registered
+// holds all accounts which have a session with webxdc
+// scheme registered except those with internet access
 const accounts_sessions: number[] = []
 
+// holds all accounts which have internet access
```
Simon-Laux: confusing/unclear comment

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -826,50 +860,120 @@ export default class DCWebxdc {
...
+    // to avoid CORS issues and to be able to cache tiles
+    const response = await rpc.getHttpResponse(
+      accountId,
+      request.url.replace('maps://', 'https://')
+    )
+    const blob = Buffer.from(response.blob, 'base64')
+    return makeResponse(null, blob, {}, mimeType)
+  } catch (error) {
+    log.error('map: load blob:', error)
+    return makeResponse(null, '', { status: 404 })
```
Simon-Laux: nitpick/correctness: we don't know exact error code, could be temporary error, not sure if leaflet uses it / if it makes a difference in this case.
I believe we should make a new core method that also returns the error code, so that we can return the correct error code here.

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -171,21 +173,43 @@ export default class DCWebxdc {
...
+      // register appropriate protocols
+      // see https://www.electronjs.org/docs/latest/api/protocol
+      if (webxdcInfo['internetAccess']) {
+        if (!sessions_with_internet_access.includes(appId)) {
+          const ses = sessionFromId(appId)
+          partition = partitionFromId(appId)
+          sessions_with_internet_access.push(appId)
+          ses.protocol.handle('webxdc', (...args) => {
+            return webxdcProtocolHandler(this.rpc, ...args)
+          })
```
Simon-Laux: nitpick: this is code appears again below (in line 203), maybe code can be restructured in a way that the code looks less duplicated?

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -826,50 +860,120 @@ export default class DCWebxdc {
...
+}
+
+/**
+ * the webxdc:// protocol is used to fetch the webxdc app files
+ * and any additional blobs (images, sounds, etc.) that are
+ * part of the webxdc package.
+ * All responses have the appropriate CSP headers set (using makeResponse)
+ *
+ * Instead of a domain we use accountId.msgId as hostname
+ * so we can identify each request from the app and check permissions
```
Simon-Laux, line 982: The other and bigger reason for this hostname format is to get different localstorage/indexdb/cookies instances for each app, so that an app can not access/change data of other apps.

But I wonder if this is the correct place to document it.

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -949,52 +1054,83 @@ function makeTitle(webxdcInfo: T.WebxdcMessageInfo, chatName: string): string {
...
+  return `persist:webxdc_${id}`
 }
 
-function sessionFromAccountId(accountId: number) {
-  return session.fromPartition(partitionFromAccountId(accountId), {
+function sessionFromId(id: number | string) {
+  return session.fromPartition(partitionFromId(id), {
     cache: false,
   })
 }
```
Simon-Laux: I think it would be more clear to have separate methods for "get partition from account id" and "get partition for app id".

But I also don't quite understand why each internet enabled app needs it's own session, wouldn't it be enough to just have two sessions per account? one with internet and one without?

we could give the maps scheme to all internet enabled webxdcs, internet enabled webxdcs won't work anyway because of the DNS block in chromium...
if we want them to work, then we could think about overwriting the "https" scheme with the maps scheme handler if that is possible instead of creating a new maps scheme.

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -949,52 +1054,83 @@ function makeTitle(webxdcInfo: T.WebxdcMessageInfo, chatName: string): string {
...
-  await session.clearStorageData()
-  await session.clearData()
+  const sessions = getAllSessionsForAccount(accountId)
+  for (const s of sessions) {
+    await s.clearStorageData()
+    await s.clearData()
+  }
 })
 
 ipcMain.handle('webxdc.getWebxdcDiskUsage', async (_, accountId: number) => {
```
Simon-Laux: this method is unused, not sure what we should do with it.

1. we don't have a storage usage screen yet
2. idea: we could add it to the account size that is shown when you are about to delete an account
3. or we could delete it and bring it back later, it is saved in the git history after all

(well you can call it from the dev console in `--devmode` for debugging purposes, but I'm not sure that counts: `await exp.runtime.getWebxdcDiskUsage()`)
image

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -949,52 +1054,83 @@ function makeTitle(webxdcInfo: T.WebxdcMessageInfo, chatName: string): string {
...
+  return session.fromPartition(partitionFromId(id), {
     cache: false,
   })
 }
 
+/**
+ * helper function to get all sessions for a given accountId
+ * for cleanup and removal of data
+ */
+function getAllSessionsForAccount(accountId: number) {
```
Simon-Laux, line 1100: This only works if the user has already opened the sessions before deleting the account, so currently this would introduce a bug where the webxdc data would stay when an account is deleted unless the user opened a webxdc and the map on that account before deleting it (in same "usage session" as the deletion).

packages/target-electron/src/deltachat/webxdc.ts
nicodh: As long as there is no "concept" how to handle internet access for webxdc apps we should keep it like that imo.

Webxdc apps have no internet access and we only have this single exception for an integrated app. I think it should be more obvious visible for the user if an integrated webxdc app has internet access and there are probably more questions around a "generic" solution.

packages/target-electron/src/deltachat/webxdc.ts
nicodh: I saw that it is unused but didn't want to add unrelated changes. On the other hand I had to rewrite the function and so I'm also fine with deleting it for now

nicodh:
Update: I reduced the sessions to just 2 types of sessions which also simplifies the cleanup process and added a reduced version of maps.xdc to avoid additional dependant decisions https://github.com/deltachat/maps/pull/7

packages/target-electron/src/deltachat/webxdc.ts
Simon-Laux: latest changes resolves this issue

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -826,50 +861,118 @@ export default class DCWebxdc {
...
+    filename = filename.substring(0, filename.length - 1)
+  }
+  if (filename.startsWith('/')) {
+    filename = filename.substring(1)
   }
 
+  const mimeType: string | undefined = Mime.lookup(filename) || ''
+  try {
+    // now we pipe the real request through our backend
+    // to avoid CORS issues and to be able to cache tiles
```
```suggestion
    // to bypass CORS restrictions and to be able to cache tiles
```

Why do we need to bypass them? Don't the tile services add `Access-Control-Allow-Origin: *`? If not, why?

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -826,50 +861,118 @@ export default class DCWebxdc {
...
+ * that are important for security, namely `Content-Security-Policy`.
+ * Failing to set CSP might result in the app being able to create
+ * an <iframe> with no CSP, e.g. `<iframe src="/no_such_file.lol">`
+ * within which they can then do whatever
+ * through the parent frame, see
+ * "XDC-01-002 WP1: Full CSP bypass via desktop app webxdc.js"
+ * https://public.opentech.fund/documents/XDC-01-report_2_1.pdf
+ */
+const makeResponse = (
+  appId: string | null,
```
WofWca: Is there not a less confusing way to specify if we want lax CSP? Why is `appId` sometimes provided and sometimes not? I now see why, but it's not immediately clear. I suggest to instead add an `options` object with an optional `cspAllowMapsImgSrc`.

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -58,8 +55,12 @@ const open_apps: {
   [instanceId: string]: AppInstance
 } = {}
 
-// holds all accounts which have a session with webxdc scheme registered
-const accounts_sessions: number[] = []
+// holds all partitionKeys that have a session created for webxdc apps
+// mainly to avoid creating multiple sessions for the same partition
```
WofWca: Saying "mainly" seems to imply that there is another reason?
Also if this is talking about `session.fromPartition` then, according to its docs, it still returns the same `Session` object if it exists. Plus we don't actually deal with `Session` objects much, we only operate partitions.

If you don't want to deal with documenting existing code as part of this MR, I suggest to remove this line (sentence).

static/xdcs/version.txt
```diff
@@ -0,0 +1 @@
+Based on https://github.com/deltachat/maps/pull/7
```
WofWca, line 1: This appears to be false? It's https://github.com/deltachat/maps/pull/6 instead?

```suggestion
Based on https://github.com/deltachat/maps/pull/6
```

packages/target-electron/src/deltachat/webxdc.ts
nicodh: > Saying "mainly" seems to imply that there is another reason?

If a session already exists and you create it with the same partitionKey you get the existing session but with the options passen to the first creation, so you can't "override" options. That's the main reason.
I reworded to 
"to make sure each session is only created once per partitionKey"

static/xdcs/version.txt
nicodh: Hmm, do you see the map with a "settings" icon or without?

Based on https://github.com/deltachat/maps/pull/7 is a reduced version (looks same as the one we have)

WofWca: It is with the settings icon, where I can change map providers.

nicodh: And you don't have it in Saved messages?

nicodh: Just tried it with a fresh account: the "shipped" map is the one without settings

nicodh:
> Do we want to support third-party maps WebXDC apps? If yes, then this MR will break them, and require that they explicitly support Delta Chat Desktop. If not, then I'd suggest to have a white-list of map service domains.

The main idea is to have an easy way to extend the map sources (i.e. to extend the mapServices list in the version with add-ons) and yes these can be added to Saved messages and replace the intergrated maps.xdc

> Has https://www.electronjs.org/docs/latest/api/web-request API been considered? i.e. intercept `https` requests.

We could got for that if we have a clear concept of use cases for this "integrated apps have Internet access"
I still don't feel comfortable with the fact that we 
1. claim that webxdc apps have no internet access
2. allow to replace the integrated app with a provided webxdc app
3. integrated apps have internet access

A more generic approach would be to ask the user once if the replaced & integrated app accesses an external URL and remember the answer

static/xdcs/version.txt
WofWca: Oh. Sorry. I checked on a new account, and indeed there is no "Settings" button.

Looks like the "maps" feature is a mess. It appears to be using the WebXDC file from some hidden message that was sent in September, probably when I first tried this MR.

WofWca: I mean to say that the version is different on different accounts, even if I don't have any maps `.xdc`s in "Saved Messages".
Thus, on an even older accounts, the maps app does not load tiles at all (because it doesn't know that it needs to use the `maps:` protocol.

WofWca:
> yes these can be added to Saved messages and replace the intergrated maps.xdc

OK, so are we willing to break existing maps apps with this? Looking at my comment above, https://github.com/deltachat/deltachat-desktop/pull/5455#discussion_r2482079436, it appears that this `maps://` approach will break the feature for all existing users.

nicodh:
> > yes these can be added to Saved messages and replace the intergrated maps.xdc
> 
> OK, so are we willing to break existing maps apps with this? Looking at my comment above, [#5455 (comment)](https://github.com/deltachat/deltachat-desktop/pull/5455#discussion_r2482079436), it appears that this `maps://` approach will break the feature for all existing users.

That's what "experimental" means. Breaking changes can occur.

WofWca:
So do we then wait for the respective maps MR (https://github.com/deltachat/maps/pull/7) to get merged? Or for DC Android / iOS client devs to confirm that they're OK with this? Or does it all just not matter?

r10s:
to sum up, what i got from one to one discussions with @nico, i think, it is fine.

at a glance, imu, it offers to `maps:` protocol to be used on desktop where `http:` protocol is not working for $reasons

- in contrast to webxdc, the maps integration is heavily bounded to the main app and there may be changes at every time we change the sendUpdate() payloads. that one needs to you `maps:` does not add much on that

- i suspect that there are many integrations used in the wild. i know only from experiments as that from china user - which was actually the reason for that change, and imu, all that is build in the main app now

- if there are other, more advanced adaptions, they will break, that is true - but that is also easy to adapt. replacing integration was always for developers (as us) to be able to adapt and test things, it was not for the end user 

- in the past, the difference between android/ios and desktop was even worse as there was no way at all to use a different server than the default

- from the view of android/ios it is fine as well as the maps.xdc will continue running there using http: - at some point, it is worth checking if routing things over maps: is possible on android/ios as well - that way, it would also get eg. proxy support

- this PR does not offer more rights to the maps integrations as initially intended, thing is that this was not working on desktop before

- it is experimental :)

static/xdcs/version.txt
WofWca: Following the discussion below, from https://github.com/deltachat/deltachat-desktop/pull/5455#issuecomment-3473982768

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -828,52 +867,142 @@ export default class DCWebxdc {
...
+  const headers = new Headers()
+  if (cspAllowMapsImgSrc) {
+    /**
+     * for apps with internet access (only maps.xdc for now)
+     * we need to allow the maps://* protocol as image src in CSP
+     * the maps protocolHandler fetches the tiles through our backend getHttpResponse
+     *
+     * note that the extended CSP is only applied to the responses of the
+     * webxdc protocol (only that can provide the appId for each request)
+     */
```
WofWca: `appId` is basically unused now, and the comment is outdated.

PR #5456, opened 2025-09-11
Add new test config option chatmail

nicodh:
so we can run all tests that are suitable one time with chatmail accounts and one time with non chatmail accounts

#skip-changelog

### DIFF
--- a/packages/e2e-tests/playwright-helper.ts
+++ b/packages/e2e-tests/playwright-helper.ts
@@ -1,4 +1,4 @@
-import { expect, Page } from '@playwright/test'
+import { expect, test as base, Page } from '@playwright/test'
 
 const { config } = await import('dotenv')
 config()
@@ -19,6 +19,18 @@ export type User = {
   address: string
   password?: string
 }
+
+export type TestOptions = {
+  chatmail: boolean
+}
+
+/**
+ * extend all tests with chatmail option
+ */
+export const test = base.extend<TestOptions>({
+  // can be overriden in the config.
+  chatmail: [true, { option: true }],
+})
 
 export async function reloadPage(page: Page): Promise<void> {
   await page.goto('https://localhost:3000/')
@@ -80,7 +92,7 @@ export async function createUser(
   expect(user.id).toBeDefined()
 
   existingProfiles.push(user)
-  console.log(`User ${user.name} wurde angelegt!`, user)
+  console.log(`User ${user.name} was created!`, user)
   return user
 }
 
@@ -249,7 +261,7 @@ export async function deleteAllProfiles(
   existingProfiles: User[]
 ): Promise<void> {
   if (existingProfiles.length < 1) {
-    throw new Error('Not existing profiles to delete!')
+    console.log('No existing profiles to delete!')
   }
   for (let i = 0; i < existingProfiles.length; i++) {
     const profileToDelete = existingProfiles[i]
--- a/packages/e2e-tests/playwright-helper.ts
+++ b/packages/e2e-tests/playwright-helper.ts
@@ -21,7 +21,7 @@ export type User = {
 }
 
 export type TestOptions = {
-  chatmail: boolean
+  isChatmail: boolean
 }
 
 /**
@@ -29,7 +29,7 @@ export type TestOptions = {
  */
 export const test = base.extend<TestOptions>({
   // can be overriden in the config.
-  chatmail: [true, { option: true }],
+  isChatmail: [true, { option: true }],
 })
 
 export async function reloadPage(page: Page): Promise<void> {
--- a/packages/e2e-tests/tests/basic-tests.spec.ts
+++ b/packages/e2e-tests/tests/basic-tests.spec.ts
@@ -1,4 +1,4 @@
-import { test, expect, type Page } from '@playwright/test'
+import { expect, type Page } from '@playwright/test'
 
 import {
   getUser,
@@ -10,6 +10,7 @@ import {
   clickThroughTestIds,
   reloadPage,
   sendMessage,
+  test,
 } from '../playwright-helper'
 
 /**
@@ -35,7 +36,12 @@ const numberOfProfiles = 2
 // https://playwright.dev/docs/next/test-retries#reuse-single-page-between-tests
 let page: Page
 
-test.beforeAll(async ({ browser }) => {
+test.beforeAll(async ({ browser, chatmail }) => {
+  // Determine which type of profiles to use based on test context
+  console.log(
+    `Running tests with ${chatmail ? 'chatmail' : 'plain email'} profiles`
+  )
+
   const contextForProfiles = await browser.newContext()
   const pageForProfiles = await contextForProfiles.newPage()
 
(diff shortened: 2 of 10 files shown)


packages/e2e-tests/tests/unencrypted-group-tests.spec.ts
nicodh, line 1: I'm not sure if we still need this file now. Now that we can run unencrypted tests on non chatmail accounts.

nicodh:
Seems now there are server errors again.
https://github.com/deltachat/deltachat-desktop/actions/runs/17665670883/job/50206753028?pr=5456#step:7:10181

[WebServer] raw response:
[WebServer] <!doctype html>
[WebServer] <html lang=en>
[WebServer] <title>500 Internal Server Error</title>
[WebServer] <h1>Internal Server Error</h1>
[WebServer] <p>The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.</p>

packages/e2e-tests/playwright-helper.ts
```diff
@@ -20,6 +20,18 @@ export type User = {
...
+export type TestOptions = {
+  chatmail: boolean
+}
+
+/**
+ * extend all tests with chatmail option
+ */
+export const test = base.extend<TestOptions>({
+  // can be overriden in the config.
+  chatmail: [true, { option: true }],
```
WofWca: `isChatmail` should make it more clear that it's a bool wherever it's used in the code.
But `chatmail` is also fine.

packages/e2e-tests/tests/basic-tests.spec.ts
```diff
@@ -35,7 +36,12 @@ const numberOfProfiles = 2
 // https://playwright.dev/docs/next/test-retries#reuse-single-page-between-tests
 let page: Page
 
-test.beforeAll(async ({ browser }) => {
+test.beforeAll(async ({ browser, chatmail }) => {
+  // Determine which type of profiles to use based on test context
```
WofWca: Outdated comment?

packages/e2e-tests/tests/chat-list-multiselect.spec.ts
```diff
@@ -30,10 +31,14 @@ const expectSelectedChats = async (chatNums: number[]) => {
...
 
-test.beforeAll(async ({ browser }) => {
+test.beforeAll(async ({ browser, chatmail }) => {
   const contextForProfileCreation = await browser.newContext()
   const pageForProfileCreation = await contextForProfileCreation.newPage()
 
+  console.log(
+    `Running multiselect tests with ${chatmail ? 'chatmail' : 'plain email'} profiles`
+  )
+
```
WofWca: Are these logs useful? Why is it added in some files but not others?

packages/e2e-tests/tests/group-tests.spec.ts
```diff
@@ -57,7 +59,9 @@ test.afterAll(async ({ browser }) => {
   const context = await browser.newContext()
   const pageForProfileDeletion = await context.newPage()
   await reloadPage(pageForProfileDeletion)
-  await deleteAllProfiles(pageForProfileDeletion, existingProfiles)
+  if (existingProfiles.length > 0) {
+    await deleteAllProfiles(pageForProfileDeletion, existingProfiles)
+  }
```
WofWca: Why does this change affects only some files?

packages/e2e-tests/tests/plain-email-tests.spec.ts
```diff
@@ -88,3 +93,39 @@ test('check "New E-Mail" option is shown and a chat can be created', async () =>
...
+  await page.getByTestId('group-name-input').fill(subject)
+  await page.locator('#addmember').click()
+
+  await page.getByTestId('add-member-search').fill(emailUserB)
+
+  const contactList = page.getByTestId('add-member-dialog').locator('li button')
+
+  await expect(contactList).toHaveCount(0)
+
+  await expect(page.getByTestId('ok')).toHaveAttribute('disabled')
```
WofWca, line 119: This can potentially break, due to search taking some time.

Edit: I mean this will result in a false "pass".

packages/e2e-tests/tests/unencrypted-group-tests.spec.ts
```diff
@@ -17,10 +18,12 @@ const numberOfProfiles = 1
 // https://playwright.dev/docs/next/test-retries#reuse-single-page-between-tests
 let page: Page
 
-test.beforeAll(async ({ browser }) => {
+test.beforeAll(async ({ browser, chatmail }) => {
   const contextForProfileCreation = await browser.newContext()
   const pageForProfileCreation = await contextForProfileCreation.newPage()
-
+  if (chatmail) {
+    test.skip(true, 'Not suitable for non chatmail profiles')
```
WofWca:
```suggestion
  if (!chatmail) {
    test.skip(true, 'Only suitable for chatmail profiles')
```

The condition is incorrect, it causes test failures.
I also simplified the log message.

packages/e2e-tests/tests/unencrypted-group-tests.spec.ts
WofWca: Maybe `plain-email-tests.spec.ts` can be merged into this one now. The `check "New E-Mail" option presence` test is useful and should not be removed IMO.

WofWca (changes requested):
Looks good overall. Same for the irrelevant changes.

packages/e2e-tests/playwright-helper.ts
nicodh: Changed it to isChatmail

packages/e2e-tests/tests/unencrypted-group-tests.spec.ts
nicodh: We now could also run tests for both scenarios but with conditions in the test:

```
  const newEmailButton = page.getByRole('button', { name: 'New E-Mail' })
  if (isChatmail) {
    await expect(newEmailButton).not.toBeVisible()
    // Same button, but double-check, by ID.
    await expect(page.locator('#newemail')).not.toBeVisible({ timeout: 1 })
  } else {
    await expect(newEmailButton).toBeVisible()
  }

```

packages/e2e-tests/tests/group-tests.spec.ts
nicodh: Not neccessary - so I removed it

packages/e2e-tests/tests/unencrypted-group-tests.spec.ts
nicodh: Fixed it

packages/e2e-tests/tests/chat-list-multiselect.spec.ts
nicodh: I added them for developement and reviews.  Can be removed

packages/e2e-tests/tests/plain-email-tests.spec.ts
nicodh: What do you mean? The search for the ok button attribute?

WofWca: Well, verifying that the list is empty. It could be empty at first, and then become populated.

packages/e2e-tests/tests/group-tests.spec.ts
WofWca: Side note: this might be useful for when the test fails, so that other tests can run properly.

PR #5461, opened 2025-09-15
Enable to scan proxy url

### LINKED ISSUE #4290: Allow to scan proxy settings
It should be possibly to scan a proxy url and confirm that it should be used in proxy settings.
Using DC_QR_PROXY
see 
https://github.com/deltachat/interface/pull/75
https://github.com/deltachat/deltachat-core-rust/pull/5895
https://github.com/deltachat/deltachat-android/pull/3334

Once we have a separate page for the proxy settings, it would be nice to start the qr code scanner from there also

see https://github.com/deltachat/deltachat-desktop/issues/4124

nicodh:
resolves #4290

image

### DIFF
--- a/packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
+++ b/packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
@@ -204,7 +204,8 @@ export default function ProxyConfiguration(
     if (props.newProxyUrl) {
       addProxy(props.newProxyUrl)
     }
-  }, [props.newProxyUrl, addProxy])
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [props.newProxyUrl]) // skip addProxy in deps since it has too many deps itself
 
   const openQrScanner = useCallback(() => {
     openDialog(ProxyQrScanner, {


packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
```diff
@@ -133,40 +155,62 @@ export default function ProxyConfiguration(
...
+      maybeValidProxyUrl,
+      accountId,
+      openAlertDialog,
+      tx,
+      updateProxyState,
+    ]
+  )
+
+  if (props.newProxyUrl) {
+    addProxy(props.newProxyUrl)
```
WofWca: I think we should make sure that this only gets invoked once, before this breaks.
Edit: actually, this is already broken. You cannot delete the proxy that you just added, because it continuously gets re-added. Closing the dialog fixes this though.

Please create an issue if you choose to merge without addressing this.

packages/frontend/src/components/dialogs/ProxyQrScanner.tsx
```diff
@@ -0,0 +1,78 @@
...
+import useAlertDialog from '../../hooks/dialog/useAlertDialog'
+import { selectedAccountId } from '../../ScreenController'
+
+import type { DialogProps } from '../../contexts/DialogContext'
+
+/**
+ * QR code scanner for proxy configuration
+ * Processes scanned QR code and calls onSuccess if it is a valid proxy QR code
+ */
+export default function ProxyQrScanner({
```
WofWca, line 18: I would prefer to have a comment about the fact that this is mostly copy-pasted.

WofWca (approved):
Works nicely overall. I didn't check the "error" paths thoroughly though.

packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
WofWca: About the updated code: have you checked if it solves the issue? Because it doesn't look like it does. Because `addProxy` changes all the time, and `useEffect` still runs more than once.
Why not just utilize `useHasChanged2`?

packages/frontend/src/components/dialogs/ProxyQrScanner.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { processQr } from '../../backend/qr'
+import useAlertDialog from '../../hooks/dialog/useAlertDialog'
+import { selectedAccountId } from '../../ScreenController'
+
+import type { DialogProps } from '../../contexts/DialogContext'
+
+/**
+ * QR code scanner for proxy configuration
+ * Processes scanned QR code and calls onSuccess if it is a valid proxy QR code
+ * copied from dialogs/ProxyConfiguration/index.tsx
```
WofWca:
```suggestion
 * copied from dialogs/QrCode.tsx
```

packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
nicodh: I tried useHasChanged2 but it didn't work, the status was not updated at all.

So with the last change, removing addProxy from useEffect dependencies it works as expected.

WofWca: `useHasChanged2` has a parameter to `trueOnFirstRun` or something like that.

OK. Ignoring linter rules is not nice, but it works for now.

WofWca: > useHasChanged2 

Also having a "ran once" bool ref is an option.

PR #5462, opened 2025-09-15
Do not quote info messages on Ctrl Up

### LINKED ISSUE #5337: unreliable UX when replying to info messages

nicodh:
closes #5337 

#skip-changelog

### DIFF
--- a/packages/frontend/src/components/composer/Composer.tsx
+++ b/packages/frontend/src/components/composer/Composer.tsx
@@ -982,7 +982,11 @@ export function useDraft(
     if (chatId == undefined || !canSend) {
       return
     }
-    const quoteMessage = (messageId: number) => {
+    const quoteMessage = async (messageId: number) => {
+      const message = await BackendRemote.rpc.getMessage(accountId, messageId)
+      if (message.isInfo || message.viewType === 'VideochatInvitation') {
+        return
+      }
       draftRef.current.quote = {
         kind: 'WithMessage',
         messageId,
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/composer/Composer.tsx
```diff
@@ -982,7 +982,11 @@ export function useDraft(
     if (chatId == undefined || !canSend) {
       return
     }
-    const quoteMessage = (messageId: number) => {
+    const quoteMessage = async (messageId: number) => {
+      const message = await BackendRemote.rpc.getMessage(accountId, messageId)
+      if (message.isInfo || message.viewType === 'VideochatInvitation') {
+        return
+      }
```
WofWca, line 989: This makes it impossible to "step over" an info message by pressing Ctrl + Up multiple times. It's `quoteMessage`, not `maybeQuoteMessage`.

WofWca (changes requested):

nicodh:
Closed in favour of #5468

PR #5465, opened 2025-09-15
Refactor e2e tests

nicodh:
#skip-changelog


WofWca (approved):
This is a lot of different stuff and it's hard to call it a refactor. Here is the most important changes:
- Un-comment non-Chatmail tests.
- Tests from `plain-email-tests` have apparently been merged into `unencrypted-group-test`, which is good. But it's hard to see if it really is the case, there is also a lot of removals in the final file.
- Some change to the ESlint config.
- the `__testForceShowNewEmailButton` hack has been removed.

I re-ran the tests on GitHub, and they didn't fail. Which means that it's good to merge.

PR #5467, opened 2025-09-16
Remove e2ee_enabled key

nicodh:
e2ee_enabled is removed in core 2.13

The existing key in desktop settings cause a failure with core >= 2.13

#skip-changelog


WofWca (approved):
The preference is unused since 22e0826d1f81340b1a4eb9a84079d822eea1c9b5, so it's safe to remove.

PR #5468, opened 2025-09-16
Cache quotable message IDs

### LINKED ISSUE #5337: unreliable UX when replying to info messages

nicodh:
resolves #5337

Skip info messages on Ctrl + Up/Down

Video invitations are not skipped, but will be removed soon anyway. 

TBD: should we also invalidate the messageId cache when messages are deleted?

#skip-changelog

### DIFF
--- a/packages/frontend/src/components/composer/Composer.tsx
+++ b/packages/frontend/src/components/composer/Composer.tsx
@@ -974,6 +974,70 @@ export function useDraft(
   )
 
   const { jumpToMessage } = useMessage()
+
+  // Cache for message IDs to avoid expensive RPC calls
+  const quotableMessageIdsCache = useRef<{
+    chatId: number | null
+    messageIds: number[] | null
+  }>({ chatId: null, messageIds: null })
+
+  const getCachedQuotableMessageIds = useCallback(async (): Promise<
+    number[]
+  > => {
+    if (chatId === null) return []
+
+    // Check if we have cached data for the current chat
+    if (
+      quotableMessageIdsCache.current.chatId === chatId &&
+      quotableMessageIdsCache.current.messageIds
+    ) {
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/composer/Composer.tsx
```diff
@@ -974,6 +974,70 @@ export function useDraft(
...
+      messageIds: filteredMessageIds,
+    }
+
+    return filteredMessageIds
+  }, [accountId, chatId])
+
+  // Invalidate message IDs cache when the chat list changes
+  useEffect(() => {
+    if (chatId === null) return
+    const cleanup = onDCEvent(accountId, 'IncomingMsg', args => {
```
WofWca, line 1031: When you send a message yourself, the list does not update (same with deletion, as you mentioned).

WofWca (commented):
That's a lot of code. I don't think that trying to maintain two message lists (on top of the one we already have) is the right approach.

Do you have something against my suggestion in https://github.com/deltachat/deltachat-desktop/issues/5337#issuecomment-3127461812 ?

> That might be solved larger refactor, i.e. making the message list available to the shortcut handling code, which would also improve performance.

PR #5471, opened 2025-09-18
Reuse MessageStore in useDraft hook

nicodh:
- Extract useDraft from composer
- Include message store
- MessageStore as Singleton

As discussed here https://github.com/deltachat/deltachat-desktop/issues/5337#issuecomment-3127461812 "make the message list available to the shortcut handling code"
To avoid loading messages when navigation up or down when selecting messages for replyTo reuse the MessageStore which already holds the visible messages in messageCache

Special case: if there is already a quoted message in a draft, and the quoted message is not in messageCache (i.e. not visible) - we jump to that message on first shortcut event and only the next up or down will navigate to the next message.

### DIFF
--- a/packages/frontend/src/components/message/MessageList.tsx
+++ b/packages/frontend/src/components/message/MessageList.tsx
@@ -15,7 +15,7 @@ import { debounce } from 'debounce'
 import { MessageWrapper } from './MessageWrapper'
 import { getLogger } from '../../../../shared/logger'
 import { KeybindAction } from '../../keybindings'
-import { useMessageList } from '../../stores/messagelist'
+import { useMessageList, type MessageListStore } from '../../stores/messagelist'
 import { BackendRemote, onDCEvent } from '../../backend-com'
 import { throttledUpdateBadgeCounter } from '../../system-integration/badge-counter'
 import { MessagesDisplayContext } from '../../contexts/MessagesDisplayContext'
@@ -127,28 +127,36 @@ type Props = {
   accountId: number
   chat: T.FullChat
   refComposer: any
+  messageListStore: MessageListStore
+  messageListState: ReturnType<MessageListStore['getState']>
+  fetchMoreBottom: () => void
+  fetchMoreTop: () => void
 }
 
-export default function MessageList({ accountId, chat, refComposer }: Props) {
+export default function MessageList({
+  accountId,
+  chat,
+  refComposer,
+  messageListStore,
+  messageListState,
+  fetchMoreBottom,
+  fetchMoreTop,
+}: Props) {
   const {
-    store: {
-      scheduler,
-      effect: { jumpToMessage, loadMissingMessages },
-      reducer: { unlockScroll, clearJumpStack },
-      activeView,
-    },
-    state: {
-      oldestFetchedMessageListItemIndex,
-      newestFetchedMessageListItemIndex,
-      messageCache,
-      messageListItems,
-      viewState,
-      jumpToMessageStack,
-      loaded,
-    },
-    fetchMoreBottom,
-    fetchMoreTop,
-  } = useMessageList(accountId, chat.id)
+    scheduler,
+    effect: { jumpToMessage, loadMissingMessages },
+    reducer: { unlockScroll, clearJumpStack },
+    activeView,
+  } = messageListStore
+  const {
+    oldestFetchedMessageListItemIndex,
+    newestFetchedMessageListItemIndex,
+    messageCache,
+    messageListItems,
+    viewState,
+    jumpToMessageStack,
+    loaded,
+  } = messageListState
   const { hideReactionsBar, isReactionsBarShown } = useReactionsBar()
 
   const messageListRef = useRef<HTMLDivElement | null>(null)
--- a/packages/frontend/src/components/message/MessageListAndComposer.tsx
+++ b/packages/frontend/src/components/message/MessageListAndComposer.tsx
@@ -16,6 +16,7 @@ import { ReactionsBarProvider } from '../ReactionsBar'
 import useDialog from '../../hooks/dialog/useDialog'
 import useMessage from '../../hooks/chat/useMessage'
 import { Viewtype } from '@deltachat/jsonrpc-client/dist/generated/types'
+import { useMessageList } from '../../stores/messagelist'
 
 const log = getLogger('renderer/MessageListAndComposer')
 
@@ -101,6 +102,14 @@ export default function MessageListAndComposer({ accountId, chat }: Props) {
 
   const regularMessageInputRef = useRef<ComposerMessageInput>(null)
   const editMessageInputRef = useRef<ComposerMessageInput>(null)
+
+  const {
+    store: messageListStore,
+    state: messageListState,
+    fetchMoreBottom,
+    fetchMoreTop,
+  } = useMessageList(accountId, chat.id)
+
   const {
     draftState,
     updateDraftText,
@@ -112,6 +121,7 @@ export default function MessageListAndComposer({ accountId, chat }: Props) {
     clearDraftStateAndUpdateTextareaValue,
     setDraftStateAndUpdateTextareaValue,
   } = useDraft(
+    messageListState,
     accountId,
     chat.id,
     chat.isContactRequest,
@@ -295,6 +305,10 @@ export default function MessageListAndComposer({ accountId, chat }: Props) {
               accountId={accountId}
               chat={chat}
               refComposer={refComposer}
+              messageListStore={messageListStore}
+              messageListState={messageListState}
+              fetchMoreBottom={fetchMoreBottom}
+              fetchMoreTop={fetchMoreTop}
             />
           </ReactionsBarProvider>
         </RecoverableCrashScreen>
--- a/packages/frontend/src/hooks/chat/useDraft.ts
+++ b/packages/frontend/src/hooks/chat/useDraft.ts
@@ -9,6 +9,7 @@ import { MessageTypeAttachmentSubset } from '../../components/attachment/Attachm
 import { KeybindAction } from '../../keybindings'
 import useMessage from './useMessage'
 import ComposerMessageInput from '../../components/composer/ComposerMessageInput'
+import { type MessageListStore } from '../../stores/messagelist'
 
 const log = getLogger('renderer/composer')
 
@@ -34,11 +35,12 @@ function emptyDraft(chatId: number | null): DraftObject {
 }
 
 export function useDraft(
+  messageListState: MessageListStore['state'],
   accountId: number,
   chatId: number | null,
   isContactRequest: boolean,
   canSend: boolean, // no draft needed in chats we can't send messages
-  inputRef: React.MutableRefObject<ComposerMessageInput | null>
+  inputRef: React.RefObject<ComposerMessageInput | null>
 ): {
   draftState: DraftObject
   onSelectReplyToShortcut: (
@@ -68,6 +70,7 @@ export function useDraft(
      */
     _setDraftStateButKeepTextareaValue,
   ] = useState<DraftObject>(emptyDraft(chatId))
+
   /**
    * `draftRef.current` gets set to `draftState` on every render.
    * That is, when you mutate the value of this ref,
@@ -246,6 +249,11 @@ export function useDraft(
   )
 
   const { jumpToMessage } = useMessage()
+
+  /**
+   * Support the Ctrl/Cmd+Up/Down shortcuts to select a message
+   * to reply to and set the quote in the draft accordingly
+   */
   const onSelectReplyToShortcut = async (
     upOrDown:
       | KeybindAction.Composer_SelectReplyToUp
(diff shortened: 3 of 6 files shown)


packages/frontend/src/stores/messagelist.ts
```diff
@@ -67,9 +91,7 @@ export function useMessageList(
   fetchMoreTop: () => void
 } {
   const store = useMemo(() => {
-    const store = new MessageListStore(accountId, chatId)
-    store.effect.loadChat()
-    return store
+    return getOrCreateMessageListStore(accountId, chatId)
```
WofWca: This is an extremely significant change that breaks a lot of stuff currently. Switching chats back and forth shows "error loading message" for a moment. When the messages do load, you end up in the middle of the chat. Loading a chat, then switching from it, then getting a message, then switching back to this chat also doesn't scroll to the last message.

My idea with "make the message list available to the shortcut handling code" was not to add global state that `useMessageList` would reference, but instead, in React terminology, to simply lift the state up, to the first common ancestor of both the MessageList and the Ctrl + Up handling code.

packages/frontend/src/hooks/chat/useDraft.ts
```diff
@@ -0,0 +1,388 @@
...
+    messageListStore.state.messageListItems
+  )
+
+  useEffect(() => {
+    const unsubscribe = messageListStore.subscribe(update => {
+      setMessageCache(update.messageCache)
+      setMessageListItems(update.messageListItems)
+    })
+    return unsubscribe
+  }, [messageListStore])
```
WofWca: This is just duplicated state, right? Why is it needed? Because `messageListStore.state` is not reactive? I'd prefer to either fix that or to have a `forceRerender` function instead.

packages/frontend/src/hooks/chat/useDraft.ts
```diff
@@ -0,0 +1,388 @@
...
+    const currQuoteMessageIdInd = messageIds.lastIndexOf(currQuote.messageId)
+    if (currQuoteMessageIdInd === -1) {
+      // maybe the message is just not in the cache (yet)
+      // but still in the full list of messages
+      // -> check if it's there
+      const fullMsgIdList = messageListItems
+        .filter(m => m.kind === 'message')
+        .map(m => m.msg_id)
+      // quoted message not found in cache
+      const fullIndex = fullMsgIdList.findIndex(m => m === currQuote.messageId)
```
WofWca: prefer `Array.includes`.

Also why not just go for `.some` right away? `filter` + `map` could be dealing with megabyte arrays here.

packages/frontend/src/hooks/chat/useDraft.ts
```diff
@@ -0,0 +1,388 @@
...
+   */
+  const onSelectReplyToShortcut = async (
+    upOrDown:
+      | KeybindAction.Composer_SelectReplyToUp
+      | KeybindAction.Composer_SelectReplyToDown
+  ) => {
+    if (chatId == undefined || !canSend) {
+      return
+    }
+    const quoteMessage = (messageId: number, jumpToOnly = false) => {
```
WofWca: I think it's better to separate a `jumpToMessage` function rather than fusing the "just jump to message" functionality into a function named `quoteMessage`.

packages/frontend/src/hooks/chat/useDraft.ts
```diff
@@ -0,0 +1,388 @@
...
+        scrollIntoViewArg: { block: 'nearest' },
+      })
+    }
+
+    const messageIds = Object.keys(messageCache)
+      .map(Number)
+      .filter(
+        id =>
+          messageCache[id]?.kind === 'message' &&
+          messageCache[id]?.chatId === chatId &&
```
WofWca: Do we expect this to be `false`?

packages/frontend/src/stores/messagelist.ts
```diff
@@ -20,6 +20,18 @@ const log = getLogger('messagelist')
...
+ * Creates a new MessageListStore instance.
+ * The store should be created at the parent component level (MessageListAndComposer)
+ * and shared between useMessageList and useDraft to avoid redundant data loading.
+ */
+export function createMessageListStore(
+  accountId: number,
+  chatId: number
+): MessageListStore {
+  return new MessageListStore(accountId, chatId)
+}
```
WofWca: How about moving this comment to `MessageListStore` and eliminating this one-line function?

packages/frontend/src/hooks/chat/useDraft.ts
```diff
@@ -0,0 +1,388 @@
...
+          messageCache[id]?.isInfo === false
```
WofWca: It would also be nice to cross-reference with the `showReply` code.

https://github.com/deltachat/deltachat-desktop/blob/77a1f88a351df49e5df38a14c3a1704a76ecdcb3/packages/frontend/src/components/message/Message.tsx#L274-L278

nicodh: Why? In this PR it's only about the shortcut which only applies if there is a composer input. So no need to check chat.canSend.

WofWca: We used to have a bug where you could execute the shortcut in a read only chat.

But this is not only about `canSend`. The fact that info messages are the only messages that you cannot reply to is only incidental. We just might disable reply for video call messages. For such a case, these two pieces of code need to be in sync. Android already has some utility functions for such logic.

nicodh: So with cross-reference you don't mean code but just a comment pointing to that code?

WofWca: Well, maybe the code will be the best, but a comment is also better than nothing I'd say. Yes, I meant a comment.

packages/frontend/src/components/message/MessageListAndComposer.tsx
```diff
@@ -98,6 +100,15 @@ export default function MessageListAndComposer({ accountId, chat }: Props) {
   const { openDialog, hasOpenDialogs } = useDialog()
   const { sendMessage } = useMessage()
 
+  // Create a store instance and share it between
+  // useMessageList and useDraft to avoid redundant data loading
```
WofWca: IDK if this comment is much useful. It's clear from the code that `messageListStore` is shared.

packages/frontend/src/stores/messagelist.ts
```diff
@@ -1195,3 +1190,20 @@ async function loadMessages(
...
+// Custom hook to reactively access message list store state
+export const useMessageListStore = (messageListStore: MessageListStore) => {
+  const [, forceUpdate] = useReducer(x => x + 1, 0)
+
+  useEffect(() => {
+    const unsubscribe = messageListStore.subscribe(() => {
+      forceUpdate()
+    })
+    return unsubscribe
+  }, [messageListStore])
```
WofWca: I'm not sure when exactly `useEffect` runs. I'm a little afraid that this can result in some updates being missed. I would suggest to use the same approach that we use here:

https://github.com/deltachat/deltachat-desktop/blob/e76e5e6196f1f8143b8ef828fa8f8e49b2f5a40c/packages/frontend/src/stores/messagelist.ts#L144-L150

Or maybe to ensure synchronous execution of `subscribe`, e.g. with `useHasChanged2`.

And in general - these two pieces of code seem to be trying to duplicate the same functionality. I think we need to DRY here.

But maybe it's just me not knowing React.

packages/frontend/src/stores/messagelist.ts
```diff
@@ -53,11 +53,12 @@ const defaultState = () =>
...
- * A hook to read a portion of messages(a view) for a given chat. It creates a store(MessageListStore)
+ * A hook to read a portion of messages(a view) for a given chat. It uses the passed store(MessageListStore)
  * for the given chat and account and loads messages on it. It always has a maximum specified number
- * of messages as per PAGE_SIZE constant.
+ * of messages as per PAGE_SIZE constant. The store is also used in useDraft to avoid redundant data loading.
  */
 export function useMessageList(
+  store: MessageListStore,
   accountId: number,
   chatId: number
```
WofWca: With this function signature it is now possible to provide a MessageListStore whose `chatId` and `accountId` are different from that of `store`. I think we should obtain them from the `store` and remove the parameters.

packages/frontend/src/stores/messagelist.ts
WofWca, line 65: Returning these doesn't seem to provide value.

packages/frontend/src/stores/messagelist.ts
```diff
@@ -53,11 +53,12 @@ const defaultState = () =>
   }) as MessageListState
 
 /*
- * A hook to read a portion of messages(a view) for a given chat. It creates a store(MessageListStore)
+ * A hook to read a portion of messages(a view) for a given chat. It uses the passed store(MessageListStore)
  * for the given chat and account and loads messages on it. It always has a maximum specified number
- * of messages as per PAGE_SIZE constant.
+ * of messages as per PAGE_SIZE constant. The store is also used in useDraft to avoid redundant data loading.
```
WofWca: All my other comments from this review aside, it doesn't seem to me that trying to decouple `useMessageList` from `MessageListStore` is good idea. What is the point of it if `useMessageList` is the only thing that is supposed to be responsible for managing (mutating) the `MessageListStore`? Basically `MessageListStore` can never exist without the corresponding `useMessageList`. But then why would we spread the creation of `MessageListStore` and the invocation of `useMessageList` so far apart?

I suggest to revert the decoupling, and to simply lift the `useMessageList()` call to inside `MessageListAndComposer`, where `useDraft` can access the return value of `useMessageList()`.
With this done, I believe all my other comments from this review will become irrelevant.

packages/frontend/src/hooks/chat/useDraft.ts
WofWca: This seems to have been done, but `jumpToOnly` has not been removed.

WofWca:
About performance: for some reason it feels very sluggish, but only in dev mode. If you build frontend in prod mode, then it's very snappy!

packages/frontend/src/stores/messagelist.ts
WofWca: Resolved thanks to https://github.com/deltachat/deltachat-desktop/pull/5471/files#r2441759745 having been resolved.

packages/frontend/src/stores/messagelist.ts
```diff
@@ -53,9 +53,9 @@ const defaultState = () =>
...
```
WofWca: With https://github.com/deltachat/deltachat-desktop/pull/5471/files#r2441759745 having been addressed, the comments also needs to get reverted.


```suggestion
 * A hook to read a portion of messages(a view) for a given chat. It creates a store(MessageListStore)
 * for the given chat and account and loads messages on it. It always has a maximum specified number
 * of messages as per PAGE_SIZE constant.
```

packages/frontend/src/components/message/MessageListAndComposer.tsx
```diff
@@ -100,6 +102,14 @@ export default function MessageListAndComposer({ accountId, chat }: Props) {
 
   const regularMessageInputRef = useRef<ComposerMessageInput>(null)
   const editMessageInputRef = useRef<ComposerMessageInput>(null)
+
+  const {
+    store: messageListStore,
+    state: messageListState,
+    fetchMoreBottom,
+    fetchMoreTop,
+  } = useMessageList(accountId, chat.id)
```
WofWca, line 111: Maybe not destructuring these would be tidier, but alright.

packages/frontend/src/components/message/MessageList.tsx
```diff
@@ -15,7 +15,7 @@ import { debounce } from 'debounce'
 import { MessageWrapper } from './MessageWrapper'
 import { getLogger } from '../../../../shared/logger'
 import { KeybindAction } from '../../keybindings'
-import { useMessageList } from '../../stores/messagelist'
+import { useMessageList, MessageListStore } from '../../stores/messagelist'
```
```suggestion
import { useMessageList, type MessageListStore } from '../../stores/messagelist'
```

packages/frontend/src/stores/messagelist.ts
```diff
@@ -10,7 +10,7 @@ import {
   defaultChatViewState,
 } from './chat/chat_view_reducer'
 import { ChatStoreScheduler } from './chat/chat_scheduler'
-import { useEffect, useMemo, useRef, useState } from 'react'
+import { useEffect, useRef, useState, useMemo } from 'react'
```
```suggestion
import { useEffect, useMemo, useRef, useState } from 'react'
```

packages/frontend/src/hooks/chat/useDraft.ts
```diff
@@ -0,0 +1,379 @@
...
+import { basename } from 'path'
+
+import { getLogger } from '../../../../shared/logger'
+import { BackendRemote, Type } from '../../backend-com'
+import { selectedAccountId } from '../../ScreenController'
+import { MessageTypeAttachmentSubset } from '../../components/attachment/Attachment'
+import { KeybindAction } from '../../keybindings'
+import useMessage from './useMessage'
+import ComposerMessageInput from '../../components/composer/ComposerMessageInput'
+import { MessageListStore } from '../../stores/messagelist'
```
WofWca:
```suggestion
import { type MessageListStore } from '../../stores/messagelist'
```

nicodh:
> As I mentioned, it would be very nice to have a separate commit for `useDraft` being moved to a separate file, otherwise the changes that this MR does to it are gonna be undecipherable.

But there is a separate commit for extract useDraft:
https://github.com/deltachat/deltachat-desktop/pull/5471/commits/71a00f1ce18781dccba5fc08664a5f7681d83aaf

WofWca:
Yes, sorry, I meant that we should not squash that commit, but squash the other ones (or almost all the other ones)

nicodh:
> Yes, sorry, I meant that we should not squash that commit, but squash the other ones (or almost all the other ones)

That means I have to merge locally or is that possible in the github UI?

WofWca:
No I don't think it's possible on GitHub, but one could squash locally and then force push to this MR.

nicodh:
> No I don't think it's possible on GitHub, but one could squash locally and then force push to this MR.

Damn, I just wanted to expand the "Squash & Merge button" to select the other option and then it just squashed an merged it immediately. Bad UI!

WofWca:
FTR we force-pushed. This is merged as 6fc4777a8b42cfc303514d6d847a5d75ce2ee914 (and the parent commit).

PR #5473, opened 2025-09-19
fix: emit core events on main proc JSON-RPC client

WofWca:
Currently this commit does not affect behavior of the app,
so it can be considered a refactor, #skip-changelog.

This is a preparation for the upcoming calls MR,
which relies on listening to events in the main process.

Related:
- e16f88b47ba6323577e2e91bcef1d16856ca82f4
  (https://github.com/deltachat/deltachat-desktop/pull/5470).

### DIFF
--- a/packages/target-electron/src/deltachat/controller.ts
+++ b/packages/target-electron/src/deltachat/controller.ts
@@ -30,17 +30,6 @@ class ElectronMainTransport extends yerpc.BaseTransport {
 }
 
 export class JRPCDeltaChat extends BaseDeltaChat<ElectronMainTransport> {}
-/**
- * A DeltaChat JSON-RPC client that does not emit events
- * (but still can be used to send regular requests).
- *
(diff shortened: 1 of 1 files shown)


packages/target-electron/src/deltachat/controller.ts
```diff
@@ -128,6 +117,45 @@ export default class DeltaChatController {
               typeof event === 'object' &&
               event.kind
             ) {
+              // A stupid hack.
```
nicodh: I would remove this line. Looks "from outside" as if we do "stupid hacks" ;-)

Better name it "temporary workaround" and add the "reason" for this workaround here: a line why there can be only one consumer with the current implementation of the event emitting/handling

WofWca: > Looks "from outside" as if we do "stupid hacks"

It is what it is 🤷‍♀️.

OK, I'll replace with "workaround"

PR #5476, opened 2025-09-19
feat: add outgoing video calls

WofWca:
image image image

TODO:
- [x] Put this behind an experimental setting.
- [x] Clean up comments.
- [x] Merge the base branch into main.
- [x] Maybe add a CHANGELOG entry. Or do it when we un-experimental the setting instead.

Otherwise this MR can be reviewed, at least the overall structure of things and the approach, not the messy TODOs.

We agreed that it's probably OK to implement the calls feature bit by bit, in such small MRs, since it's all going to be behind an "experimental" toggle for a while.
But turning this MR into the second "main" branch is also an option.
Although I'd prefer the former.

Related:
- https://github.com/orgs/deltachat/projects/81/views/1.
- https://github.com/deltachat/deltachat-android/pull/3785.
- https://github.com/deltachat/deltachat-ios/pull/2638.

#skip-changelog since this is part of a not yet finished feature

### DIFF
--- a/packages/target-electron/src/ipc.ts
+++ b/packages/target-electron/src/ipc.ts
@@ -57,9 +57,6 @@ let dcController: typeof DeltaChatController.prototype
 export function getDCJsonrpcRemote() {
   return dcController.jsonrpcRemote
 }
-export function getDCJsonrpcClient() {
-  return dcController.jsonrpcRemote.rpc
-}
 
 /** returns shutdown function */
 export async function init(cwd: string, logHandler: LogHandler) {
--- a/packages/target-electron/src/windows/html_email.ts
+++ b/packages/target-electron/src/windows/html_email.ts
@@ -20,7 +20,7 @@ import { isInviteLink, truncateText } from '@deltachat-desktop/shared/util.js'
 import { tx } from '../load-translations.js'
 import { open_url } from '../open_url.js'
 import { loadTheme } from '../themes.js'
-import { getDCJsonrpcClient } from '../ipc.js'
+import { getDCJsonrpcRemote } from '../ipc.js'
 import { getLogger } from '../../../shared/logger.js'
 
 import * as mainWindow from './main.js'
@@ -459,7 +459,7 @@ function makeBrowserView(
   if (allow_remote_content) {
     const callback = async (req: Request) => {
       try {
-        const response = await getDCJsonrpcClient().getHttpResponse(
+        const response = await getDCJsonrpcRemote().rpc.getHttpResponse(
           account_id,
           req.url
         )


nicodh:
If we merge this into main instead "waiting" for a full feature branch being complete (which is fine to me) I would put it behind a experimental feature and add (WIP) to point out that is it is not only an experimental feature but also a not yet completed feature

WofWca:
> add (WIP)

Add it where? In the toggle's description?

nicodh:
> > add (WIP)
> 
> Add it where? In the toggle's description?

Either in the title or mention it in the description

packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
```diff
@@ -514,9 +515,34 @@ function ChatNavButtons({ chat }: { chat: T.FullChat }) {
     })
   }, [openDialog, chatId])
 
+  const openConfirmationDialog = useConfirmationDialog()
```
nicodh, line 518: This might also go into the calls webapp later, see my comment in https://github.com/deltachat/calls-webapp/issues/28#issuecomment-3322756229

If not, we should check for camera access here and maybe already show the video stream of the camera.

All for later, but maybe add a comment here?

packages/target-electron/package.json
```diff
@@ -56,6 +57,7 @@
   "dependencies": {
     "@deltachat/jsonrpc-client": "catalog:",
     "@deltachat/stdio-rpc-server": "catalog:",
+    "calls-webapp#build-for-deltachat-desktop": "catalog:",
```
nicodh, line 60: Should we at least rebase the build-for-deltachat-desktop before merging this? And/or ask directly for a review there to get this into main?

packages/target-electron/src/ipc.ts
```diff
@@ -53,6 +54,9 @@ const log = getLogger('main/ipc')
 const app = rawApp as ExtendedAppMainProcess
 
 let dcController: typeof DeltaChatController.prototype
+export function getDCJsonrpcRemote() {
+  return dcController.jsonrpcRemote
+}
 export function getDCJsonrpcClient() {
```
nicodh: Maybe remove this method now, since it's only used in html_emails and there we could go with 

await getDCJsonrpcRemote().rpc.getHttpResponse instead

packages/target-electron/src/windows/video-call.ts
```diff
@@ -0,0 +1,297 @@
...
+  })
+}
+
+const SCHEME_NAME = 'calls-webapp-scheme'
+const DUMMY_HOST_NAME = 'calls-webapp-dummy-host'
+
+/**
+ * @see `origin_no_opaque` in the Tauri version.
+ */
+function isOriginGood(url: string) {
```
nicodh, line 275: Rename to "isValidOrigin" or "isValidProtocolUrl"?

packages/target-electron/src/windows/video-call.ts
```diff
@@ -0,0 +1,297 @@
...
+        requestingOrigin,
+        ...details,
+      })
+    }
+  )
+  ses.setPermissionRequestHandler((_wc, permission, callback, details) => {
+    callback(permission_handler(permission, details))
+  })
+
+  const port = (() => {
```
nicodh: rename to messagePort or webAppMessagePort to make it a bit more self explaining?

```suggestion
 // see https://www.electronjs.org/docs/latest/api/message-channel-main
  const webAppMessagePort = (() => {
```

packages/target-electron/src/windows/video-call.ts
```diff
@@ -0,0 +1,297 @@
...
+            'turn:49.12.217.82',
+          ],
+          username: 'ohV8aec1',
+          credential: 'zo3theiY',
+        },
+      ]),
+    })
+  }
+
+  const offerP = new Promise<string>(r => {
```
nicodh: Rename to avoid confusion with Port. Since we have a communication with the webapp via port and a communication between a caller and a receiver via Promises and an offerPayload (in another file) it might help to have a concise naming.

```suggestion
  const offerPromise = new Promise<string>(r => {
```

packages/target-electron/src/windows/video-call.ts
```diff
@@ -0,0 +1,297 @@
...
+export function startOutgoingVideoCall(accountId: number, chatId: number) {
+  log.info('starting outgoing video call', { accountId, chatId })
+
+  const ses = session.fromPartition(`calls-webapp_${accountId}`)
+
+  if (!ses.protocol.isProtocolHandled(SCHEME_NAME)) {
+    ses.protocol.handle(SCHEME_NAME, returnCallsWebappFile)
+  }
+
+  const win = new BrowserWindow({
```
nicodh, line 26:
```suggestion
  const callWindow = new BrowserWindow({
```

packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
WofWca: I don't get it why?

packages/target-electron/package.json
WofWca: There is no strict need to rebase, the app is working just fine. But OK, I will rebase.

packages/target-electron/src/windows/video-call.ts
WofWca: I don't think there is a need to be that specific. There is only one window after all in the context of this file.

packages/target-electron/src/windows/video-call.ts
WofWca: IDK, "valid" sounds like "we managed to parse it".

packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
nicodh: if the dialog is in the calls webapp we just have to open it, no additional dialog needed here

The comment could be:
#TODO implement a check for camera access here, since the webapp will just hang, showing "Connecting" if there is no camera

WofWca: I think the calls-webapp should handle the camera access error. In fact, I already made not fail if there is only mic but no camera.

WofWca:
Done, force-pushed.

PR #5480, opened 2025-09-21
feat: add incoming video calls

WofWca:


WofWca:
So, what about the changelog? Do we skip adding an entry for now?

nicodh:
> So, what about the changelog? Do we skip adding an entry for now?

Yes, just to make obvious that this is WIP

WofWca:
Oh, and was this an approval of just this MR, or is it together with https://github.com/deltachat/deltachat-desktop/pull/5476?

nicodh:
> Oh, and was this an approval of just this MR, or is it together with #5476?

Just for this MR (although now also #5481 needs to be merged into this) I'm just testing it

PR #5481, opened 2025-09-22
fix: properly `rpc.endCall()`, close call window

WofWca:

### DIFF
--- a/packages/target-electron/src/windows/video-call.ts
+++ b/packages/target-electron/src/windows/video-call.ts
@@ -28,6 +28,13 @@ export function startOutgoingVideoCall(accountId: number, chatId: number) {
 
   ;(async () => {
     const { offer, onAnswer } = await offerPromise
+    if (offer == null) {
+      log.info("calls-webapp didn't return an offer, aborting outgoing call")
+      // We expect this code path to be taken
+      // only if the window already got closed, but let's be defeinsive.
+      closeWindow()
+      return
+    }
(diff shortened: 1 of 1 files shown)


packages/target-electron/src/windows/video-call.ts
```diff
@@ -318,6 +331,15 @@ function openVideoCallWindow<T extends CallDirection>(
       },
     })),
     windowClosed,
+    closeWindow() {
```
nicodh: Is there a reason why you use method shorthand syntax here? We usually use object property definitions.
I consider it more readable and there is a certain risk https://www.totaltypescript.com/method-shorthand-syntax-considered-harmful

packages/target-electron/src/windows/video-call.ts
```diff
@@ -318,6 +331,15 @@ function openVideoCallWindow<T extends CallDirection>(
...
+      if (win.webContents.isDevToolsOpened()) {
```
nicodh, line 368: This triggers a warning:
UnhandledPromiseRejectionWarning: TypeError: Object has been destroyed

Not an issue but can we maybe avoid it?

packages/target-electron/src/windows/video-call.ts
WofWca: No particular reason.
OK, I'll change it.

packages/target-electron/src/windows/video-call.ts
WofWca: Added an `isDestroyed()` check.

PR #5483, opened 2025-09-23
feat: video calls: show other's avatar

WofWca:
#skip-changelog because we'll add it later.

This changes the calls-webapp URL to add `accountId`,
`chatId` and `messageId` to it, in order to determine
which avatar we need to get. Similarly to our WebXDC implementation.

Note that if a user doesn't have an avatar,
the avatar won't be displayed at all.
https://github.com/deltachat/calls-webapp/pull/29 would improve this a little.

image

### DIFF
--- a/packages/target-electron/src/windows/video-call.ts
+++ b/packages/target-electron/src/windows/video-call.ts
@@ -661,11 +661,6 @@ function formatHost(
     : `${chatIdOrMessageId.callMessageId}.none.${accountId}.calls-webapp-dummy-host`
 }
 /**
- * check if the numbers used in host are valid
- * and the format (see formatHost) is correct
- * @returns `null` if the host is invalid.
- */
-/**
  * @see {@linkcode formatHost}.
  * @returns `null` if the host is invalid.
  */
--- a/packages/target-electron/src/windows/video-call.ts
+++ b/packages/target-electron/src/windows/video-call.ts
@@ -665,6 +665,10 @@ function formatHost(
  * and the format (see formatHost) is correct
  * @returns `null` if the host is invalid.
  */
+/**
+ * @see {@linkcode formatHost}.
+ * @returns `null` if the host is invalid.
+ */
 function parseHost(host: string): null | {
   accountId: number
   chatIdOrMessageId:
--- a/packages/target-electron/src/windows/video-call.ts
+++ b/packages/target-electron/src/windows/video-call.ts
@@ -660,6 +660,11 @@ function formatHost(
     ? `${chatIdOrMessageId.callMessageId ?? 'none'}.${chatIdOrMessageId.chatId}.${accountId}.calls-webapp-dummy-host`
     : `${chatIdOrMessageId.callMessageId}.none.${accountId}.calls-webapp-dummy-host`
 }
+/**
+ * check if the numbers used in host are valid
+ * and the format (see formatHost) is correct
+ * @returns `null` if the host is invalid.
+ */
 function parseHost(host: string): null | {
   accountId: number
   chatIdOrMessageId:
--- a/packages/target-electron/src/windows/video-call.ts
+++ b/packages/target-electron/src/windows/video-call.ts
@@ -20,6 +20,7 @@ export function startOutgoingVideoCall(accountId: number, chatId: number) {
 
   const { offerPromise, windowClosed, closeWindow } = openVideoCallWindow(
     accountId,
+    { chatId },
     CallDirection.Outgoing,
     {}
   )
@@ -93,6 +94,7 @@ function openIncomingVideoCallWindow(
 
   const { answerPromise, windowClosed, closeWindow } = openVideoCallWindow(
     accountId,
+    { callMessageId },
     CallDirection.Incoming,
     {
       callerWebrtcOffer,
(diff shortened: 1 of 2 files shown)


packages/target-electron/src/windows/video-call.ts
```diff
@@ -526,41 +541,86 @@ const CSP =
...
     ...responseInit,
-    headers: {
-      'Content-Security-Policy': CSP,
-    },
+    headers,
   })
 }
 
 const SCHEME_NAME = 'calls-webapp-scheme'
-const DUMMY_HOST_NAME = 'calls-webapp-dummy-host'
```
nicodh, line 625: Why did you remove the DUMMY_HOST_NAME constant? The string is used in 2 places so using a constant makes sense imo

packages/target-electron/src/windows/video-call.ts
```diff
@@ -572,10 +632,72 @@ function isOriginGood(url: string) {
...
+      }
+} {
+  const [messageIdStr, chatIdStr, accountIdStr, dummyHostName, ...rest] =
+    host.split('.')
+  const [callMessageId, chatId, accountId] = [
+    parseInt(messageIdStr),
+    parseInt(chatIdStr),
+    parseInt(accountIdStr),
+  ]
+  const isNumGood = (num: number) => Number.isFinite(num) && num >= 0
```
nicodh: tbh I really don't like "good" in variable naming. I would say it's highly unused in software context.

Good and bad are moral judgments.

In this place it doesn't matter since it's only used some lines after, but the isOriginGood sounds very strange to me.

Why not just use  "isValid" here?

packages/target-electron/src/windows/video-call.ts
```diff
@@ -572,10 +632,72 @@ function isOriginGood(url: string) {
   }
   return (
     urlParsed.protocol === SCHEME_NAME + ':' &&
-    urlParsed.host === DUMMY_HOST_NAME &&
+    parseHost(urlParsed.host) != null &&
     urlParsed.port === ''
   )
 }
+function formatHost(
```
nicodh, line 647:
```suggestion
/**
 * @returns the host string in format
 *  `{callMessageId|none}.{chatId|none}.{accountId}.calls-webapp-dummy-host`
 */
 function formatHost(
```

packages/target-electron/src/windows/video-call.ts
```diff
@@ -572,10 +632,72 @@ function isOriginGood(url: string) {
...
+      }
+): string {
+  return (
+    `${chatIdOrMessageId.callMessageId ?? 'none'}.` +
+    `${chatIdOrMessageId.chatId ?? 'none'}.` +
+    `${accountId}.` +
+    `calls-webapp-dummy-host`
+  )
+}
+function parseHost(host: string): null | {
```
nicodh, line 667:
```suggestion
/**
 * check if the numbers used in host are valid
 * and the format (see formatHost) is correct
 * @returns `null` if the host is invalid.
 */
function parseHost(host: string): null | {
```

packages/target-electron/src/windows/video-call.ts
```diff
@@ -132,6 +134,18 @@ const enum CallDirection {
 
 function openVideoCallWindow<T extends CallDirection>(
   accountId: number,
+  /**
+   * Whatever is known, either chatId or messageId, should be passed.
```
nicodh:
```suggestion
   * Depending on the call direction we only know
   * the chat ID or the call message ID
```

nicodh (approved):
Reviewed by testing & reading

Some additional comments and concerns about the naming (again :-)

packages/target-electron/src/windows/video-call.ts
WofWca: Because the host name is now consists of not just this string, and the 2 places where it's used are 2 complementary adjacently placed functions.

packages/target-electron/src/windows/video-call.ts
WofWca: I force-pushed, adding `type HostStr`. See if it's good.

packages/target-electron/src/windows/video-call.ts
WofWca:
```suggestion
/**
 * @see {@linkcode formatHost}.
 * @returns `null` if the host is invalid.
 */
function parseHost(host: string): null | {
```

The first 2 lines are redundant IMO. This info can be inferred from the return type.

packages/target-electron/src/windows/video-call.ts
WofWca: > Good and bad are moral judgments.

"Bad milk" doesn't mean "this milk has committed sins".

> I would say it's highly unused in software context.

I don't think so, I see it all the time.

OK, `valid` sounds good here. `isValidId`. I changed it.

packages/target-electron/src/windows/video-call.ts
WofWca: This doesn't have to be call direction though.

WofWca:
Force-pushed, addressing some comments. Please see if it's good and mark as resolved.

packages/target-electron/src/windows/video-call.ts
nicodh: We could name it DUMMY_HOST_APPENDIX - but not worth to discuss further...

nicodh (approved):

PR #5485, opened 2025-09-23
refactor: `startOutgoingVideoCall`: return Promise

WofWca:
This should make the front-end console print an error
if `startOutgoingVideoCall` throws an error.
This does not affect functionality.

### DIFF
--- a/packages/target-electron/src/windows/video-call.ts
+++ b/packages/target-electron/src/windows/video-call.ts
@@ -27,7 +27,7 @@ export function startOutgoingVideoCall(accountId: number, chatId: number) {
 
   const jsonrpcRemote = getDCJsonrpcRemote()
 
-  ;(async () => {
+  const callHandledPromise = (async () => {
     const { offer, onAnswer } = await offerPromise
     if (offer == null) {
       log.info("calls-webapp didn't return an offer, aborting outgoing call")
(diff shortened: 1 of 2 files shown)


packages/target-electron/src/windows/video-call.ts
```diff
@@ -60,7 +60,11 @@ export function startOutgoingVideoCall(accountId: number, chatId: number) {
     }
     log.info('Received answer from callee')
     onAnswer(answer)
+
+    return await done
```
nicodh, line 65: Sorry, naming again: shouldn't we name this afterCallEnded? 
And
onDone => onCallEnded ?
donePromise => afterCallPromise

It's really a bit hard to read through all these nested promises, that's why I'm a bit nit picky...

WofWca: "done" does not always mean "call ended", because we can be "done" while the call is still being conducted on another device:

https://github.com/deltachat/deltachat-desktop/blob/f22567312296e6c0bdf0ac41d53b586a1cd1cc0e/packages/target-electron/src/windows/video-call.ts#L437-L441

What would be your suggestion given this information?

nicodh: What about
afterCallHandlingEnded
onCallHandlingEnded
afterCallHandlingPromise

I think it's worth to be pretty verbose here...

WofWca: OK, I renamed the variables introduced in this MR. I'll make a follow-up.

WofWca: `callHandledPromise` sounds good?

PR #5486, opened 2025-09-23
feat: show caller's name on incoming call

WofWca:
Also set their avatar as the window icon.

image

#skip-changelog because we'll add it later.


nicodh:
This is how it looks like on Linux btw.

image

nicodh (approved):
Reviewed by testing & reading

Needs to be rebased

PR #5488, opened 2025-09-26
improvement: calls: use in-app "Answer?" prompt

WofWca:
- It looks nicer.
- Allows you to toggle video and mute mic before you accept the call.

image

See https://github.com/deltachat/calls-webapp/pull/37.

Security-wise, not that it's critical, but note that now
we pass the offer to the app immediately, so hypothetically
the app could auto-accept the call without waiting
for user interaction, if it wanted to.

Let's not remove the old prompt code for now.

The updated calls-webapp has been built from ~~<https://github.com/deltachat/calls-webapp/commit/63fd46ee132029635e838b81e6f61d84019688d1>~~ https://github.com/deltachat/calls-webapp/commit/0fa0c9dd616183070ed9b58e4ada6b1c0382349e (latest main).

### DIFF
--- a/packages/target-electron/src/windows/video-call.ts
+++ b/packages/target-electron/src/windows/video-call.ts
@@ -315,7 +315,13 @@ function openVideoCallWindow<T extends CallDirection>(
     })
   }
 
-  if (callerWebrtcOffer != undefined) {
+  /**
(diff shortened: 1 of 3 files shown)


nicodh (approved):
Reviewed by testing and reading

PR #5490, opened 2025-09-26
fix: calls window: a way to go fullscreen, zoom

WofWca:
#skip-changelog because this is an experimental feature

### DIFF
--- a/packages/target-electron/src/windows/video-call.ts
+++ b/packages/target-electron/src/windows/video-call.ts
@@ -13,6 +13,7 @@ import { getLogger } from '@deltachat-desktop/shared/logger'
 import { getDCJsonrpcRemote } from '../ipc'
 import { pathToFileURL } from 'node:url'
 import type { T } from '@deltachat/jsonrpc-client'
+import { tx } from '../load-translations'
 
 const log = getLogger('windows/video-call')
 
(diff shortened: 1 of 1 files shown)


nicodh (commented):
I still don't see a menu with these changes?

WofWca:
This is because I set autohide to true.
I think these are most useful as keyboard shortcuts, so this is the primary point of this change.

WofWca:
Oh, I didn't mention. You have to press Alt to show the menu.

nicodh:
> Oh, I didn't mention. You have to press Alt to show the menu.

So is that meant as "hidden feature"? You won't find it if you don't know the shortcut.
The only shortcuts that do not work yet (without menu) are Ctr++ to zoom in and Ctrl+0 to reset the zoom.

But I don't mind to have that in the code, might be extended later.

WofWca:
Well, I guess one can put it that way, but I don't think it's that hidden. Menu auto-hide is pretty common IMO, together with the Alt button. Plus, the shortcuts are also common. For example in browsers nowadays there is no such visible menu to toggle the zoom level.

WofWca:
> The only shortcuts that do not work yet (without menu) are Ctr++ to zoom in and Ctrl+0 to reset the zoom.

No, this MR also fixes fullscreen, at least on Windows.

nicodh:
> > The only shortcuts that do not work yet (without menu) are Ctr++ to zoom in and Ctrl+0 to reset the zoom.
> 
> No, this MR also fixes fullscreen, at least on Windows.

What I meant: with this change F11 for full screen and Ctrl-- to zoom out work (and didn't before), but Ctr++ to zoom in and Ctrl+0 don't work (at least on Linux)

Edit: Ctrl+0 works but Ctrl++ not

WofWca:
It has to be the usual Ctrl + Shift + "+"

PR #5491, opened 2025-09-27
Improve display of call messages

nicodh:
Changes:

1. adds phone icons to messages
2. changes outgoing meta data color in dark mode
3. disable edit & forward for call messages

Reasons:
1. Similar to https://github.com/deltachat/deltachat-android/pull/3920 although the UI might still change
2. we had the same color (green) for dark & light mode, but the green has very low contrast to a dark background, the change is more similar to the color in Android
3. call messages should not be forwardable & editable - could be confusing since call messages are more some kind of log

TBD: I used a "filled" phone icon since it seemed to me more visible. Maybe we replace the icon in the navbar also?

image

Dark mode:
image

#skip-changelog as this is part of an experimental WIP feature

### DIFF
--- a/packages/frontend/scss/message/_message.scss
+++ b/packages/frontend/scss/message/_message.scss
@@ -100,7 +100,8 @@ $info-text-max-width: 400px;
       }
       &.call {
         > .text {
-          margin-inline-end: 24px;
+          --call-icon-size: 24px;
+          margin-inline-end: var(--call-icon-size);
           white-space: nowrap;
           font-weight: bold;
           span:last-child {
@@ -111,8 +112,8 @@ $info-text-max-width: 400px;
             position: absolute;
             top: 18px;
             transform: translateY(-50%);
-            width: 24px;
-            height: 24px;
+            width: var(--call-icon-size);
+            height: var(--call-icon-size);
             @include mixins.color-svg(
               './images/icons/phone_filled.svg',
               var(--messageOutgoingStatusColor),
--- a/packages/frontend/scss/message/_message.scss
+++ b/packages/frontend/scss/message/_message.scss
@@ -104,13 +104,11 @@ $info-text-max-width: 400px;
           margin-inline-end: var(--call-icon-size);
           white-space: nowrap;
           font-weight: bold;
-          span:last-child {
-            margin-inline-end: 6px;
-          }
           &::after {
             content: '';
             position: absolute;
             top: 18px;
+            inset-inline-end: 8px;
             transform: translateY(-50%);
             width: var(--call-icon-size);
             height: var(--call-icon-size);
--- a/packages/frontend/scss/message/_message.scss
+++ b/packages/frontend/scss/message/_message.scss
@@ -98,6 +98,29 @@ $info-text-max-width: 400px;
           cursor: pointer;
         }
       }
+      &.call {
+        > .text {
+          margin-inline-end: 24px;
+          white-space: nowrap;
+          font-weight: bold;
+          span:last-child {
+            margin-inline-end: 6px;
+          }
+          &::after {
+            content: '';
+            position: absolute;
+            top: 18px;
+            transform: translateY(-50%);
+            width: 24px;
+            height: 24px;
+            @include mixins.color-svg(
+              './images/icons/phone_filled.svg',
+              var(--messageOutgoingStatusColor),
+              100%
+            );
+          }
+        }
+      }
 
       & > .text {
         color: var(--messageText);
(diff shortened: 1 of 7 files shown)


packages/frontend/scss/message/_message.scss
```diff
@@ -98,6 +98,29 @@ $info-text-max-width: 400px;
...
+          font-weight: bold;
+          span:last-child {
+            margin-inline-end: 6px;
+          }
+          &::after {
+            content: '';
+            position: absolute;
+            top: 18px;
+            transform: translateY(-50%);
+            width: 24px;
```
WofWca: Some vars would be nice

packages/frontend/scss/message/_message.scss
```diff
@@ -98,6 +98,29 @@ $info-text-max-width: 400px;
           cursor: pointer;
         }
       }
+      &.call {
+        > .text {
+          margin-inline-end: 24px;
+          white-space: nowrap;
```
WofWca, line 105: Do these always fit on narrow screens? And for all locales?

packages/frontend/themes/dark.scss
```diff
@@ -6,6 +6,7 @@
   $bgImage: url('./images/background_dark.svg'),
   $bgMessageBubbleIncoming: #303030,
   $bgMessageBubbleOutgoing: #354535,
+  $messageOutgoingStatusColor: #cdb7b5,
```
WofWca: Still seems too dark to me, maybe doesn't even pass accessibility contrast test?

WofWca (commented):
Some preliminary comments.

packages/frontend/themes/dark.scss
nicodh: Well I "borroughed" the color from Android.

I set it to #e4d4d3 which succeeds in the contrast test but is very close to the normal text color.

packages/frontend/scss/message/_message.scss
nicodh: I can't test it for all languages but the longest possible text is "Outgoing call <1 minute" which means 3 words it should be sufficient.

I did not apply different styles for rtl languages since it's more a question of taste wether the icon should be before he text or after

packages/frontend/scss/message/_message.scss
nicodh: Hmm, IDK, but if you prefer I don't mind

packages/frontend/scss/message/_message.scss
```diff
@@ -98,6 +98,30 @@ $info-text-max-width: 400px;
...
+          }
+          &::after {
+            content: '';
+            position: absolute;
+            top: 18px;
+            transform: translateY(-50%);
+            width: var(--call-icon-size);
+            height: var(--call-icon-size);
+            @include mixins.color-svg(
+              './images/icons/phone_filled.svg',
```
WofWca: Is it just me or is the icon missing?

packages/frontend/scss/message/_message.scss
WofWca: Oh, so the point of this is to "remove" the newline that is returned from Core. This doesn't seem like the right approach TBH, but OK.

`white-space: normal;` would be better then IMO, in order to allow breaking, but then the icons starts getting displayed in an incorrect place.

packages/frontend/themes/_themebase.scss
```diff
@@ -135,12 +136,13 @@ $scrollbarTransparency: 0.34 !default;
   --infoMessageBubbleText: white;
   --messageIncomingBg: #{$bgMessageBubbleIncoming};
   --messageOutgoingBg: #{$bgMessageBubbleOutgoing};
-  --messageOutgoingStatusColor: #4caf50; // Message Bubble - Buttons
+  --messageOutgoingStatusColor: #{$messageOutgoingStatusColor}; // Message Bubble - Buttons
```
WofWca, line 139: The comment is not relevant since a long time BTW (but this is not about this MR).

packages/frontend/scss/message/_message.scss
```diff
@@ -98,6 +98,30 @@ $info-text-max-width: 400px;
...
         }
       }
+      &.call {
+        > .text {
+          --call-icon-size: 24px;
+          margin-inline-end: var(--call-icon-size);
+          white-space: nowrap;
+          font-weight: bold;
+          span:last-child {
+            margin-inline-end: 6px;
```
WofWca: Why not just add these 6px to `.text`? The selector looks rather fragile.

packages/frontend/themes/dark.scss
WofWca: Hm yea, this is very close to existing text color. At this point I'm not sure if we even need this color then.

image

packages/frontend/scss/message/_message.scss
nicodh: Ah yes, I wanted to suggest to use another icon, with filled ends for better visibility.

I will replace the existing phone icon and check it in. Let me know what you think.

WofWca: IDK, one might say that it stands out a little compared to the other "not filled" icons.

image

But maybe it's the other icons that need to be changed then. It's fine either way.

WofWca (approved):
The overflow will probably need to get fixed eventually. But this is better for the majority of cases.

PR #5492, opened 2025-09-28
Replace deprecated MutableRefObject

nicodh:
#skip-changelog not changes any code, just a deprecated type definition


WofWca (approved):
This MR only affects types.
The JSDoc of `MutableRefObject` says to use `RefObject`, without details. So I guess it's good.

PR #5494, opened 2025-09-29
Improvements to Dialog Bottom Margin

Simon-Laux:
- **remove dialog margin**
- **re-add dialog margin to settings. It is a seperate element, because that way it is inside of the scroll area, not taking away from it, only visible when you scrolled to the end.**
- **reintroduce bottom space for gallery, again by adding the margin to the scroll content and not the container**
- **add spacing to remaining dialogs. using br line break beacuse it works just fine and is easy to do and understand. again, idea here is to add it to the content and not outside of it for the case that the dialog is scrollable.**

Before this pr scroll-able content vanishes at an invisible line:
image
After this pr the content is visible until the edge of the dialog:
image

### DIFF
--- a/packages/frontend/src/components/Settings/Settings.tsx
+++ b/packages/frontend/src/components/Settings/Settings.tsx
@@ -14,7 +14,7 @@ import Advanced from './Advanced'
 import Profile from './Profile'
 import Dialog, { DialogBody, DialogHeader } from '../Dialog'
 import EditProfileDialog from '../dialogs/EditProfileDialog'
-import SettingsSeparator from './SettingsSeparator'
+import SettingsSeparator, { SettingsEndSeparator } from './SettingsSeparator'
 import useDialog from '../../hooks/dialog/useDialog'
 import useTranslationFunction from '../../hooks/useTranslationFunction'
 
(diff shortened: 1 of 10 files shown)


packages/frontend/src/components/Settings/Settings.tsx
```diff
@@ -14,7 +14,7 @@ import Advanced from './Advanced'
 import Profile from './Profile'
 import Dialog, { DialogBody, DialogHeader } from '../Dialog'
 import EditProfileDialog from '../dialogs/EditProfileDialog'
-import SettingsSeparator from './SettingsSeparator'
+import SettingsSeparator, { SettingsEndSeparator } from './SettingsSeparator'
```
nicodh, line 17: Tbh I don't like the idea to use an extra component just to add some styling. A margin and a border color should be applied via css instead the  
```
<hr  className={styles.settingsSeparator} />
```
 component.
 
Is there a specific reason like accessibility?

Simon-Laux: reason is that we already do this for the other separator, so did the same thing for consistency in the settings. also this is less to write/autocomplete.

Simon-Laux:
> To me the previous display looked as well as the new one. Is it really worth to add these additional SettingsEndSeparator and br tags?

the setting screen visually looks the same, but when it is too large for the dialog then you see the difference.

PR #5496, opened 2025-09-30
allow absolute DC_ACCOUNTS_DIR in browser edition

Simon-Laux:
this makes it possible/easier to check accounts of bots for debugging.


nicodh (approved):
Browser package is for development only so fine to add that imo

PR #5498, opened 2025-09-30
Display message error even for messages that are not OutFailed

link2xt:
Message may have an error not only whet it is in OutFailed state. Since change <https://github.com/chatmail/core/pull/7222> incoming messages may have an error if they fail to download.

With this change the error is displayed when there is an error regardless of the message status.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
   - do not create a group if the sender includes self in the To field
 
 ### Fixed
+- Display message error even for messages that are not OutFailed #5498
 - fix app picker sometimes incorrectly showing "Offline"
 - allow using the app picker even when offline, as long as the app store data is cached
 - if adding an app from the app picker fails, show an error dialog
--- a/packages/frontend/src/components/message/Message.tsx
+++ b/packages/frontend/src/components/message/Message.tsx
@@ -763,6 +763,7 @@ export default function Message(props: {
             fileMime={message.fileMime || null}
             direction={direction}
             status={status}
+            error={message.error || null}
             downloadState={message.downloadState}
             isEdited={message.isEdited}
             hasText={text !== null && text !== ''}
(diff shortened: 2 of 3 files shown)


link2xt:
This is how failed message looks like with this change:
image

Once UI display errors, we will be able to remove "[!]" from the message text.

nicodh:
> Once UI display errors, we will be able to remove "[!]" from the message text.

What kind of display errors do you mean? You mean make it more obvious than the red exclamation mark in the right lower corner? I agree that is no very obvious...

link2xt:
> Not sure how to test that. Maybe by some manipulation in the database?

I have tested by creating a wrapper for the core naed `rpc-server-wrapper.sh`:
```
#!/usr/bin/env bash
export DELTACHAT_FAIL_ON_RECEIVING_FULL_MSG=1
exec /home/user/.cargo/bin/deltachat-rpc-server
```

Then started desktop with `DELTA_CHAT_RPC_SERVER=$(which rpc-server-wrapper.sh) pnpm -w dev:electron --allow-unsafe-core-replacement`. Environment variable `DELTACHAT_FAIL_ON_RECEIVING_FULL_MSG` temporary sets `fail_on_receiving_full_msg` option that makes reception of all messages fail for testing with the latest `main` core.

>> Once UI display errors, we will be able to remove "[!]" from the message text.
>
> What kind of display errors do you mean? You mean make it more obvious than the red exclamation mark in the right lower corner? I agree that is no very obvious...

Without this PR exclamation mark is not shown in the lower right corner. This is why in the core we add `[❗]` to the message. This is why in the screenshot there is an exclamation mark in the lower right corner and in the message. We will be able to remove `[❗]`  from the message text in the core once this PR is merged.

Error display in the lower right corner is good enough, it was just not there at all for incoming messages without this PR.

PR #5499, opened 2025-09-30
upgrade electron from `37.1.0` to `37.6.0` to fix high GPU usage bugs on macOS Tahoe

### LINKED ISSUE #5497: Update electron to latest version
This will help macOS Tahoe users, as electron before v39.0.0-alpha.7, v38.2.0, v37.6.0 and v36.9.2 contains bug with shadows that drives WindowServer process crazy.

Simon-Laux:
closes #5497

why not upgrade electron to a newer major version like 39?
doing that would require more reading of changlog and testing, because it is a bigger change, so for now I think it's better to just get the bugfix that we can still release fast without much testing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
   - withdraw all QR codes when one is withdrawn
   - do not create a group if the sender includes self in the To field
 - minor visual improvements to keyboard shortcut preview #5495
+- upgrade electron from `37.1.0` to `37.6.0`
+  - this fixes high GPU usage bugs on macOS Tahoe
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
   - withdraw all QR codes when one is withdrawn
   - do not create a group if the sender includes self in the To field
 - minor visual improvements to keyboard shortcut preview #5495
+- upgrade electron from `37.1.0` to `37.6.0` #5499
```
WofWca: Probably should be in "Fixed" then.

nicodh (approved):

WofWca (approved):

PR #5500, opened 2025-10-01
refactor: simplify `video-calls.ts` (chatId)

WofWca:
We can now get the `chatId` directly, instead of asynchronously
determening it from `messageId`.
This allows us to simplify the host name
and remove the `getChatInfo` helper function.

Note that now all calls in the same chat have the same origin.
Which is OK.

This also renames a type `T` which was conflicting. This became evident to me just after these changes.

TODO:
- [x] Updgrade core, this MR depends on that. Then rebase and test.


nicodh (approved):

PR #5501, opened 2025-10-01
fix: WebXDC apps having `isSecureContext == false`

WofWca:
This has probably been introduced in
a570afac0f9396307feaeb63ed394dba3c15108f.
Apparently we shouldn't call `protocol.registerSchemesAsPrivileged`
twice, because the previous one gets overridden.

Thanks to @hpk42 and @ralphtheninja for discovering this, and apologies for carelessness.

#skip-changelog because the bug has not been released.

I have verified that this fixes it, by running the Webxdc Test app.

### DIFF
--- a/packages/target-electron/src/index.ts
+++ b/packages/target-electron/src/index.ts
@@ -32,6 +32,7 @@ if (rc['help'] === true || rc['h'] === true) {
   process.exit()
 }
 
+import { callsWebappElectronScheme } from './windows/video-call.js'
 protocol.registerSchemesAsPrivileged([
   {
     scheme: 'webxdc',
@@ -64,9 +65,8 @@ protocol.registerSchemesAsPrivileged([
       stream: true, // needed for audio playback
     },
   },
+  callsWebappElectronScheme,
 ])
-import { registerCallsWebappSchemeAsPrivileged } from './windows/video-call.js'
-registerCallsWebappSchemeAsPrivileged()
 
 const app = rawApp as ExtendedAppMainProcess
 app.rc = rc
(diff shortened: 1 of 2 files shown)


packages/target-electron/src/index.ts
```diff
@@ -32,6 +32,7 @@ if (rc['help'] === true || rc['h'] === true) {
   process.exit()
 }
 
+import { callsWebappElectronScheme } from './windows/video-call.js'
 protocol.registerSchemesAsPrivileged([
```
nicodh, line 36:
```suggestion
// Note: make sure to call registerSchemesAsPrivileged only once with all schemes
// you want to register, since following calls will override previous registered schemes
protocol.registerSchemesAsPrivileged([
```

WofWca:
I'd put it upstream, in Electron docs.

ralphtheninja:
> I have verified that this fixes it, by running the Webxdc Test app.

Does the webxdc test app also check for `window.crypto.subtle`?

ralphtheninja:
I can confirm this branch fixes the problems in the `TOTP` app.

nicodh:
> Does the webxdc test app also check for `window.crypto.subtle`?

No, but it shows wether the isSecureContext is true or false (in the Navigator section)

packages/target-electron/src/index.ts
WofWca: OK, I re-checked, and this is already in the [documentation](https://www.electronjs.org/docs/latest/api/protocol#protocolregisterschemesasprivilegedcustomschemes) of the method:

> can be called only once

WofWca: So I suggest to merge without adding this comment.

PR #3813, opened 2024-05-10
simon/i2872 update message parser

Simon-Laux:
- **update `@deltachat/message_parser_wasm` to `0.10.0`**
   - now parses internationalised Links, some minor bugfixes, added emoji related helper functions
- **replace emoji regex with message-parser function**
- fix: quoted messages with single emoji as text look too big

closes https://github.com/deltachat/deltachat-desktop/issues/2872
closes #3806

PR #833, opened 2019-06-04
Refactor messagebody + subset of markdown

Simon-Laux:
(`**bold**`,`*italics*`,`~~striketrough~~`,`__underline__`)
(**bold**,*italics*,~~striketrough~~,__underline__)

Closes https://github.com/deltachat/deltachat-desktop/issues/491

PR #5502, opened 2025-10-02
Use Linkifyjs as message parser

nicodh:
Mainly for following reasons:

- the old message parser was a separate project that was not really maintained anymore
- there were several bugs with parsing links not being fixed for a long time
- it was not easy to understand the whole code to fix stuff
- the lib needed a special import handling via rust => wasm
- the new lib supports mentions (to be used in future)

Downsides: Markdown is not supported anymore, but it was experimental, only worked in desktop and only if both user had it enabled

### DIFF
--- a/packages/frontend/src/tests/botcommand.test.ts
+++ b/packages/frontend/src/tests/botcommand.test.ts
@@ -93,7 +93,7 @@ describe('linkify-plugin-botcommand', () => {
 
     // although we want to limit the bot commands to word boundaries,
     // it is not possible to do so with the current implementation of linkify.
-    // this test prooves that this is still the case and our filter workaround
+    // this test proves that this is still the case and our filter workaround
     // is needed.
     it('Is not limited to word boundaries', () => {
       expect(
--- a/packages/shared/shared-types.d.ts
+++ b/packages/shared/shared-types.d.ts
@@ -61,6 +61,8 @@ export interface DesktopSettingsType {
   syncAllAccounts: boolean
   /** @deprecated The last used file location for the save dialog is now only kept in memory and not persisted anymore between sessions. */
   lastSaveDialogLocation: string | undefined
+  /** @deprecated */
+  experimentalEnableMarkdownInMessages?: boolean
   enableWebxdcDevTools: boolean
   /** set to false to disable the confirmation dialog for loading remote content */
   HTMLEmailAskForRemoteLoadingConfirmation: boolean


Raiden-GH:
Sorry to interfere, but aren't there any good alternatives to Linkifyjs that support Markdown? 
Markdown support is often requested by users in the forum and groups. I also use it frequently myself. 
Many users will probably consider it a step backward that Markdown support is now disappearing completely, especially since most messengers offer the option to format text.

nicodh:
> Sorry to interfere, but aren't there any good alternatives to Linkifyjs that support Markdown? Markdown support is often requested by users in the forum and groups. I also use it frequently myself. Many users will probably consider it a step backward that Markdown support is now disappearing completely, especially considering that most messengers offer the option to format text.

Well we also offered it only partly. So you could only use it if you knew the other user also had it enabled and was on desktop too.

Yes that feature might come back some day but for now I didn't find a solid, lightweight library that supports both.

nicodh:
The failing code check is not related to this PR. I added some fixes here: https://github.com/deltachat/deltachat-desktop/pull/5503

Simon-Laux:
<details><summary>FTR: some context/clarification about the last two of the listed problems</summary>

> the lib needed a special import handling via rust => wasm

plan was to add it to core, but I blocked that for desktop, because I mistakenly thought, that it would be faster and better to run the parsing in the ui thread via wasm, but blocking the ui thread is bad, because it can result in lagging and is not multi threaded.
But long story short: plan was to have it in core and wasm was only a way to use it before we got it integrated into core.

> the new lib supports mentions (to be used in future)

message-parser supports those too, pr is in project resurrection: https://github.com/deltachat/message-parser/pull/49 

But it is not just parsing "@username" anyway, there will be some challenges since usernames are not unique and emails can change and thus the wire and storage format can't rely on them, but those are to be discussed in the future when we actually work on it.

</details>

---

linkifyjs fixes link parsing and markdown support was not that good anyway (https://github.com/deltachat/message-parser/issues/23), so I agree with the decision to go for linkifyjs as the next logical step to fix the issues and reduce complexity.
Personally I will also miss markdown, but I estimate it will come back better and cross-platform in a few years.

Raiden-GH:
> but I estimate it will come back better and cross-platform in a few years.

In a few **years**? 😢

nicodh:
> > but I estimate it will come back better and cross-platform in a few years.
> 
> In a few **years**? 😢

That's a very pessimistic estimation...

WofWca:
So is this ready for review?

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@
   - withdraw all QR codes when one is withdrawn
   - do not create a group if the sender includes self in the To field
 - minor visual improvements to keyboard shortcut preview #5495
+- replace rust/wasm message-parser with javascript library
```
Simon-Laux: should say which js lib

packages/frontend/package.json
```diff
@@ -50,20 +56,25 @@
...
+    "@types/mocha": "^10.0.7",
+    "@types/node": "catalog:",
     "@types/react": "^19.1.0",
     "@types/react-dom": "^19.1.0",
     "@types/react-window": "^1.8.8",
     "@types/react-window-infinite-loader": "^1.0.9",
+    "chai": "^5.1.1",
     "esbuild": "^0.25.0",
     "esbuild-plugin-inline-worker": "^0.1.1",
+    "mocha": "^10.7.0",
```
Simon-Laux, line 70: the "shared" package imports those as well, better switch/migrate to pnpm workspace dependencies to avoid duplicated dependencies, by making sure all packages use the same version.

nicodh:
The current state of this is (afaik):
- it parses links, emails, hashes and bot commands
- it also renders domains & Urls without protocol as links
```
example.com becomes <a href="https://example.com">example.com</a>
```


TODO:
it has some known bugs, maybe most important is https://github.com/nfrasser/linkifyjs/issues/476
which means it renders fediverse addresses wrong(!) since it renders the part after the first @ as email adddress 
(our current message parser renders fediverse addresses just as text)

But there is a possible workaround in PR #5526

We should also carefully check if the performance is good with this new library

packages/frontend/src/components/message/MessageParser.tsx
```diff
@@ -1,75 +1,116 @@
...
-})
+const log = getLogger('renderer/message-parser')
+
+/**
+ * returns an array with emojis if the first token of str
+ * is of type emoji, composed emojis are just one item in the array
+ */
+export function extractEmojisFromFirstToken(
+  str: string,
+  emojiOnly = false // only return emojis if the whole string is just one emoji token
```
WofWca, line 26: Wouldn't it be more appropriate to make a separate function? So far I don't really understand the body of the function, but its interface seems overly complex, and the name seems to be correct for only the `emojiOnly = true` usage. I mean, why do we need it to returns the list of emojis? Is there no simpler way to write it? Without using `Segmenter`s and stuff?

packages/frontend/package.json
```diff
@@ -15,7 +15,8 @@
     "watch:ts": "pnpm build:ts -- -w",
     "watch:types": "npx tsc --noEmit --pretty -w --preserveWatchOutput",
     "watch:static": "node ../../bin/copy.js ../../static ./html-dist & node ../../bin/copy.js ./static ./html-dist & node ../../bin/copy.js ../../images ./html-dist/images",
-    "watch:scss": "pnpm build:scss --watch"
+    "watch:scss": "pnpm build:scss --watch",
+    "test": "NODE_OPTIONS=\"--enable-source-maps --import ./bin/register-ts-node.mjs\" mocha 'src/tests/**/*.ts'"
```
WofWca, line 19: It would be nice to add this to CI I guess?

packages/frontend/package.json
```diff
@@ -25,17 +26,20 @@
...
+    "@types/linkifyjs": "^2.1.7",
     "classnames": "^2.5.1",
     "debounce": "^1.2.1",
     "emoji-js-clean": "^4.0.0",
     "emoji-mart": "^5.6.0",
     "filesize": "^10.1.4",
     "immutable": "^4.3.7",
     "jsqr": "^1.4.0",
+    "linkify-plugin-hashtag": "^4.3.2",
+    "linkify-plugin-mention": "^4.3.2",
```
WofWca: Isn't this unused?

```suggestion
```

The lock file will also need to be updated.

packages/frontend/src/components/message/MessageParser.tsx
```diff
@@ -1,75 +1,116 @@
...
+      }
+      const url = new URL(fullUrl)
+      let suspicousUrl = false
+      const stripLastSlash = (url: string) => {
+        if (url.endsWith('/')) {
+          url = url.slice(0, -1)
+        }
+        return url
+      }
+      // according to https://developer.mozilla.org/de/docs/Web/API/URL/hostname
```
WofWca:
```suggestion
      // according to https://developer.mozilla.org/docs/Web/API/URL/hostname
```

packages/frontend/src/utils/linkify-plugin-bot-command/botCommand.ts
```diff
@@ -0,0 +1,67 @@
...
+const BotCommandToken = createTokenClass('botcommand', {
+  isLink: true,
+  toHref() {
+    return this.toString()
+  },
+})
+
+/**
+ * Bot command parser plugin for linkify
+ * Finds strings with a single trailing slash at the beginning of a word
```
WofWca:
```suggestion
 * Finds strings with a single leading slash at the beginning of a word
```

packages/frontend/src/components/message/MessageParser.tsx
```diff
@@ -1,75 +1,116 @@
...
+              ascii_hostname: url.hostname,
+              punycode_encoded_url: url.href,
+              original_hostname: elm.v, // TODO: change naming here
+            }
+          : null,
+        scheme: '',
+        linkText: elm.v,
+      }
+      destination.hostname = url.hostname
+      destination.scheme = url.protocol.replace(':', '')
```
WofWca: Why not just initialize with these values right away?

packages/frontend/src/components/message/MessageParser.tsx
```diff
@@ -126,47 +156,28 @@ function renderElement(
...
 }
 
-/** render in preview mode for ChatListItem summary and for quoted messages,
- *  not interactive (links can not be clicked) just looks more similar to the message in the chatview/message-list */
+/**
+ * render in preview mode for ChatListItem summary and for quoted messages,
+ * not interactive (links can not be clicked) just looks more similar to
+ * the message in the chatview/message-list
+ */
 function renderElementPreview(
```
WofWca: It seems that we don't need this function anymore or am I missing something?

packages/frontend/package.json
nicodh: It is already. pnpm test in work dir calls all test scripts in all projects https://github.com/deltachat/deltachat-desktop/blob/d10aaf362bfeffd58a16277b75a0d41756b677af/.github/workflows/test.yml#L34

packages/frontend/src/components/message/MessageParser.tsx
nicodh: It is used to handle composed emojis which have a different string length than not composed emojis. The Segmenter is a way to distinguish the "segments" in the string, where a composed emoji starts and ends.
I didn't want to repeat the Segmenter stuff and so I combined the 2 functions that need it into one.
Otherwise we would have 3 functions, one for the splitting which is called by getEmojiCount and getFirstEmoji

packages/frontend/src/components/message/MessageParser.tsx
nicodh: It is in use in the parseAndRenderMessage function if the second parameter is true

WofWca: I see that it's in use, I'm saying that the function doesn't do anything really, and we could just output the raw text. No need to "recognize newlines".

packages/frontend/src/components/message/MessageParser.tsx
```diff
@@ -101,20 +131,20 @@ function renderElement(
...
-    case 'Linebreak':
+    case 'nl':
       return <span key={key}>{'\n'}</span>
 
-    case 'Text':
-      return <span key={key}>{elm.c}</span>
+    case 'text':
+      return <span key={key}>{elm.v}</span>
     default:
       //@ts-ignore
```
WofWca: BTW these `@ts-ignore` don't seem to be needed (anymore)?

packages/frontend/src/components/message/MessageParser.tsx
nicodh: You are right. It's not needed anymore

packages/frontend/src/components/message/MessageParser.tsx
WofWca: But the function name and the docstring are still perhaps too implementation-focused, aren't they? The user doesn't need to know about "token"s, WDYT?

packages/frontend/src/components/message/MessageParser.tsx
Simon-Laux, line 1: Some background on why I think this was here:
I guess this was an attempt to use exhaustive switch checking: https://www.typescriptlang.org/docs/handbook/2/narrowing.html#exhaustiveness-checking

But there is also the question if linkifyjs provides a specific types (`element_type = 'text'|'url'|'hashtag'>....`) or just a generic type of string (`element_type = string`).

packages/frontend/src/components/message/MessageParser.tsx
Simon-Laux: unicode seems to call that a grapheme cluster, likify parser outputs tokens. others seem to call it "perceived" emojis.

I agree that terminology is a bit confusing, also I think the method signature is a bit hard to understand at first.

packages/frontend/src/components/message/MessageParser.tsx
Simon-Laux: context: this comment is about removal of `//@ts-ignore` in default case of the switch statement

packages/frontend/package.json
nicodh: I agree but will do that in a separate MR after this is merged

packages/frontend/src/components/message/MessageParser.tsx
nicodh: I renamed it to extractFirstEmojisFromString

packages/frontend/src/tests/botcommand.test.ts
```diff
@@ -0,0 +1,113 @@
...
+    it('Does not work with just a slash', () => {
+      expect(linkify.test('/', 'botcommand')).to.equal(false)
+    })
+
+    it('Does not work with slashes not at word boundaries', () => {
+      expect(linkify.test('test/help', 'botcommand')).to.equal(false)
+      expect(linkify.test('path/to/file', 'botcommand')).to.equal(false)
+      expect(linkify.test('1/test', 'botcommand')).to.equal(false)
+      expect(linkify.test('_/test', 'botcommand')).to.equal(false)
+    })
```
WofWca: These tests seem incorrect, and they don't catch what I think is a bug: (a string `I like something/somethingelse` contains a bot command, according to the plugin. Looks like what we're supposed to be testing is whether the string contains a bot command, and not whether the whole string is a bot command.

CHANGELOG.md
```diff
@@ -29,6 +29,9 @@
   - withdraw all QR codes when one is withdrawn
   - do not create a group if the sender includes self in the To field
 - minor visual improvements to keyboard shortcut preview #5495
+- replace rust/wasm message-parser with javascript lib linkify
```
WofWca: When the MR is approved, yet another CHANGELOG update will need to happen, to move this to the appropriate place (why haven't we abandoned this practice yet?)

packages/frontend/package.json
nicodh: see #5560

packages/frontend/src/tests/botcommand.test.ts
nicodh: Ok, this is a blocker. Bot commands should be marked only when at a "word start" i.e. in the beginning of a string or after a space

nicodh: Made a fix

packages/frontend/src/utils/linkify/parseElements.ts
```diff
@@ -0,0 +1,30 @@
...
+        if (start > 0 && !/\s/.test(message[start - 1])) {
+          element.t = 'text'
+        }
+      }
+      // here we could also disable email links in fediverse mentions
+      // or hashtags that have no preceeding space
+      return element
+    })
+    return elements
+  } catch (error) {
```
WofWca, line 25: Do we ever expect this to fail? If so, I think we should render the original message text. I.e. maybe just return one "text" node.

packages/frontend/src/tests/botcommand.test.ts
```diff
@@ -89,25 +91,149 @@ describe('linkify-plugin-botcommand', () => {
       expect(linkify.test('/', 'botcommand')).to.equal(false)
     })
 
-    it('Does not work with slashes not at word boundaries', () => {
-      expect(linkify.test('test/help', 'botcommand')).to.equal(false)
-      expect(linkify.test('path/to/file', 'botcommand')).to.equal(false)
-      expect(linkify.test('1/test', 'botcommand')).to.equal(false)
-      expect(linkify.test('_/test', 'botcommand')).to.equal(false)
+    // check that the filter is needed - it should not parse bot
```
WofWca: What does "check that the filter is needed" mean here?

packages/frontend/src/utils/linkify/parseElements.ts
```diff
@@ -0,0 +1,30 @@
...
+export function parseElements(message: string): linkify.MultiToken[] {
+  try {
+    const elements = linkify.tokenize(message).map(element => {
+      if (element.t === 'botcommand') {
+        const start = element.startIndex()
+        // do not render botcommands if they
+        // not start at index 0 or after a space
+        if (start > 0 && !/\s/.test(message[start - 1])) {
+          element.t = 'text'
+        }
```
WofWca, line 18: So is it not possible to achieve this inside the Linkify plugin itself?

nicodh: Not by a plugin it seems. I tried various approaches. The problem is that once a new state is started  in the Finit-State-Machine pattern that is used here, there is no access to the previous and there is no config option that can be passed to be executed in the parser to check the previous token.

I had a look why it is working for https, but it turned out that "https" is a row of characters registered as SLASH_SCHEME and is only detected if it builds a complete token which is not the case if it doesn't start after a space (then it's just part of a WORD token).

We can't do the same for just a single slash followed by multiple other possible characters.

We could ask for a change for that on upstream, but it might be hard to get a feature in that is not needed by the existing plugins/packages. Only case would be if you don't want mentions to be valid if they are not at a word boundary like foo#bar but something like 😀#me should work maybe...

It's questionable to introduce a new library that needs "workarounds" from the very beginning but the workaround seems pretty straight forward to me and the lib covers a lot (all?) other cases we have right now.

WofWca: > feature in that is not needed by the existing plugins/packages. Only case would be if you don't want mentions to be valid if they are not at a word boundary

Well this _does_ sound like something that should be done upstream. It hashtags in the middle of a word don't make sense.

But alright, having this workaround is fine.

nicodh: > Well this _does_ sound like something that should be done upstream. It hashtags in the middle of a word don't make sense.

I added an issue

https://github.com/nfrasser/linkifyjs/issues/521

packages/shared/shared-types.d.ts
```diff
@@ -61,7 +61,6 @@ export interface DesktopSettingsType {
   syncAllAccounts: boolean
   /** @deprecated The last used file location for the save dialog is now only kept in memory and not persisted anymore between sessions. */
   lastSaveDialogLocation: string | undefined
-  experimentalEnableMarkdownInMessages: boolean
   enableWebxdcDevTools: boolean
```
WofWca, line 66:
```suggestion
  /** @deprecated */
  experimentalEnableMarkdownInMessages?: boolean
  enableWebxdcDevTools: boolean
```

I think we should not remove the properties from the type definition in such a case, because in reality it still exists: we don't actually clean it up from storage.

packages/frontend/src/tests/botcommand.test.ts
```diff
@@ -0,0 +1,241 @@
...
+      expect(linkify.test('/ConfigDebug', 'botcommand')).to.equal(true)
+    })
+
+    it('Does not work with just a slash', () => {
+      expect(linkify.test('/', 'botcommand')).to.equal(false)
+    })
+
+    // although we want to limit the bot commands to word boundaries,
+    // it is not possible to do so with the current implementation of linkify.
+    // this test prooves that this is still the case and our filter workaround
```
WofWca:
```suggestion
    // this test proves that this is still the case and our filter workaround
```

WofWca:
FTR, this seems to somewhat revert 5e94c780e469f575ff0a031b67bae3bcd17b45c3 (https://github.com/deltachat/deltachat-desktop/pull/833). We used a similar library, `linkify-it`, before that.

Edit: see also
- 0257c50f5ae2217e644ae27af9f3dbf5e28a2127 (https://github.com/deltachat/deltachat-desktop/pull/3813).

PR #5503, opened 2025-10-02
Fix some type errors

nicodh:
Our code validation detected some type errors it didn't detect before...

About the missing translations: I'm not sure if they are translated correct. Seems there were introduced here:
https://github.com/deltachat/deltachat-desktop/commit/46ecc3b0c1ca9f4a8c3bd7106abd70aa17642782#diff-f7e1c983754b9480c25c1dea84f638f2c5e02d29231c0bbd1cac534000c2ac54R89
but never added?  Running pnpm check gave warnings for the missing translations

#skip-changelog

### DIFF
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -91,14 +91,5 @@
   },
   "accounts_list_item_menu_label": {
     "message": "Actions with Profile"
-  },
-  "a11y_delivery_status_in_fresh": {
-    "message": "Message is new"
-  },
-  "a11y_delivery_status_in_seen": {
-    "message": "Message is seen"
-  },
-  "a11y_delivery_status_in_noticed": {
-    "message": "Message was notified"
   }
 }
--- a/packages/frontend/src/components/message/MessageMetaData.tsx
+++ b/packages/frontend/src/components/message/MessageMetaData.tsx
@@ -113,16 +113,15 @@ export default function MessageMetaData(props: Props) {
             <span className='visually-hidden'>
               {tx(
                 `a11y_delivery_status_${
-                  status as Exclude<
-                    typeof status,
-                    // '' is not supposed to happen.
-                    // The others are not supposed to happen
-                    // as long as direction is outgoing.
-                    | ''
-                    | (typeof direction extends 'outgoing'
-                        ? 'in_fresh' | 'in_seen' | 'in_noticed'
-                        : never)
-                  >
+                  error !== null
+                    ? 'error'
+                    : (status as Exclude<
+                        typeof status,
+                        // '' is not supposed to happen.
+                        // The others are not supposed to happen
+                        // as long as direction is outgoing.
+                        '' | 'in_fresh' | 'in_seen' | 'in_noticed'
+                      >)
                 }`
               )}
             </span>


_locales/_untranslated_en.json
```diff
@@ -91,5 +91,14 @@
...
     "message": "Actions with Profile"
+  },
+  "a11y_delivery_status_in_fresh": {
+    "message": "Message is new"
+  },
+  "a11y_delivery_status_in_seen": {
+    "message": "Message is seen"
+  },
+  "a11y_delivery_status_in_noticed": {
+    "message": "Message was notified"
```
WofWca: The error has been introduced in https://github.com/deltachat/deltachat-desktop/pull/5498. The type check that I wrote in 46ecc3b0c1ca9f4a8c3bd7106abd70aa17642782 was supposed to guard against such errors.

I think that the proper way to fix this is to make sure that the (visually hidden) text matches the icon, instead of introducing these strings.

packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
```diff
@@ -154,7 +154,7 @@ export const AudioRecorder = ({
       ?.stop()
       .getMp3()
       .then(([buffer, _blob]) => {
-        saveVoiceAsDraft(new Blob(buffer, { type: 'audio/mp3' }))
+        saveVoiceAsDraft(new Blob(buffer as any, { type: 'audio/mp3' }))
```
WofWca, line 157: It doesn't seem to me like this is the right way to address these type errors. It seems that it would be more correct to specify the concrete types as generic type arguments, e.g. here

https://github.com/deltachat/deltachat-desktop/blob/d887c6f078158fd91c7fc8386a38e1d511a83678/packages/frontend/src/components/AudioRecorder/Encoder.ts#L19

```typescript
private dataBuffer: Int8Array<ArrayBuffer>[] = []
```

But I am fine with this.

Edit: see https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-9.html#libdts-changes

WofWca (changes requested):

packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
nicodh: I tried various types but always ended in typescript errors, so I decided it's not worth here...

_locales/_untranslated_en.json
nicodh: But the text is passed to the tx function and the key is dynamically created. The lines were introduced in 46ecc3b 

What do you mean with " the (visually hidden) text matches the icon"?

WofWca: > But the text is passed to the tx function and the key is dynamically created

TypeScript narrows down the type of `typeof direction`. It used to not include strings starting with ~~`out_`~~ `in_`, before we added the `|| error != null` condition.

> What do you mean with " the (visually hidden) text matches the icon"?

If `error !== null`, we display an error icon, instead of the `status`:

https://github.com/deltachat/deltachat-desktop/blob/d4b70ad6f1e13c2f72218b39e70798245e52eb69/packages/frontend/src/components/message/MessageMetaData.tsx#L110

But this override has not been applied to the text:

https://github.com/deltachat/deltachat-desktop/blob/d4b70ad6f1e13c2f72218b39e70798245e52eb69/packages/frontend/src/components/message/MessageMetaData.tsx#L113-L128

WofWca: The text should always say "error" if the icon is "error".

WofWca: Correction: "It used to not include strings starting with ~~`out_`~~ in_"

nicodh: Ok, having a closer look at this I realized that after Exclude till line 125 it's all about the type restriction. But I don't understand why you exclude  'in_fresh' | 'in_seen' | 'in_noticed' only conditionally. They will/should never occur here?

So this should be sufficient:

```typescript
{tx(
      `a11y_delivery_status_${
        error !== null ? 'error' : status as Exclude<
          typeof status,
          // '' is not supposed to happen.
          // The others are not supposed to happen
          // as long as direction is outgoing.
          | '' | 'in_fresh' | 'in_seen' | 'in_noticed'  
        >
      }`
   )}
  ```

Comment (WofWca) [reply _locales/_untranslated_en.json]: Because this doesn't catch the error. With https://github.com/deltachat/deltachat-desktop/pull/5498, they can in fact occur, TypeScript is right.

Comment (nicodh) [reply _locales/_untranslated_en.json]: Well the condition in line 93 is:
(direction === 'outgoing' || error !== null)

the first condition excludes "in_" states, the second will set the key to 'error' so how can "in_" states reach that part?

Comment (WofWca) [reply _locales/_untranslated_en.json]: Oh. Sorry. I didn't read your code thoroughly and assumed that it only changes types but not behavior.
Still, this assertion is not that strong, it does not catch the errors that it used to catch.

Comment (nicodh) [reply _locales/_untranslated_en.json]: The syntax is really evil to read - that's why I on first read thought these strings would be part of the key, while in fact they just were part of the type exclude

File: packages/frontend/src/stockStrings.ts [hunk 3]
```diff
@@ -20,7 +20,6 @@ export async function updateCoreStrings() {
   strings[C.DC_STR_VIDEO] = tx('video')
   strings[C.DC_STR_AUDIO] = tx('audio')
   strings[C.DC_STR_FILE] = tx('file')
-  strings[C.DC_STR_ENCRYPTEDMSG] = tx('encrypted_message')
```
WofWca, line 23: What error does this fix?

nicodh: It's removed from https://github.com/chatmail/core/blob/main/src/stock_str.rs

I just found it when checking translations. So yes, maybe should not part of this PR but should be removed anyway.
There are some more obsolete strings, so I will make another PR removing more of these

WofWca: `DC_STR_AEAP_EXPLANATION_AND_LINK` has _not_ been removed.

PR #5508, opened 2025-10-03
fix: untranslated stock strings, add type safety

WofWca:
Add type safety and add back some missing strings.

The first commit adds type safety, the second commit adds missing strings.

This partially reverts f0d264b857c3b3176a100a877553ce6d1725107f,
which removed `DC_STR_AEAP_EXPLANATION_AND_LINK`,
which apparently should not have been removed. See
https://github.com/deltachat/deltachat-desktop/pull/5503#discussion_r2401112411.

### DIFF
--- a/packages/frontend/src/stockStrings.ts
+++ b/packages/frontend/src/stockStrings.ts
@@ -129,6 +129,7 @@ export async function updateCoreStrings() {
     [C.DC_STR_SETUP_CONTACT_QR_DESC]: tx('qrshow_join_contact_hint'),
     [C.DC_STR_SECURE_JOIN_GROUP_QR_DESC]: tx('qrshow_join_group_hint'),
     [C.DC_STR_NOT_CONNECTED]: tx('connectivity_not_connected'),
+    [C.DC_STR_AEAP_EXPLANATION_AND_LINK]: tx('aeap_explanation'),
     [C.DC_STR_GROUP_NAME_CHANGED_BY_YOU]: tx('group_name_changed_by_you'),
     [C.DC_STR_GROUP_NAME_CHANGED_BY_OTHER]: tx('group_name_changed_by_other'),
     [C.DC_STR_GROUP_IMAGE_CHANGED_BY_YOU]: tx('group_image_changed_by_you'),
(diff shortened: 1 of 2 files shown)


packages/frontend/src/stockStrings.ts
```diff
@@ -5,183 +5,221 @@ import { BackendRemote } from './backend-com'
...
+    [C.DC_STR_SELF]: tx('self'),
+    [C.DC_STR_DRAFT]: tx('draft'),
+    [C.DC_STR_VOICEMESSAGE]: tx('voice_message'),
+    [C.DC_STR_IMAGE]: tx('image'),
+    [C.DC_STR_GIF]: tx('gif'),
+    [C.DC_STR_VIDEO]: tx('video'),
+    [C.DC_STR_AUDIO]: tx('audio'),
+    [C.DC_STR_FILE]: tx('file'),
+    // [C.DC_STR_E2E_AVAILABLE]: tx('DC_STR_E2E_AVAILABLE'),
+    // [C.DC_STR_ENCR_TRANSP]: tx('DC_STR_ENCR_TRANSP'),
```
nicodh, line 79: Shouldn't we cleanup here now?
It seems DC_STR_ENCR_TRANSP & DC_STR_ENCR_NONE are not needed anymore (and deprecated) while DC_STR_E2E_AVAILABLE is not deprecated and still available in https://github.com/chatmail/core/blob/main/src/stock_str.rs (as E2eAvailable) and used in core

At the same time we have no translation for DC_STR_E2E_AVAILABLE ? 

Our script bin/test_for_missing_translations.sh doesn't check if a translation string is commented so we should avoid having such comments

WofWca: > Shouldn't we cleanup here now?

Maybe. But these commented out strings were explicitly ignored (judging by the fact that they are commented out), so I kept it as is for now. The rest can be done in follow-ups.

> At the same time we have no translation for DC_STR_E2E_AVAILABLE ?

Apparently. git history shows some though, e.g. here it was removed: 332e8eed1ee82b6eb6e12348d5e5d4547f2ccaae.

> Our script bin/test_for_missing_translations.sh doesn't check if a translation string is commented so we should avoid having such comments

Is the script even useful though? What does it do that `getMessageFunction` type ( 46ecc3b0c1ca9f4a8c3bd7106abd70aa17642782 ) doesn't?

nicodh (approved):

PR #5511, opened 2025-10-05
Style refactoring proxy dialog

### LINKED ISSUE #5507: proxy layout broken for some proxy addresses, wrong title
when a proxy address is a bit longer, the layout of the proxy dialog is broken, eg. the "switch off" button is not visible:

Image

the uses proxy address is quite normal in lengths, it is a random pick from some source. 

it appears when scrolling (but one has to know that this is even possible as there is no scrollbar :)

Image

fix is to eg. wrap the address

moreover, feel free to create a separate issue for that, the title should read "Proxy", not "Settings"

nicodh:
resolves #5507 

- make dialog a bit wider (500px)
- use textarea as input with field-sizing: content
- max-width for the proxy label with title showing the full url to keep consistent layout
- remove display: flex of dialog body to keep adaptive width

image


packages/frontend/src/components/dialogs/ProxyConfiguration/styles.module.scss
```diff
@@ -34,6 +30,9 @@
 
 .proxyLabel {
   margin-bottom: 6px;
+  max-width: 320px;
+  overflow: hidden;
+  text-overflow: ellipsis;
```
WofWca, line 35: We still have line breaks.

Image

Maybe it's better to either add `white-space: nowrap` or to remove `overflow: hidden` and add `word-break: break-word;`

packages/frontend/scss/login/_login.scss
```diff
@@ -195,6 +195,7 @@ div.delta-form-group {
     font-family: inherit;
     box-shadow: none !important;
     resize: vertical;
+    field-sizing: content;
```
WofWca, line 198: Note that this also affects the "edit profile" dialog:

Image

Looks fine IMO.

packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
```diff
@@ -370,7 +370,7 @@ export default function ProxyConfiguration(
       <DialogFooter>
         {showNewProxyForm && (
           <form>
-            <DeltaInput
+            <DeltaTextarea
```
WofWca, line 373: Now one cannot submit the form with "Enter". I would say that keeping this an `<input>` is better. One does not usually need to edit the proxy line much.

packages/frontend/src/components/dialogs/ProxyConfiguration/styles.module.scss
```diff
@@ -34,6 +30,9 @@
 
 .proxyLabel {
   margin-bottom: 6px;
+  max-width: 320px;
```
WofWca, line 33: This is a wrong approach. The buttons get shifted when this element reaches 320px in width:

Image

WofWca (approved):
This is not a refactor, but it fixes the issue.

packages/frontend/src/components/dialogs/ProxyConfiguration/styles.module.scss
nicodh: Why is that a problem? You can enter line breaks but can't do anything with that.

packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
nicodh: I would prefer to make the whole URL visible by wrapping it into the text area which is not achievable in a text input afaik

packages/frontend/src/components/dialogs/ProxyConfiguration/styles.module.scss
nicodh: I can only reproduce that when I narrow the window to less than 500px which is an extreme edge case. And the form is still usable , only doesn't looks nice. I consider that good enough.

But if you have know one, a better solution would be welcome

WofWca: > I can only reproduce that when I narrow the window to less than 500px

Not the case for me. try playing around with zoom maybe.

> I consider that good enough.

I agree.

packages/frontend/src/components/dialogs/ProxyConfiguration/styles.module.scss
WofWca: WDYM "can't do anything with that"?
Not really a problem, but maybe can be considered inconsistent. Sometimes we have line breaks, sometimes we truncate.

PR #5516, opened 2025-10-06
chore: upgrade Core to 2.17.0

nicodh:


packages/target-electron/src/windows/video-call.ts
```diff
@@ -312,9 +312,7 @@ function openVideoCallWindow<T extends CallDirection>(
     return port1
   })()
 
-  // @ts-expect-error ts(2339) This API will be introduced later.
   if (jsonrpcRemote.rpc.iceServers != undefined) {
-    // @ts-expect-error ts(2339) This API will be introduced later.
```
WofWca, line 317: This doesn't do what I did in my merge request, but I guess I can make a follow-up later.

nicodh: We can also merge the video-call from your branch. This was just to avoid the error: src/windows/video-call.ts(315,3): error TS2578: Unused '@ts-expect-error' directive.

WofWca (approved):
This also upgrades Tauri Rust dependencies.
Otherwise a regular Core upgrade.
The E2E test failure is due to flakiness.

packages/target-electron/src/windows/video-call.ts
WofWca: I'll do it.

PR #5521, opened 2025-10-07
Update translations and add missing stock strings

nicodh:


_locales/_untranslated_en.json
```diff
@@ -91,5 +91,26 @@
...
+    "message": "Missed call"
+  },
+  "call_outgoing": {
+    "message": "Outgoing call"
+  },
+  "channel_left_by_you": {
+    "message": "You left the Channel"
+  },
+  "qrshow_join_channel_hint": {
+    "message": "Scan to join Channel."
```
WofWca: It is pointless to add these to `untranslated.json` because the core apparently provides the default English strings.

https://github.com/chatmail/core/blob/9ec0332483a56b97d0a5a4b00e6ae642cdc670c8/src/stock_str.rs#L428-L444

Or maybe it would work for "future-proofing" where we wait for https://github.com/deltachat/deltachat-android/pull/3785 to get merged, which would add these strings, but then it's not that much of future proofing, because the string names (keys) don't match and we'll still have to update the code.

packages/frontend/src/stockStrings.ts
```diff
@@ -217,8 +216,16 @@ export async function updateCoreStrings() {
     [C.DC_STR_MESSAGE_ADD_MEMBER]: tx('member_x_added'),
     [C.DC_STR_YOU_REACTED]: tx('reaction_by_you'),
     [C.DC_STR_REACTED_BY]: tx('reaction_by_other'),
-    [C.DC_STR_SECUREJOIN_WAIT]: tx('secure_join_wait_timeout'),
+    [C.DC_STR_SECUREJOIN_WAIT]: 'deprecated',
```
WofWca: This is really weird. Why not just add these strings to the "ignore list" above? There is no need to show 'deprecated' to the user instead of the default core's English string.

https://github.com/chatmail/core/blob/9ec0332483a56b97d0a5a4b00e6ae642cdc670c8/src/stock_str.rs#L415-L416

CHANGELOG.md
```diff
@@ -39,8 +40,7 @@
 
 <a id="2_11_1"></a>
 
-## [2.11.1] - 2025-09-01
-
+trans
```
```suggestion
## [2.11.1] - 2025-09-01

```

packages/frontend/src/stockStrings.ts
nicodh: Of course - you are right

CHANGELOG.md
```diff
@@ -39,8 +40,6 @@
 
 <a id="2_11_1"></a>
 
-## [2.11.1] - 2025-09-01
```
WofWca, line 42: Still an issue.

packages/frontend/src/stockStrings.ts
```diff
@@ -62,7 +62,15 @@ export async function updateCoreStrings() {
...
-    | C.DC_STR_CONTACT
+    | C.DC_STR_CHANNEL_LEFT_BY_YOU
+    | C.DC_STR_CANCELED_CALL
+    | C.DC_STR_DECLINED_CALL
+    | C.DC_STR_INCOMING_CALL
+    | C.DC_STR_MISSED_CALL
+    | C.DC_STR_OUTGOING_CALL
+    | C.DC_STR_SECURE_JOIN_CHANNEL_QR_DESC
+    | C.DC_STR_SECUREJOIN_WAIT
+    | C.DC_STR_AEAP_EXPLANATION_AND_LINK
```
WofWca, line 73: I'll create an issue in order to not forget to add these later.

WofWca (approved):
With the changelog fixed, this will be good to merge.

I didn't check the translations themselves.

PR #5526, opened 2025-10-08
Avoid email links in Fediverse mentions (edited)

nicodh:
Linkify renders fediverse links like @user@mastodon.social as 
```
<a href="https://mastodon.social/@user">@user@mastodon.social</a>
```

This is also a test how much effort it is to adapt the linkify library

#skip-changelog as this is still experimental and based on https://github.com/deltachat/deltachat-desktop/pull/5502

#public-preview to enable testing


Simon-Laux:
we decided not to do that in https://github.com/deltachat/message-parser/issues/82, have you seen that?
is there a new discussion on this, that includes mobile support?

r10s:
yip, i am also sceptical about that, for reason linked above, we may want to use a similar format ourselves one day, and this PR would make things even harder.

but also wrt consistency  with android/ios

nicodh:
> yip, i am also sceptical about that, for reason linked above, we may want to use a similar format ourselves one day, and this PR would make things even harder.
> 
> but also wrt consistency with android/ios

Ok, I'm also not opinionated about these links. This started as a fix for the way linkify parses these links (same as github here does):

```
@user@mastodon.social is parsed a @ sign plus an email address
@<a href="mailto:user@mastodon.social">user@mastodon.social</a>
```

Since convertFediverseMentions is just a wrapper it's easy to remove if we don't need it anymore (i.e. if the upstream issue is fixed).

So I would keep it, rename it to fixEmailFediverseLinks and convert Fediverse Links to plain text without click handling to have the exact same result as with the existing parser.

I just saw the "low hanging fruit" to render the links to work as expectd and picked it. :-)

nicodh:
> we may want to use a similar format ourselves one day, and this PR would make things even harder.

I think once we want mentions the first step would be to add a select for contacts if a @ sign is entered. The implementation would probably then be displayed as @username but needs to be @contactId or mention:contactId something similar "under the hood". So it will never be a simple @username - especially since we allow spaces etc. in usernames...

hpk42:
FWIW i am in favor of linkyfing @user@tld.org which i believe is a Mastodon/Pleroma/etc. shortform of RFC7033's  spec of 'acct:user@example.org' . Not rendering it as email address certainly makes sense. 

As to https://github.com/deltachat/message-parser/issues/82 and prior discussion, i agree with @ndh that @-completion shouldn't interfere much with how messages are rendered in the message list. If it causes problems, we can drop it, but it think it's not a problem, really.  Am fine with whatever last word bjoern wants to speak here :)

Simon-Laux:
I also like the idea of fediverse addresses, but since we already removed features for consistency with mobile, I'm only for it when mobile wants to do it as well.

hpk42:
On second thought i am not sure we can derive a canonical link from the '@user@example.org' notation -- it involves RFC7033 web finger network queries.  If that's true, it would require core to resolve it, and means this PR shouldn't parse it as email address and be done.

Simon-Laux:
> If that's true, it would require core to resolve it, and means this PR shouldn't parse it as email address and be done.

a quick jsonrpc api for that would be doable, also there is also a jsonrpc api for http requests.

<details>
<summary>looks simple enough:</summary>

```
curl https://fosstodon.org/.well-known/webfinger?resource=acct:treefit@fosstodon.org
```
```json
{"subject":"acct:treefit@fosstodon.org","aliases":["https://fosstodon.org/@treefit","https://fosstodon.org/users/treefit"],"links":[{"rel":"http://webfinger.net/rel/profile-page","type":"text/html","href":"https://fosstodon.org/@treefit"},{"rel":"self","type":"application/activity+json","href":"https://fosstodon.org/users/treefit"},{"rel":"http://ostatus.org/schema/1.0/subscribe","template":"https://fosstodon.org/authorize_interaction?uri={uri}"},{"rel":"http://webfinger.net/rel/avatar","type":"image/png","href":"https://cdn.fosstodon.org/accounts/avatars/109/183/043/582/226/717/original/d5d371ac42479a2a.png"}]}
```

```sh
$ curl -s https://fosstodon.org/.well-known/webfinger?resource=acct:treefit@fosstodon.org | jq -r '.links[] | select(.rel == "self") | .href'
https://fosstodon.org/users/treefit

$ curl -s https://pixelfed.social/.well-known/webfinger?resource=acct:treefit@pixelfed.social | jq -r '.links[] | select(.rel == "self") | .href'
https://pixelfed.social/users/treefit
```
</details>

So this request can be done when user clicks on it and when it failed it could open a dialog with the error offering to copy the handle.

But only makes sense if all platforms would do it.

hpk42:
sure, web-finger requests are doable, but it's a bit outside of this PR's scope, also given that bjoern already is reserved and android/ios will not jump on it. Can be reconsidered later IMO.

r10s:
> Am fine with whatever last word bjoern wants to speak here :)

thanks :)

beside _not_ rendering as an email address, i would postpone all that. while surely being nice, it is not the most impactful thing to take care for atm. we already have so many moving targets

and i would have said that even without knowing that even core and web requests are needed or  to be discussed :)

WofWca (commented):
I don't think we need this at all, looking at the usefulness to complexity ratio.

I did not verify the correctness of the implementation. But the fact that `clearFediverseMentions` takes the whole message and outputs the whole message means that the function could potentially mess it up in unrelated ways.

The unrelated changes look OK (but I'd prefer them to be a part of a separate MR).

I'll need time to understand the code.

WofWca:
To those confused by the previous comments: this MR was previously about linkifying these Fediverse handles as Fediverse handles instead of emails, but now it's about not linkifying them at all.

nicodh:
>  takes the whole message and outputs the whole message 

It removes the email element and adds the text value and the token to the existing (previous) text element. That's it.

nicodh:
After some discussions with WofWca we decided to close this PR and leave the fediverse mentions as they are - an @ sign and a (non-existing) email link similar to the other clients

Simon-Laux:
> an @ sign and a (non-existing) email link **similar to the other clients**
>
> ~ nico last comment above this one, emphasis: mine

the iOS client doesn't do this and the message-parser also did not do this after this behavior was filed as a bug: https://github.com/deltachat/message-parser/issues/82
In my understanding the expected behavior is do not linkify, display as text.

PR #5527, opened 2025-10-08
adapt to `FullChat.contacts` removal

Simon-Laux:
# TODO
- [ ] we should figure out how to make async dialogs using suspense (with error handling/boundary) to ensure that the dialog does not resize/flicker when opened. 

depends on https://github.com/chatmail/core/pull/7282

### DIFF
--- a/packages/frontend/src/components/dialogs/ViewGroup.tsx
+++ b/packages/frontend/src/components/dialogs/ViewGroup.tsx
@@ -146,23 +146,76 @@ export const useGroup = (accountId: number, chat: T.FullChat) => {
     [tx, openDialog, chat.id, accountId]
   )
 
-  type GroupContacts = typeof group.contacts
-  /**
-   * setGroupContacts is only called after changes from "outside" triggered by
-   * DCEvent "ContactsChanged" to update the group contacts in local state
-   *
-   * It takes a "setter" argument to make sure, the update always happens
-   * on the latest group state
-   */
-  const setGroupContacts = useCallback(
-    (setter: (oldContacts: GroupContacts) => GroupContacts) => {
-      setGroup((group: T.FullChat) => {
-        const newContacts = setter(group.contacts)
-        return { ...group, contacts: newContacts }
+  const [pastContacts, setPastContacts] = useState<T.Contact[]>([])
+
+  useEffect(() => {
+    BackendRemote.rpc
+      .getContactsByIds(accountId, group.pastContactIds)
+      .then((pastContacts: { [id: number]: T.Contact }) => {
+        setPastContacts(
+          group.pastContactIds.map((id: number) => pastContacts[id])
+        )
       })
-    },
-    []
-  )
+  }, [accountId, group.pastContactIds])
+
+  const [groupContacts, setGroupContacts] = useState<T.Contact[]>([])
+
+  useEffect(() => {
+    BackendRemote.rpc
+      .getContactsByIds(accountId, group.contactIds)
+      .then((groupContacts: { [id: number]: T.Contact }) => {
+        setPastContacts(group.contactIds.map((id: number) => groupContacts[id]))
+      })
+  }, [accountId, group.contactIds])
+
+  useEffect(() => {
+    return onDCEvent(
+      accountId,
+      'ContactsChanged',
+      ({ contactId: changedContactId }) => {
+        // update contacts in case a contact changed
+        // while this dialog is open (e.g. contact got blocked)
+        //
+        // Loading the initial `pastContacts`
+        // and `groupContacts` is taken care of in different places.
+
+        let contactIdsToReload: Array<T.Contact['id']>
+        if (changedContactId == null) {
+          contactIdsToReload = [...group.pastContactIds, ...group.contactIds]
+        } else {
+          if (
+            !group.pastContactIds.includes(changedContactId) &&
+            !group.contactIds.includes(changedContactId)
+          ) {
+            // No need to do anything, the contact has nothing to do
+            // with this group. For performance.
+            return
+          }
+
+          contactIdsToReload = [changedContactId]
+        }
+
+        BackendRemote.rpc
+          .getContactsByIds(accountId, contactIdsToReload)
+          .then((contactsToUpdate: { [id: string]: T.Contact }) => {
+            // Making sure to only update the contacts
+            // that are already present in the lists,
+            // because we're doing it in an async way.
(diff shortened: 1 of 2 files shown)


WofWca:
The TODO doesn't sound like a blocker? The flicker only happens once when you open it, right?

Simon-Laux:
> The TODO doesn't sound like a blocker? The flicker only happens once when you open it, right?

technically no, but I personally think that it is a big enough UX regression to investigate solutions. (I'm super annoyed by the flickering)

packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
```diff
@@ -398,6 +397,21 @@ function ChatHeading({ chat }: { chat: T.FullChat }) {
...
+      .getContact(accountId, chat.contactIds[0])
+      .then(setFirstChatContact)
+      .catch(
+        log.error.bind(
+          null,
+          'error fetching first contact of chat for chat subtitle'
+        )
+      )
+  }, [chat.contactIds, accountId])
+
```
WofWca: Do you not like `useRpcFetch`?

Simon-Laux: I forgot about it.

Simon-Laux:
seems like the flickering also breaks the e2e test

WofWca:
> seems like the flickering also breaks the e2e test

I'll take a look.

WofWca:
This is not about flickering. It's because all members are shown as "Past members":

6f66d15ef0b6300b9876c82c9238d7d2433e917e

packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
```diff
@@ -456,7 +460,10 @@ function ChatHeading({ chat }: { chat: T.FullChat }) {
     }
   }
 
-  const subtitle = chatSubtitle(chat)
+  const subtitle = chatSubtitle(
+    chat,
+    firstChatContact.result?.ok ? firstChatContact.result?.value : null
```
WofWca:
```suggestion
    firstChatContact.result?.ok ? firstChatContact.result.value : null
```

packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
```diff
@@ -398,6 +397,11 @@ function ChatHeading({ chat }: { chat: T.FullChat }) {
   const openViewProfileDialog = useOpenViewProfileDialog()
   const accountId = selectedAccountId()
 
+  const firstChatContact = useRpcFetch(BackendRemote.rpc.getContact, [
+    accountId,
+    chat.contactIds[0],
```
WofWca: This could be `undefined`, which would result in an error. Behavior-wise it's OK, but it will spam the console:

https://github.com/deltachat/deltachat-desktop/blob/1dbf369b51bad410e11d7c4ee0a726ded143f6e3/packages/frontend/src/hooks/useFetch.ts#L141-L153

An "ideal" solution would be to pass `null` instead of `[accountId, contactId]` to `useRpcFetch`.

WofWca:
About the flicker: I'm not sure if "suspense" is really what we need. We know the number of contact beforehand, so why not just display placeholders in place of them? Or at least one div of height depending on the number of contacts..

Simon-Laux:
> About the flicker: I'm not sure if "suspense" is really what we need. We know the number of contact beforehand, so why not just display placeholders in place of them? Or at least one div of height depending on the number of contacts..

I went with the single div solution for now.

another solution could be to delay all data loading by waiting for all data before showing the dialog, to emulate the previous behavior. but it is good enough for now and will change anyways once we also use virtualized lazy loaded lists for the contact lists.

packages/frontend/src/components/dialogs/ViewGroup.tsx
```diff
@@ -392,8 +388,18 @@ function ViewGroupInner(
                     <PseudoListItemShowQrCode onClick={() => showQRDialog()} />
                   </>
                 )}
+                {group.contactIds.length != 0 && groupContacts.length == 0 && (
+                  <div /* placeholder to keep layout from jumping around while contact info is loaded */
+                    style={{
+                      height:
+                        group.contactIds.length *
+                        64 /* 64px is the height of a contact list item */,
```
WofWca, line 396: Would be nice to add cross-reference comments

https://github.com/deltachat/deltachat-desktop/blob/845455ef3652a22dbd270116edd10545e0d24fe2/packages/frontend/scss/misc/_avatar.scss#L4-L6

packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
```diff
@@ -367,7 +369,11 @@ function chatSubtitle(chat: Type.FullChat) {
       } else if (chat.isDeviceChat) {
         return tx('device_talk_subtitle')
       }
-      return chat.isEncrypted ? null : chat.contacts[0].address
+      if (chat.isEncrypted) {
+        return null
+      } else {
+        return firstContact?.address
```
WofWca: Very minor thing, but this causes unnecessary jumpiness.

```suggestion
        return firstContact != null ? firstContact.address : tx('loading')
```

(Although it'd be more proper to pass the entire `firstChatContactFetch` to also handle error).

WofWca (approved):
Works, looks good.
Can't wait for groups to start loading blazingly fast!

WofWca:
(We shouldn't merge this before we upgrade core to https://github.com/chatmail/core/pull/7282 ? Because performance would actually get a little worse).

But no big deal

packages/frontend/src/components/dialogs/ViewGroup.tsx
Simon-Laux: I think it is fine for now when it is hardcoded, but in the future we could also read it from css vars like we do for the chat-list-item height.

WofWca: Well a comment wouldn't hurt, but alright.

Simon-Laux:
I merged the core PR today and without this PR I get this error:
image

> We shouldn't merge this before we upgrade core to https://github.com/chatmail/core/pull/7282 ?

So in order that current main desktop and current main core can work together, I'll merge this now.
Note that this PR does not depend on the core PR, so this will not break desktop main, I just confirmed that it works.

PR #5530, opened 2025-10-08
adapt to jsonrpc api change: chat_type as string union

Simon-Laux:
depends on https://github.com/chatmail/core/pull/7285


packages/frontend/src/stockStrings.ts
Simon-Laux, line 1: this extra `typeof` is because I changed how the constants are exported to fix some weird type issues

link2xt:
I have released core 2.26.0 with this breaking change, so this needs to be merged now.

nicodh:
We will do the release 2.25.0 before and then merge it

packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
```diff
@@ -536,7 +527,7 @@ function ChatNavButtons({
         // Core only allows placing calls in chats of type "single"
         // (but not e.g. in groups consisting of 2 members).
         // https://github.com/chatmail/core/blob/738dc5ce197f589131479801db2fbd0fb0964599/src/calls.rs#L147
-        chat.chatType === C.DC_CHAT_TYPE_SINGLE &&
+        chat.chatType === "Single" &&
```
WofWca, line 530:
```suggestion
        chat.chatType === 'Single' &&
```

This will fix all lint warnings.

WofWca (approved):
I hope you don't mind that I rebased this. There was only one conflict which was easy to resolve.

I have checked all the places, and none seem to mix up the new types.
I have also checked that the old variable names are no longer in the code base.

This adds some type casts, but the situation is still better than what we had before.

This can be merged as soon as we upgrade core.

nicodh:
Cherry-picked the commits of this PR to #5739

nicodh:
Closed as the changes were moved to https://github.com/deltachat/deltachat-desktop/pull/5739

PR #5537, opened 2025-10-09
refactor: fix potential `newProxyUrl` bug

WofWca:
This unfortunately appears to trigger an existing bug. A race condition where `batchGetConfig` resolves _after_ the initial `addProxy`.
The result is that the proxy just doesn't get added in the end.

https://github.com/deltachat/deltachat-desktop/blob/9298502b681059c1b60b2098c8f98b9adfa82b00/packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx#L112-L147

That bug needs to get fixed, then this MR can be merged.


nicodh:
I added a new PR #5539 to fix that

WofWca:
Superseded by https://github.com/deltachat/deltachat-desktop/pull/5539.

Issue #5536, opened 2025-10-09
remove "invite to video chat" alert


r10s:
when tapping the "Call" button, there is still the "Do you want to invite to video chat..." questions, we want to strike that all together for now:

<img width="735" height="329" alt="Image" src="https://github.com/user-attachments/assets/71f6c354-8a75-4a90-939c-0340026dc657" />

- calls are much more lightweight
- it is not hard to cancel when accidentally tapped
- worse is that the receiver get a "missed call" message

WofWca:
I suggest to add delay before the call actually starts, after opening the window. Otherwise you start ringing basically instantly. New users might be annoyed by this, "I just wanted to see what this button does".

PR #5538, opened 2025-10-10
improvement: remove "Start Call?" prompt

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/5536.

Maybe it' better to have this delay inside the app itself, but let's start with just desktop.

### DIFF
--- a/packages/target-electron/src/windows/video-call.ts
+++ b/packages/target-electron/src/windows/video-call.ts
@@ -386,7 +386,7 @@ function openVideoCallWindow<D extends CallDirection>(
   const query = callDirection === CallDirection.Incoming ? '?playRingtone' : ''
   const hash =
     callDirection === CallDirection.Outgoing
-      ? '#startCall'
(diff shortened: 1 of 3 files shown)


packages/target-electron/src/windows/video-call.ts
```diff
@@ -396,6 +395,21 @@ function openVideoCallWindow<D extends CallDirection>(
     extraHeaders: 'Content-Security-Policy: ' + CSP,
   })
 
+  if (callDirection === CallDirection.Outgoing) {
+    // Do not `#startCall` immediately. Give the user a moment to cancel it
+    // before it starts ringing.
+    // Note that after we actually `#startCall`,
+    // the offer will be generated almost immediately,
+    // because the app is "warmed up" (e.g. see `iceCandidatePoolSize`).
+    // So basically this sets a minumum delay, rather than an extra delay.
```
nicodh:
```suggestion
    // So basically this sets a minimum delay, rather than an extra delay.
```

PR #5539, opened 2025-10-10
Avoid potential race condition 

nicodh:
- add newProxyUrl from props after loading settings
- extract validateProxyUrl
- readd some button styles

#skip-changelog

### DIFF
--- a/packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
+++ b/packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
@@ -107,7 +107,7 @@ export default function ProxyConfiguration(
     )
   }, [])
 
-  const isValidProxyUrl = useCallback(
+  const validateProxy = useCallback(
     async (proxyUrl: string, existingProxies: string[]): Promise<boolean> => {
       let errorMessage: string | null = null
       let proxyValid = false
@@ -133,7 +133,12 @@ export default function ProxyConfiguration(
   )
 
   useEffect(() => {
-    const loadSettings = async () => {
+    // load current proxy settings from backend
+    // and add incomingProxyUrl if given
+    // This is the case when the proxy url was scanned
+    // from general qr code scanner in an existing account.
+    // Then this dialog is opened with the scanned url as prop.
+    const init = async () => {
       try {
         const { proxy_enabled, proxy_url } =
           await BackendRemote.rpc.batchGetConfig(accountId, [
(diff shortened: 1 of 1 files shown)


WofWca:
Is there a way to summarize what this MR does? The commits messages don't seem to do the job.

nicodh:
> Is there a way to summarize what this MR does? The commits messages don't seem to do the job.

That's why I listed the points in the description.

> add newProxyUrl from props after loading settings

this avoids the race condition (starting in line 154)

> extract validateProxyUrl

now the validation is needed in  2 places so I extracted the function,
also improves the usability since you get an error message if trying to add an existing proxy again

> readd some button styles

the styles were removed as a side effect of the last change in the already merged PR

packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
```diff
@@ -420,6 +424,7 @@ export default function ProxyConfiguration(
                 {tx('proxy_add')}
               </Button>
               <Button
+                aria-label={tx('qrscan_title')}
                 className={styles.scanQrButton}
```
WofWca, line 431: Why is this needed? The buttons are already properly labelled, aren't they?

packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
```diff
@@ -115,15 +141,26 @@ export default function ProxyConfiguration(
...
+          if (incomingProxyUrl) {
+            // This is the case when the proxy url was scanned
+            // from general qr code scanner in an existing account.
+            // Then this dialog opens and the scanned url is passed via props.
+            const proxyValid = await isValidProxyUrl(incomingProxyUrl, proxies)
+            if (proxyValid) {
+              proxies.push(incomingProxyUrl)
+              activeProxy = incomingProxyUrl
+              enabled = true
+            }
```
WofWca, line 167: This approach doesn't work. The proxy doesn't actually get saved to core. Try adding a proxy this way, then closing the "Proxy settings" dialog and reopening it again.

nicodh: updateSettings was missing :-(

packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
```diff
@@ -157,56 +197,22 @@ export default function ProxyConfiguration(
...
       if (proxyValid) {
-        try {
-          const parsedUrl = await BackendRemote.rpc.checkQr(accountId, proxyUrl)
-          proxyValid = parsedUrl.kind === 'proxy'
-        } catch (error) {
-          log.error('checkQr failed with error', error)
-          errorMessage = unknownErrorToString(error)
-          proxyValid = false
-        }
+        setProxyState(prev => ({
```
WofWca, line 204: Why did we have to switch from `updateProxyState` to `setProxyState`.

packages/frontend/src/components/dialogs/ProxyConfiguration/index.tsx
```diff
@@ -91,19 +96,40 @@ export default function ProxyConfiguration(
...
+  const maybeValidProxyUrl = useCallback((url: string): boolean => {
+    const parts = url.split('://')
+    return (
+      parts.length === 2 &&
+      parts[0].length >= 2 && // shortest protocol is ss://
+      parts[1].length >= 1 // host
+    )
+  }, [])
+
+  const isValidProxyUrl = useCallback(
```
WofWca: The function name seems to imply that it has no side effects (the dialog). I'd name it `validateProxyUrl`.

PR #5543, opened 2025-10-10
Add info dialogs for experimental settings

### LINKED ISSUE #5533: reword "Calls" option, add alert dialog to settings
the option "Calls" at "Advanced / Experimental" should be renamed to **Debug Calls**

when enabled, the following message should appear on the screen (probably using a standard alert,  btw, these alerts are useful for many other experiments, it is anyway good to introduce that pattern)

> Thanks for helping to debug  "Calls"!
>
> You can now debug calls using the phone-icon in one-to-one-chats
>
> The experiment is about making decentralised calls work and reliable at all, not about options or UI.
> We're happy about focused feedback at support.delta.chat

the text is pretty temporary and will (hopefully!) not be needed in many releases, so no need to translate it.

see https://github.com/deltachat/deltachat-android/pull/3785#discussion_r2413578745 for some reasoning.

this is how the corresponding dialogs look like on android/ios:

Image

Image

nicodh:
resolves #5533 

#skip-changelog as it is just an experimental feature

### DIFF
--- a/packages/frontend/src/components/Settings/ExperimentalFeatures.tsx
+++ b/packages/frontend/src/components/Settings/ExperimentalFeatures.tsx
@@ -9,44 +9,53 @@ import useTranslationFunction from '../../hooks/useTranslationFunction'
 import SettingsSwitch from './SettingsSwitch'
 import { runtime } from '@deltachat-desktop/runtime-interface'
 import useDialog from '../../hooks/dialog/useDialog'
-import ConfirmationDialog from '../dialogs/ConfirmationDialog'
+import AlertDialog from '../dialogs/AlertDialog'
 
 export function ExperimentalFeatures() {
   const tx = useTranslationFunction()
   const { openDialog } = useDialog()
 
   const showExperimentalInfoDialog = async (
-    settingsKey: keyof SettingsStoreState['desktopSettings'],
+    settingsKey: keyof Pick<
+      SettingsStoreState['desktopSettings'],
+      | 'enableAVCallsV2'
+      | 'enableBroadcastLists'
+      | 'enableOnDemandLocationStreaming'
+    >,
     updatedValue: boolean
   ) => {
     if (!updatedValue) {
       return
     }
-    let header = ''
-    let message = ''
+    let message: string
     // The strings are copy-pasted from
     // https://github.com/deltachat/deltachat-android/blob/2385b236c7ed9eb0e26ef819d8274936877b7023/src/main/java/org/thoughtcrime/securesms/preferences/AdvancedPreferenceFragment.java
-    if (settingsKey === 'enableAVCallsV2') {
-      header = 'Thanks for helping to debug "Calls"!'
-      message =
-        '• You can now debug calls using the "phone" icon in one-to-one-chats' +
-        "\n\n• The experiment is about making decentralised calls work and reliable at all, not about options or UI. We're happy about focused feedback at support.delta.chat"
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/Settings/DesktopSettingsSwitch.tsx
```diff
@@ -37,11 +37,12 @@ export default function DesktopSettingsSwitch({
...
       value={value}
       onChange={async () => {
+        const newValue = !settingsStore.desktopSettings[settingsKey]
         await SettingsStoreInstance.effect.setDesktopSetting(
           settingsKey,
-          !settingsStore.desktopSettings[settingsKey]
+          newValue
         )
-        callback?.()
+        callback?.(newValue)
```
nicodh, line 45: This is not reliable since we don't check if the update was successfull, but I think it's ok to skip that here

packages/frontend/src/components/Settings/ExperimentalFeatures.tsx
```diff
@@ -1,31 +1,79 @@
...
+  const showExperimentalInfoDialog = async (
+    settingsKey: keyof SettingsStoreState['desktopSettings'],
+    updatedValue: boolean
+  ) => {
+    if (!updatedValue) {
+      return
+    }
+    let header = ''
+    let message = ''
+    if (settingsKey === 'enableAVCallsV2') {
```
WofWca:
```suggestion
    // The strings are copy-pasted from
    // https://github.com/deltachat/deltachat-android/blob/2385b236c7ed9eb0e26ef819d8274936877b7023/src/main/java/org/thoughtcrime/securesms/preferences/AdvancedPreferenceFragment.java
    if (settingsKey === 'enableAVCallsV2') {
```

packages/frontend/src/components/Settings/ExperimentalFeatures.tsx
```diff
@@ -1,31 +1,79 @@
...
+        '• If enabled you will find a map icon above the message list, which opens a map with shared locations of your contacts' +
+        '\n\n• Sharing your own location is only available in mobile clients'
+    }
+    message +=
+      '\n\n• If you want to quit the experimental feature, you can disable it at "Settings / Advanced"'
+    openDialog(ConfirmationDialog, {
+      header,
+      message,
+      confirmLabel: tx('ok'),
+      cb: () => {},
```
WofWca: The "cancel" button does nothing.

packages/frontend/src/components/Settings/ExperimentalFeatures.tsx
```diff
@@ -1,31 +1,79 @@
...
+    })
+  }
 
   return (
     <>
       {runtime.getRuntimeInfo().target === 'electron' && (
         <DesktopSettingsSwitch
           settingsKey='enableAVCallsV2'
-          label={tx('calls')}
```
WofWca, line 17: The string needs to get removed then I guess.

packages/frontend/src/components/Settings/ExperimentalFeatures.tsx
nicodh: I will switch to AlertDialog and then also apply the changes you made in your PR, ok?

WofWca: I haven't checked if the Alert dialog will work, but if the only difference is that it doesn't have a cancel button, then sounds good.

PR #5544, opened 2025-10-10
refactor: improve type safety in experimental opts

WofWca:


packages/frontend/src/components/Settings/ExperimentalFeatures.tsx
```diff
@@ -16,27 +16,39 @@ export function ExperimentalFeatures() {
   const { openDialog } = useDialog()
 
   const showExperimentalInfoDialog = async (
-    settingsKey: keyof SettingsStoreState['desktopSettings'],
+    settingsKey: keyof Pick<
+      SettingsStoreState['desktopSettings'],
+      | 'enableAVCallsV2'
+      | 'enableBroadcastLists'
+      | 'enableOnDemandLocationStreaming'
```
nicodh, line 23: So this is just to make sure no keys are passes that are not handled, right?
What is the use case? A developer calling this with any non valid key? Idk, seems a bit exaggerated...?

WofWca: This parameter change followed from switching to exhaustive checking of `settingsKey`, i.e. requiring that `header` and `message` be properly initialized. As a nice side-effect yes, we get the ability to tell if the call to `showExperimentalInfoDialog` is incorrect and that the function will misbehave.

nicodh: Ok, I will apply that in a commit I just finish...

WofWca:
Already merged

https://github.com/deltachat/deltachat-desktop/blob/620560153b543cb55f0e5b782098e24ba5888604/packages/frontend/src/components/Settings/ExperimentalFeatures.tsx#L18-L52

PR #5545, opened 2025-10-10
chore: upgrade calls-webapp

WofWca:
This includes the 'Swap "Accept" and "End Call" buttons' patch.

#skip-changelog because this is experimental


WofWca:
Superseded

Issue #5529, opened 2025-10-08
call window does not always show correct avatar


r10s:
when a user is using the "fallback" avatar (color and letter), this one is not shown in the call window.

instead, a different "fallback" avatar is shown, resulting in unnecessary confusion.

<img width="686" height="580" alt="Image" src="https://github.com/user-attachments/assets/b31e4f59-0e2e-4521-b923-f25a8d81940d" /><br><i>different avatars</i>

this is not a big thing, however, for the first impression it is be good to have that correct

WofWca:
This needs to be implemented either in the calls-webapp itself, or we need to turn the initial into an image somehow. Maybe with SVG, or offscreen canvas.

r10s:
ftr, android is using an "offscreen canvas" (in a drawable), not sure how practical or complex is that on desktop.

alternative would be to show _no_ avatar instead of "wrong" fallback, that could probably be very easy and is also good enough for now

WofWca:
If the "SVG" approach works, we might want to extract the "get first character" utility function out of https://github.com/deltachat/deltachat-desktop/pull/5502/files#diff-edf5bc4298a0ce2a42613ce04127f8dddd41014fd0a1aa302964ed26e4fa01aeR42-R48 and used it both on the front-end and the back-end (Node).

PR #5546, opened 2025-10-10
fix: calls: show initial letter img if no avatar

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/5529.

The SVG approach works nicely.

image  image image

TODO:
- [x] Address TODO comments.

#skip-changelog because this is experimental.

### DIFF
--- a/packages/target-electron/src/windows/video-call.ts
+++ b/packages/target-electron/src/windows/video-call.ts
@@ -1,3 +1,4 @@
+import escapeHTML from 'escape-html'
 import {
   BrowserWindow,
   dialog,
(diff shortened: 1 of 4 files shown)


packages/target-electron/src/windows/video-call.ts
```diff
@@ -626,15 +626,27 @@ async function returnIndexHtmlOrAvatar(request: GlobalRequest) {
...
+    const chat = await jsonrpcRemote.rpc.getBasicChatInfo(
       parsedHost.accountId,
       parsedHost.chatId
     )
-    if (profileImage == null || profileImage == '') {
-      // TODO shouldn't we display an initial letter avatar then?
-      return makeResponse('', { status: 404 })
+    if (chat.profileImage == null || chat.profileImage == '') {
+      const firstLetter = chat.name[0]
+      // TODO un-XSS. Although this should be safe already,
```
WofWca: I guess we can utilize https://www.npmjs.com/package/escape-html. It's pretty light.

nicodh: But that's just a 44 line function which replaces 5 char codes. I would say use that function but no need to import the lib. I don't think that that is even something that can be "protected" IP

WofWca: Umm what IP?

> I would say use that function but no need to import the lib

This would make the reader of the code to worry about whether this is a correct implementation.

WofWca: Oh, intellectual property. Got it.

Simon-Laux: you could probably also try `Intl.Segmenter('en', { granularity: 'grapheme' });`, I think there is no need to check if it is an emoji first in this case.

Simon-Laux:
For sake of consistency it may make sense to also set the emoji font to the same we use in the main window:

if it exists "apple emoji", else the copy of "google noto color emoji" that we ship with the app.

https://github.com/deltachat/deltachat-desktop/blob/5094f29a898cd3fda05fe81e578f5f46c1340466/static/fonts/fonts.css#L41-L44

WofWca:
I am afraid this is gonna be hard with this SVG approach. Apparently we would need to embed the font inside the SVG file itself:
https://stackoverflow.com/questions/30024943/how-to-specify-font-family-in-svg

I suggest to tackle this later.

packages/target-electron/src/windows/video-call.ts
```diff
@@ -630,15 +631,34 @@ async function returnIndexHtmlOrAvatar(request: GlobalRequest) {
...
       parsedHost.chatId
     )
-    if (profileImage == null || profileImage == '') {
-      // TODO shouldn't we display an initial letter avatar then?
-      return makeResponse('', { status: 404 })
+    if (chat.profileImage == null || chat.profileImage == '') {
+      const initial = avatarInitial(chat.name)
+      // Style (font size, etc) are mostly copy-pasted from `.avatar`
+      // (in `_avatar.scss`).
+      return makeResponse(
```
WofWca, line 645:
```suggestion
      // (in `_avatar.scss`).
      // TODO: somehow use the same font that we use in main window
      // (`EmojiMart`). See
      // https://github.com/deltachat/deltachat-desktop/pull/5546#issuecomment-3418636124.
      return makeResponse(
```

WofWca:
> Or is the screenshot provided here a svg?

I didn't get it.

> it shows a different emoji like in the description here

Hmmm, unfortunate. Maybe we need to add `Apple Color Emoji` to the list of fonts.

nicodh:
> > Or is the screenshot provided here a svg?
> 
> I didn't get it.

I meant  is this screenshot based on a svg avatar or an emoji (I assume the latter)
image

WofWca:
Yes, the user has no avatar, only a name that starts with an emoji.

WofWca:
Please try again. I have added some more fonts.

nicodh:
Now I don't even see the reduced emoji but just a grey avatar and an error message:
avatar:1  Failed to load resource: the server responded with a status of 404 (Not Found)

WofWca:
Umm I didn't change anything in that regard.

nicodh:
Tried again, now it loads but look the same as before your (forced) update. But I'm not sure if that is font-family related. In the source code of the avatar in the chat window I see 

```
<div class="content" style="--local-avatar-color: #0062da;">👨&zwj;👩&zwj;👦&zwj;👦</div>
```

So that probably can't be displayed with the css rule
background-image: url(/avatar);

So I think it's not worth doing more on this. I just wondered how it happened that this multi emoji is working...

image

WofWca:
Thanks for checking. 

It would be nice to also check the resulting SVG text through dev tools then if you think that this is not about fonts.

PR #5548, opened 2025-10-10
Change icon order

nicodh:
As discussed with @r10s

- phone icon should be the first icon with some margin to the other icons
- last used apps close to apps&media icon

image

#skip-changelog - just a minor change


r10s:
sorry for being unclear, i wanted to suggest to put the phone icon left of menu:

499891132-8f4cabf9-11af-4225-8fba-9741ce190dbc-2

or:

499891132-8f4cabf9-11af-4225-8fba-9741ce190dbc


that way, we have a stable position for the button, and it also looks better as icon-types are separated

very nice, to have some spacing and put all "app related" visually together

Issue #5547, opened 2025-10-10
calls: make debug log easier accessible


r10s:
android/ios use [calls-webapp compiled with `NODE_ENV=debug`](https://github.com/deltachat/calls-webapp?tab=readme-ov-file#building), resulting in a button in the lower left corner where you can directly access the log of the calls window.

on desktop, it is more complicated to access the log, and on macOS, i did not find a method at all (debug log and shortcuts always refer to the main window)

also for consistency, we should use `NODE_ENV=debug` on desktop as well for now

WofWca:
> on desktop, it is more complicated to access the log

What is complicated? Opening the developer tools?

> and on macOS, i did not find a method at all (debug log and shortcuts always refer to the main window)

Have you tried pressing Alt to show the menu? Did you execute Ctrl + Shift + I while the call window is focused?

r10s:
there is the shortcut ⌥⌘ + I, that opens the debug view. however, for me, it always opens the log of the main window.

but even if that is working, that's more complicated to explain to some tester than saying: "hit the button lower right" ...

<img width=320 src=https://github.com/user-attachments/assets/44106273-0034-4e6f-9d2f-90b44282727b>
<img width=320 src=https://github.com/user-attachments/assets/b65f3d73-5071-4d75-a823-ee3f91503c90>

... esp. as that is possible on all UI then -  none of the information shown above are accessible for me on macOS - but i've seen even linux user struggling.

there may be other ways to fix that, but using `NODE_ENV=debug` seems to be super-simple. it is anyway only a temporary thing. is there anything that speaks against that?

WofWca:
> is there anything that speaks against that?

No, nothing non-obvious.

PR #5549, opened 2025-10-10
fix: use debug build of calls-webapp

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/5547.
Apparently Chromium developer tools are not good enough.

~~Note that the Eruda icon is not displayed properly because our CSP won't let Eruda load its fonts:~~

image

#skip-changelog because the feature is experimental

### DIFF
--- a/packages/target-electron/package.json
+++ b/packages/target-electron/package.json
@@ -34,7 +34,7 @@
     "build": "pnpm build:locales && pnpm build:backend && pnpm --filter=@deltachat-desktop/frontend build && pnpm build:compose-frontend && pnpm build:calls-webapp",
     "build:locales": "pnpm -w translations:convert",
     "build:backend": "node ./bin/build.js",
-    "build:calls-webapp": "node ../../bin/copy.js ./node_modules/calls-webapp#build-for-deltachat-desktop ./html-dist/calls-webapp",
+    "build:calls-webapp": "node ../../bin/copy.js ./node_modules/calls-webapp#debug-build ./html-dist/calls-webapp",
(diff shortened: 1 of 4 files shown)


nicodh:
Just had a look (maybe too fast :-)

I get

```
Refused to load the font '<URL>' because it violates the following Content Security Policy directive: "default-src 'none'". Note that 'font-src' was not explicitly set, so 'default-src' is used as a fallback.

15:50:18.530 (index):71 empty URL hash:  calls-webapp-scheme://87.133.calls-webapp-dummy-host/
15:50:18.533 (index):16 Refused to load the image 'data:image/gif;base64,R0lGODlhAQABAIAAAP///wAAACH5BAEAAAAALAAAAAABAAEAAAICRAEAOw==' because it violates the following Content Security Policy directive: "img-src 'self'".

oe @ (index):16
15:50:18.535 (index):16 Refused to load the image 'data:image/gif;base64,R0lGODlhAQABAIAAAP///wAAACH5BAEAAAAALAAAAAABAAEAAAICRAEAOw==' because it violates the following Content Security Policy directive: "img-src 'self'".

```

WofWca:
Yes, the last 3 lines have already been present. The font thing is new.
I edited the description. The issue is that Eruda icons are not properly displayed because of this.

Since the debug build is a temporary measure, I suggest to ignore this. Unless someone says that having these icons is critical.

nicodh:
> Since the debug build is a temporary measure, I suggest to ignore this. Unless someone says that having these icons is critical.

But I see nothing now? So nothing to click on for the logs.

image

WofWca:
Oh. I guess it's different on macOS. OK, I will make the CSP more lax.

WofWca:
I force-pushed, now the icon is displayed properly. Thanks for testing!

PR #5550, opened 2025-10-10
fix: make link_core scripts also run on nixos by using `#!/usr/bin/env bash` instead of `#!/bin/bash`

Simon-Laux:


Simon-Laux:
the node script already had this and there were no issues that I know of
 
https://github.com/deltachat/deltachat-desktop/blob/f92514e44308d07e7e742ed02916b921d3da1412/bin/link_core/link_version.js#L1

PR #5553, opened 2025-10-11
improvement: use account label initial as avatar

WofWca:
This makes the profile tags (labels) more usefult,
because otherwise these labels are only ever displayed
in the profile hover "tooltip".

image

TODO:
- [ ] sync with Android and iOS DC clients. Although the "accounts" display there is a little different, so maybe the "be consistent" requirements here are not that strict.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - some strings being untranslated
 
 ### Changed
+- Use profile label first letter (or emoji) as its avatar when no avatar set, instead of the profile public name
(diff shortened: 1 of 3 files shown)


r10s:
the avatar shown for self should be the same as shown for others.

so, using the profile-tag for the avatar is a non-goal and would add confusion and weaken a UI element for the many

(the profile tag is already for the few and pretty much advanced)

let's close this pr.

in general, it is better to first file an issue or a feature request to agree on, btw

PR #5556, opened 2025-10-12
chore: calls-webapp: get from npmjs.com

WofWca:
Instead of github.com.
Because this causes trouble for our Flatpak build,
which can't fetch packages from anything other than npmjs.com.

#skip-changelog


nicodh:
I think we should publish that in the  @deltachat account.

WofWca:
Moved to `@deltachat` npmjs.com org and rebased.

WofWca:
Tested that it works.

PR #5559, opened 2025-10-13
Handle invalid qr code properly

### LINKED ISSUE #5555: add profile: confusing error dialog when scanning/pasting a wrong string
> this bug was initially reported by @cve-1312  at #5554 - thanks a lot!

when scanning or pasting a QR code to set up a profile, and the scanned QR code or the pasted string is _wrong_ (eg. `foo bar`), the user gets a dialog showing the string and offering to copy it to the clipboard - esp. on pasting, this  does not make sense as the string is already on the clipboard.  but also otherwise, this is not what the user wants to do here.

also, the other button "OK" is unclear - "OK what?"

we should replace the dialog by a showing the more telling error message `qraccount_qr_code_cannot_be_used`, this is what eg. android/iOS is doing

the weird dialog after _pasting_ a bad QR code:

when _scanning_ a bad QR code, the app even becomes unusable - both buttons do not work for me:

Image

what others are showing instead:

**to be clear:** when a correct QR code as  `dcaccount:https://mehl.cloud/cgi-bin/newemail.py` is scanned or pasted, things are fine: the user gets back to the "enter name" field with the new server selected

nicodh:
resolves #5555

tbd: should we show the copy dialog when a user with configured account scanned an invalid qr code? Or just a simple alert dialog with just a OK button?

The text can't be copied from a simple alert dialog, since we have user-select: false on the content of the Alert dialog

Maybe we should just change that, so allow select the content of Alert dialogs always?

### DIFF
--- a/packages/frontend/src/hooks/useProcessQr.ts
+++ b/packages/frontend/src/hooks/useProcessQr.ts
@@ -423,16 +423,26 @@ export default function useProcessQR() {
         return callback?.()
       }
 
-      // Just show the contents of the scanned QR code if it is correct but
+      // Just show the contents of the scanned QR code
+      // so the user can see what it contains,
       // we don't know what to do with it ..
-      openDialog(CopyContentAlertDialog, {
-        message:
-          qr.kind === 'url'
-            ? tx('qrscan_contains_url', url)
-            : tx('qrscan_contains_text', url),
-        content: url,
-        cb: callback,
-      })
+      if (isLoggedIn) {
+        openDialog(CopyContentAlertDialog, {
+          message:
+            qr.kind === 'url'
+              ? tx('qrscan_contains_url', url)
+              : tx('qrscan_contains_text', url),
+          content: url,
+        })
+        // this closes the underlying scan dialog immediately
+        // while the copyContent dialog is still open
+        return callback?.()
+      } else {
+        await openAlertDialog({
+          message: tx('qraccount_qr_code_cannot_be_used'),
+        })
+        return callback?.()
+      }
     },
     [
       openAlertDialog,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - some strings being untranslated
 - VCard (share contact) avatars not having color (all being gray instead) #5552
 - some emoji avatars displaying incorrect emoji
+- handle invalid qr code properly #5555
 
 ### Changed
 - update `@deltachat/stdio-rpc-server` and `deltachat/jsonrpc-client` to `2.17.0`


packages/frontend/src/hooks/useProcessQr.ts
```diff
@@ -423,16 +423,26 @@ export default function useProcessQR() {
...
+      if (isLoggedIn) {
+        openDialog(CopyContentAlertDialog, {
+          message:
+            qr.kind === 'url'
+              ? tx('qrscan_contains_url', url)
+              : tx('qrscan_contains_text', url),
+          content: url,
+        })
+        // this closes the underlying scan dialog immediately
+        // while the copyContent dialog is still open
```
WofWca, line 438: Why is this needed?

packages/frontend/src/hooks/useProcessQr.ts
```diff
@@ -423,16 +423,26 @@ export default function useProcessQR() {
...
+        return callback?.()
+      } else {
+        await openAlertDialog({
+          message: tx('qraccount_qr_code_cannot_be_used'),
+        })
```
WofWca, line 443: `!isLoggedIn` does not always imply that the intent of scanning the QR code is to create an account. This could also be "add as second device", in which case I don't think this string applies.

WofWca, line 443: Additionally, this dialog is only shown if the scanned QR is not otherwise recognized above. For example, if you try to scan a proxy QR in the "scan invitation code" dialog, you'll get "You must log in first". Which I guess means that the MR doesn't fully resolve the issue.

WofWca:
> tbd: should we show the copy dialog when a user with configured account scanned an invalid qr code? Or just a simple alert dialog with just a OK button?

I think having the "Copy" is nice. Also this is what Android is doing in its "New Contact -> Scan QR" dialog (not to say that it's doing the right thing).
But also I don' think that `user-select: none;` really is needed. Why would you prevent users from being able to copy stuff.

packages/frontend/src/hooks/useProcessQr.ts
WofWca: Although Android has the same behavior. IDK if this is intentional. But if so, I think a comment about this would be nice.

nicodh: > you'll get "You must log in first"

this will only happen, if you scan a valid proxy url. So maybe we should show a bit more information for that use case:
1. show the user that a (valid) proxy url was detected
2. maybe give some explanation how to use a proxy before onboarding
3. maybe show a dialog: do you want to setup a proxy before onboarding (and show the appropriate dialog if the user confirms)

But I would say all that is beyond the scope of this MR i.e. can be done in a followup

packages/frontend/src/hooks/useProcessQr.ts
nicodh: Since this is the only scenario where return the callback without awaiting the previous action since openDialog is not async.

Not returning immediately results in the multiple copyDialogs to be opened

r10s:
the "Copy" dialog should definetly not being shown, only the alert.

an additional "Copy" button seems superfluous as well, it would be more confusing than helping for the avg user.

also android/ios do not have them, see screenshots at #5555

the button "Copy to clipboard" mentioned above is about the generic scanner, that can scan contacts/proxies/secure-join - and plain-text

WofWca:
Yes, I was talking about the "generic" scanner, and not showing the copy button in the "can't create account with this QR" dialog.

packages/frontend/src/hooks/useProcessQr.ts
WofWca: > But I would say all that is beyond the scope of this MR i.e. can be done in a followup

In that case the commit message and the description needs to be updated not to say that it resolves the issue.
This MR I'd say only covers a very rare case where the scanned QR has nothing to do with Delta Chat.

packages/frontend/src/hooks/useProcessQr.ts
WofWca: It looks like adding `await openDialog` without invoking the callback is the solution. But OK, let's not require it in this MR.

packages/frontend/src/hooks/useProcessQr.ts
nicodh: But that is exactly what was the described use case in the issue:

Title: "a wrong qr code"

> and the scanned QR code or the pasted string is wrong (eg. foo bar)

and the title of this MR says "handle invalid qr code"

And this MR also fixes the issue that the dialog ends up in a blocked UI which is also described in the issue.

That's why I say other use cases should be handled in followups. Which means valid QR codes that are not suitable for InstantOnboarding etc.

WofWca: OK, let's then open an issue so that we don't forget.

nicodh: I think to show appropriate dialogs for all use cases we have to provide more context to the processQrCode hook than only the isUserLoggedIn variable.

nicodh: > OK, let's then open an issue so that we don't forget.

I created #5563

PR #5560, opened 2025-10-13
Move shared deps to catalog

nicodh:
#skip-changelog

depends on #5502


Simon-Laux:
would be easier to review if this pr would merge into the `linkify` branch

PR #5562, opened 2025-10-13
fix: clean up `useProcessQR` a little

WofWca:
#skip-changelog because this is insignificant, mostly a refactor.


nicodh:
Wouldn't it make sense to do the cleanup after the MR #5559 or to make a MR to the fix-invalid-qrcode-handling branch?

WofWca:
I didn't base this MR on top of that MR because that MR is still in progress. I will resolve the conflicts, but I think we will not see any.

nicodh (approved):

Issue #5563, opened 2025-10-13
Handle all qr code scans properly


nicodh:
We have several contexts to scan a QR code and all are handled in useProcessQrCode where we have no context except the isUserLoggedIn boolean.

- [ ]  ~~define a matrix for all contexts with the expected result~~
- [x] implement the appropriate handling
- [ ] add e2e tests for these cases

edit: postpone a matrix until we have a more consistent strategy for the different possible paths to scan a qr code

WofWca:
At a glance, defining a matrix is too complex of a solution to a problem that sounds simple.

It seems like we should stop using the same "handle QR" function in all contexts. Each context should have separate handling code.

Well, maybe it's easier to look up how Android does it.

nicodh:
> At a glance, defining a matrix is too complex of a solution to a problem that sounds simple.
 Well I thought so too, but we have quite a lot different qr codes and contexts now:

Scan contexts:

1. Account creation from Welcome Screen:
- Add as second device
- Use other server: Invite Code / Login etc.
- Scan Proxy
2. Account creation from Account Scan
- Invite code, Login etc.
4. Settings->Proxy->scan Proxy

For example right now we have stuff like: (scanning an invite code in the add second device scan)

<img width="512" height="633" alt="Image" src="https://github.com/user-attachments/assets/2aa7caf0-4923-4e54-8fb6-a5cd6be42e87" />

Or when scanning a mailto: when being on a chatmail server

<img width="520" height="630" alt="Image" src="https://github.com/user-attachments/assets/35805371-92a8-4444-a286-558ae2c57e34" />

nicodh:
After investigating a bit: we have these contexts for scanning a qr code:

1. Onboarding => Add as second device
2. Onboarding => Use existing server => Invite Code
3. Onboarding => set/add proxy

As configured user

1. QrCode Button in navbar
2. New contact QrCode Button in new Chat
3. in InvalidUnencryptedMailDialog & VerifiedContactsRequiredDialog  => Invite contact with "Scan QR Code"
4. ProxySettings => set/add proxy  (as configured user)

I also found:

- [ ] A Vcard is always scanned as non encrypted account, even if a key is provided (same on Android) https://github.com/chatmail/core/issues/6906
[...]

PR #5565, opened 2025-10-16
fix: handle wrong qr code scans

nicodh:
related to #5563 

Introduce new type SCAN_CONTEXT_TYPE to handle qr codes scans more appropriate.

For now it just has impact for the "Add second device" & the "Use other server" context.


packages/shared/constants.ts
```diff
@@ -35,3 +35,9 @@ export const enum NOTIFICATION_TYPE {
   REACTION,
   WEBXDC_INFO,
 }
+
+export const enum SCAN_CONTEXT_TYPE {
+  DEFAULT = 'DEFAULT', // default context, no restrictions on qr code types
```
nicodh: TBD: we have a lot of use cases where we just pass the "DEFAULT" and only 2 where we pass other contexts. So we could make the parameter optional and have less code. At the same time it increases the risk to "forget" the scan context parameter in new use cases and introduce new wrong handlings in future.

packages/frontend/src/components/helpers/PseudoListItem.tsx
```diff
@@ -160,7 +161,7 @@ export const PseudoListItemAddContactOrGroupFromInviteLink = ({
   const contact = contactFetch?.result?.ok ? contactFetch.result.value : null
 
   const onClick = () => {
-    processQr(accountId, inviteLinkTrimmed)
+    processQr(accountId, inviteLinkTrimmed, SCAN_CONTEXT_TYPE.DEFAULT)
```
nicodh, line 164: TBD: the qr button here comes with the text "New contact" but in fact it opens the general qr code scanner. We could restrict this context to only allow qr codes that in fact "introduce a new contact".
In Android the scanner opened here has no limitations like that...

The same applies to the dialogs in InvalidUncryptedMail & ProtectionStatusDialog.

packages/shared/constants.ts
```diff
@@ -35,3 +35,9 @@ export const enum NOTIFICATION_TYPE {
...
+  OTHER_SERVER = 'OTHER_SERVER', // onboarding with another server than the default server
```
nicodh: The naming is somehow inconsistent, but so is the wording in the UI.
Clicking on the "Use other server" button you find a "Scan Invitation Code" button which can be used to scan a DC Account url which is not an invitation code but sets the chatmail server to another URL...

Simon-Laux: Back then when the button text was chosen, it was "invitation to a server", that's where the wording came from.

AFAIK now it can be both invitation code to a group/user or dcaccount code (server account creation link/code).

packages/shared/constants.ts
Simon-Laux: makes more sense to protect against risk of forgetting cases by making things more explicit in this case IMO

packages/shared/constants.ts
Simon-Laux:
```suggestion
  OTHER_SERVER = 'OTHER_SERVER', // on-boarding with another server than the default server or contact/group invite code
```

packages/shared/constants.ts
```diff
@@ -35,3 +35,9 @@ export const enum NOTIFICATION_TYPE {
   REACTION,
   WEBXDC_INFO,
 }
+
+export const enum SCAN_CONTEXT_TYPE {
```
Simon-Laux: the comments here should be jsdoc comments so that they appear on hover in IDEs

packages/frontend/src/components/helpers/PseudoListItem.tsx
Simon-Laux: I think it is fine in those cases, as there are specific dialogs for scanning account creation codes
image
and add as second device codes:
image

Simon-Laux: although the multi-device setup dialog text should be updated, the account switcher is not behind a "logout" action anymore. ideally it should also just say "you are about to sync your existing profile to this device, this will not overwrite your current profile, continue?" or sth. like that. and the ok button would just create a new unconfigured profile and execute the code again, like it is done for scanning `dcaccount:` links

packages/shared/constants.ts
```diff
@@ -35,3 +35,12 @@ export const enum NOTIFICATION_TYPE {
...
```
WofWca: This does not belong to `shared` because it's only used on the front-end.

WofWca (approved):
I have confirmed that the "add as second device" and "scan invitation code" workflows are not broken, at least for `askVerifyContact` and `login`.
And this works otherwise.

Issue #5566, opened 2025-10-16
Improve withdraw/revive invitation code dialogs


nicodh:
The dialogs to withdraw or revive an invitation code have no explicit buttons for "withdraw" or "revive" in the dialogs, only header with "Revoke QR code" or "Revive QR code" and cancel/ok buttons which is pretty misleading since the text describes the existing state and the header the (possible) next state

<img width="527" height="240" alt="Image" src="https://github.com/user-attachments/assets/dbeea8d5-7a91-46fe-8b8c-1722a8bb4ecb" />

Simon-Laux:
closed by https://github.com/deltachat/deltachat-desktop/pull/5567

PR #5567, opened 2025-10-16
fix: show correct button label (#5566)

nicodh:
#skip-changelog

image


nicodh:
> This looks better to me, but I probably don't have the power to approve this. Maybe this is somehow intentional and needs to be in sync with Android and iOS.

I don't think so, but since it's a quite small change maybe @r10s can have a look

Simon-Laux (approved):

Simon-Laux:
Showing concrete action verb[^1] on confirmation buttons is always a good change in my view.

[^1]: as long as it is small enough to fit on the button.

PR #5572, opened 2025-10-18
Set fix height for drafts

### LINKED ISSUE #4314: style: make composer's quote height constant

nicodh:
resolves #4314

#skip-changelog - just a minor change

### DIFF
--- a/packages/frontend/scss/composer/_composer.scss
+++ b/packages/frontend/scss/composer/_composer.scss
@@ -24,17 +24,22 @@
       padding: 9px;
       border-bottom: 1px solid
         color-mix(in srgb, var(--messageQuotedText) 15%, transparent);
-      height: 64px;
 
       .quote {
         // this is just overwrites for the quote styles defined in _message.scss
         margin: 4px;
+        .quoted-text {
+          // Set fixed height to reduce layout shifts
+          // when changing the quote rapidly, e.g. with the Ctrl + Up shortcut.
+          // See https://github.com/deltachat/deltachat-desktop/issues/4314.
+          height: 24px;
+        }
       }
 
       // The "remove attachment / quote" buttons.
       button:not(.quote-background) {
         margin-inline-start: auto;
-        justify-content: right;
+        justify-content: flex-end;
       }
 
       .message-attachment-media {


packages/frontend/scss/composer/_composer.scss
```diff
@@ -33,6 +34,7 @@
       // The "remove attachment / quote" buttons.
       button:not(.quote-background) {
         margin-inline-start: auto;
+        justify-content: right;
```
WofWca: This is not RTL-dependent. Why not fix it properly, e.g. by setting `flex-grow` or `flex-shrink`?

packages/frontend/scss/composer/_composer.scss
```diff
@@ -24,6 +24,7 @@
       padding: 9px;
       border-bottom: 1px solid
         color-mix(in srgb, var(--messageQuotedText) 15%, transparent);
+      height: 64px;
```
WofWca: At some zoom levels a part of the third line can be seen. IDK if we're good with this.

Image

And this seems to make these lines pointless:

https://github.com/deltachat/deltachat-desktop/blob/58d3415c0d2537e6e0b0fe29d6c0dd347ffe950f/packages/frontend/scss/message/_message.scss#L582-L586

nicodh: .quote-text applies to all ocurrences of quotes (also in message list) - so we would need a special rule if draft is in composer upper-row.

WofWca: Oh. You're right. OK, I guess it's fine then.

WofWca:
```suggestion
      // Set fixed height to reduce layout shifts
      // when changing the quote rapidly, e.g. with the Ctrl + Up shortcut.
      // See https://github.com/deltachat/deltachat-desktop/issues/4314.
      height: 64px;
```

Simon-Laux: where does this "64px" come from? wouldn't it make sense to derive it from font size with `calc()` and `em`/`rem`?

also may make sense to also overwrite the line limit in this case, so it's always only two lines.

packages/frontend/scss/composer/_composer.scss
Simon-Laux: or at-least `flex-end` instead of `right`, which adapts to text direction

PR #5573, opened 2025-10-18
chore: format on save in vscode

nicodh:
#skip-changelog


Simon-Laux:
maybe we also need to set the formatter to prettier somewhere

PR #5576, opened 2025-10-18
fix: potential race in WebXDC attachment info

WofWca:
Plus simplify the code, and improve the comment a little.

This is a follow-up to b352801fb8d2e816e364aadced44235299a6baa0
(https://github.com/deltachat/deltachat-desktop/pull/5227).

Partially addresses https://github.com/deltachat/deltachat-desktop/pull/5576.

#skip-changelog because this is insignificant.

### DIFF
--- a/packages/frontend/src/components/attachment/messageAttachment.tsx
+++ b/packages/frontend/src/components/attachment/messageAttachment.tsx
@@ -239,7 +239,8 @@ export function DraftAttachment({
 }) {
   const isViewTypeWebxdc = attachment.viewType === 'Webxdc'
   const [attachmentIdToLoad, setAttachmentIdToLoad] = useState(attachment.id)
-  if (useHasChanged2(attachment.fileName)) {
+  const hasFileNameChanged = useHasChanged2(attachment.fileName)
+  if (hasFileNameChanged) {
     // Only reload webxdc info if filename has changed, because
     // the `id` itself could be changing as often as we update the draft.
     setAttachmentIdToLoad(attachment.id)


WofWca:
The test failure is unrelated. The server is really slow.

packages/frontend/src/components/attachment/messageAttachment.tsx
```diff
@@ -235,37 +237,17 @@ export function DraftAttachment({
...
-          })
-          .finally(() => {
-            setIsLoadingWebxdcInfo(false)
-          })
-      }
-    }
-  }, [accountId, attachment.id, attachment.viewType, attachment.fileName])
+  const isViewTypeWebxdc = attachment.viewType === 'Webxdc'
+  const [attachmentIdToLoad, setAttachmentIdToLoad] = useState(attachment.id)
+  if (useHasChanged2(attachment.fileName)) {
```
Simon-Laux:
```suggestion
  const hasFileNameChanged = useHasChanged2(attachment.fileName)
  if (hasFileNameChanged) {
```

to make it extra clear that rules of hooks are not violated (no conditional hooks).

Simon-Laux (approved):
Looks good and removes code.

packages/frontend/src/components/attachment/messageAttachment.tsx
WofWca: > to make it extra clear that rules of hooks are not violated (no conditional hooks).

IDK, those are caught at compile time. The fact that build succeeds means that there is no such error, so there is no need for this I'd say.

Simon-Laux: is it caught if you do this?
```js
if (useHook1() && useHook2()) {}
```

WofWca: I just tried it. Yes.

Simon-Laux: good then not necessary, but may still be confusing to react novices.

WofWca: Alright, if just for the sake of clarity then let's apply.

PR #5585, opened 2025-10-20
refactor: split up `ChatList`'s search sections

WofWca:
This should not affect behavior, besides performance.

This is mostly about moving the code around,
without much changes to the logic.
Which is easy to see with a diff viewer that recognizes code moves.

Commit-by-commit review should be easy. Out of all the commits,
``refactor: split up `ChatList`'s search sections`` is the main one.

Regarding performance: this ensures that we do not send
the contact and messages search queries (with an empty query)
to the backend as soon as the chat list is rendered.
Now the search queries will only get sent when the query is not
empty, i.e. when we actually do need to show the search results.
This is thanks to the fact that the new "search reults" components
are unrendered initially.
Namely, previously we would always load _all_ the contacts
from the backend.

This makes switching between accounts faster.

### DIFF
--- a/packages/frontend/src/components/chat/ChatList.tsx
+++ b/packages/frontend/src/components/chat/ChatList.tsx
@@ -636,7 +636,7 @@ function ContactAndMessageSearchResults({
       messageResultIds,
       messageCache,
       queryStr,
-      isSingleChatSearch: queryChatId != null,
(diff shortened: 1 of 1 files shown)


nicodh (approved):
Reviewed by testing & reading

Besides the performance improvements the split makes the code better readable!

WofWca:
Rebased.

PR #5587, opened 2025-10-21
fix: Revert wrong translation of DC_STR_CANT_DECRYPT_OUTGOING_MSGS

Hocuri:
When we notice that we can't decrypt some outgoing message, then we add a device message DC_STR_CANT_DECRYPT_OUTGOING_MSGS (175). However, it looks like we forgot to add this to Transifex, and lately it was wrongly set to `systemmsg_cannot_decrypt` in https://github.com/deltachat/deltachat-desktop/commit/aef000a3cdfd46b89d45840241afba49009cb1d5

This PR reverts this.

### DIFF
--- a/packages/frontend/src/stockStrings.ts
+++ b/packages/frontend/src/stockStrings.ts
@@ -25,6 +25,8 @@ export async function updateCoreStrings() {
     | C.DC_STR_SERVER_TURNED_OFF
     | C.DC_STR_SYNC_MSG_SUBJECT
     | C.DC_STR_SYNC_MSG_BODY
+    | C.DC_STR_CANT_DECRYPT_OUTGOING_MSGS
+    | C.DC_STR_CANTDECRYPT_MSG_BODY
 
     // Deprecated, see
     // https://github.com/chatmail/core/blob/main/deltachat-ffi/deltachat.h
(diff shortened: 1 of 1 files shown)


Hocuri:
Hm, I don't know why CI fails, and would be glad if someone else could push a fix

Also, you could actually also remove the DC_STR_CANTDECRYPT_MSG_BODY translation while you're on it: https://github.com/chatmail/core/pull/7327

packages/frontend/src/stockStrings.ts
```diff
@@ -71,6 +71,7 @@ export async function updateCoreStrings() {
     | C.DC_STR_SECURE_JOIN_CHANNEL_QR_DESC
     | C.DC_STR_SECUREJOIN_WAIT
     | C.DC_STR_AEAP_EXPLANATION_AND_LINK
+    | C.DC_STR_CANT_DECRYPT_OUTGOING_MSGS
```
WofWca: This section is for deprecated strings. But `DC_STR_CANT_DECRYPT_OUTGOING_MSGS` is not marked as deprecated.

https://github.com/chatmail/core/blob/498a8318734de4a11bed6f31a7a1fd36e385e705/deltachat-ffi/deltachat.h#L7681-L7685

Of course the same goes for the new "call" string, but that's for another MR perhaps..

Otherwise looks good.

Hocuri: `DC_STR_CANT_DECRYPT_OUTGOING_MSGS` is removed in deltachat.h now: https://github.com/chatmail/core/blob/main/deltachat-ffi/deltachat.h

I don't think we need to deprecate stock strings before removing them? We can just remove them from core, and then directly remove them from all UIs.

WofWca: I meant the type definition above. It has a section for strings that are marked deprecated in core, and a section for strings that are absent for some other reason.

nicodh: I moved the string up and added C.DC_STR_CANTDECRYPT_MSG_BODY

Both will be removed with next core release anyway...

Hocuri:
Thanks!

PR #5589, opened 2025-10-22
refactor: touch up useDraft

WofWca:


nicodh (approved):
Reviewed by reading & testing

This is mainly about renaming & commenting plus some minor cleanups

Concerning the comments:
I really would like to have some more comments on the fact that we have 2 states here and why. Like
_Note: since the textarea value changes on every keystroke, and we don't want to trigger React re-renders on every keystroke (performance) the textarea manages its own state and the text is copied to draftState. We only sync to React state when needed._

Maybe I will propose some comments in a separate followup.

WofWca:
I'm still working on `useDraft` right now, so IDK if it's worth it, we'll probably have conflicts.

PR #5590, opened 2025-10-22
feat(ui): Implement drag-and-drop to reorder accounts

nicodh:
This is a "take over" of PR #5279 which is somehow abandoned

It adds some missing changes and a basic e2e test

### DIFF
--- a/packages/frontend/src/components/AccountListSidebar/AccountItem.tsx
+++ b/packages/frontend/src/components/AccountListSidebar/AccountItem.tsx
@@ -254,7 +254,6 @@ export default function AccountItem({
     <div
       className={classNames(styles.accountWrapper, {
         [styles.isSticky]: isSticky,
-        'sticky-account': isSticky,
       })}
     >
       <button
--- a/packages/frontend/src/components/AccountListSidebar/AccountListSidebar.tsx
+++ b/packages/frontend/src/components/AccountListSidebar/AccountListSidebar.tsx
@@ -64,7 +64,6 @@ export default function AccountListSidebar({
 
   const [{ accounts: noficationSettings }] = useAccountNotificationStore()
 
-  // Use drag and drop hook
   const {
     draggedAccountId,
     dropIndicator,
--- a/packages/frontend/src/components/AccountListSidebar/styles.module.scss
+++ b/packages/frontend/src/components/AccountListSidebar/styles.module.scss
@@ -72,6 +72,7 @@
         position: sticky;
         bottom: var(--als-avatar-margin);
         top: var(--als-avatar-margin);
+        // Only needed when this account is scrolled _above_ the visible region.
         z-index: map.get(
           variables.$z-index,
           account-list-sidebar-scope-sticky-account


packages/frontend/src/components/AccountListSidebar/AccountListSidebar.tsx
```diff
@@ -78,6 +85,91 @@ export default function AccountListSidebar({
   }
 
   const [syncAllAccounts, setSyncAllAccounts] = useState(true)
+
+  // Drag and drop handlers
```
WofWca: I think it'd be nice to separate all this into a hook.

packages/frontend/src/components/AccountListSidebar/AccountListSidebar.tsx
```diff
@@ -78,6 +85,91 @@ export default function AccountListSidebar({
...
+  const handleDragLeave = () => {
+    setDropIndicator(null)
+  }
+
+  const handleDragEnd = () => {
+    setDraggedAccountId(null)
+    setDropIndicator(null)
+  }
+
+  const handleDrop = async (e: React.DragEvent) => {
```
WofWca: Since this function doesn't utilize `await` much, I think It'd be nice to convert it to a sync one. Sync is easier to follow.

packages/frontend/src/components/AccountListSidebar/AccountListSidebar.tsx
```diff
@@ -183,17 +275,37 @@ export default function AccountListSidebar({
                 ⚠️
               </button>
             ) : (
-              accountsFetch.lingeringResult?.value.map(id => (
-                <AccountItem
+              accountsFetch.lingeringResult?.value.map((id, index) => (
+                <div
```
WofWca: While, based on my tests, screen-reader accessibility seems to be unaffected, I think it would be nice not to wrap `<li>` elements into anything else.

A simple solution would be to change this tag to `<li>` and change the `AccountItem` tag to `div`.

packages/frontend/src/components/AccountListSidebar/AccountListSidebar.tsx
```diff
@@ -63,6 +63,13 @@ export default function AccountListSidebar({
 
   const [{ accounts: noficationSettings }] = useAccountNotificationStore()
 
+  // Drag and drop state
+  const [draggedAccountId, setDraggedAccountId] = useState<number | null>(null)
+  const [dropIndicator, setDropIndicator] = useState<{
+    index: number
+    position: 'top' | 'bottom'
```
WofWca: Looking at behavior, it seems that `{ index: 0, position: 'bottom' }` is not much different from `{ index: 1, position: 'top' }`? But I guess it's fine.

packages/e2e-tests/tests/account-reordering.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+  const userA = getUser(0, existingProfiles)
+  const userB = getUser(1, existingProfiles)
+  const userC = getUser(2, existingProfiles)
+
+  // Wait for account items to be rendered and finish loading
+  const accountItems = page.locator('[data-testid^="account-item-"]')
+  await waitForAccountItemsToFinishLoading(page)
+
+  // Get initial account order
+  const initialOrder = await accountItems.allTextContents()
```
WofWca, line 76: One can utilize something similar to

https://github.com/deltachat/deltachat-desktop/blob/e75c0d712323734fd2ee2e8ad97c78e0bc915375/packages/e2e-tests/tests/basic-tests.spec.ts#L200-L205

packages/e2e-tests/tests/account-reordering.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+  // Perform drag-and-drop: move userC to position after userB
+  await accountC.dragTo(accountB, {
+    targetPosition: { x: 0, y: 20 }, // Drop on top half of (before) userB
+    force: true,
+  })
+  await waitForAccountItemsToFinishLoading(page)
+
+  const orderAfterDrag = await accountItems.allTextContents()
+  // should be [' C', ' B', ' A'] if no previous accounts existed
+  expect(orderAfterDrag).not.toEqual(initialOrder)
```
WofWca:
```suggestion
```

Duplicated.

packages/frontend/src/components/AccountListSidebar/AccountListSidebar.tsx
```diff
@@ -78,6 +85,91 @@ export default function AccountListSidebar({
...
+          (dropIndicator.position === 'bottom' &&
+            dragIndex < accounts.length - 1))
+      ) {
+        // This case is handled by dropping on the adjacent item, so we can ignore it.
+      } else {
+        setDraggedAccountId(null)
+        setDropIndicator(null)
+        return
+      }
+    }
```
WofWca: This looks more complex than it should be.
But alright, let's leave this for later.

WofWca (approved):
Seems to work nicely! And even if it doesn't, it doesn't look like this can break things much.

None of my comments are critical, but it would be nice to address at least the "HTML tags" thing.

Thanks for finishing this up!

Edit: also, seems like a squash merge would be appropriate.

packages/frontend/src/components/AccountListSidebar/AccountListSidebar.tsx
nicodh: I swapped the tags. For accessibility I plan to add a menu item "Move to top" to each account like in ANdroid

packages/frontend/src/components/AccountListSidebar/AccountListSidebar.tsx
nicodh: I agree - and moved it into an own hook

packages/frontend/src/components/AccountListSidebar/styles.module.scss
```diff
@@ -75,15 +85,6 @@
     margin-top: var(--als-avatar-margin);
 
     &.isSticky {
-      position: sticky;
-      bottom: var(--als-avatar-margin);
-      top: var(--als-avatar-margin);
-      // Only needed when this account is scrolled _above_ the visible region.
```
WofWca, line 81: Why does the comment have to be removed?

packages/frontend/src/components/AccountListSidebar/AccountListSidebar.tsx
```diff
@@ -63,12 +64,16 @@ export default function AccountListSidebar({
 
   const [{ accounts: noficationSettings }] = useAccountNotificationStore()
 
-  // Drag and drop state
-  const [draggedAccountId, setDraggedAccountId] = useState<number | null>(null)
-  const [dropIndicator, setDropIndicator] = useState<{
-    index: number
-    position: 'top' | 'bottom'
-  } | null>(null)
+  // Use drag and drop hook
```
WofWca:
```suggestion
```

packages/frontend/src/components/AccountListSidebar/AccountItem.tsx
```diff
@@ -251,9 +251,10 @@ export default function AccountItem({
   // for a different account, and upon initial render.
 
   return (
-    <li
+    <div
       className={classNames(styles.accountWrapper, {
         [styles.isSticky]: isSticky,
+        'sticky-account': isSticky,
```
WofWca:
```suggestion
```

Unused?

packages/frontend/src/components/AccountListSidebar/AccountListSidebar.tsx
```diff
@@ -78,6 +90,7 @@ export default function AccountListSidebar({
   }
 
   const [syncAllAccounts, setSyncAllAccounts] = useState(true)
+
```
WofWca, line 92:
```suggestion
```

nicodh:
> Good catch on the sticky stuff.

Would be nice to have a test for that, maybe with https://playwright.dev/docs/api/class-page#page-set-viewport-size

PR #5591, opened 2025-10-22
Fix react warnings in stores & AudioRecorder

nicodh:
#skip-changelog - no visible changes

### DIFF
--- a/packages/frontend/src/stores/messagelist.ts
+++ b/packages/frontend/src/stores/messagelist.ts
@@ -10,7 +10,7 @@ import {
   defaultChatViewState,
 } from './chat/chat_view_reducer'
 import { ChatStoreScheduler } from './chat/chat_scheduler'
-import { useEffect, useRef, useMemo, useState } from 'react'
+import { useEffect, useMemo, useState } from 'react'
 import { useDebouncedCallback } from 'use-debounce'
 import { debounce } from 'debounce'
 import { getLogger } from '@deltachat-desktop/shared/logger'
@@ -78,20 +78,17 @@ export function useMessageList(
   // when it's released.
   const settingsStore = useSettingsStore()[0]
 
-  const incomingMessageAudioElementRef = useRef<HTMLAudioElement>(null)
-
-  useMemo(() => {
-    if (incomingMessageAudioElementRef.current == null) {
-      incomingMessageAudioElementRef.current = document.createElement('audio')
-      incomingMessageAudioElementRef.current.src = './audio/sound_in.wav'
-    }
-
-    const volume = settingsStore?.desktopSettings.inChatSoundsVolume
-    if (volume != null) {
-      // Note that `volume` could be 0.
-      incomingMessageAudioElementRef.current!.volume = volume
-    }
-  }, [settingsStore?.desktopSettings.inChatSoundsVolume])
+  const incomingMessageAudioElement = useMemo(() => {
+    const el = document.createElement('audio')
+    el.src = './audio/sound_in.wav'
+    return el
+  }, [])
+  const volume = settingsStore?.desktopSettings.inChatSoundsVolume
+  if (volume != null) {
+    // Note that `volume` could be 0.
+    // eslint-disable-next-line react-hooks/immutability
+    incomingMessageAudioElement.volume = volume
+  }
 
   useEffect(() => {
     const cleanup = [
@@ -107,8 +104,8 @@ export function useMessageList(
           // Note that the element might already be playing,
           // if we received two or more messages rapidly.
           // In that case it could be nice to play multiple sounds in parallel.
-          incomingMessageAudioElementRef.current!.currentTime = 0
-          incomingMessageAudioElementRef.current!.play()
+          incomingMessageAudioElement.currentTime = 0
+          incomingMessageAudioElement.play()
         } else {
           store.log.debug(
             `chatId of IncomingMsg event (${chatId}) doesn't match id of selected chat (${eventChatId}). Skipping.`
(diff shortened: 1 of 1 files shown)


nicodh:
tbd: most of the warnings are caused by access to ref during rendering, so in many cases the solution would be to wrap these access in useEffect...

packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
```diff
@@ -57,7 +63,16 @@ const Timer = () => {
...
+  const [totalWidth, setTotalWidth] = useState(0)
+  useLayoutEffect(() => {
+    const el = volumeBarRef.current
+    if (!el) return
+    const updateWidth = () => setTotalWidth(el.clientWidth)
+    const ro = new ResizeObserver(updateWidth)
+    ro.observe(el)
+    updateWidth()
+    return () => ro.disconnect()
+  }, [])
```
WofWca: Does this really make no behavioral changes?

packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
```diff
@@ -57,7 +63,16 @@ const Timer = () => {
 const VolumeMeter = (prop: { volume: number }) => {
   const steps = 10
   const volumeBarRef = useRef<HTMLDivElement>(null)
-  const totalWidth = Number(volumeBarRef.current?.clientWidth) || 0
+  const [totalWidth, setTotalWidth] = useState(0)
+  useLayoutEffect(() => {
+    const el = volumeBarRef.current
+    if (!el) return
```
WofWca: Do we expect the element to be nullish here? Looks like that would be a bug?

packages/frontend/src/stores/messagelist.ts
```diff
@@ -78,19 +78,20 @@ export function useMessageList(
...
+    }
+
     const volume = settingsStore?.desktopSettings.inChatSoundsVolume
     if (volume != null) {
       // Note that `volume` could be 0.
-      incomingMessageAudioElement.volume = volume
+      incomingMessageAudioElementRef.current!.volume = volume
     }
-  }
+  }, [settingsStore?.desktopSettings.inChatSoundsVolume])
```
WofWca: I would go for `useMemo`. Then it complains about the fact that you can't set `volume`, but I'd simply ignore the warning.
I don't think that we want a situation where `incomingMessageAudioElementRef.current` is `null`, but with this change it's now possible. I think ignoring the "react/immutable" warning is better than having a non-null assertion.

packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
nicodh: It doesn't harm and if someone adds a condition to the rendering of the element if will prevent an error

packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
nicodh: I tested it. ResizeObserver should do the job instead the previous direct ref.current access

packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
WofWca: > will prevent an error

It will prevent an error from being visible to the developer, but it will not prevent the error itself. The effect runs only once. If the element is not present on the first render, then we will never change `totalWidth` from 0.

Looks like we need to utilize `ref={(newRef) => doStuff(newRef)}` to prevent such an error.

But I would be fine with not suppressing the error first.

WofWca:
> wrap these access in useEffect

I don't think that `useEffect` is the solution in most of the cases.

https://react.dev/reference/react/useEffect :

> synchronize a component with an external system

There is no "external system" most of the time.

Especially that `useEffect` doesn't run synchronously, so such an approach is very likely to introduce bugs, as I mentioned in https://github.com/deltachat/deltachat-desktop/pull/5591#discussion_r2451406094.

packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
WofWca: What I'm saying is that it looks like this actually fixes a bug, i.e. now the size updates on resize, and not only on re-render.

WofWca:
About `useRef` again: React docs themselves seem to suggest this approach of doing stuff with the ref during render: https://react.dev/reference/react/useRef#avoiding-recreating-the-ref-contents

I am not what this means for us, maybe that we can ignore the warnings. Or maybe there is a way to work around them somehow, e.g. by separating such usages into a hook. For example, I noticed that `useHasChanged2` does not produce the "accessing ref during render" warnings for some reason.

nicodh:
> I don't think that useEffect is the solution in most of the cases.

That's why I said tbd.
The linter says: "Refs should only be accessed outside of render, such as in event handlers or effects"

So it seems there is no easy generic approach  to fix all these warnings.

For the AudioRecorder I have another idea, but that means every single of the 70 warnings has to be fixed separately :-((

packages/frontend/src/components/AudioRecorder/AudioRecorder.tsx
WofWca: Niiiiiice!

packages/frontend/src/stores/messagelist.ts
```diff
@@ -78,19 +78,20 @@ export function useMessageList(
...
```
```suggestion
  const incomingMessageAudioElement = useMemo(() => {
    const el = document.createElement('audio')
    el.src = './audio/sound_in.wav'
    return el
  }, [])
  {
    const volume = settingsStore?.desktopSettings.inChatSoundsVolume
    if (volume != null) {
      // Note that `volume` could be 0.
      incomingMessageAudioElement.volume = volume
    }
  }
```

This is what I meant in https://github.com/deltachat/deltachat-desktop/pull/5591/files#r2451406094.

PR #5592, opened 2025-10-23
refactor: use state instead ref

nicodh:
Removes another 16 react warnings

useState should do the same job as useRef as it is read-only so the initial state should be kept

#skip-changelog - refactoring only


WofWca (approved):
Not sure if this affects performance, but yep, looks good. Makes sense and does not affect behavior.

PR #5599, opened 2025-10-24
refactor: don't require file name to set draft

WofWca:
The Core is able to figure out the file name itself.


packages/frontend/src/components/composer/menuAttachment.tsx
```diff
@@ -55,7 +55,7 @@ export default function MenuAttachment({
         for (const filePath of filePaths) {
           await sendMessage(accountId, selectedChat.id, {
             file: filePath,
-            filename: basename(filePath),
+            // filename
```
nicodh, line 58: left over?

WofWca: I'd keep it, because it makes it clear that one can specify `filename` here.

PR #5601, opened 2025-10-24
perf: make message quoting instant

WofWca:
TODO:
- [x] Address the TODOs.
- [x] Merge the base into main, up to 1 commit.

### DIFF
--- a/packages/frontend/src/hooks/chat/useDraft.ts
+++ b/packages/frontend/src/hooks/chat/useDraft.ts
@@ -365,6 +365,47 @@ export function useDraft(
     [draftState, inputRef, saveAndRefetchDraft, debouncedSaveAndRefetchDraft]
   )
 
+  const quoteMessage = useMemo(
+    () =>
+      setAndDebouncedSaveAndRefetchDraft == null ||
+      debouncedSaveAndRefetchDraft == null
+        ? null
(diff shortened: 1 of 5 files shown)


packages/frontend/src/hooks/chat/useDraft.ts
```diff
@@ -513,3 +536,62 @@ export function useDraft(
...
+   * to show the returned quote in the UI,
+   * or whether we need to re-fetch the quote from core right away.
+   */
+  needRefetch: boolean
+} {
+  return {
+    needRefetch: !(
+      // Obscure `viewType`s usually have some special quote text.
+      // For example, quotes with a contact attachment
+      // have `text` '<person emoji> <Contact name>',
```
nicodh, line 679: What do you mean here? A Vcard? I tried it and I think Vcards don't need refetch either.

What viewTypes do need a refetch?

WofWca: Yes, a VCard. Try first quoting VCard message with no text, then re-entering the chat. You'll see that the text changes.

image


> What viewTypes do need a refetch?

I suggest to use a "white-list" approach. I.e. all message types that are not in the white list.
To name a few: WebXDC, Voice messages, Videos

image image

WofWca:
Rebased.

PR #5604, opened 2025-10-26
refactor: move `loadingDraft` state to `useDraft`

WofWca:
To make things less imperative.
Maybe now the value of the variable is not the same in all cases,
but the behavior should be pretty much the same,
or maybe even better.

### DIFF
--- a/packages/frontend/src/hooks/chat/useDraft.ts
+++ b/packages/frontend/src/hooks/chat/useDraft.ts
@@ -49,6 +49,13 @@ export function useDraft(
   inputRef: React.RefObject<ComposerMessageInput | null>
 ): {
   draftState: DraftObject
+  /**
+   * Whether the initial loading of the draft is being performed,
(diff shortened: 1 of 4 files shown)


packages/frontend/src/hooks/chat/useDraft.ts
```diff
@@ -255,7 +271,6 @@ export function useDraft(
       } else {
         clearDraftStateButKeepTextareaValue()
       }
-      inputRef.current?.setState({ loadingDraft: false })
     },
     [abortController, clearDraftStateButKeepTextareaValue, inputRef]
```
nicodh: Why is inputRef included here? It is not used in the callback and saveAndRefetchDraft_ is called by updateDraftText anyway on each change?

WofWca: Hm yep, I think I simply missed it. Can be removed, I'll force-push.

PR #5605, opened 2025-10-26
fix: crash when clicking settings when loading

WofWca:
It happens e.g. when you click "Settings"
before the settings have loaded.
The non-null assertion was incorrect.

The crash usually happens when running Playwright tests.


Simon-Laux:
- possibly related: https://github.com/deltachat/deltachat-desktop/issues/4200

PR #5606, opened 2025-10-26
test: add tests for composer's draft

WofWca:


packages/e2e-tests/tests/composer.spec.ts
```diff
@@ -0,0 +1,304 @@
...
+    isChatmail
+  )
+
+  await contextForProfileCreation.close()
+  page = await browser.newPage()
+  await reloadPage(page)
+
+  chatList = page.getByLabel('Chats').getByRole('tablist')
+  textarea = page.getByRole('textbox', { name: 'Write a message' })
+  composerSection = page.getByRole('region', { name: 'Write a message' })
```
nicodh, line 63: Do we really want to rely on translated wordings? I see the benefit of using findByRole as described [here](https://qaexpertise.com/playwright/an-in-depth-understanding-of-getbyrole-in-playwright/) but any change of the used translations would break the test.
Checking for accessibility as mentioned in the referred post is a separate concern imo.
Why do you prefer getRole to getByTestid?

packages/e2e-tests/tests/composer.spec.ts
```diff
@@ -0,0 +1,304 @@
...
+
+  const context = await browser.newContext()
+  const pageForProfileDeletion = await context.newPage()
+  await reloadPage(pageForProfileDeletion)
+  await deleteAllProfiles(pageForProfileDeletion, existingProfiles)
+  await context.close()
+})
+
+async function createNDummyChats(n: number) {
+  for (let i = 0; i < n; i++) {
```
nicodh: Move to playwrightHelper? It's a duplicate of createNDummyChats in chat-list-multiselect.spec

packages/e2e-tests/tests/composer.spec.ts
```diff
@@ -0,0 +1,304 @@
...
+  await expect(textarea).not.toBeDisabled()
+
+  await textarea.pressSequentially(text, { delay: 50 })
+
+  // Seems to be necessary, otherwise it's flaky.
+  // This might indicate a bug in our app, actually.
+  await expect(textarea).toHaveText(text)
+}
+
+test.describe('draft', () => {
```
nicodh, line 100: This is a very scarce test description - maybe rename it to "test suite for draft tests"

packages/e2e-tests/tests/composer.spec.ts
```diff
@@ -0,0 +1,304 @@
...
+
+    await someOtherChat.click()
+    await expect(composerSection).not.toContainText(messageText, {
+      timeout: 300,
+    })
+    await expect(composerReply).not.toBeVisible({ timeout: 1 })
+
+    // This is currently failing sometimes, probably due to some
+    // `ChatlistItemChanged` race. TODO.
+    // await expect(targetChat).toContainText(`Draft: Reply`)
```
nicodh, line 154: Very cool that we have a test for that now!

packages/e2e-tests/tests/composer.spec.ts
WofWca: > but any change of the used translations would break the test

The same can be said about changing class names. I think in this regard the translation is even more reliable. Class names and similar are an implementation detail, which can be changed by us at any time.
These are E2E tests, and `getByRole` describes semantics better than some "test ID".
And yes, this also tests accessibility along the way.

Playwright docs recommend the same https://playwright.dev/docs/locators :

> We recommend prioritizing role locators to locate elements, as it is the closest way to how users and assistive technology perceive the page.

Although perhaps the presence of a composer textarea is something that is never going to change, thus it can be accessed by ID, instead of accessible label, so, regarding the textarea, I'm 50:50 here.

packages/e2e-tests/tests/composer.spec.ts
WofWca: Um I don't think so. "test suite for ... tests" is what `describe` implies IMU.

packages/e2e-tests/tests/composer.spec.ts
WofWca: Makes sense. I force-pushed.

packages/e2e-tests/tests/composer.spec.ts
nicodh: > The same can be said about changing class names.

Yes I agree I also try to avoid class names as locator. But the name of data-testid attribute  is explicit saying why it's needed. And moving stuff around or rewording would not brake the tests.

Even if someone adds an exclamation mark "Write a message!" it would break the tests. And as soon as we have many locators that rely on a specific wording it might get confusing

What about using the translation key and translate it in the test? Then we are independent of the wording and still have proper semantics

WofWca: OK, I force-pushed to use the ID for the textarea.

> Even if someone adds an exclamation mark "Write a message!"

"well ackchually" it wouldn't, because `getByRole... { name: 'Write a message' }` would still match it.

> What about using the translation key and translate it in the test?

I thought about this, but I think it's not worth it. Reading and writing tests is gonna become harder, plus again this reduced their "end-to-end"-ness. For example, what happens if we accidentally mix up two strings? Replace 'Write a message' with 'Search'. The tests would keep passing. I'm not saying that this paticular example is likely, but yep, this is about the code being separate from the E2E tests.

Updating tests when translations change should not be hard, and we don't expect it to happen often.

Oh, and also I think it's actually a good think that updating translations breaks the tests. This forces us to re-evaluate whether the strings still make sense in the context that they are used in.

nicodh (approved):

packages/e2e-tests/tests/composer.spec.ts
nicodh: ok, arguments accepted! :-)

packages/e2e-tests/tests/composer.spec.ts
nicodh: Or almost have one, if the bug is fixed :-)

PR #5607, opened 2025-10-26
fix: missing css class (found via typescript warning)

Simon-Laux:
#skip-changelog because there is no user visible change


WofWca (approved):
FYI there is a lot more of such warnings.

PR #5610, opened 2025-10-26
refactor: don't use basename for file name

WofWca:
Using `basename()` in the browser environment is not correct,
because it's just a polyfill, and e.g. it doesn't work properly
on Windows, i.e. `basename()` returns the whole path unchanged.

Note that now the file name will be momentarily _not_ shown,
until draft has been re-fetched (see `DraftAttachment`).
But that's probably better than showing the entire path on Windows.

TODO:
- ~~[ ] Also remove the `parse(path)`.~~
  That's a separate topic, can do later.
- ~~[ ]~~ Consider writing our own `basename` which we'll be using only for displaying the file name when `filename == null && path != null`, e.g. here:
  https://github.com/deltachat/deltachat-desktop/blob/dc11bbc7f9de67bda59782fdd226378306928c52/packages/frontend/src/components/attachment/messageAttachment.tsx#L315
  
  Also can be done later.
- [x] When done, see if this comment can be removed:
  https://github.com/deltachat/deltachat-desktop/blob/6b3a03e401d1317e541ce23b9f639f5afca4e332/packages/frontend/src/hooks/chat/useDraft.ts#L212-L215

  Edit: already removed on current main

### DIFF
--- a/packages/frontend/src/components/composer/Composer.tsx
+++ b/packages/frontend/src/components/composer/Composer.tsx
@@ -65,7 +65,7 @@ const Composer = forwardRef<
     updateDraftText: (text: string, InputChatId: number) => void
     addFileToDraft: (
       file: string,
-      fileName: string,
+      fileName: string | null,
       viewType: T.Viewtype
     ) => Promise<void>
     removeFile: () => void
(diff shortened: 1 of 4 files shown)


WofWca:
Ah I just found out that we are intentionally using `path`.

https://github.com/deltachat/deltachat-desktop/blob/5aff81758d1c6cbfa8992614ba8935b7cdf8714c/packages/frontend/bin/build-frontend-ts.mjs#L39-L41

https://github.com/browserify/path-browserify/issues/1

nicodh (approved):
Reviewed by reading

PR #5615, opened 2025-10-27
fix: some more `useDraft` races

WofWca:
`abortController` (introduced in
95e808afc56bf7cef7ecf063896d33878601cc1a)
was nice, but it's not enough, e.g. to guard against
multiple concurrent `__reloadDraft`s.
Let's just finally migrate to the good old `useRpcFetch` approach.

TODO:
- [ ] Merge the base such that only 1 commit remain.
- [x] Merge https://github.com/deltachat/deltachat-desktop/pull/5613 to fix E2E tests.


WofWca:
This is based on a rejected MR.
Edit: that is to say, by default this MR should be rejected, unless someone strongly feels that this MR implements the right way.

WofWca:
OK, let's close then.

PR #5618, opened 2025-10-28
feat: do not use HTTPS request for default instant onboarding

link2xt:
Depends on the core PR: <https://github.com/chatmail/core/pull/7360>
Using domain name in dcaccount scheme works
similarly to existing scheme,
but no HTTPS request is made
and credentials are generated locally on the client.


link2xt:
CI currently fails because of no core update.

link2xt:
We have core 2.23.0 now, so once it is updated, this PR can be rebased.

WofWca (approved):
I'm not sure exactly what it does, but it's the Core's business. Looks good, the tests have been fixed.

nicodh (approved):

PR #5630, opened 2025-10-29
fix: improper display of blocked contacts

### LINKED ISSUE #5493: Blocked Contacts dialog has weird scrollbars
Image

There is an additional scrollbar shown, which should not be there. 
And IMO there is too much spacing left and right around the blocked contacts and text of the items is cut off even though there is still plenty of space available:

Image

---

This dialog still doesn't use a virtual list, doing that that may make sense. -> also a search would make sense, but that would be a new feature request.

nicodh:
resolves #5493

#skip-changelog just a minor styling fix

### DIFF
--- a/packages/frontend/scss/contact/_contact.scss
+++ b/packages/frontend/scss/contact/_contact.scss
@@ -6,7 +6,7 @@
   display: flex;
   flex-direction: row;
   align-items: center;
-  width: 300px;
+  width: 100%;
   .chat-list & {
     width: unset; // otherwise you can scroll sidewards sometimes
     padding: 0px 10px;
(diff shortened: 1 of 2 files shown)


packages/frontend/scss/contact/_contact.scss
```diff
@@ -6,7 +6,7 @@
   display: flex;
   flex-direction: row;
   align-items: center;
-  width: 300px;
+  width: 100%;
```
WofWca, line 9: Now super long contact names are a little ugly. Not just in this dialog.

Image

WofWca (approved):

PR #5631, opened 2025-10-29
fix: avoid console error

nicodh:
We should avoid to have errors in console.

#skip-changelog no user visible change

### DIFF
--- a/packages/frontend/src/components/chat/ChatListItem.tsx
+++ b/packages/frontend/src/components/chat/ChatListItem.tsx
@@ -307,7 +307,9 @@ function ChatListItemError({
  */
 function RegularChatListItem({
   chat,
-  chatListItem: _chatListItem, // to avoid the console error about unrecognized prop
+  // To avoid the console error about unrecognized DOM prop,
+  // simply remove it from `rest`.
+  chatListItem: _chatListItem,
   onClick,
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/chat/ChatListItem.tsx
```diff
@@ -307,6 +307,7 @@ function ChatListItemError({
  */
 function RegularChatListItem({
   chat,
+  chatListItem: _chatListItem, // to avoid the console error about unrecognized prop
```
WofWca:
```suggestion
  // To avoid the console error about unrecognized DOM prop,
  // simply remove it from `rest`.
  chatListItem: _chatListItem,
```

nicodh: What is "simply remove it from `rest`." for?

WofWca: Well the error is caused by the fact that this property is present on the `rest` object. I thought  this makes it a little more clear to someone who has no idea of what error we're talking about.

nicodh: Don't think that line is enough to make that clear :-) - but why not...

PR #5637, opened 2025-10-30
perf: `field-sizing` for composer height, not JS

WofWca:
Before / after:

image
image

As you can see, the single key processing time went down from ~7ms to ~4ms.

### DIFF
--- a/packages/frontend/scss/composer/_composer.scss
+++ b/packages/frontend/scss/composer/_composer.scss
@@ -70,6 +70,9 @@
     .microphone-button {
       background: none;
       border: none;
+
+      // So that the composer doesn't make them shrink as you type text.
+      flex-shrink: 0;
     }
 
     .attachment-button {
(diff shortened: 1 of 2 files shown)


packages/frontend/scss/composer/_composer.scss
```diff
@@ -148,13 +151,21 @@
...
+      $line-height: 24px;
+      line-height: $line-height;
       margin-top: 8px;
       margin-bottom: 8px;
       overflow-y: hidden;
       background-color: var(--composerBg);
       color: var(--composerText);
 
+      &.use-field-sizing-css-prop {
+        field-sizing: content;
```
Simon-Laux, line 163: note: firefox doesn't support this property yet: https://bugzilla.mozilla.org/show_bug.cgi?id=1832409

not sure if we need to document that, since browser version is not a release target, so probably not a big issue.
And it will probably be supported by the time we would release a browser version.. (since browser edition release is not planned currently, so will take a long time ;)

WofWca: I have implemented dynamic support checking, see the JS.

PR #5639, opened 2025-10-30
chore: add git cliff and prepare-relase script

nicodh:
To avoid huzzle with CHANGELOG conflicts and for a convenient prepare release process

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,24 +1,5 @@
 # Changelog
 
-## [Unreleased][unreleased]
-
-### Added
-- enable drag & drop for the account order
-
-### Fixed
-- show appropriate messages when scanning wrong qr codes #5563
-- fix: remove the option to select info messages with Ctrl + Up/Down #5337
-- fix: open https links case insensitive #5625
-- don't start chat for mailto links on chatmail accounts
-
-### Changed
-- Removed "Watch Sent Folder" preference. #5611
-
-### Changed
-- replace rust/wasm message-parser with javascript lib linkify
-  - removes experimental markdown support
-  - fixes some link parsing issues
-
 <a id="2_22_0"></a>
 
 ## [2.22.0] - 2025-10-17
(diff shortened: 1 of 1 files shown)


pnpm-lock.yaml
```diff
@@ -6,36 +6,9 @@ settings:
...
-      version: 2.1.35
-    sass:
-      specifier: ^1.86.3
-      version: 1.86.3
     typescript:
       specifier: ^5.9.2
       version: 5.9.2
-    typescript-plugin-css-modules:
-      specifier: ^5.2.0
-      version: 5.2.0
```
WofWca, line 53: How did this happen? Why is it that only `typescript` remains? Removing `typescript` also seems to work.
Whatever I guess.

cliff.toml
```diff
@@ -0,0 +1,94 @@
...
+    { message = "^refactor", group = "Refactor"},
+    { message = "^perf", group = "Performance"},
+    { message = "^test", group = "Tests"},
+    { message = "^style", group = "Styling"},
+    { message = "^chore\\(release\\): prepare for", skip = true},
+    { message = "^chore", group = "Miscellaneous Tasks"},
+    { message = "^build", group = "Build system"},
+    { message = "^docs", group = "Documentation"},
+    { message = "^ci", group = "CI"},
+    { message = ".*", group = "Other"},
```
WofWca, line 28: These two values seem to be conflicting. I suggest to set `filter_unconventional = false`, so that the changelog can be later adjusted and nothing is missed.

CHANGELOG.md
```diff
@@ -3195,543 +3194,6 @@ This section is only relevant to contributors.
 
 - remove version from titlebar
 
-## [0.999.1] - 2020-02-02
```
WofWca, line 3203: Why do we have to delete the old entries?

And then why do we still keep the links to them at the very bottom of the file?

bin/update_desktop_version.js
```diff
@@ -26,33 +53,47 @@ packageDirectories
...
+    console.log(
+      "\n❌ Check failed, make sure you have run 'pnpm update:target-versions'"
+    )
+    process.exit(1)
+  } else {
+    console.log('\n✅ All versions are in sync!')
+  }
+} else {
+  console.log(`\n✅ All versions updated to ${version}!`)
 }
```
WofWca, line 99: Are all the changes in this file irrelevant?

cliff.toml
```diff
@@ -0,0 +1,94 @@
...
+            [{{ release.version | trim_start_matches(pat="v") }}]: \
+                https://github.com/deltachat/deltachat-desktop\
+                /compare/v{{ release.previous.version | trim_start_matches(pat="v") }}...v{{ release.version | trim_start_matches(pat="v") }}
+        {% endif -%}
+    {% else -%}
+        [unreleased]: https://github.com/deltachat/deltachat-desktop\
+            /compare/v{{ release.previous.version | trim_start_matches(pat="v") }}...HEAD
+    {% endif -%}
+{% endfor %}
+"""
```
WofWca: This is unused?

cliff.toml
WofWca, line 28:
```suggestion
    { message = "^feat", group = "Features / Changes"},
    { message = "^fix", group = "Fixes"},
    { message = "^refactor", group = "Refactor", skip = true},
    { message = "^perf", group = "Performance"},
    { message = "^test", group = "Tests", skip = true},
    { message = "^style", group = "Styling"},
    { message = "^chore\\(release\\): prepare for", skip = true},
    { message = "^chore", group = "Miscellaneous Tasks", skip = true},
    { message = "^build", group = "Build system", skip = true},
    { message = "^docs", group = "Documentation"},
    { message = "^ci", group = "CI", skip = true},
    { message = ".*", group = "Other"},
```

Users don't need to know about refactors and stuff.

bin/update-changelog.sh
```diff
@@ -0,0 +1,91 @@
...
+
+# Get the latest tag
+LATEST_TAG=$(git describe --tags --abbrev=0 2>/dev/null || echo "")
+
+if [ -n "$NEW_VERSION" ]; then
+    # Releasing a specific version - generate for unreleased commits
+    echo "📝 Preparing release ${NEW_VERSION}"
+
+    # Generate unreleased changelog
+    git cliff --unreleased --tag $NEW_VERSION  --prepend CHANGELOG.md
```
WofWca, line 27: Does this actually work? Isn't it supposed to be `pnpm exec git-cliff`?

WofWca (approved):
Finallyy!!

I didn't test the new `.sh` scripts, but they look sane, as much as I can decipher, and are anyway only meant to be run by whoever is in charge of releasing.

I tried running `pnpm exec git-cliff --unreleased --tag 2.23.0  --prepend CHANGELOG.md` and the output looks alright.

This will probably need more adjustments as we go, but at a start we'll be manually checking and adjusting the CHANGELOG file.

cliff.toml
nicodh: >  I suggest to set filter_unconventional = false

Yes makes sense

I just copied this part for core cliff.toml since I thought it makes sense to have the same rules.

CHANGELOG.md
nicodh: Why should we keep 5 year old changes in the changelog? But I'm not opinionated on the removal, so if there are reasons to keep them, fine for me

But yes if removed the links should be removed too

cliff.toml
nicodh: As said before, I think it makes sense to have the same rules as Core

bin/update-changelog.sh
nicodh: Well for me it worked. And this script has to be executed locally only.

cliff.toml
nicodh: Yes, since I moved the creation of the bottom link to the update script. I will remove it

cliff.toml
WofWca: I don't think so. Firstly Core is a library, and its changelog is made for software developers who can actually read and understand that stuff. Secondly, why do we have to have the same rules? Thirdly, I'm not sure if the Core's rules are good either. If the changelog is equivalent to the git commit history, then one can just look at the git commit history.

nicodh:
After looking at the core CHANGELOG (which has ~ 7000 lines) I decided not to remove the old releases

Concerning the conventional commits:
it can make sense to include a change like switching from rust message-parser to linkify since that my result in new/unexpected rendering, same for upgrade of electron, especially for the community builds it might help to have such changes in the CHANGELOG

Core lists rust upgrades in CI
We listed core upgrades in "Changes" which is not separated from "Features" with current cliff config

I think we should improve/adapt the cliff.toml step wise to our needs.

It anyways has only impact for the manually executed release

cliff.toml
WofWca: So this has not been addressed?

nicodh: The rules in this file we be first applied in the next release. I will then try some variants and see what fits.

I just don't want to have to filter too many commits

link2xt: Having the same set of prefixes (`ci`, `feat`, `fix`, `refactor`, `build`, `docs` etc. and not e.g. `doc`) makes some sense, but otherwise it can be improved, core config is just good enough but we have not put much effort into improving it.

PR #5640, opened 2025-10-30
chore: add commit lint warning

nicodh:
Show warnings if a commit message doesn't follow the conventional commit format

### DIFF
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,0 +1,1 @@
+npx --no -- commitlint --edit $1 || true
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -1,0 +1,51 @@
+// Commitlint configuration for enforcing conventional commits
+// https://commitlint.js.org/
+
+export default {
+  extends: ['@commitlint/config-conventional'],
+
+  rules: {
+    // Allow lowercase for types to match git-cliff configuration
+    // Using severity 1 (warning) instead of 2 (error) to allow non-conventional commits
+    'type-case': [1, 'always', 'lower-case'],
+
+    // Enforce these specific types (warning only)
+    'type-enum': [
+      1,
+      'always',
+      [
+        // User-facing changes (appear in changelog)
+        'feat', // New features
+        'fix', // Bug fixes
+        'refactor', // Code refactoring
+        'perf', // Performance improvements
+        'style', // Code style changes
+
+        // Non-user-facing changes (not in changelog)
+        'docs', // Documentation
+        'test', // Tests
+        'chore', // Maintenance
+        'ci', // CI/CD
+        'build', // Build system
+      ],
+    ],
+
+    // Subject should not be empty (warning only)
+    'subject-empty': [1, 'never'],
+
+    // Subject should not end with period (warning only)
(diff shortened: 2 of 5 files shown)


.husky/commit-msg
```diff
@@ -0,0 +1 @@
+npx --no -- commitlint --edit $1 || true
```
WofWca, line 1: Why `npx` and not `pnpx`?

commitlint.config.js
```diff
@@ -0,0 +1,51 @@
...
+    'subject-case': [0],
+
+    // Body should have blank line (warning only)
+    'body-leading-blank': [1, 'always'],
+
+    // Footer should have blank line (warning only)
+    'footer-leading-blank': [1, 'always'],
+
+    // Max line length (warning only)
+    'header-max-length': [1, 'always', 100],
```
WofWca, line 49: That's twice as long as usually recommended, IDK if it even makes sense to have this rule then.

package.json
```diff
@@ -5,6 +5,7 @@
   "version": "2.22.0",
   "scripts": {
     "preinstall": "node ./bin/check-nodejs-version.js",
+    "prepare": "husky",
```
WofWca, line 8: I'm not really understanding the point of the Husky projects when we already have `git config --local core.hooksPath .githooks`. I tested it and it seems to work alright.
Maybe we can avoid it?

```suggestion
    "prepare": "git config --local core.hooksPath .githooks",
```

Also maybe this should be optional, and specified in DEVELOPMENT.md.

WofWca (commented):
Overall a nice addition, but I'd say it should be optional, for the sake of security.

link2xt:
I also thought about having something like this (https://github.com/chatmail/core/pull/7524#issuecomment-3612521579), but having commit hook does not fix this. It is usually not a problem that commit author forgets to write proper commit messages, either they don't know because they never read CONTRIBUTING.md/DEVELOPMENT.md, or they don't care. If they don't care, they first merge one-line PR suggestions as automatic "Update <filename>" github-generated commits, then will click "squash and merge" in the end to merge non-conventional commit with autogenerated commit message and non-conventional title or even "create a merge commit" (whatever is the last state of the green button on github).

Whatever automatic protection is, it should enforce the commits that go to `main` to be conventional commits, not the commits that go to the PR, and not locally but on GitHub. Pre-commit hook should be on the server, but as far as I see GitHub does not support this. I once implemented such protection for SVN, but there commit is essentially created by the server.

This can be probably implemented with GitHub "Rulesets": https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/about-rulesets
There is a setting that explicitly enforces commit message patterns but it is locked behind "Enterprise":
image
This can probably be emulated by requiring [linear history](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/available-rules-for-rulesets#require-linear-history) essentially only allowing rebase and squash merging, then checking all commits coming from the PR in CI. To prevent bypassing this with squash-merging into a non-conventional commit, enable "require merge queue" rule so CI checks what ends up on the main branch before it is actually merged.

nicodh:
Closing after some internal discussion

PR #5642, opened 2025-10-30
refactor: remove `useRef` from `useDraft`

WofWca:
This change has a negative side-effect,
see the new TODO comment.

It was paralyzing to have this `useRef`
because of how hard it was to understand
all these state mutations, and the purpose of them.

Performance-wise this commit should not have much of an impact,
because even before this commit we have still been re-rendering
after each update to the draft state. The re-render would happen
after `getDraft` inside of `saveAndRefetchDraft_`.
With this commit we now re-render twice:
immediately upon setting the draft, and then after `getDraft`.

This MR is mainly a preparation for https://github.com/deltachat/deltachat-desktop/pull/5643.
However, it's valuable on its own, to make changing `useDraft` easier.

## History

`draftRef = useRef` was introduced 5 years ago, in
4bbdbe390b03e15eb059ca6af15a44156413ee44
(https://github.com/deltachat/deltachat-desktop/pull/1951),
as a workaround for our misunderstanding
of how React (or rather JavaScript?) works
(also see cfb2acd744c4b0e795cd17cc08c754102c83b39d,
which says "not sure why the state change does not work...").
Namely, the fact that `setState` (`setDraft`)
does not synchronously update the state returned from `useState`.

Later we embraced the usage of `useRef` to avoid unwanted re-renders.
Namely, see the new TODO comment again.
We also relied on this not re-rendering for setting the quote,
i.e. see how we used to incorrectly type cast `as Type.MessageQuote`
before e0a90c9f383e3a18acd72442fb320f730d7c9dfe.
This approach also caused issues later, but has been worked around:
- https://github.com/deltachat/deltachat-desktop/issues/4337
- d0b1ddd37e6bfe7f6be36ecb7fd283f6af880653
    (https://github.com/deltachat/deltachat-desktop/pull/4395).

### DIFF
--- a/packages/frontend/src/hooks/chat/useDraft.ts
+++ b/packages/frontend/src/hooks/chat/useDraft.ts
@@ -1,4 +1,4 @@
-import { useState, useRef, useEffect, useCallback, useMemo } from 'react'
+import { useState, useEffect, useCallback, useMemo } from 'react'
 import { T } from '@deltachat/jsonrpc-client'
 
 import { getLogger } from '../../../../shared/logger'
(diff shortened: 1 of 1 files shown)


nicodh (approved):
Reviewed by reading & testing

Still a bit hard to read but definitely better than before. The composed function names are long but the verbosity helps to understand the cascading

WofWca:
The verbosity is going away in https://github.com/deltachat/deltachat-desktop/pull/5643!
(edit: well, not completely, but still)

WofWca:
Rebased.

Issue #5252, opened 2025-07-02
Typing in a chat, then clearing it makes the place of chat in chatlist switch, being distracting and annoying


farooqkz:
- Operating System (Linux/Mac/Windows/iOS/Android): N/A
- Delta Chat Version: 1.58.2 (git: v1.58.2-makepkg)
- Expected behavior: There should be some `debounce` or something for switching. Like if I cleared the composer and didn't type anything for 2 seconds, then the switch will happen.
- Actual behavior: It switches as soon as you clear the composer or type something
- Steps to reproduce the problem: Select a chat which is in the bottom of the chat list. Type something. It'll come to the top. Then clear it. It'll immediately go to the bottom.

WofWca:
This has to do with us saving the draft on throttle, see https://github.com/deltachat/deltachat-desktop/pull/4144
I think we should save the draft only when necessary.

nicodh:
Bug means: "Something that is not working" - distracting or annoying means enhancement or polish

Simon-Laux:
>  I think we should save the draft only when necessary.

Ideally it would save before switching chat and also at an interval to save long messages in case that something unexpected happens, like a crash of a react component or a crash of the app. - So long messages don't get completely lost in those unexpected cases.

PR #5643, opened 2025-10-30
fix: debounce instead of throttle to save draft

WofWca:
fix: debounce instead of throttle to save draft

And also save the draft when switching chats, or minimizing the app.

Reviewing commit-by-commit should be easier.

This is based on https://github.com/deltachat/deltachat-desktop/pull/5642.

The performance, based on Chromium profiling and looking at how long
`keypress` event handler takes, seems to be pretty much unaffected
compared to when we relied on `throttle` in `ComposerMessageInputProps`.
It's still ~4ms.
And otherwise it's possible that it will actually improve,
because we send the draft to the backend much more rarely now.
An indirect thing it's that the chat list has to re-render
and to be refetched much less often, because `ChatlistItemChanged`
fires when the draft gets saved.

image

Closes https://github.com/deltachat/deltachat-desktop/issues/5252.

This should also improve the situation with
https://github.com/chatmail/issue-stargate/issues/13
(but I'm not completely sure that things
aren't going to get worse at first).

This is also the "proper" resolution of
https://github.com/deltachat/deltachat-desktop/issues/3733,
which has previously been resolved by
d1fbc7ac3b16b607504a333d36f6cf51180cdae3
(https://github.com/deltachat/deltachat-desktop/pull/4144).
So this commit also somewhat reverts that one.

This also fixes a minor bug with `showSendButton`.

TODO:
- [x] Merge the base, https://github.com/deltachat/deltachat-desktop/pull/5642,
  such that only 3 commits remain.
- [ ] As this is quite a big change on quite a critical component of the app, consider doing some manual tests (though remember that we have E2E tests).
- [x] I just remembered that there is a problem with this. There is some code that calls `__reloadDraft`, which also relies on checking whether the draft is empty. But, since the draft doesn't always get saved to the core, this can result in the draft getting overridden.
  I will fix this, but the fix will not change the general approach of this MR. So this can still be reviewed otherwise.
  Edit: I submitted an MR: https://github.com/deltachat/deltachat-desktop/pull/5663

### DIFF
--- a/packages/e2e-tests/tests/composer.spec.ts
+++ b/packages/e2e-tests/tests/composer.spec.ts
@@ -25,6 +25,7 @@ let page: Page
 let chatList: Locator
 let textarea: Locator
 let composerSection: Locator
+const numDummyChats = 3
 const getChat = (chatNum: number) =>
   chatList.getByRole('tab', { name: `Some chat ${chatNum}` })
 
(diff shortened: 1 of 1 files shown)


WofWca:
Rebased.

PR #5644, opened 2025-10-30
feat: merge images and video to "gallery" tab

Simon-Laux:
likely not the correct way (because of the extra callback, but maybe not as bad as I think?), just for testing at this point.

This is a pr so that r10s can test it, because I had not enough chats with videos.

### DIFF
--- a/packages/frontend/src/components/Gallery.tsx
+++ b/packages/frontend/src/components/Gallery.tsx
@@ -36,7 +36,7 @@ import { useTranslationWritingDirection } from '../hooks/useTranslationFunction'
 
 const log = getLogger('renderer/Gallery')
 
-type MediaTabKey = 'webxdc_apps' | 'images' | 'video' | 'audio' | 'files'
+type MediaTabKey = 'webxdc_apps' | 'gallery' | 'audio' | 'files'
 
 type GalleryElement = (
   props: GalleryAttachmentElementProps & {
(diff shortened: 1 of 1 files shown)


r10s (commented):
great stuff! it is working as expected.

cannot say much to the code, tho.

background: this PR came out of some discussions about streamlining the "all media" view across platforms, other PR at https://github.com/deltachat/deltachat-ios/pull/2861 and https://github.com/deltachat/deltachat-android/pull/3976

Raiden-GH:
Sorry for interfering, but I think separating photos and videos into different tabs was an advantage. Especially for profiles with many pictures and videos. It's simply more organized, and users can find a specific video/image faster. Just because other messengers combine pictures and videos doesn't mean it's a good thing. Maybe a filter in the tab would be a good compromise?

r10s:
yip, a filter could do, but that is out of scope of this issue and would be a feature request.

the PR it is about making delta chat UIs consistent, and the combine is a very small tradeoff, if at all. no larger amounts of user ever complained that android/ios should split their gallery - and an additional tab would also come at costs wrt usability. the cross checking of whatsapp/telegram/signal underlined the impression, we did not use that as a reason in the first place

EDIT: moreover, the consistency also help as lot on multi-device, as one finds  the same things at the same places on eg. android and desktop.
the combine if also better if you get eg. a number of photos and a video from your friend's holidays, you have all in one place, whatever comes from the cam of your friend.
wrt a filter, not even telegram has one, so we could be better here - but that also underlines it is not really needed by many

r10s:
all in all, this PR lgtm - is there smth. technically blocking it?

nicodh (approved):
Reviewed by reading & testing

PR #5646, opened 2025-10-31
More contrast for icons

nicodh:
As discussed with @r10s 

The grey color of icon buttons often looked like "disabled" (dialog header buttons, SettingsButton etc.)

Additionally the contrast (especially in dark mode) was often too low

image

image

The color is now with 20% less contrast than the foreground text color

tbd: we still have grey icons in Fullscreen-Attachement view (gallery, avatar) and in the composer input. Should we change that too?

#skip-changelog

### DIFF
--- a/packages/frontend/scss/misc/_context_menu.scss
+++ b/packages/frontend/scss/misc/_context_menu.scss
@@ -63,7 +63,7 @@
     justify-content: flex-start;
 
     & .left-icon {
-      background-color: var(--iconBackground);
+      background-color: var(--iconColorPrimary);
       margin: 0 5px 0 0;
     }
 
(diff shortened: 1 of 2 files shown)


r10s:
> tbd: we still have grey icons in Fullscreen-Attachement view (gallery, avatar) and in the composer input. Should we change that too?

we should at least make "emoji" and "voicemessage" the same grey as the "attach button", that is mainly visible in dark mode:

Screenshot 2025-10-31 at 13 21 59

for the fullscreen-attachment-view: maybe being more grey is okay there, to have a more neutral overall view. maybe a hover effect (as for dialogs) would do to let it appear less disabled

Simon-Laux (approved):

PR #5651, opened 2025-11-01
bug: Use regular title bar on macOS

### LINKED ISSUE #5649: Moving the Delta Chat window is too difficult because there is no title bar
- Operating System (Linux/Mac/Windows/iOS/Android): macOS 15.7.1 (24G231)
- Delta Chat Version: 2.22.0 (git: v2.22.0)
- Expected behavior: There is a clear and obvious way to move the Delta Chat window
- Actual behavior: The area where I can click to move the Delta Chat window changes based on how long the chat title is
   1. Rename a group to have a very long name (100+ characters)
   2. Resize the window to occupy the left half of a MacBook Pro display
   3. Try to drag the window to the other side of the screen
   4. Be frustrated by the microscopic click areas which are usable for this task.
- Screenshot (Draggable areas boxed in red):
Image
- Logs: N/A

Please consider turning off the Electron "feature" that hides the title bar and extends application content up into the area where the title bar is supposed to be.

s0ph0s-dog:
Fixes #5649 by removing the special-case that hides the title bar on macOS.

Turns out this is really easy to change!

### DIFF
--- a/packages/target-electron/src/windows/main.ts
+++ b/packages/target-electron/src/windows/main.ts
@@ -1,7 +1,6 @@
 import debounce from 'debounce'
 import electron, { BrowserWindow, Rectangle, session } from 'electron'
 import { isAbsolute, join, sep } from 'path'
-import { platform } from 'os'
 import { fileURLToPath } from 'url'
 import { Session } from 'electron/main'
 


nicodh:
Thank you for contributing to Deltachat desktop!

> Turns out this is really easy to change!

It's easy to change if we always hide it. But it seems many people prefer to have it hidden  and there was put quite some effort to have it hidden
https://github.com/deltachat/deltachat-desktop/pull/4300

we also had a previous PR that changed it to be shown by default
https://github.com/deltachat/deltachat-desktop/commit/0d0f197d9ac82adef8db411f1cfe3383eaa3f862

Unfortunately it seems not possible to hide/unhide it dynamically. We could add a setting to let the user configure it, but that needs to be discussed, since we try to have not too many setting options to not confuse people.

On the other hand usability has a higher priority than design (if in doubt)

@Simon-Laux you implemented most of the stuff around the hidden titlebar iirc, what is your opinion on this?

Simon-Laux:
I would prefer fixing/improving the custom title bar as it looks much better and is also what many other apps are doing.

Even macOS system apps like finder do this title bar with controls/content. _(though it is not feasible to use the system ui libraries for a better titlebar, because electron is very web focused, so we are stuck with replicating it as best as we can.)_
And I'm sure we can still improve in that regard, current implementation takes heavy inspiration from discord and element, maybe not the best examples.

For the mean-time I would agree with an op-in experimental setting to enable the system title-bar, but I would like to keep the current default. And ideally remove it again once the custom titlebar is fixed.

nicodh:
Maybe let the name be part of the draggable area and narrow the profile click to the avatar would be a solution, but tbh I doubt how important it is for people to have a titleBar less window

nicodh:
> For the mean-time I would agree with an op-in experimental setting to enable the system title-bar, but I would like to keep the current default. And ideally remove it again once the custom titlebar is fixed.

What would be the fix for a custom title bar?

Simon-Laux:
> What would be the fix for a custom title bar?

https://github.com/deltachat/deltachat-desktop/issues/5649#issuecomment-3476287196 would be a start

> Maybe let the name be part of the draggable area and narrow the profile click to the avatar would be a solution, 

no, that would be too confusing for users. they would not find the option to view the group.

r10s:
i would also like to keep the "macOS stylish" bar, it is known and loved by many ppl.

only if it is unfixable, we should go back to the classic ones. we should not do an option for that, not even experimental.

ftr,  not  _only_ left and right of the title the windows is draggable, also _above_ the title. however, wondering, why the title cannot be used for dragging as well as for tapping? this usually is not an either-or (see the new draggable profiles, which as still tappable)

Simon-Laux:
> why the title cannot be used for dragging as well as for tapping?

electron limitation, https://github.com/electron/electron/issues/47854

we could add a native dependency to workaround that (like https://github.com/noobfromph/electron-click-drag-plugin).

There are also some concrete ideas in https://github.com/deltachat/deltachat-desktop/issues/5649#issuecomment-3476287196 to improve it even without a new native dependency, I believe we should start with those and then think about adding the native npm package.

slightly off topic: In tauri it is probably already solvable, because the window dragging works differently there and it is also easier to contribute to tauri to use other system apis than to build/patch/contribute-to electron. (unfortunately tauri is not ready in other areas)

s0ph0s-dog:
Alright, I see that this is not the preferred solution. I'll close this PR and leave the issue open to be dealt with when there's a solid design consensus for how to make the window consistently draggable.

PR #5652, opened 2025-11-01
refactor: dedup `avatarInitial` fn

WofWca:
And this allows to reduce the complexity
of the `extractFirstEmojisFromString` function
(renamed to `countEmojisIfOnlyContainsEmoji`),
which now only does one thing.

`tsconfig`s need to get updated in order for it to not complain
about `Property 'Segmenter' does not exist on type 'typeof Intl'`.


Simon-Laux (approved):

nicodh (approved):
Reviewed by reading & testing

PR #5653, opened 2025-11-01
refactor: count emojis: remove Linkify dependency

WofWca:
This might also improve performance.
I have tested that it still works:

image

Related:
- aa486d6ae60a8221d0a55a02f4d77b3af87aec2e (https://github.com/deltachat/deltachat-desktop/pull/3102).
- 0257c50f5ae2217e644ae27af9f3dbf5e28a2127 (https://github.com/deltachat/deltachat-desktop/pull/3813)

TODO:
- [ ] Merge the base such that only 1 commit remains.

### DIFF
--- a/packages/frontend/src/components/conversations/emoji.ts
+++ b/packages/frontend/src/components/conversations/emoji.ts
@@ -2,7 +2,6 @@
 // We only really need the emoji data of this module
 import EmojiConvertor from 'emoji-js-clean'
 import { getLogger } from '../../../../shared/logger'
-import { countEmojisIfOnlyContainsEmoji } from '../message/MessageParser'
 
 const log = getLogger('renderer/emoji')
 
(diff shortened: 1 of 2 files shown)


nicodh:
This seems to have an effect how composed emojis are rendered:
in this PR:
image

In current main:
image

WofWca:
Hmmmm, yep. Thanks for checking. `/^\p{Extended_Pictographic}{1,8}$/u.test('👨‍👨‍👧‍👧')` is `false` for some reason. `/^\p{Extended_Pictographic}{1,8}$/u.test('👩🏾‍✈️')` is also `false`.

https://unicode.org/reports/tr44/#Extended_Pictographic :

> for pictographic symbols, as well as reserved ranges in blocks largely associated with emoji characters. This enables segmentation rules involving emoji to be specified stably, even in cases where an existing non-emoji pictographic symbol later comes to be treated as an emoji.

PR #5657, opened 2025-11-01
refactor: remove unnecessary `__reloadDraft`

WofWca:
This is apparently no longer necessary after
2b3cc19c50f4b0d7910de6777fc04d15fc5ce4b8
(https://github.com/deltachat/deltachat-desktop/pull/4394)
Apparently the issue was about the fact that
the composer will not get un-disabled unless we explicitly call
`setText()` or `setState({ loadingDraft: false })`.

This basically reverts
- fdf080dacdc1874deb65491bd64d416939271eb9
  (https://github.com/deltachat/deltachat-desktop/pull/4085).
- 2265667639c76ca14cf46858746562e1a39612d5
  (https://github.com/deltachat/deltachat-desktop/pull/3547).

Now the `disabled` state is controlled
with the reactive `draftIsLoading`, and the chat state
is reloaded on `ChatModified` events,
so such issues should no longer be happening,
and this imperative code can be removed

I added a test for those bugs: https://github.com/deltachat/deltachat-desktop/pull/5656. I also ran the test on this code, and it passed.

Together with https://github.com/deltachat/deltachat-desktop/pull/5663, this will allow us to remove `__reloadDraft`.

### DIFF
--- a/packages/frontend/src/components/composer/Composer.tsx
+++ b/packages/frontend/src/components/composer/Composer.tsx
@@ -19,12 +19,7 @@ import { replaceColonsSafe } from '../conversations/emoji'
 import { Quote } from '../message/Message'
 import { DraftAttachment } from '../attachment/messageAttachment'
 import { useSettingsStore } from '../../stores/settings'
-import {
-  BackendRemote,
-  EffectfulBackendActions,
-  onDCEvent,
-  Type,
-} from '../../backend-com'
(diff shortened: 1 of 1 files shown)


nicodh (approved):

WofWca:
Rebased.

WofWca:
Alright, the new tests passed, both locally and in CI, let's merge.

PR #5663, opened 2025-11-03
fix: missing "Replace draft?" prompt

WofWca:
~~This also includes https://github.com/deltachat/deltachat-desktop/pull/5675, which I suggest to review first.~~
Edit: merged separately.

Namely, when doing
- 'Share Profile'
- `webxdc.sendToChat()`
- bot command click
- `mailto:...?body=...` click

The issue was described in
https://github.com/deltachat/deltachat-desktop/pull/5643.
The problem was that, since we save the draft the the backend
on debounce and not immediately, it is not enough
to utilize `BackendRemote.rpc.getDraft()` to check whether
a draft is present.

So let's instead treat the source of truth for the draft state
as being inside of `useDraft`, and overall move more draft logic
to inside of `useDraft`.

The problem with this, however, is that `useDraft` is not always
rendered with the `accountId` and `chatId` that we need,
so we have to store the "set draft request"
until the correct chat has been selected and loaded.

Also add tests for this issue.

### DIFF
--- a/packages/frontend/src/hooks/chat/useCreateDraftMesssage.ts
+++ b/packages/frontend/src/hooks/chat/useCreateDraftMesssage.ts
@@ -1,15 +1,28 @@
 import { useCallback } from 'react'
 
 import useChat from './useChat'
-import useConfirmationDialog from '../dialog/useConfirmationDialog'
-import useTranslationFunction from '../useTranslationFunction'
-import { BackendRemote } from '../../backend-com'
+import { getLogger } from '@deltachat-desktop/shared/logger'
+import type { T } from '@deltachat/jsonrpc-client'
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import type { runtime } from '@deltachat-desktop/runtime-interface'
+
+const log = getLogger('useCreateDraftMessage')
 
 export type CreateDraftMessage = (
   accountId: number,
   chatId: number,
   messageText: string,
-  file?: { path: string; name?: string }
+  file?: {
+    path: string
+    name?: string
+    viewType?: T.Viewtype
+    /**
+     * If the provided file is a temp file, set this to `true`,
+     * and the file will be deleted (with {@linkcode runtime.removeTempFile})
+     * when we're done adding it to the draft (or if something failed).
+     */
+    deleteTempFileWhenDone: boolean
+  }
 ) => Promise<void>
 
 /**
@@ -19,45 +32,29 @@ export type CreateDraftMessage = (
  * replace it.
  */
 export default function useCreateDraftMessage() {
-  const tx = useTranslationFunction()
-  const openConfirmationDialog = useConfirmationDialog()
   const { selectChat } = useChat()
 
   return useCallback<CreateDraftMessage>(
     async (accountId, chatId, messageText, file) => {
-      const draft = await BackendRemote.rpc.getDraft(accountId, chatId)
-
       selectChat(accountId, chatId)
 
-      if (draft) {
-        const { name } = await BackendRemote.rpc.getBasicChatInfo(
-          accountId,
-          chatId
-        )
-
-        // Ask if the draft should be replaced
-        const continueProcess = await openConfirmationDialog({
-          message: tx('confirm_replace_draft', name),
-          confirmLabel: tx('replace_draft'),
-        })
-
-        if (!continueProcess) {
-          return
-        }
+      if (window.__setDraftRequest != undefined) {
+        log.error('previous createDraftMessage has not worked?')
       }
-
-      await BackendRemote.rpc.miscSetDraft(
+      window.__setDraftRequest = {
         accountId,
         chatId,
-        messageText,
-        file?.path || null,
-        file?.name || null,
-        null,
-        file ? 'File' : 'Text'
-      )
-
-      window.__reloadDraft?.()
+        file: file
+          ? {
+              ...file,
+              // Should we make `viewType` required?
+              viewType: file.viewType ?? 'File',
+            }
+          : undefined,
+        text: messageText,
+      }
+      window.__checkSetDraftRequest?.()
     },
-    [tx, openConfirmationDialog, selectChat]
+    [selectChat]
   )
 }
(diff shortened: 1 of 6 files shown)


WofWca:
I merged https://github.com/deltachat/deltachat-desktop/pull/5675 into this, which allowed to remove the `chatId` comparison.

packages/frontend/src/hooks/chat/useDraft.ts
```diff
@@ -492,6 +503,105 @@ export function useDraft(
     debouncedSaveAndRefetchDraft,
   ])
 
+  const handleSetDraftRequest = useCallback(() => {
```
nicodh, line 510:
```suggestion

  /**
   * Handle `window.__setDraftRequest` which might have been set
   * before the chat was loaded to set a draft with specified text and/or file.
   */
  const handleSetDraftRequest = useCallback(() => {
```
A comment here would help to understand what this function does without reading all code

packages/frontend/src/hooks/chat/useCreateDraftMesssage.ts
nicodh, line 1: Typo in filename: useCreateDraftMess**s**age

packages/frontend/src/hooks/chat/useDraft.ts
```diff
@@ -492,6 +503,105 @@ export function useDraft(
...
+    debouncedSaveAndRefetchDraft,
+    draftIsLoading,
+    draftState,
+    openAlertDialog,
+    openConfirmationDialog,
+    saveAndRefetchDraft,
+    tx,
+  ])
+  handleSetDraftRequest()
+  const handleSetDraftRequestEffectEvent = useEffectEvent(handleSetDraftRequest)
```
nicodh, line 599: Here some comments would also help, since this is a bit hard to read...

packages/e2e-tests/tests/composer.spec.ts
```diff
@@ -56,9 +58,28 @@ test.beforeAll(async ({ browser, isChatmail }) => {
...
+  })
+}
+async function testDraftIsEmpty() {
+  await expect(textarea).toBeEmpty({ timeout: 500 })
+
+  await expect(composerSection).toHaveText('', { timeout: 500 })
+  // The above check should be enough, but let's double check
+  // with these extra ones.
+  await expect(
+    composerSection.getByRole('region', { name: /reply|attachment|/i })
```
nicodh, line 78: when would this be visible?

packages/e2e-tests/tests/composer.spec.ts
```diff
@@ -56,9 +58,28 @@ test.beforeAll(async ({ browser, isChatmail }) => {
   chatList = page.getByLabel('Chats').getByRole('tablist')
   textarea = page.locator('textarea#composer-textarea')
   composerSection = page.getByRole('region', { name: 'Write a message' })
+  composerReply = composerSection.getByRole('region', { name: 'Reply' })
   await createNDummyChats(page, 3, 'Some chat ')
```
nicodh: "Some chat " should better go to a variable to make obvious that it's reused in several places (not part of this PR though...)

packages/frontend/src/hooks/chat/useCreateDraftMesssage.ts
WofWca: The file name has not been changed in this MR, it can be updated in a follow-up.

packages/frontend/src/hooks/chat/useDraft.ts
WofWca: What's hard to read? Would a comment above `handleSetDraftRequest` not suffice?

packages/e2e-tests/tests/composer.spec.ts
WofWca: As the comment says, it shouldn't be visible as long as the line above didn't throw.

nicodh: I don't mean in this test but the selector. I couldn't find a region with name /reply/attachment/ when testing various scenarios

packages/frontend/src/hooks/chat/useDraft.ts
nicodh: I mean this section. what is the reason why we use useEffectEvent here

```typescript
  const handleSetDraftRequestEffectEvent = useEffectEvent(handleSetDraftRequest)
   useEffect(() => {
    window.__checkSetDraftRequest = handleSetDraftRequestEffectEvent
    return () => {
      if (window.__checkSetDraftRequest === handleSetDraftRequestEffectEvent) {
        window.__checkSetDraftRequest = undefined
      }
    }
  }, [])
```

WofWca:
> The PR does more than the title says so maybe a better name could be given

What would that be? You mean adding tests? Or that another MR that I mentioned?

nicodh:
> > The PR does more than the title says so maybe a better name could be given
> 
> What would that be? You mean adding tests? Or that another MR that I mentioned?

Well skipping the prompt if it is not needed is just a detail. I would say the most important change here is to introduce a new setDraftRequest handling.
So refactor: introduce global setDraftRequest or something like that

WofWca:
Oh... Yeah, it's worded poorly, I didn't see that. The MR is about NOT skipping the prompt when we should actually show it. This is a bug fix. I'll adjust the commit and MR name.

packages/frontend/src/hooks/chat/useDraft.ts
WofWca: Well, I mean, just because we can. For basically the same reasons as one uses `useCallback` vs a raw function.
We just want to use the latest dependencies, that's it. But things wouldn't be much different if we didn't use `useEffectEvent` I think.

WofWca:
Rebased, added a comment, changed commit message

packages/e2e-tests/tests/composer.spec.ts
WofWca: It's the "quote reply" section:

image

nicodh (approved):

WofWca:
Rebased.

PR #5666, opened 2025-11-03
fix: don't ask "Replace Draft?" unnecessarily

WofWca:
Don't ask if nothing is actually going to get lost.

Edit: I had to change up the tests, because it started selecting chats by the name of the contact in draft.

TODO:
- [x] Adjust tests for the new behavior

### DIFF
--- a/packages/e2e-tests/tests/composer.spec.ts
+++ b/packages/e2e-tests/tests/composer.spec.ts
@@ -11,6 +11,7 @@ import {
   makeDummyContactInviteLink,
   selectChat as selectChatByName,
   sendMessage,
+  getChat,
 } from '../playwright-helper'
 
 test.describe.configure({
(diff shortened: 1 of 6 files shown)


packages/e2e-tests/tests/composer.spec.ts
```diff
@@ -415,34 +416,34 @@ test.describe('draft', () => {
       await replaceDraftDialog.getByRole('button', { name: 'Cancel' }).click()
     }
 
-    const somePriorDraftText =
-      'Draft text before bot command has been clicked' + Math.random()
+    // If have file but no text, it's safe replace text.
```
nicodh:
```suggestion
    // If there is a file but no text, it's safe to set the text.
```

nicodh (approved):
Reviewed by reading & testing

PR #5667, opened 2025-11-03
fix: remove some key codes from keyhandler

### LINKED ISSUE #5595: Dvorak keyoard shortcut errors on mac
Delta chat does not respond to shortcuts correctly on a different keyboards. while using the dvorak-right-handed (DRH) keyboard, copying text and pressing the shortcut key for paste (cmd+v) within a chat on the DRH keyboard - (which corresponds to the keys `cmd+,` on a standard qwerty keyboard) - instead of pasting, delta chat opens the settings popup. There is no keyboard shortcut that pastes copied text on the DRH keyboard. 

- Operating System (Linux/Mac/Windows/iOS/Android): MacOS 15.6.1 
- Delta Chat Version: 2.15.0 (git: v2.15.0)
- Expected behavior: `cmd+v` pastes copied text into chat entry
- Actual behavior: `cmd+v` instead mimics the shortcut `cmd+,`
   - on mac, change keyboard layout to Dvorak Right-handed keyboard
   - copy any text
   - open delta chat, and open a chat 
   - press cmd+v to paste the copied text (on qwerty, this corresponds to `cmd+`,
   - delta chat opens the settings popup, indicating that the app is not tracking keyboard layout change (other apps paste normally using the dvorak-right-handed keyboard)

nicodh:
resolves #5595

This is just a workaround to solve the issue that copy & paste shortcuts (and others) might not be working on different keyboard layouts.

Since copy & paste usage has a much higher priority than maybe having no shortcut to open the settings. And any keyCode can collide with some generic shortcut depending on the keyboard layout on the current machine.

The current approach with only checking the event.key is not a final solution since it might not be possible to use non latin keyboards for shortcuts.

A more proper solution would be to use a library (maybe like https://github.com/mckravchyk/shocut which supports non-latin keyboards it seems) or implement a similar approach ourself

edit: added a simple isLatin check to provide a fallback for non latin keyboards

### DIFF
--- a/packages/frontend/src/keybindings.ts
+++ b/packages/frontend/src/keybindings.ts
@@ -78,6 +78,7 @@ export function keyDownEvent2Action(
   // When modifying this, don't forget to also update the corresponding
   // `aria-keyshortcuts` properties, and the "Keybindings" help window.
   if (!ev.repeat) {
+    // check if key is latin, if not we have to use ev.code instead of ev.key
     const isLatin = ev.key && /^\p{Script=Latin}$/u.test(ev.key)
     // fire only on first press
     if (ev.altKey && ev.code === 'ArrowDown') {
@@ -159,7 +160,7 @@ export function keyDownEvent2Action(
       }
     } else if (
       ((ev.metaKey || ev.ctrlKey) && ev.key === '/') ||
-      (!isLatin && ev.code === 'KeySlash')
+      (!isLatin && ev.code === 'Slash')
     ) {
       return KeybindAction.KeybindingCheatSheet_Open
     }


packages/frontend/src/keybindings.ts
```diff
@@ -78,6 +78,7 @@ export function keyDownEvent2Action(
   // When modifying this, don't forget to also update the corresponding
   // `aria-keyshortcuts` properties, and the "Keybindings" help window.
   if (!ev.repeat) {
+    const isLatin = ev.key && /^\p{Script=Latin}$/u.test(ev.key)
```
WofWca, line 82: How about some comments? At least link to an issue?

packages/frontend/src/keybindings.ts
```diff
@@ -154,7 +157,10 @@ export function keyDownEvent2Action(
       if ((ev.target as HTMLElement)?.id === 'composer-textarea') {
         return KeybindAction.MessageList_PageDown
       }
-    } else if ((ev.metaKey || ev.ctrlKey) && ev.key === '/') {
+    } else if (
+      ((ev.metaKey || ev.ctrlKey) && ev.key === '/') ||
+      (!isLatin && ev.code === 'KeySlash')
```
WofWca: Have you tested this? Looks like this key does not exist.

```suggestion
      (!isLatin && ev.code === 'Slash')
```

packages/frontend/src/keybindings.ts
nicodh: git blame will show the commit and the comment in this PR. The issue is not about non-latin but Dvorak right-handed.

But I will add an explaining comment why we care for isLatin

WofWca: I don't think git history is a good place to explain the semantics of the code. One still properly names variables even if you can look at git blame to find the issue that a commit solves. And the same should go for comments.

nicodh:
> Does anyone else implement such a solution?

It's based on the approach described here https://www.velopen.com/blog/keyboard-shortcuts-on-non-latin-alphabet-layouts/ if the key is non-latin go for the keyCode.

We could use that library but I'm not sure if we need the contexts which the lib introduces. If there is a solid library that covers all cases and is used by many I agree to use it (as said before)

Basically we rely on the event.key (since that is what the user should see when typing the key as text). Only if the key is non latin we go for the code.

As discussed there is no proper solution for all scenarios, so I would prefer this simple approach.

packages/frontend/src/keybindings.ts
nicodh: As said the issue is about Dvorak right-handed to about non-latin keys

PR #5673, opened 2025-11-04
chore: upgrade Playwright

WofWca:
The old one has a vulnerability.

After this, make sure to `cd packages/e2e-tests && pnpm exec playwright install`.

### DIFF
--- a/packages/e2e-tests/package.json
+++ b/packages/e2e-tests/package.json
@@ -4,7 +4,7 @@
   "type": "module",
   "license": "GPL-3.0-or-later",
   "devDependencies": {
-    "@playwright/test": "^1.50.0",
+    "@playwright/test": "^1.56.1",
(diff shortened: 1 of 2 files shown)


WofWca:
Unfortunately now the test is broken, I think it's flaky. The purpose of this `page.reload()` is unknown:

https://github.com/deltachat/deltachat-desktop/blob/19018c958fa38ff83be93ab287f7a3bf90ea9029/packages/e2e-tests/tests/qrcode-tests.spec.ts#L302-L303

The `afterAll` hook doesn't run because it's on this screen. The "new account" has not been deleted at the end of the test.

5fe6bd4fd226838219c8b6f8d04acee5e87a3e2b

Maybe we could spam Escape, but it doesn't always work, sometimes it just leaves an empty screen. And also I'm not sure how this would interact with the `existingProfiles` array.

PR #5675, opened 2025-11-04
fix: some bugs when switching chats

WofWca:
While working on `useDraft`, I found myself trying to implement
complex logic to manage outdated async functions,
deciding when and when not to set state.

However, `key` is just much simpler and better.

We do not rely on any kind of cross-chat state,
so having `key={chat.id}` makes sense.

The performance for switching chats appears to be about the same.
Without `key` / with `key`, for an empty chat:

image
image

Removing the code that supports variable `chatId`
can be done in follow-up MRs / commits.

Edit: this appears to also improve message list rendering performance, i.e. now it's less often that we `fetchMoreTop` immediately after switching chats. And this is also less jumpy visually.


WofWca:
CC @Simon-Laux because you mentioned this, otherwise feel free to ignore.

PR #5668, opened 2025-11-03
test: add Ctrl + Up shortcut tests

WofWca:
TODO:
- [x] They're a bit flaky right now, due to the raciness of the composer. Hopefully that situation will be improved soon, then this can be merged.
  I ran the tests locally with debug stepping, and they passed.
  Update: OK, maybe flakiness has been fixed by a Playwright upgrade. If not, we can try https://github.com/deltachat/deltachat-desktop/pull/5914 or to temporarily disable the tests. So this is good to merge IMO.

PR #5676, opened 2025-11-04
fix: simplify `useDraft` and fix small issues

WofWca:
https://github.com/deltachat/deltachat-desktop/pull/5675 now gives us much more freedom, so we can now significantly simplify `useDraft`.

Sorry about making you watch me add all these lines and then remove them again...

TODO:
- [x] Review and merge https://github.com/deltachat/deltachat-desktop/pull/5675 and https://github.com/deltachat/deltachat-desktop/pull/5663, which this MR is based on.
- [x] Merge the base into main, such that only 8 commits remain.
- [x] Fix test failures?

### DIFF
--- a/packages/frontend/src/hooks/chat/useDraft.ts
+++ b/packages/frontend/src/hooks/chat/useDraft.ts
@@ -228,7 +228,7 @@ export function useDraft(
           debounce(saveAndRefetchDraft, 15_000),
     [saveAndRefetchDraft]
   )
-  // Flush the draft to backend when switching chats.
+  // Flush the draft to backend when ~~switching chats~~ unmounting.
   // Note that specifying `chatId` as a dependency is not necessary,
   // because `debouncedSaveAndRefetchDraft` itself already depends on it.
   useEffect(() => {


WofWca:
OK, again as in https://github.com/deltachat/deltachat-desktop/pull/5668, the flakiness seems to have been resolved, which is IIRC was one of the main reasons I haven't marked this as ready. Tests are also passing locally.
Now it should be good.

packages/frontend/src/hooks/chat/useDraft.ts
```diff
@@ -108,49 +112,26 @@ export function useDraft(
...
+  const skipLoadingDraft = chatId === null
+  const draftIsLoading = skipLoadingDraft ? false : draftIsLoading_
+  const effectRanRef = useRef(false)
   useEffect(() => {
-    // The fact that `abortController` updated means that its dependencies,
-    // i.e. `[accountId, chatId]`, updated.
-    return () => {
-      // This is now the "old" `abortController`. Let's abort it.
-      abortController.abort()
+    if (effectRanRef.current) {
```
nicodh, line 120: So the only reason to have effectRanRef is to log a warning line? Why do wee need that?

WofWca: Well, for the same reason that we need regular logging and TypeScript.
Also, this makes it clear for the reader that we only expect this effect to run once.

nicodh: For the reader it causes more confusion to see a reference that has no real usage imo

WofWca: Well IDK, it's a tradeoff I guess.
If you insist, I'll remove it.

nicodh: I don't insist :-)

nicodh (approved):
Let's merge this afther the upcoming release

PR #5677, opened 2025-11-04
fix: not being able to type question mark

WofWca:
The bug was introduced in
f3630f587d4b109e0144d9917fee5c7235687cc9
(https://github.com/deltachat/deltachat-desktop/pull/5667).


WofWca:
The test failure seems unrelated, I will investigate.

PR #5680, opened 2025-11-05
feat: add multitransport

### LINKED ISSUE #5648: list, remove and add transports
- [ ] add a `ListTransportDialog` similar to the proxy view that allows removing transport and selecting default transport  
this dialog is opened where "Password and Accounts" dialog were opened before - each item in the list opens another "Password and Accounts" dialog then
- [ ] adding transport, for now, should be done by scanning an DCACCOUNT:-QR code from the normal scanner (instead of adding a new profile)

cc @link2xt does that make sense for the current state of multi-transport? which PR should be checked out for testing? otherwise, feel free to edit this issue directly :)

counterpart of https://github.com/deltachat/deltachat-android/issues/3970

nicodh:
resolves #5648

Needs local core with https://github.com/chatmail/core/pull/7348 merged ✅

~~Be aware not to use with existings accounts see https://github.com/deltachat/deltachat-desktop/issues/5648#issuecomment-3474884528~~

### DIFF
--- a/packages/frontend/src/components/dialogs/Transports/index.tsx
+++ b/packages/frontend/src/components/dialogs/Transports/index.tsx
@@ -141,6 +141,7 @@ export default function TransportsDialog(
   return (
     <Dialog
       fixed
+      onClose={onClose}
       width={500}
       dataTestid='transports-dialog'
       canOutsideClickClose={true}
@@ -157,9 +158,7 @@ export default function TransportsDialog(
               <div className={styles.transportRow} key={transport.addr}>
                 <div
                   onClick={() => changeDefaultTransport(transport)}
-                  className={classNames(styles.transportItem, {
-                    [styles.transportLabel]: transport.isDefault,
-                  })}
+                  className={styles.transportItem}
                 >
                   <span className={styles.transportRadioButton}>
                     <input
@@ -173,10 +172,10 @@ export default function TransportsDialog(
                       readOnly
                     />
                   </span>
-                  <span id={`transport-label-${index}`}>
+                  <label id={`transport-label-${index}`}>
                     {transport.addr}{' '}
                     {transport.isDefault ? ` (${tx('def')})` : ''}
-                  </span>
+                  </label>
                 </div>
                 <div>
                   <Button
--- a/packages/frontend/src/components/dialogs/Transports/styles.module.scss
+++ b/packages/frontend/src/components/dialogs/Transports/styles.module.scss
@@ -1,17 +1,9 @@
-.transportList {
-  flex-grow: 1;
-  overflow-y: auto;
-}
-
 .container {
-  margin: 0 10px 0 30px;
+  margin-inline-start: 30px;
+  margin-inline-end: 10px;
   display: flex;
   flex-direction: column;
   flex: 1 1 auto;
-  h3 {
-    margin: 0;
-    margin-top: 10px;
-  }
 }
 
 .transportRow {
@@ -22,7 +14,7 @@
   &:hover {
     background-color: var(--globalHoverBg);
   }
-  div {
+  & > div {
     display: flex;
   }
   .transportRadioButton {
@@ -36,11 +28,6 @@
   .transportItem {
     cursor: pointer;
   }
-}
-
-.explain {
-  margin-inline-end: 20px;
-  margin-top: 10px;
 }
 
 $iconSize: 1.3em;
@@ -71,11 +58,4 @@ $iconSize: 1.3em;
 
 .addTransportButton {
   padding: 10px;
-}
-
-@media (max-height: 500px) {
-  .transportList {
-    flex-grow: 0;
-    overflow-y: visible;
-  }
 }


packages/frontend/src/components/dialogs/Transports/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+  const openConfirmationDialog = useConfirmationDialog()
+
+  const changeDefaultTransport = useCallback(
+    async (transport: EnteredLoginParam) => {
+      await BackendRemote.rpc.setConfig(
+        accountId,
+        'configured_addr',
+        transport.addr
+      )
+      await BackendRemote.rpc.stopIo(accountId)
```
link2xt: These manual `stopIo` and `startIo` calls should not be needed anymore, the core now does this when `configured_addr` is set.

nicodh: Just pulled the latest state of the core multitransport PR. When I now add a new transport I see a lot of events in the logs and they don't stop any more:

121.6s [i]core/event: 2 src/imap.rs:778: 0 mails read from "INBOX".
121.6s [i]core/event: 2 src/imap/select_folder.rs:248: uid/validity change folder INBOX: new 1/1762373735 previous 3/1762373687.
121.6s [i]core/event: 2 src/imap/select_folder.rs:248: uid/validity change folder INBOX: new 3/1762373687 previous 1/1762373735.
121.6s [D]core/event: 2 ConnectivityChanged { kind: 'ConnectivityChanged' }
121.6s [D]core/event: 2 ImapInboxIdle { kind: 'ImapInboxIdle' }
121.6s [i]core/event: 2 src/scheduler.rs:685: IMAP session in folder "INBOX" supports IDLE, using it.
121.6s [i]core/event: 2 src/imap/select_folder.rs:248: uid/validity change folder INBOX: new 1/1762373735 previous 3/1762373687.
121.6s [i]core/event: 2 src/imap/idle.rs:39: Skipping IDLE in "INBOX" because there may be new mail.
121.6s [i]core/event: 2 src/scheduler.rs:425: IMAP loop iteration for inbox finished, keeping the session.
121.6s [i]core/event: 2 src/imap.rs:881: resync_folder_uids: Collected 0 message IDs in INBOX.
121.6s [i]core/event: 2 src/imap/select_folder.rs:248: uid/validity change folder INBOX: new 3/1762373687 previous 1/1762373735.
121.6s [i]core/event: 2 src/imap.rs:585: fetch_new_msg_batch(INBOX): UIDVALIDITY=1762373687, UIDNEXT=3.
121.6s [i]core/event: 2 src/imap/select_folder.rs:248: uid/validity change folder INBOX: new 1/1762373735 previous 3/1762373687.
121.6s [i]core/event: 2 src/imap.rs:778: 0 mails read from "INBOX".
121.6s [i]core/event: 2 src/imap/select_folder.rs:248: uid/validity change folder INBOX: new 3/1762373687 previous 1/1762373735.
121.6s [i]core/event: 2 src/imap.rs:881: resync_folder_uids: Collected 0 message IDs in INBOX.
121.6s [i]core/event: 2 src/imap/select_folder.rs:248: uid/validity change folder INBOX: new 1/1762373735 previous 3/1762373687.
121.6s [i]core/event: 2 src/imap.rs:585: fetch_new_msg_batch(INBOX): UIDVALIDITY=1762373735, UIDNEXT=1.

link2xt: It is broken by the latest WIP commit https://github.com/chatmail/core/pull/7348/commits/37a6304e928972507f533753c629f632b566e8b2, the branch is not usable at the moment.

But once it is working stop_io and start_io will not be necessary.

nicodh: > These manual `stopIo` and `startIo` calls should not be needed anymore, the core now does this when `configured_addr` is set.

I just tested with the latest state of https://github.com/chatmail/core/pull/7348 . Adding and switching transports works again now.  But without stopIo & startIo in https://github.com/deltachat/deltachat-desktop/blob/add-multitransport/packages/frontend/src/components/dialogs/Transports/index.tsx#L64 I have to switch to another account and back to get a new transport setting effective.

link2xt: I added commit https://github.com/chatmail/core/pull/7348/commits/14cc5977192b491613cddb819448dc843872ab0b that removes `configured_addr` from cache.

`start_io` also clears cache for all config values, this is why with start_io it worked.

nicodh: Ok, now it switches the used transport immediately.

One minor thing is still unexpected: when calling BackendRemote.rpc.getConfig(accountId, 'configured_addr') AFTER the transport switch it still returns the previous transport address although it sends messages using the new address.

After switching to another account getConfig shows the correct new address.
So it seems there is still another cache involved.

Can be tested with this PR in desktop by opening the EditAccountAndPassword dialog after changing a transport setting.

packages/frontend/src/components/dialogs/Transports/index.tsx
```diff
@@ -0,0 +1,203 @@
...
+        </div>
+      </DialogBody>
+      <DialogFooter>
+        <Button
+          className={styles.addTransportButton}
+          onClick={() => openQrScanner()}
+          aria-label={tx('add_transport')}
+          title={tx('add_transport')}
+        >
+          ＋
```
link2xt: This is probably some "+" sign, but for me it does not display:
Image

I can only search for it, it does not display there either: https://unicodeplus.com/U+FF0B

packages/frontend/src/components/dialogs/Transports/index.tsx
link2xt: I fixed caching with https://github.com/deltachat/deltachat-desktop/pull/5680/commits/51ac9575a7a195ea9f1122527536762f6c2d5245

packages/frontend/src/components/dialogs/Transports/index.tsx
nicodh: I replaced it with an icon

nicodh:
I also added the label (Default)  after the default transport in the list of transport

nicodh:
e2e tests still need to be updated (since EditAccountAndPassword is not directly accessible in Advanced settings anymore

link2xt:
Multi-transport is currently not supported together with multi-device, so maybe display EditAccountAndPassword link for bcc_self=0 and transport list button for bcc_self=1?

We will add multi-device support, but users anyway have to update all devices first, so for the next release adding second transport with multi-device should not be possible.

nicodh:
@r10s there are some UI questions around the  fact that adding transports should not be possible when bcc_self/Multi-Device is enabled. Should we hide the complete Transport Settings dialog and just show the EditAccountAndPassword or show the dialog but hide the Add button or show it and then show a warning adding a transport will disable Multi-Device mode?

r10s:
> Multi-transport is currently not supported together with multi-device, so maybe display EditAccountAndPassword link for bcc_self=0 and transport list button for bcc_self=1?

@link2xt i am wondering if core could not return an error, similar to https://github.com/chatmail/core/issues/7400 - then not all UI need to special case that - and all remove special casing in some months

packages/frontend/src/components/dialogs/BasicScanner.tsx
```diff
@@ -57,7 +46,7 @@ export default function ProxyQrScanner({
   }, [])
 
   return (
-    <Dialog onClose={onClose} dataTestid='proxy-qrscan-dialog'>
+    <Dialog onClose={onClose} dataTestid='transport-qrscan-dialog'>
```
WofWca: This dialog is now used for both proxy and transport, thus the ID name seems inappropriate?

packages/frontend/src/components/dialogs/BasicScanner.tsx
```diff
@@ -29,25 +25,18 @@ export default function ProxyQrScanner({
...
-        const { qr } = await processQr(accountId, data)
-        if (qr.kind === 'proxy') {
-          onSuccess(data)
-          onClose()
-        } else {
-          handleError(new Error(tx('proxy_invalid')))
-        }
-      }
+      onSuccess(data)
+      onClose()
```
WofWca, line 37: We probably shouldn't close the dialog here then, and instead only close it if the QR type is appropriate (which would have to be done by the user of this component).

packages/frontend/src/components/dialogs/Transports/styles.module.scss
```diff
@@ -0,0 +1,78 @@
+.transportList {
+  flex-grow: 1;
+  overflow-y: scroll;
```
WofWca:
```suggestion
  overflow-y: auto;
```

makes the scrollbar only appear when it's needed.

WofWca:
Is the MR's description still true? Isn't the MR already in Core 2.28? And is it still unsafe to test with existing accounts?

packages/frontend/src/components/Settings/Advanced.tsx
```diff
@@ -104,6 +100,16 @@ export default function Advanced({ settingsStore }: Props) {
           <ImapFolderHandling settingsStore={settingsStore} />
         </>
       )}
+      <SettingsButton
+        onClick={() => {
+          openTransportSettings()
+        }}
```
WofWca:
```suggestion
        onClick={openTransportSettings}
```

packages/frontend/src/components/Settings/Advanced.tsx
```diff
@@ -104,6 +100,16 @@ export default function Advanced({ settingsStore }: Props) {
...
+        dataTestid='open-transport-settings'
+      >
+        {tx('transport_settings')}
+      </SettingsButton>
+
+      <ImapFolderHandling settingsStore={settingsStore} />
```
WofWca: Are we supposed to show two `ImapFolderHandling` components at the same time? The other is a few lines above.

Image

packages/frontend/src/components/dialogs/Transports/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+      <DialogBody className={styles.transportDialogBody}>
+        <div className={styles.container}>
+          <div className={styles.transportList}>
+            {transports.map((transport, index) => (
+              <div className={styles.transportRow} key={index}>
+                <div
+                  onClick={() => changeDefaultTransport(transport)}
+                  className={classNames(styles.transportItem, {
+                    [styles.transportLabel]: transport.isDefault,
+                  })}
```
WofWca:
```suggestion
                  className={styles.transportItem}
```

`transportLabel` is unused.

Also `styles.radioButton` and `styles.transportDialogBody`.

packages/frontend/src/components/dialogs/Transports/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+                >
+                  <span className={styles.transportRadioButton}>
+                    <input
+                      id={`proxy-${index}`}
+                      name='proxy-selection'
```
WofWca: Not that it affects behavior, but this is not `proxy`.

packages/frontend/src/components/dialogs/Transports/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+                      type='radio'
+                      value={transport.addr}
+                      checked={transport.isDefault}
+                      className={styles.radioButton}
+                      aria-labelledby={`transport-label-${index}`}
```
WofWca, line 171: There is no element with such ID.

packages/frontend/src/components/dialogs/Transports/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+              <div className={styles.transportRow} key={index}>
+                <div
+                  onClick={() => changeDefaultTransport(transport)}
+                  className={classNames(styles.transportItem, {
+                    [styles.transportLabel]: transport.isDefault,
+                  })}
+                >
+                  <span className={styles.transportRadioButton}>
+                    <input
+                      id={`proxy-${index}`}
```
WofWca: This ID is unused, right? Probably, because it's also unused in `ProxyItemRow` where this has been copied from.

packages/frontend/src/components/dialogs/Transports/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+                          styles.trash
+                        )}
+                        aria-hidden='true'
+                      />
+                    </Button>
+                  )}
+                </div>
+              </div>
+            ))}
+          </div>
```
WofWca, line 218: Accessibility seems to suck, unfortunately, and, as I just realized, this seems to apply to the proxy dialog as well. Maybe what we need is to have a separate `<section aria-label=...>` for each transport item. Otherwise you can't be sure which transport the "edit" and "delete" buttons belong to.

Maybe there is an existing pattern for this case.

I could address this separately.

packages/frontend/src/components/dialogs/Transports/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+      <DialogHeader
+        title={tx('transports')}
+        onClose={onClose}
+        dataTestid='transports-settings'
+      />
+      <DialogBody className={styles.transportDialogBody}>
+        <div className={styles.container}>
+          <div className={styles.transportList}>
+            {transports.map((transport, index) => (
+              <div className={styles.transportRow} key={index}>
```
WofWca: Specifying `key={index}` is pointless because it's done by React by default. A better key is needed.

I'm not sure if `address` is enough, I suspect that one can have multiple transports with the same address but with different IMAP connection settings and stuff.

packages/frontend/src/components/dialogs/Transports/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+        message:
+          'It is not possible to add transports when multi-device mode is enabled.',
+      })
+      return
+    }
+    openDialog(BasicQrScanner, {
+      onSuccess: async (result: string) => {
+        const { qr } = await processQr(accountId, result)
+        if (qr.kind === 'account') {
+          await BackendRemote.rpc.addTransportFromQr(accountId, result)
```
WofWca, line 96: It seems that this can take a while, so maybe we should have a "loading" indicator.

packages/frontend/src/components/dialogs/Transports/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+import { processQr } from '../../../backend/qr'
+import Icon from '../../Icon'
+
+/**
+ * Dialog for transports configuration
+ */
+export default function TransportsDialog(
+  props: DialogProps & {
+    accountId: number
+    configured: boolean
```
WofWca: This parameter is unused.

packages/frontend/src/components/dialogs/Transports/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+    newTransport?: string
+  }
+) {
+  const tx = useTranslationFunction()
+  const openAlertDialog = useAlertDialog()
+  const { accountId, onClose } = props
+
+  // used in  new transport form
```
WofWca, line 34: What does this comment mean?

packages/frontend/src/components/dialogs/Transports/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+    async (transport: EnteredLoginParam) => {
+      await BackendRemote.rpc.setConfig(
+        accountId,
+        'configured_addr',
+        transport.addr
+      )
+      SettingsStoreInstance.effect.setCoreSetting(
+        'configured_addr',
+        transport.addr
+      )
```
WofWca: This does `setConfig()` twice, the second one inside of `setCoreSetting`. Do we need to do it twice?

packages/frontend/src/components/dialogs/Transports/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+    },
+    [openDialog]
+  )
+
+  return (
+    <Dialog
+      fixed
+      width={500}
+      dataTestid='transports-dialog'
+      canOutsideClickClose={false}
```
WofWca:
```suggestion
      canOutsideClickClose={true}
```

There is no possible "unsaved progress" so I think it's fine to close it on outside click.

packages/frontend/src/components/dialogs/Transports/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+        accountId,
+        'configured_addr',
+        transport.addr
+      )
+      SettingsStoreInstance.effect.setCoreSetting(
+        'configured_addr',
+        transport.addr
+      )
+
+      setTransports(prev =>
```
WofWca: The way that `setTransports` is currently used, I don't see a great benefit in it. The benefit could be that the change is instantly visible to the user, before `setConfig` even finishes, but this benefit is not there, because we `await rpc.setConfig()` before `setTransports`.

I suggest to either call `setTransports` immediately, or to utilize `useRpcFetch().refresh()` instead. Or maybe even have a combination, something like `setOptimisticTransports`.

nicodh:
> adding transport, for now, should be done by scanning an DCACCOUNT:-QR code from the normal scanner (instead of adding a new profile)

I will add that in another PR

packages/frontend/src/components/dialogs/Transports/index.tsx
nicodh: Can/should be done in a followup PR

packages/frontend/src/components/dialogs/Transports/index.tsx
nicodh: I don't think so. At least right now as we have configured_address holding only the address as identifier...

WofWca: You're right. OK, `address` it is then, let's hope that this doesn't break at some point.

packages/frontend/src/components/dialogs/BasicScanner.tsx
nicodh: Sounds reasonable but can/should go into a followup PR

packages/frontend/src/components/Settings/Advanced.tsx
```diff
@@ -74,17 +80,6 @@ export default function Advanced({ settingsStore }: Props) {
...
-        onClick={() => {
-          openDialog(EditAccountAndPasswordDialog, {
-            settingsStore,
-          })
-        }}
-        dataTestid='open-account-and-password'
-      >
-        {tx('pref_password_and_account_settings')}
-      </SettingsButton>
       <SettingsButton
```
WofWca, line 84:
```suggestion
      <SettingsHeading>{tx('pref_server')}</SettingsHeading>
      <SettingsButton
```

WofWca (commented):
I don't think that the "server" heading is supposed to be removed?

packages/frontend/src/components/dialogs/Transports/index.tsx
```diff
@@ -0,0 +1,234 @@
...
+      openDialog(EditAccountAndPasswordDialog, {
+        addr: transport.addr,
+      })
+    },
+    [openDialog]
+  )
+
+  return (
+    <Dialog
+      fixed
```
WofWca, line 143:
```suggestion
    <Dialog
      onClose={onClose}
      fixed
```

Otherwise the dialog doesn't properly close with outside click, resulting it re-appearing when re-opening the "Settings" dialog.

packages/frontend/src/components/dialogs/Transports/styles.module.scss
```diff
@@ -0,0 +1,81 @@
...
+.addTransportButton {
+  padding: 10px;
+}
+
+@media (max-height: 500px) {
+  .transportList {
+    flex-grow: 0;
+    overflow-y: visible;
+  }
+}
```
WofWca: What is this supposed to do? It doesn't seem to affect behavior. Same for the other `.transportList` selector.

packages/frontend/src/components/dialogs/Transports/styles.module.scss
```diff
@@ -0,0 +1,81 @@
...
+  .transportItem {
+    cursor: pointer;
+  }
+}
+
+.explain {
+  margin-inline-end: 20px;
+  margin-top: 10px;
+}
+
```
WofWca:
```suggestion
```

Unused.

packages/frontend/src/components/dialogs/Transports/styles.module.scss
```diff
@@ -0,0 +1,81 @@
...
+  display: flex;
+  border-bottom: 1px solid var(--separatorColor);
+  padding: 10px 0 10px 10px;
+  justify-content: space-between;
+  &:hover {
+    background-color: var(--globalHoverBg);
+  }
+  div {
+    display: flex;
+  }
```
WofWca: This affects all `div` children of `transportRow`.

packages/frontend/src/components/dialogs/Transports/styles.module.scss
```diff
@@ -0,0 +1,81 @@
...
+
+.container {
+  margin: 0 10px 0 30px;
+  display: flex;
+  flex-direction: column;
+  flex: 1 1 auto;
+  h3 {
+    margin: 0;
+    margin-top: 10px;
+  }
```
WofWca:
```suggestion
```

Apparently unused.

packages/frontend/src/components/dialogs/Transports/index.tsx
```diff
@@ -0,0 +1,234 @@
...
+                  <span className={styles.transportRadioButton}>
+                    <input
+                      id={`transport-${index}`}
+                      name='transport-selection'
+                      type='radio'
+                      value={transport.addr}
+                      checked={transport.isDefault}
+                      className={styles.radioButton}
+                      aria-labelledby={`transport-label-${index}`}
+                      readOnly
```
WofWca, line 172: Another accessibility issue. This will break eventually. The `span` should become a label, and we should listen to click events on the `<input>` itself.

nicodh:
> I don't think that the "server" heading is supposed to be removed?

Idk, I think it will change. We don't mention "server" but only transports now. But the section should have a headline so I readded it for now

packages/frontend/src/components/dialogs/ConfigureProgressDialog.tsx
```diff
@@ -93,17 +93,11 @@ export function ConfigureProgressDialog({
...
-              } else if (existingTransport.addr !== configuration.addr) {
-                // multiple transports are not supported yet
-                throw new Error(
-                  'Multi transport is not supported right now. Check back in a few months!'
-                )
-              }
+            if (
+              existingTransports.length === 1 &&
+              existingTransports[0].addr === ''
+            ) {
```
WofWca, line 99: When do we ever expect this? I tried "manual log in" and "instant onboarding" and I have not managed to trigger this condition. In the former case there are 0 transports, in the latter the "qr" path is taken.

Thankfully this only affects whether we'll select the "device messages" chat after login.

packages/frontend/src/components/dialogs/Transports/styles.module.scss
```diff
@@ -0,0 +1,81 @@
+.transportList {
+  flex-grow: 1;
+  overflow-y: auto;
+}
+
+.container {
+  margin: 0 10px 0 30px;
```
WofWca: This is not gonna work as intended together with "add transport" button in RTL mode.

Image

WofWca (approved):
I tested it again, it seems to work alright.

As long as the remaining issues (e.g. accessibility) will not be forgotten and addressed in follow-ups, let's merge.

ralphtheninja:
Congratulations! This is a milestone!

PR #5681, opened 2025-11-05
fix: attaching file sometimes resetting text

WofWca:
See the new comment on `addFileToDraft`.

The issue is easy to reproduce with

```diff
diff --git a/packages/frontend/src/components/composer/Composer.tsx b/packages/frontend/src/components/composer/Composer.tsx
index https://github.com/deltachat/deltachat-desktop/commit/bd995140f62aebc77d8a9b529acb4f9e26bb56e6..13d790cb8 100644
--- a/packages/frontend/src/components/composer/Composer.tsx
+++ b/packages/frontend/src/components/composer/Composer.tsx
@@ -525,6 +525,7 @@ const Composer = forwardRef<
             fileName,
             file_content.split(';base64,')[1]
           )
+          await new Promise(r => setTimeout(r, 2000))
           await addFileToDraft(path, fileName, msgType)
           // delete file again after it was sucessfuly added
           await runtime.removeTempFile(path)
```

1. Ctrl + C a file from your OS.
2. Ctrl + V it in the composer.
3. Type something before the file shows up.

This was an unfortunate side-effect of the recent migration
from `useRef` to `useState` in `useDraft`.
62d796b89ab7b5d3174444098f444c12287755eb
(https://github.com/deltachat/deltachat-desktop/pull/5642).

This also moves `handleSetDraftRequestEffectEvent` to inside
`useEffect`, otherwise it starts warning
about refs access during render.
I don't think this will affect behavior.

The warning suppressed by
`eslint-disable-next-line react-hooks/set-state-in-effect`
stopped existing, I am not sure why. Probably a bug in the linter.
So I removed the "ignore" line.

TODO:
- [x] Merge the base, up to the last commit

### DIFF
--- a/packages/frontend/src/hooks/chat/useDraft.ts
+++ b/packages/frontend/src/hooks/chat/useDraft.ts
@@ -99,17 +99,34 @@ export function useDraft(
     inputRef.current?.focus()
   })
 
-  const [
-    draftState,
-    /**
-     * This will not save the draft to the backend.
-     */
-    setDraftState,
-  ] = useState<DraftObject>(() => emptyDraft())
+  const [draftState, _setDraftState] = useState<DraftObject>(() => emptyDraft())
+  const draftStateRef = useRef(draftState)
+  /**
+   * This is like a regular React's `setState` with a function argument,
+   * but it applies the function synchronously, and returns the new state.
+   *
+   * This will not save the draft to the backend.
+   */
+  // We need this workaround to be able to call `miscSetDraft` without waiting
+  // for a re-render to see the final `draftState`.
+  // Because the next re-render might just not ever happen,
+  // e.g. because the composer got unmounted / re-created with another key.
+  //
+  // Otherwise a `useEffect` would probably be enough for `rpc.miscSetDraft()`.
+  const setDraftState = useCallback(
+    (
+      draftSetter: (prevDraftState: DraftObject) => DraftObject
+    ): DraftObject => {
+      draftStateRef.current = draftSetter(draftStateRef.current)
+      _setDraftState(draftStateRef.current)
+      return draftStateRef.current
+    },
+    []
+  )
 
   const clearDraftState = useCallback(() => {
-    setDraftState(emptyDraft())
-  }, [])
+    setDraftState(() => emptyDraft())
+  }, [setDraftState])
 
   const [draftIsLoading_, setDraftIsLoading] = useState(true)
   const skipLoadingDraft = chatId === null
@@ -126,7 +143,6 @@ export function useDraft(
     if (skipLoadingDraft) {
       return
     }
-    // eslint-disable-next-line react-hooks/set-state-in-effect
     setDraftIsLoading(true)
     BackendRemote.rpc
       .getDraft(accountId, chatId)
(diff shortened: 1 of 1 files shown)


WofWca:
I'm thinking that instead of this we might want to ensure to never call the `useDraft`'s callbacks asynchronously, like I mentioned in https://github.com/deltachat/deltachat-desktop/issues/5993#issue-3870919908.
To achieve that we can instead `setState` immediately, but pass a promise to inside of it, that resolves to the file path or whatever.

WofWca:
TL;DR I think this MR is good to merge. Below I consider the alternative approaches.

> but pass a promise to inside of it

I have pondered this concept for a while, and came to the conclusion that this approach is not always the best one. One example: app picker. There we first download the app and then add it to the draft. The "loading" state here, I argue, clearly belongs to the app picker and not the composer. It would be weird if we added this `Promise<string>` to the draft but then, upon encountering a download error, told the user that "setting the draft failed because of a network error". Rather the app picker should handle the network.

But even in such a situation it's still possible to call the latest `addFileToDraft` and not the one captured by an async function. One way to do it is something like this:

```typescript
  const [filePromise, setFilePromise] = useState<Promise<string> | null>(null)
  const addFileToDraft2 = useEffectEvent(addFileToDraft)
  useEffect(() => {
    if (filePromise == null) {
      return
    }
    let outdated = false
    filePromise.then(path => {
      if (outdated) {
        return
      }
      addFileToDraft2, fileName, viewType)
    })
    return () => {
      outdated = true
    }
  }, [filePromise])
```

This code can even be in `useDraft`, but then there is a weird situation where we have both `draftState.file` and this pending `Promise`, i.e. basically "two files at the same time". So maybe it's better to have this code somewhere else.

I also considered doing `rpc.miscSetDraft` inside of `useEffect` whose dependency is `draftState`, but that would have a downside (or is it an upside??) which is: if you switch chats while there is a pending promise, then when it finally resolves and calls `addFileToDraft`, the `useEffect` will not run (because, again, the component is already unmounted).
Since this would change behavior, I did not consider it too thoroughly.

I also worked on splitting up the `file` part of the draft and the `text` part of it, so that setting a `file` would not affect the text, but there the problem was that `rpc.miscSetDraft` still needs the full draft state.

nicodh (approved):
I see that this PR works and fixes the problem but I couldn't reproduce the issue it fixes without "faking" a delay after pasting a file to have time to type.

If we remove some of the references in the comment it is not adding extra lines so fine to merge imo

packages/frontend/src/hooks/chat/useDraft.ts
```diff
@@ -99,17 +99,39 @@ export function useDraft(
...
+   */
+  // We need this workaround to be able to call `miscSetDraft` without waiting
+  // for a re-render to see the final `draftState`.
+  // Because the next re-render might just not ever happen,
+  // e.g. because the composer got unmounted / re-created with another key.
+  //
+  // Otherwise a `useEffect` would probably be enough for `rpc.miscSetDraft()`.
+  // See
+  // - https://react.dev/learn/queueing-a-series-of-state-updates
+  // - https://medium.com/geographit/accessing-react-state-in-event-listeners-with-usestate-and-useref-hooks-8cceee73c559
```
nicodh: I would move the references to the PR description or the commit message but not leave them in the comment (except the react.dev source maye)

WofWca: I also wasn't too sure if those are much useful. Removed all the links.

PR #5686, opened 2025-11-06
feat: new broadcast channel behaviour & core update 2.25

### LINKED ISSUE #5669: Adapt to the new channels
When updating core to v2.24.0, which as https://github.com/chatmail/core/pull/7268 merged, some adaptions are needed:

- [ ] There is a new QR type `DC_QR_ASK_JOIN_BROADCAST = 204` (CFFI) / `AskJoinBroadcast` (JsonRPC)
  - It should be handled just as `DC_QR_ASK_VERIFYGROUP` and `DC_QR_ASK_VERIFYCONTACT`, just that the dialog shown after scanning uses the translated string `qrscan_ask_join_channel`
- [ ] Cloning an outgoing broadcast channel should be impossible
- [ ] In the profile view of broadcast channels, `Add member` should be hidden, and instead `QR Invite Code` should be shown
- [ ] While creating a broadcast channel:
  - the "Add Members"/"Add Recipients" option should be hidden.
  - The translated string `chat_new_channel_hint` should be shown.
- [ ] UI-testing using the list at https://github.com/chatmail/core/issues/6884

Android PR for reference: https://github.com/deltachat/deltachat-android/pull/3886

nicodh:
Implements new broadcast channel creation & invites

resolves #5669

Best reviewed commit by commit

#public-preview (for testing)

The failing e2e test seems to be related to the changes

Preview is available here:
https://download.delta.chat/desktop/preview/

### DIFF
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -84,6 +84,6 @@
     "message": "Do you want to join the channel \"%1$s\"?"
   },
   "instant_onboarding_confirm_channel": {
-    "message": "Create a profile to join channel %1$s."
+    "message": "Do you want to create a new profile and join the channel \"%1$s\"?"
   }
 }
--- a/packages/frontend/src/backend/qr.ts
+++ b/packages/frontend/src/backend/qr.ts
@@ -8,7 +8,7 @@ export type VerifyContactQr = Extract<T.Qr, { kind: 'askVerifyContact' }>
 
 export type VerifyGroupQr = Extract<T.Qr, { kind: 'askVerifyGroup' }>
 
-export type VerifyChannelQr = Extract<T.Qr, { kind: 'askJoinBroadcast' }>
+export type JoinBroadcastQr = Extract<T.Qr, { kind: 'askJoinBroadcast' }>
 
 export type LoginQr = Extract<T.Qr, { kind: 'login' }>
 
--- a/packages/frontend/src/components/chat/ChatListContextMenu.tsx
+++ b/packages/frontend/src/components/chat/ChatListContextMenu.tsx
@@ -232,9 +232,8 @@ export function useChatListContextMenu(): {
             label: tx('edit_channel'),
             action: () => onViewGroup(singleChat.id),
           },
-        // Clone Group
+        // Clone Group (note that broadcasts should not be cloned!)
         singleChat &&
-          !isOutBroadcast &&
           isGroup && {
             label: tx('clone_chat'),
             action: () => {
--- a/packages/frontend/src/contexts/InstantOnboardingContext.tsx
+++ b/packages/frontend/src/contexts/InstantOnboardingContext.tsx
@@ -5,7 +5,7 @@ import type {
   AccountQr,
   LoginQr,
   QrWithUrl,
-  VerifyChannelQr,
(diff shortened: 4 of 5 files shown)


packages/frontend/src/components/chat/ChatListContextMenu.tsx
```diff
@@ -234,6 +234,7 @@ export function useChatListContextMenu(): {
           },
         // Clone Group
         singleChat &&
+          !isOutBroadcast &&
```
WofWca: This does nothing functionally, and I'd say also doesn't make sense semantically, because `!isOutBroadcast` is implied by `isGroup`.

packages/frontend/src/backend/qr.ts
```diff
@@ -8,6 +8,8 @@ export type VerifyContactQr = Extract<T.Qr, { kind: 'askVerifyContact' }>
 
 export type VerifyGroupQr = Extract<T.Qr, { kind: 'askVerifyGroup' }>
 
+export type VerifyChannelQr = Extract<T.Qr, { kind: 'askJoinBroadcast' }>
```
WofWca: Why is it named this way? This neither reflects the core's name nor the meaning, apparently?

_locales/_untranslated_en.json
```diff
@@ -79,5 +79,11 @@
   },
   "accounts_list_item_menu_label": {
     "message": "Actions with Profile"
+  },
+  "qrscan_ask_join_channel": {
+    "message": "Do you want to join the channel \"%1$s\"?"
+  },
+  "instant_onboarding_confirm_channel": {
+    "message": "Create a profile to join channel %1$s."
```
WofWca:
```suggestion
    "message": "Do you want to create a new profile and join the channel \"%1$s\"."
```

In line with

https://github.com/deltachat/deltachat-desktop/blob/63c2c938dd86976e7d3e94e19bd1243475fc70a9/_locales/en.xml#L635-L637

But it doesn't matter much, we'll pull the translations from upstream anyway.

WofWca (commented):
I tested instant onboarding with this MR, it works. Also tested joining with existing profile. Works.
So all the commits up to the `refactor: deduplicate...` one seem good and IMO can be merged.
I'll come back with a review of the refactor commit.

we'll probably want to re-check the implementation in with Android.

packages/frontend/src/components/chat/ChatListContextMenu.tsx
nicodh: I would say it is "explicitly" saying that broadcast channel should not be cloned. 
But maybe that info should go to a comment then?

WofWca: IDK, It already clearly says that this button is only available for groups.

But it's not a big deal.

packages/frontend/src/backend/qr.ts
nicodh: Ok, the naming is inconsistent but in several scopes:
Channel <=> Broadcast is not consistent: the translations use only channel, the core uses broadcast
 askVerifyGroup could also be named askJoinGroup
 
 But I agree that it might be more consistent here to stick to the kind type

PR #5688, opened 2025-11-07
Update classic email login wordings

### LINKED ISSUE #5687: update strings and some details in "Classic E-Mail Login"
- for the title, use `manual_account_setup_option` instead of  `login_explain`
- for the info below the password field, use `login_advanced_hint` instead of `login_no_servers_hint`. make sure, linebreaks are rendered correctly
- remove the hint `login_subheader` at the very bottom
- for the "expand button", use `menu_more_options` instead of `menu_advanced`
- use the placeholders `automatic` for all expaneded (now "more") options - do not use  `default_value_as_above` or `def`
- check that the strings `login_explain`, `login_header`, `login_subheader`, `login_no_servers_hint` are no longer used elsewhere, they will be removed soon

counterpart of https://github.com/deltachat/deltachat-ios/pull/2874 (has screenshots) and https://github.com/deltachat/deltachat-android/pull/3996

nicodh:
resolves #5687

### DIFF
--- a/packages/frontend/src/components/LoginForm.tsx
+++ b/packages/frontend/src/components/LoginForm.tsx
@@ -12,7 +12,6 @@ import { I18nContext } from '../contexts/I18nContext'
 const log = getLogger('renderer/loginForm')
 
 import type { Credentials } from './Settings/DefaultCredentials'
-import { useSettingsStore } from '../stores/settings'
 
 const Socket = {
   automatic: 'automatic',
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/LoginForm.tsx
```diff
@@ -145,15 +148,15 @@ export default function LoginForm({ credentials, setCredentials }: LoginProps) {
             </div>
           )}
 
-          <p className='text'>{tx('login_no_servers_hint')}</p>
+          {!isChatmail && <p className='text'>{tx('login_advanced_hint')}</p>}
```
r10s: thinking it over, that hint makes sense for configured accounts as well, even if chatmail currently, they are advanced, should not changed to shared, and if changed, may become unencrypted. cmp. https://github.com/deltachat/deltachat-android/pull/4005#issuecomment-3504267145

```suggestion
          <p className='text'>{tx('login_advanced_hint')}</p>
```

r10s (approved):
the hint can be shown unconditionally, sorry for the noise, see comment, otherwise lgtm (tested, cannot say much to the code)

nicodh:
Failing test is not related

PR #5690, opened 2025-11-08
test: reduce composer test flakiness

WofWca:
The tests are still failing, but now it's (almost?) always at the same place. That place is an actual bug,
~~for fixing which I have made an MR: https://github.com/deltachat/deltachat-desktop/pull/5681.~~
Edit: umm, no, it's still failing...


WofWca:
Thanks
I'll try to remember to revert it when the bug is fixed.

Edit: I ran it a few times locally also. This should make them pass.

PR #5693, opened 2025-11-08
fix: no leave group for contact requests

### LINKED ISSUE #5678: Don't show a 'Leave' button in contact request groups
- Expected behavior: There is no 'Leave' button in contact request groups
- Actual behavior: There is a 'Leave' button, but it doesn't do anything when clicked
  - Get a contact request group
  - Right-click on it, or click on the three-vertical-dots menu

Closed core PR that would have made it possible to leave contact request groups, for reference: https://github.com/chatmail/core/pull/7007

nicodh:
resolves #5678

### DIFF
--- a/packages/frontend/src/components/chat/ChatListContextMenu.tsx
+++ b/packages/frontend/src/components/chat/ChatListContextMenu.tsx
@@ -326,7 +326,8 @@ export function useChatListContextMenu(): {
         singleChat &&
           isGroup &&
           singleChat.isEncrypted &&
-          singleChat.isSelfInGroup && {
+          singleChat.isSelfInGroup &&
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/chat/ChatListContextMenu.tsx
```diff
@@ -326,7 +326,8 @@ export function useChatListContextMenu(): {
         singleChat &&
```
Simon-Laux, line 326: I just noticed that this var naming "singleChat" is a bit confusing (too similar to the `ChatType: "Single"`), not related to the pr though.

Simon-Laux (approved):

PR #5695, opened 2025-11-09
fix: invert pinch zoom direction

### LINKED ISSUE #5641: Unexpected behaviour for pinch gesture on Mac
Zooming out currently zooms in

see https://github.com/deltachat/deltachat-desktop/issues/3893#issuecomment-3468437110

nicodh:
resolves #5641

### DIFF
--- a/packages/frontend/src/components/ImageCropper/index.tsx
+++ b/packages/frontend/src/components/ImageCropper/index.tsx
@@ -370,7 +370,12 @@ export default function ImageCropper({
       const absDelta = Math.abs(ev.deltaY)
       // NOTE(maxph): I'm not sure about this, but there are many sources stating that 'wheel' delta on touchpad is smaller (somewhere like 50 vs 100 at maximum)
       // so here we treat small delta as not a mouse wheel and zoom in different direction
-      const dir = absDelta < 50 ? 1 : -1
+      let dir = absDelta < 50 ? 1 : -1
+      const { isMac } = runtime.getRuntimeInfo()
+      if (isMac) {
+        // on macOS it seems we have to invert zoom direction
+        dir *= -1
+      }
       const normDelta = Math.min(8, absDelta)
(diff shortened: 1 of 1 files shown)


nicodh:
> This might fix it for the pinch gesture on Mac, but this inverts it for the scroll wheel.

Yes I know, but that's what I experienced in my tests. Did you test it with your touchpad? On mine the behaviour was wrong on Linux too. But maybe since I had to use an external touchpad which behaved like a Mac??

WofWca:
No, I only tested with a mouse.

WofWca:
@nicodh You seem to have edited my comment?

Have you tried using the mouse scroll wheel? It's supposed to be "scroll up to zoom in, scroll down to zoom out", which it is not in this MR.

nicodh:
> @nicodh You seem to have edited my comment? Sorry not by purpose - I will restore it
> 
> Have you tried using the mouse scroll wheel? It's supposed to be "scroll up to zoom in, scroll down to zoom out", which it is not in this MR.

Hmm, for me I have the opposite behaviour. What do you see in main branch when opening the full screen view (like in media gallery). There I see: scroll down => zoom in, scoll up => zoom out
The same behaviour I have in apps like gThumbs etc.

But most important: the scroll behaviour is th opposite of the behaviour in the Image Cropper. Can you confirm that? (in main branch)

nicodh:
> No, I only tested with a mouse.

So for me even Ctr + mouse move has the same zoom direction as other applications (and other components in desktop) with this PR.

Before it was just the opposition direction in the ImageCropper.

But of course the PR title is wrong then and I will change it

WofWca:
For me it's "scroll up to zoom in" in the fullscreen image view. Same is on the main branch for the image cropper.

nicodh:
It seems the zoom direction is not reliable for various touchpads on linux, so we just invert it for Mac for now where we know it works.

PR #5697, opened 2025-11-10
feat: add member filter in group profile

nicodh:
- added the expand icon as proposed
- collapse related chats when search is active
- changed the dialog to have a fixed height to avoid jumps when filtering
- moved ViewGroup to ViewGroup/index.tsx
- extract inline styles to style.modules.scss

Discussed final UI with @r10s

Based on #5234

image

image


r10s:
ftr, let's wait with merging until https://github.com/deltachat/deltachat-desktop/issues/5707 is resoved; we'll have a call about that these days

trufae:
eta wen

trufae:
https://github.com/deltachat/deltachat-desktop/issues/5707 is merged, whats the blocker now?

nicodh:
> #5707 is merged, what´s the blocker now?

We decided to add that feature to a (still missing) 3 dot menu. So there is nothing visible in normal state and in the context menu there are 2 items:
Filter members
Encryption info

The reason is, that there is no extra line for the number of members anymore so the serach icon would be placed "somewhere"  with unclear context see https://github.com/deltachat/deltachat-desktop/issues/5707#issuecomment-3541738154

Sorry for the delay, there were so many other priorities the last weeks...

trufae:
Sounds good. Thanks for the explanation :) no hurry at all, just wondering about the state of the things. Good work

nicodh:
Closed in favour of #6456

PR #5700, opened 2025-11-10
fix: hide or move email address

nicodh:
- don't show the mail address below single chat header
- move address & verified by contact information to the profile bottom

Address should no be displayed here:
image

### DIFF
--- a/packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
+++ b/packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
@@ -366,7 +366,7 @@ function chatSubtitle(chat: Type.FullChat) {
       } else if (chat.isDeviceChat) {
         return tx('device_talk_subtitle')
       }
-      return null
+      return chat.isEncrypted ? null : chat.contacts[0].address
     }
   }
   return 'ErrTitle'


packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
```diff
@@ -366,7 +366,7 @@ function chatSubtitle(chat: Type.FullChat) {
       } else if (chat.isDeviceChat) {
         return tx('device_talk_subtitle')
       }
-      return chat.contacts[0].address
+      return null
```
WofWca: Is this supposed to be in line with Android? Because it's not. This way the address is never displayed, which I don't think should be the case at least for regular (unencrypted) emails.

https://github.com/deltachat/deltachat-android/blob/85f679c69fa23ca48d36b0e188c65a16e2543823/src/main/java/org/thoughtcrime/securesms/ConversationTitleView.java#L81-L97

packages/frontend/src/components/dialogs/ViewProfile/index.tsx
```diff
@@ -371,6 +355,26 @@ export function ViewProfileInner({
...
+              )}
+              {contact.address && (
+                <div className={styles.addressLine}>
+                  <i className='material-svg-icon material-icon-server' />
+                  {addressLine}
+                </div>
+              )}
+            </div>
+          )}
         </>
```
WofWca, line 378: I wonder if we should blindly copy this UI arrangement from Android. With long "chats in common" lists this looks weird to me (actually, with short ones too). It's not expected that there is anything else below the list. I think this info should be just below "Last seen".

Image

r10s: currently, it is fine if it is hard to reach the address and partly broken/not-useful/hard-to-explain verified - both are more for debugging currently, and should not be that much in ones face

we will iterate over that, but atm it is about an easy fix and consistency

packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
nicodh: Added a condition

PR #5701, opened 2025-11-10
style: change link color

### LINKED ISSUE #5698: change linkified text color
linkified text just have a bottom border, but uses the same color as the normal text, it should use some link accent color, see:

https://github.com/deltachat/deltachat-android/pull/4011

nicodh:
resolves #5698


r10s (approved):
great, great, great - improving things by removing lines.

only bad, if one gets paid by lines - but who would come to that totally weird idea?

WofWca (commented):
Works on both themes, and it looks good.

PR #5703, opened 2025-11-11
fix: padding & hide icon in blocked contacts

### LINKED ISSUE #5661: "blocked contacts" layout glitches
the "blocked contact" dialog has some minor UI glitches:

- the padding at the bottom should be the same as the padding at the top
- the list should be left-aligned with the title
- i am not sure what the symbol 🚫 is doing there? does it have some meaning - or is it added unconditionally? in the latter case, just remove it, it is clear, this is about blocking, also for consistency with other UI

Image

nicodh:
resolves #5661

### DIFF
--- a/packages/frontend/scss/contact/_contact.scss
+++ b/packages/frontend/scss/contact/_contact.scss
@@ -67,7 +67,10 @@
   padding-bottom: 20px;
   padding-inline-start: 10px !important;
   // hide blocked contact icon
-  .contact-name .display-name i {
+  .material-icon-blocked {
     display: none;
+  }
+  p {
+    padding-inline-start: 20px;
   }
 }


packages/frontend/scss/contact/_contact.scss
```diff
@@ -61,3 +61,13 @@
     }
   }
 }
+
+// less padding for unblock contacts list
+.blocked-contacts {
+  padding-bottom: 20px;
+  padding-inline-start: 10px !important;
+  // hide blocked contact icon
+  .contact-name .display-name i {
```
WofWca:
```suggestion
  .material-icon-blocked {
```

WofWca (commented):
Now the `blockedContacts.length === 0` case looks uglier:

Image

nicodh:
> Now the `blockedContacts.length === 0` case looks uglier:

added some padding

nicodh:
> In the future this will break, but that's a problem for future us.

Why?

WofWca:
Because this adds a bunch of overrides for styles defined elsewhere. If one updates the code and styles in that elsewhere, or adds another `<p>`, then it will not be styled as expected.

nicodh:
> Because this adds a bunch of overrides for styles defined elsewhere. If one updates the code and styles in that elsewhere, or adds another `<p>`, then it will not be styled as expected.

I don't agree. The styles are all below the .blocked-contacts class, so any unexpected changes in the behavior can only happen there and the dev tools will point to the right place where the styles come from.

Only questionable point is the file where this styles are written down, which is the result of comparing the effort/result of:
1. adding an own stylesheet for that 2 rules or
2. introduce a new prop to be passed to the Contact component

WofWca:
> The styles are all below the .blocked-contacts class, so any unexpected changes in the behavior can only happen there

No. If one changes `.dialogContent`'s padding, then it will get out of sync with the code added in this MR. Same will happen if one decides to properly fix the contact list item to not have such a huge left padding. And if someone adds a `<p>` element to inside of the contact list, without explicitly checking if the "blocked contacts" dialog looks good, then they will not know that the new `<p>` has an extra padding in the "blocked contacts dialog" context.

nicodh:
> No. If one changes `.dialogContent`'s padding, then it will get out of sync with the code added in this MR. Same will happen if one decides to properly fix the contact list item to not have such a huge left padding. And if someone adds a `<p>` element to inside of the contact list, without explicitly checking if the "blocked contacts" dialog looks good, then they will not know that the new `<p>` has an extra padding in the "blocked contacts dialog" context.

But changes in generic components always have effect on multiple places/usages. Changing the padding in the contact list will not only affect the blocked contacts but all places that show the contact list.

So yes we would like to have a consistent styling/layout through all dialogs and reused components but this PR is not the place for such a refactoring.

WofWca:
> No.

OK, sorry I misunderstood you there. I'm not saying that these styles will break some other places.

> But changes in generic components always have effect on multiple places/usages. Changing the padding in the contact list will not only affect the blocked contacts but all places that show the contact list.

Yes, and that is the problem. We keep adding hacks instead of fixing the issue.

PR #5704, opened 2025-11-11
fix: hide member count if self not in group

### LINKED ISSUE #5699: only show group member count if dcChat.canSend() is true
in the group's title bar, when `dcChat.canSend()` is false (ex. when the joining process via invite link is not finished), instead of "N members" it should display `…`, since otherwise it will display "1 member" or other outdated number for groups you left

in the group profile, the "N members" label should be hidden if `dcChat.canSend()` is false

android implementation: https://github.com/deltachat/deltachat-android/pull/4014/commits/01ac63714260fa9c53fd206e61aa207262bcbd77

nicodh:
resolves #5699

### DIFF
--- a/packages/frontend/src/components/dialogs/ViewGroup.tsx
+++ b/packages/frontend/src/components/dialogs/ViewGroup.tsx
@@ -449,9 +449,7 @@ function ViewGroupInner(
                   ? tx('n_members', group.contactIds.length.toString(), {
                       quantity: group.contactIds.length,
                     })
-                  : tx('n_members', ' ', {
-                      quantity: group.contactIds.length,
-                    })
+                  : ''
                 : tx('n_recipients', group.contactIds.length.toString(), {
                     quantity: group.contactIds.length,
                   })}
--- a/packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
+++ b/packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
@@ -345,7 +345,7 @@ function chatSubtitle(chat: Type.FullChat) {
           quantity: chat.contactIds.length,
         })
       } else {
-        return '...'
+        return '…'
       }
     } else if (
       chat.chatType === C.DC_CHAT_TYPE_SINGLE &&


packages/frontend/src/components/dialogs/ViewGroup.tsx
```diff
@@ -445,9 +445,13 @@ function ViewGroupInner(
...
-                ? tx('n_members', group.contactIds.length.toString(), {
-                    quantity: group.contactIds.length,
-                  })
+                ? group.selfInGroup
+                  ? tx('n_members', group.contactIds.length.toString(), {
+                      quantity: group.contactIds.length,
+                    })
+                  : tx('n_members', ' ', {
+                      quantity: group.contactIds.length,
+                    })
```
WofWca: For some languages (including mine) the result is gonna sound off.

I think we don't need to remove this label here. The "View Group" looks different on Desktop compared to Android. On Android the number of members is a "subtitle" just below the group name. If we show the list of members, then it's fine to also keep the title.

If it's insisted that the "N members" thing needs to be removed, then I suggest to either add a proper string, or to remove the element completely, though Accessibility will then have to be taken into account.

packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
```diff
@@ -340,9 +340,13 @@ function chatSubtitle(chat: Type.FullChat) {
...
     if (chat.chatType === C.DC_CHAT_TYPE_GROUP) {
-      return tx('n_members', [String(chat.contactIds.length)], {
-        quantity: chat.contactIds.length,
-      })
+      if (chat.selfInGroup) {
+        return tx('n_members', [String(chat.contactIds.length)], {
+          quantity: chat.contactIds.length,
+        })
+      } else {
+        return '...'
```
WofWca: This will also be displayed if you left / have been kicked from the group.

nicodh: but that's another bug then: selfInGroup should not return true if you left the group or has been kicked off the group!

packages/frontend/src/components/dialogs/ViewGroup.tsx
nicodh: As I already mentioned in the Issue: hiding the number but showing the whole list is a bit paradox. Let's wait for @r10s opinion on that

packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
WofWca: No, I mean `...` will be displayed in such a case. IMU `...` is supposed to mean "loading" or something.

nicodh: Wait  - maybe I misunderstood the whole Issue: I assumed we should never show the number of members if we left a group or are not member anymore (maybe we don't get updates of member changes anymore) but the issue is just about "when the joining process via invite link is not finished".

I will change that...

WofWca: Well apparently that's how Android is doing it right now, and I am pointing out that this might not be the best solution.

nicodh: After some discussion s with @adbenitez 
The issue is about both cases:
1. you are in a join group process
2. you are not member anymore in the group

In both cases we should not show the member count in the Chat header but still show the list of members in the group profile (although it only shows one member in case 1. and might be outdated in case 2.)

The group profile display is a bit paradox then, but the issue is mainly to avoid confusion about the wrong header in case 1. ?
@r10s did I get it right? :-)

r10s: yes, you got it right.

it is a little drawback, that the "..." is shown as well when one left the group, but that was accepted as the benefit outweights it, and tracking the state is out of scope for now and not worth the effort. see discussion at https://github.com/deltachat/deltachat-android/issues/3997

and yes, for the "profile", the whole label should be removed. if would not spent much effort on that in this PR, it is fine, if only a grey bar is left.

in a subsequent PR, the "N members" title should become a centered subtitle as on android/iOS, it is an oversight from me that this was not adapted yet (it was considered an important difference to one-to-one profiles, that group profiles have a subtitle, but also it looks better and is needed for consistency)
EDIT: created an issue for that part at https://github.com/deltachat/deltachat-desktop/issues/5707

packages/frontend/src/components/dialogs/ViewGroup.tsx
r10s: the whole "N members" title should be hidden, not only the number? but not sure if i got the discussion correctly

nicodh: Yes I will adapt that. I'm just changing it

packages/frontend/src/components/dialogs/ViewGroup.tsx
```diff
@@ -445,9 +445,11 @@ function ViewGroupInner(
...
             >
               {!isBroadcast
-                ? tx('n_members', group.contactIds.length.toString(), {
-                    quantity: group.contactIds.length,
-                  })
+                ? group.selfInGroup
+                  ? tx('n_members', group.contactIds.length.toString(), {
+                      quantity: group.contactIds.length,
+                    })
+                  : ''
```
WofWca, line 452: The element itself is still displayed in such a case, but it's fine since we'll rewrite this stuff anyway soon: https://github.com/deltachat/deltachat-desktop/issues/5707

WofWca (approved):
This does what's intended (as per the comments), and it works.

PR #5708, opened 2025-11-12
fix: replace iframe in connectivity dialog

### LINKED ISSUE #5706: Connectivity dialog empty (sometimes)
When opening the Connectivity dialog it (often) shows no content. In the dev tools it shows the correct ConnectivityHTML and styles but the iframe content is not visible

nicodh:
resolves #5706

Iframe is not (reliably) working since we stopped disabling the IsolateSandboxedIframes feature (which is deprecated)

I also tested with tauri and it looks fine there


WofWca:
Maybe it's simpler to just re-disable `IsolateSandboxedIframes` for now.

nicodh:
> Maybe it's simpler to just re-disable `IsolateSandboxedIframes` for now.

Maybe, at least safer for the current release

nicodh:
Closed in favor of #5711

PR #5709, opened 2025-11-12
chore: upgrade `@deltachat/calls-webapp`

WofWca:
It includes the patch for showing the "P2P" / "non-P2P" text.

I have tested it. The app is still working, and the new text appears.


nicodh (approved):
Reviewed by testing

PR #5713, opened 2025-11-12
Fix #4820 - Edit last message when pressing the UP arrow key

### LINKED ISSUE #4820: ArrowUp key in composer should open edit message for last sent message
This is a common convenience feature that many messengers have (telegram, discord, element).
Would fit nicely with the existing shortcuts for choosing a message to answer-to/quote.

Timing: can be done already now or if not I'll later address this in my planned composer refactoring

trufae:
Closes https://github.com/deltachat/deltachat-desktop/issues/4820.

### DIFF
--- a/packages/frontend/src/components/composer/Composer.tsx
+++ b/packages/frontend/src/components/composer/Composer.tsx
@@ -121,7 +121,8 @@ const Composer = forwardRef<
   const messageEditing = useMessageEditing(
     accountId,
     chatId,
-    editMessageInputRef
+    editMessageInputRef,
+    regularMessageInputRef
   )
   /**
    * Use this only if you're sure that what you want to do with the ref
@@ -322,6 +323,12 @@ const Composer = forwardRef<
       if (ev.type === 'keydown' && ev.code === 'Escape') {
         setShowEmojiPicker(false)
         setShowAppPicker(false)
+        if (messageEditing.isEditingModeActive) {
+          messageEditing.cancelEditing()
+          setTimeout(() => {
+            regularMessageInputRef.current?.focus()
+          })
+        }
       }
     }
     // these options are needed, otherwise emoji mart sometimes eats the keydown event
@@ -334,7 +341,7 @@ const Composer = forwardRef<
       document.removeEventListener('keydown', onKey, opt)
       document.removeEventListener('keyup', onKey, opt)
     }
-  }, [shiftPressed])
+  }, [shiftPressed, messageEditing, regularMessageInputRef])
 
   useEffect(() => {
     if (!showEmojiPicker) return
@@ -757,7 +764,8 @@ export default Composer
 function useMessageEditing(
   accountId: number,
   chatId: T.BasicChat['id'],
-  editMessageInputRef: React.RefObject<ComposerMessageInput | null>
+  editMessageInputRef: React.RefObject<ComposerMessageInput | null>,
+  regularMessageInputRef: React.RefObject<ComposerMessageInput | null>
 ) {
   const tx = useTranslationFunction()
   const { userFeedback } = useContext(ScreenContext)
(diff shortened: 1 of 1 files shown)


trufae:
not sure how to address the e2e tests

nicodh (changes requested):
Edit message should only be possible for own messages. The current implementation also tries to edit messages from others (if the last message is from another user).

This is not possible and should be avoided

nicodh:
> not sure how to address the e2e tests

The error is not related to this PR. Rebasing to main should solve it

packages/frontend/src/components/composer/Composer.tsx
```diff
@@ -134,6 +134,40 @@ const Composer = forwardRef<
...
+        .find(item => item.kind === 'message' && item.msg_id > 0)
+      if (!lastMessageItem || lastMessageItem.kind !== 'message') {
+        return
+      }
+      // Load the full message
+      const messages = await BackendRemote.rpc.getMessages(accountId, [
+        lastMessageItem.msg_id,
+      ])
+      const message = messages[lastMessageItem.msg_id]
+      if (message && message.kind === 'message' && message.state >= 10) {
```
nicodh: Check should be message.fromId === C.DC_CONTACT_ID_SELF to exclude other authors messages.

But there are more checks necessary. See https://github.com/deltachat/deltachat-desktop/blob/main/packages/frontend/src/components/message/Message.tsx#L287

nicodh: It would make sense to have an own function isMessageEditable in https://github.com/deltachat/deltachat-desktop/blob/main/packages/frontend/src/components/message/messageFunctions.ts to avoid redundant code

trufae: what about now?

packages/frontend/src/components/composer/Composer.tsx
```diff
@@ -134,6 +134,45 @@ const Composer = forwardRef<
     [chatId, updateDraftText]
   )
 
+  const onArrowUpWhenEmpty = useCallback(async () => {
+    if (messageEditing.isEditingModeActive) {
+      return
+    }
+    try {
+      // Get message list items for the chat
+      const messageListItems = await BackendRemote.rpc.getMessageListItems(
```
nicodh: This call and the next BackendRemote.rpc.getMessages call can be avoided by passing the messageCache from MessageListAndComposer to Composer as a prop.

```
messageCache={messageListState.messageCache}
```
and then use the messageCache similar to Line 451 in [useDraft](https://github.com/deltachat/deltachat-desktop/blob/main/packages/frontend/src/hooks/chat/useDraft.ts#L451)

```suggestion
     // Dummy code!
      const messageIds = Object.keys(messageCache)
      .map(Number)
      
    const message = messageCache[messageIds.length - 1]
```

trufae: thanks! i didnt know that was possible. i'm not really into react/ui stuff dev and i assume there are lots of patterns im missing, so thanks for the careful reviews :)

trufae:
i found a couple of behaviours that are not working as i would expect and i'm trying to fix them in a separate commit, can you check it out? the behaviour now is good

* ESC key cancels the current editing
* Enter/Return focus the input field again

trufae:
i removed the comment
```
-    // TODO focus the "regular message" input?
-    // Or is it not a bug but a feature,
-    // so that you don't accidentally send the draft
-    // right after editing a message?
-    // Though this doesn't apply if you send a message by clicking
-    // the "send" button with a pointer device (mouse).
-    //
-    // Maybe instead, to guard against this, we could simply disable
-    // the "send" function for ~1 second after the edit is done.
```

because i think that's solved with my patch

nicodh (changes requested):
reviewed by testing & reading

Looks good now!

 Just a rebase would be nice, should fix the failings e2e tests

trufae:
Rebased

trufae:
No matter how many times i rebase the e2e tests are always red

PR #5718, opened 2025-11-14
Prepare release 2.25.1

nicodh:


WofWca (approved):
Looks good, just a version bump.
This should settle the fuss caused by "2.25.0", and fix an important bug.

PR #5722, opened 2025-11-14
fix: avatar initials not being always uppercase

WofWca:
The issue has been introduced in
18c4250ad55f52f91073c84724549dfa13c5322d
(https://github.com/deltachat/deltachat-desktop/pull/5652).


r10s (approved):
tested, works as expected :)

cannot say much to the code, tho lgtm :)

Issue #5723, opened 2025-11-14
attaching vcard on 2.25.1 is broken, memory usage >10 gb after some seconds


r10s:
when trying to attach a vcard on 2.25.1 (macOS), i just get an empty list.

tried to restart and also to use smaller profiles (first thought it is just slow), still not working.

and: once i tried opening the list, the app is broken - and switching to other chats does no longer work at all until i restart the app

https://github.com/user-attachments/assets/d9d1f8ec-b719-41e5-9f08-734ca9b0b168

nicodh:
Can't reproduce that on MacOS. Neither on building from main nor on the provided DeltaChat-2-25-1-universal.dmg from https://download.delta.chat/desktop/v2.25.1/

Anything errors in the logs? Or in the dev tools console?

edit: I now saw it once in current release on Linux, but even after intensive testing with some accounts with many contacts I don't see it anymore

r10s:
i am using the arm64 build - i did not see anything special in the log - but when trying to open the log, the list suddenly appears, probably minutes after i opened it. could not reproduce that again, now the list is just always gone again. may it be that it is incredibly slow?

EDIT: while waiting some minutes with the laptop on the legs, it becomes really hot, and battery shows a warning that delta chat is using "significant energy" - afk for now, can have a closer look tomorrow

Raiden-GH:
I can confirm this bug.

r10s:
i tried over, using a small account (the others are disaled by "only sync selected profile") - after some seconds memory usage for the **rpc-server** goes from 20 mb to **up to >10 gb**. the log does not show anything special, video and log below:

https://github.com/user-attachments/assets/96c37a71-3679-4a45-980e-2e6fb1e343e2

https://gist.github.com/r10s/52cbd0cb0423acc7d076f1a21f142dfd

WofWca:
> and switching to other chats does no longer work at all

This seems to indicate that the Core is not responding to JSON-RPC requests.

I am unable to reproduce on Windows.

How about the "New Chat" dialog? Does it work fine?

WofWca:
This is the dialog that gets opened:

https://github.com/deltachat/deltachat-desktop/blob/31a2944b66261401b9036dee7d669bf60e64a74c/packages/frontend/src/components/dialogs/SelectContact/index.tsx#L15-L119

[...]

PR #5725, opened 2025-11-15
fix: app hang on attach contact sometimes

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/5723.

It seems that sometimes this causes front-end to basically
infinitely spam the backend with `get_contacts_by_ids`.

`isContactLoaded` also starts returning `true`
as soon as we have started loading a contact,
and not only when we have loaded it.

This might not completely fix such a potential bug,
but this should at least make it not happen in practice.

TODO:
- [ ] Verify that the bug no longer happens. This can be done after merging the MR,
  in which case the issue should only be closed when it is confirmed that the bug is fixed.


WofWca:
Yep, I noticed that, I don't know why that's happening, maybe some `react-infinite-loader` internals.

PR #5727, opened 2025-11-15
perf: don't load chats, contacts, gallery twice

WofWca:
It seems that `react-window-infinite-loader` calls `isItemLoaded`
before React gets a chance to re-render with the new state of
e.g. `chatLoadState`, and so `react-window-infinite-loader`
calls `loadMoreItems` twice.

Let's convert this state to a ref then so that the changes to it
are reflected immediately.
It is fine to do so because apparently the `isItemLoaded` function
is not used for rendering, but only for determining whether
`loadMoreItems` needs to get called.

Related:
- https://github.com/deltachat/deltachat-desktop/pull/5725#pullrequestreview-3467878299.

TODO:
- [x] Make a release, because this MR probably shouldn't go into the next one.

### DIFF
--- a/packages/frontend/src/components/chat/ChatList.tsx
+++ b/packages/frontend/src/components/chat/ChatList.tsx
@@ -58,11 +58,6 @@ import { useHasChanged2 } from '../../hooks/useHasChanged'
 const log = getLogger('ChatList')
 const useMultiselectLog = getLogger('ChatListMultiselect')
 
-const enum LoadStatus {
-  FETCHING = 1,
-  LOADED = 2,
-}
-
 /**
  * This component holds either a list of chats OR the result
  * of a search query including chats, contacts and messages.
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/chat/ChatList.tsx
```diff
@@ -749,52 +744,43 @@ export function useLogicVirtualChatList(chatListIds: number[]) {
   const chatCacheRef = useRef<typeof chatCache>({})
   chatCacheRef.current = chatCache
 
-  const [chatLoadState, setChatLoading] = useState<{
-    [id: number]: undefined | LoadStatus.FETCHING | LoadStatus.LOADED
-  }>({})
+  const loadingChats_ = useRef(new Set<T.BasicChat['id']>())
```
nicodh: I would prefer to use loadingChatsRef as name (as we did for chatCacheRef). It's easier to distinguish and more self explaining

packages/frontend/src/components/chat/ChatList.tsx
```diff
@@ -856,7 +836,15 @@ export function useLogicVirtualChatList(chatListIds: number[]) {
     }
   }, [accountId])
 
-  return { isChatLoaded, loadChats, chatCache }
+  return {
+    /**
+     * This function is not reactive, i.e. it doesn't get re-created
+     * when we start loading more items.
```
nicodh, line 842:
```suggestion
     * when we start loading more items.  It also returns true if a
     * chat is not yet loaded but is in the process of being loaded.
```

nicodh: I suggest to add this comment or rename the function since the current name could be misleading

packages/frontend/src/components/chat/ChatList.tsx
```diff
@@ -883,37 +871,37 @@ function useMessageResults(
   const [messageCache, setMessageCache] = useState<{
     [id: number]: T.MessageSearchResult | undefined
   }>({})
-  const [messageLoadState, setMessageLoading] = useState<{
-    [id: number]: undefined | LoadStatus.FETCHING | LoadStatus.LOADED
-  }>({})
+  const loadingMessages_ = useRef(new Set<T.MessageSearchResult['id']>())
```
nicodh: same as above loadingMessagesRef

packages/frontend/src/components/chat/ChatList.tsx
```diff
@@ -883,37 +871,37 @@ function useMessageResults(
...
+      loadingMessages.delete(id)
+    }
   }
 
   return {
     messageResultIds,
+    /**
+     * This function is not reactive, i.e. it doesn't get re-created
+     * when we start loading more items.
+     */
```
nicodh, line 903: see other comment

packages/frontend/src/components/chat/ChatList.tsx
WofWca: Adding `Ref` makes sense, though I wanted to mark it as "private", so how about `loadingChatsRef_`?

packages/frontend/src/components/chat/ChatList.tsx
WofWca: Can be done in a follow-up

packages/frontend/src/components/chat/ChatList.tsx
nicodh: Hmm, imo private only applies to class variables and inheritance, but I assume you mean it as "should not be modified". I'm not convinced that there is a benefit from that underscore but won't insist to remove it

WofWca: I removed the `ref` vars entirely in some placed, and renamed in others.

PR #5733, opened 2025-11-17
fix: tauri: user-notify deleted images when added to notifications on macOS.

### LINKED ISSUE #5211: tauri: GIF, JPEG, and PNG files generate an error message for the recipient
When sending GIF, JPEG, and PNG files as files, images, or stickers, the following error message is displayed on the recipient's side (same DC-Version):

`loading message 734 failed: failed to get /Users/User/Library/Application
Support/chat.delta.desktop.tauri/accounts/8efae49e-bf79-4fba-8300-c2b185096e6d/dc.db-
blobs/06ec500d967633a574877e75994c999.png size in bytes: No such file or directory (os
error 2)`

If the recipient sends the same image, file, or sticker to the sender, the error message disappears and everything is displayed normally.

macOS 15.5
DC-Desktop 1.59.1 (Tauri)

### LINKED ISSUE #5212: tauri: sender's avatar not displayed
When starting a new chat with an existing contact, their avatar is displayed correctly for the sender. However, a question mark appears in the recipient's profile instead of the sender's avatar. If the recipient replies, the recipient's avatar also disappears from the sender's profile and is replaced by a question mark.
A similar problem also occurs with group avatars. It's likely caused by the same issue. That's why I'm describing the bug here.
If a group member sets an avatar, it is displayed for them, but only a question mark appears for the other members.

macOS 15.5
DC-Desktop 1.59.1 (Tauri)

Image

Simon-Laux:
So we copy them now in the user-notify crate.

It was not a core issue core after all, but our notification crate or macOS System API itself.

closes #5212
closes #5211

### DIFF
--- a/packages/target-tauri/crates/user-notify/examples/test.rs
+++ b/packages/target-tauri/crates/user-notify/examples/test.rs
@@ -1,4 +1,4 @@
-use std::{collections::HashMap, path::PathBuf, time::Duration};
+use std::{collections::HashMap, time::Duration};
 
 use tokio::{signal::ctrl_c, spawn, time::sleep};
 use user_notify::{NotificationCategory, NotificationCategoryAction, get_notification_manager};


packages/target-tauri/crates/user-notify/src/platform_impl/mac_os/builder.rs
```diff
@@ -53,9 +53,21 @@ fn build(
...
+            // file needs to be copied, because macOS will move it to internal datastore and delete it afterwards.
+            let copied_file_path = env::temp_dir().join(format!(
+                "{}.{}",
+                uuid::Uuid::new_v4().to_string(),
+                original_path
+                    .extension()
+                    .and_then(|e| e.to_str())
+                    .unwrap_or_default()
+            ));
+            fs::copy(original_path, &copied_file_path)?;
```
WofWca, line 66: Should this be async?

WofWca (approved):
Wow, this is really odd behavior. Good that you managed to find that! I only managed to find https://developer.apple.com/documentation/usernotifications/unnotificationattachment as a source:

> The system limits the amount of storage space allocated for attachments for each app. To delete attachments, use the methods of the [UNUserNotificationCenter](https://developer.apple.com/documentation/usernotifications/unusernotificationcenter) class to remove the notification requests that contain those attachments.

Didn't test, but the code looks sensible.
Also there is a clippy warning.

packages/target-tauri/crates/user-notify/src/platform_impl/mac_os/builder.rs
Simon-Laux: method is not async, so it would need a greater refactoring to make it async.
Combining Objective C and rust async code is also a bit annoying.., let's think about it again if it really becomes a performance issue.

Simon-Laux:
> Didn't test, but the code looks sensible.

I tested it with the example and managed to reproduce the issue (it deleted my test image every time).
After my change only the temporary copy was deleted.

Couldn't test it in the actual tauri release, because I didn't have the correct signing keys and don't wanted to deal with getting them. We can test it properly with the first release after this is merged

PR #5734, opened 2025-11-17
chore: add FreeBSD to distribution platforms

nicodh:
Add FreeBSD to the list of distribution channels in "Release status issue"


nicodh:
@feld if you agree we would also list FreeBSD on https://delta.chat/en/download

feld:
I think that's a good idea. Let's add it and I'll continue to coordinate the release distribution with the team

PR #5735, opened 2025-11-17
refactoring: use https protocol for maps

nicodh:


packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -214,6 +214,23 @@ export default class DCWebxdc {
         })
         await ses.closeAllConnections()
 
+        ses.webRequest.onBeforeRequest(
+          {
+            urls: ['<all_urls>'],
+          },
+          async (details, cb) => {
```
```suggestion
          (details, cb) => {
```

packages/target-electron/src/deltachat/webxdc.ts
WofWca: This probably needs to be renamed then, e.g. `cspAllowAllHttpsImgSrc`

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -214,6 +214,23 @@ export default class DCWebxdc {
...
+              cancelRequest = false
+            } else if (details.url.startsWith('devtools://')) {
+              cancelRequest = !DesktopSettings.state.enableWebxdcDevTools
+            } else if (details.url.startsWith('https://')) {
+              cancelRequest = !internetAccess
+            }
+            cb({ cancel: cancelRequest })
+          }
+        )
+
```
WofWca, line 233: What's the use of this piece of code? Is it relevant for this MR? Things seem to be working fine without it.

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -1024,13 +1040,13 @@ const makeResponse = ({
 }
 
 /**
- * the maps:// protocol may only be used by apps
- * that are allowed to access the internet
+ * this should only be allowed for apps that are
```
```suggestion
 * this should only be used by apps that are
```

packages/target-electron/src/deltachat/webxdc.ts
nicodh: This is just an additional layer in case the other restrictions do not work for any reason. Electron calls this before any request and if cancelRequest is true aborts it.

It was my first approach to  intercept https requests and call rpc.getHttpResponse but then I decided for reusing the existing protocolHandler implementation.

PR #5738, opened 2025-11-17
change: move group member count

### LINKED ISSUE #5707: profile consistency: center "N members" string
during recent profile cleanups, the "N members" string was moved below the title and centered. 

reason is to have a immediate, clear distinction between group and one-to-one profiles.

moreover, it was regarded to look better, clearer and less cluttering. it also makes it easier to show part of the profile consistently in the chat in the future when a chat gets created, similar to signal/whatsapp

for consistency with android/iOS, this should be done on desktop as well:

before / after (roughly):

nicodh:
resolves #5707

### DIFF
--- a/packages/frontend/scss/dialogs/_group-styles.scss
+++ b/packages/frontend/scss/dialogs/_group-styles.scss
@@ -84,4 +84,5 @@ input.group-name-input {
 
 .group-profile-subtitle {
   text-align: center;
+  padding-top: 6px;
 }


packages/frontend/scss/dialogs/_group-styles.scss
```diff
@@ -81,3 +81,7 @@ input.group-name-input {
     padding-inline-start: 20px;
   }
 }
+
+.group-profile-subtitle {
+  text-align: center;
```
r10s, line 86: played around a bit, giving the subtitle a little more whitespace looks much better, also in comparison with android/ios.

```suggestion
  text-align: center;
  padding-top: 6px;
```

Issue #5731, opened 2025-11-15
Adapt to withdrawable broadcast channel invites


Hocuri:
When updating to a core with https://github.com/chatmail/core/pull/7439 merged (most probably v2.27.0), `check_qr()` can return `DC_QR_REVIVE_JOINBROADCAST` or `DC_QR_WITHDRAW_JOINBROADCAST`. You need to mostly do the same things as when a group invite is revived/withdrawn.

- [ ] When `check_qr()` returns `DC_QR_WITHDRAW_JOINBROADCAST`, show the same dialog as `DC_QR_WITHDRAW_VERIFYGROUP`, but with the string `withdraw_joinbroadcast_explain` rather than `withdraw_verifygroup_explain`
- [ ] When `check_qr()` returns `DC_QR_REVIVE_JOINBROADCAST`, show the same dialog as `DC_QR_REVIVE_VERIFYGROUP`
- [ ] When the user clicks on "Reset QR code", and the chat type is broadcast, use the string `withdraw_joinbroadcast_explain`

See https://github.com/deltachat/deltachat-android/pull/4022/.

PR #5739, opened 2025-11-17
Update core 2.27.0

nicodh:


WofWca (approved):
This also upgrades ESBuild.

Regarding whether https://github.com/deltachat/deltachat-desktop/pull/5530 has been correctly rebased, I only checked `MainScreen.tsx`.

Looking at the core CHANGELOG, it appears that we'll need to take care of https://github.com/deltachat/deltachat-desktop/issues/5731 (<https://github.com/chatmail/core/pull/7439>). I am not sure if it's correctly marked as "breaking" though, since the build is passing in this MR. Maybe we're lacking some type exhaustivity checks.

nicodh:
> Looking at the core CHANGELOG, it appears that we'll need to take care of #5731 ([chatmail/core#7439](https://github.com/chatmail/core/pull/7439)). I am not sure if it's correctly marked as "breaking" though, since the build is passing in this MR. Maybe we're lacking some type exhaustivity checks.

Yes #5731 should be fixed soon in a followup

PR #5743, opened 2025-11-19
feat: show warning when disabling multi-device mode

### LINKED ISSUE #5752: Rename "Send copy to self" to "Multi-device mode" and always display it
`bcc_self` setting now means "multi-device mode". It should not be hidden for chatmail profiles anymore as we want to minimized the difference between chatmail and non-chatmail profiles to prepare for multi-transport.

Corresponding Android PR already implementing this: https://github.com/deltachat/deltachat-android/pull/4019

Similar iOS issue: https://github.com/deltachat/deltachat-ios/issues/2908

nicodh:
Also resolves #5752 by replacing the old send copy to self wording with the new multi device texts

image

When disabling multi device mode:

image

### DIFF
--- a/packages/frontend/src/components/Settings/ChatsAndMedia.tsx
+++ b/packages/frontend/src/components/Settings/ChatsAndMedia.tsx
@@ -65,10 +65,10 @@ export default function ChatsAndMedia({
           <SettingsHeading>{tx('delete_old_messages')}</SettingsHeading>
           <Autodelete settingsStore={settingsStore} />
           <SettingsSeparator />
-          <SettingsHeading>{tx('pref_backup')}</SettingsHeading>
-          <Backup />
         </>
       )}
+      <SettingsHeading>{tx('pref_backup')}</SettingsHeading>
+      <Backup />
     </>
   )
 }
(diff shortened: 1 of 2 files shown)


r10s:
sorry for not filing some concrete instructions beforehand.

the "multi-device mode" switch and strings as such are correct, however

- it needs to be shown unconditionally for chatmail/non-chatmail (before, it was non-chatmail only)
- for chatmail, the option "Chats and Media / Delete Old Messages / Delete Messages From Server" shall be removed - for non-chatmail, is stays available for now


@adbenitez you were driving that, is that correct? a migration or so was not needed at the end, right?

nicodh:
Now the settings look like this:

Non Chatmail:

image

Chatmail:

image

@r10s  I assume "Show Classic E-Mails" can be hidden on non chatmail?

r10s:
yes - show-email should not be available for chatmail as well

but that is already the case on main, it seemed to sneak in by this PR again :)

r10s (approved):
functionality wise, this looks good to me now, also tested a bit.

a minor thing is that showMultiDeviceWarning lacks a "Cancel" button - if this is an easy thing to add, i would go for that. but that is no blocker.

i cannot say much to the code.

Issue #5744, opened 2025-11-19
Footer on the image overlaps it and cannot be clicked to zoom the image


link2xt:
Here is an example:
<img width="436" height="391" alt="Image" src="https://github.com/user-attachments/assets/c56193c6-bcf9-496a-825b-262391417ee2" />

Red border is the `<footer>` element to which I added border style. It is not clickable, I have to click above or below it to zoom the image. With images that have small height this is easy to notice because they have only a few pixels above and below the footer.

r10s:
i want that wizard!

WofWca:
This is especially problematic for videos. You can't really use the controls.

One simple solution is just not to take the whole width.

ralphtheninja:
> One simple solution is just not to take the whole width.

I had the same thought. Let the little black square rest in the bottom right corner.

ralphtheninja:
<img width="671" height="257" alt="Image" src="https://github.com/user-attachments/assets/8024162e-735c-4327-837a-3df13aa4fc26" />

Removing the "left" property in the css fixes it.

PR #5745, opened 2025-11-19
Fix: Reduce image footer width in media only messages

ralphtheninja:
Reduce the width of the footer so it doesn't cover the image.

Closes https://github.com/deltachat/deltachat-desktop/issues/5744

### DIFF
--- a/packages/frontend/src/components/message/styles.module.scss
+++ b/packages/frontend/src/components/message/styles.module.scss
@@ -26,7 +26,7 @@
     bottom: 5px;
     margin-inline-start: 10px;
     position: absolute;
-    right: 5px;
+    inset-inline-end: 5px;
   }
 
   // &.withReactionsNoText {


ralphtheninja:
Do we need changelog for this?

Simon-Laux:
> Do we need changelog for this?

if is a bug fix so would make sense, But changelogs are now generated from commit messages like in chatmail-core.
So the pr title becomes the changelog entry. read more at https://github.com/deltachat/deltachat-desktop/blob/main/CONTRIBUTING.md#commit-messages-and-changelog

ralphtheninja:
> if is a bug fix so would make sense, But changelogs are now generated from commit messages like in chatmail-core.
> So the pr title becomes the changelog entry. read more at https://github.com/deltachat/deltachat-desktop/blob/main/CONTRIBUTING.md#commit-messages-and-changelog

Roger that. So the quality of the PR title is more important now than before. I'll update this.

PR #5756, opened 2025-11-22
change: Improve About Screen: copy every info to clipboard, add space usage report and more

Simon-Laux:
depends on https://github.com/chatmail/core/pull/7486 (released in core 2.18)

Changes:
- larger about dialog
- move copy button to top, so users find it faster
- copy button now copies all visible info and more (like the full runtime info object)
- refactor a bit to cleanup code and make it more resistant to errors using RpcFetch hook
- remove duplicated mention of runningUnderARM64Translation
- add space usage report

screenshots:
image image

### DIFF
--- a/packages/frontend/src/components/dialogs/About.tsx
+++ b/packages/frontend/src/components/dialogs/About.tsx
@@ -36,16 +36,16 @@ function useAboutInfo() {
       : { accountInfoError: 'loading' }
     : { accountInfoError: 'no account selected' }
 
-  const spaceUsageReportFetch = useRpcFetch(
-    BackendRemote.rpc.getSpaceUsageReportString,
+  const storageUsageReportFetch = useRpcFetch(
+    BackendRemote.rpc.getStorageUsageReportString,
     accId ? [accId] : null
   )
 
-  const spaceUsageReport: string = spaceUsageReportFetch
-    ? spaceUsageReportFetch.lingeringResult
-      ? spaceUsageReportFetch.lingeringResult.ok
-        ? spaceUsageReportFetch.lingeringResult.value
-        : JSON.stringify(spaceUsageReportFetch.lingeringResult.err)
+  const storageUsageReport: string = storageUsageReportFetch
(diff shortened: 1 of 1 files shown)


r10s:
i know this is a draft, but already now: 

the debug info before were already questionable, but we should not manifest them in "about", directly in the face of the user in a usually expected "nice" dialog. also the split with log is questionable

i am afk just now (weekend ;), just saying to not put more effort in here

we will discuss that in the ui/ux team next week

Simon-Laux:
> we will discuss that in the ui/ux team next week

awesome, putting my thoughts here then for consideration:

> but we should not manifest them in "about", directly in the face of the user in a usually expected "nice" dialog.

true, makes the dialog ugly and huge.

debug info could be a separate dialog.

For the about dialog:
- should show versions and what runtime is used
- additional runtime info may or may not be interesting (could be part of versions table):
   - User Agent on Browser Edition
   - used core path on non-tauri edition (maybe only dev-mode as this is not interesting for users and already part of log)
   - whether it runs under rosetta
- I don't really like the links included in the translated text, they use some string replace which may break if translators translate the part of the string that should be substituted
- and the links also lack a symbol that they open the web. (I came across this while researching for my answer to a forum topic https://support.delta.chat/t/do-not-suddenly-open-links-in-the-browser-without-warning/4327/2)

For the debug info:
- new dialogs are super easy to make in desktop (as long as they don't need new custom design)
- I think the table makes it slightly more readable, also because python and rust tests already ensure the format of get_info to be "key=value[linebreak]repeat" which is automatically converted to a dict/object by jsonrpc bindings.
- copy output format json is easiest, allows us to easily include runtime info and other desktop specific debug info.

> also the split with log is questionable

log is a file. a hacky solution could be to print debug info to log when you open either current log file or the log folder.

There is also no direct share option, maybe it could be nice to have a button to share it to a chat (re using webxdc share to chat logic) or to a predefined support contact?

r10s:
closing in favor of https://github.com/deltachat/deltachat-desktop/issues/5772

PR #5757, opened 2025-11-22
fix: only fetch system info one time on startup not every time you switch accounts

Simon-Laux:
only fetch and log system info on startup, its content does not change anyway.
It is still useful to have, so that log contains core and sqlite version atleast once.


Simon-Laux:
haven't seen #5755, so closing this

PR #5759, opened 2025-11-24
chore: upgrade core to 2.28.0

nicodh:


WofWca:
Have we ensured the things that need to be ensured for the 2.28 Core release? i.e. that transports can't be created accidentally and stuff (see the private chat)?

nicodh:
> Have we ensured the things that need to be ensured for the 2.28 Core release? i.e. that transports can't be created accidentally and stuff?

That should be addressed in #5680 - I'm just on it. But we need the new core in main and rebase the PR first

nicodh:
Any idea why github/workflows/build-tauri-preview.yml  is not valid anymore?
Must be related to the rustc upgrade
 Error: The template is not valid. .github/workflows/build-tauri-preview.yml (Line: 40, Col: 16): hashFiles('**/Cargo.lock') failed. Fail to hash files under directory '/Users/runner/work/deltachat-desktop/deltachat-desktop'

WofWca:
> Any idea why github/workflows/build-tauri-preview.yml is not valid anymore?
> Must be related to the rustc upgrade

I doubt it. The action step runs right after checkout, and it's working fine on other OSes.
This probably has to do with the runner or with the `cache` action.

Let's watch https://github.com/orgs/community/discussions/180160

WofWca (approved):
So a Rust upgrade plus a Core upgrade, plus some minor Rust package upgrades.

I have checked that the new untranslated strings' key match the ones currently in Android, so they'll work automatically when we pull translations.
https://github.com/deltachat/deltachat-android/blob/d41679f2c03f06359a2757793bbda6421a93ddcd/src/main/res/values/strings.xml

As was said, the Core's breaking changes are to be addressed in follow-up MRs.

WofWca:
BTW, I suggest to ignore the CI failure, we can merge with it.

PR #5760, opened 2025-11-24
feat: use confirm instead alert dialog for multi-device mode

nicodh:
see https://github.com/deltachat/deltachat-desktop/pull/5743#pullrequestreview-3492492806 for context

needed a refactor of the switch component to make the visual state controllable from outside

### DIFF
--- a/packages/frontend/src/components/Settings/CoreSettingsSwitch.tsx
+++ b/packages/frontend/src/components/Settings/CoreSettingsSwitch.tsx
@@ -20,7 +20,7 @@ type Props = {
    * Called before the setting is changed. Return false to prevent the change.
    * Can be async to show confirmation dialogs.
    */
-  beforeChange?: (currentValue: boolean) => boolean | Promise<boolean>
+  beforeChange?: (newValue: boolean) => boolean | Promise<boolean>
 }
 
 /*
@@ -57,7 +57,7 @@ export default function CoreSettingsSwitch({
     if (beforeChange) {
       setIsProcessing(true)
       try {
-        const shouldProceed = await beforeChange(storeValue)
+        const shouldProceed = await beforeChange(!storeValue)
         setIsProcessing(false)
         if (!shouldProceed) {
           // user cancelled, don't change the setting
--- a/packages/frontend/src/components/Settings/Advanced.tsx
+++ b/packages/frontend/src/components/Settings/Advanced.tsx
@@ -32,10 +32,10 @@ export default function Advanced({ settingsStore }: Props) {
     })
   }
   const confirmDisableMultiDevice = async (
-    currentValue: boolean // current state (true = enabled)
+    newValue: boolean // (true => multi-device enabled)
   ): Promise<boolean> => {
     // Show a warning when user wants to disable multi-device
-    if (currentValue) {
+    if (!newValue) {
       const confirmed = await confirmDialog(
         openDialog,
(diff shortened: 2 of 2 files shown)


packages/frontend/src/components/Switch/index.tsx
```diff
@@ -9,19 +9,17 @@ type Props = {
...
-  }, [props.checked])
+  // this component just shows the state of props.checked
+  // and triggers the onChange callback when the user toggles the switch
+  const checked = props.checked ?? false
 
   const toggle = (event: React.ChangeEvent<HTMLInputElement>) => {
     const v = event.target.checked
     props.onChange(v)
-    setChecked(v)
+    // No local state update! Parent controls everything
```
WofWca, line 22: These comments are redundant. The code implements a common pattern.

packages/frontend/src/components/Switch/index.tsx
```diff
@@ -9,19 +9,17 @@ type Props = {
   checked?: boolean
   label?: string
   id?: string
-  onChange: (value: boolean) => void
+  onChange: (value: boolean) => void | Promise<void>
```
WofWca, line 12: What's the point of this extra `Promise<void>`? It's unused, so just `void` is fine.

packages/frontend/src/components/Switch/index.tsx
```diff
@@ -9,19 +9,17 @@ type Props = {
   checked?: boolean
```
WofWca: This is probably the right time to also mark `checked` prop as required. I have checked that this will need no other code adjustments, except for the removal of `const checked = props.checked ?? false`.


```suggestion
  checked: boolean
```

packages/frontend/src/components/Settings/CoreSettingsSwitch.tsx
```diff
@@ -16,7 +16,11 @@ type Props = {
   description?: string
   disabled?: boolean
   disabledValue?: boolean
-  callback?: (updatedValue: boolean) => void
+  /**
+   * Called before the setting is changed. Return false to prevent the change.
+   * Can be async to show confirmation dialogs.
+   */
+  beforeChange?: (currentValue: boolean) => boolean | Promise<boolean>
```
WofWca: I think `newValue` would be more intuitive, but it's fine.

packages/frontend/src/components/Settings/ImapFolderHandling.tsx
```diff
@@ -15,41 +15,22 @@ export default function ImapFolderHandling({ settingsStore }: Props) {
...
-    if (!multiDeviceActive) {
-      openAlertDialog({
-        message: tx('pref_multidevice_change_warn'),
-      })
-    }
-  }
-
   return (
     <>
+      <SettingsSeparator />
```
WofWca: The separator should be outside the component, it's easy to do.

packages/frontend/src/components/Settings/CoreSettingsSwitch.tsx
```diff
@@ -28,32 +32,56 @@ export default function CoreSettingsSwitch({
...
+          // user cancelled, don't change the setting
+          return
+        }
+      } catch (_error) {
+        setIsProcessing(false)
+        return
+      }
+    }
+
+    // Change the setting
```
WofWca: These two comments also seem a little redundant

PR #5762, opened 2025-11-24
change: open webxdc apps link in local help

### LINKED ISSUE #5742: say "In-chat apps", not "Webxdc"
the "Help" menu of an opened app currently says "What is Webxdc?" and links to https://webxdc.org

however, "webxdc" is a technical term we do not want to use when facing the avg end user.

- no action required for the menu title: the string as such will be changed by the next translation update to "About Apps", cmp. https://github.com/deltachat/deltachat-android/pull/4025 

- for the link: instead of opening the webxdc.org page that does not answer the question for the avg end user, open the internal offline help at `#webxdc` - we just revised the content; it will look like https://delta.chat/en/help#webxdc on  the next offline help update (but no need to wait, already now, the section exist

nicodh:
resolves #5742


WofWca (approved):
Didn't test, but looks like it should work.

PR #5765, opened 2025-11-24
fix: show different styles for empty bio

### LINKED ISSUE #5749: Settings dialog says "Bio" in my profile preview when no bio is set
I have no bio set in my profile.

The settings dialog just displays the default placeholder "Bio" under my display name:
Image

It makes sense as a placeholder here:
Image

But not in the preview and not in the same font as the actually set bio. It is currently indistinguishable from having `Bio` string set as your bio.

nicodh:
resolves #5749 

Show empty bio in italic font and reduced opacity and make it clickable to edit 

image

### DIFF
--- a/packages/frontend/src/components/Settings/Profile.tsx
+++ b/packages/frontend/src/components/Settings/Profile.tsx
@@ -33,12 +33,13 @@ export default function Profile({ settingsStore, onStatusClick }: Props) {
       className={styles.profile}
       onClick={onStatusClick}
       data-testid='edit-profile-button'
+      title={`${tx('pref_edit_profile')}`}
     >
       <LargeProfileImage
         initials={initials}
         color={settingsStore.selfContact.color}
         imagePath={profileImagePath}
-        disableFullscreen={false}
+        disableFullscreen={true}
       />
       <div className={styles.profileDetails}>
         <div className={styles.profileDisplayName}>{profileName}</div>


r10s:
even though i suggested "italics", it is unclear if that is an improvement overall. it seems better to revert that change [^1] - sorry for the noise, but that was not totally clear to me before

but we can target larger issues that came up, more than one time, in the past in practise:

the click areas in that part are too cluttered. there are (now) two different mouse pointer, the avatar is sometimes enlargeable (if an image is set), sometimes not (if not image is set - but the one also gets not offer to set an image). 

let's just have the whole avatar/name/bio cell a single, large click-area, with the same hover effect as the other menu items.

- this way, the user can tap the avatar and get forward in setting one
- same for name or name/bio (without special placeholder style)
- it is clear to land always in the same dialog
- it would be same as on android/ios
- small disadvantage is viewing ones own avatar requires a second tap - but this is not done often, that's fine

[^1]: the new style raises questions, it does not answer much. only the theoretical, if you set your own bio to literal "Bio" and forgot about it, and want the answer without tapping :) it is not uncommon nowadays to have minimal layout where a placeholder and labels looks like a button etc. and here it seems totally fine. it is not random user input or you can be tricked. even if we have some nice placeholder style - there is not much comparison for the user, they will barely notice.
also wrt consistency with android/ios, and maybe similar places, it is not worth to introduce another style and maybe follow-up work.

nicodh:
@r10s adapted the profile as discussed (except for the smaller image which would be a bit more effort to change)

r10s:
EDIT: the following is not needed, maybe it turns out the currently layout is nice as well, so let's leave it as is for now

<details>
<summary>totally optional suggestion</summary>

tried to do a code-level-suggestion for the avatar size, however, failed :)

still, the avatar seems a little too large to appear homogeneous. the following uses 75% of current width/height.

moreover, the first horizontal line is replaced by some whitespace (the additional whitespace should not enlarge padding, but margin)

before / after:

 &nbsp; &nbsp; 

</details>

nice placeholder color, btw

PR #5768, opened 2025-11-24
fix: pass pointer events through footer

### LINKED ISSUE #5763: style: reactions are shifted to left in message bubble
- Operating System (Linux/Mac/Windows/iOS/Android): all
- Delta Chat Version: main branch

They used to be on the opposite end of the time element.

This is a regression introduced in d73ec257e9d384f950698d0d39a05d6ce091fc7c (https://github.com/deltachat/deltachat-desktop/pull/5745), we missed this.

Image

nicodh:
resolves #5763 

reverts d73ec257e9d384f950698

### DIFF
--- a/packages/frontend/scss/message/_message.scss
+++ b/packages/frontend/scss/message/_message.scss
@@ -211,7 +211,6 @@ $info-text-max-width: 400px;
   .metadata {
     margin-top: 10px;
     margin-bottom: -7px;
-    pointer-events: all;
   }
 
   .module-message__img-attachment {
--- a/packages/frontend/src/components/Reactions/styles.module.scss
+++ b/packages/frontend/src/components/Reactions/styles.module.scss
@@ -4,7 +4,6 @@
   position: relative;
   line-height: 1; // So that it doesn't change the message height.
   top: 16px;
-  pointer-events: all;
 }
 
 .emoji {
--- a/packages/frontend/src/components/message/styles.module.scss
+++ b/packages/frontend/src/components/message/styles.module.scss
@@ -28,7 +28,12 @@
     margin-inline-start: 10px;
     position: absolute;
     inset-inline-end: 5px;
+    /** allow to click through the footer to the message body */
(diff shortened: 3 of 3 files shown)


packages/frontend/src/components/message/styles.module.scss
```diff
@@ -24,9 +24,11 @@
 
   &.onlyMedia {
     bottom: 5px;
+    inset-inline-start: 0;
     margin-inline-start: 10px;
     position: absolute;
     inset-inline-end: 5px;
+    pointer-events: none;
```
WofWca, line 32: I think something like

```scss
pointer-events: none;
> * {
  pointer-events: initial;
}
```

would have been more thorough, and not require the logic being spread across multiple files. Prevent potential future bugs with things being unclickable.

packages/frontend/scss/message/_message.scss
```diff
@@ -211,6 +211,7 @@ $info-text-max-width: 400px;
   .metadata {
     margin-top: 10px;
     margin-bottom: -7px;
+    pointer-events: all;
```
WofWca: `all` seems to be SVG-only.

```suggestion
    pointer-events: initial;
```

packages/frontend/src/components/Reactions/styles.module.scss
```diff
@@ -4,6 +4,7 @@
   position: relative;
   line-height: 1; // So that it doesn't change the message height.
   top: 16px;
+  pointer-events: all;
```
WofWca:
```suggestion
  pointer-events: initial;
```

See above.

packages/frontend/src/components/message/styles.module.scss
WofWca, line 32: A comment would be nice.

nicodh: It didn't work with the direct descendants only condition but works fine without

PR #5770, opened 2025-11-25
Reduce react warnings

nicodh:
Reduces the number of warnings from  37 to 16. Many of the warnings seem inappropriate so just adding a line to disable the rook was often used.

Can be best reviewed commit by commit.

### DIFF
--- a/packages/frontend/src/components/conversations/Timestamp.tsx
+++ b/packages/frontend/src/components/conversations/Timestamp.tsx
@@ -104,11 +104,12 @@ const UpdatingTimestamp = (props: TimestampProps) => {
 const relativeTimeThreshold = 24 * 60 * 60 * 1000 // one day
 
 export default function Timestamp(props: TimestampProps) {
+  const [currentTime] = useState(() => Date.now())
+
   // if older than one week we don't need to update timestamps
   const isOld = useMemo(
-    () => props.timestamp < Date.now() - relativeTimeThreshold,
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    []
+    () => props.timestamp < currentTime - relativeTimeThreshold,
+    [props.timestamp, currentTime]
   )
 
   if (isOld) {


nicodh:
@WofWca we have 7 warnings just from the [ChatlistContext](https://github.com/deltachat/deltachat-desktop/blob/d249fa5e098d4f8d35b8ec9885be4af33652e501/packages/frontend/src/contexts/ChatContext.tsx) file. Maybe you can have a look how to fix/avoid them?

packages/frontend/src/components/conversations/Timestamp.tsx
```diff
@@ -105,7 +105,13 @@ const relativeTimeThreshold = 24 * 60 * 60 * 1000 // one day
 
 export default function Timestamp(props: TimestampProps) {
   // if older than one week we don't need to update timestamps
-  if (props.timestamp < Date.now() - relativeTimeThreshold) {
+  const isOld = useMemo(
+    () => props.timestamp < Date.now() - relativeTimeThreshold,
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
WofWca: Why do we have to disable the rule here? This seems incorrect, this should depend on at least `props.timestamp`.

packages/frontend/src/components/ConnectivityToast.tsx
```diff
@@ -71,6 +75,10 @@ export default function ConnectivityToast() {
     [maybeNetwork]
   )
 
+  // Update ref to enable recursive calls without circular dependency
+  // eslint-disable-next-line react-hooks/refs
+  tryMaybeNetworkIfOfflineAfterXmsRef.current = tryMaybeNetworkIfOfflineAfterXms
+
```
WofWca: What's the point of all this complexity if we have to disable the linter anyway?
My suggestion below should be a little simpler, although the entire function looks like it needs to be rewritten.

```suggestion
  const tryMaybeNetworkIfOfflineAfterXms = useCallback(
    function tryMaybeNetworkIfOfflineAfterXms_(ms: number) {
      setTimeout(() => {
        // This is a hack to get the current network state by abusing the setNetworkState function.
        // Not pretty but works.
        setNetworkState(([connectivity, error]: [Connectivity, string]) => {
          if (connectivity === Connectivity.NOT_CONNECTED) {
            log.debug(
              `We are still not online after ${ms}  milli seconds, try maybeNetwork again`
            )
            maybeNetwork()
          } else if (ms < 30000) {
            tryMaybeNetworkIfOfflineAfterXms_(2 * ms)
          } else {
            log.debug(
              `We tried reconnecting with waiting for more then 30 seconds, now stop`
            )
          }

          // Keep state unchanged
          return [connectivity, error]
        })
      }, ms)
    },
    [maybeNetwork]
  )

```

packages/frontend/src/contexts/RovingTabindex.tsx
```diff
@@ -43,10 +43,11 @@ import React, {
 export function useRovingTabindex(elementRef: RefObject<HTMLElement | null>) {
   const context = useContext(RovingTabindexContext)
 
+  // If the active element has not been chosen yet,
+  // let' keep the default behavior (tabindex="0")
```
WofWca: Why did it have to be moved?

nicodh: unintended

packages/frontend/src/components/conversations/Timestamp.tsx
```diff
@@ -104,8 +104,15 @@ const UpdatingTimestamp = (props: TimestampProps) => {
...
 
 export default function Timestamp(props: TimestampProps) {
+  const [currentTime] = useState(() => Date.now())
+
   // if older than one week we don't need to update timestamps
-  if (props.timestamp < Date.now() - relativeTimeThreshold) {
+  const isOld = useMemo(
+    () => props.timestamp < currentTime - relativeTimeThreshold,
+    [props.timestamp, currentTime]
+  )
```
WofWca, line 113: Utilizing `useState` over `useMemo` for `currentTime` seems to be just a hack to suppress the warning here or am I wrong? Let's better suppress it explicitly.


```suggestion
  // if older than one week we don't need to update timestamps
  const isOld = useMemo(
    // eslint-disable-next-line react-hooks/purity
    () => props.timestamp < Date.now() - relativeTimeThreshold,
    [props.timestamp]
  )
```

packages/frontend/src/components/ConnectivityToast.tsx
nicodh: Feel free to add a commit to this PR

packages/frontend/src/components/conversations/Timestamp.tsx
nicodh: Both variants seem fine to me

PR #5771, opened 2025-11-25
test: withdraw group invite link

nicodh:


packages/e2e-tests/tests/group-tests.spec.ts
```diff
@@ -348,6 +348,70 @@ test('Readd user to group', async () => {
   await page.getByTestId('view-group-dialog-header-close').click()
 })
 
+test('Withdraw group invite link', async ({ browserName }) => {
+  if (browserName.toLowerCase().indexOf('chrom') > -1) {
```
WofWca, line 363:
```suggestion
  if (browserName.toLowerCase().includes('chrom')) {
```

packages/e2e-tests/tests/group-tests.spec.ts
```diff
@@ -348,6 +348,70 @@ test('Readd user to group', async () => {
...
+
+  // check for info message containing "waiting"
+  const lastMessage = page.locator('#message-list li').last()
+  await expect(lastMessage).toContainText(userA.name)
+  await expect(lastMessage).toContainText('Member Me removed', {
+    ignoreCase: true,
+  })
+  // the group chat opens but composer is not available
+  const composer = page.locator('textarea#composer-textarea')
+  await expect(composer).not.toBeVisible({ timeout: 1 })
```
WofWca, line 432: This part of the test isn't doing much is it? It would behave the same as with a non-withdrawn QR, except you'd get re-added to the group a few moments later.
I think we need to wait for a few seconds to confirm that the QR is not working.

nicodh:
The failure in e2e test is really strange. The moment the qr code is scanned and the group chat is opened it should immediately show the invite message but obviously that didn't happen even after 60 seconds...

The error is:
Locator: getByRole('list', { name: 'Messages' }).getByRole('listitem').filter({ hasText: 'Alice invited you to join this group' })
Expected: visible
Timeout: 60000ms
Error: element(s) not found

And the screenshot confirms that:
image
 
But in local tests it always looks like this

image

WofWca:
This also passes for me locally. I also merged `main` into this branch and the tests keep passing.

WofWca:
Sooo, has the issue been fixed with the last commit? What was it about?

nicodh:
> Sooo, has the issue been fixed with the last commit? What was it about?

I just initiated the test again.

I made a rebase earlier this day but the test failed after that again with 2 failing tests...

nicodh:
I will trigger the test one more time to see if it's reliable now

Issue #3989, opened 2024-06-29
clicking deltachat icon in swaybar's system tray offers inconsistent behavior (and can crash app)


dkg:
- OS: Debian stable (12.5) amd64
- Deltachat-desktop version: 1.46.1

I'm running `deltachat-desktop` under [sway](https://github.com/swaywm/sway), with no Xwayland or X11 support on the system at all.  I've launched deltachat by using electron's (chromium's) ozone backend for wayland:

```
deltachat-desktop --ozone-platform=wayland
```

When delta first starts up, it is *only* rendered as an icon in the system tray section of swaybar.  If i click that icon, the full deltachat interface appears.

If i click it again, the deltachat interface disappears.

But if i click it a third time, deltachat closes entirely (the app appears to crash), emitting the following message to stdout or stderr:

```
[3179:0628/212044.664566:FATAL:wayland_toplevel_window.cc(1123)] Should not be called with kMinimized state
Trace/breakpoint trap
```

When the app is running but minimized, right-clicking the icon in the system tray creates a menu with three items:

- Open DeltaChat
- Minimize
- Quit

if i choose "Open DeltaChat", that also causes the crash described above.  "Quit" appears to close the app cleanly.

Note that the menu doesn't interact well with sway -- it wants to consume the entire screen, and it doesn't share the screen at all well with the main app itself.

I'm happy to test or diagnose things if that would be useful.

farooqkz:
As a workaround for this issue, use DeltaChat with Xwayland. I've been using it for sometime and it's fine.

hpk42:
Well, dkg was describing a system setup that explicitely not uses wayland
so recommending to use wayland does not help much i am afraid.

farooqkz:
@Simon-Laux There are other reports which specificly target wayland on my side. Any on your side?

farooqkz:
> Well, dkg was describing a system setup that explicitely not uses wayland so recommending to use wayland does not help much i am afraid.
> […](#)
> On Mon, Jul 01, 2024 at 05:23 -0700, Farooq Karimi Zadeh wrote: As a workaround for this issue, use DeltaChat with Xwayland. I've been using it for sometime and it's fine. -- Reply to this email directly or view it on GitHub: [#3989 (comment)](https://github.com/deltachat/deltachat-desktop/issues/3989#issuecomment-2200004161) You are receiving this because you are subscribed to this thread. Message ID: ***@***.***>

[...]

PR #5774, opened 2025-11-26
fix: do not minimize the window before hiding it

link2xt:
Minimizing the window with mainWindow.window.minimize() before hiding it with mainWindow.window.hide()
results in a crash when mainWindow.window.show()
is called later when running on Wayland:
`[3179:0628/212044.664566:FATAL:wayland_toplevel_window.cc(1123)] Should not be called with kMinimized state`

I have tested the fix in `labwc` by clicking the icon to hide and show the window multiple times,
as well as hiding the window manually by clicking
on the window name in the taskbar (wlr/taskbar module of waybar) and then restoring it via the tray icon.
This also fixed the problem when running `niri`,
I cannot reproduce the crash with it as well.

Closes <https://github.com/deltachat/deltachat-desktop/issues/3989>

### DIFF
--- a/packages/target-electron/src/tray.ts
+++ b/packages/target-electron/src/tray.ts
@@ -55,12 +55,9 @@ function mainWindowIsVisible() {
   return mainWindow.window.isVisible() && mainWindow.window.isFocused()
 }
 
-export function hideDeltaChat(minimize?: boolean) {
+export function hideDeltaChat() {
   if (!mainWindow.window) {
     throw new Error('window does not exist, this should never happen')
-  }
-  if (minimize === true) {
-    mainWindow.window.minimize()
   }
   mainWindow.window.hide()
   if (process.platform === 'linux') tray?.setContextMenu(getTrayMenu() as Menu)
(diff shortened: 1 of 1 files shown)


link2xt:
The code seems to have been introduced in 768d2222a7ee30100388d5d8996fe6b267b8bccf, don't know if there is any reason for calling `minimize`.

link2xt:
Likely needs to be tested on macOS, maybe this messes up with the dock. I only tested on Linux.

WofWca:
> don't know if there is any reason for calling minimize.

I suspect that the reason for this was the same as in 1dd9409866027d4c425bd2ef31c9557d31585843, i.e. to reflect the new state faster, probably so that rapidly clicking the tray icon also rapidly shows / hides the window.

> Likely needs to be tested on macOS, maybe this messes up with the dock.

[Electron docs on `win.minimize()`](https://www.electronjs.org/docs/latest/api/browser-window#winminimize) state:

> On some platforms the minimized window will be shown in the Dock.

so yep, maybe.

I have checked that we do not check `isMinimized()`, and do not use `restore()`, and also read the Electron docs and they do not mention minimizing the window affecting other properties of `window` that we rely on.

Semantically I also don't think that there is necessity to minimize the window on top of hiding it.

I have tested this on Windows and found no issues that are caused by this MR, although maybe I'm just not sure what to look for. Clicking on notifications show the window regardless of whether it is focused / minimized / hidden. Clicking on tray icon toggles between hidden and visible.

When someone tests this on macOS, I think it's good to merge.
Thanks, good work!

PR #5775, opened 2025-11-27
build: limit esbuild to 0.25.x

nicodh:
esbuild-plugin-inline-worker has a dependency to esbuild@latest
which is 0.27.0 now and requires MacOS X 12.x!

Since the plugin was not updated since last release 4 years ago it should be safe to restrict it to esbuild 0.25.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,8 @@
       "esbuild"
     ],
     "overrides": {
-      "stylus": "npm:empty-npm-package@1.0.0"
+      "stylus": "npm:empty-npm-package@1.0.0",
+      "esbuild-plugin-inline-worker>esbuild": "^0.25.0"
     }
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


WofWca:
I'm not sure if I understand the point of this.

> which is 0.27.0 now and requires MacOS X 12.x

Is it that we need to support building on older macOS?

Also can't we avoid specifying the esbuild version twice in our project? e.g. with catalog

https://github.com/deltachat/deltachat-desktop/blob/f24f788f54f5c0de13c481fe4ef67fa68248e16d/packages/target-electron/package.json#L87

nicodh:
>> Is it that we need to support building on older macOS?

Not neccessarily*, but imo it doesn't make sense to exclude support for nothing. If we decide one day we need esbuild 0.27 we could do so, but now it's just because a plugin that had no updates since 4 years has a dependency defined as "latest".

*I have a Macbook which is limited to macOS 12 and it still works very well with all other stuff...

> Also can't we avoid specifying the esbuild version twice in our project? e.g. with catalog

I tried that, but it seems that pnpm does not support catalog references in the override section.

Issue #5773, opened 2025-11-26
cleanup: remove outdated options and experiments


r10s:
after internal discussions, the following options  should be removed alltogether:

- "Settings / Advanced / Encryption":  
the whole section incl. "Export Keys", "Send Autocrypt Setup Message", "Learn More".  
this continues confusing user and/or raises false expectations - for both, avg user and power-user with some easily outdated knowledge.  
if expert user really want to inspect their keys, we updated the faq at https://github.com/deltachat/deltachat-pages/pull/1209

- "Settings Advanced / Experimental / Related Chats":
this option is of limited benefits, raises questions, often standing in the way of other improvements and raising additional maintenance effort. we probably also do not want to have that in the UI.

- "Settings Advanced / Experimental / Chat Audit Log":
this option is of limited benefits, and raises questions. we probably also do not want to have that in the UI.

code and feature cleanup is needed in general to keep us focused. and spending resources on things with more impact.

but it was good to have these experiments, to know what we finally do not want.

WofWca:
Do we need to completely nuke the features, right now? Or would removing the UI be enough? Then remove the code itself when it causes problems.
I heard @Simon-Laux has been doing some work on the "Chat Audit Log" feature recently?

If yes, will Android take care of removing the strings?

https://github.com/deltachat/deltachat-desktop/blob/9084e8968038d207529b2f2881972db0e62a5505/_locales/en.xml#L507-L510

Simon-Laux:
> I heard @Simon-Laux has been doing some work on the "Chat Audit Log" feature recently?

doesn't matter, it is not really used, delta chat does not have big groups with many messages and moderation, that's where an audit log would be actually useful.

It would be easy to bring back should we have a better concept for it. Maybe I will revive it in my fork when I fork, who knows. For now I agree with bjoern that we should keep the official version on point.

I never really used both features much, maybe because I don't really use the clone group feature, which I guess "related chats" could be useful for.

Simon-Laux:
> Do we need to completely nuke the features, right now?

There are still saved on git and currently the core APIs for them still exist, and even if we would remove the core APIs they are trivial to rebuild:
- get related chats is basically an SQL query 
- chat audit dialog is a message-list filter to only show info messages[^1]


[^1]: if you don't care about efficiency then you could even get the normal list and filter it in the UI code

nicodh:
[...]

PR #5776, opened 2025-11-27
Merge three dot menues

### LINKED ISSUE #5325: How is "Three dot menu" different from "Chat list item context menu"?
- Delta Chat Version: 1.60.1

`ThreeDotMenu` and `ChatListContextMenu` are almost identical.

https://github.com/deltachat/deltachat-desktop/blob/5a8b3fcd72bce930a44492c5f4ce564afe10d379/packages/frontend/src/components/ThreeDotMenu.tsx#L66-L201

https://github.com/deltachat/deltachat-desktop/blob/af7e222a87b9a200a6d0c1d8cdd49f2c23dd343a/packages/frontend/src/components/chat/ChatListContextMenu.tsx#L166-L317

But not completely. Some options that are present in one are missing in another. For example, "Search in chat" and "Ephemeral messages" are only available in `ThreeDotMenu`, "Clone chat", "View chat", "Clear chat", "Encryption info" are only available in `ChatListContextMenu`.
What this results in:
- You have to figure out which menu you need to open to get to the option you need
- Behavior inconsistency. We, as part of the core v2 migration, recently updated the `ChatListContextMenu`, but did not update `ThreeDotMenu`.

Can we just keep `ChatListContextMenu` and remove `ThreeDotMenu`? Or remove the duplicate items? This would not be good on small screens though, where the chat list is not always visible, as on Android.
At the very least we should them reuse the same code, to avoid inconsistencies.

Just in case, keep in mind that chat list item multiselect is coming: #5297.

WDYT?

### LINKED ISSUE #5769: Cannot leave the channel from 3-dot menu
I am running the latest `main` version.

Inside the channel I joined I cannot leave it from 3-dot menu:
Image

I can right-click it in the chatlist and leave there, but I expected an option in 3-dot menu as well, similarly to the "Leave group" in group chats.

Also not clear what does "Block contact" means here. It is a channel, not 1:1 chat with a contact.

nicodh:
resolves #5325, resolves #5769

TBD:

1. should we show the Encryption info item in chat list at all?
2. the generic "dangerColor" seems a bit inappropriate (too pale) - do we need 2 different colors for danger & error?
3. I moved profile to "more options" in 3dot menu since there is the direct link to the chat header which is much more convenient

Current state:
Main difference between both menues: chatlist has "pin", main chat has "search" and "more options" 

```
context menu:            three-dot-menu menu:


                         search in chat
                         mute notifications
                         --------
pin chat                     
mute notifications       archive chat
archive chat             disappearing messages
--------                 --------
profile
clone chat **
--------                 
block contact *          block contact *
leave group **           leave group/channel
                         clear chat
delete chat              delete chat
encryption info          more options:
                                 profile
                                 clear chat
                                 encryption info
                                 chat audit log



*one-to-one only
**groups only
```

image

Note: I will rename the file and the class ChatListContextMenu => ChatContextMenu but after being approved for better reviewing

### DIFF
--- a/packages/frontend/src/components/chat/ChatContextMenu.tsx
+++ b/packages/frontend/src/components/chat/ChatContextMenu.tsx
@@ -120,7 +120,7 @@ function buildArchiveAndPinMenuItems(
 
   /*      Archive  UnArchive  Pin  UnPin
   pinned      y       n       n      y
-  archived    n       y       y      n
+  archived    n       y       n      n
   normal      y       n       y      n
   */
 
@@ -401,11 +401,11 @@ export function useChatContextMenu(): {
     }
 
     const onDisappearingMessages = () => {
-      if (activeChatId === null) {
+      if (relatedChat === undefined) {
         return
       }
       openDialog(DisappearingMessages, {
-        chatId: activeChatId,
+        chatId: relatedChat.id,
       })
     }
 
@@ -536,9 +536,11 @@ export function useChatContextMenu(): {
     await openContextMenu({
       ...mouseEventToPosition(event),
       items: menu,
-      ariaAttrs: {
-        'aria-label': tx('chat_list_item_menu_label'),
-      },
+      ariaAttrs: isMainView
+        ? { 'aria-labelledby': 'three-dot-menu-button' }
+        : {
+            'aria-label': tx('chat_list_item_menu_label'),
+          },
     })
     setActiveContextMenuChatIds([])
   }


Simon-Laux:
> 1. should we show the Encryption info item in chat list at all?

may not even be useful for groups anymore with key contacts? there are no opportunisitc groups anymore, are there?
And for contacts you can already/still access it from the profile view.

r10s:
> 1. should we show the Encryption info item in chat list at all?

"encryption info" is mostly useful for one-to-ones, that is even mentioned in [advanced faq](https://delta.chat/en/help#how-can-i-manually-check-encryption-information).

"group encryption info" is more a debug thing, it does not make sense to upgrade that to three-dot-menu, it would better be placed in group profile as well - but as we do not have the three-dot-menu there atm, let's leave that in the chatlist context menu for now (we anyway stay with two menus) or remove it (if we only have one menu)

there is the idea to have a "three dot menu" also in group profiles

> 2. the generic "dangerColor" seems a bit inappropriate (too pale) - do we need 2 different colors for danger & error?

let's go for `#ff0c16` for "red danger or error text", this is what we're using on android as well. it works nicely on light and dark mode

but i'd call the color `dangerTextColor` - it must not be used for red background and white text (as eg. in the buttons)

Screenshot 2025-12-01 at 18 36 29

r10s:
for the "more options" - that can also be flattened, there will stay only "clear chat" after https://github.com/deltachat/deltachat-desktop/issues/5773

this also looks better as the "red options" are at the end then ("Clear" and "Delete")

packages/frontend/themes/_themebase.scss
```diff
@@ -4,7 +4,7 @@ $hover-contrast-change: 2%;
 // set light theme variables as fallback
 $themeDarkOrLight: light !default;
 $colorPrimary: #42a5f5 !default;
-$colorDanger: #f96856 !default;
+$colorDanger: #ff0c16 !default;
```
WofWca, line 7: The contrast in fact is not satisfactory on either of the themes (dark or light), especially with the "hover highlight".

Image

To fix that we should use different colors for dark and light theme, as @nicodh said.
This also goes for `$btnDangerBackground` BTW.
Maybe this is for another MR though.

packages/e2e-tests/tests/basic-tests.spec.ts
```diff
@@ -99,10 +99,24 @@ test('start chat with user', async ({ browserName }) => {
...
-  await expect(
-    page.locator('.chat-list .chat-list-item').filter({ hasText: userA.name })
-  ).toHaveCount(1)
-  console.log(`Chat with ${userA.name} created!`)
+  const chatListItem = page
+    .locator('.chat-list .chat-list-item')
+    .filter({ hasText: userA.name })
+    .first()
+  await expect(chatListItem).toBeVisible()
+  // check for context menu items
```
WofWca:
```suggestion

  // check for context menu items
```

It's somewhat of a separate test, so let's separate it with a new line at least.

packages/e2e-tests/tests/group-tests.spec.ts
```diff
@@ -138,6 +138,20 @@ test('create group', async ({ browserName }) => {
     .getByTestId(`account-item-${userB.id}`)
     .locator('.styles_module_accountBadgeIcon')
   await expect(badgeNumber).toHaveText('1')
+  // check for context menu items
```
WofWca:
```suggestion

  // check for context menu items
```

packages/e2e-tests/tests/basic-tests.spec.ts
```diff
@@ -99,10 +99,24 @@ test('start chat with user', async ({ browserName }) => {
...
+  await chatListItem.click({
+    button: 'right',
+  })
+  await expect(page.getByRole('menu').getByRole('menuitem')).toHaveText([
```
WofWca: How about adding tests for the "three-dot menu" as well while we're at it?

(BTW note that there is another place which explicitly tests context menu:)

https://github.com/deltachat/deltachat-desktop/blob/17ba3b04f6314c704ac53f525ef52d4dab910946/packages/e2e-tests/tests/chat-list-multiselect.spec.ts#L193-L210

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -0,0 +1,463 @@
...
+        label: tx('menu_delete_chat'),
+        action: onDeleteChats,
+        danger: true,
+      },
+      isMainView && {
+        label: tx('clear_chat'),
+        action: onClearChat,
+        danger: true,
+      },
+      !isMainView && encryptionInfoItem,
```
WofWca, line 530: TBH this looks really odd after the "danger" buttons:

Image

Maybe move it up? Right along "view profile / group".

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -0,0 +1,463 @@
...
+      {
+        label: tx('menu_delete_chat'),
+        action: onDeleteChats,
+        danger: true,
+      },
+      isMainView && {
+        label: tx('clear_chat'),
+        action: onClearChat,
+        danger: true,
+      },
```
WofWca:
```suggestion
      isMainView && {
        label: tx('clear_chat'),
        action: onClearChat,
        danger: true,
      },
      {
        label: tx('menu_delete_chat'),
        action: onDeleteChats,
        danger: true,
      },
```

Let's swap the two. Makes sense, and is in line with Android and Telegram.

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -0,0 +1,463 @@
...
+          label: tx('menu_leave_channel'),
+          action: () => onLeaveGroupOrChannel(singleChat),
+          danger: true,
+        },
+      showLeaveGroupOption && {
+        label: tx('menu_leave_group'),
+        action: () => onLeaveGroupOrChannel(singleChat),
+        danger: true,
+      },
+      ephemeralMessagesMenuItem,
```
WofWca: How about moving this to just below "mute chat"?

packages/e2e-tests/tests/group-tests.spec.ts
```diff
@@ -138,6 +138,20 @@ test('create group', async ({ browserName }) => {
...
+  await chatListItem.click({
+    button: 'right',
+  })
+  await expect(page.getByRole('menu').getByRole('menuitem')).toHaveText([
+    'Pin Chat',
+    'Mute Notifications',
+    'Archive Chat',
+    'View Profile',
```
WofWca:
```suggestion
    'Edit Group',
```

`menu_edit_group` string is unused in this version of the MR. I think we shouldn't change this to "View Profile".

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -0,0 +1,463 @@
...
+
+  const selfInGroup = isGroup && singleChat.isSelfInGroup
+
+  return [
+    // View Profile (for single chats)
+    !isGroup &&
+      !singleChat.isDeviceTalk && {
+        label: tx('menu_view_profile'),
+        action: onViewProfile,
+      },
```
WofWca: This is also shown for channels, and the button just opens the first recipient's profile.

I think it's best to avoid using `isGroup` with `!`, and it's best to use `chatType` directly.

Edit: OK, never mind, this bug is also present on `main`.

packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
```diff
@@ -487,7 +487,17 @@ function ChatNavButtons({
...
 }) {
   const tx = useTranslationFunction()
-  const onClickThreeDotMenu = useThreeDotMenu(chat)
+  const { openContextMenu } = useChatContextMenu()
+  const onClickThreeDotMenu = useCallback(
+    (event: React.MouseEvent) => {
+      const activeChat = {
+        ...chat,
+        kind: 'ChatListItem',
+      } as unknown as T.ChatListItemFetchResult & { kind: 'ChatListItem' }
```
WofWca: Is it hard to avoid a type cast here?

packages/e2e-tests/tests/group-tests.spec.ts
nicodh: That was explicitly requested by @r10s

nicodh:
I applied some of the suggestions concerning the order of items. And introduced a new (reduced) chat list type which the full chat is converted to.

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -0,0 +1,508 @@
...
+    chatType: chat.chatType,
+    isPinned: chat.pinned,
+    isArchived: chat.archived,
+    isMuted: chat.isMuted,
+    isDeviceTalk: chat.isDeviceChat,
+    isSelfTalk: chat.isSelfTalk,
+    isSelfInGroup: chat.selfInGroup,
+    isContactRequest: chat.isContactRequest,
+    isEncrypted: chat.isEncrypted,
+    dmChatContact:
```
WofWca, line 64:
```suggestion
    // https://github.com/chatmail/core/blob/a3328ea2de1e675b1418b4e2ca0c23f88828c558/deltachat-jsonrpc/src/api/types/chat_list.rs#L130-L146
    dmChatContact:
```

packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
```diff
@@ -487,7 +487,13 @@ function ChatNavButtons({
   lastUsedApps: T.Message[]
 }) {
   const tx = useTranslationFunction()
-  const onClickThreeDotMenu = useThreeDotMenu(chat)
+  const { openContextMenu } = useChatContextMenu()
+  const onClickThreeDotMenu = useCallback(
+    (event: React.MouseEvent) => {
+      openContextMenu(event, [], chat.id, chat)
```
WofWca: Soo, you can pass an empty array? I think `null` would be more explicit. Plus I think this behavior needs a bit of documentation, i.e. when it's OK to pass `null` / `[]`.

WofWca: Plus I think this is the cause of the bug where we always show "unmute chat" and "unarchive chat" even if it's not muted and is archived, but I'm not sure.

Image

nicodh: Should be fixed now

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -0,0 +1,516 @@
...
+    const isMainView = !!selectedChat
+    const singleChat: ChatListItem | false = isMainView
+      ? fullChatToChatListItem(selectedChat)
+      : chatListItems.length === 1
+        ? chatListItems[0]
+        : false
+
+    if (chatListItems.length === 0 && singleChat) {
+      chatListItems.push(singleChat)
+    }
```
WofWca: I think mutating arguments will cause a bug at some point, unless there is a bug already.

nicodh:
I changed the order of items, since I removed pin from archived chats which seemed useless imo. But that resulted in always showing mute menu item in first place on archived items, where it would be nicer to have unarchive as first item.

Tests needed to be adapted.

packages/frontend/src/components/chat/ChatContextMenu.tsx
nicodh: Hmm, maybe you are right. Might be confusing. I will fix that.

nicodh: I refactored the implementation. To avoid too much testing when reviewing I created e2e tests to make sure that the changes have the expected output

packages/frontend/themes/_themebase.scss
nicodh: I would prefer to sole that in another MR. This one is already big enough...

WofWca: https://github.com/deltachat/deltachat-desktop/issues/5911

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -0,0 +1,525 @@
...
+  const tx = useTranslationFunction()
+  const [activeContextMenuChatIds, setActiveContextMenuChatIds] = useState<
+    number[]
+  >([])
+
+  /**
+   * @param chatListItems array of selected chat list items (empty if opened from main chat view)
+   * @param selectedChatId id of the currently selected chat (null if none selected)
+   * @param selectedChat the currently active chat (only in main view)
+   */
```
WofWca, line 318: Moving this doc string makes some IDEs not show the docs for `openContextMenu`.

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -0,0 +1,525 @@
...
+): (ContextMenuItem | false)[] {
+  if (!singleChat) return []
+
+  const onViewProfile = async () => {
+    const fullChat =
+      selectedChat ??
+      (await BackendRemote.rpc.getFullChatById(accountId, singleChat.id))
+    if (!fullChat) {
+      throw new Error('chat was not found')
+    }
```
WofWca:
```suggestion
```

We don't expect this to happen.

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -0,0 +1,525 @@
...
+    openViewGroupDialog(
+      fullChat as T.FullChat & { chatType: 'Group' | 'OutBroadcast' }
+    )
+  }
+
+  const selfInGroup =
+    singleChat.chatType === 'Group' && singleChat.isSelfInGroup
+
+  return [
+    // View Profile (for single chats)
```
WofWca:
```suggestion
```

no longer for "single chats".

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -0,0 +1,525 @@
...
+  const { unselectChat } = useChat()
+  const tx = useTranslationFunction()
+  const [activeContextMenuChatIds, setActiveContextMenuChatIds] = useState<
+    number[]
+  >([])
+
+  /**
+   * @param chatListItems array of selected chat list items (empty if opened from main chat view)
+   * @param selectedChatId id of the currently selected chat (null if none selected)
+   * @param selectedChat the currently active chat (only in main view)
```
WofWca: These are really confusing. All three seem to mean the same thing just looking at their docs.

For "selected" as in "the currently opened" chat I switched the terminology to `activeChat`, 5c664bf98a62ef217b33f5de1cdcc3152316f3f8. Maybe doing the same would help here.

Is it that `selectedChat` must _not_ be provided if it's not the main view? What happens if someone does that accidentally?

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -0,0 +1,525 @@
...
+    }
+    // true if opened from main chat view (not from list)
+    const isMainView = !!selectedChat
+
+    const multipleChatsSelected = chatListItems.length > 1
+
+    // Determine use case:
+    // 1. multiple selected chats in chat list (multipleChatsSelected = true, singleChat = undefined)
+    // 2. single selected chat in chat list
+    // 3. no selected chat in chat list, but main view chat
```
WofWca: What's "Determine use case" about?

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -0,0 +1,525 @@
...
+function buildViewEditMenuItems(
+  singleChat: ChatListItem | undefined,
+  selectedChat: T.FullChat | undefined,
+  openDialog: ReturnType<typeof useDialog>['openDialog'],
+  openViewGroupDialog: ReturnType<typeof useOpenViewGroupDialog>,
+  openViewProfileDialog: ReturnType<typeof useOpenViewProfileDialog>,
+  accountId: number,
+  tx: ReturnType<typeof useTranslationFunction>
+): (ContextMenuItem | false)[] {
+  if (!singleChat) return []
```
WofWca: Looks like this condition should be outside of this function.

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -0,0 +1,525 @@
...
+      isMainView && {
+        label: tx('search_in_chat'),
+        action: onSearchInChat,
+      },
+      !isMainView && pin,
+      archive,
+      ephemeralMessagesMenuItem,
+      muteMenuItem,
+      { type: 'separator' },
+      ...(!isMainView ? viewEditMenuItems : []),
```
WofWca, line 489:
```suggestion
      ...(!isMainView && !multipleChatsSelected
        ? buildViewEditMenuItems(
            singleChat,
            selectedChat,
            openDialog,
            openViewGroupDialog,
            openViewProfileDialog,
            accountId,
            tx
          )
        : []),
```

Let's not spread related the logic far apart.

The variable will need to be removed.

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -0,0 +1,525 @@
...
+    }
+  }
+}
+
+/**
+ * Builds view/edit profile menu items for chats
+ */
+function buildViewEditMenuItems(
+  singleChat: ChatListItem | undefined,
+  selectedChat: T.FullChat | undefined,
```
WofWca: FYI where this function is used, `selectedChat` is always `undefined` according to TypeScript.
I'm not sure if this means that we should remove this parameter.

WofWca: As with `openContextMenuHandler`, these are hard to understand. For example, nothing is saying that these even have to be the same chat (with the same `id`).

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -0,0 +1,525 @@
...
+    }
+    if (
+      fullChat.chatType !== 'InBroadcast' &&
+      fullChat.chatType !== 'Mailinglist'
+    ) {
+      openViewProfileDialog(accountId, fullChat.contactIds[0])
+    } else {
+      openDialog(MailingListProfile, {
+        chat: fullChat as T.FullChat & {
+          chatType: 'InBroadcast' | 'Mailinglist'
```
WofWca:
```suggestion
          chatType: typeof fullChat.chatType
```

For type safety. Would also be nice to do this in `onViewGroup`.

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -0,0 +1,525 @@
...
+  chats: ChatListItem[],
+  accountId: number,
+  tx: ReturnType<typeof useTranslationFunction>
+): ContextMenuItem {
+  const batchAction = (fn: (chat: ChatListItem) => Promise<unknown>) => () =>
+    Promise.all(chats.map(chat => fn(chat)))
+
+  const hasMutedChats = chats.some(item => !item.isMuted)
+
+  if (hasMutedChats) {
```
WofWca:
```suggestion
  const hasUnmutedChats = chats.some(item => !item.isMuted)

  if (hasUnmutedChats) {
```

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -0,0 +1,525 @@
...
+    // Build the complete menu
+    const menu: (ContextMenuItem | false)[] = [
+      isMainView && {
+        label: tx('search_in_chat'),
+        action: onSearchInChat,
+      },
+      !isMainView && pin,
+      archive,
+      ephemeralMessagesMenuItem,
+      muteMenuItem,
```
WofWca, line 487:
```suggestion
      buildMuteMenuItem(chatListItems, accountId, tx),
```

No extra indirection.

packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
WofWca: No, still the same. `buildMuteMenuItem` is provided an empty array. The same with "Archive", `buildArchiveAndPinMenuItems`. And also thus the buttons do nothing.

WofWca: Continuing in https://github.com/deltachat/deltachat-desktop/pull/5776#discussion_r2678477531

WofWca:
Ooops, sorry I must have checked out the old branch? Let me re-test.

packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
WofWca: OK, sorry, this has been resolved. I tested it on the old branch state.

packages/e2e-tests/tests/group-tests.spec.ts
WofWca, line 337: Is this intended to be in this MR?

packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
nicodh: I see it correct?

Muted & Archived
Image

Unmuted & Archived

Image

packages/frontend/src/components/chat/ChatContextMenu.tsx
WofWca: So you decided not to fix this issue?

packages/e2e-tests/tests/group-tests.spec.ts
nicodh: This is NOT in the current PR. see https://github.com/deltachat/deltachat-desktop/blob/main/packages/e2e-tests/tests/group-tests.spec.ts#L364C1-L364C35

It seems github has problems syncing with git... ??

Do you see this in git diff main locally?

WofWca: Yes. it's part of 08e88fff0e66f0ec023c70bf66b307a490e81d5e. see `git diff HEAD~19 -- packages/e2e-tests/tests/group-tests.spec.ts`

WofWca: It's because that commit is already merged into main, but it's also included in this branch. This branch is not based on that commit.

nicodh: But this diff should show "files changed", like when running `git diff main`

Otherwise reviewing is quite hard...

WofWca: It shows what `git diff main...origin/merge-three-dot-menues` (with three dots) would show, which I think is what one would want, because this MR does indeed contain a commit that introduced this test.

packages/frontend/src/components/chat/ChatContextMenu.tsx
nicodh: >  All three seem to mean the same thing just looking at their docs.

No, ChatListItem has different props as FullChat. It would be nice if ChatListItem would just be a subtype of Fullchat but that has to be changed in the JSON RPC Api.

> What happens if someone does that accidentally

Well that's why there is a documented parameter. If that disturbs you we could only export wrapper functions with more distinct names and non ambigous parameter like:

openChatContextMenuInMainView
openChatContextMenuInChatList

WofWca: Sounds like an option. Or maybe instead making the function take a union type object (`{ ... } | { ... }`), where it's clear which of the two the arguments is, and which properties actually must be present.

packages/frontend/src/components/chat/ChatContextMenu.tsx
nicodh: I started a separate refactoring PR #5929 since this PR is already too overloaded.

Btw.: when refactoring I found that I the wording "selectedChatId" in chatlist context menu is really misleading. It is NOT the selected chatlist item that received the right click to open the context menu which would be expected but the active chat in main view.

I fixed that too in the refactoring PR

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -0,0 +1,559 @@
...
+  openViewGroupDialog: ReturnType<typeof useOpenViewGroupDialog>,
+  openViewProfileDialog: ReturnType<typeof useOpenViewProfileDialog>,
+  accountId: number,
+  tx: ReturnType<typeof useTranslationFunction>
+): (ContextMenuItem | false)[] {
+  const onViewProfile = async () => {
+    const fullChat = await BackendRemote.rpc.getFullChatById(
+      accountId,
+      relatedChat.id
+    )
```
WofWca: Would be nice to have a comment about why we need to `getFullChatById` despite already having it (`relatedChat: T.FullChat`).

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -0,0 +1,559 @@
...
+      })
+    }
+
+    const onSearchInChat = () => {
+      window.__chatlistSetSearch?.('', activeChatId)
+      setTimeout(
+        () => ActionEmitter.emitAction(KeybindAction.ChatList_FocusSearchInput),
+        0
+      )
+    }
```
WofWca, line 418: Using `activeChatId` here is sematically incorrect. If one day we decide to show this menu item for chat list items, this will cause a bug.

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -0,0 +1,559 @@
...
+
+    event.preventDefault()
+
+    setActiveContextMenuChatIds(chatListItems.map(item => item.id))
+    await openContextMenu({
+      ...mouseEventToPosition(event),
+      items: menu,
+      ariaAttrs: {
+        'aria-label': tx('chat_list_item_menu_label'),
+      },
```
WofWca:
```suggestion
      ariaAttrs: isMainView
        ? { 'aria-labelledby': 'three-dot-menu-button' }
        : {
            'aria-label': tx('chat_list_item_menu_label'),
          },
```

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -0,0 +1,559 @@
...
+  pinned      y       n       n      y
+  archived    n       y       y      n
+  normal      y       n       y      n
+  */
+
+  // Determine which menu items to show based on chat states
+  if (chats.every(chat => chat.isPinned)) {
+    return { pin: unPin, archive }
+  } else if (chats.every(chat => chat.isArchived)) {
+    return { pin: null, archive: unArchive }
```
WofWca, line 131: Is there an explanation for this change getting smuggled in? This used to be `{ pin, archive: unArchive }`.
Also this is now out of sync with the table above.

nicodh: Showing Pin chat for an archived chat results in silently unArchiving the chat which seems wrong to me. Either I want to have a chat archived OR I want to have it pinned. So it should not be possible to pin an archived chat

nicodh: Yes the comment table needs to be updated then

packages/frontend/src/components/chat/ChatContextMenu.tsx
nicodh: I changed it

packages/frontend/src/components/chat/ChatContextMenu.tsx
WofWca: > results in silently unArchiving the chat which seems wrong to me.

To me it's fine. What would be expected? That it gets pinned inside the archive? FTR in Telegram yes, there is a way to pin chats inside the archive.

If that's the case we'll probably also need MRs / issues for Android to remove the "pin" button.

Issue #5715, opened 2025-11-12
"Esc" when editing a message should cancel the edit


dkg:
This is about the default keybindings.

In DeltaChat 2.22.0, when the compose window is active, and I'm editing a message, hitting "Esc" should behave the same as clicking the X in the "Edit Message" pane above the compose window.

WofWca:
We also utilize Esc to cancel the quote, so we'll need to some logic there.

Edit: actually, it's not possible to have both the quote and the "editing" mode to be present in the composer, so there is not gonna be a lot of logic. Just "if in editing mode, exit edit mode. if in regular mode, remove quote".

WofWca:
This seems to have been addressed in https://github.com/deltachat/deltachat-desktop/pull/5713. Thanks, @trufae!

https://github.com/deltachat/deltachat-desktop/blob/af5fcc8465bd5678f46099e6cc4a638a59434441/packages/frontend/src/components/composer/Composer.tsx#L326-L331

PR #5783, opened 2025-11-28
fix: don't change reply when editing message

WofWca:
- **refactor: refer to composer by class name**
- **fix: don't change reply when editing message**

Previously you could select and remove the reply with
`Ctrl + ArrowUp` / `ArrowDown` and `Escape`.
`Escape` is especially problematic because it collides with
"exit message editing mode". See
https://github.com/deltachat/deltachat-desktop/issues/5715.

I have checked that "Escape" no longer cancels reply when in editing mode.

### DIFF
--- a/packages/frontend/src/components/composer/ComposerMessageInput.tsx
+++ b/packages/frontend/src/components/composer/ComposerMessageInput.tsx
@@ -242,7 +242,11 @@ export default class ComposerMessageInput extends React.Component<
                 : '')
             }
             style={{ '--maxLines': maxLines } as React.CSSProperties}
-            id='composer-textarea'
+            id={
+              this.props.isMessageEditingMode
+                ? 'composer-textarea-edit'
+                : 'composer-textarea-non-edit'
+            }
             ref={this.textareaRef}
             rows={1}
             // intent={this.state.error ? 'danger' : 'none'}
--- a/packages/frontend/src/keybindings.ts
+++ b/packages/frontend/src/keybindings.ts
@@ -121,14 +121,14 @@ export function keyDownEvent2Action(
       ev.code === 'ArrowUp' &&
       (ev.ctrlKey || ev.metaKey) &&
       !(ev.ctrlKey && ev.metaKey) && // Both at the same time
-      (ev.target as HTMLElement)?.id === 'composer-textarea'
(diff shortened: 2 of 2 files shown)


packages/frontend/src/components/composer/ComposerMessageInput.tsx
```diff
@@ -236,13 +236,17 @@ export default class ComposerMessageInput extends React.Component<
         {({ writingDirection }) => (
           <textarea
             className={
-              'message-input-area' +
+              'create-or-edit-message-input' +
```
nicodh, line 239: is class name really the best choice? I would say it is unexpected that it is not only used for css but also as a selector in MessageList.
Did you consider using data-role, data-element or just data-identifier to avoid that ambiguity?

WofWca: No, I did not consider that.

That would probably be better.
But IDK if it's that much better. There is no fundamental difference between `data` and `class`.

I'll come back to this.

nicodh: > There is no fundamental difference between data and class

Well I would say for a developer a data attribute means it has a special purpose while a class name is not expected to have a relevance besides css.

> I'll come back to this.

You mean by changing this PR or in a followup?

WofWca: > You mean by changing this PR or in a followup?

Either way works for me.

nicodh: So I would say change it in this PR.

WofWca: I have considered this, and I think that the way it is now is better. I think that having two ways (a data attribute and a class name) to refer to the same (kind of an) is not nice.

nicodh: Well they have different scopes - could even have different names. But I have no strong opinion on that so let's hope noone will change the class name for whatever reason and miss the dual usage

WofWca: IDK, I don't thin that doing `querySelector` by class name is _that_ unusual. We do it in multiple other places in our codebase.

PR #5784, opened 2025-11-28
change: disable delete_server_after

### LINKED ISSUE #5778: "Delete Messages from Device" gone
the PR https://github.com/deltachat/deltachat-desktop/pull/5743 removed the option "Delete Messages from **Device**" - where it should only remove the option "Delete Messages from **Server**" :)

as we want to maintain the deletion from device, the option "Delete Messages from Device" needs to come back, at best before a release.

what also comes to mind: we need to reset the then removed option, see https://github.com/deltachat/deltachat-desktop/issues/5752#issuecomment-3562764664 - not sure if that was done. alternatively, restore both option for now and remove "Delete Messages from Device" again when actually ignored by core

nicodh:
resolves #5778

### DIFF
--- a/packages/target-electron/src/deltachat/migration.ts
+++ b/packages/target-electron/src/deltachat/migration.ts
@@ -6,7 +6,7 @@ import { existsSync, lstatSync } from 'fs'
 import { join } from 'path'
 import { Logger } from '../../../shared/logger.js'
 import { mkdir, readdir, rename, rm, rmdir, stat } from 'fs/promises'
-import { DcEvent, RawClient } from '@deltachat/jsonrpc-client'
+import type { DcEvent, RawClient } from '@deltachat/jsonrpc-client'
 
 /**
  *
@@ -207,7 +207,7 @@ export async function disableDeleteFromServerConfig(
         log.info(
           `Disabling delete_server_after for chatmail account ${accountId}`
         )
-        await rpc.setConfig(accountId, 'delete_server_after', '0')
+        await rpc.setConfig(accountId, 'delete_server_after', null)
       }
     }
   } catch (error) {


packages/target-electron/src/deltachat/migration.ts
```diff
@@ -181,3 +181,39 @@ export async function migrateAccountsIfNeeded(
...
+      )
+      if (
+        is_chatmail === '1' &&
+        delete_server_after !== null &&
+        delete_server_after !== '0'
+      ) {
+        log.info(
+          `Disabling delete_server_after for chatmail account ${accountId}`
+        )
+        await rpc.setConfig(accountId, 'delete_server_after', '0')
```
WofWca:
```suggestion
        await rpc.setConfig(accountId, 'delete_server_after', null)
```

Android goes for `null`, which apparently means "reset":

https://github.com/deltachat/deltachat-android/pull/4019/files#diff-a8d42cd3356633700cac279767659e49a613752194ee76dedd22a765ea9a4dbaR114

packages/target-electron/src/deltachat/controller.ts
```diff
@@ -204,6 +207,10 @@ export default class DeltaChatController {
       // `get_next_event`, and that is the renderer process's JSON-RPC client.
       false
     )
+    await disableDeleteFromServerConfig(
```
WofWca, line 210: Do we have to await this? If we do then it'd be nice to do the migration in parallel for all accounts, instead of doing it in sequence, slowing down the startup process.

Well unless we expect that this code is to be removed very soon.

packages/target-electron/src/deltachat/migration.ts
```diff
@@ -6,7 +6,7 @@ import { existsSync, lstatSync } from 'fs'
 import { join } from 'path'
 import { Logger } from '../../../shared/logger.js'
 import { mkdir, readdir, rename, rm, rmdir, stat } from 'fs/promises'
-import { DcEvent } from '@deltachat/jsonrpc-client'
+import { DcEvent, RawClient } from '@deltachat/jsonrpc-client'
```
WofWca:
```suggestion
import type { DcEvent, RawClient } from '@deltachat/jsonrpc-client'
```

packages/target-electron/src/deltachat/controller.ts
nicodh: Well this is only executed once so I would say no harm if there is one start with a bit delay. And we don't expect users to have a vast number of accounts.

nicodh:
> We probably need to create an issue in order to not to forget to remove this code when core removes the option? Or at least a TODO.
> 
> https://github.com/deltachat/deltachat-android/pull/4019/files#diff-a8d42cd3356633700cac279767659e49a613752194ee76dedd22a765ea9a4dbaR114

Shouldn't we get a compile error when trying to set/read a setting key that does not exist in core? Not sure if that was added in https://github.com/deltachat/deltachat-desktop/commit/e17e0aa1d62d9510db2f6f9203fa0c18d31692a5

PR #3379, opened 2023-09-05
feat: add "related chats" to group profile

link2xt:
This is using experimental core API https://github.com/deltachat/deltachat-core-rust/pull/4649

For merging:
- [x] Adapt to new API once the core is updated https://github.com/deltachat/deltachat-core-rust/pull/4693
- [x] Rename from "Shared Chats" into "Related Chats" in the group profile and add a translation string
- [x] Hide by default and add a new experimental setting to enable it

PR #2042, opened 2021-01-11
add chat audit log

Simon-Laux:
### TODO
- [x] needs to reference a core that has https://github.com/deltachat/deltachat-core-rust/pull/2132 merged 
- [ ] optional: virtual lazyloaded list (like chatlist)
    - [ ] reload list/items when state changes (pending/sending system messages)
- [x] changelog entry: "added chat audit log view"
### See also
Core PR: https://github.com/deltachat/deltachat-core-rust/pull/2132
Forum Feature Proposal: https://support.delta.chat/t/show-audit-log-of-group/512

PR #5786, opened 2025-11-29
change: remove experimental feature auditLog & relatedChats

### LINKED ISSUE #5773: cleanup: remove outdated options and experiments

nicodh:
resolves #5773


packages/shared/shared-types.d.ts
```diff
@@ -32,7 +32,6 @@ export interface DesktopSettingsType {
    */
   enableAVCallsV2: boolean
   enableBroadcastLists: boolean
-  enableChatAuditLog: boolean
```
WofWca, line 35: As usual, I think that we should not remove the property from the type definition, but instead mark it as optional and add a `@deprecated` comment. See other such properties.
This is because the property could actually be present on the settings object, because we don't perform the "deletion" migration.

The same goes for the other property.

WofWca (commented):
FTR this reverts
- https://github.com/deltachat/deltachat-desktop/pull/3379
- https://github.com/deltachat/deltachat-desktop/pull/2042

I am not sure if we really want to remove these, but given that we do, this MR does the job.

There are some residuals left, I'll open another MR to remove them. Edit: https://github.com/deltachat/deltachat-desktop/pull/5801.

packages/shared/shared-types.d.ts
nicodh: > This is because the property could actually be present on the settings object

But we never know when that will not be the case anymore in the future?

r10s: what happens if a property is present on the (persisted) settings object that is unknown to current running app?

cause that can easily happen e.g. by an oversight in the past or in the future - tracking all things we ever did is fragile, and bloats code

WofWca: > But we never know when that will not be the case anymore in the future?

I didn't get it.

> what happens if a property is present on the (persisted) settings object that is unknown to current running app?

Nothing really.

What could hypothetically happen though is that first we delete a property and then decide to add one with the same name years later, possibly with a different type. That would cause a bug then.

nicodh: > > But we never know when that will not be the case anymore in the future?
> 
> I didn't get it.

I mean we will never get rid of the deprecated optional settings key

> What could hypothetically happen though is that first we delete a property and then decide to add one with the same name years later, possibly with a different type. That would cause a bug then.

I would say "hypothetically" in "some years" is not worth to keep it in the code

nicodh: And the names are specific enough that it's very unlike to happen

WofWca: It's not like one line of code is a big maintenance burden, but OK, whatever.

WofWca (approved):

PR #5791, opened 2025-12-01
chore: upgrade to core 2.29.0

nicodh:


Simon-Laux:
it solves the "Deleting message makes context stuck/freeze" bug, could be worth a mention in the changelog, but I haven't followed if we released a version with that bug, so maybe not.
https://github.com/chatmail/core/issues/7490

WofWca:
No I don't think we have. The last release was with Core 2.25.0.

PR #5795, opened 2025-12-02
Allow add transport from main scanner

nicodh:


packages/frontend/src/hooks/useProcessQr.ts
```diff
@@ -291,18 +298,14 @@ export default function useProcessQR() {
...
-          if (!userConfirmed) {
-            return callback?.()
+          if (!confirmed) {
+            return
           }
-
-          const new_accountId = await addAndSelectAccount()
-          await startInstantOnboarding(new_accountId, { ...parsed, qr })
+          await BackendRemote.rpc.addTransportFromQr(accountId, url)
+          openTransportSettings(accountId)
```
WofWca:
```suggestion
          openDialog(TransportsDialog, {
            accountId,
          })
```

No need for this indirection I think

packages/frontend/src/hooks/useProcessQr.ts
```diff
@@ -291,18 +298,14 @@ export default function useProcessQR() {
        */
       if (qr.kind === 'account') {
         if (isLoggedIn) {
-          const userConfirmed = await openConfirmationDialog({
-            message: tx('qraccount_ask_create_and_login_another', qr.domain),
-            confirmLabel: tx('login_title'),
-            dataTestid: 'ask-create-account',
+          const confirmed = await openConfirmationDialog({
+            message: `${tx('confirm_add_transport')}\n ${qr.domain}`,
```
WofWca, line 296:
```suggestion
            message: `${tx('confirm_add_transport')}\n ${qr.domain}`,
            confirmLabel: tx('add_transport'),
```

nicodh:
> Though I wonder if we need a more prominent "confirm" dialog. I think that setting the server that you'll route all your traffic through is quite a big deal.

To route your traffic through that transport it needs an additional "set as default" action.

WofWca:
But it's not like we show a confirmation dialog before that.

nicodh:
> But it's not like we show a confirmation dialog before that.

But if you compare it with the current situation - there we don't have a "special" dialog either. So I consider that safe enough

WofWca:
Hmm yep. But it doesn't mean that it's good. Also in that case you get a brand new account, and it's pretty clear what happened, unlike the "transport" case (who even understands what that means besides us).

Issue #5696, opened 2025-11-09
"enter" acts as "escape" in some dialogs; this seems unexpected


r10s:
- right tap a chat
- select "delete" from the context menu
- hit "enter" key

-> as the dialog disappears one would expect that the "delete" action, as being rightmost and outstanding button was confirmed.

but actually nothing has happened - which may lead to the situation that the user believes they deleted smth but actually did not.

i see the point that an accidental hitting of enter may be worse as well, and also the outstanding button is great, and no-focus-by-default is fine as well, all that should not change.

to fix the issue, i would just do nothing if "Enter" is hit and no button is focused.

that would make it clear to the user that no action was taken place - if they want to do nothing, they should hit "Escape" - and if they actually want to delete, hit "tab -> enter" or click the button

WofWca:
It's probably time to start using `<form>`s.

A potential hack (or a solution?) is to apply `type="button"` to such "cancel" buttons, and keep `type="submit"` on the "confirm" buttons.

WofWca:
Another approach would be to remove the `:focus-visible` thing for the buttons in such "confirm" dialogs, and instead always show the outline of the focused button, i.e. not only in the "keyboard-only" mode, so that it's always clear that "cancel" is the "default" button.

https://github.com/deltachat/deltachat-desktop/blob/5aff81758d1c6cbfa8992614ba8935b7cdf8714c/packages/frontend/scss/_global.scss#L110-L112

r10s:
_in general,_ always visible focus rect would just not look good for the many using the mouse mainly. and it would clutter their ui.

but also for non-keyboard - it is not good if sometimes the focus rect is shown on opening, and sometimes after hitting "tab".

it is better to stay with showing focus rect on first "tab", this is also what most other apps are doing (in the past, and for some ui it is different, but there the focus rect looks far less in your face)

so, no, do not always show the outline. also, _for the issue at hand,_ iiuc, we do not want to have "enter" hitting the delete button - unless focused explicitly

WofWca:
In such a case it seems that what we need is "autoblur". Chromium automatically focuses the first element inside the dialog. I guess `blur()`ing all the buttons should work. Or like `useEffect( setTimeout( document.activeElement?.blur()`.

PR #5796, opened 2025-12-02
Change copy dialog button order

### LINKED ISSUE #5792: button order in "Copy Invite Link" unexpected
the standard option in "Copy Invite Link" should be rightmost, and using a "blue" button.

left of that the cancel button as usual.

currently it is the other way round - happended to me several times that i _thought_ i copied the link, but i hit "cancel" (user do not read, me included :)

# wrong button order

Image

# better

Image

nicodh:
resolves #5792 

The dialog now is also affected by #5696 

With the wrong button order it worked to copy with "enter" since the copy button was the first focusable element. Now "enter" just cancels...

### DIFF
--- a/packages/frontend/src/components/dialogs/QrCodeCopyConfirmationDialog.tsx
+++ b/packages/frontend/src/components/dialogs/QrCodeCopyConfirmationDialog.tsx
@@ -41,7 +41,11 @@ export default function QrCodeCopyConfirmationDialog({
   }
 
   return (
-    <Dialog onClose={onClose}>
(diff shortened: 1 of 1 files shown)


r10s:
> With the wrong button order it worked to copy with "enter" since the copy button was the first focusable element. Now "enter" just cancels...

that is weird, indeed. unless smth is selected with tab or the default button is destructive, one expects "Enter" to hit the default button ... but i've seen there is already a `fix-enter-behavior` branch :)

PR #5797, opened 2025-12-02
fix: focus primary action button

### LINKED ISSUE #5696: "enter" acts as "escape" in some dialogs; this seems unexpected

nicodh:
resolves #5696

By adding a setAutoFocus prop to FooterActionButton we can control which button  should be focused after rendering.

By setting type="button" we enable the on enter => click behaviour for that button.

In general using form directive and submit button would be more comprehensive but is more relevant for dialogs with multiple form elements on it.

requestAnimationFrame should put the focus event to the end of the rendering process although it is not guaranteed that it works in all scenarios

### DIFF
--- a/packages/frontend/src/components/Button/index.tsx
+++ b/packages/frontend/src/components/Button/index.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useEffect, useRef } from 'react'
 import classNames from 'classnames'
 
 import styles from './style.module.scss'
@@ -8,6 +8,8 @@ export type ButtonProps = React.ButtonHTMLAttributes<HTMLButtonElement> & {
   // borderless means button element has no border and transparent background
   // and is of type button for accessibility reasons
   styling?: 'primary' | 'danger' | 'borderless'
+  // automatically focus this button when it mounts
+  setAutoFocus?: boolean
 }
 
 export default function Button({
@@ -15,10 +17,30 @@ export default function Button({
   active = false,
   styling,
   className,
+  setAutoFocus = false,
   ...props
 }: ButtonProps) {
+  const buttonRef = useRef<HTMLButtonElement>(null)
+
+  useEffect(() => {
+    if (setAutoFocus && buttonRef.current) {
+      // Use requestAnimationFrame to ensure the button is fully rendered
+      let cancelled = false
+      requestAnimationFrame(() => {
+        if (!cancelled) {
+          buttonRef.current?.focus()
+        }
+      })
+      return () => {
+        cancelled = true
+      }
+    }
+  }, [setAutoFocus])
+
   return (
     <button
+      ref={buttonRef}
+      type='button'
       className={classNames(
         styles.button,
         active && styles.active,
--- a/packages/frontend/src/components/Dialog/FooterActionButton.tsx
+++ b/packages/frontend/src/components/Dialog/FooterActionButton.tsx
@@ -6,13 +6,12 @@ import type { ButtonProps } from '../Button'
 
 import styles from './styles.module.scss'
 
-export default function FooterActionButton({
-  children,
-  ...props
-}: ButtonProps) {
+const FooterActionButton = ({ children, ...props }: ButtonProps) => {
   return (
     <Button className={styles.footerActionButton} {...props}>
       {children}
     </Button>
   )
 }
+
+export default FooterActionButton
--- a/packages/frontend/src/components/dialogs/ConfirmationDialog.tsx
+++ b/packages/frontend/src/components/dialogs/ConfirmationDialog.tsx
@@ -61,6 +61,7 @@ export default function ConfirmationDialog({
             {cancelLabel || tx('cancel')}
           </FooterActionButton>
           <FooterActionButton
+            setAutoFocus
             styling={isConfirmDanger ? 'danger' : 'primary'}
             onClick={() => handleClick(true)}
             data-testid='confirm'
--- a/packages/frontend/src/components/dialogs/QrCodeCopyConfirmationDialog.tsx
+++ b/packages/frontend/src/components/dialogs/QrCodeCopyConfirmationDialog.tsx
@@ -61,6 +61,7 @@ export default function QrCodeCopyConfirmationDialog({
             onClick={onCopy}
             data-testid='confirm-qr-code'
             styling='primary'
+            setAutoFocus
           >
             {tx('global_menu_edit_copy_desktop')}
           </FooterActionButton>


nicodh:
> Have you tried the native `autofocus`? Did it not work? Why?

Yes. That was my first approach. Maybe the modal of the dialog "steals" the focus?

WofWca:
Oh, we've been through the `autofocus` stuff already, actually: https://github.com/deltachat/deltachat-desktop/pull/4496#discussion_r1918730504. In short, React doesn't apply this native attribute to elements. Edit: https://github.com/facebook/react/issues/23301

Either way, are we sure that this is how we want to solve this? Do we really want "Spacebar" to act as "confirm"? Even for destructive actions?

nicodh:
We could exclude it by handling the specific key event, but I doubt if that is necessary. Any confirm dialog shows up after some action. So it needs at least 2 unintended actions to execute some unintended action.

You have a better idea?

WofWca:
With the approach I suggested in https://github.com/deltachat/deltachat-desktop/issues/5696 I believe you can only "confirm" with "Enter"?

Otherwise there is another UI pattern to have the button disabled for a few hundred milliseconds right after the dialog has been opened. Firefox browser uses it for some "confirm" popups.
But in this case I believe the "autofocus" approach will also not work? Because one doesn't want to suddenly change the focus after an arbitrary timeout.
Or maybe it can work if we make the button disabled but still focusable somehow.

nicodh:
> With the approach I suggested in #5696 I believe you can only "confirm" with "Enter"?

If we just add different types it has no effect without the focus. If the focus is on a button again space activates the onClick.

If we wrap all dialogs in a form tag and add a submit type to the button it might work, but it's not sufficient just to wrap the footer in a form as the focus is not on that form (I tried that), so it would affect all dialogs.

If you consider it important to avoid "space" events we could intercept the onClick event in the Button component and filter by event.key.

nicodh:
Where I really would like to have a form => submit behaviour would be in the create account or create group dialogs.

packages/frontend/src/components/Button/index.tsx
```diff
@@ -8,17 +8,39 @@ export type ButtonProps = React.ButtonHTMLAttributes<HTMLButtonElement> & {
...
+  setAutoFocus = false,
   ...props
 }: ButtonProps) {
+  const buttonRef = useRef<HTMLButtonElement>(null)
+
+  useEffect(() => {
+    if (setAutoFocus && buttonRef.current) {
+      // Use requestAnimationFrame to ensure the button is fully rendered
+      let cancelled = false
+      requestAnimationFrame(() => {
```
WofWca, line 29: I suspect that `requestAnimationFrame` is not the thing that we need to use.

packages/frontend/src/components/Button/index.tsx
```diff
@@ -8,17 +8,39 @@ export type ButtonProps = React.ButtonHTMLAttributes<HTMLButtonElement> & {
...
+        if (!cancelled) {
+          buttonRef.current?.focus()
+        }
+      })
+      return () => {
+        cancelled = true
+      }
+    }
+  }, [setAutoFocus])
```
WofWca, line 38: The approach suggested in https://github.com/facebook/react/issues/23301#issuecomment-2817039386 could be better, because that approach doesn't make the button receive focus every time it's mounted, but instead only when the dialog opens, i.e. it's the actual native `autofocus` behavior.

WofWca (approved):
Hmmm yep, you're right. I also tried applying `tabindex={0}` on the form, and it didn't work. Here is my diff

<details><summary>diff</summary>

```diff
diff --git a/packages/frontend/src/components/dialogs/ConfirmationDialog.tsx b/packages/frontend/src/components/dialogs/ConfirmationDialog.tsx
index c1ad7a55b..075ad4728 100644
--- a/packages/frontend/src/components/dialogs/ConfirmationDialog.tsx
+++ b/packages/frontend/src/components/dialogs/ConfirmationDialog.tsx
@@ -46,6 +46,14 @@ export default function ConfirmationDialog({
 
   return (
     <Dialog onClose={handleClose} dataTestid={dataTestid}>
+    <form
+      // tabIndex={0}
+      method='dialog'
+      onSubmit={e => {
+        console.log('form submit', e)
+        handleClick(true)
+      }}
+    >
       {header && <DialogHeader title={header} />}
       <DialogBody>
         <DialogContent paddingTop={header === undefined}>
@@ -54,21 +62,24 @@ export default function ConfirmationDialog({
       </DialogBody>
       <DialogFooter>
         <FooterActions>
+          <input hidden></input>
           <FooterActionButton
             onClick={() => handleClick(false)}
             data-testid='cancel'
+            type='button'
           >
             {cancelLabel || tx('cancel')}
           </FooterActionButton>
           <FooterActionButton
             styling={isConfirmDanger ? 'danger' : 'primary'}
-            onClick={() => handleClick(true)}
             data-testid='confirm'
+            type='submit'
           >
             {confirmLabel || tx('yes')}
           </FooterActionButton>
         </FooterActions>
       </DialogFooter>
+    </form>
     </Dialog>
   )
 }

```

</details>

OK, it's probably fine to just auto-focus the confirm button.
But either the "temporarily disable" or "only confirm on Enter" thing would be nice on top of this.

This MR, however, probably doesn't address the issue completely, as I suspect that there are more dialogs that are like this.

r10s:
yes - #5696 is however only about "enter acting as escape", without technical suggestions about focus - it seems fine if just no button responds to "enter" - and in case of destructive actions, maybe even better.

but i would leave that part open. but what i would not like is that the focus rect becomes always visible in some dialogs by the fix - this is what my last comment is about

r10s:
without being deep into these focus and keyboard discussions, i just tried the PR in practise:

it seems just fine for the "delete message" dialog

enter just does nothing by default, as it would be destructive - but if you want to use the keyboard, you can select a button using tab. i do not know, however, if there are side effects for other dialogs

nicodh:
Ok I will change the behaviour. Sorry if I had read the description details more carefully - it is already described there. I just remembered the Issue wrong.

nicodh:
Closed in favour of #5807

PR #5798, opened 2025-12-02
improvement: calls: disable out video initially

WofWca:
See https://github.com/deltachat/calls-webapp/issues/78
for discussion.

Marking as Draft because apparently it's not clear yet whether we want this.


WofWca:
We are apparently going for an explicit choice when starting a call: https://github.com/deltachat/deltachat-desktop/pull/6017

PR #5799, opened 2025-12-02
fix: edit profile accessibility

### LINKED ISSUE #5794: "Edit profile" is not accessible
- Delta Chat Version:
- Expected behavior: Can open "Edit profile" dialog using just keyboard.
- Actual behavior: Can't
- Screenshots:
- Logs:

The bug has been introduced in https://github.com/deltachat/deltachat-desktop/pull/5765.

nicodh:
resolves #5794


WofWca (approved):
This is gonna be a button whose label is "profile image. Alice. A software engineer and a nature lover", which is not good, but it's better than the button being completely inaccessible, so let's go for it.

PR #5800, opened 2025-12-02
draft: prettier pre-messages for files

Simon-Laux:
only experiment at this point, ~~this info may come from core later.~~ **Update**: it now comes from core. I changed the purpose of this branch to show how UI can present it in a prettier way, by disabling that core adds the info to the message text and displaying custom ui.

to run the experiment I recommend the browser version with fresh accounts (you need to delete them if you already make some with the multi transport branch, because the migrations are incompatible at the moment).
```
cd packages/target-browser
DELTA_CHAT_RPC_SERVER=/path/to/core/target/debug/deltachat-rpc-server WEB_PASSWORD=1234 pnpm run start
```

the core pr you need to build https://github.com/chatmail/core/pull/7558
```
cargo build -p deltachat-rpc-server
```

image

**Update**: now it looks prettier for files:

https://github.com/user-attachments/assets/ab217654-b57c-4011-8c63-f99fc9cc41cf

### DIFF
--- a/packages/frontend/src/components/message/Message.tsx
+++ b/packages/frontend/src/components/message/Message.tsx
@@ -733,7 +733,7 @@ export default function Message(props: {
       Voice: 'voice_message',
       Webxdc: 'webxdc_app',
     }
-    const viewType: T.Viewtype = (message as any).fullMessageViewType
+    const viewType: T.Viewtype = (message as any).postMessageViewType
     const viewTypeTranslationKey = viewTypeTranslations[viewType]
     const fileSize = filesize(message.fileBytes)
     const metadata = `[${viewTypeTranslationKey ? tx(viewTypeTranslationKey) : viewType} - ${fileSize}]`


Simon-Laux:
since core adds this info to the end of the message text by default, I changed the purpose of this branch.
Now it is for experimenting with custom display for those Pre-Messages.

This is **not** important right now, just an experiment, we could complete it soon when we want and have extra time or wait until we have image thumbnails in Pre-Messages and then continue it.

packages/frontend/src/components/message/Message.tsx
```diff
@@ -716,9 +717,28 @@ export default function Message(props: {
...
+      File: 'file',
+      Gif: 'gif',
+      Image: 'image',
+      Sticker: 'sticker',
+      Vcard: 'contact',
+      Video: 'video',
+      Voice: 'voice_message',
+      Webxdc: 'webxdc_app',
+    }
+    const viewType: T.Viewtype = (message as any).fullMessageViewType
```
Simon-Laux: the "as any" is here because I didn't want to build the api types.

Simon-Laux:
This is just for reference for one possible approach to do it. but it is not time for it right now.

PR #5805, opened 2025-12-03
Create chat on enter

nicodh:


packages/frontend/src/components/Button/index.tsx
```diff
@@ -19,6 +20,7 @@ export default function Button({
 }: ButtonProps) {
   return (
     <button
+      type={props.type || 'button'}
```
WofWca, line 24: Does this do anything? I think this gets overridden by `{...props}` below?

nicodh: only if props.type is defined but it's optional so a default is needed

nicodh:
> Hmmm, maybe it would be nice to first apply `type` to all `<button>`s, plus enable https://github.com/jsx-eslint/eslint-plugin-react/blob/HEAD/docs/rules/button-has-type.md rule, so that we are sure that we do not have any buttons that can accidentally submit such forms?

Well that would be quite some work. We have at least 88 buttons in the code. But this is only relevant for dialogs and there only for the html < button > tag
The form is only wrapped around some specific dialogs so even if we forget one or two the risk is not worth to apply changes to ~80 files imo

packages/frontend/src/components/Button/index.tsx
WofWca: No I mean the whole line does nothing.

WofWca:
> Well that would be quite some work. We have at least 88 buttons in the code.

For me it would be easier to do this work than to manually review whether we've missed any buttons that need to be marked as `type="button"`.
I'll do that if you have no objections.

packages/frontend/src/components/Button/index.tsx
nicodh: this line adds a type attribute to the <button tag either with the type passed in the props or if none is passed with value 'button'

WofWca: Oh, I got it. Indeed it does, sorry. Hope that this doesn't become a bug one day.

nicodh:
But the risk to introduce some side effects by changing so many files is bigger, than the risk having a case where enter in a dialog triggers the wrong button imo

WofWca:
I don't think so. We do not and should not rely on `<button>` having `type="submit"` outside of `<form>`s. The vast majority of our `<button>`s are `type="button"` sematically. And specifying `type` for all buttons is semantically the right thing to do.

packages/frontend/src/components/Button/index.tsx
nicodh: > Hope that this doesn't become a bug one day.

If that becomes a bug that means the default type='button' has some unexpected side effect, which is my argument in our other dialog here

nicodh:
> > Well that would be quite some work. We have at least 88 buttons in the code.
> 
> For me it would be easier to do this work than to manually review whether we've missed any buttons that need to be marked as `type="button"`. I'll do that if you have no objections.

I would say that should go into another PR since it affects much more dialogs & components then the ones involved here

packages/frontend/src/components/screens/WelcomeScreen/OnboardingScreen.tsx
```diff
@@ -44,33 +49,36 @@ export default function OnboardingScreen(props: Props) {
...
+              <p className={styles.welcomeScreenTitle}>
+                {tx('welcome_chat_over_email')}
+              </p>
+            </div>
+            <div className={styles.welcomeScreenButtonGroup}>
+              <Button
+                type='submit'
+                className={styles.welcomeScreenButton}
+                styling='primary'
+                onClick={props.onNextStep}
```
WofWca:
```suggestion
```

Don't call `onNextStep` twice.

packages/frontend/src/components/dialogs/CreateChat/index.tsx
```diff
@@ -663,8 +663,25 @@ export function CreateGroup(props: CreateGroupProps) {
...
+        } else {
+          // TODO: handle error
+        }
+      })
+      .finally(onClose)
+  }
+
   return (
-    <>
+    <form onSubmit={submitForm}>
```
WofWca, line 684: This causes double scroll

Image

Technically this is also the case for "create channel" but there it only happens on super short windows sizes, currently.

WofWca (changes requested):
I did not check whether all the buttons that need it have been applied `type="button"` to. Expecting that https://github.com/deltachat/deltachat-desktop/pull/5811 will take care of that.

I have checked that the callbacks are the same, that they do what they did before this MR.

We should also attach forms to dialogs (later).

When my code comments are addressed and #5811 is merged, I this MR is also good to merge.

packages/frontend/src/components/dialogs/CreateChat/index.tsx
nicodh: Fixed it with display: contents

packages/frontend/src/components/Button/index.tsx
```diff
@@ -19,7 +20,8 @@ export default function Button({
 }: ButtonProps) {
   return (
     <button
-      type='button'
+      // eslint-disable-next-line react/button-has-type
+      type={props.type || 'button'}
```
nicodh, line 24: It seems the react/button-has-type rule is too strict. This is a valid assignment but wrong detected by eslint as error

PR #5806, opened 2025-12-03
refactor: use `<dialog>.closedby`, remove hacks

WofWca:
Remove the `preventDefault` hack that we used for
`canOutsideClickClose` and `canEscapeKeyClose`.
These are now available natively.
But not on Safari (WebKit) though as of writing.

TODO:
- [ ] wait for WebKit to implement this so that this works as intended in the Tauri version on Linux.
  https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog#browser_compatibility


WofWca:
Closing because we don't know when it's going to get fixed.

PR #5807, opened 2025-12-03
fix: blur buttons in dialogs

### LINKED ISSUE #5696: "enter" acts as "escape" in some dialogs; this seems unexpected

nicodh:
resolves #5696

After some discussions: with this PR we just avoid to have a default focus in dialogs except where it is wanted (for example in Instinct Onboarding where we want to focus the "Create Profile" button and the name input field).

Now enter does not trigger the first focusable element in dialogs so it has no effect at all, except when the user tabs the focus to an element.

### DIFF
--- a/packages/frontend/src/components/Dialog/Dialog.tsx
+++ b/packages/frontend/src/components/Dialog/Dialog.tsx
@@ -20,9 +20,12 @@ type Props = React.PropsWithChildren<{
   // takes full screen and is transparent
   unstyled?: boolean
   dataTestid?: string
-  /** per default the first element in a modal dialog is focused
-   * but we remove that focus to avoid unexpected behaviours
-   * set this to true to keep the default focus behavior */
+  /**
+   * per default the first element in a modal dialog is focused
+   * but we remove that focus if the first element is a button or
+   * a 'button like' element to avoid unexpected behaviours
+   * set this to true to keep the default focus behavior also for buttons
+   */
   allowDefaultFocus?: boolean
 }>
 


nicodh:
tbd: we could also skip the blur for alert dialogs so that enter/space would close them

packages/frontend/src/components/Dialog/Dialog.tsx
```diff
@@ -20,6 +20,7 @@ type Props = React.PropsWithChildren<{
   // takes full screen and is transparent
   unstyled?: boolean
   dataTestid?: string
+  skipAutoBlur?: boolean
```
WofWca: I'd add a comment about why we do it.

packages/frontend/src/components/Dialog/Dialog.tsx
```diff
@@ -83,6 +85,10 @@ const Dialog = React.memo<Props>(
       // calling showModal is "only" the way to have ::backdrop
       dialog.current?.showModal()
       dialog.current!.style.display = 'flex'
+      // Remove focus from any auto-focused element
+      if (!skipAutoBlur && document.activeElement instanceof HTMLElement) {
+        document.activeElement.blur()
+      }
```
WofWca: This breaks things. It's executed every time a dialog is re-rendered, so e.g. if a dialog contains an input, it gets blurred on every key press. For example, see the "add members" dialog.

Perhaps moving this inside of a separate `useEffect` with empty dependencies (or just `skipAutoBlur`) should work.

nicodh: Well it is already in a useEffect but I didn't see that the dependency was missing in the existing code which is a bit strange.

I will add it and add some conditions for the element type

packages/frontend/src/components/screens/WelcomeScreen/index.tsx
```diff
@@ -73,6 +73,7 @@ export default function WelcomeScreen({ selectedAccountId, ...props }: Props) {
         canOutsideClickClose={false}
         onClose={onClose}
         dataTestid='onboarding-dialog'
+        allowDefaultFocus={true}
```
nicodh, line 76: So far this is the only place where this prop makes sense: it will focus the "Create Account" button which is the primary action for the InstantOnboarding Screen

packages/frontend/src/components/Dialog/Dialog.tsx
```diff
@@ -93,7 +88,23 @@ const Dialog = React.memo<Props>(
...
+          )
+
+        if (
+          (tagName === 'button' || isButtonLikeInput) &&
+          !document.activeElement.hasAttribute('autofocus')
+        ) {
+          // Remove focus from auto-focused buttons or button like elements
+          document.activeElement.blur()
+        }
+      }
```
WofWca, line 109: This should be inside a separate `useEffect` because otherwise `showModal()` depends on `allowDefautFocus` for no apparent reason.

packages/frontend/src/components/Dialog/Dialog.tsx
```diff
@@ -30,6 +20,10 @@ type Props = React.PropsWithChildren<{
   // takes full screen and is transparent
   unstyled?: boolean
   dataTestid?: string
+  /** per default the first element in a modal dialog is focused
+   * but we remove that focus to avoid unexpected behaviours
+   * set this to true to keep the default focus behavior */
```
WofWca:
```suggestion
   * but we remove that focus so that e.g. the user doesn't press a button
   * (with Space or Enter) which they have not explicitly focused,
   * for example the "cancel" button in the "Delete Chat?" dialog.
   * set this to true to keep the default focus behavior.
   *
   * We only prevent focus on button elements,
   * so if the first element in the dialog is e.g. a text input,
   * this prop has no effect */
```

WofWca (approved):
Alright, the bug seems to have been fixed (although I didn't re-test), and the behavior is more conservative than on my last review.

packages/frontend/src/components/Dialog/Dialog.tsx
nicodh: but allowDefautFocus won't change once the dialog is mounted. So this useEffect will only (but reliably) run once when a dialog is mounted

WofWca: Sure, the behavior is the same. It's just the semantics, future-proofing.

PR #5813, opened 2025-12-05
Error backup too new

### LINKED ISSUE #4187: In add second device dialogue, the copy button does not give any feedback that the copy to clipboard has been done.

### LINKED ISSUE #5810: wrong error message when setting up second device
when trying to set up a second device from a newer version, we get the error `backupTooNew`; the UI should show an alert with the text `multidevice_receiver_needs_update`.

however, instead, always an generic error is shown:

Image

this is how it should look like:

in fact, `backupTooNew` is used in generic scanner, however, this is not the normal flow ppl go (and you anyways never get into setting up a second device from there)

i came over that with 2.25.0 scanning a code from 2.29.0 - to check out if the update part will be communicated to the user, unfortunately, it will not, but just now, there is also not much we can do on that.

so, this is no blocker for the next release, but it should be fixed soon to have the next update experience smoother

nicodh:
fixes #5810 , closes #4187

Adds a new test for "backup too old"

### DIFF
--- a/packages/e2e-tests/tests/backup-tests.spec.ts
+++ b/packages/e2e-tests/tests/backup-tests.spec.ts
@@ -84,9 +84,7 @@ test('shows warning when scanning backups that are newer than supported', async 
 
   await sendBackupDialog.getByRole('button', { name: 'Continue' }).click()
 
-  const copyButton = sendBackupDialog.getByRole('button', { name: 'Copy' })
-  await expect(copyButton).toBeVisible()
-  await copyButton.click()
+  await sendBackupDialog.getByRole('button', { name: 'Copy' }).click()
   await page.getByTestId('alert-ok').click()
 
   const clipboardContent = await page.evaluate(async () => {
@@ -104,6 +102,7 @@ test('shows warning when scanning backups that are newer than supported', async 
   }, mutatedContent)
 
   await sendBackupDialog.getByRole('button', { name: 'Cancel' }).click()
+  // Confirm cancellation
   await page.getByTestId('confirm-dialog').getByTestId('confirm').click()
 
   await clickThroughTestIds(page, [


packages/e2e-tests/tests/backup-tests.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  page = await browser.newPage()
+  await reloadPage(page)
+})
+
+test.afterEach(async () => {
+  for (let i = 0; i < 5; i++) {
+    await page.keyboard.press('Escape')
+  }
+  existingProfiles = await loadExistingProfiles(page)
```
WofWca, line 48:
```suggestion
```

packages/e2e-tests/tests/backup-tests.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+  const settingsDialog = page.getByTestId('settings-dialog')
+  await expect(settingsDialog).toBeVisible()
+
+  await settingsDialog
+    .getByRole('button', { name: 'Add Second Device' })
+    .click()
+
+  const sendBackupDialog = page
+    .locator('dialog')
+    .filter({ hasText: 'Add Second Device' })
```
WofWca, line 82: Note that this Locator would also math the Settings dialog (if we didn't auto-close it)

packages/e2e-tests/tests/backup-tests.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+  await expect(sendBackupDialog).toBeVisible()
+
+  await sendBackupDialog.getByRole('button', { name: 'Continue' }).click()
+
+  const copyButton = sendBackupDialog.getByRole('button', { name: 'Copy' })
+  await expect(copyButton).toBeVisible()
+  await copyButton.click()
```
WofWca:
```suggestion
  await sendBackupDialog.getByRole('button', { name: 'Copy' }).click()
```

A successful `click()` implies that the button is visible.

packages/e2e-tests/tests/backup-tests.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+    /^DCBACKUP(\d+):/,
+    (_match, version) => `DCBACKUP${Number(version) + 1}:`
+  )
+
+  await page.evaluate(async text => {
+    await navigator.clipboard.writeText(text)
+  }, mutatedContent)
+
+  await sendBackupDialog.getByRole('button', { name: 'Cancel' }).click()
+  await page.getByTestId('confirm-dialog').getByTestId('confirm').click()
```
WofWca, line 106:
```suggestion
  // Confirm cancellation
  await page.getByTestId('confirm-dialog').getByTestId('confirm').click()
```

WofWca (approved):
Looks good, makes sense.

The unrelated change is also OK, although maybe it's too intrusive to show an entire dialog, it's better than nothing.

packages/frontend/src/components/dialogs/SetupMultiDevice/SendBackupDialog.tsx
```diff
@@ -70,6 +72,9 @@ export function SendBackupDialog({ onClose }: DialogProps) {
     if (qrContent) {
       runtime.writeClipboardText(qrContent)
```
Simon-Laux, line 73: you need to await this, otherwise you won't know if it succeeded.

packages/frontend/src/components/dialogs/SetupMultiDevice/SendBackupDialog.tsx
```diff
@@ -70,6 +72,9 @@ export function SendBackupDialog({ onClose }: DialogProps) {
...
       setCopied(true)
+      openAlertDialog({
+        message: tx('copied_to_clipboard'),
+      })
```
Simon-Laux, line 77: toast or sth. else that is less annoying to click away may be better

PR #5814, opened 2025-12-05
change: update local help

r10s:
this esp. updates the help entry for removed "export keys" and "encryption settings"


r10s:
yeah, it is auto generated by bin/help/update-local-help.sh - i should have said that in the pr description

WofWca:
> yeah, it is auto generated by bin/help/update-local-help.sh - i should have said that in the pr description

Nah I think it's already clear from the fact that the generated files are already present in out source code. This MR only updates them.

PR #5820, opened 2025-12-06
change: update multi device warning

nicodh:
It is not blocking anmyore and shows

image

### DIFF
--- a/packages/frontend/src/components/dialogs/Transports/index.tsx
+++ b/packages/frontend/src/components/dialogs/Transports/index.tsx
@@ -226,13 +226,6 @@ export default function TransportsDialog(
         >
           <Icon icon='plus' size={16} />
         </Button>
-        <p>
-          Note if using multi-device:
-          <br />
-          before changing or adding transports make sure all other devices have
-          at least <strong>version 2.33.0</strong> installed. Otherwise they
-          will run out of sync.
-        </p>
       </DialogFooter>
     </Dialog>
   )


packages/frontend/src/components/dialogs/Transports/index.tsx
```diff
@@ -77,11 +77,10 @@ export default function TransportsDialog(
       'bcc_self'
     )
     if (multiDeviceMode === '1') {
-      openAlertDialog({
+      await openAlertDialog({
```
WofWca, line 80: Unfortunately the returned Promise only resolves if the dialog is closed with the "OK" button. Otherwise the code gets stuck here forever.

https://github.com/deltachat/deltachat-desktop/blob/19018c958fa38ff83be93ab287f7a3bf90ea9029/packages/frontend/src/components/dialogs/AlertDialog.tsx#L21-L51

Maybe this is good for guarding against people messing up their setups.

I'll make an MR to fix this.

packages/frontend/src/components/dialogs/Transports/index.tsx
```diff
@@ -77,11 +77,10 @@ export default function TransportsDialog(
...
         message:
-          'It is not possible to add transports when multi-device mode is enabled.',
+          'Note if using multi-device:\nbefore changing or adding transports make sure all other devices have at least version 2.33.0 installed. Otherwise they will run out of sync.',
```
WofWca, line 82: IDK if "they will run out of sync" is clear. Run out of sync in what way? You said that the sent messages will not appear on old versions. I think that "they will run out of sync" is too weak of a wording.

WofWca (commented):
This warning is not shown when scanning a "new account" QR, e.g. <dcaccount:https://nine.testrun.org/new>.

The MR on its own is an improvement, but if this is the only guardrail that we have in place, I think many people might get their Delta Chats messed up.

packages/frontend/src/components/dialogs/Transports/index.tsx
nicodh: What else could "run out of sync" mean? The devices are not synced anymore.

WofWca: Synced in what way? What's supposed to be in sync and what is not? Encrytion / decryption keys? App themes? Accounts? Photos?

Also this doesn't say how hard it is to get them back in sync.

packages/frontend/src/components/dialogs/Transports/index.tsx
WofWca: https://github.com/deltachat/deltachat-desktop/pull/5822

packages/frontend/src/components/dialogs/Transports/index.tsx
nicodh: I would not be too verbose in a warning that will be shown on any add transport. Let's see what the other clients will do about it. We still can change that dialog anytime if appropriate

packages/frontend/src/components/dialogs/Transports/index.tsx
WofWca: Merged.

WofWca (approved):
As I mentioned, I fear that people will be messing up their Delta Chats on accident, e.g. by scanning "new account" QRs.

This particular MR is a good change either way.

PR #5821, opened 2025-12-06
doc: add a hint for running multiple instances

nicodh:


WofWca (approved):
Makes sense.

But maybe better to move this somewhere else, out of the very introduction.

PR #5824, opened 2025-12-06
refactor: simplify `selectChat`

WofWca:
- **refactor: move "switch to archive on selectChat"**
- **refactor: simplify `selectChat`**

Let's stop providing this unused async API and be more React'ive.

This fixes a bunch of React linter warnings
(access to refs during render)
(see https://github.com/deltachat/deltachat-desktop/pull/5770#issuecomment-3574735021).

Moving the `Switch to "archived" view` code was a requirement for the last, i.e. main, commit.
I have tested that that still works.

### DIFF
--- a/packages/frontend/src/contexts/ChatContext.tsx
+++ b/packages/frontend/src/contexts/ChatContext.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useRef, useState } from 'react'
+import React, { useCallback, useEffect, useState } from 'react'
 
 import { ActionEmitter, KeybindAction } from '../keybindings'
 import { markChatAsSeen, saveLastChatId } from '../backend/chat'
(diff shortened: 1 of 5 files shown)


packages/frontend/src/contexts/ChatContext.tsx
```diff
@@ -171,6 +123,16 @@ export const ChatProvider = ({
 
   unselectChatRef.current = unselectChat
```
nicodh, line 124: Although this PR removes some react warnings we should avoid introducing new ones. So either wrap this in a useEffect or add a line // ignore...

WofWca: What warning is introduced by this MR? This line is unchanged, the warning has been there before.

nicodh: Ok, you are right, it's not a new one but it should be addressed when refactoring this file imo

PR #5826, opened 2025-12-06
chore: prepare release 2.33.0

nicodh:


WofWca (approved):
See my thoughts about the release-ability state of current main in https://github.com/deltachat/deltachat-desktop/pull/5820.

Otherwise this looks like a regular release.

The changelog is better, but still includes some entries that I think should be omitted or reworded more clearly.

PR #5827, opened 2025-12-07
feat: show info dialog in useProcessQr

nicodh:
To have the same confirmation dialog when scanning a DC ACCOUNT qr code from main scanner

### DIFF
--- a/packages/frontend/src/components/dialogs/Transports/index.tsx
+++ b/packages/frontend/src/components/dialogs/Transports/index.tsx
@@ -17,15 +17,14 @@ import classNames from 'classnames'
 import useDialog from '../../../hooks/dialog/useDialog'
 import { processQr } from '../../../backend/qr'
 import Icon from '../../Icon'
-import { TranslationKey } from '@deltachat-desktop/shared/translationKeyType'
 
 export const addTransportConfirmationDialog = async (
   domain: string,
   multiDeviceMode: boolean,
   openConfirmationDialog: (options: { message: string }) => Promise<boolean>,
-  tx: (key: TranslationKey, ...args: any[]) => string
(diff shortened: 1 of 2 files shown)


nicodh:
These changes should go into the release so I will rebase #5826 when this PR is merged

packages/frontend/src/hooks/useProcessQr.ts
```diff
@@ -201,6 +203,9 @@ export default function useProcessQR() {
     [openConfirmationDialog, startInstantOnboardingFlow, tx]
   )
 
+  const multiDeviceMode =
+    (settingsStore && settingsStore.settings['bcc_self'] === '1') ?? false
```
WofWca, line 207: I think It's not nice to specify a default value here. I think this should be happening inside of `addTransportConfirmationDialog`.

packages/frontend/src/components/dialogs/Transports/index.tsx
```diff
@@ -17,6 +17,24 @@ import classNames from 'classnames'
...
+  openConfirmationDialog: (options: { message: string }) => Promise<boolean>,
+  tx: (key: TranslationKey, ...args: any[]) => string
+): Promise<boolean> => {
+  let message = `${tx('confirm_add_transport')}\n ${domain}`
+  if (multiDeviceMode) {
+    message +=
+      '\n\nNote if using multi-device:\nbefore changing or adding transports make sure all other devices have at least version 2.33.0 installed. Otherwise they will run out of sync.'
+  }
+  const confirmed = await openConfirmationDialog({
+    message,
```
WofWca, line 33:
```suggestion
    message,
    confirmLabel: tx('add_transport'),
```

PR #5828, opened 2025-12-08
use new 'Edit Transport' wording instead of deprecated 'Password and Account'

r10s:
this works in practise - on creating new profile "Use Classic Email as Transport" is shown - otherwise "Edit Transport".

however, the `isChatmail` condition seems wrong - it should instead be a `isEditingProfile` or so, but not sure how to change it, anyone, please feel free to commit to this PR :)


nicodh:
> - it should instead be a isEditingProfile or so

The wording is mainly for being displayed when onboarding and use a classic email. Later when you want to edit the credentials you go anyway via transports->edit so the headline "Edit profile" would not be unexpected there

r10s:
k, yeah, it's probably fine. i was just wondering, as in general, we want to tune down these "isChatmail" checks - and here it is not really about chatmail but about create/edit. but i agree, create is then non-chatmail, so all is fine.

PR #5838, opened 2025-12-09
build: update dependencies

nicodh:
various minor updates

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
     "@eslint/compat": "^2.0.0",
     "@eslint/eslintrc": "^3.3.3",
     "@eslint/js": "^9.39.2",
+    "@types/node": "^22.19.3",
     "@types/ignore-walk": "^4.0.3",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -62,23 +62,26 @@
...
+    "prettier": "^3.7.4",
     "typescript": "catalog:",
-    "typescript-eslint": "^8.46.2",
+    "typescript-eslint": "^8.51.0",
     "xml-js": "^1.6.11",
-    "yaml": "^2.8.1"
+    "yaml": "^2.8.2"
+  },
+  "dependencies": {
+    "@types/node": "^22.19.3"
```
WofWca: Shouldn't this be a dev dependency?

WofWca (commented):
Apart from `@types/node` looks good.
I didn't check all the dependencies' source code, of course.

This only upgrades dev dependencies, not runtime ones?

PR #5843, opened 2025-12-09
build: upgrade electron to v39.x

### LINKED ISSUE #5816: Electron version seems to lag behind multiple months
The electron version seems to be lacking multiple months behind, e.g. v37.7.1 was released in mid October, almost two months ago, and the latest flatpak of Delta Chat desktop seems to be using v37.6.0. Since then there have been releases almost every week. Also, there's a guide that seems to suggest [only v39.x (soon v40.x) is really safe](https://www.electronjs.org/docs/latest/tutorial/security#16-use-a-current-version-of-electron) and it's not the best idea to stay on older major versions.

Perhaps it might be worth doing some sort of automation to pull in and test newer electron versions quicker, since I assume e.g. the webxdc sandbox otherwise won't really be reliable.

- Delta Chat Version: 2.25.3 (git: flathub)
- Expected behavior: uses latest v39.x electron release or not more than 1-2 weeks behind
- Actual behavior: seems to come with a two months old v37.x electron release instead
- Steps to reproduce the problem: compare https://github.com/electron/electron/releases with "About" dialog in deltachat desktop
- Screenshots:
- Logs:

nicodh:
resolves #5816

### DIFF
--- a/packages/target-electron/package.json
+++ b/packages/target-electron/package.json
@@ -82,7 +82,7 @@
     "chai": "^5.1.1",
     "chokidar": "^3.6.0",
     "debounce": "^1.2.0",
-    "electron": "^37.6.0",
+    "electron": "^39.2.6",
     "electron-builder": "^24.13.3",
     "esbuild": "^0.25.0",
     "mocha": "^10.7.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -455,8 +455,8 @@ importers:
         specifier: ^1.2.0
         version: 1.2.1
       electron:
-        specifier: ^37.6.0
-        version: 37.6.0
+        specifier: ^39.2.6
+        version: 39.2.6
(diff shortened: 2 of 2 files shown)


WofWca (commented):
There are quite some changes, possibly fixing some of our bugs
https://www.electronjs.org/blog/electron-38-0

For example, the Wayland thing could be important https://www.electronjs.org/blog/electron-38-0#removed-electron_ozone_platform_hint-environment-variable ?

And of course the removal of macOS 11 support.

Importantly, v39 removes Chromium's support for `host-rules`. https://www.electronjs.org/blog/electron-39-0#deprecated---host-rules-command-line-switch . We gotta re-check the DNS prefetch leak fix I guess.

WofWca:
I have also confirmed that this doesn't re-introduce https://github.com/deltachat/deltachat-desktop/issues/5706.

Issue #5785, opened 2025-11-28
in webxdc, allow to open dialog on clicking https: etc.


adbenitez:
currently, when clicking an `https:`, `geo:`, etc. links in an webxdc, nothing happens.

however, this is too strict and eliminates a lot of legitimate usecases, as having an "about" page, a "donate" link - user-entered links in an editor or spreadsheet.

there are lots of ideas around about whitelisting links somehow or even checking the code, after lots of internal discussion, however, it seems fine to show an dialog as 

```
Do you want to open this link?

https://this-is-some-link/?skdhfkh+foo+bar

[ Cancel ]     [ Copy ]  [ Open ]
```

when targeting this issue, 

- [ ] add the dialog on opening https: links - show the raw ASCII domains, and not some unicode representation [^1] 
- [ ] whitelist protocols we want to support (for the whitelist check https://github.com/deltachat/deltachat-android/pull/4066)
- [ ] make sure, the links cannot be opened silently without the dialog being shown

you can use  https://github.com/webxdc/webxdc-test is links section for testing, it includes punycode links

counterparts:
https://github.com/deltachat/deltachat-android/issues/4054
https://github.com/deltachat/deltachat-ios/issues/2924

[^1]: showing plain ascii is needed to avoid [homograph attack](https://en.wikipedia.org/wiki/IDN_homograph_attack) - eg. the "а" in `wikipediа.org` is _not_ an ASCII-`a` this kind of stuff is used to trick users. by showing raw ASCII punycode, this is discoverable, `wikipediа.org` would be shown as `xn--wikipedi-86g.org`

WofWca:
> the links cannot be opened silently without the dialog being shown

So what's the decision, are we supposed to make an exception and skip the dialog for DC invite links or not? (following the discussion in https://github.com/deltachat/deltachat-desktop/issues/5750#issuecomment-3561752482).

adbenitez:
> > the links cannot be opened silently without the dialog being shown
> 
> So what's the decision, are we supposed to make an exception and skip the dialog for DC invite links or not? (following the discussion in [#5750 (comment)](https://github.com/deltachat/deltachat-desktop/issues/5750#issuecomment-3561752482)).

for invite links we show a dialog as well, the dialog of "do you want to chat with X" that normally appears when user clicks invite links

WofWca:
Yes, but do we skip the warning dialog before we show the "chat with ...?" dialog (which is text-injectable)?

[...]

PR #5845, opened 2025-12-09
fix: remove WebXDC wrapper iframe

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/3321, at least for the top frame (which is the most common use case)

Partially reverts 750721ca5b8e4e9c1c376c95cec470eefefd474c
(https://github.com/deltachat/deltachat-desktop/pull/3098),
together with b3f66a5a11a60ccb61ae4381d41abb729a1a235d
(https://github.com/deltachat/deltachat-desktop/pull/5451).

This also makes non-`target="_blank"` links be handled
by the main window, as originally intended.

I have tested:
- Exit fullscreen with Escape
- That `href` of info messages works
  - Does not reload the page if the window is already open and the href is just a hash (`#`).
  - That it opens the app at the required page if it's closed initially.
- That navigation to other origins is prevented: I tried clicking links (also non-`target="_blank"`), and navigating to other `webxdc:` instance origins - it's prevented.
- `<iframe>` navigation still works
- `visibilitychange` thing is fixed

  image
- That there are no more words "frame" in the codebase that refer to this wrapper WebXDC iframe that we're removing here.
  
For comparison, see our Tauri WebXDC implementation:

https://github.com/deltachat/deltachat-desktop/blob/a22770b19d9d9485c7b39bb6c273b6c720cdc5c7/packages/target-tauri/src-tauri/src/webxdc/commands_main_window.rs#L391


nicodh (commented):
I found different behaviour between main and that branch when clicking on the mailto link in the test-webxdc app's Links section. 

In current main just an empty page is loaded, but now the link is opened by deltachat and you get a dialog which says that mailto links can't be handled.

Is that expected?

WofWca:
You're right, I missed that. I updated the commit message and the MR description.

So this is basically another bugfix, which will also come in handy for https://github.com/deltachat/deltachat-desktop/issues/5785, i.e. we'll be able to handle non-`target="_blank"` links in the top frame.

nicodh (approved):

Nice that we get rid of the extra iframe now and can avoid the executeJavaScript

PR #5846, opened 2025-12-09
fix: add transport for DCLOGIN

### LINKED ISSUE #5844: scanning DCACCOUNT:/DCLOGIN: should behave same and create a new transport
instead of creating a new profile, one should always being asked whether to add a transport.

for DCACCOUNT:, this is already the case, however, we should do the same for DCLOGIN

- one can argue that DCLOGIN: is a bit different, however, these QR codes are pretty rare, and we do not want to have special paths for them, that introduce bugs and maintainance effort

- from the view of the user, these codes are pretty much the same

- passing DCLOGIN to addTransportFromQr() is idempotent - accidental multiple scanning does not do harm

- less can go wrong - if ppl play around, adding a transport is less harmful than a new profile where they do not know how to switch, send from different ones etc.

counterpart of https://github.com/deltachat/deltachat-ios/issues/2946

nicodh:
resolves #5844

### DIFF
--- a/packages/frontend/src/components/dialogs/Transports/index.tsx
+++ b/packages/frontend/src/components/dialogs/Transports/index.tsx
@@ -19,12 +19,12 @@ import { processQr } from '../../../backend/qr'
 import Icon from '../../Icon'
 
 export const addTransportConfirmationDialog = async (
-  domain: string,
+  domainOrAddress: string,
   multiDeviceMode: boolean,
   openConfirmationDialog: (options: { message: string }) => Promise<boolean>,
   confirmLabel: string
 ): Promise<boolean> => {
-  let message = `${confirmLabel}\n ${domain}`
+  let message = `${confirmLabel}\n ${domainOrAddress}`
   if (multiDeviceMode) {
     message +=
       '\n\nNote if using multi-device:\nbefore changing or adding transports make sure all other devices have at least version 2.33.0 installed. Otherwise they will run out of sync.'


packages/frontend/src/components/dialogs/Transports/index.tsx
```diff
@@ -96,9 +96,9 @@ export default function TransportsDialog(
     openDialog(BasicQrScanner, {
       onSuccess: async (result: string) => {
         const { qr } = await processQr(accountId, result)
-        if (qr.kind === 'account') {
+        if (qr.kind === 'account' || qr.kind === 'login') {
           const confirmed = await addTransportConfirmationDialog(
-            qr.domain,
+            qr.kind === 'account' ? qr.domain : qr.address,
```
WofWca, line 101: This parameter is still named `domain`. It needs to be renamed.

https://github.com/deltachat/deltachat-desktop/blob/c63297d5bf1cc0007ad657ae05ac9661fbb9ccae/packages/frontend/src/components/dialogs/Transports/index.tsx#L22

packages/frontend/src/components/dialogs/Transports/index.tsx
WofWca, line 109: Is Core actually supposed to be able to handle `login` QRs? `addTransportFromQr` docs state

>      * Adds a new email account as a transport
>      * using the server encoded in the QR code.

Edit: OK, it does. Its docs need to be updated.

https://github.com/chatmail/core/blob/f7ae2abe5262b51d7f4633a5dc1237931a07169e/src/configure.rs#L161-L172

packages/frontend/src/hooks/useProcessQr.ts
```diff
@@ -316,35 +319,6 @@ export default function useProcessQR() {
...
-          const userConfirmed = await openConfirmationDialog({
-            message: tx('qrlogin_ask_login_another', qr.address),
-            confirmLabel: tx('login_title'),
-          })
-
-          if (!userConfirmed) {
-            return callback?.()
-          }
-
-          const new_accountId = await addAndSelectAccount()
```
WofWca, line 339: Apparently this was the last place where `addAndSelectAccount` was used. I think this means that we should add a comment that says that `addAndSelectAccount` of `ScreenContext` is unused so it can be removed.

https://github.com/deltachat/deltachat-desktop/blob/a5c02433976909fe01a64f3049c96d0d101ae9d8/packages/frontend/src/contexts/ScreenContext.tsx#L9-L10

packages/frontend/src/hooks/useProcessQr.ts
```diff
@@ -294,10 +297,10 @@ export default function useProcessQR() {
        * 2. unconfigured account: set instant onboarding chatmail instance
```
WofWca: The comment needs to be updated I think. This part is only supposed to apply to `DCACCOUNT`, but now it seems that it applies to both.

r10s:
> If we urgently need it for the release

no, a release is totally fine without this PR. 

DCLOGIN is a super-rare thing, scanning that from main scanner even more

(this is also the reason, we really want to have the exact same code paths as DCACCOUNT. since years that DCLOGIN always gets far more resources than it deserves :)

packages/frontend/src/hooks/useProcessQr.ts
```diff
@@ -284,20 +282,24 @@ export default function useProcessQR() {
...
+       * see https://github.com/deltachat/interface/blob/main/uri-schemes.md#DCLOGIN
        *
-       * contains the url to a chatmail instance which will be used in the
-       * instant onboarding process initiated by this QR code scan
        *
        * Scenarios depending on the account status:
-       * 1. configured account: Ask the user if they want to create a new profile on the given chatmail instance
-       * 2. unconfigured account: set instant onboarding chatmail instance
+       * 1. configured account: Ask the user if they want to add a new relay to the current profile
+       * 2. unconfigured account: set instant onboarding instance
```
WofWca, line 296:
```suggestion
       * 2. unconfigured account: set instant onboarding chatmail instance
```

What's the reason for removing "Chatmail"? Can it be something other than Chatmail?

adbenitez: dclogin is even mainly useful for non-chatmail instances, for chatmail you will use dcaccount

WofWca: Oh, right. I thought that this is about `dcaccount`.

PR #5854, opened 2025-12-10
fix: also define 'process.env.NODE_ENV' in the build script for electron backend

Simon-Laux:
this may fix the issue where it tries to load `.env` file in production release builds that were made with `pnpm build4production`.


Simon-Laux:
We have this "define" already in the script that builds the frontend, it was basically missing in the script that builds the backend,

Issue #5856, opened 2025-12-11
wrong "+" icon when changing profile order


r10s:
when changing the order of profiles (so, moving profiles), a "+" icon is shown beside the dragged avatar.

this is widely regarded as a symbol for "copy", so wrong here.

it should just be removed (no replacement or so).

moreover, on mac, the cursor on moving is weird - not sure where that comes from, i would not ship a special one, if in doubt, just use the standard arrow, or the "grabbing hand" (if available)

<img width="110" height="308" alt="Image" src="https://github.com/user-attachments/assets/ace4ed5b-d652-4ad9-9d69-7e8451453e11" />

r10s:
ah, according to https://www.w3schools.com/cssref/playit.php?filename=playcss_cursor&preval=move , this seems to be the "copy" cursor - so we should change that to "move" or "auto" - i would go for "auto", there is already the image dragged shown, so no need for another signal, it is quite standard to use the default cursor in these cases

PR #5857, opened 2025-12-11
style: apply "move" cursor on account list drag

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/5856.

There is more that can be done to improve the drag and drop feedback,
e.g. to show the "not allowed" cursor when trying to drag an account
over a message list, but this shoud cover the most common cases.

Please test this on macOS, check whether the issue is resolved.


nicodh (approved):
Reviewed by reading & testing

PR #5860, opened 2025-12-11
remove unused stock strings

r10s:
this PR removes two since some time unused stock strings, avoiding a warning in the log


nicodh:
As long as strings are not removed from json-rpc we keep them in that file (but put deprecated strings in the "omit" section) to make sure we are not missing changes

PR #5863, opened 2025-12-12
Show error on add relay

### LINKED ISSUE #5868: Adding Relays/transports: configure failure is neither handled nor shown to the user
- Delta Chat Version: main 2a6069eab3e299ca645b9b9024ecb294d6cd4dfe

### Steps to reproduce the problem:

Use the following qr code link to trigger an configuration failure
```
dclogin://myself@example.com?p=url%2FEncoded%5C%40passw%40rd&v=1
```

You'll observer that nothing happens in the UI, the relay is not added and there is no error dialog shown to the user.

The log shows an unhandled promise rejection:
```
Unhandled Rejection: Konfiguration fehlgeschlagen. Fehler: Could not find your mail server.

Please check your internet connection. PromiseRejectionEvent {isTrusted: true, reason: {…}, type: 'unhandledrejection', target: Window, currentTarget: Window, …} {code: -1, message: 'Konfiguration fehlgeschlagen. Fehler: Could not fi…l server.\n\nPlease check your internet connection.'}
```

### Expected:
Error should be shown to the user.

Showing the error will also be needed for https://github.com/chatmail/core/issues/7608

nicodh:
resolves #5868 
For testing: on a classic email profile set "move mails to deltachat folder" to true or "Only fetch from Deltachat folder".

Core throws an error when adding a transport which was ignored before.

Refactored to a hook to make the call easier

### DIFF
--- a/packages/frontend/src/hooks/dialog/useAddTransportDialog.ts
+++ b/packages/frontend/src/hooks/dialog/useAddTransportDialog.ts
@@ -1,0 +1,52 @@
+import { useCallback } from 'react'
+
+import { BackendRemote } from '../../backend-com'
+import useAlertDialog from './useAlertDialog'
+import useConfirmationDialog from './useConfirmationDialog'
+import useTranslationFunction from '../useTranslationFunction'
+
+type AddTransportDialogFn = (
+  accountId: number,
+  transportString: string,
+  domainOrAddress: string,
+  multiDeviceMode: boolean,
+  confirmLabel?: string
+) => Promise<boolean>
+
+export default function useAddTransportDialog(): AddTransportDialogFn {
+  const tx = useTranslationFunction()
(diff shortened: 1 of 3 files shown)


packages/frontend/src/hooks/dialog/useAddTransportDialog.ts
```diff
@@ -0,0 +1,52 @@
...
+        message,
+      })
+      if (!confirmed) {
+        return false
+      }
+      try {
+        await BackendRemote.rpc.addTransportFromQr(accountId, transportString)
+      } catch (e) {
+        openAlertDialog({
+          message: 'Relay could not be added. ' + (e as Error).message,
```
Simon-Laux, line 44: this string should be translated I guess

PR #5864, opened 2025-12-13
change: split transport label

nicodh:
This is how the other clients show the transport list

image

### DIFF
--- a/packages/frontend/src/components/dialogs/Transports/index.tsx
+++ b/packages/frontend/src/components/dialogs/Transports/index.tsx
@@ -186,8 +186,12 @@ export default function TransportsDialog(
                     />
                   </span>
                   <label id={`transport-label-${index}`}>
-                    {transport.addr}{' '}
-                    {transport.isDefault ? ` (${tx('def')})` : ''}
+                    <strong>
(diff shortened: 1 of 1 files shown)


nicodh:
> The code looks alright, but I personally don't think that such obfuscation is useful.

Is that why you didn't approve it? As said above this is the way the other clients do it. Since we in general try to have similar UIs in all clients a review should not be rejected based on personal taste, especially in the UI where there will always be different opinions .

WofWca:
> Is that why you didn't approve it?

No, that was because I thought that you're expecting a review from @r10s

nicodh:
> No, that was because I thought that you're expecting a review from @r10s

That was more since I didn't expect you to review. :-)

r10s:
> The code looks alright, but I personally don't think that such obfuscation is useful

reasoning was:

- do not confuse ppl with email addresses, ppl will try to enter that somehow "ahh, email, i know how this works!!!" - we had that at tones of other places before, and have lots of experience with that. do not only say, we cannot be used as an email client, act like it :)

- apart from that, what really matters is the server part - and that would be hidden quite behind the random string (which also sometimes is much longer than nine characters)

PR #5865, opened 2025-12-13
fix: add min/max width to vcard

### LINKED ISSUE #5839: shared vcard is too short
Image

nicodh:
resolves #5839

image

### DIFF
--- a/packages/frontend/src/components/message/styles.module.scss
+++ b/packages/frontend/src/components/message/styles.module.scss
@@ -61,7 +61,7 @@
   padding-top: 0;
   padding-inline-end: 15px;
   min-width: 200px;
-  max-width: min(260px, 100%);
+  max-width: 260px;
   .avatar {
     margin-bottom: 0;
   }


packages/frontend/src/components/message/styles.module.scss
```diff
@@ -60,20 +60,26 @@
   padding: 9px;
   padding-top: 0;
   padding-inline-end: 15px;
+  min-width: min(220px, 100%);
+  max-width: min(260px, 100%);
```
WofWca: I believe there should be a less complex way to express this, whatever this is expressing. `max-width` seems like it can be just `260px` or am I wrong?

packages/frontend/src/components/message/styles.module.scss
```diff
@@ -60,20 +60,26 @@
...
     padding-inline-start: 10px;
     justify-content: center;
+    min-width: 0;
   }
   .displayName {
     margin-bottom: 0;
     margin-top: 10px;
     font-weight: 200;
     font-size: medium;
+    max-width: min(200px, 100%);
```
WofWca: Do we need to have two places where we specify widths? One in `displayName` and one in `.vcard`.

packages/frontend/src/components/message/styles.module.scss
nicodh: No. If the wrapping component is smaller than 260px it would overlap. So this rule says if 100% is less than 260px then take 100%

packages/frontend/src/components/message/styles.module.scss
nicodh: Ok, Vcard alone might be sufficient

nicodh:
> Did you take the screenshot on the current version of the branch? Because this doesn't seem to work.

Fixed it with setting min-width to a fix number

nicodh:
Rebased...

packages/frontend/src/components/message/styles.module.scss
```diff
@@ -60,20 +60,25 @@
   padding: 9px;
   padding-top: 0;
   padding-inline-end: 15px;
+  min-width: 200px;
+  max-width: min(260px, 100%);
```
WofWca: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/max-width

> max-width overrides [width](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/width), but [min-width](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/min-width) overrides max-width.

So is the `100%` much useful? I think it would only work on widths between 260px and 200px. If the parent is smaller than 200px, then we'd overflow.

Also this causes some weird behavior. The message bubble is stretched to fit the text, but the vcard itself remains narrow. IDK if it's Chromium being buggy, but it is what it is:

Image

nicodh: This can only be fixed with a static max-width which seems the better compromise imo

PR #5869, opened 2025-12-13
fix: add missing error dialog when adding transport fails

Simon-Laux:


nicodh:
This seems to be a duplicate of #5863 which also includes a refactoring

Simon-Laux:
duplicate of #5863

PR #5871, opened 2025-12-13
fix: improve webxdc data cleanup and fix closing and cleaning up webxdc apps when chat is deleted.

### LINKED ISSUE #5758: improve webxdc web data cleanup
webxdc deletion on desktop currently uses the `WebxdcInstanceDeleted` event.
But there are cases when this event is not reliable, like when deleting chats (https://github.com/chatmail/core/issues/6670) or when dc crashes after the event was emitted.

Android had an even bigger problem where old webxdc data was never deleted, which is being solved right now https://github.com/deltachat/deltachat-android/pull/4050

Would be nice to argument the cleanup in desktop with the new core API [`get_existing_msg_ids()`](https://github.com/chatmail/core/pull/7487) (released in core 2.18) to catch the cases where the `WebxdcInstanceDeleted` event doesn't work.

Could be a throttled function that runs only once at a time and is triggered by the `WebxdcInstanceDeleted` event, which sends all ids of webxdcs that were opened to core and core then returns the ids that still exist and desktop deletes the data for the ids that don't exist anymore. 

But at-least on electron edition it may be hard to know all origins, because a "get all Origins that have data" method, like android has, is missing from the electron API.
so a few alternative ideas to get the message ids:
- so we might need to send big ranges of ids to core to let core check it
- or there may be a way to extract the opened webxdc ids from the window positions that we store in ui config keys like "ui.desktop.webxdcBounds.281425" (those get created when you open a webxdc), but I don't know if there is a core API yet to return all ui config keys.

example of the ui keys in my account database:
Image

Simon-Laux:
changes: 
- save last bounds also on startup not only when closing app.
- cleanup old webxdc app webdata of webxdc apps that don't exist anymore
- cleanup webxdc data when deleting a chat (workaround for missing webxdc instance deleted event in that case, https://github.com/chatmail/core/issues/6670)

closes #5758

I tested that this works correctly when deleting single messages, that it works with ephemeral messages and when deleting a chat.

this pr needs a core with https://github.com/chatmail/core/pull/7579 merged.

### DIFF
--- a/packages/target-electron/src/deltachat/webxdc.ts
+++ b/packages/target-electron/src/deltachat/webxdc.ts
@@ -984,21 +984,19 @@ export default class DCWebxdc {
 
   webxdcCleanupRunning = false
 
-  async webxdcCleanup(accountId: number) {
-    if (this.webxdcCleanupRunning) {
-      log.warn('webxdcCleanup is still running, it can only run once at a time')
-      return
-    }
-    this.webxdcCleanupRunning = true
-
+  private async webxdcCleanupInner(accountId: number) {
     // get all webxdc's using the lastBounds ui config keys, which are created as soon as you open or close a webxdc app
     // this workaround is needed because electron does not currently have an api to list all origins which have data.
     // see https://github.com/deltachat/deltachat-desktop/issues/5758
     const uiConfigKeys = await this.rpc.getAllUiConfigKeys(accountId)
-    const boundsKeys = uiConfigKeys.filter(key =>
+    const boundsKeys: string[] = uiConfigKeys.filter(key =>
       key.startsWith('ui.desktop.webxdcBounds.')
     )
-    const instanceIds = boundsKeys.map(key =>
+    if (boundsKeys.length == 0) {
+      log.warn('webxdcCleanup has nothing to do, boundsKeys is empty')
+      return
+    }
+    const instanceIds: number[] = boundsKeys.map(key =>
       Number(key.replace('ui.desktop.webxdcBounds.', ''))
     )
     // check if they message ids exist
@@ -1019,7 +1017,22 @@ export default class DCWebxdc {
     log.info(
       `webxdcCleanup cleared data of ${nonExistingInstanceIds.length} deleted webxdc apps in account ${accountId}`
     )
-    this.webxdcCleanupRunning = false
+  }
+
+  async webxdcCleanup(accountId: number) {
+    if (this.webxdcCleanupRunning) {
+      log.warn('webxdcCleanup is still running, it can only run once at a time')
+      return
+    }
+    this.webxdcCleanupRunning = true
+
+    try {
+      await this.webxdcCleanupInner(accountId)
+    } catch (err) {
+      log.warn('webxdcCleanup failed:', err)
+    } finally {
+      this.webxdcCleanupRunning = false
+    }
   }
 
   async removeWebxdcAppData(accountId: number, instanceId: number) {


Simon-Laux:
we could also think about whether we also want to run the cleanup logic once on startup.

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -985,6 +978,65 @@ export default class DCWebxdc {
...
+    this.webxdcCleanupRunning = true
+
+    // get all webxdc's using the lastBounds ui config keys, which are created as soon as you open or close a webxdc app (previous versions set this only on close)
+    // this workaround is needed because electron does not currently have an api to list all origins which have data.
+    // see https://github.com/deltachat/deltachat-desktop/issues/5758
+    const uiConfigKeys = await this.rpc.getAllUiConfigKeys(accountId)
+    const boundsKeys = uiConfigKeys.filter(key =>
+      key.startsWith('ui.desktop.webxdcBounds.')
+    )
+    const instanceIds = boundsKeys.map(key =>
```
nicodh: If boundsKeys is empty all next steps can be skipped

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -985,6 +978,65 @@ export default class DCWebxdc {
...
+      stillExistingInstanceIds,
+      nonExistingInstanceIds,
+    })
+    for (const id of nonExistingInstanceIds) {
+      this.removeWebxdcAppData(accountId, id)
+    }
+    log.info(
+      `webxdcCleanup cleared data of ${nonExistingInstanceIds.length} deleted webxdc apps in account ${accountId}`
+    )
+    this.webxdcCleanupRunning = false
```
nicodh: Maybe wrap in a try ... final condition in case an error occurs before this line is reached

packages/target-electron/src/deltachat/webxdc.ts
WofWca: I don't think we need extra logic for this corner case

packages/frontend/src/system-integration/webxdc.ts
```diff
@@ -19,6 +19,11 @@ export function initWebxdc() {
   BackendRemote.on('WebxdcInstanceDeleted', (accountId, { msgId }) => {
     runtime.notifyWebxdcInstanceDeleted(accountId, msgId)
   })
+  BackendRemote.on('ChatDeleted', accountId => {
+    // This is only needed as long as it is not part of 'WebxdcInstanceDeleted'
```
WofWca:
```suggestion
    // This is only needed as long as chat deletion
    // still doesn't fire the 'WebxdcInstanceDeleted' event
    // for WebXDC instances inside the chat.
```

packages/runtime/runtime.ts
```diff
@@ -118,7 +118,17 @@ export interface Runtime {
     payload: number[]
   ): void
   notifyWebxdcMessageChanged(accountId: number, instanceId: number): void
-  notifyWebxdcInstanceDeleted(accountId: number, instanceId: number): void
+  /** notify runtime that a webxdc instance was deleted.
+   * It then closes it's window if it was open and deletes its web data (like localstorage).
+   *
+   * It also starts the webxdc cleanup which cleans up webxdc app data of app instances that don't exist anymore.
```
```suggestion
   * It also starts the webxdc cleanup which cleans up webxdc app data of other app instances that don't exist anymore.
```

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -985,6 +978,65 @@ export default class DCWebxdc {
...
+  webxdcCleanupRunning = false
+
+  async webxdcCleanup(accountId: number) {
+    if (this.webxdcCleanupRunning) {
+      log.warn('webxdcCleanup is still running, it can only run once at a time')
+      return
+    }
+    this.webxdcCleanupRunning = true
+
+    // get all webxdc's using the lastBounds ui config keys, which are created as soon as you open or close a webxdc app (previous versions set this only on close)
```
```suggestion
    // get all webxdc's using the lastBounds ui config keys, which are created as soon as you open or close a webxdc app
```

I'd omit this part. It's good as part of a commit message, but might get confusing after a while.

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -509,10 +509,11 @@ export default class DCWebxdc {
         delete open_apps[`${accountId}.${msg_id}`]
       })
 
-      webxdcWindow.once('close', () => {
+      const saveBounds = () => {
         const lastBounds = webxdcWindow.getBounds()
         this.setLastBounds(accountId, msg_id, lastBounds)
-      })
+      }
+      webxdcWindow.once('close', saveBounds.bind(this))
```
WofWca, line 520:
```suggestion
      // Note that this also runs when closing the window
      // as a result of its message getting deleted.
      // This is fine, we'll still clean it up next time.
      webxdcWindow.once('close', saveBounds.bind(this))
```

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -985,6 +978,65 @@ export default class DCWebxdc {
...
   }
 
+  webxdcCleanupRunning = false
+
+  async webxdcCleanup(accountId: number) {
+    if (this.webxdcCleanupRunning) {
+      log.warn('webxdcCleanup is still running, it can only run once at a time')
+      return
+    }
+    this.webxdcCleanupRunning = true
```
WofWca: Nitpick: perhaps this thing with "only one at a time" is unnecessary.
It makes it so that if we do need to run another one, we will not run it because the previous one hasn't finished.

Something like `asyncThrottle` / `asyncDebounce` (or maybe just debounce) would be more appropriate.

packages/target-tauri/src-tauri/src/webxdc/commands_main_window.rs
```diff
@@ -110,18 +110,19 @@ pub(crate) async fn on_webxdc_message_deleted(
...
-    instance_id: u32,
+    instance_id: Option<u32>,
 ) -> Result<(), Error> {
-    if let Some((window_label, _)) = webxdc_instances
-        .get_webxdc_for_instance(account_id, instance_id)
-        .await
-    {
-        if let Some(window) = app.get_window(&window_label) {
-            window.destroy()?;
+    if let Some(instance_id) = instance_id {
```
WofWca, line 115: So this is basically unfinished and is to be addressed in https://github.com/deltachat/deltachat-desktop/pull/5872

WofWca (commented):
So far I only tested this without Core. Deleting single messages works.
Will wait for Core upgrade to test again.
But overall looks OK.
Not relying on core (edit: I mean the `WebxdcInstanceDeleted` event) here is probably the right thing.

To test, go to `%APPDATA%\DeltaChat\Partitions\webxdc_123\IndexedDB` and see if directories get deleted.

packages/target-tauri/src-tauri/src/webxdc/commands_main_window.rs
Simon-Laux: I fixed the rust error, so it compiles again.

packages/target-electron/src/deltachat/webxdc.ts
Simon-Laux: sure those methods would be better, but there are not used in the back-end yet, so they would change our dependencies, which I didn't want to do in this pr. this could be an improvement in a follow-up pr.

Simon-Laux: Also technically this should be throttled per account, so it can run in parallel for different accounts.

PR #5872, opened 2025-12-13
fix: tauri: improved webxdc cleanup, now webxdc app data is also cleaned up when deleting a chat

Simon-Laux:
- [ ] test on windows
- [x] test on linux
- [ ] test on macOS
- [ ] code todo: only run once at a time (per account id)
- [ ] macos "get all data store ids" api has a minimum macos version, so check if it affects us.

I did this pr as pro-bono work, it is the tauri counterpart to https://github.com/deltachat/deltachat-desktop/pull/5871


nicodh:
Closing since currently noone is working on it.

PR #5873, opened 2025-12-13
refactor: tauri: fix clippy warnings

Simon-Laux:
- **fix: tauri: fix clippy warnings**
- **make an exception for the module_inception in themes**

### DIFF
--- a/packages/target-tauri/crates/user-notify/src/platform_impl/xdg/category.rs
+++ b/packages/target-tauri/crates/user-notify/src/platform_impl/xdg/category.rs
@@ -32,6 +32,6 @@ impl Display for XdgNotificationCategory {
             TransferError => "transfer.error",
             Custom(category) => category,
         };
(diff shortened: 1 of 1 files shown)


packages/target-tauri/crates/user-notify/src/platform_impl/xdg/category.rs
```diff
@@ -29,7 +31,7 @@ impl XdgNotificationCategory {
             TransferComplete => "transfer.complete",
             TransferError => "transfer.error",
             Custom(category) => category,
-        }
-        .to_owned()
+        };
+        write!(f, "{}", category_string)
```
WofWca:
```suggestion
        write!(f, "{category_string}")
```

This is what Clippy wants.

WofWca (approved):
This doesn't look like this will affect behavior.
I wouldn't call this a "fix" though, a refactor rather.

PR #5875, opened 2025-12-13
feat: add badge counter to window title

### LINKED ISSUE #5808: Indicate the number of unread messages in window title
I'm running DeltaChat 2.25.3 on GNU/Linux on Wayland (using sway).

There are two things that are typically visible to me when delta is not in the foreground:

- the icon in the tray
- the titlebar of the window

When i have zero unread messages, the icon is just the delta balloon.  When i have more than zero unread messages, the icon gets a little blue dot on it as well.

However, the title bar is always just named "Delta Chat".  For parity with the icon, the the title bar could be changed to something like "Delta Chat ⭐"  when new messages are available.  Or, for even higher-density feedback, it could be changed to "Delta Chat (13)" to indicate the number of available messages.

By way of comparison, Signal Desktop with 5 unread messages has its title bar set to "Signal (5)".

Simon-Laux:
- **tauri: show badge counter in window title**
- **electron: add badge count to title**

Test the change in the electron edition:
- [x] test on windows
- [x] test on macOS
- [ ] test on linux 
  - [x] KDE
  - [x] Gnome
  - [ ] sway

closes #5808

### DIFF
--- a/packages/target-electron/runtime-electron/runtime.ts
+++ b/packages/target-electron/runtime-electron/runtime.ts
@@ -24,6 +24,7 @@ import type { dialog, app, IpcRenderer, webUtils } from 'electron'
 import type { LocaleData } from '@deltachat-desktop/shared/localize.js'
 import type { getLogger as getLoggerFunction } from '@deltachat-desktop/shared/logger.js'
 import type { setLogHandler as setLogHandlerFunction } from '@deltachat-desktop/shared/logger.js'
+import { appName } from '@deltachat-desktop/shared/constants.js'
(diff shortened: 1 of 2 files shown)


Simon-Laux:
> This part is not covered, apparently?

true, but I don't think this is needed on macOS, as there we already have a working badge and the native window title is shown where you would expect to see the count. (when switching windows and probably also in the dock on hover.)

Simon-Laux:
Let's merge it and we can improve it in a follow up pr if necessary.

PR #5876, opened 2025-12-14
feat: prettier About Dialog and introduce Protocol View Dialog

### LINKED ISSUE #5772: feat: prepend relevant debug infos to logs
The item "View log" in Settings->Advanced should open a dialog showing a (limited?) dump of the current log entries and relevant infos about the system, storage etc. in a textarea, similar to the view in the mobile clients.

The text should be editable and a share and a copy button should be available.

A warning should be shown: "This log may contain sensitive information. If you want to post it publicly you may examine and edit it beforehand"

For context see https://github.com/chatmail/core/pull/7486

and

https://github.com/deltachat/deltachat-desktop/pull/5756

Simon-Laux:
- move browser version gathering to browser runtime
- new about screen:
  - no debug info, instead a minimal dialog with a logo, description, app version and a few weblinks.
  - git version is only shown when DC is started with `--devmode` (git version shows branch and commit, useful during development)
- new protocol view screen
  - just a big text field where you can readact the log befor copying
  - offers options: reload, copy to clipboard and share to chat
  - contains the new space usage report

closes #5772

---

# Progress

### core features:
- [x] new about dialog
- [x] implement reading log in tauri and browser edition
- [x] implement share to chat
- [x] transfer remaining to do comments here into to do items, work on them or both

### lower priority:
- [ ] add resistance if either call fails it should still show the rest (get account info and get storage usage report in `LogDialog`)
- [ ]  check if `useDialog`/`useAlertDialog` hook can trigger an unwanted reload (which would loose user edits) - > `openAlertDialog` dependency in `update` callback
- [ ] add user feedback when copy was successful
- [ ] on sharing: close about or settings dialog as well if it is still open. (or easier: close them when opening the log)
- [ ] optional: think about making runtime info section prettier / think about re-adding a prettier versions section
- [ ] IDEA: import support contact vcard and pass a parameter to display it on top in share dialog?
---

# Screenshots

about dialog in light theme about dialog in dark theme

protocol dialog

### Other Editions

tauri edition
browser edition

### DIFF
--- a/packages/frontend/src/components/Dialog/DialogHeader.tsx
+++ b/packages/frontend/src/components/Dialog/DialogHeader.tsx
@@ -36,6 +36,7 @@ type Props = PropsWithChildren<{
    */
   onContextMenuClick?: (_event: any) => void
   dataTestid?: string
+  className?: string
 }>
 
 export default function DialogHeader(props: Props) {
@@ -50,7 +51,7 @@ export default function DialogHeader(props: Props) {
   const dataTestid = props.dataTestid ?? 'dialog-header'
   const tx = useTranslationFunction()
   return (
-    <header className={classNames(styles.dialogHeader)}>
+    <header className={classNames(styles.dialogHeader, props.className)}>
       {onClickBack && (
         <BackButton onClick={onClickBack} data-testid={dataTestid + '-back'} />
       )}
--- a/packages/frontend/src/components/dialogs/Log/index.tsx
+++ b/packages/frontend/src/components/dialogs/Log/index.tsx
@@ -15,6 +15,8 @@ import useAlertDialog from '../../../hooks/dialog/useAlertDialog'
 import useContextMenu from '../../../hooks/useContextMenu'
 
 import styles from './styles.module.scss'
+import WebxdcSaveToChatDialog from '../WebxdcSendToChat'
+import useDialog from '../../../hooks/dialog/useDialog'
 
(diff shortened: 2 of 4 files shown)


r10s:
i see, this is still a draft, however, some high-level feedback, to stay focused:

the "View Log" dialog is pretty nice, some comments:

-  iirc, we added the `** this log may contain sensitive information ... **` on purpose atop of the log, that way it is preserved, if it is about to be shared with other tools or so

- we indeed may want it translatable, but we should drive that from android. so just use hard coded string for now

- a "reload" is not needed nor wanted for now, might even do harm as it is not present elsewhere and on reloading otherwise fine information get lost. it is also not really clear or visible to the user what will happen. if really needed, closing and reopening is good enough

- the [sections] and partly used json could probably be streamlined and flattened

for the "about dialog": 

- we want to have _less_ there, not _more_. so the additional links (source code, forum), kind of another menu, icons and slogan (at the end forgotten to maintain) are not wanted - only a single link to homepage, help, view log, done. this pushes focus on the more important things and matches the level of detail used elsewhere. also, the links a not exhaustive and already in the menu

- the centering and delta chat logo is nice, so let's go for smth as: EDIT: here is the draft:<br>Screenshot 2025-12-15 at 18 34 25


if needed we can do add more things there, but that is another PR - focus of this PR is not adding things to "about" but making log accessible

Simon-Laux:
superseded by #5882

PR #5878, opened 2025-12-14
fix: replace 404 transifex link in metainfo.xml with one that works

fluoriteByte:
https://www.transifex.com/delta-chat/public/ doesn't exist, but https://www.transifex.com/delta-chat/ does

this avoids a warning  in appstream validation for flatpak (also dont ask what im doing, secret)


fluoriteByte:
all these wasted runners, for a single metadata file, wonderful

fluoriteByte:
WHEN DID I CLOSE THIS???

PR #5882, opened 2025-12-16
feat: prettier About Dialog and introduce Protocol View Dialog (+ added storage usage report to that dialog)

### LINKED ISSUE #5772: feat: prepend relevant debug infos to logs

Simon-Laux:
simpler version of #5876

---

## what it does

- move browser version gathering to browser runtime
- new about screen:
  - no debug info, instead a minimal dialog with a logo, description, app version and a few weblinks.
  - git version is only shown when DC is started with `--devmode` (git version shows branch and commit, useful during development)
- new protocol view screen
  - just a big text field where you can readact the log before copying
  - offers options: copy to clipboard and share to chat
  - contains the new storage usage report

closes #5772

## progress

- [x] simplify
  - [x] remove slogan and most links from about dialog
  - [x] remove unused icons again 
  - [x] rm reload option
  - [x] put the warning into log
  - [x] streamline/deduplicate runtime info and flatten it (making runtime info section prettier / think about re-adding a prettier versions section)
  - [x] make sure hooks can not trigger reloads
- [x] fix: on sharing: close about or settings dialog as well if it is still open. (or easier: close them when opening the log) 
- [x] add resistance if either call fails it should still show the rest (get account info and get storage usage report in `LogDialog`)
- [x] add missing feedback on copy to clipboard
- [x] hardcode warning string
- [x] normal buttons instead of settings sections for the buttons
- [x] make about dialog prettier with css (button size and optimize spacing) -> good enough for now, if someone else wants to play with the css, feel free to do so.

possibly out of scope for this pr:
- [ ] IDEA: import support contact vcard and pass a parameter to display it on top in share dialog?

## screenshots

image

image

### DIFF
--- a/packages/frontend/src/components/dialogs/About/styles.module.scss
+++ b/packages/frontend/src/components/dialogs/About/styles.module.scss
@@ -14,9 +14,7 @@
 }
 
 .buttonContainer {
-  width: 100%;
   display: flex;
-  margin-bottom: 42px;
-  justify-content: space-evenly;
-  margin-top: 43px;
+  margin: 42px 0;
+  gap: 20px;
 }


packages/frontend/src/components/dialogs/About/styles.module.scss
```diff
@@ -0,0 +1,22 @@
...
+.appName {
+  margin-bottom: 15px;
+}
+
+.buttonContainer {
+  width: 100%;
+  display: flex;
+  margin-bottom: 42px;
+  justify-content: space-evenly;
+  margin-top: 43px;
```
r10s: the buttons look a bit place randomly, with their different widths centering at additional axis.

the following puts the buttons together and centeres them in the middle.

```suggestion
  display: flex;
  margin: 42px 0;
  gap: 20px;
```

btw, the buttons are higher than in other dialogs - but that might be a general issue? that we still have different types? if so, we should create a dedicated issue for that

Simon-Laux: > btw, the buttons are higher than in other dialogs

if you mean that there is more space bellow: I didn't use the dialog footer buttons, as your design has them more like in the content of the dialog like the macOS about system dialog does.

if you mean they are too big and generally don't looks good:
I agree, that's why I went for the settings buttons in the first iteration, anyways we can still improve this in a follow up pr.

r10s: it is about the height of the buttons, not about padding/margin

the height should be unified, unless there are good reasons for different heights

Simon-Laux: I don't see the difference in height. if you measured the screenshots above, then you need to know that the screenshot scale is different, so you would need to open the screenshots or try the branch to measure correctly.
But maybe you mean sth. different?

r10s: thanks for following up!

maybe the following image explains better what i mean :) 
left: too high, somehow different class or style,
right: correct

Simon-Laux: strange they showed really similar height in the inspector for me, but that inspector feels a bit buggy anyway recently (takes long to load all the time). seems like those buttons are slightly different. I agree that they should be unified more. though that should be a new issue, to consolidate button styles.

Simon-Laux: https://github.com/deltachat/deltachat-desktop/issues/5887

r10s:
ftr,  before / after on merging:

PR #5884, opened 2025-12-16
fix: disable email address field when editing existing transports

### LINKED ISSUE #5870: Changing email address dialog needs to be updated
When I want to change my email address, then I need to use the multi transport feature to do so. https://support.delta.chat/t/how-to-migrate-your-profile-to-another-relay-version-2-33/4441

In the multi-transport, transports are identified by their address, so changing the address will add a new transport.

Currently I get this dialog when attempting to change my email address. 
Image
It tells me that it is not possible at the moment, but it is possible since multitransport is out.

I think we need to improve this flow, I have two ideas:
A: update text in the dialog to tell you that you need to instead add a new transport, could link to the guide in the forum (or an faq entry once we have one)
B: make the field read only or otherwise make clear that it is not editable

CC @r10s @link2xt

Simon-Laux:
also remove the ""coming soon"" dialog. Though I keept the error when
address changed.

closes #5870

### DIFF
--- a/packages/frontend/src/components/LoginForm.tsx
+++ b/packages/frontend/src/components/LoginForm.tsx
@@ -29,9 +29,15 @@ const CertificateChecks = {
 type LoginProps = React.PropsWithChildren<{
   credentials: Credentials
   setCredentials: (credentials: Credentials) => void
+  /** whether editing existing account */
+  isEdit?: true
 }>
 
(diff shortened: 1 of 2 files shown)


Simon-Laux:
slight issue with this is, that we don't anymore/yet have a styling for disabled DeltaInput fields, so it doesn't look disabled at first glance. But that can be fixed in a follow-up pr.

PR #5896, opened 2025-12-22
fix: more space/height for connectivity dialog, because it can be larger since adding quota for multi transport core#7591

Simon-Laux:
dialog got larger in https://github.com/chatmail/core/pull/7630
The new height is enough for the current limit of 5 transports.

### DIFF
--- a/packages/frontend/src/components/dialogs/ConnectivityDialog.tsx
+++ b/packages/frontend/src/components/dialogs/ConnectivityDialog.tsx
@@ -63,9 +63,9 @@ function ConnectivityDialogInner() {
         <iframe
           style={{
             border: 0,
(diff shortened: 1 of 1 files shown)


nicodh:
It seems a height of 380px instead 620px is sufficient, at least in my tests:

620px:
image

380px:

image

Simon-Laux:
Ideally height would adapt, but that's not possible with the sandboxed iframe which, since I haven't found a way to access the height of its content.

I guess we could weaken the sandbox to allow us to read height from parent/host, because we kinda trust the core output to be sanitised. Though that should be done in a dedicated pr and IDK if it is worth the extra work right now.

Issue #4663, opened 2025-02-16
Mov file is not displayed in the media library


Raiden-GH:
A mov file is not displayed in the media library. In the chat it appears as a file. If you click on it, the OS plays it. 
It can be exported. If you import it back into a chat in the same profile afterwards, it is displayed normally both in the chat itself and in the media library.
This is very strange, as mov files usually work without any problems and the video in question also works if it is newly imported. The file cannot be corrupted either, as it displays correctly in DC-Android and DC-iOS if the profile was transferred from DC-Desktop.

macOS 15.3.1
DC-Desktop 1.54

![Image](https://github.com/user-attachments/assets/ce4efc11-a70f-46ea-922c-63f50964dbbf)

Simon-Laux:
These are the code places that likely need to be changed.

https://github.com/deltachat/deltachat-desktop/blob/72825b788b9a0114c70553356849276fa5ba8874/packages/frontend/src/components/attachment/Attachment.tsx#L28-L33

https://github.com/deltachat/deltachat-desktop/blob/72825b788b9a0114c70553356849276fa5ba8874/packages/frontend/src/components/attachment/Attachment.tsx#L47-L49

But note that they are ancient, maybe meanwhile there is a much better way to handle this, like ty to load it and register an on error event listener on the `<video>` element. https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/error_event

nicodh:
Closed as duplicate of #4779

nicodh:
@Raiden-GH can you provide a sample file? Especially for the export - import stuff I would like to reproduce the behaviour you described

Raiden-GH:
Unfortunately, I can't send the video because it's private. The file alone doesn't seem to be the problem. 
It was sent by a non-DC user unencrypted via an HTML email. I can reproduce the bug if I forward the email and file to another account or chat. So, somehow, the combination of HTML email and video file seems to be the problem. However, if I export and send it via a new HTML email, the bug doesn't occur. 


![Image](https://github.com/user-attachments/assets/79fc3791-3398-4fd3-b30f-4df7370c5abb)

WofWca:
This might have to do with incorrect MIME detection by whoever sets `Message.fileMime`.

> If you import it back into a chat in the same profile afterwards, it is displayed normally both in the chat itself and in the media library.

I would guess that this is because resending it sets a different MIME type than originally.

[...]

PR #5900, opened 2026-01-01
fix: all "image" pickers to accept same formats

WofWca:
...and also use the translated "Image" for the filter.

Some of the "Image" pickers would accept e.g. `.jpg`,
but not `.jpeg`. Let's unify them.

Related:
- https://support.delta.chat/t/extend-upload-filter-with-all-rendered-image-formats/4539?u=wofwca.
- https://github.com/deltachat/deltachat-desktop/issues/4663.


nicodh (approved):
Makes sense!

This adds jpeg and apng as new formats that should be accepted. Both work fine in Chrome

nicodh:
Shouldn't we use the constant also here https://github.com/deltachat/deltachat-desktop/blob/main/packages/frontend/src/components/message/MessageListAndComposer.tsx#L92
?

WofWca:
I think you're right. Those are for the "drop" event. I'll make another MR.

Issue #5899, opened 2025-12-25
malformed WebXDC app can crash the renderer, break a chat


feld:
- Operating System: FreeBSD
- Delta Chat Version: 2.33.0
- Expected behavior: ?? just doesn't crash
- Actual behavior: permanently breaks a chat
- Steps to reproduce the problem: attach a malformed webxdc app

- Logs:

```
2025-12-25T21:02:59.252Z	main/deltachat/webxdc 	ERROR	"\nfile:///usr/local/share/deltachat-desktop/bundle_out/index.js:4870:17,\nprocess.processTicksAndRejections (node:internal/process/task_queues:105:5),\nasync AsyncFunction.<anonymous> (node:electron/js2c/browser_init:2:53695)"	"failed to load webxdc icon for:"	{"code":-1,"message":"unable to locate the end of central directory record"}
2025-12-25T21:02:59.317Z	renderer/hooks/useFetch	ERROR	[]	"error while executing fetch"	{"code":-1,"message":"unable to locate the end of central directory record"}
2025-12-25T21:02:59.321Z	main/deltachat/webxdc 	ERROR	"\nfile:///usr/local/share/deltachat-desktop/bundle_out/index.js:4870:17,\nprocess.processTicksAndRejections (node:internal/process/task_queues:105:5),\nasync AsyncFunction.<anonymous> (node:electron/js2c/browser_init:2:53695)"	"failed to load webxdc icon for:"	{"code":-1,"message":"unable to locate the end of central directory record"}
[...]

PR #5902, opened 2026-01-02
fix: crash on adding invalid .xdc to draft

WofWca:
Fixes https://github.com/deltachat/deltachat-desktop/issues/5899.

This has apparently been introduced in
e701aec874c7c3e7363068d2f7395345aa2d37ba.

The issue is that even if the new state is the same
as per `Object.is`,
> React may still need to call your component
> before skipping the children

so we have to perform this equality check ourself.

https://react.dev/reference/react/useState#setstate-caveats.

I have confirmed that this fixes the bug.

I suspect that I might have planted similar bugs in the codebase.
I am not really sure how to search for them,
except maybe some complex regex.

### DIFF
--- a/packages/frontend/src/components/attachment/messageAttachment.tsx
+++ b/packages/frontend/src/components/attachment/messageAttachment.tsx
@@ -247,7 +247,10 @@ export function DraftAttachment({
     isViewTypeWebxdc ? [selectedAccountId(), attachmentIdToLoad] : null
   )
   const hasFileNameChanged = useHasChanged2(attachment.fileName)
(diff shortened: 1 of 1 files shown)


nicodh (approved):
Reviewed by reading & testing

nicodh:
> This has apparently been introduced in [e701aec](https://github.com/deltachat/deltachat-desktop/commit/e701aec874c7c3e7363068d2f7395345aa2d37ba).

Well the real cause is the fact that our draft ID can change even if the draft is not changed. We should have a fixed draft ID from core to avoid all the workarounds...

PR #5903, opened 2026-01-03
fix: recognize more formats as img on drop event

WofWca:
Follow-up to e2ecf678341cc3d408cd6db4e469bc093c43e9f3
(https://github.com/deltachat/deltachat-desktop/pull/5900).


WofWca:
OK, actually this might be to good for `.gif`s and `.apng`s, because those should get `viewType === 'Gif'` probably, but it's anyway probably better than them getting `'File'` `viewType`.

https://github.com/deltachat/deltachat-desktop/blob/86577f081566f62eed22f8d87f29aed280913701/packages/frontend/src/components/message/MessageListAndComposer.tsx#L158

https://github.com/deltachat/deltachat-desktop/blob/6ce1676bf9dac405a53072a8217c047cde91ba18/packages/frontend/src/hooks/chat/useDraft.ts#L345-L366

WofWca:
Either way, this doesn't seem to affect behavior? I tried `.gif`s and `.webp`s, and they are displayed in the draft and in the message list the same way with and without this change.

PR #5905, opened 2026-01-03
fix: allow Escape key to navigate back to the chat list

omidz4t:
### Overview
This change makes the **Escape** key work as a "Back" button. It helps users navigate the app faster, especially on small screens where the chat list is hidden while a chat is open.

### What I did:
*   **Unified Navigation:** I mapped the `Escape` key to a new global action called `Chat_Unselect`. Now, pressing `Escape` behaves exactly like clicking the "Back" arrow in the top corner.
*   **Smart Handling in Composer:** I updated the message area so that `Escape` is "smart."
    *   **First press:** Closes the emoji picker, app picker, or cancels message editing/replying.
    *   **Second press:** Navigates back to the chat list.
*   **Fixed Focus Issues:** I removed an old piece of code that was forcing the cursor back into the text box when `Escape` was pressed. This allows the chat window to close properly.
*   **Context Awareness:** I ensured that `Escape` only triggers navigation if there are no open dialogs (like Settings or Profile), so it doesn't interfere with closing windows.

### Why I did this:
Previously, the `Escape` key only worked for small things like clearing search or canceling a reply. By making it a general "Back" button, the app feels more natural and consistent with modern desktop standards, providing a much better experience for users on small windows or laptops.

---
*This PR was prepared with the assistance of AI, under human supervision, and has been manually tested for correctness.*

### DIFF
--- a/packages/frontend/src/components/composer/Composer.tsx
+++ b/packages/frontend/src/components/composer/Composer.tsx
@@ -213,68 +213,68 @@ const Composer = forwardRef<
     messageEditing.isEditingModeActive || draftIsLoading
       ? null
       : async () => {
-          if (chatId === null) {
-            throw new Error('chat id is undefined')
-          }
-          if (!(draftState.text.length > 0) && !draftState.file) {
-            log.debug(`Empty message: don't send it...`)
-            return
-          }
-
-          const preSendDraftState = draftState
-          const sendMessagePromise = sendMessage(accountId, chatId, {
-            text: draftState.text,
-            file: draftState.file || undefined,
-            filename: draftState.fileName || undefined,
-            quotedMessageId:
-              draftState.quote?.kind === 'WithMessage'
-                ? draftState.quote.messageId
-                : null,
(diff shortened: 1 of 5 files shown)


nicodh:
> I agree with the idea, but the MR doesn't look merge-able so far. Is it meant as a prototype?

You mean the merge conflicts?

I'm more concerned about changing the behaviour of escape in chatview on non small views. I don't see why we should need it there. Except if the focus would land in the chatlist and you could immediately switch to other chats with key up/down.

But for smaller screens this PR it makes sense imo

Edit: and if we merge it then better before #5909 because that would have many merge conflicts then

WofWca:
> You mean the merge conflicts?

No I mean the code. e.g. the fact that `ChatList_ExitSearch` is now not working on this MR.

nicodh:
Closed in favour of #5920 which is based on this PR

nicodh:
Closed in favour of the followup #5920

Issue #5779, opened 2025-11-27
"Ctrl + Up to quote" shortcut doesn't work


WofWca:
- Operating System (Linux/Mac/Windows/iOS/Android): All
- Delta Chat Version: latest main

This has probably been introduced in https://github.com/deltachat/deltachat-desktop/pull/5713.

Issue #5682, opened 2025-11-05
Ctrl + Shift + F shortcut doesn't work on latin keyboard layout


WofWca:
- Operating System (Linux/Mac/Windows/iOS/Android): All
- Delta Chat Version: latest main

The bug has been introduced in f3630f587d4b109e0144d9917fee5c7235687cc9 (https://github.com/deltachat/deltachat-desktop/pull/5667).

PR #5909, opened 2026-01-05
Fix keyboard zoom shortcuts

### LINKED ISSUE #5908: Zoom shortcuts fail in Fullscreen view
Pressing Ctrl/Meta with + does not zoom into the image but zooms only the buttons in full screen view.

And pressing Ctrl/Meta with - on a german keyboard layout opens the Shortcut help page, since the evt.code equals Slash not -

The latter happens in any view not only in FullScreen dialogs

nicodh:
resolves #5908

- another try to solve the evt.key vs evt.code problem on non-standard keyboard layouts
- use own zoom shortcuts handler in FullScreen views since the built-in electron zoom handler fails

### DIFF
--- a/packages/frontend/src/hooks/useZoomKeyboardShortcuts.ts
+++ b/packages/frontend/src/hooks/useZoomKeyboardShortcuts.ts
@@ -6,6 +6,9 @@ import type React from 'react'
  * Hook to handle Ctrl+Plus (with fallback to 0), Ctrl+Minus,
  * and Ctrl+0 keyboard shortcuts for zooming.
  * Returns refs to connect to react-zoom-pan-pinch's TransformWrapper utils.
+ * Captures and `preventDefault()`s the respective shortcuts,
+ * primarily because the same shortcuts are used
+ * to change the zoom level of the entire app.
  */
 export function useZoomKeyboardShortcuts() {
   const zoomInRef = useRef<(() => void) | null>(null) as React.RefObject<
@@ -20,7 +23,7 @@ export function useZoomKeyboardShortcuts() {
 
   useEffect(() => {
     const handleKeyDown = (ev: KeyboardEvent) => {
-      if (ev.ctrlKey || ev.metaKey) {
+      if ((ev.ctrlKey || ev.metaKey) && !ev.altKey) {
         if (ev.key === '+' || ev.key === '=') {
           ev.preventDefault()
           ev.stopPropagation()


packages/frontend/src/hooks/useZoomKeyboardShortcuts.ts
```diff
@@ -0,0 +1,47 @@
...
+  const zoomOutRef = useRef<(() => void) | null>(null) as React.RefObject<
+    (() => void) | null
+  >
+  const zoomResetRef = useRef<(() => void) | null>(null) as React.RefObject<
+    (() => void) | null
+  >
+
+  useEffect(() => {
+    const handleKeyDown = (ev: KeyboardEvent) => {
+      if (ev.ctrlKey || ev.metaKey) {
```
WofWca:
```suggestion
      if ((ev.ctrlKey || ev.metaKey) && !ev.altKey) {
```

But I did not include `&& !ev.shiftKey` because IDK maybe some layouts need Shift to get these `key`s.

packages/frontend/src/hooks/useZoomKeyboardShortcuts.ts
```diff
@@ -0,0 +1,47 @@
+import { useEffect, useRef } from 'react'
+
+import type React from 'react'
+
+/**
+ * Hook to handle Ctrl+Plus (with fallback to 0), Ctrl+Minus,
+ * and Ctrl+0 keyboard shortcuts for zooming.
+ * Returns refs to connect to react-zoom-pan-pinch's TransformWrapper utils.
+ */
```
WofWca, line 12:
```suggestion
 * Captures and `preventDefault()`s the respective shortcuts,
 * primarily because the same shortcuts are used
 * to change the zoom level of the entire app.
 */
```

WofWca (approved):
Regarding the "zoom shortcuts" part of this MR:
looks alright and works, at least for me. The feature itself makes sense. The scope where this event catching is taking place is limited, so even if it's buggy, it should not affect much.

Regarding another rewrite of the shortcut handling code: I do not hope that it's going to be bug-free (and the code doesn't look like it either), but seems to work for me. All shortcuts are working, and bugs like https://github.com/deltachat/deltachat-desktop/issues/5779 or https://github.com/deltachat/deltachat-desktop/issues/5682 did not get re-introduced. I did not test DVORAK layout though. So let's just roll with it.

packages/frontend/src/hooks/useZoomKeyboardShortcuts.ts
nicodh: I don't understand this suggestion and the comment? How is this related to Shift and why might it be needed?

WofWca: You don't want to execute the shortcut when Alt is pressed. The same would go for Shift, but shift might be necessary on some layouts to get the required character, `key`, e.g. you'd actually have to press `Ctrl + Shift + =` to get a `plus`. For that reason you are handling both: `ev.key === '+' || ev.key === '='`. But maybe on some layouts this is the same for `Shift`.

nicodh: Yes Shift might be (and is) needed. On german keyboard layout the "=" needs Shift.

PR #5910, opened 2026-01-06
feat: move freely-selected emoji to end

r10s:
android and iOS both show freely-selected, random emoji at the end of the list, but even without these examples, showing them at the end is more logical and consistent, as other emoji are also not shown at the beginning, but at its place - which is the end as '...' is at the end as well.

moreover, this leave the standard reactions always at the same place.

apart from that, it just looks better :)

a minor, sure, but annoys me since forever :)

before / after:

Screenshot 2026-01-06 at 16 21 52 &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; Screenshot 2026-01-06 at 16 21 38


WofWca (commented):
I agree, this looks better.
Code-wise also looks trivial.

PR #5914, opened 2026-01-07
test: fix flakiness: don't "cache" `Locator`s

WofWca:
Playwright seems to hold the old element inside a Locator
for some time, which sometimes leads to `expect()`s
falsely passing, and causing really hard-to-debug
test failures down the line.
I believe this has not been happening in earlier Playwright versions.


WofWca:
OK, actually it seems that the flakiness has been resolved. I no longer observe it in https://github.com/deltachat/deltachat-desktop/pull/5668.

So this MR is no longer needed, it seems.

IDK if it's due to 5cc9a14a197b75095ab9d544ff1a17b1bae1ec24 (https://github.com/deltachat/deltachat-desktop/pull/5690) or a Playwright upgrade.

PR #5916, opened 2026-01-09
Switch profile for sharing

### LINKED ISSUE #5840: in "Forward" allow selecting the target profile
this issue is about adding the option to forward messages to another profile:

- by default, everything should stay as is - same forward dialog, same list, same profile

- in addition, add the own profile avatar tappable to the title bar - not yet totally sure where, upper right corner, ("close" stays mostright) may be nice, as the focus stays "Forward to", and also a logic read, "Forward to", "Profile", "Chat"

- tapping the avatar shows a menu with the profiles

- on selection, the chatlist changes to the new profile

- on selecting a chat for a different profile, a full profile switch is required then (so also in main window) as we used to show the destination chat on forwarding and ask "Forward message to xxx?" as usual 

alternatively, a three dot menu could do as well, esp. if easier for the beginning

what we do not want to have here, is a beforehand choice as discussed when sharing from external on mobile (there, it is only discussed for share-to-delta, not for forwarding, where defaulting to same is often what one wants)

cmp https://github.com/deltachat/deltachat-ios/issues/2912 and https://github.com/deltachat/deltachat-android/issues/4055

nicodh:
resolves #5840

image

Note 1: Although we use the wording "profile" instead "account" now, it is hard to change that in the code since there are ~1600 occurences of "account". So I used profile only in he e2e tests.

Note 2: In some languages "Switch profile" can get pretty long like in bulgarian: "Превключване към друг акаунт" - but I decided in doubt it is more important to see the profile switcher than the filter input field

### DIFF
--- a/packages/e2e-tests/tests/forward-message.spec.ts
+++ b/packages/e2e-tests/tests/forward-message.spec.ts
@@ -1,4 +1,5 @@
 import { expect, type Page } from '@playwright/test'
+import path from 'path'
 
 import {
   getUser,
@@ -26,6 +27,9 @@ let existingProfiles: User[] = []
 const numberOfProfiles = 2
 
 let page: Page
+
+const fixturesPath = path.join(import.meta.dirname, '..', 'fixtures')
+const zipPath = path.join(fixturesPath, 'test.zip')
 
 test.beforeAll(async ({ browser, isChatmail }) => {
   const contextForProfileCreation = await browser.newContext()
@@ -249,7 +253,7 @@ test('forward message to another account', async () => {
   ).toBeVisible()
 })
 
-test('forward message with webxdc attachment', async () => {
+test('forward message with file attachment', async () => {
   const userA = getUser(0, existingProfiles)
   const userB = getUser(1, existingProfiles)
 
@@ -338,4 +342,67 @@ test('forward message with webxdc attachment', async () => {
     .locator('.message')
     .filter({ has: forwardedWebxdcMessage })
   await expect(forwardedMessage).toContainText('Forwarded')
+
+  // Now test forwarding a zip file attachment
+  const fileChooserPromise = page.waitForEvent('filechooser')
+  await page.getByTestId('open-attachment-menu').click()
+  await page.getByRole('menuitem', { name: 'File' }).click()
+
+  // Upload the zip file
+  const fileChooser = await fileChooserPromise
+  await fileChooser.setFiles(zipPath)
+
+  // Wait for the file to appear in the draft
+  const attachmentDraftPreview = page.locator(
+    '.attachment-quote-section .message-attachment-generic'
+  )
+  await expect(attachmentDraftPreview).toBeVisible()
+  await expect(attachmentDraftPreview).toContainText('test.zip')
+
+  // Add a message text
+  await page
+    .locator('textarea.create-or-edit-message-input')
+    .fill('Test file for forwarding')
+
+  // Send the message to Saved messages
+  await page.locator('button.send-button').click()
+
+  // Verify the file message appears in the chat
+  const fileMessage = page.locator('.message.outgoing').last()
+  await expect(fileMessage).toBeVisible()
+  await expect(fileMessage.locator('.msg-body')).toContainText('test.zip')
+
+  // Right-click on the file message and forward it
+  await fileMessage.click({ button: 'right' })
+  await page.locator('[role="menuitem"]').filter({ hasText: 'Forward' }).click()
+
+  // Wait for the SelectChat dialog to appear
+  await expect(page.getByRole('dialog')).toContainText('Forward to')
+
+  // Click the switch account button to forward to another account
+  await page.getByTestId('switch-account-button').click()
+  await expect(page.getByTestId('select-account-dialog')).toBeVisible()
+
+  // Click on user A's account (forward back to original account)
+  await page.getByTestId(`account-select-${userA.id}`).click()
+  await expect(page.getByRole('dialog')).toContainText('Forward to')
+
+  // Forward to Saved Messages
+  await page
+    .getByRole('dialog')
+    .locator('.chat-list-item')
+    .filter({ hasText: 'Saved Messages' })
+    .click()
+
+  // Verify the forwarded file message is visible
+  const forwardedFileMessage = page.locator('.message').last()
+  await expect(forwardedFileMessage).toBeVisible()
+  await expect(forwardedFileMessage.locator('.msg-body')).toContainText(
+    'test.zip'
+  )
+  await expect(forwardedFileMessage.locator('.msg-body .text')).toContainText(
+    'Test file for forwarding'
+  )
+  // Verify it's marked as forwarded
+  await expect(forwardedFileMessage).toContainText('Forwarded')
 })


WofWca:
> it is hard to change that in the code since there are ~1600 occurences of "account". So I used profile only in he e2e tests.

I personally don't think that we should be changing the wording in the code base.

packages/frontend/src/components/dialogs/ForwardMessage/index.tsx
```diff
@@ -21,56 +28,158 @@ type ForwardMessageProps = {
...
   const tx = useTranslationFunction()
   const { openDialog } = useDialog()
   const { selectChat } = useChat()
   const { forwardMessage, jumpToMessage } = useMessage()
 
+  useEffect(() => {
+    getConfiguredAccounts().then(accounts => {
+      setHasMultipleAccounts(accounts.length > 1)
+    })
+  }, [])
```
WofWca:
```suggestion
  const configuredAccountsFetch = useFetch(getConfiguredAccounts, [])
  const hasMultipleAccounts = configuredAccountsFetch.result?.ok
    ? configuredAccountsFetch.result.value.length > 1
    : false
```

packages/frontend/src/components/dialogs/ForwardMessage/index.tsx
```diff
@@ -21,56 +28,158 @@ type ForwardMessageProps = {
 export default function ForwardMessage(props: ForwardMessageProps) {
   const { message, onClose } = props
 
-  const accountId = selectedAccountId()
+  const currentAccountId = selectedAccountId()
+  const [targetAccountId, setTargetAccountId] = useState(currentAccountId)
+  const [hasMultipleAccounts, setHasMultipleAccounts] = useState(false)
```
WofWca:
```suggestion
```

packages/frontend/src/components/chat/ChatList.tsx
```diff
@@ -729,8 +729,11 @@ function translate_n(
 }
 
 /** functions for the chat virtual list */
-export function useLogicVirtualChatList(chatListIds: number[]) {
-  const accountId = selectedAccountId()
+export function useLogicVirtualChatList(
+  chatListIds: number[],
+  overrideAccountId?: number
+) {
+  const accountId = overrideAccountId ?? selectedAccountId()
```
WofWca, line 736: This hook is not able to handle `accountId` changing. Namely, `chatCache` is keyed by `chatId`, but `chatId`s can be the same across accounts, which leads to the chat list showing chats from different accounts when switching accounts in the "forward" dialog.

I suggest putting as much as possible behind React `key={accountId}`.

packages/frontend/src/components/dialogs/SelectAccountDialog.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  return (
+    <Dialog width={350} onClose={onClose} dataTestid='select-account-dialog'>
+      <DialogHeader onClose={onClose} title={tx('switch_account')} />
+      <DialogBody className={styles.dialogBody}>
+        {loading ? (
+          <div className={styles.loading}>{tx('loading')}</div>
+        ) : (
+          <ul className={styles.accountList}>
+            {accounts.map(account => (
```
WofWca:
```suggestion
        {accountsFetch.loading ? (
          <div className={styles.loading}>{tx('loading')}</div>
        ) : !accountsFetch.result.ok ? (
          <div>
            {tx('error_x', unknownErrorToString(accountsFetch.result.err))}
          </div>
        ) : (
          <ul className={styles.accountList}>
            {accountsFetch.result.value.map(account => (
```

packages/frontend/src/components/dialogs/SelectAccountDialog.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const tx = useTranslationFunction()
+  const [accounts, setAccounts] = useState<T.Account[]>([])
+  const [loading, setLoading] = useState(true)
+
+  useEffect(() => {
+    getConfiguredAccounts().then(accounts => {
+      setAccounts(accounts)
+      setLoading(false)
+    })
+  }, [])
```
WofWca:
```suggestion

  const accountsFetch = useFetch(getConfiguredAccounts, [])
```

packages/frontend/src/components/dialogs/SelectAccountDialog.tsx
```diff
@@ -0,0 +1,95 @@
+import React, { useEffect, useState } from 'react'
+
+import { avatarInitial } from '@deltachat-desktop/shared/avatarInitial'
+import { runtime } from '@deltachat-desktop/runtime-interface'
+import { getConfiguredAccounts } from '../../backend/account'
+import Dialog, { DialogBody, DialogHeader } from '../Dialog'
+import useTranslationFunction from '../../hooks/useTranslationFunction'
+
+import type { T } from '@deltachat/jsonrpc-client'
```
WofWca:
```suggestion
```

packages/frontend/src/components/dialogs/SelectAccountDialog.tsx
```diff
@@ -0,0 +1,95 @@
+import React, { useEffect, useState } from 'react'
```
WofWca:
```suggestion
import React from 'react'
```

packages/frontend/src/components/dialogs/ForwardMessage/index.tsx
```diff
@@ -1,18 +1,25 @@
-import React from 'react'
+import React, { useEffect, useState } from 'react'
```
WofWca:
```suggestion
import React, { useState } from 'react'
```

See https://github.com/deltachat/deltachat-desktop/pull/5916/changes#r2678680057

packages/frontend/src/components/dialogs/SelectAccountDialog.tsx
```diff
@@ -0,0 +1,95 @@
...
+    })
+  }, [])
+
+  const handleAccountClick = (accountId: number) => {
+    onSelect(accountId)
+    onClose()
+  }
+
+  return (
+    <Dialog width={350} onClose={onClose} dataTestid='select-account-dialog'>
```
WofWca, line 30: Have you considered reusing the contact list component for this? I think contacts are quite similar to accounts.

packages/frontend/src/components/chat/ChatList.tsx
WofWca, line 736: Another idea with slightly different functionality: instead of having this "override account" thing, instead actually switch the account as a result of the "switch profile" dialog action, and then re-open the "forward" dialog on that account, with the state (message ID and account ID) persisted on `globalThis` or something.

WofWca:
> when switching the profile, at least saved-messages' last message is not updated (add "this is a text message" to saved-messages, forward that message to another profile - the list will always show "this is a text message" - no matter which profile is selected)

This is because of https://github.com/deltachat/deltachat-desktop/pull/5916#discussion_r2678699256

r10s:
> for Note 1: Although we use the wording "profile" instead "account" now, it is hard to change that in the code since there are ~1600 occurences of "account". So I used profile only in he e2e tests.

agree - it it fine to keep "account" internally, at least for old code. also core itself talks about "account" and "transports" and not about "profiles" and "relays". this is fine, it is also that we change UI wording more often, if smth does not work out as expected or slightly changes

> Note 2: In some languages "Switch profile" can get pretty long like in bulgarian: "Превключване към друг акаунт" - but I decided in doubt it is more important to see the profile switcher than the filter input field

that seems to be a but in the translation, it seems to be much longer than needed, it literally says "Switch to another Account"/"Zu einem anderen Konto wechseln" and not "Switch profile"/"Profil wechseln". i add an issue to the translation

packages/frontend/src/components/dialogs/SelectAccountDialog.tsx
nicodh: Imo it doesn't make sense to overload the contact list just to save a 100 line dialog which is more explicit.

packages/frontend/src/components/chat/ChatList.tsx
nicodh: I agree and applied your suggestion

packages/e2e-tests/tests/forward-message.spec.ts
```diff
@@ -0,0 +1,255 @@
...
+  // paste invite link in account of userB
+  await clickThroughTestIds(page, ['qr-scan-button', 'show-qr-scan', 'paste'])
+  const confirmDialog = page.getByTestId('confirm-start-chat')
+  await expect(confirmDialog).toContainText(userA.name)
+
+  await page.getByTestId('confirm-start-chat').getByTestId('confirm').click()
+  await expect(
+    page.locator('.chat-list .chat-list-item').filter({ hasText: userA.name })
+  ).toHaveCount(1)
+})
```
WofWca, line 98: This is merely a test setup which doesn't have to do with message forwarding, so it should be in `beforeAll` (but not necessarily inline, maybe as a function call).

packages/e2e-tests/tests/forward-message.spec.ts
```diff
@@ -0,0 +1,255 @@
...
+/**
+ * Helper function to right-click on a message and click Forward
+ */
+async function openForwardDialog(pageRef: Page, messageText: string) {
+  const message = pageRef.locator('.message').filter({ hasText: messageText })
+  await message.click({ button: 'right' })
+  await pageRef
+    .locator('[role="menuitem"]')
+    .filter({ hasText: 'Forward' })
+    .click()
```
WofWca:
```suggestion
  await pageRef.getByRole('menuitem', { name: 'Forward' }).click()
```

Same for other occurrences of `locator('[role="menuitem"]')`.

packages/frontend/src/components/dialogs/SelectAccountDialog.tsx
```diff
@@ -0,0 +1,102 @@
...
+        {accountsFetch.loading ? (
+          <div className={styles.loading}>{tx('loading')}</div>
+        ) : !accountsFetch.result.ok ? (
+          <div>
+            {tx('error_x', unknownErrorToString(accountsFetch.result?.err))}
+          </div>
+        ) : (
+          <ul className={styles.accountList}>
+            {accountsFetch.result.value
+              .filter(account => account.id !== selectedAccountId())
```
WofWca: After you have switched the account once, it is impossible to go back to the current account.
I think we should either remove this filtering or check `targetAccountId` instead of `selectedAccountId()` (a.k.a. "active account").

packages/frontend/src/components/dialogs/SelectAccountDialog.tsx
```diff
@@ -0,0 +1,102 @@
...
+                    </div>
+                    <div className={styles.accountInfo}>
+                      <div className={styles.displayName}>
+                        {account.kind === 'Configured'
+                          ? account.displayName || account.addr
+                          : tx('unconfigured_account')}
+                      </div>
+                      {account.kind === 'Configured' && account.addr && (
+                        <div className={styles.emailAddress}>
+                          {account.addr}
```
WofWca:
```suggestion
                          {account.privateTag ?? account.addr}
```

Maybe not ideal and inconsistent, but this should make accounts more easily recognizable. On top of us seemingly wanting to avoid showing email addresses like fire.

The class name will need to be changed to reflect this.

packages/frontend/src/components/dialogs/SelectAccountDialog.module.scss
```diff
@@ -0,0 +1,84 @@
...
+  flex-shrink: 0;
+  width: 48px;
+  height: 48px;
+}
+
+.avatarImage,
+.avatarInitial {
+  width: 100%;
+  height: 100%;
+  border-radius: 50%;
```
WofWca:
```suggestion
  border-radius: 10%;
```

To be consistent with the main account list. Makes them easier to recognize, and I think to understand the UI

Image

packages/frontend/src/components/dialogs/SelectAccountDialog.module.scss
```diff
@@ -0,0 +1,84 @@
...
+  object-fit: cover;
+}
+
+.avatarInitial {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  color: var(--avatarLabelColor);
+  font-size: 22px;
```
WofWca:
```suggestion
  font-size: 26px;
```

Again, to be consistent with the main account list.

But then it would be nice to either reuse the styles or to cross-reference the duplicated code.

packages/frontend/src/components/dialogs/ForwardMessage/styles.module.scss
```diff
@@ -0,0 +1,27 @@
...
+.switchAccountText {
+  line-height: 1;
+  font-size: 16px;
+}
+
+.switchAccountIcon {
+  margin-inline-end: 10px;
+  transform: rotate(180deg);
+}
+
```
WofWca:
```suggestion
```

unused

packages/frontend/src/components/dialogs/SelectAccountDialog.tsx
WofWca: Another thing to consider: removing the account from the list makes the list less recognizable. Imagine if you have always had 3 accounts. But then you encounter this list and you see only two. Might throw you off a little. So yep, maybe it makes sense to show all the accounts here.

packages/frontend/src/components/dialogs/SelectAccountDialog.tsx
```diff
@@ -0,0 +1,102 @@
...
+import { unknownErrorToString } from '../helpers/unknownErrorToString'
+
+type Props = {
+  onSelect: (accountId: number) => void
+} & DialogProps
+
+export default function SelectAccountDialog({ onSelect, onClose }: Props) {
+  const tx = useTranslationFunction()
+
+  const accountsFetch = useRpcFetch(getConfiguredAccounts, [])
```
WofWca:
```suggestion
import { useFetch } from '../../hooks/useFetch'
import { unknownErrorToString } from '../helpers/unknownErrorToString'

type Props = {
  onSelect: (accountId: number) => void
} & DialogProps

export default function SelectAccountDialog({ onSelect, onClose }: Props) {
  const tx = useTranslationFunction()

  const accountsFetch = useFetch(getConfiguredAccounts, [])
```

Although this does not currently affect behavior, `useFetch` is correct here. `useRpcFetch` is about invoking the function with `BackendRemote.rpc` as `this`, which is not needed here.

https://github.com/deltachat/deltachat-desktop/blob/fad619d65fe8aab929be0a3109f065724161d146/packages/frontend/src/hooks/useFetch.ts#L298-L314

packages/frontend/src/components/dialogs/SelectAccountDialog.tsx
nicodh: Applied in 7f7da53

packages/frontend/src/components/dialogs/SelectAccountDialog.module.scss
nicodh: Applied in 7f7da53

packages/frontend/src/components/dialogs/ForwardMessage/styles.module.scss
nicodh: Applied in 7f7da53

packages/frontend/src/components/dialogs/SelectAccountDialog.tsx
nicodh: Applied in 7f7da53

packages/frontend/src/components/dialogs/SelectAccountDialog.tsx
nicodh: Applied in 7f7da53

packages/frontend/src/components/dialogs/SelectAccountDialog.module.scss
nicodh: Applied in 7f7da53

packages/e2e-tests/tests/forward-message.spec.ts
nicodh: I added a createChat method to playwrightHelper in https://github.com/deltachat/deltachat-desktop/pull/5956 so once that is merged I will use it here and move the stuff to beforeAll

packages/frontend/src/components/dialogs/ForwardMessage/index.tsx
```diff
@@ -21,56 +30,164 @@ type ForwardMessageProps = {
...
+            chat.id
+          )
+        } else {
+          await forwardMessage(currentAccountId, message.id, chat.id)
+        }
+      } catch (e) {
+        void openDialog(AlertDialog, {
+          message: unknownErrorToString(e),
+        })
+        return false
```
WofWca, line 156:
```suggestion
```

nicodh:
The e2e tests are expected to fail now, as long as https://github.com/chatmail/core/pull/7755 is not merged and released

packages/frontend/src/components/dialogs/ForwardMessage/index.tsx
nicodh: So you say it would be better not to show the user the error than shwoing it?

Since the most probable errors are file writing or copy errors it might be helpful to see the error (might be the disk is full or similar)

PR #5918, opened 2026-01-09
chore: upgrade Core to 2.37.0

nicodh:


packages/target-tauri/src-tauri/Cargo.toml
Simon-Laux, line 1: update to cargo.lock is missing

PR #5919, opened 2026-01-09
chore: stop using deprecated strings, remove dead code

r10s:
stop using the strings `pref_chats_and_media`, `add_recipients`, `verified_contact_required_explain` and remove dead code referring them

for "Chats and Media" -> "Chats" reasoning, refer to https://github.com/deltachat/deltachat-android/pull/4128


r10s:
>  There are still some residual usages of titleMembersOrRecipients,
which, however, do not affect behavior.

thanks for careful review, this was an oversight, i added a commit.

>  Are we not supposed to also remove n_recipients?

no - n_recipients reads "123 Subscribers" now and is there to stay, as the subtitle in channels uses the latter.

having an extra wording group only for emails turned out to be too much, was never completely consistent anyways, and even wrong for some months,  saying "Subscribers" to normal emails :) but no one cared, therefore, we removed it

PR #5920, opened 2026-01-10
feature: improve escape key handling

nicodh:
Based on #5905

Additionaly handles/fixes following scenarios:

- handle only composer related escape key events in composer and stop propagation
- on composer escape unselect chat only in smallScreenMode (to open the chatlist)
- on composer escape in non smallScreenMode focus the selected chat in the chat list to easily navigate up or down to other chats

### DIFF
--- a/packages/frontend/src/components/composer/Composer.tsx
+++ b/packages/frontend/src/components/composer/Composer.tsx
@@ -333,10 +333,10 @@ const Composer = forwardRef<
           setShowAppPicker(false)
           handled = true
         }
-        if (messageEditing.isEditingModeActive) {
+        if (!handled && messageEditing.isEditingModeActive) {
           messageEditing.cancelEditing()
           handled = true
-        } else if (draftState.quote) {
+        } else if (!handled && draftState.quote) {
           removeQuote()
           handled = true
         }


packages/frontend/src/components/composer/Composer.tsx
```diff
@@ -314,19 +309,53 @@ const Composer = forwardRef<
...
+          // No picker/edit mode/quote to close
+          if (smallScreenMode) {
+            // In small screen mode, unselect the chat to go back to chatlist
+            ActionEmitter.emitAction(KeybindAction.Chat_Unselect)
+          } else {
+            // Focus the chatlist item to enable arrow key navigation between chats
+            ActionEmitter.emitAction(KeybindAction.ChatList_FocusItems)
+          }
         }
+        ev.stopPropagation()
```
WofWca, line 358: What's the purpose of this?

packages/frontend/src/components/composer/Composer.tsx
```diff
@@ -314,19 +309,53 @@ const Composer = forwardRef<
...
             regularMessageInputRef.current?.focus()
           })
+        } else {
+          // No picker/edit mode/quote to close
+          if (smallScreenMode) {
+            // In small screen mode, unselect the chat to go back to chatlist
+            ActionEmitter.emitAction(KeybindAction.Chat_Unselect)
+          } else {
+            // Focus the chatlist item to enable arrow key navigation between chats
+            ActionEmitter.emitAction(KeybindAction.ChatList_FocusItems)
```
WofWca, line 355: Just to note, our focusing behavior is getting a bit convoluted IMO. Escaping from search gets you to the composer, and escaping from the composer gets you back to the chat list. Perhaps it's fine though.

packages/frontend/src/components/composer/Composer.tsx
nicodh: That's the focus of the related chat list item as described in the PR description.

WofWca: No, I mean `ev.stopPropagation()`

nicodh:
> Also IIRC @Simon-Laux has been advocating for keeping all the shortcut handling code inside of `keybindings.ts`. 

Since we want to close the App picker or the Emoji picker or the edit mode it doesn't make sense to listen to the event in keybindings and then soemhow bubble down (or use a context) to call the appropriate functions.

This way escape key events "below" Composer are all handled in Composer and nowhere else (that's why we call stopPropagation)

We don't want to catch escape key events from other origins. We could add an id to the app picker input field but the emoji picker needs another identifier since we can't change that

nicodh:
>  I discovered a bug (not introduced in this MR) while testing this where closing the emoji picker also cancels message editing

I added a fix in this PR

Simon-Laux:
> Also IIRC @Simon-Laux has been advocating for keeping all the shortcut handling code inside of keybindings.ts.

Escape key can be treated as an exception to this preference, as we never want to allow to re-map it, and it is not a keyboard "shortcut", it is a standard user action with a clear meaning.

Also, generally I'm open for different better context aware systems. `keybindings.ts` was also an attempt to build something that we can easily extend later to allow users to change their keyboard shortcuts. (by separating the action from the actual shortcut/keypress detection logic).

> I had an idea of instead opening the chat in an actual <dialog> instead so that all this is handled natively.

Sounds very wrong to me as you need tons of outside clicking (chat list, account switcher, …), but could be done for overlays like context menu, app picker and emoji picker.
Though I could be wrong in my estimation, I have not looked into `<dialog>`-elements for a while, so my understanding could be outdated or incomplete. My biggest concern about it is that it could mess up accessibility or just not work at all.

On the note of the emoji picker, we should really fork it because original author abandoned it, made an issue for that: https://github.com/deltachat/deltachat-desktop/issues/5935

WofWca:
> Sounds very wrong to me as you need tons of outside clicking (chat list, account switcher, …)

No, I mean for just the small screen mode.
What we have now is this Android-like "screen" system - chat list screen / single chat screen.

packages/frontend/src/components/composer/Composer.tsx
nicodh: But that's what a typical workflow looks like. What would you expect where to jump to when escaping search?

packages/frontend/src/components/composer/Composer.tsx
nicodh: I commented below

packages/frontend/src/components/composer/Composer.tsx
WofWca: That "escape" always moves you up in hierarchy sounds intuitive. In this case Escaping from the search should not move you to the composer.

But yep, maybe what we have now makes the most sense, I'm not sure

Simon-Laux:
> No, I mean for just the small screen mode.

Sure, if it doesn't make the code too complex and works well, then it makes sense.

PR #5921, opened 2026-01-11
chore: add python3 to nix flake, because `bin/link_core/build_and_link_local_core.py` script needs it

Simon-Laux:
related to #5917

#skip-changelog

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -73,6 +73,7 @@
 
           # coding
           pnpm
+          python3 # for bin/link_core/build_and_link_local_core.py script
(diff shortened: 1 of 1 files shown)


nicodh:
WHat's the reason to add it? It Is not neccessary to have python installed to develop desktop. Only if you want to connect to a local core using the provided script.

Simon-Laux:
The `nix develop` command should give me a dev environment where everything that is needed is installed.
If we add python scripts, then that environment should also contain python.

PR #5925, opened 2026-01-11
remove: option to disable realtime-channels

### LINKED ISSUE #5885: Remove real-time switch
see https://github.com/deltachat/deltachat-android/pull/4094 and https://github.com/deltachat/deltachat-android/pull/4094#issuecomment-3625862865

Simon-Laux:
- **remove: option to disable realtime-channels**
- **reset config key to default on startup**

closes #5885

### DIFF
--- a/packages/frontend/src/components/Settings/Advanced.tsx
+++ b/packages/frontend/src/components/Settings/Advanced.tsx
@@ -96,8 +96,6 @@ export default function Advanced({ onClose, settingsStore }: Props) {
       <SettingsHeading>{tx('pref_experimental_features')}</SettingsHeading>
       <ExperimentalFeatures />
 
-      <SettingsSeparator />
-
       {/*
         don't show it on electron yet, as the message "not available on this runtime/platform"
         would confuse users as long as tauri is not the default */}


packages/frontend/src/components/Settings/Advanced.tsx
```diff
@@ -98,12 +98,6 @@ export default function Advanced({ onClose, settingsStore }: Props) {
 
       <SettingsSeparator />
```
WofWca: This separator might not make sense now in some cases.

packages/frontend/src/ScreenController.tsx
```diff
@@ -102,6 +102,14 @@ export default class ScreenController extends Component {
...
           .setConfig(accountId, 'disable_idle', null)
           .catch(() => log.error("failed to reset 'disable_idle' config key"))
+
+        // 2025-01-11: The `webxdc_realtime_enabled` setting was removed.
+        // Revert it to the default if it was changed in the past.
+        BackendRemote.rpc
+          .setConfig(accountId, 'webxdc_realtime_enabled', null)
+          .catch(() =>
+            log.error("failed to reset 'webxdc_realtime_enabled' config key")
+          )
```
WofWca, line 112:
```suggestion
        BackendRemote.rpc
          .batchSetConfig(accountId, {
            disable_idle: null,
            webxdc_realtime_enabled: null,
          })
          .catch(() =>
            log.error('failed to reset deprecated config vars to defaults')
          )
```

Simon-Laux: Also thought about it, but combining them makes it potentially less reliable when setting one of them fails at runtime.
There is no static type-checking for config keys yet, so it is not unlikely that one or the other key may vanish without us noticing, which could break the reset of the other keys if they were combined.

packages/frontend/src/components/Settings/Advanced.tsx
```diff
@@ -98,16 +98,15 @@ export default function Advanced({ onClose, settingsStore }: Props) {
...
```
WofWca:
```suggestion
```

r10s:
ftr, the FAQ are updated as well meanwhile, summing up reasoning: https://delta.chat/en/help#who-sees-my-ip-address

PR #5926, opened 2026-01-11
Try scanning twice to get the expected message

nicodh:


WofWca (approved):
Not nice, but as long as it works.

Maybe we should investigate whether there is a real bug at play here, may be a Core bug.

PR #5929, opened 2026-01-12
refactoring: use separate function to avoid overloading

nicodh:
This refactoring is a followup to #5776 

- provide wrapper to the openContextMenuHandler to have less ambigous parameter
- change the variable names to a more meaningful wording

### DIFF
--- a/packages/frontend/src/components/chat/ChatContextMenu.tsx
+++ b/packages/frontend/src/components/chat/ChatContextMenu.tsx
@@ -76,7 +76,7 @@ function buildArchiveAndPinMenuItems(
   accountId: number,
   chats: ChatListItem[],
   tx: ReturnType<typeof useTranslationFunction>,
-  selectedChatId: number | null
+  activeChatId: number | null
 ): { pin: ContextMenuItem | null; archive: ContextMenuItem } {
   const batchAction = (fn: (chat: ChatListItem) => Promise<unknown>) => () =>
     Promise.all(chats.map(chat => fn(chat)))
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -320,70 +321,104 @@ export function useChatContextMenu(): {
...
+      event,
+      activeChat,
+      [fullChatToChatListItem(activeChat)],
+      activeChat.id,
+      true
+    )
+  }
+
+  /**
+   * @param relatedChat the chat the context menu is opened for (undefined if multiple chats are selected)
```
WofWca, line 366: Related to what?
Maybe `relevant` is better or am I misunderstanding this?

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -517,6 +553,7 @@ export function useChatContextMenu(): {
 
   return {
     activeContextMenuChatIds,
-    openContextMenu: openContextMenuHandler,
+    openContextMenu: openChatListContextMenu,
```
WofWca, line 556:
```suggestion
    openChatListContextMenu,
```

(will need further adjustments)

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -320,70 +321,104 @@ export function useChatContextMenu(): {
...
+    activeChatId: number | null
   ) => {
-    if (chatListItems.length === 0 && activeChat === undefined) {
+    // only if chatListItems contains a single chat, we need the full chat
+    const selectedChat =
+      chatListItems.length === 1
+        ? await BackendRemote.rpc.getFullChatById(
+            accountId,
+            chatListItems[0].id
+          )
```
WofWca, line 338: This delays the menu opening. Would be nice to avoid this.

packages/frontend/src/components/chat/ChatContextMenu.tsx
nicodh: Well we have a ChatContextMenu and a related Chat so I would say that is self explaining?

WofWca: If it's self-explaining then it should be just `chat`? "related" just adds confusion.

PR #5932, opened 2026-01-12
fix: remove misleading hover effect in Keybindings and update styles

r10s:
this PR removes the misleading hover effect from the keybindings dialog.

in the second commit, while on that, the layout was tweaked a big, making it more modern and more matching to overall appearance. also making the headline stand out more than the keys.

before / after:

### DIFF
--- a/packages/frontend/scss/misc/_keyboard_hint_box.scss
+++ b/packages/frontend/scss/misc/_keyboard_hint_box.scss
@@ -9,7 +9,6 @@
       border-radius: 6px;
       border: 1px var(--keybindingKeyBorder) solid;
       box-shadow: 0px 3px var(--keybindingKeyShadow);
-      color: var(--keybindingKeyFontColor);
       display: inline-block;
       font-weight: 400;
       padding: 3px 8px 1px;
--- a/packages/frontend/themes/_themebase.scss
+++ b/packages/frontend/themes/_themebase.scss
@@ -117,7 +117,6 @@ $scrollbarTransparency: 0.34 !default;
   /* Keybindings Cheatsheet */
   --keybindingHintsBoxBackground: #{$keybindingHintsBoxBackground};
   --keybindingKeyBorder: #{$keybindingKeyBorder};
-  --keybindingKeyFontColor: black;
   --keybindingKeyShadow: #{$keybindingKeyShadow};
   --keybindingActionBackground: #{$keybindingActionBackground};
   /* Floating Action Button */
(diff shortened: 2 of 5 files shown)


packages/frontend/scss/misc/_keyboard_hint_box.scss
```diff
@@ -6,21 +6,13 @@
...
       border: 1px var(--keybindingKeyBorder) solid;
       box-shadow: 0px 3px var(--keybindingKeyShadow);
       color: var(--keybindingKeyFontColor);
       display: inline-block;
-      font-size: 14px;
       font-weight: 400;
-      padding: 3px 10px;
-
-      &:hover,
-      &.pressed {
```
WofWca, line 20: Hover is not the only way to achieve this effect. As you can see, it's also the `pressed` class, which we apply programmatically when you press the relevant key.

packages/frontend/scss/misc/_keyboard_hint_box.scss
```diff
@@ -6,21 +6,13 @@
     padding: 5px 0;
 
     .key {
-      background: var(--keybindingKeyBackground);
```
WofWca, line 9: This also removes the usage of the variable without removing its declaration from `themes/`.

packages/frontend/scss/misc/_keyboard_hint_box.scss
r10s: ah, i was not aware of that, was just wondering. 

still, pressing then eg. `ctrl-n` and getting the "new chat dialog" looks odd to me - otoh `esc` does nothing with "reply to message" but closes (as expected). and pressing `cmd` and then `enter` results in a redraw error, letting enter keep appearing .pressed

i restored it, but thb, this looks a bit over-engineerd to me. do you know what the usecase is?

packages/frontend/scss/misc/_keyboard_hint_box.scss
r10s: thanks for the hint, i added a commit it remove it

packages/frontend/scss/misc/_keyboard_hint_box.scss
WofWca: > do you know what the usecase is?

I wasn't the one who introduced it, but if I were to guess, it's probably just a cool-looking thing, maybe a little helpful.

r10s:
~~just notices, dark-mode does not look good now, back to draft for a moment~~ EDIT: fixed

PR #5937, opened 2026-01-14
perf: open WebXDCs a little faster

WofWca:
- **perf: load WebXDCs a little faster**
- **perf: open WebXDCs even faster (a little)**


WofWca:
I didn't mention: hypothetically this can lead to the app icon not being shown on the "Source code" menu item sometimes.

PR #5940, opened 2026-01-14
fix: break long words in dialogs

### LINKED ISSUE #5939: Long names are not truncated in confirm dialogs
Image

Image

nicodh:
Resolves #5939

image

This also has the advantage that even very long names are completely visible


nicodh:
> I had to deal with something like this recently, although I don't fully understand all the details of how this works.
> 
> But doesn't this go against the idea of #4722, which introduced the `whitespace` class?

the whitespace class is needed to preserve stuff like line breaks. But break-word is just for long words that need to break to fit into the parent container

PR #5941, opened 2026-01-14
fix: remove email from profile

### LINKED ISSUE #5938: Remove email address from profile
remove the email address unconditionally from the profile; for the usual reasons, to not let ppl get confused by the address etc. this is one of the last occurances btw.

the address stays in the "encryption info" unconditionally and in the chat's subtitle for email-contacts

cmp https://github.com/deltachat/deltachat-android/issues/4161 and  https://github.com/deltachat/deltachat-ios/issues/2983

nicodh:
resolves #5938

### DIFF
--- a/packages/frontend/src/components/dialogs/ViewProfile/index.tsx
+++ b/packages/frontend/src/components/dialogs/ViewProfile/index.tsx
@@ -227,15 +227,11 @@ export function ViewProfileInner({
     Number(useThemeCssVar('--SPECIAL-chatlist-item-chat-height')) || 64
 
   let displayName = contact.displayName
-  let addressLine = contact.address
   let statusText = contact.status
   let avatarPath = contact.profileImage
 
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/dialogs/ViewProfile/index.tsx
```diff
@@ -227,15 +227,11 @@ export function ViewProfileInner({
     Number(useThemeCssVar('--SPECIAL-chatlist-item-chat-height')) || 64
 
   let displayName = contact.displayName
-  let addressLine = contact.address
   let statusText = contact.status
   let avatarPath = contact.profileImage
 
-  if (isDeviceChat) {
```
r10s, line 234: ftr, i was stumbling upon that removal, but the device talk subtitle is not needed as the following description goes into even more details. it is also not shown on android/iOS. good catch!

PR #5942, opened 2026-01-14
fix: more contrast for danger color

### LINKED ISSUE #5911: Danger color needs more contrast
see https://github.com/deltachat/deltachat-desktop/pull/5776#discussion_r2592403669

nicodh:
resolves #5911 

checked on https://webaim.org/resources/contrastchecker/ to have at least 4.5:1 contrast

### DIFF
--- a/packages/frontend/themes/dark.scss
+++ b/packages/frontend/themes/dark.scss
@@ -11,7 +11,7 @@
   $bgPrimary: #222,
   $bgSecondary: #333,
   $chatListItemSelectedBg: #4c6e7d,
-  $colorDanger: #ff0c16,
+  $colorDanger: #ff3d3d,
   $colorNone: #a0a0a0,
   $colorPrimary: #42a5f5,
   $newChatButtonBg: #a0a0a0,
--- a/packages/frontend/themes/_themebase.scss
+++ b/packages/frontend/themes/_themebase.scss
@@ -4,7 +4,7 @@ $hover-contrast-change: 2%;
 // set light theme variables as fallback
 $themeDarkOrLight: light !default;
 $colorPrimary: #42a5f5 !default;
-$colorDanger: #ff0c16 !default;
+$colorDanger: #c70000 !default;
 $colorNone: #a0a0a0 !default;
 $colorLessContrast: #676767 !default;
 
--- a/packages/frontend/themes/dark_amoled.scss
+++ b/packages/frontend/themes/dark_amoled.scss
@@ -11,7 +11,7 @@
   $bgPrimary: #000,
   $bgSecondary: #222,
   $chatListItemSelectedBg: #4c6e7d,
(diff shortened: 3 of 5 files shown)


packages/frontend/themes/dark.scss
```diff
@@ -11,7 +11,7 @@
   $bgPrimary: #222,
   $bgSecondary: #333,
   $chatListItemSelectedBg: #4c6e7d,
-  $colorDanger: #ff0c16,
+  $colorDanger: #ff3d3d,
```
WofWca, line 14:
```suggestion
  $colorDanger: #ff4c4c,
```

This ensures that it also passes the contrast check when it's on the "menu item hover" background.
Although personally I'd make it a little more white still.

packages/frontend/themes/dark.scss
WofWca, line 27: Note that on the dark theme the "danger" buttons' background is still too bright to display white text on them

Image

Taking the color that we use in the light theme ( #C70000 ) would make the test pass:

Image

If you say that it's too bright, #800000 would also work.

WofWca (approved):
This MR at its current state indeed makes things better, but I'm not sure if we can say that it closes the issue.
I did not check though whether this MR follows our "color design guidelines" if we have any.

r10s:
this may have more contrast now, but it does no longer look good. the tone looks more clumsy now. dark theme looks weird as well.

the button might be regarded even as too covered and kind of disabled, which is an accessibility fail as well.

i doubt, this is an overall improvement.

iirc, we improved contrast already in the PR before, so - is there really another request or need? i can have another look, but just now busy with other things ...

packages/frontend/themes/dark.scss
nicodh: So here the contrast with white font is sufficient but not with the dark background. Seems it is not possible to achieve both.

nicodh:
> this may have more contrast now, but it does no longer look good. the tone looks more clumsy now. dark theme looks weird as well.
> 
> the button might be regarded even as too covered and kind of disabled, which is an accessibility fail as well.
> 
> i doubt, this is an overall improvement.

Seems to be a conflict between "nice" design and accessibility. Maybe we have to look how other pass solve that. As mentioned above it seems impossible even to find a button color that has enough contrast with black background and white font-color.

So closing now till we have a better solution

nicodh:
Ok, seems the colors in Signal have not enough contrast either:

image

image

r10s:
ftr, on the linked page, the "contrast 4.5" is meant for "normal text", not sure if buttons, that are otherwise outstanding, fits to that, even if they have the same pt-size as floating text

so, i doubt, these values should rule everything

WofWca:
> Ok, seems the colors in Signal have not enough contrast either:

TBH Signal is not the best example when it comes to accessibility. Screen reader and keyboard-only support there is not great, for example.

PR #5943, opened 2026-01-15
fix: WebXDC: fix 404 for weird file names

WofWca:
e.g. file names with spaces.
Add `decodeURIComponent(url.pathname)`.
Let the dog spin.

I have tested some other `.xdc`s, and the one which has assets that have spaces in them. Both work.


nicodh (approved):
Reviewed by reading

PR #5944, opened 2026-01-15
improvement: WebXDC: respond with /index.html to /

WofWca:
This is useful for apps that utilize various "routers".
It is common for websites not to have this explicit `/index.html`,
so let's support this behavior.

See
- https://support.delta.chat/t/web-server-behavior-index-html-etc/2735?u=wofwca.
- https://github.com/webxdc/webxdc_docs/issues/61.

### DIFF
--- a/packages/target-electron/src/deltachat/webxdc.ts
+++ b/packages/target-electron/src/deltachat/webxdc.ts
@@ -1122,6 +1122,9 @@ async function webxdcProtocolHandler(
   const cspAllowHttpsImgSrc = open_apps[id].internet_access
 
   let filename = url.pathname
+  if (filename === '/' || filename === '') {
+    filename = '/index.html'
+  }
   // remove leading / trailing "/"
   if (filename.endsWith('/')) {
     filename = filename.substring(0, filename.length - 1)


nicodh:
Is this change agreed with the other clients or is it just a fix for a lacking implementation in desktop.

You wrote in the forum "But this issue is mostly about the WebXDC spec"

I couldn't find anything about it in the docs neither defining it nor excluding it.

WofWca:
> I couldn't find anything about it in the docs neither defining it nor excluding it.

Yes, that's the point. I proposed to specify that in the spec by that forum post.

> Is this change agreed with the other clients or is it just a fix for a lacking implementation in desktop.

No, I haven't checked what other clients do.
It's just a thing that should make some apps work without hacks, while not breaking any other apps (well in theory it could break some hypothetical apps, but that would be really weird on their part).

nicodh:
> > I couldn't find anything about it in the docs neither defining it nor excluding it.
> 
> Yes, that's the point. I proposed to specify that in the spec by that forum post.

But why was it closed as "not planned" then?

So I think the best way would be: add a test to the webxdc-test app. Check if other clients support that behaviour and if so we implement it in desktop.

Otherwise it could result in different behaviour which should not happen. Someone develops and tests with desktop but then it turns out it doesn't work in the mobile clients.

Can you maybe describe what the concrete use case is?

WofWca:
> But why was it closed as "not planned" then?

Because we've been doing this "issue gardening" thing again and I moved it to the forum.

> add a test to the webxdc-test app.

Makes sense. I'll do that later.

> Can you maybe describe what the concrete use case is?

Simple use case: an existing multi-page app might have a link to its home page as `/` and not `/index.html`. If that is the case, then you would have to adapt the app for webxdc environment more. With this MR you don't need to.

nicodh:
Just discussed this and decided to keep it as it is since it if at all should no be fixed in desktop

r10s:
to amend, it is on purpose to be explicit here, and not try to be smart. other apps one wants to port may regard other files as main, which may work on some server implicitly. it is easy to change these lines to explicit index.html - most apps one ports anyway are single file. also, we expect the same explicit behaviour also in info.href

PR #5945, opened 2026-01-16
refactor: synchronize search input font size

nicodh:
- always use font-size 16px (the only visible change in this PR)
- remove redundant search input styles
- remove unused searchInputBackgroundColor

Followup to comment in https://github.com/deltachat/deltachat-desktop/pull/5916#pullrequestreview-3646906047

### DIFF
--- a/packages/frontend/src/components/SearchInput/styles.module.scss
+++ b/packages/frontend/src/components/SearchInput/styles.module.scss
@@ -9,22 +9,14 @@ $closeButtonSize: 24px;
   border-radius: 3px;
 }
 
-.searchInput {
-  background-color: var(--searchInputBackgroundColor);
-  border: 0;
+input.searchInput {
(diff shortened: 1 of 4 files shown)


packages/frontend/src/components/SearchInput/styles.module.scss
```diff
@@ -9,22 +9,14 @@ $closeButtonSize: 24px;
   border-radius: 3px;
 }
 
-.searchInput {
-  background-color: var(--searchInputBackgroundColor);
-  border: 0;
+input.searchInput {
```
WofWca, line 12: So which styles take precedence depends on the order in which they're defined here, I believe? Not super nice, but alright.

nicodh: no this is just to override the global style in packages/frontend/scss/misc/_search-input.scss above

PR #5948, opened 2026-01-16
style: fix poor visibility of timestamp

WofWca:
See previous attempt:
https://github.com/deltachat/deltachat-desktop/pull/4259.

Apparently the color was initially written for the light theme,
but now the color is not much different from the regular color text
in the light theme, but in the dark theme it's really pale.

Let's not be introducing another color just for this,
and let's simply use the normal color for this.

### DIFF
--- a/packages/frontend/scss/message/_message-detail.scss
+++ b/packages/frontend/scss/message/_message-detail.scss
@@ -17,7 +17,3 @@
   font-weight: 300;
   padding-inline-end: 5px;
 }
-
-.module-message-detail__unix-timestamp {
-  color: #eeefef;
-}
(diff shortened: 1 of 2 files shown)


r10s:
i think, the super-low contrast, barely visible is on purpose. could as well be white on white. it is meant for copy+pasting and for debugging.

i'd rather remove that unix-timestamp alltogether. i doubt it was ever of any worth (maybe to find something for debugging in the database

WofWca:
I created https://github.com/deltachat/deltachat-desktop/pull/5951 to remove it.

WofWca:
Superseded by https://github.com/deltachat/deltachat-desktop/pull/5951

PR #5954, opened 2026-01-19
feat: clearer 'clear chat' confirmation button

r10s:
- repeat the command on the button (instead of 'Yes')
- mark the action as 'dangerous'

before / after:

Screenshot 2026-01-19 at 16 31 17

Screenshot 2026-01-19 at 16 29 22


nicodh:
Our e2e tests are somehow flaky today although nothing really changed. But that obscured that one tests failed due to the changed wording

nicodh (approved):

PR #5955, opened 2026-01-19
feat: tweak chat menus

r10s:
this PR tweaks the chat menu order, without raising other questions, if an item is needed or not.

- tune down "encryption info" - by the previous "last" position, it was exposed very much. also, the "danger red" should be last

- do not unnecessarily interrupt menu items, so the menu items that stay together in "context menu", stay together in "three dot menu" as well

- having "archive" as the last item of a group makes sense

-  slightly adapt to android (however, menu is different in general there)


nicodh (approved):

r10s:
sorry to let you fix the e2e tests again 😳 i should learn how to fo that :)

Issue #5890, opened 2025-12-19
Improvements on "Up Arrow" to mean "Edit previous message"


dkg:
Thanks for the fix to #4820!

It's very nice to see up-arrow working to edit the last message when it's one that i just sent.

The ergonomics are still a bit clunkier than they could be though.  As of 2.33.0:

- if i have sent the last message in the chat, this works.  as soon as someone else sends another message, though, it doesn't work.  My guess is the logic is "if i sent the last message, edit it".  but it should be "edit the last message i sent"
- if i have sent the last two messages `A` and `B`, and i want to edit `A`, pressing up only works to get me to editing `B`,  i have to manually go mousey-mousey clicky-clicky to get to editing `A`.  The logic here needs a bit more sophistication about whether we are currently editing, whether the message body has changed at all, and where the cursor is in the message, to decide whether to treat an `UpArrow` keystroke in this way.  But i don't think it's particularly ambiguous.  Note that #5716 might also play a role in making this approach ergonomic though.

nicodh:
While I agree that the first issue should be fixed (see #5959 ) I doubt if the second makes sense: I would consider it a rare use case that you want to edit the second to last message and would not like to see additional complexity being added to the key handling just for that.

We could consider adding composer specific key shortcuts like "e" for edit and "r" for respond (or "reactions"?)
So you would just have to jump to a message with arrow up and the hit e or r.

WofWca:
> We could consider adding composer specific key shortcuts like "e" for edit and "r" for respond (or "reactions"?)

You mean "message-specific" then?

> So you would just have to jump to a message with arrow up and the hit e or r.

This sounds nice, though I would be careful with adding shortcuts that don't require a modifier key. This could conflict with some screen readers' shortcuts. E.g. in NVDA IIRC you use "S" and "D" to navigate between landmarks.
So if we do want such shortcuts, I'd make them `Ctrl + E` or `Ctrl + R`. Or maybe Alt. Or Ctrl + Shift.

WofWca:
https://github.com/deltachat/deltachat-desktop/pull/5959 says "partly resolves", so reopening.

nicodh:
Solved with #5963

PR #5962, opened 2026-01-24
Show shortcuts for messages

nicodh:
Related to #5890

Based on #5963

Show shortcuts for save, edit, reply, copy and delete in context menu

image

tbd: showing the shortcuts in the context menu seems helpful, but comes with a price: we have to add more key event listener logic to the ContextMenu.

Maybe in a followup - if we want this also for other context menues - we can extract that logic to a hook

### DIFF
--- a/packages/e2e-tests/tests/basic-tests.spec.ts
+++ b/packages/e2e-tests/tests/basic-tests.spec.ts
@@ -181,15 +181,15 @@ test('message menu items presence', async () => {
     .first()
   await someRegularOutgoingMessage.click({ button: 'right' })
   await expect(page.getByRole('menu').getByRole('menuitem')).toHaveText([
-    'Reply',
+    'ReplyCtrl+R',
     'Forward',
-    'Save Message',
+    'Save MessageCtrl+S',
     'React',
-    'Edit',
-    'Copy Text',
+    'EditCtrl+E',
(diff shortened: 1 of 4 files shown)


r10s (commented):
showing shortcuts in the context menu clutters appearance unnecessarily for the majority of users, degrading usability for the many.

it is helpful only for a minority, and even for them not all the time, most time it clutters there as well.

there are apps that show shortcuts in the context menu, it makes sense for more advanced apps eg. with lots of options that are hard to remember. however, for our design philosophy, we shall no do that.

it's great to have the shortcuts, and also good to show them in the main menu. ppl into shortcuts will discover them there.

nicodh:
> also good to show them in the main menu

You mean the main menu at the top bar of electron? Or in the Keybinding Cheat sheet?

r10s:
> You mean the main menu at the top bar of electron? Or in the Keybinding Cheat sheet?

both.

so, i am saying it is good enough what we do now.

and suggesting to close this PR unmerged.

in general, it may be a good idea to add more options to the main menu in case things are selected or so, but i am not suggesting to dive into that currently, it is good enough

nicodh:
Closing in favour of implementing only https://github.com/deltachat/deltachat-desktop/pull/5963

PR #5963, opened 2026-01-24
feature: shortcuts for messages

nicodh:
Adds shortcuts for save, edit, reply, copy and delete

Related to https://github.com/deltachat/deltachat-desktop/issues/5890

### DIFF
--- a/packages/frontend/src/components/message/Message.tsx
+++ b/packages/frontend/src/components/message/Message.tsx
@@ -50,7 +50,10 @@ import InvalidUnencryptedMailDialog from '../dialogs/InvalidUnencryptedMail'
 import Button from '../Button'
 import VCardComponent from './VCard'
 
-import { matchesLetterShortcut } from '../../keybindings'
+import {
+  matchesLetterShortcut,
+  matchesNonLetterShortcut,
+} from '../../keybindings'
 
 import styles from './styles.module.scss'
 
@@ -621,8 +624,7 @@ export default function Message(props: {
         !e.metaKey &&
         !e.altKey &&
         !e.shiftKey &&
-        message &&
-        e.key === 'Delete' &&
+        matchesNonLetterShortcut(e, 'Delete', 'Delete') &&
         message
       ) {
         e.preventDefault()
--- a/packages/frontend/src/keybindings.ts
+++ b/packages/frontend/src/keybindings.ts
@@ -112,8 +112,8 @@ export function matchesLetterShortcut(
  * This prevents capturing unintended shortcuts like Ctrl+- when we want Ctrl+/
  * (on German keyboards, the Slash position produces '-').
  */
-function matchesNonLetterShortcut(
-  ev: KeyboardEvent,
+export function matchesNonLetterShortcut(
+  ev: KeyboardEvent | React.KeyboardEvent<Element>,
   symbol: string,
   code: string
 ): boolean {


nicodh:
These shortcuts still need to be added to the CheatSheet but since I will refactor/redesign the CheatSheet, that will happen later in a followup

packages/frontend/src/components/message/Message.tsx
WofWca: The name is no longer right. Maybe something like `commonAttrs`.

packages/frontend/src/components/message/Message.tsx
```diff
@@ -545,6 +547,69 @@ export default function Message(props: {
...
+        enterEditMessageMode(message)
+        return
+      }
+
+      // Check if Ctrl/Cmd+"r" is pressed to reply to message
+      if (
+        isCtrlOrMetaKeyPress &&
+        matchesLetterShortcut(e, 'r') &&
+        chat.canSend &&
+        !message.isInfo
```
WofWca: These things will again get out of sync like this. We only just unified the chat list item and "three-dot" menus. Functions such as `isMessageEditable` would be nice here.

packages/frontend/src/components/message/Message.tsx
```diff
@@ -545,6 +547,69 @@ export default function Message(props: {
...
+        message.savedMessageId !== null
+      ) {
+        e.preventDefault()
+        e.stopPropagation()
+        BackendRemote.rpc.deleteMessages(accountId, [message.savedMessageId])
+        return
+      }
+
+      // Check if Ctrl/Cmd+"d" is pressed to delete message
+      if (isCtrlOrMetaKeyPress && matchesLetterShortcut(e, 'd')) {
```
WofWca: Why not the "Delete" key itself? In Browesers "Ctrl + D" is associated with "bookmark".

packages/frontend/src/components/message/Message.tsx
```diff
@@ -545,6 +547,69 @@ export default function Message(props: {
...
+        e.preventDefault()
+        e.stopPropagation()
+        BackendRemote.rpc.saveMsgs(accountId, [message.id])
+        return
+      }
+
+      // Check if Ctrl/Cmd+"u" is pressed to unsave message
+      if (
+        isCtrlOrMetaKeyPress &&
+        matchesLetterShortcut(e, 'u') &&
```
WofWca: Maybe Ctrl + Shift + S would be better? Or just Ctrl + S.

packages/frontend/src/components/message/Message.tsx
```diff
@@ -545,6 +547,69 @@ export default function Message(props: {
...
+      // Check if Ctrl/Cmd+"r" is pressed to reply to message
+      if (
+        isCtrlOrMetaKeyPress &&
+        matchesLetterShortcut(e, 'r') &&
+        chat.canSend &&
+        !message.isInfo
+      ) {
+        e.preventDefault()
+        e.stopPropagation()
+        setQuoteInDraft(message)
```
WofWca: I think it would be more handy if this was "react" because for "reply" we already have the Telegram-style Ctrl + Up in the composer.

packages/frontend/src/components/message/Message.tsx
```diff
@@ -545,6 +547,69 @@ export default function Message(props: {
     ref,
     tabIndex: rovingTabindex.tabIndex,
     onKeyDown: (e: React.KeyboardEvent) => {
+      // Handle letter shortcuts with Ctrl/Cmd modifier
+      const isCtrlOrMetaKeyPress =
+        (e.ctrlKey || e.metaKey) && !e.altKey && !e.shiftKey && message
+
+      // Check if Ctrl/Cmd+"e" is pressed to enter edit mode
```
WofWca, line 557: These comments are a little redundant?

packages/frontend/src/components/message/Message.tsx
r10s: +1, unless there are other larger apps using "ctrl+u for unsafe" i would also find one of the two more intuitive. to not accidentally toggle, maybe ctrl+shift+s

packages/frontend/src/components/message/Message.tsx
nicodh: I used showReactionsUi from useReactionsBar now

packages/frontend/src/components/message/Message.tsx
nicodh: I changed it

packages/frontend/src/components/message/Message.tsx
nicodh: I agree and changed it. Downside: Mac has no "Delete" key

packages/frontend/src/components/message/Message.tsx
nicodh: I don't agree...

packages/frontend/src/components/message/Message.tsx
WofWca: But you can execute it with `Fn + Backspace`, right?

packages/frontend/src/components/message/Message.tsx
```diff
@@ -541,10 +543,86 @@ export default function Message(props: {
...
+        message.savedMessageId !== null
+      ) {
+        e.preventDefault()
+        e.stopPropagation()
+        BackendRemote.rpc.deleteMessages(accountId, [message.savedMessageId])
+        return
+      }
+
+      // Check if Delete key is pressed to delete message
+      if (e.key === 'Delete' && message) {
```
WofWca:
```suggestion
      if (
        !e.ctrlKey &&
        !e.metaKey &&
        !e.altKey &&
        !e.shiftKey &&
        message &&
        e.key === 'Delete' &&
        message
      ) {
```

WofWca: Isn't `matchesNonLetterShortcut` supposed to be used here? (I don't know)

https://github.com/deltachat/deltachat-desktop/blob/17ab48f1a5259da5acb75d374be5312b035537bc/packages/frontend/src/keybindings.ts#L109-L127

WofWca (approved):
https://github.com/deltachat/deltachat-desktop/pull/5963#discussion_r2727518128 hasn't been fully addressed.
But otherwise this seems to be working alright.
I have not managed to come up with how "Delete" without modifier keys can conflict with anything.

packages/frontend/src/components/message/Message.tsx
r10s: often it is just `backspace`, eg. in the file manager,  just `backspace` deletes selected files.

but tbh, i would find that in an app that is also focused on text input too tricky, it will too easy goes wrong

but: is there a shortcut to open the context menu?  "ctrl+enter" does not work for me (often used on mac, windows has its own key iirc)

that seems to be far more interesting, as one has to remember only one key and can then chose the command with up/down (which is already working)

and that also solves the issue when some dedicated shortcuts are not working on some system (seems to happen all the time :)

WofWca: > it will too easy goes wrong

I'm not sure what your point is? What can go wrong?

> but is there a shortcut to open the context menu? "enter" or "ctrl+enter" (usually is not working

On non-Mac keyboards a key for this exists, left of the right Ctrl. For macOS it might be F11 or Shift + F11. People who use keyboard a lot should be familiar with it. https://apple.stackexchange.com/questions/32715/how-do-i-open-the-context-menu-from-a-mac-keyboard
And yes, it works with Delta Chat Desktop.

PR #5964, opened 2026-01-24
test: retry failing e2e tests 3 times

nicodh:


packages/e2e-tests/playwright.config.ts
```diff
@@ -15,7 +15,7 @@ export default defineConfig<TestOptions>({
   fullyParallel: true,
   /* Fail the build on CI if you accidentally left test.only in the source code. */
   forbidOnly: !!process.env.CI,
-  /* Retry on CI only */
+  /* No individual test retries! - we use file-level retries instead */
```
WofWca: Isn't it enough to specify `retries` here instead of per-test? Docs say that it should be enough.

https://playwright.dev/docs/test-retries#serial-mode

packages/e2e-tests/tests/composer.spec.ts
```diff
@@ -516,7 +518,10 @@ test('gets focused when selecting a chat', async () => {
 })
 
 test.describe('Ctrl + Up shortcut', () => {
-  test.describe.configure({ mode: 'serial' })
+  test.describe.configure({
+    mode: 'serial',
+    retries: process.env.CI ? 3 : 0,
```
WofWca: How about this one?

WofWca:
Indeed this works. Tests failed here, and re-ran: https://github.com/deltachat/deltachat-desktop/pull/5992/changes/4ffc6b03ebc402db470811151c31d2b5656d08d9

PR #5967, opened 2026-01-25
feat: calls: don't enable video if caller did not

WofWca:
Related:
- https://github.com/deltachat/deltachat-desktop/pull/5798.
- https://github.com/deltachat/calls-webapp/issues/78.

I have tested this, but I did not test the caller not enabling video (I don't have a version that supports this).

### DIFF
--- a/packages/target-electron/src/windows/video-call.ts
+++ b/packages/target-electron/src/windows/video-call.ts
@@ -28,7 +28,9 @@ export function startOutgoingVideoCall(accountId: number, chatId: number) {
     accountId,
     chatId,
     CallDirection.Outgoing,
-    {}
+    {
+      noOutgoingVideoInitially: false,
+    }
   )
 
   const jsonrpcRemote = getDCJsonrpcRemote()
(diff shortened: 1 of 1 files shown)


WofWca:
> I con confirm that if the caller disables video immediately after calling the receiver has no video option at all.

Ummm that is not supposed to happen. Are you using Core 2.39.0?

nicodh:
> > I con confirm that if the caller disables video immediately after calling the receiver has no video option at all.
> 
> Ummm that is not supposed to happen. Are you using Core 2.39.0?

Yes but the reason was that the camera can only be assigned to one instance so it was not related.

So what exactly is the expected behaviour? According to the title I expected that if the caller disables video the receiver should just receive an audio call.

WofWca:
The caller has to specify that they did not enable video by providing a boolean to the "Start Call" Core function. Which only the new iOS PR does currently.
Desktop doesn't yet have the button to start an audio-only call.

I guess we could check if the camera is enabled when the call is started, for example, to handle the case where the device doesn't have camera at all, but I suggest to do it in a future MR.

PR #5968, opened 2026-01-25
Fix ad-hoc signing error for previews

nicodh:


WofWca:
Maybe we should try upgrading Electron-builder first?

nicodh:
This seems to solve the problem for the preview build. It shouldn't break anything for the production build where CSC_IDENTITY_AUTO_DISCOVERY is not set to false

nicodh:
Closing since the changes were already merged with the electron upgrade PR #6027

PR #5970, opened 2026-01-26
chore: upgrade Electron

WofWca:
Breaking changes don't seem to affect us
https://www.electronjs.org/blog/electron-40-0#breaking-changes.
I have tested calls, checked WebXDC WebRTC isolation.

### DIFF
--- a/packages/target-electron/package.json
+++ b/packages/target-electron/package.json
@@ -82,7 +82,7 @@
     "chai": "^5.1.1",
     "chokidar": "^3.6.0",
     "debounce": "^1.2.0",
-    "electron": "^39.2.6",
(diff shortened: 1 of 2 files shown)


nicodh:
Are there reasons to update now? 2 weeks after release might be a bit early?

WofWca:
> Are there reasons to update now?

Not really.

> 2 weeks after release might be a bit early?

As long as we have time until our release.
Also I thought that the conclusion of https://github.com/deltachat/deltachat-desktop/issues/5816 was that we don't want to run out of date much.

nicodh (approved):
Since it might still take some time till the next release it's maybe better to upgrade now and find potential bugs better now than later

PR #5973, opened 2026-01-27
    tauri: fix: add some translation substitutions

WofWca:
TODO:
- [ ] Merge the base such that only one commit remains


WofWca:
This doesn't do much

PR #5975, opened 2026-01-27
fix: add `unsafe-eval` CSP for calls to fix Eruda

WofWca:
Related:
- https://github.com/deltachat/deltachat-desktop/issues/5547.
- https://github.com/deltachat/calls-webapp/pull/20.

### DIFF
--- a/packages/target-electron/src/windows/video-call.ts
+++ b/packages/target-electron/src/windows/video-call.ts
@@ -591,13 +591,13 @@ function handleCallEnd(
 }
 
 // See https://github.com/deltachat/calls-webapp/pull/20.
-// font-src is needed for Eruda (debug) build and should be removed
-// when the "calls" feature goes to production. See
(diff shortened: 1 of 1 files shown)


nicodh:
Why don't we put that behind a flag/env var?
NODE_ENV=DEBUG or DC_DEBUG ?

We need this only to execute commands in eruda, right? The logging works even without the relaxation and running commands in the console seems a very specific developer use case.

WofWca:
> We need this only to execute commands in eruda, right?

Yes.

> Why don't we put that behind a flag/env var?

IDK, we already include Eruda itself and going out of our way to say that calls are for developers only, do we need even more levels of developer-ness?

nicodh (approved):
Hopefully the comment is sufficient to not forget the removal...

But most probably we will remove Eruda then anyway

PR #5976, opened 2026-01-27
fix: also apply content protection to call window

WofWca:


WofWca:
I tested just now, and yes, it works.

PR #5977, opened 2026-01-27
fix: cleanup dialog paddings

nicodh:
Started as a fix for this

Before
image

After
image

this PR aims to cleanup and simplify the dialog stylings. No paddingBottom or paddingTop parameter needed anymore. Less code and less extra styles.

The few dialogs without footer that need a bottom padding can define that in the affected component directly.

I tested as many dialogs I could find and didn't see any quirks so far.

### DIFF
--- a/packages/frontend/scss/dialogs/_qr-code.scss
+++ b/packages/frontend/scss/dialogs/_qr-code.scss
@@ -3,7 +3,6 @@
 .qr-code-switch {
   display: flex;
   background-color: var(--globalBackground);
-  margin-bottom: 30px;
 
   button {
     @include mixins.button-reset;
--- a/packages/frontend/src/components/Dialog/styles.module.scss
+++ b/packages/frontend/src/components/Dialog/styles.module.scss
@@ -92,7 +92,7 @@ $paddingVertical: 20px;
   /* override additional top padding/margins of the first element
      in dialog content, the whole dialog already has 20px
    */
-  & > :first-child {
+  &:not(.allowTopPadding) > :first-child {
     margin-top: 0 !important;
     padding-top: 0 !important;
   }
--- a/packages/frontend/src/components/Dialog/DialogContent.tsx
+++ b/packages/frontend/src/components/Dialog/DialogContent.tsx
@@ -7,11 +7,22 @@ import styles from './styles.module.scss'
 
 type Props = PropsWithChildren<{
   className?: string
+  allowTopPadding?: boolean
 }>
 
-export default function DialogContent({ children, className }: Props) {
+export default function DialogContent({
+  children,
+  className,
+  allowTopPadding,
+}: Props) {
   return (
-    <div className={classNames(styles.dialogContent, className)}>
+    <div
+      className={classNames(
+        styles.dialogContent,
+        allowTopPadding && styles.allowTopPadding,
+        className
+      )}
+    >
       {children}
     </div>
   )
--- a/packages/frontend/src/components/dialogs/QrCode.tsx
+++ b/packages/frontend/src/components/dialogs/QrCode.tsx
@@ -198,7 +198,7 @@ export function QrCodeShowQrInner({
   return (
     <>
       <DialogBody className='show-qr-dialog-body'>
-        <DialogContent className='show-qr-dialog-content'>
+        <DialogContent className='show-qr-dialog-content' allowTopPadding>
           {svgUrl && (
             <img
               style={{
@@ -206,6 +206,7 @@ export function QrCodeShowQrInner({
                 height: '100%',
                 userSelect: 'none',
                 marginBottom: '20px',
+                marginTop: '20px',
               }}
               className='show-qr-dialog-qr-image'
               src={svgUrl}


packages/frontend/src/components/Dialog/styles.module.scss
```diff
@@ -85,22 +88,19 @@ $paddingVertical: 20px;
...
   line-height: 22px;
 
-  &.paddingTop {
-    padding-top: $paddingVertical;
-  }
-
-  &.paddingBottom {
-    padding-bottom: $paddingVertical;
+  /* no top padding except for buttons */
+  :first-child:not(button) {
```
WofWca: Is there logic behind overriding margins for everything _except_ buttons or is this just a hack? Where does this apply? Only the onboarding dialogs?

packages/frontend/src/components/Dialog/styles.module.scss
```diff
@@ -85,22 +88,19 @@ $paddingVertical: 20px;
...
+    margin-top: 0 !important;
+    padding-top: 0 !important;
```
WofWca: TBH I think this not the right approach still. Effectively forcefully settings fixed margins on the contents. Previously one at least had control over the vertical margins that they want to use. Now it's always overridden with `!important` with no way to change that except using another `!important`.

This also effectively forces you to always match the vertical margin of the first element to be equal to the `20px`, as, for example, with `.group-profile-subtitle`, and `EncryptionInfo.tsx`, because otherwise it's unbalanced.

Personal opinion: a rule of thumb should be that we specify the same vertical padding and margin for each individual element if the intent is to have equal spacing at its top and its bottom, instead of spreading the values around the element's parents and children.

packages/frontend/src/components/Dialog/styles.module.scss
```diff
@@ -2,7 +2,6 @@
 
 $headerButtonSize: 30px;
 $paddingHorizontal: 30px;
-$paddingVertical: 20px;
```
WofWca, line 5: It doesn't look like the variable needs to be removed, because the 20px value is also used in the footer?

packages/frontend/src/components/Dialog/styles.module.scss
```diff
@@ -85,22 +88,19 @@ $paddingVertical: 20px;
...
 }
 
 .dialogFooter {
   width: 100%;
   background-color: var(--bp4DialogBgPrimary);
-  padding-bottom: $paddingVertical;
+  padding-bottom: 20px;
   padding-inline-start: $paddingHorizontal;
   padding-inline-end: $paddingHorizontal;
-  padding-top: $paddingVertical;
```
WofWca, line 103: Does this removal have to do with the style cleanup? Or is this a separate change? This made the "scan QR" dialog uglier IMO.

Image

And overall for scrollable dialogs the margins seem off to me?

Image

Though yes, the "Delete message?" dialog does look better.

packages/frontend/src/components/Dialog/styles.module.scss
```diff
@@ -38,6 +37,11 @@ $paddingVertical: 20px;
     width: unset;
 
     justify-content: center;
+    padding: 0;
+    /* top needs always 20px padding but we have to make
+    sure that first element does not add additional padding
+    */
+    padding-top: 20px;
```
WofWca: The "scan QR" dialog got uglier

Image

packages/frontend/src/components/Dialog/styles.module.scss
nicodh: I agree that this is not the final  perfect approach, but it improves already for following reasons:

1. we do not need addition params be passed to dialogs
2. more general behaviour with only a few exceptions

For a "final" approach we probably have to refactor a lot of dialogs.

> specify the same vertical padding and margin for each individual element 

but the point is that those paddings and margins might sum up.

packages/frontend/src/components/Dialog/styles.module.scss
nicodh: This is a "special" case. The qr-code-switch should have a bottom-margin here. 

But just always add a top-padding to dialog content does not work as long as we have dialogs:
- with header or without
- with a tab switch or without
-- with content below the "content" section
- with footer or without
- etc.

packages/frontend/src/components/Dialog/styles.module.scss
nicodh: What is the reason to declare a variable that is only used one time?

WofWca: It's used 2 times. Once in the `.dialog`, once in the `.dialogFooter`.

packages/frontend/src/components/Dialog/styles.module.scss
nicodh: And I would say in dialog content we would not expect to care about the top padding for each content element. headlines hx and paragraphs p come with an own padding/margin which would have to be adapted without this !important

packages/frontend/src/components/Dialog/styles.module.scss
nicodh: see comment below

packages/frontend/src/components/Dialog/styles.module.scss
nicodh: I added the varaiable again

nicodh:
As discussed I added a parameter to avoid the top padding

packages/frontend/scss/dialogs/_qr-code.scss
```diff
@@ -3,6 +3,7 @@
 .qr-code-switch {
   display: flex;
   background-color: var(--globalBackground);
+  margin-bottom: 30px;
```
WofWca: Do we actually want to make the QR code even smaller?
And this also changes the design of the "scan QR" view, it adds more padding between the tabs and the video.

Image

PR #5979, opened 2026-01-28
chore: Replace once_cell with std library

joshuamegnauth54:
`LazyLock` is stable as of 1.80. Delta Chat's MSRV is 1.88.

I also changed a `Result<_, ()>` to an `Option<_>`. It was being used that way anyway.  :grin:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2007,7 +2007,6 @@ dependencies = [
  "mime",
  "mime_guess",
  "notify",
- "once_cell",
  "percent-encoding",
  "png 0.18.0",
  "rand 0.9.2",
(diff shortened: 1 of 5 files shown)


WofWca (commented):
Thanks!

The switch to std is nice (I don't remember why I haven't used `std` from the get go), but I disagree with the Result -> Option change. The fact that `Err` is `()` is only because I couldn't manage to pass the actual error to the users of `DUMMY_LOCALHOST_PROXY_URL`

joshuamegnauth54:
Alright, fair enough. :grin:  I reverted the `Option` bit.

PR #5980, opened 2026-01-28
Redesign shortcut hints

nicodh:
We have more shortcuts now and expect more to have. So maybe a good time to group the shortcuts into sections and redesign the cheat sheet a bit.

Before 

Image

Image

After

Image

Image

### DIFF
--- a/packages/frontend/scss/misc/_keyboard_hint_box.scss
+++ b/packages/frontend/scss/misc/_keyboard_hint_box.scss
@@ -45,6 +45,7 @@
       font-size: 14px;
       color: var(--colorText);
       flex: 1;
+      padding-top: 4px;
     }
 
     .shortcut-bindings {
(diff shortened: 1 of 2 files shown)


r10s:
some comments from the screenshots:
- for the in-between-headlines, just use white (default) background, that will look clearer and nicer. in other words, just use the in-between-headlines also used for the settings dialog, eg. at "Chats"
- the vertically entered headlines look lost for larger blocks, better align them to the top, so that the eye does not need to scan for the beginning. also on scrolling, that is much clearer - otherwise you see a shortcut with no clue

nicodh:
> some comments from the screenshots:

I applied your suggestions. Let me know if there are other ideas what could be improved

r10s:
- the font size is too large sometimes (see below, for the cursor keys)
- it should be always upper case (see below, cmd+M vs cmd+r)
- settings is doubled (see below)
- on mac, command-tab does not switch between chats (this is control tab, before it was shown correctly)
- use a little more padding atop of the labels, so that label and first shortcut-key appear aligned (that is a minor, it is good enough also without, if that becomes comples)

Screenshot 2026-01-30 at 13 45 08

PR #5981, opened 2026-01-28
refactor: `noUncheckedIndexedAccess` to TSConfig

WofWca:
Here are its docs: https://www.typescriptlang.org/tsconfig/#noUncheckedIndexedAccess

It happens often that we have to catch such errors manually.
I think it would be better to just force ourselves to handle the possibility
of the index access returning `undefined`,
or at the very least using the "trust me bro" operator, a.k.a. "no-null assertion",
a.k.a. the `!` (`arr[num]!.stuff()`).

TODO:
- [x] At the current state of the MR it shows waaaaay too many errors, and breaks the build.
  We need to introduce it gradually somehow, if we do decide that we want `noUncheckedIndexedAccess` at all.

### DIFF
--- a/packages/e2e-tests/tsconfig.json
+++ b/packages/e2e-tests/tsconfig.json
@@ -5,7 +5,8 @@
     "module": "nodenext",
     "target": "ES2022",
     "allowJs": false,
-    "esModuleInterop": true
+    "esModuleInterop": true,
+    "noUncheckedIndexedAccess": true
   },
   "include": ["./**/*.ts", "./**/*.js"]
 }
--- a/packages/frontend/package.json
+++ b/packages/frontend/package.json
@@ -4,7 +4,7 @@
   "type": "module",
   "license": "GPL-3.0-or-later",
   "scripts": {
-    "check:types": "tsc --noEmit",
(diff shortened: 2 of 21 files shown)


nicodh:
While in general I agree that this change makes sense, I really would like  to get the existing warnings fixed before introducing new rules.

WofWca:
I disabled the warnings in productions, so now they're basically only reported in the IDE.
Note that there is still a lot of them.
Let me know if you think that this is a good enough solution.

WofWca:
I also pushed a second commit where I address some of the warnings. Otherwise `pnpm test` fails.

nicodh (approved):
Reviewed by reading

Let's try to eliminate the warnings step by step in future

WofWca:
> Let's try to eliminate the warnings step by step in future

Do you mean that this MR does not do it the way you'd like? How would you like it instead?

nicodh:
> > Let's try to eliminate the warnings step by step in future
> 
> Do you mean that this MR does not do it the way you'd like? How would you like it instead?

No I mean the we should be able to run tsc --noEmit --noUncheckedIndexedAccess true" some day :-)

WofWca:
Let's move towards that!

PR #5983, opened 2026-01-29
fix: make menues consistent

### LINKED ISSUE #5974: inconsistent "View" menus
there are three "View" menus available in the app, having different item order, and a duplicate entry (at least on macos).

main window / webxdc window / html window:

Image
Image
Image

esp. "Developer" and "Float on Top" are at different positions, let's change that to the following for all menu items (non matching items, of course, are left out):

```
Zoom In
Zoom Out
Reset
------------------
Float On Top
Toggle Full Screen
------------------
Language        >
------------------
Developer       >     Developer Tools
                      Open Log Folder
```

that way, 
- the first 4 items are always identical
- order is same otherwise
- last item is always "Developer" (an unconditionally submenu, even if there is only one entry "Developer Tools")

please note, if there are some electron constraints, then we can think over.

for "Open Current Log File": originally, this was the way for the user to send logs to devs - but that does not have the vars and is replaced by "Settings / View Log" - so, not sure, if that is of any use now - devs will find the "current log" file also by "log folder" - and the user should not use it for sending. maybe it does even harm, as ppl easily send the wrong log.

nicodh:
resolves #5974

### DIFF
--- a/packages/target-electron/src/windows/video-call.ts
+++ b/packages/target-electron/src/windows/video-call.ts
@@ -255,7 +255,10 @@ function openVideoCallWindow<D extends CallDirection>(
           { type: 'separator' },
           { role: 'togglefullscreen' },
           { type: 'separator' },
-          { role: 'toggleDevTools' },
+          {
+            label: tx('global_menu_view_developer_desktop'),
+            submenu: [{ role: 'toggleDevTools' }],
+          },
         ],
       },
     ])


packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -440,6 +431,20 @@ export default class DCWebxdc {
...
+                      label: tx('global_menu_view_developer_desktop'),
+                      submenu: [
+                        {
+                          label: tx('global_menu_view_developer_tools_desktop'),
+                          role: 'toggleDevTools',
+                        } as MenuItemConstructorOptions,
+                      ],
+                    },
+                  ]
+                : []),
```
WofWca, line 447: No need for hiding this in a submenu if it's just one item. I think.
This is also consistent with the "calls" window menu now.

```suggestion
              ...(DesktopSettings.state.enableWebxdcDevTools
                ? [
                    { type: 'separator' } as MenuItemConstructorOptions,
                    {
                      label: tx('global_menu_view_developer_tools_desktop'),
                      role: 'toggleDevTools',
                    } as MenuItemConstructorOptions,
                  ]
                : []),
```

WofWca: Oh, the original issue says that it should be in a submenu

packages/target-electron/src/windows/video-call.ts
```diff
@@ -249,15 +249,13 @@ function openVideoCallWindow<D extends CallDirection>(
...
-          { type: 'separator' },
-          { role: 'resetZoom' },
           { role: 'zoomIn' },
           { role: 'zoomOut' },
+          { role: 'resetZoom' },
           { type: 'separator' },
-
           { role: 'togglefullscreen' },
+          { type: 'separator' },
+          { role: 'toggleDevTools' },
```
WofWca: The issue says

> (an unconditionally submenu, even if there is only one entry "Developer Tools")

Issue #5892, opened 2025-12-20
Account reordering: there is no threshold when the move starts, it starts moving the account before indicator is shown


Simon-Laux:
https://github.com/user-attachments/assets/0581361c-87cb-4c32-9327-3634775fa12e

It should not modify the account list if you haven't dragged it enough to make an indicator (blue line) appear.
In most cases this happens, you just dragged it by accident while you wanted to click on it.

- Delta Chat Version: main branch 5816000d902938e449adae9f076d7ca0f9902f1f
- Expected behavior: only move account when indicator was shown, which shows where it will be moved
- Actual behavior: slight accidental drag already moves it, without showing indicator
- Steps to reproduce the problem: slightly drag account down and observe that it moves, see video
- Screenshots: see video

WofWca:
It has to do with this piece of code

https://github.com/deltachat/deltachat-desktop/blob/2a6069eab3e299ca645b9b9024ecb294d6cd4dfe/packages/frontend/src/hooks/useAccountDragAndDrop.ts#L71-L85

https://github.com/deltachat/deltachat-desktop/pull/5590#discussion_r2461618933

TBH I don't see in what situations we'd actually want to change order when the item is dropped on itself. So, unless I'm missing something, I think we should always return if `dragIndex === dropIndex`, without requiring any other condition.
@sadraiiali

PR #5990, opened 2026-01-29
fix: account reordering when dropping on itself

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/5892.

I tested it a little. I have not found any unintuitive behavior
with this new implementation.


nicodh (approved):
Reviewed by testing & reading

PR #5992, opened 2026-01-29
fix: reduce chat React state leaking between accs

WofWca:
This might fix, or at least make it less prevalent,
the problem that sometimes you get into the "Archived chats"
folder after switching accounts, despite the fact
that neither the previous nor the new chat are archived.

I have tested this, and things work pretty much normally.

### DIFF
--- a/packages/frontend/src/contexts/ChatContext.tsx
+++ b/packages/frontend/src/contexts/ChatContext.tsx
@@ -6,7 +6,7 @@ import { BackendRemote } from '../backend-com'
 
 import type { RefObject, PropsWithChildren } from 'react'
 import type { T } from '@deltachat/jsonrpc-client'
-import { useRpcFetch } from '../hooks/useFetch'
+import { useFetch } from '../hooks/useFetch'
 import { getLogger } from '@deltachat-desktop/shared/logger'
 import { useHasChanged2 } from '../hooks/useHasChanged'
 
@@ -48,6 +48,10 @@ type Props = {
 
 export const ChatContext = React.createContext<ChatContextValue | null>(null)
 
+/**
+ * This context tries to be capable of handling `accountId` changing,
+ * i.e. it does not need to be behind `key={accountId}`.
+ */
 export const ChatProvider = ({
   children,
   accountId,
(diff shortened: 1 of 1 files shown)


packages/frontend/src/contexts/ChatContext.tsx
```diff
@@ -58,19 +62,38 @@ export const ChatProvider = ({
...
+    // to prevent data leaking between accounts, and weird race-y bugs.
+    //
+    // We usually utilize `key={accountId}` for this,
+    // but applying `key=` to a React context provider
+    // makes _all_ its descendants (as opposed just the components
+    // that depend on the context) re-render, including `AccountListSidebar`.
+    accountId === chatFetch.lingeringResult.value.accountId
+      ? chatFetch.lingeringResult.value.chat
+      : undefined
+  const chatNoLinger =
```
nicodh, line 90: Looking again at the ChatContext I wonder why we need to assign chatNoLinger and loadingChat to the context value.

It seems they are never used by the children?

WofWca: Indeed they are not. But they probably should be. At least the "loading" state. In fact I think that it's better not to use lingering values by default. But back when I introduced the "no linger" variant, we have only had the "with linger" variant, and I decided not to change the behavior d9764acbd9d401baa77fcb640dd6f97d34e72940, at least not in the same commit.

nicodh: > Indeed they are not. But they probably should be.

Then I would vote for remove them and reintroduce them when they are used. When the first PR introduced them I expected a sooner follow up that would use them...

But that's not part of this PR of course

nicodh (approved):
Reviewed by reading

This should not introduce new behaviour except for an assumed edge case which I could not reproduce but "in theory" it should be more resilient against timing/race conditions now compared to before

WofWca:
> assumed edge case which I could not reproduce

May I know what that edge case is?

nicodh:
> May I know what that edge case is?

The one you mentioned "the problem that sometimes you get into the "Archived chats""

PR #6002, opened 2026-01-30
refactor: remove pointless `key={index}`

WofWca:
From <https://react.dev/learn/rendering-lists>:

> You might be tempted to use an item’s index
> in the array as its key.
> In fact, that’s what React will use
> if you don’t specify a key at all.


nicodh (approved):
Reviewed by reading

nicodh:
Just realized this introduced an error in the console which is more annoying then a "pointless" imo

PR #6004, opened 2026-01-30
refactor: use JSX instead of function calls

WofWca:
See https://stackoverflow.com/questions/46965309/react-functional-component-calling-as-function-vs-as-component.
Some say that calling a function is faster,
but the places affected by this commit are not performance-critical.


nicodh:
So what is this about? Just point to a stack overflow discussion does not replace a reason what is fixed with this change or why it is needed

WofWca:
I think you're right, that question doesn't explain much.

This article is probably better: https://kentcdodds.com/blog/dont-call-a-react-function-component

In short, this can lead to crashes. I think we already had one like this already.

WofWca:
Force-pushed to adjust the commit message.

PR #6005, opened 2026-01-30
improvement: add arrow key nav to "React" menu

WofWca:


nicodh:
Hmm, before you could navigate with tab key - so why you think arrow key is better?

WofWca:
The "React" widget IMO is basically a "menu" pattern. In fact we already have `role='menu'` on it. For menus it is expected that you can navigate its items with arrow keys, see https://www.w3.org/WAI/ARIA/apg/patterns/menubar/.
And overall I think that this is just more natural to navigate it with arrow keys.

nicodh (approved):
Reviewed by testing & reading

PR #6008, opened 2026-02-01
Remove some unneeded heap allocs

joshuamegnauth54:
This is in no way exhaustive, but if it's accepted I can fix a few of the other simple cases I found as well.

Most of the fixes for this batch involve taking ownership of a &str to create a PathBuf or String. In the fixed cases, a borrowed Path or the &str or Cow could be used directly. In other cases, Strings were cloned to make PathBufs instead of consumed directly. SafePathBufs cloned PathBufs instead of using them directly as well.

I removed a few unneeded from_strs too. `PathBuf::from_str` is infallible so it would never return an error.

I force pushed to fix the `cargo fmt` check failing. It doesn't seem like the fix is related to this patch though.

### DIFF
--- a/packages/target-tauri/src-tauri/src/chat_background_image.rs
+++ b/packages/target-tauri/src-tauri/src/chat_background_image.rs
@@ -97,7 +97,7 @@ async fn copy_background_image_file_<R: tauri::Runtime>(
     let bg_images_dir = get_background_images_dir(&app)
         .context("failed to get the chat background images directory")?;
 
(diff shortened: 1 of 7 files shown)


WofWca (approved):
Thanks! Looks good, nice to see some potential errors gone.

The theme code looks pretty much equivalent to me, except for the changed logging, which is OK.

> I force pushed to fix the cargo fmt check failing. It doesn't seem like the fix is related to this patch though.

It is related, because you made that line shorter.

PR #6009, opened 2026-02-02
Check core rust version

nicodh:
- always wrap paths with path.join
- check node version


WofWca (approved):
Some proper path joining + extra checks. Did not test, but looks sane overall.
If this works for you, then should be good to merge.

Issue #5969, opened 2026-01-26
better scannable "Add Second Device" QR code


r10s:
there are several reports, that the QR code of the "Add Second Device" screen is hard hard to scan, esp. by some poor scanners.

the situation got worse by some iroh update, adding more data to the QR code (eg. add-second-device Qr code is more dense than most invite-links)

this issue is about **improving the showing side**, esp. when when the QR code is shown on a small, maybe dirty and dark screen

the currently used `dc_backup_provider_get_qr_svg` is anyway deprecated, so we should use `dc_create_qr_svg(dc_backup_provider_get_qr())`, which will shown more QR code when using the same svg width - also contrast to background is maximized. the borders are removed, but they're anyway no longer state-of-the-art nor good looking

the subtitle can probably be removed alltogether, it does not add much, context is pretty clear

counterpart of https://github.com/deltachat/deltachat-android/issues/4052 (has a **Screenshot**)

PR #6011, opened 2026-02-03
Show second device hints above qr code

### LINKED ISSUE #6010: Second device qr code sometimes not fully visible
Depending on the window ratio it can happen that the qr code is not fully visible:

Image

nicodh:
Show the hints above the qr code to avoid space problems

resolves #6010 

image

Very small window

image

extreme landscape ratio 

image


nicodh:
As a followup there should be a fix for #5969

Issue #3760, opened 2024-04-09
Reaction selection dialogue closes unexpectedly when a new message comes


farooqkz:
- Operating System (Linux/Mac/Windows/iOS/Android): Linux
- Delta Chat Version: v1.44.1
- Expected behavior: The reaction selection dialogue shouldn't close
- Actual behavior: It closes, of course
- Steps to reproduce the problem: 1. select a message in a chat to react to and open reaction selection dialogue to find an appropriate reaction and send it. 2. wait for a new message in the same chat to arrive. 3. see that the dialogue closes immediately upon new message arrival wasting whatever search you were making among emojies.
- Screenshots: N/A
- Logs: N/A

WofWca:
Happens on 1.44.0, so it's not because of #3751
Also FYI this isn't happening for the right-click menu, it stays open on new messages

PR #6015, opened 2026-02-04
fix: mobile: "more reactions" picker getting closed

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/5418.

Let's go for the more conservative "wheel" event instead.
But this might cause it to be kept open on touch screen devices
and touch pads.
But it is not super important
to close the "React" menu on scroll anyway.

I have tested this with a mouse for a bit. Seems to work as usual,
except it doesn't close on resize anymore.

Related:
- 26f8d98dc2f9dc6d98ed99b90629b0606d48f62e
  (https://github.com/deltachat/deltachat-desktop/pull/3817 ,
  https://github.com/deltachat/deltachat-desktop/issues/3760).

### DIFF
--- a/packages/frontend/src/components/message/MessageList.tsx
+++ b/packages/frontend/src/components/message/MessageList.tsx
@@ -280,74 +280,65 @@ export default function MessageList({
   const pendingProgrammaticSmoothScrollTo = useRef<null | number>(null)
   const pendingProgrammaticSmoothScrollTimeout = useRef<number>(-1)
 
-  const onScroll = useCallback(
-    (ev: React.UIEvent<HTMLDivElement> | null) => {
-      if (!messageListRef.current) {
-        return
-      }
-      if (scheduler.isLocked('scroll') === true) {
-        return
+  const onScroll = useCallback(() => {
+    if (!messageListRef.current) {
+      return
+    }
+    if (scheduler.isLocked('scroll') === true) {
+      return
+    }
+
+    const distanceToTop = messageListRef.current.scrollTop
+    const distanceToBottom =
+      messageListRef.current.scrollHeight -
+      messageListRef.current.scrollTop -
+      messageListRef.current.clientHeight
+
+    const isNewestMessageLoaded =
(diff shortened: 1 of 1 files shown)


nicodh:
I can't reproduce the emoji-picker-closes-on-resize issue. For me it stays open if I resize the window. Is there another way to reproduce it?

Edit: ah found it . it's only about the "more emojis picker"

packages/frontend/src/components/message/MessageList.tsx
```diff
@@ -737,6 +729,7 @@ export const MessageListInner = React.memo(
   (props: {
     onScroll: (event: React.UIEvent<HTMLDivElement>) => void
```
nicodh, line 730: Shouldn't we change the signature here, since onScroll has no argument anymore?

WofWca:
> The diff say 69 lines removed and 74 added, but it seems that the only "real" changes are:
removed

yes, it's easier to review with "hide whitespace" and word-diff coloring. Maybe I should have mentioned that.

> Is that correct?

Yes

> onScroll is triggered by the ResizeObserver?

Not quite. The 'scroll' event is fired by the browser when the scrollable element is resized. But yes, the idea is like this.

> but the behaviour should be the same when scrolling with a touchpad?

I haven't tested, but I believe this MR makes it so that the emoji picker doesn't close if you scroll with the touch pad or touch screen. I think this is fine.
Moreover, I would say that the "proper" behavior would be to disable scrolling when the emoji picker is open, i.e. to have the behavior of context menus and dialogs.

packages/frontend/src/components/message/MessageList.tsx
WofWca: I though about this, and my reasoning for keeping this is "the component provides the argument, and it's up to the parent component whether it wants to use that argument". Basically the same as for `onClick` event of regular HTML elements - they provide the argument, but we don't use it most of the time.

WofWca: It doesn't cost us much to keep this argument, it is even natural to simply reflect what React already does.

PR #6016, opened 2026-02-04
chore: upgrade calls-webapp

WofWca:
- Shows remote avatar when remote video is disabled.
- Supports new iOS calls implementation,
  https://github.com/deltachat/deltachat-ios/pull/2909.

I have tested this.


WofWca:
Going to be addressed in another MR.

PR #6017, opened 2026-02-04
feat: add audio/video select menu

nicodh:
Follow up to https://github.com/deltachat/deltachat-desktop/pull/5967

image

### DIFF
--- a/packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
+++ b/packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
@@ -6,6 +6,7 @@ import React, {
   useCallback,
   useContext,
   useMemo,
+  useId,
 } from 'react'
 import { C } from '@deltachat/jsonrpc-client'
 
@@ -651,6 +652,7 @@ function AppIcons({
 function CallButton({ chat }: { chat: T.FullChat }) {
   const tx = useTranslationFunction()
   const accountId = selectedAccountId()
+  const elId = useId()
 
   const onContextMenu = useContextMenu(
     [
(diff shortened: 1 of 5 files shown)


packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
```diff
@@ -656,3 +647,40 @@ function AppIcons({
...
+        action: () => {
+          runtime.startOutgoingVideoCall(accountId, chat.id, true)
+        },
+      },
+    ],
+    { 'aria-label': tx('start_call') }
+  )
+
+  return (
+    <Button
```
WofWca, line 682: Prefer ID instead of just a label

```suggestion
  const elId = useId()

  const onContextMenu = useContextMenu(
    [
      {
        label: tx('audio_call'),
        icon: 'phone',
        action: () => {
          runtime.startOutgoingVideoCall(accountId, chat.id, false)
        },
      },
      {
        label: tx('video_call'),
        icon: 'camera',
        action: () => {
          runtime.startOutgoingVideoCall(accountId, chat.id, true)
        },
      },
    ],
    { 'aria-labelledby': elId }
  )

  return (
    <Button
      id={elId}
```

packages/runtime/runtime.ts
```diff
@@ -133,7 +133,11 @@ export interface Runtime {
   /**
    * Initiates and conducts the video call fully, from start to end.
    */
-  startOutgoingVideoCall(accountId: number, chatId: number): void
+  startOutgoingVideoCall(
+    accountId: number,
+    chatId: number,
+    cameraEnabled: boolean
```
```suggestion
    params: { cameraEnabled: boolean }
```

Makes it more clear what this `true` / `false` mean wherever the function is used.

WofWca: Also maybe it's better to stick to `noOutgoingVideoInitially`?
Edit: I mean the name

WofWca (commented):
The code looks good, and things also work as written, but I think we should not merge this until we have Core the API to actually convey the info about whether this is a video or an audio call to the remote party, i.e. https://github.com/chatmail/core/pull/7740. Because otherwise the button is kind of misleading. You think that you're starting an audio-only call, but for the callee the camera is enabled by default.

packages/runtime/runtime.ts
nicodh: noOutgoingVideoInitially sounds for me more confusing 

would you agree on
startWithCameraEnabled ?

nicodh: ok, but then with the name below

WofWca: All are fine to me

ralphtheninja:
> Regarding the UI / UX: not sure if it's the best to have people click two times (maybe we should have two separate buttons presented right away, maybe one hidden behind the three-dot menu), but at least Signal utilizes the same approach, so this is good.

Why not a toggle button? One image is the camera with a red line/cross over it and the other state shows an enabled camera. Then you only need one click to toggle the state.

WofWca:
I though about that a little, and the problem with that is that it basically introduces a "default" mode for calls, i.e. not requiring the user to make an explicit choice. But then again, I am not sure if it's good or not.

You are still free to toggle the camera however you like after you have already started the call.

nicodh:
I remember that these issues were alreday discussed. (at least live, maybe online too somewhere)

PR #6018, opened 2026-02-04
fix: avoid double toggle fullscreen item on Mac

nicodh:


packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -431,7 +431,10 @@ export default class DCWebxdc {
                   }
                 },
               },
-              { role: 'togglefullscreen' },
+              // electron adds "Toggle Full Screen" to the menu on macOS automatically
```
WofWca: Do we know if that is supposed to happen or if it's just a bug? Because one day it might get fixed, and we're gonna have the opposite problem, of `togglefullscreen` not being present at all.

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -431,7 +431,13 @@ export default class DCWebxdc {
                   }
                 },
               },
-              { role: 'togglefullscreen' },
+              // due to bug https://github.com/electron/electron/issues/49048
+              // electron always adds "Toggle Full Screen" to the menu on macOS
+              // TODO: remove this workaround when we upgrade to electron v42
+              // or when the fix is backported
+              ...(!isMac
```
WofWca, line 438: This only fixes it for the WebXDC window but not other windows. See screenshots in https://github.com/deltachat/deltachat-desktop/issues/5974.

packages/target-electron/src/deltachat/webxdc.ts
nicodh: I added a comment

nicodh:
I tested this on MacOS and the workaround fixed the issue

PR #6019, opened 2026-02-05
chore: upgrade Core to 2.40.0

nicodh:


packages/target-electron/src/windows/video-call.ts
```diff
@@ -48,7 +48,8 @@ export function startOutgoingVideoCall(accountId: number, chatId: number) {
     const callMessageId = await jsonrpcRemote.rpc.placeOutgoingCall(
       accountId,
       chatId,
-      offer
+      offer,
+      false
```
WofWca, line 52:
```suggestion
      true
```

This is a parameter named `has_video`, which should be `true`, because above `noOutgoingVideoInitially: false,`.
But it's not a big deal because https://github.com/deltachat/deltachat-desktop/pull/6017 is coming.

nicodh: will be immediately removed as you say

PR #6022, opened 2026-02-06
Improve image resolution

### LINKED ISSUE #4914: Avatars with enhanced image quality do not work
DC-Desktop 1.56 continues to reduce new profile pictures to 256x256 and a maximum of 20 KiB. 
I've also tested it with DC-Android 1.56.1 and DC-iOS 1.54.5. With these versions, the avatar is saved in significantly better quality and more than 20 KiB. Therefore, I assume this isn't a core bug. 
If a profile is used on other devices besides DC-Desktop and the new avatar is imported into DC-Android or DC-iOS, the new profile picture will also be displayed in DC-Desktop with better quality.

macOS 15.3.2
DC-Desktop 1.56

nicodh:
resolves #4914

### DIFF
--- a/packages/frontend/src/components/ImageCropper/index.tsx
+++ b/packages/frontend/src/components/ImageCropper/index.tsx
@@ -31,9 +31,6 @@ import { copyToBlobDir } from '../../utils/copyToBlobDir'
 //
 // * on export we just cut the bounding box (targetWidth, targetHeight) around cursor position and rotate/flip it
 
-/** Scale factor for the circular clip (slightly smaller circle) */
-const CIRCLE_CLIP_SCALE = 0.9
-
 export default function ImageCropper({
   filepath,
   shape,
@@ -133,8 +130,8 @@ export default function ImageCropper({
     const scaleY = naturalHeight / displayHeight
 
     // For circle shape, crop a slightly smaller region to match the visible clip
-    const cropW = shape === 'circle' ? resultW * CIRCLE_CLIP_SCALE : resultW
-    const cropH = shape === 'circle' ? resultH * CIRCLE_CLIP_SCALE : resultH
+    const cropW = resultW
+    const cropH = resultH
 
     // Output keeps requested dimensions (full resolution)
     const outputWidth = Math.round(resultW * scaleX)
@@ -280,7 +277,7 @@ export default function ImageCropper({
     const hw = targetWidth.current / zoom.current / 2
     const hh = targetHeight.current / zoom.current / 2
     if (shape === 'circle') {
-      const r = Math.min(hw, hh) * CIRCLE_CLIP_SCALE
+      const r = Math.min(hw, hh)
       return `circle(${r}px at ${x}px ${y}px)`
     } else {
       return (
--- a/packages/frontend/src/components/ImageCropper/styles.module.scss
+++ b/packages/frontend/src/components/ImageCropper/styles.module.scss
@@ -66,20 +66,20 @@
 
 .imageCropperControls {
   position: absolute;
-  top: 6px;
+  top: 0;
   right: 0;
-  display: flex;
-  flex-direction: column;
-  gap: 6px;
+  width: 80px;
   z-index: 10;
 }
 
 .imageCropperControlsButton {
-  display: flex;
-  padding: 12px;
+  display: inline;
+  padding: 10px;
   align-items: center;
   justify-content: center;
   background: transparent;
   border: none;
   cursor: pointer;
+  width: 40px;
+  height: 40px;
 }


packages/frontend/src/components/LargeProfileImage/styles.module.scss
```diff
@@ -22,7 +19,12 @@ $imageSize: 100px;
   object-position: center;
   pointer-events: none;
   width: $imageSize;
+}
 
+/** only for non image avatars */
+span.largeProfileImageArea {
```
WofWca, line 25: Would be nice to have an explicit class to select for this.

packages/frontend/src/components/ImageCropper/index.tsx
```diff
@@ -87,11 +87,25 @@ export default function ImageCropper({
...
     }
 
+    if (!userModified.current) {
+      // User didn't modify anything, just copy the original file
+      rememberLastUsedPathPromise.then(({ setLastPath }) =>
+        setLastPath(dirname(filepath))
+      )
+      const blob_path = await copyToBlobDir(filepath)
+      onResult(blob_path)
+      onClose()
```
WofWca, line 112: This allows one to set a non-square avatar. Is this fine?

packages/frontend/src/components/LargeProfileImage/styles.module.scss
WofWca, line 25: Also, is this the only place where we need to adjust something to support transparent avatars?

WofWca (commented):
This does not address the problems that I mentioned in the issue, nor does it allow for cropping images ~~larger~~ smaller than 512x512, but maybe it's better than what we have now.

Image

nicodh:
> This does not address the problems that I mentioned in the issue

what do you mean? You only mentioned that it is not possible to set desiredWidth to 4096 which is not a UI issue but a code change.

> nor does it allow for cropping images larger than 512x512

Do you mean the result or the loaded image? I tried loading bigger images and I could crop them.
Edit: bigger images hide the tool buttons so crop & move work only with scroll (Shift Scroll) but flip is not possible
But the issue was about "enhanced image quality" not about missing UI buttons

When saving an image with rpc.setChatProfileImage core resizes & compresses it.
I saved a 3.9 MB image with size 1920x1080 and the result was an 44kb image with size 512x288px size.

So nothing related to desktop.

> but maybe it's better than what we have now.

Did you compare it? In my tests it was much better then before

packages/frontend/src/components/ImageCropper/index.tsx
nicodh: Maybe not - the user sees that only a part is displayed but he probably would expect that only the visible part of the image is saved

packages/frontend/src/components/LargeProfileImage/styles.module.scss
nicodh: You mean images with transparent parts like png? png currently only work when core has not to resize or compress them, otherwise they are converted to jpeg loosing the transparency

WofWca: Yes, I mean transparency. Your commit message seems to indicate that we want to properly display transparency, but we have a lot more avatar components, which I would assume would also need to be updated?

WofWca:
> what do you mean? You only mentioned that it is not possible to set desiredWidth to 4096 which is not a UI issue but a code change.

I am talking about this comment https://github.com/deltachat/deltachat-desktop/issues/4914#issuecomment-3048638670. Everything that I said there still applies. Even with 512px the cropper is not properly displayed, as you can see on the screenshot that I posted above in this MR. And you cannot zoom further than some amount (not sure if it's 512px).

> Do you mean the result or the loaded image? I tried loading bigger images and I could crop them.

Sorry, I meant the thing that I said in the issue. That is that you cannot make the final image smaller than 512px. If the initial image is 512x512, then you cannot zoom at all.

> Edit: bigger images hide the tool buttons so crop & move work only with scroll (Shift Scroll) but flip is not possible
But the issue was about "enhanced image quality" not about missing UI buttons

Ummm I'm not sure how you got the buttons to hide. Maybe you're talking about the vertical scroll at smaller screen heights? Is that how the buttons get "hidden"?

> Did you compare it? In my tests it was much better then before

The image quality is indeed better, but the UI and the functionality of the image cropper component is degraded. Again, see my comment  https://github.com/deltachat/deltachat-desktop/issues/4914#issuecomment-3048638670.

WofWca (commented):
I tested this for a bit. The issues that I mentioned in my comment are resolved. Though the cropper is still scrollable on smaller screens (both narrow and short), but it's bearable.
So far this seems like an overall improvement, including the unrelated changes.
Although I didn't look to closely at what exactly the code does.

packages/frontend/src/components/LargeProfileImage/styles.module.scss
nicodh: No the other ones need not to be update. The class largeProfileImageArea is only used in the component LargeProfileImage and the background-color should not be applied if it's an image holding that class (as it was before)

For a png with a transparent background the result looked like this:

Image

The change removes the ground background color, showing the image as it will look like in the profile later

r10s:
why was the layout changed? i looked more polished before:

before / after:


&nbsp;&nbsp;&nbsp;


i agree, one can think over polishing the old layout (save button is not primary, padding), but the new looks even more rough (button colors not matching other buttons, unclear area, misaligned appearing "save" button).

unless there is a technical constraint, that makes the change necessary, i would stay with the old one. in both cases we can iterate over layout in another PR

nicodh:
> why was the layout changed? i looked more polished before:
> 
 the issue was about big images mainly - then the layout was completely broken, like the screenshot of WofWca shows

nicodh:
The initial issue was the bad image resolution - the reason was, that the crop was based on the displayed image (even if the natural resolution was much bigger) - but when I solved that it caused problems in the layout so I had to fix that.

I can have a look to get it more similar to the initial layout - shouldn't be too hard

r10s:
great, having smth as before, sounds great!

only if it turns out to be too hard, let's merge as is. 
the feature "high quality avatar" then rules the layout of the avatar selector - which we also can improve given then new constraints.

nicodh:
This is the current state:

image

packages/frontend/src/components/ImageCropper/index.tsx
WofWca: Added `isSquare` check

packages/frontend/src/components/LargeProfileImage/styles.module.scss
WofWca: Oh I see. So when the image finally gets set, Core still removes transparency, so the commit 0cc3ca23cf89b1cb741f45fdcd19bf47e8682cbc only hypothetically adds transparency support.

> as it was before

I think you mean aea10696154c59022b709393a084be6e09fcd5a3.

> The change removes the ground background color, showing the image as it will look like in the profile later

That is only true with the light theme where the background is white.

packages/frontend/src/components/ImageCropper/index.tsx
```diff
@@ -30,6 +30,10 @@ import { copyToBlobDir } from '../../utils/copyToBlobDir'
 //   cursor (0, 0) is always at the top-left corner of non-rotated image
 //
 // * on export we just cut the bounding box (targetWidth, targetHeight) around cursor position and rotate/flip it
+
+/** Scale factor for the circular clip (slightly smaller circle) */
+const CIRCLE_CLIP_SCALE = 0.9
```
WofWca: Apparently because of this it's impossible to set an uncropped image now.
If you have an image that is already cropped the way you want, it will still get cropped a little, no matter how you zoom.
Edit: OK, actually it _is_ possible, but only as long as you haven't pressed any of the buttons (which is not intuitive).

nicodh: I removed the CIRCLE_CLIP_SCALE and rearranged the buttons a bit to have no overlap with the image

PR #6023, opened 2026-02-06
chore: prepare release 2.40.0

nicodh:


r10s:
there is already core 2.41.0, which esp makes disabling "Move to Delta Chat" and instructions in device message and at https://delta.chat/legacy-move easier

PR #6026, opened 2026-02-07
Add channel/group descriptions

nicodh:
Needs https://github.com/chatmail/core/pull/7829 to be merged in  core

### DIFF
--- a/packages/frontend/src/components/dialogs/ViewGroup.tsx
+++ b/packages/frontend/src/components/dialogs/ViewGroup.tsx
@@ -594,7 +594,6 @@ export function EditGroupNameDialog({
             />
           </div>
           <DeltaInput
-            key='groupname'
             id='groupname'
             placeholder={!isBroadcast ? tx('group_name') : tx('channel_name')}
             value={groupName}
@@ -621,7 +620,6 @@ export function EditGroupNameDialog({
             </p>
           )}
           <DeltaTextarea
-            key='description'
             id='description'
             placeholder={tx('chat_description')}
             value={groupDescription}


r10s:
did not test, one comment:

when tapping the info message "Chat Description changed" it is very likely that the user wants to _see_ the changed chat description.

easiest and probably best way to do that, is to open the profile, this is what [android](https://github.com/deltachat/deltachat-android/pull/4239) is doing, after short discussion

r10s:
ftr, the "Description" string is meanwhile available as `chat_description`.

for the info messages, the stock strings  DC_STR_GROUP_DESCRIPTION_CHANGED_BY_YOU and  DC_STR_GROUP_DESCRIPTION_CHANGED_BY_OTHER must be set to `chat_description_changed_by_you` resp. `chat_description_changed_by_other`


and, when tapping the info messages with type=DC_INFO_GROUP_NAME_CHANGED, the profile shall be opened directly; this is the much more probable action the user want to achieve on tapping them.

cmp. https://github.com/deltachat/deltachat-android/pull/4239 and https://github.com/deltachat/deltachat-ios/issues/3020

nicodh:
The strings are updated but there is still a wrong string 
`"[Chat description changed. To see this and other new features, please update the app]" `
coming from core ( is fixed in https://github.com/chatmail/core/pull/7939 )

packages/frontend/scss/dialogs/_group-styles.scss
```diff
@@ -78,3 +78,11 @@ input.group-name-input {
   padding-top: 6px;
   margin-bottom: 20px;
 }
+
+.group-profile-description {
+  text-align: center;
```
WofWca, line 83: Is the centered layout what we want? Looks odd IMO

image

packages/frontend/src/components/message/Message.tsx
```diff
@@ -734,6 +736,10 @@ export default function Message(props: {
         if (isWebxdcInfo) {
           // open or focus the webxdc app
           openWebxdc(message)
+        } else if (message.systemMessageType === 'GroupDescriptionChanged') {
+          openViewGroupDialog(
+            chat as T.FullChat & { chatType: 'Group' | 'OutBroadcast' }
+          )
```
WofWca: Let's not typecast.

```suggestion
        } else if (
          message.systemMessageType === 'GroupDescriptionChanged' &&
          (chat.chatType === 'Group' || chat.chatType === 'OutBroadcast')
        ) {
          openViewGroupDialog(
            chat as typeof chat & { chatType: typeof chat.chatType }
          )
```

packages/frontend/src/components/message/Message.tsx
```diff
@@ -734,6 +736,10 @@ export default function Message(props: {
         if (isWebxdcInfo) {
           // open or focus the webxdc app
           openWebxdc(message)
+        } else if (message.systemMessageType === 'GroupDescriptionChanged') {
```
WofWca: Note that this only gets executed if `message.infoContactId != null`. Maybe this should eventually be properly specified by Core.

packages/frontend/src/components/dialogs/ViewGroup.tsx
```diff
@@ -284,14 +304,20 @@ function ViewGroupInner(
   const onClickEdit = () => {
     openDialog(EditGroupNameDialog, {
       groupName,
+      groupDescription: groupDescription ?? '',
```
WofWca: We should not just have a default value here. Instead we should wait for the description to load. Maybe by making the "edit" button do nothing if we're not loaded.

packages/frontend/src/components/dialogs/ViewGroup.tsx
```diff
@@ -573,6 +615,18 @@ export function EditGroupNameDialog({
                 : tx('please_enter_channel_name')}
             </p>
           )}
+          <DeltaTextarea
+            key='description'
```
WofWca:
```suggestion
```

No point in it?

packages/frontend/scss/dialogs/_group-styles.scss
nicodh: Well I would say for short descriptions it looks weird to have them moved to the left, while Avatar, name and member counter are centered

Image

r10s:
> Is the centered layout what we want? Looks odd IMO

i missed that bit, i agree, the centered layout is odd.

we should do the same as for the contact's bio, which is also fine

> Well I would say for short descriptions it looks weird to have them moved to the left, while Avatar, name and member counter are centered

member count should be below avatar and name, so we have the centered things atop, followed by the left-aligned things  - as description and memberlist

i created a dedicated issue for that https://github.com/deltachat/deltachat-desktop/issues/6164

PR #6027, opened 2026-02-08
Update electron 40.2.1

### LINKED ISSUE #6014: "toggle full screen" appears twice
"toggle full screen" appears twice. this is a leftover of https://github.com/deltachat/deltachat-desktop/issues/5974

Image

nicodh:
resolves #6014

Also adds a fix for the mac preview build signing error

### DIFF
--- a/packages/target-electron/build/afterPackHook.mjs
+++ b/packages/target-electron/build/afterPackHook.mjs
@@ -183,11 +183,14 @@ async function deleteNotNeededPrebuildsFromUnpackedASAR(
 }
 
 async function setFuses(context) {
-  // Skip fuse flipping for temporary arch-specific builds when creating a universal build.
+  // Skip fuse flipping for temporary arch-specific builds when creating a macOS universal build.
   // electron-builder creates separate arm64 and x64 builds in *-temp directories first,
   // then merges them. If we flip fuses on these temp builds, the CodeResources signatures
   // won't match and the universal merge will fail.
-  if (context.appOutDir.includes('-temp')) {
+  const isMac =
+    context.electronPlatformName === 'darwin' ||
+    context.electronPlatformName === 'mas'
+  if (isMac && context.appOutDir.includes('-temp')) {
     console.log(
       'Skipping fuse flipping for temporary universal build:',
       context.appOutDir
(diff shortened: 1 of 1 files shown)


nicodh:
I tested this on MacOS and can confirm that both issues are fixed: the preview build can be run on MacOS and the duplicate toggle items are removed

nicodh:
Should be included in the next release

packages/target-electron/build/afterPackHook.mjs
```diff
@@ -189,6 +183,18 @@ async function deleteNotNeededPrebuildsFromUnpackedASAR(
 }
 
 async function setFuses(context) {
+  // Skip fuse flipping for temporary arch-specific builds when creating a universal build.
+  // electron-builder creates separate arm64 and x64 builds in *-temp directories first,
+  // then merges them. If we flip fuses on these temp builds, the CodeResources signatures
+  // won't match and the universal merge will fail.
+  if (context.appOutDir.includes('-temp')) {
```
WofWca: Maybe we should be more conservative here, and only check the last (or second to last) component of the path? Using `appOutDir.split(path.sep)`
Also maybe we should limit this to macOS builds?

https://nodejs.org/api/path.html#pathsep

packages/target-electron/build/afterPackHook.mjs
```diff
@@ -210,18 +216,30 @@ async function setFuses(context) {
...
   console.log('Applying electron fuses to:', appPath)
+
+  // For macOS arm64/universal builds without proper code signing (e.g. preview builds),
+  // we need to reset the ad-hoc signature after flipping fuses, otherwise the app
+  // will fail to launch with "Code Signature Invalid" errors on Apple Silicon.
+  const isMac =
+    context.electronPlatformName === 'darwin' ||
+    context.electronPlatformName === 'mas'
+  const needsAdHocReset =
+    isMac && process.env.CSC_IDENTITY_AUTO_DISCOVERY === 'false'
```
WofWca, line 234: Is this env var (`CSC_IDENTITY_AUTO_DISCOVERY`) actually relevant or is this just a hack to detect if we're doing a prod build?
It was introduced in https://github.com/deltachat/deltachat-desktop/pull/1959. The docs say that it's supposed to completely disable signing or not really?

https://github.com/deltachat/deltachat-desktop/blob/417030588db6ce01dccfaca73764d1f521cb1ff5/docs/DEVELOPMENT.md#L332-L338

nicodh:
> The version is 40.2.1

of course - changed it

packages/target-electron/build/afterPackHook.mjs
nicodh: This whole ad-hoc signing stuff is weird - the docs are unclear here:


"To disable Code Signing when building for macOS leave all the above vars unset except for CSC_IDENTITY_AUTO_DISCOVERY which needs to be set to false. This can be done by running export CSC_IDENTITY_AUTO_DISCOVERY=false.

Another way — set mac.identity to null. You can pass additional configuration using CLI as well: -c.mac.identity=null. If you are building for ARM, you likely actually want to use ad-hoc signing, in which case you should set mac.identity to -."

So it says "disable Code Signing" but also "Another way — set mac.identity to null" and then "set mac.identity to -" for ad-hoc signing. There is also no reason why it used to work and then stopped (with no changes in all relevant parts)

So I tried various stuff and finally it works now again. It should not affect the production build since CSC_IDENTITY_AUTO_DISCOVERY defaults to true.

And I added these changes to this PR since otherwise it would no be possible to verify that the electron upgrade fixes the MacOS specific problem.

PR #6029, opened 2026-02-08
chore: upgrade electron-builder

nicodh:


packages/target-electron/build/gen-electron-builder-config.js
```diff
@@ -78,6 +78,7 @@ build['asarUnpack'] = [] // ['./node_modules/@deltachat/stdio-rpc-server']
 
 build['afterPack'] = './build/afterPackHook.mjs'
 build['afterSign'] = './build/afterSignHook.cjs'
+build['npmRebuild'] = false
```
WofWca, line 81: Why is this needed?

https://www.electron.build/app-builder-lib.interface.commonconfiguration#npmrebuild

> Whether to [rebuild](https://docs.npmjs.com/cli/rebuild) native dependencies before starting to package the app.

WofWca (approved):
Looks pretty simple, although I don't understand the macOS build stuff fully,
The comment change seems OK, as well as the env var check.

PR #6031, opened 2026-02-09
refactor: move loadSettings into useEffect

nicodh:
Since https://github.com/deltachat/deltachat-desktop/commit/d32e85ec1353730a512cbdfff6c9c4dba5816dd4 we have a lot of react warnings.
Although we fixed a lot there are stil ~10 left since quite some time now.
So let's start to fix them now...


nicodh:
> I'm not sure why this fixes the warning though? We still do `setState` inside of `useEffect`. Maybe now React can see that it's being done asynchronously and is chill about it.

I think the linter is smart enough to detect that the async handling is poperly handled now (with the cancellation flag set to true on unmount)

PR #6034, opened 2026-02-09
fix: supress ignorable eslint warning

nicodh:


packages/frontend/src/components/Settings/Appearance.tsx
```diff
@@ -176,7 +176,11 @@ function BackgroundSelector({
         `position:absolute;top:${y}px;left:${eventPos.x}px;`
       )
       if (desktopSettings?.chatViewBgImg?.startsWith('color: ')) {
-        /* eslint react-hooks/immutability: "warn" */
+        // linter throws a a immutable error and a "modify in useEffect is
+        // not allowed" warning here
+        // both are false positives, the modifications happens in an event
+        // handler so the element can be modified and is not immutable
+        // eslint-disable-next-line
```
WofWca: I don't think this can be called a "false positive". The linter is talking about the access that is happening a few lines above in a `useEffect`

https://github.com/deltachat/deltachat-desktop/blob/e2ecf678341cc3d408cd6db4e469bc093c43e9f3/packages/frontend/src/components/Settings/Appearance.tsx#L157-L162

nicodh: You are right. I fixed it by introducing a getColorInput function and use that inside useEffect and the event handler so no DOM access happens during render.

nicodh:
> It seems to me that this merely obfuscates whatever the linter is trying to warn about. We don't expect `document.getElementById('color-input')` to ever change, so `getColorInput()` is always equivalent to `colorInput`. I suggest to simply add an eslint-ignore comment instead.

The difference is that the element is not defined during rendering anymore. But it's not worth to discuss this longer so let's go with an eslint-ignore

PR #6035, opened 2026-02-09
refactor: avoid ref access during render

nicodh:
The variable rightLimit is now calculated inside of showAfter to avoid the ref access during render


packages/frontend/src/components/ContextMenu.tsx
```diff
@@ -153,7 +158,9 @@ export function ContextMenuLayer({
     }
 
     // Displaying Menu
-    setPosition({ top, left })
+    const rightLimit =
+      layerRef.current.clientLeft + layerRef.current.clientWidth
```
WofWca, line 162: This should be above the "displaying menu" comment.

Issue #5957, opened 2026-01-21
play voice messages in background, "global player"


r10s:
instead of stopping voice messages when the corresponding chat is left, it should **continue playing**. even if profiles are switched.

to keep control over the playback in that case, we need to have a **global player**, integrate to the main window.

discussed with @nicodh a bit, we want to have the player at the bottom of the window - that way, there is less "bumping" of lots of content. to make visually clear the player is _beside_ the currently selected profile, and not necessarily playing content from there, we can re-use the color scheme from the profile switcher, which is similar from scope.

the player should have a **play/pause** button, **sendername** and **time** display and a **close** button.

when tapping sendername, **go to message** 

when another voice message is started, this should **replace** the existing playback, if any.

<details>
<summary>subsequent improvement for follow-up
issues/pr</summary>

**once these basics are done,** the issue can be closed and we can consider more improvements, maybe some are so easy to be part of this issue already 

- having  prev/next button
- having a seek slider 
- starting the next voice message automatically
- doing the same for [DC_MSG_AUDIO](https://c.delta.chat/group__DC__MSG.html#ga510c9ae279a2e5ee32ba49796a747e83),  this makes Delta Chat a simple audio player again
- having the option to select different speeds for voice messages by a button (peek what other messengers are offering here)
- last, maybe moste advanced: displaying a waveform below the seek slider in the chat

in the forum eg. https://support.delta.chat/t/voice-messages-should-be-paused-when-you-click-on-another-voice-message/1932
</details>

counterpart of https://github.com/deltachat/deltachat-android/issues/4183 and https://github.com/deltachat/deltachat-ios/issues/2987

Simon-Laux:
merging https://github.com/deltachat/deltachat-desktop/issues/5575 into here:
- [x] only one voice message should play at a time.

WofWca:
> to make visually clear the player is beside the currently selected profile

Where would that be?

nicodh:
> Where would that be?

So far we discussed:

either at the bottom:

<img width="1002" height="800" alt="Image" src="https://github.com/user-attachments/assets/ffaaeea8-a660-4c03-816b-bf06be29302c" />

or above the settings icon in the account list bar. 

The latter might be better visible but has 1. less horizontal space and 2. depends on the height and the number of accounts

WofWca:
> either at the bottom:

Why not make it span only the "accounts" and "chat list" sections? It doesn't need to be so wide to also take up the message list?
[...]

PR #6036, opened 2026-02-09
WIP: feat: add global voice message player

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/5957.

### DIFF
--- a/packages/frontend/src/ScreenController.tsx
+++ b/packages/frontend/src/ScreenController.tsx
@@ -23,6 +23,7 @@ import { ChatProvider, UnselectChat } from './contexts/ChatContext'
 import { ContextMenuProvider } from './contexts/ContextMenuContext'
 import { InstantOnboardingProvider } from './contexts/InstantOnboardingContext'
 import { SmallScreenModeMacOSTitleBar } from './components/SmallScreenModeMacOSTitleBar'
+import { GlobalVoiceMessagePlayerProvider } from './contexts/GlobalVoiceMessagePlayerContext'
 
 const log = getLogger('renderer/ScreenController')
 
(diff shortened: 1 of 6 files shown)


WofWca:
@nicodh let me know if you see anything icky so far.

nicodh:
Hmm, looks not as expected. Seems there is some css missing?

Or do you want me just to look at the code and the concept so far?

image

WofWca:
Yes, just the overall architecture so far. That we use a context, that we mute the original elements.

nicodh:
So yes - a context makes sense here. The name should not be GlobalVoiceMessagePlayer imo if we already plan to use it as a generic audio player later.

What do you mean with "mute the original elements"?
Is it already decided what to show in the actual voice message when it is played in the audio player? If it should show a synchronized audio player like the one we have already then the replacement with AudioAttachment doesn't make sense. Or what is the idea here?

WofWca:
> What do you mean with "mute the original elements"?

The original elements being the ones that we currently display in the message list and the gallery. The sound should be coming only from the global player instead, that's the idea.

> Is it already decided what to show in the actual voice message when it is played in the audio player? If it should show a synchronized audio player like the one we have already

Yes, I think so. Do you have better ideas?

> then the replacement with AudioAttachment doesn't make sense

Umm why?

nicodh:
Ah looking again I see that the AudioAttachment is just a wrapper for the AudioPlayer  👍

WofWca:
Ahh, so portal in 17c677576cfc48124fddab192f2e238b13dbc9a1 doesn't save us from `key={accountId}`: the `<audio>` element still gets re-created, losing playback position. I think we'll need to manually insert the element.
See also
- https://github.com/facebook/react/issues/13044
- https://github.com/facebook/react/issues/3965
- https://www.npmjs.com/package/react-reverse-portal
- https://github.com/Paol-imi/react-reparenting

WofWca:
Superseded by https://github.com/deltachat/deltachat-desktop/pull/6072 and https://github.com/deltachat/deltachat-desktop/pull/6062.

PR #6037, opened 2026-02-10
fix: supress warning for setState in effect

nicodh:


packages/frontend/src/components/dialogs/QrCode.tsx
```diff
@@ -142,6 +142,10 @@ export function QrCodeShowQrInner({
         const url = URL.createObjectURL(
           new Blob([qrCodeSVG], { type: 'image/svg+xml' })
         )
+        // setState inside an effect is needed here since
+        // Blob URL creation requires an effect for cleanup
+        // and another render is needed to display the QR code image
```
WofWca, line 147: The comment IMO is not very clear, the last line especially.

nicodh: setState triggers a new render which is needed since we need the URL

That's what I meant

PR #6038, opened 2026-02-10
refactor: avoid warnings in messageList

nicodh:
use layout effect and state instead ref & useHasChanged2

### DIFF
--- a/packages/frontend/src/components/message/MessageList.tsx
+++ b/packages/frontend/src/components/message/MessageList.tsx
@@ -20,6 +20,7 @@ import { throttledUpdateBadgeCounter } from '../../system-integration/badge-coun
 import { MessagesDisplayContext } from '../../contexts/MessagesDisplayContext'
 import useTranslationFunction from '../../hooks/useTranslationFunction'
 import useKeyBindingAction from '../../hooks/useKeyBindingAction'
+import { useHasChanged2 } from '../../hooks/useHasChanged'
 import { useReactionsBar } from '../ReactionsBar'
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/message/MessageList.tsx
```diff
@@ -850,11 +851,10 @@ export const MessageListInner = React.memo(
 
       onScroll(...args)
     }
-    const hasChatChanged = useHasChanged2(chat)
     const switchedChatAt = useRef(0)
-    if (hasChatChanged) {
+    useLayoutEffect(() => {
       switchedChatAt.current = Date.now()
-    }
+    }, [chat])
```
WofWca: I'm not sure how an effect is better?

> useEffect is a React Hook that lets you [synchronize a component with an external system.](https://react.dev/learn/synchronizing-with-effects)

Maybe it's more proper to utilize this pattern then https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes.

WofWca (commented):
The `unreadCount` change looks great, but IDK about the new `useLayoutEffect`.

packages/frontend/src/components/message/MessageList.tsx
nicodh: Setting a timestamp could also considered a side effect
"The Effect Hook lets you perform side effects in function components:"

But again, not worth to discuss further - I switched back to the useHasChanged2

WofWca (approved):

PR #6043, opened 2026-02-11
refactor: handle new Core `getNextEventBatch` API

WofWca:
- **refactor: extract into fn JSON-RPC event handling**
- **refactor: handle new Core `getNextEventBatch` API**

See https://github.com/chatmail/core/pull/7825.

This MR ensures that we properly handle events
even when they are retrieved with the new `getNextEventBatch()`
Core API, which is going to begin being the case
when that MR is merged.

I have tested this with old and new Core. Event logging and calls still work.

### DIFF
--- a/packages/target-electron/src/deltachat/controller.ts
+++ b/packages/target-electron/src/deltachat/controller.ts
@@ -1,5 +1,11 @@
 import { app as rawApp, ipcMain } from 'electron'
-import { yerpc, BaseDeltaChat, type T } from '@deltachat/jsonrpc-client'
+import {
+  yerpc,
+  BaseDeltaChat,
+  type T,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  type RawClient,
+} from '@deltachat/jsonrpc-client'
 import { getRPCServerPath } from '@deltachat/stdio-rpc-server'
 
 import { getLogger } from '../../../shared/logger.js'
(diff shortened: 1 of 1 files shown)


packages/target-electron/src/deltachat/controller.ts
```diff
@@ -221,3 +180,80 @@ export default class DeltaChatController {
...
+}
+
+function handleEventResponse(
+  result: T.Event,
+  jsonrpcRemote_: DeltaChatController['_jsonrpcRemote']
+): void {
+  // A workaround.
+  // Intercept the events that go to the renderer
+  // and manually fire them on this JSON-RPC client.
+  // See comments below about why we don't call `rpc.getNextEvent()`
```
link2xt: I don't see any comments below or any explanation why the hack is needed.

link2xt: There is a comment that says there can only be one consumer, but it is above now.

WofWca:
```suggestion
  // See comments above about why we don't call `rpc.getNextEvent()`
```

WofWca:
Hmm, IDK. Maybe https://delta.chat/en/2025-02-11-why-jsonrpc-bindings-exist is enough then?

WofWca:
Ah, this, unfortunately, doesn't actually work for Core <= 2.43.0, because the first `isEventResponse` throws then and we never get to `if else`. I suggest to just upgrade Core at this point. https://github.com/deltachat/deltachat-desktop/pull/6065.

Issue #6042, opened 2026-02-11
who_can_call_me option


link2xt:
The option is available in the core, but was not added to the UI:
https://github.com/chatmail/core/pull/7682

link2xt:
Corresponding Android issue: https://github.com/deltachat/deltachat-android/issues/4152

PR #6044, opened 2026-02-11
feat: add "Who can call me" setting

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/6042.

image

image

I have tested this. Setting to "nobody" makes the "incoming call window" not show up.
However, on Desktop it's then impossible to accept the call from the notification - we do not have any "click" handling for "call" messages.

TODO:
- [ ] Agree on the wordings, the position of the menu item, the condition when it's shown.

Related:
- https://github.com/chatmail/core/pull/7682.
- https://github.com/deltachat/deltachat-android/issues/4152.

### DIFF
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -100,5 +100,17 @@
   },
   "prevent_dialog_spam_checkbox": {
     "message": "Don't allow the app to open this dialog again"
+  },
+  "who_can_call_me": {
+    "message": "Who can call me"
+  },
+  "who_can_call_me_everybody": {
+    "message": "Everybody"
+  },
+  "who_can_call_me_contacts": {
+    "message": "Contacts"
+  },
+  "who_can_call_me_nobody": {
+    "message": "Nobody"
   }
 }
--- a/packages/frontend/src/components/Settings/Notifications.tsx
+++ b/packages/frontend/src/components/Settings/Notifications.tsx
@@ -12,8 +12,11 @@ import SettingsSwitch from './SettingsSwitch'
 import SettingsHeading from './SettingsHeading'
 import SettingsSelector from './SettingsSelector'
 import SmallSelectDialog from '../SmallSelectDialog'
-import SettingsStoreInstance from '../../stores/settings'
+import SettingsStoreInstance, { useSettingsStore } from '../../stores/settings'
 import useDialog from '../../hooks/dialog/useDialog'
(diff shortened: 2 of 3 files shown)


r10s:
ftr, i am not sure about the "Contacts" wording, for me, that is pretty vague, and is also not matching the use of contacts elsewhere - we eg. show gossiped group members on purpose in the contact list - but these are the ones one want to exclude iiuc

for "everyone" setting:  that does not exist even today, nor was there a request iirc. requests were only about disabling calls. did not check in details what comparable apps are offering here.

so, not sure if we want to introduce that state. one anyways always gets a (maybe muted) notification when a call is declined due to some settings or other reasons (there can be tons of reasons) and one can easily call or message back. maybe this is good enough for that usecase for now and not worth introducing a maybe theoretical usecase "everyone".

just some ideas, iirc, that option was not discussed much yet.

adbenitez:
I also feel this option is a bit ahead of its time, we don't even have the calls feature finished nor a proper concept of "my contacts" as mentioned by r10s, the contact list is a mess

nicodh:
So let's put that "on hold" for now until there is a clear "strategy" concerning the wording. But nevertheless it works already as expected if the word "Contacts" is read as "user who are in my contact list" and not as "users I see in any group members list"

r10s:
discussed with @adbenitez: for now, let's only introduce a single toggle, below "Mentions":

> **Calls**                                                   
> Show call screen on incoming calls 

if disabled, calls are still notified, if otherwise notifications arrive for that contact.

(in other words, if disabled, this is "nobody", if enabled, this is "accepted contacts")

this is what actually was requested, it does avoid "contact" wording and questions, and avoids adding features we need to maintain and to take care for, that may not be important in practise, see above

WofWca:
> Show call screen on incoming calls

Given the fact that it's currently impossible to accept a call by clicking the "call" message on Desktop, this becomes simply "disable incoming calls".

WofWca:
I implemented the suggestion in https://github.com/deltachat/deltachat-desktop/pull/6079.

r10s:
> Given the fact that it's currently impossible to accept a call by clicking the "call" message on Desktop

yip, but we should add that feature before calls go out of experimental, i created an issue for that https://github.com/deltachat/deltachat-desktop/issues/6081

r10s:
let's close this issue in favor of #6079 - might totally be that we want a tristate at some point, but that is then

PR #6054, opened 2026-02-17
chore: prepare release 2.43.0

nicodh:


packages/target-browser/package.json
```diff
@@ -2,7 +2,7 @@
   "private": true,
   "name": "@deltachat-desktop/target-browser",
   "type": "module",
-  "version": "2.35.0",
+  "version": "2.42.0",
```
WofWca: Should be 43

nicodh: The PR was incomplete. I will fix that

PR #6058, opened 2026-02-23
fix: extraneous padding on full screen image view

WofWca:
Introduced in
https://github.com/deltachat/deltachat-desktop/pull/5977.

FYI the `unstyled` prop is only `true` on `FullscreenAvatar`
and `FullscreenMedia`,
so this change is not epxected to have side effects.

I have checked that it works.


nicodh (approved):
Reviewed by reading & testing

PR #6059, opened 2026-02-23
fix: remove notification when its message deleted

WofWca:
- **refactor: bin notifications also by `messageId`**
- **fix: remove notification when its message deleted**

I have tested this. Also tested notifications otherwise.
Everything seems to be working.

### DIFF
--- a/packages/target-electron/src/notifications.ts
+++ b/packages/target-electron/src/notifications.ts
@@ -7,6 +7,7 @@ import type { DcNotification } from '../../shared/shared-types.js'
 import { getLogger } from '../../shared/logger.js'
 
 import type { NativeImage, IpcMainInvokeEvent } from 'electron'
+import { DCJsonrpcRemoteInitializedP } from './ipc.js'
 
 /**
  * Notification related functions to:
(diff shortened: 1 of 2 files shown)


WofWca:
Note that the coding approach sort of goes against the grain: I did not introduce a `runtime` method for this.
Because the code is quite trivial, and adding a runtime method would consist mostly of glue code. If another target (browser or Tauri) wishes to add the same code, it should be trivial for them, there is no need to have common logic separated on the front end.

packages/target-electron/src/notifications.ts
```diff
@@ -145,15 +158,28 @@ function showNotification(_event: IpcMainInvokeEvent, data: DcNotification) {
...
 }
 
+function clearNotificationsForMessage(
+  _: unknown,
+  accountId: number,
+  chatId: number,
+  messageId: number
+) {
+  notifications[accountId]?.[chatId]?.[messageId]?.forEach(notify => {
+    notify.close()
```
nicodh, line 172: Maybe also remove notifications[accountId][chatId][messageId] from the notifications here? Although it's probably not much harm to leave stale entries in the map so no blocker imo

WofWca:
> What is meant by
> "refactor: bin notifications also by messageId"?

I think it's this term https://en.wikipedia.org/wiki/Data_binning
Although I'm not 100% sure if it applies here.

PR #6062, opened 2026-02-25
feat: when playing voice message, pause other ones

WofWca:
- **refactor: use `src` attr instead of `<source>` el**
- **feat: when playing voice message, pause other ones**
- **refactor: use global `<audio>` for voice messages**

This is the first step towards https://github.com/deltachat/deltachat-desktop/issues/5957.

We introduce a hidden global `<audio>` element
and a `MediaPlayerMutexContext` along with it,
and secretly mute all the "real" audio elements,
those that are in the `AudioPlayer` component.

I think this is useful enough to be mergeable.
~~However, I am still not 100% sure about whether this is the right path towards implementing the "global player" in its completeness,~~
~~so if you are afraid to waste time reviewing this, then don't review yet.~~

### DIFF
--- a/packages/frontend/src/components/AudioPlayer/index.tsx
+++ b/packages/frontend/src/components/AudioPlayer/index.tsx
@@ -3,6 +3,7 @@ import React, { useContext, useEffect, useRef } from 'react'
 import styles from './styles.module.scss'
 import classNames from 'classnames'
 import { getLogger } from '@deltachat-desktop/shared/logger'
+import { ForceMutedAudioPlayer } from './ForceMutedAudioPlayer'
 import {
   MediaPlayerMutexContext,
   type MediaPlayerMutexContextValue,
@@ -46,7 +47,9 @@ export function AudioPlayer(
   }, [mediaPlayerMutexCtx.currentSrc, props.src])
 
   return (
-    <AudioPlayerNonMutex
+    // Muted because it's gonna be the <audio> element inside of
+    // `MediaPlayerMutexContext` that actually plays the media.
+    <ForceMutedAudioPlayer
       ref={ref}
       {...props}
       {...mediaPlayerMutexCtx.eventListeners}
--- a/packages/frontend/src/contexts/MediaPlayerMutexContext.tsx
+++ b/packages/frontend/src/contexts/MediaPlayerMutexContext.tsx
@@ -1,4 +1,4 @@
-import React, { createContext, useState } from 'react'
+import React, { createContext, useCallback, useState } from 'react'
 import { getLogger } from '@deltachat-desktop/shared/logger'
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 import { type runtime } from '@deltachat-desktop/runtime-interface'
@@ -13,8 +13,14 @@ export type MediaPlayerMutexContextValue = {
   currentSrc: string | null
   eventListeners: {
     onPlay: React.ReactEventHandler<HTMLMediaElement>
+    onPause: React.ReactEventHandler<HTMLMediaElement>
+    onSeeking: React.ReactEventHandler<HTMLMediaElement>
+    onRateChange: React.ReactEventHandler<HTMLMediaElement>
+    onVolumeChange: React.ReactEventHandler<HTMLMediaElement>
   }
 }
+
+const audioEl = document.createElement('audio')
 
 const noContextErrStr =
   'tried to use MediaPlayerMutexContextValue outside of context provider'
@@ -29,6 +35,18 @@ export const MediaPlayerMutexContext =
       onPlay: () => {
         log.warn(noContextErrStr)
       },
+      onPause: () => {
+        log.warn(noContextErrStr)
+      },
+      onSeeking: () => {
+        log.warn(noContextErrStr)
+      },
+      onRateChange: () => {
+        log.warn(noContextErrStr)
+      },
+      onVolumeChange: () => {
+        log.warn(noContextErrStr)
+      },
     },
   })
 
@@ -36,6 +54,16 @@ export function MediaPlayerMutexProvider({
   children,
 }: React.PropsWithChildren<{}>) {
   const [currentSrc, setCurrentSrc] = useState<string | null>(null)
+
+  const setSrcAndPlay = useCallback((newSrc: string | null) => {
+    setCurrentSrc(newSrc)
+
+    const newSrcStr = newSrc ?? ''
+    if (audioEl.getAttribute('src') !== newSrcStr) {
+      audioEl.src = newSrcStr
+    }
+    audioEl.play()
+  }, [])
 
   return (
     <MediaPlayerMutexContext.Provider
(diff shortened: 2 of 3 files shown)


packages/frontend/src/contexts/GlobalVoiceMessagePlayerContext.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+    const newSrcStr = newSrc ?? ''
+    if (audioEl.src !== newSrcStr) {
+      audioEl.src = newSrcStr
+    }
+    audioEl.play()
+  }, [])
+
+  return (
+    <GlobalVoiceMessagePlayerContext.Provider
```
nicodh: The player is also used to play attached audio files not onlöy voice messages so we should name it GlobalAudioMessagePlayerContext imo

packages/frontend/src/contexts/GlobalVoiceMessagePlayerContext.tsx
```diff
@@ -0,0 +1,110 @@
...
+          // it will be a `file:///...` URL.
+          // Note that `getAttribute()` can work around that,
+          // but we don't do it.
+          onPause: e => {
+            if (e.currentTarget.src !== audioEl.src) {
+              return
+            }
+            if (
+              e.currentTarget.ended
+              // && audioEl.duration - audioEl.currentTime < 3
```
nicodh: why not remove this comment?

packages/frontend/src/contexts/GlobalVoiceMessagePlayerContext.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+export function GlobalVoiceMessagePlayerProvider({
+  children,
+}: React.PropsWithChildren<{}>) {
+  const [currentSrc, setCurrentSrc] = useState<string | null>(null)
+
+  const setSrcAndPlay = useCallback((newSrc: string | null) => {
+    setCurrentSrc(newSrc)
+
+    if (audioEl == null) {
```
nicodh: Can this ever happen? createElement always returns an Element for known tags and audioEl is a constant

packages/frontend/src/contexts/MediaPlayerMutexContext.tsx
```diff
@@ -0,0 +1,110 @@
...
+  eventListeners: {
+    onPlay: React.ReactEventHandler<HTMLMediaElement>
+    onPause: React.ReactEventHandler<HTMLMediaElement>
+    onSeeking: React.ReactEventHandler<HTMLMediaElement>
+    onVolumeChange: React.ReactEventHandler<HTMLMediaElement>
+  }
+}
+
+const audioEl = document.createElement('audio')
+audioEl.controls = true
```
nicodh: Are the controls for the in future visible global player at the bottom bar? Since in this PR the controls will never be visible nor used if I get it right

packages/frontend/src/components/AudioPlayer/index.tsx
```diff
@@ -1,20 +1,66 @@
...
 
 import styles from './styles.module.scss'
 import classNames from 'classnames'
+import { getLogger } from '@deltachat-desktop/shared/logger'
+import { AudioPlayerNoGlobalMuteable } from './AudioPlayerMuteable'
+import { GlobalVoiceMessagePlayerContext } from '../../contexts/GlobalVoiceMessagePlayerContext'
+
+const log = getLogger('AudioPlayer')
 
 type Props = React.AudioHTMLAttributes<HTMLAudioElement> & {
```
nicodh, line 14: So according to the type the Props AudioPlayer would accept event listeners (onPlay, onPause) passed as props but those would just be overridden by {...globalPlayerCtx.eventListeners}

Can we express that in the Props type here?

packages/frontend/src/components/AudioPlayer/index.tsx
```diff
@@ -1,20 +1,66 @@
...
+    }
+
+    return () => {
+      el.pause()
+      el.currentTime = 0
+    }
+  }, [globalPlayerCtx.currentSrc])
+
+  return (
+    <AudioPlayerNoGlobalMuteable
```
nicodh: Again about naming: AudioPlayerNoGlobalMuteable is not intuitive to understand. What about AudioPlayerProxy ? That would make clear this element is just used as a "pass through" element to soemthing else

packages/frontend/src/contexts/GlobalVoiceMessagePlayerContext.tsx
WofWca: Renamed to `MediaPlayerMutexContext`.

packages/frontend/src/components/AudioPlayer/index.tsx
WofWca: Renamed it to `AudioPlayerNonMutexMuteable`.
The "proxy" nature of it emerges by us applying the `forceMute` attribute, so I don't think it's good to name it like that.

WofWca:
I split the last commit into two. I pushed a few small refactors, addressing some comments. Also added `ratechange` listener, and already implemented proper `getAttribute('src')`.

packages/frontend/src/contexts/MediaPlayerMutexContext.tsx
WofWca: Yes, I removed for now it, thanks

packages/frontend/src/components/AudioPlayer/index.tsx
nicodh: Hmm, that NonMutexMuteable seem even more confusing to - especially the fact the Mutex and Muteable sound somehow similar and not clear if the Non refers to Mutex or Muteable or MutexMuteable?

How about ForceMutedAudioPlayer ? That would make it obvious that it is force muted so not the one really playing. Tht might be sufficient to get an idea when reading the name.

Btw.: MediaPlayerMutexContext sounds fine for me!

WofWca: > ForceMutedAudioPlayer

That would be `ForceMuteableAudioPlayer` then, because it's only force-muted when `forceMute={true}`.

I can also suggest `AudioPlayerNonMutex2`. No need to specify `muteable` since it's only an extra prop that it supports.

nicodh: But will there be a case where the AudioPlayerNonMutexMuteable is used without forceMute?  Do we even need to expose that to the props?

WofWca: Good point.
Though I think it's a little more intuitive if there is a prop, because there is already a `muted` prop on `HTMLMediaElement`, they're very similar.

nicodh: But if the Player is named ForceMutedAudioPlayer and we skip the if(forceMute) condition inside and always set src.connect(zeroGain) then it should be obvious

For me it was more confusing that the prop could be set to true or false from outside

WofWca: Alright, let's go for that.
I force-pushed, implemented the suggestion.

nicodh (approved):

PR #6065, opened 2026-02-27
chore: upgrade core to 2.44.0

nicodh:


link2xt:
This will need #6043

nicodh:
> This will need #6043

Yes that's why it's a draft

PR #6067, opened 2026-02-27
chore: add FreeBSD to list of OSes for esbuild

feld:
This saves me from having to patch this in our build environment for each new release

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,8 @@
       "os": [
         "darwin",
         "win32",
-        "linux"
+        "linux",
+        "freebsd"
       ],
       "cpu": [
         "x64",


feld:
After more testing, this fixes that one problem but my efforts to make pnpm stop downloading the prebuilt rpc servers / cores for other OSes/arches has hit a wall.

If I just patch the package.json to only list "freebsd" in the os list, that fixes it... so I may have to keep patching this anyway

WofWca:
So do you think that it's semantically correct to add `freebsd` to `supportedArchitectures`? IMU you made it work on FreeBSD, but you have to do some patching, as you say? Or is this patching only about the offline build, which is only required by package repositories, and is not directly related to `supportedArchitectures`?

nicodh:
As WofWca mentioned it might be misleading if we list freebsd but in fact it needs patching to run it with freebsd.

And if it doesn't really fix your problem ".. so I may have to keep patching this anyway" we should not merge it imo

PR #6069, opened 2026-02-28
fix: do not crop stickers

### LINKED ISSUE #6053: stickers are not properly displayed
stickers should be displayed completely, they shouldn't be cropped, so if they are bigger than the expected sizes, their bigger side must be scaled down to fit the expected frame size, this is how it looks currently:

Image

this is how they are shown in the sticker frame/cell in android:

Image

nicodh:
resolves #6053 

Preview Desktop:

image

Same chat on Android:

image

Images used for testing:

petito-moreno

petito-moreno-landscape

### DIFF
--- a/packages/frontend/src/components/attachment/messageAttachment.tsx
+++ b/packages/frontend/src/components/attachment/messageAttachment.tsx
@@ -57,6 +57,8 @@ export default function Attachment({
     }
   }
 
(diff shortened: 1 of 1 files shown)


WofWca:
According to the Android screenshot, it seems that Android also utilizes a square frame with `object-fit: contain;`,

PR #6070, opened 2026-02-28
fix: overflow in delete account dialog

### LINKED ISSUE #5842: overflow in delete account dialog
Image

in general everywhere in the app where an email address is shown, it should be wrapped in a component that truncates the text with "..."

nicodh:
resolves #5842

Also shows private account tag or profile signature instead of email (which is only shown as fallback if neither exists.

In the confirmation dialog explanation the displayname is used - if that is ambiguous it's in the users responsibility imo

Here an example:

The fact that the explanation with the very long accountname  makes the dialog scrollable is acceptable I would say. Such a  long name without spaces is very unlikely

image

### DIFF
--- a/packages/frontend/src/components/screens/AccountDeletionScreen/styles.module.scss
+++ b/packages/frontend/src/components/screens/AccountDeletionScreen/styles.module.scss
@@ -50,7 +50,7 @@
   flex-direction: column;
   margin-inline-start: 8px;
   justify-content: center;
-  max-width: 240px;
+  min-width: 0;
   word-wrap: break-word;
 }
 


packages/frontend/src/components/screens/AccountDeletionScreen/styles.module.scss
```diff
@@ -50,6 +50,8 @@
   flex-direction: column;
   margin-inline-start: 8px;
   justify-content: center;
+  max-width: 240px;
```
WofWca:
```suggestion
  min-width: 0;
```

`max-width` doesn't work for narrower screens and is fragile overall.

I have tested this.

Image

PR #6072, opened 2026-02-28
feat: global voice message player

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/5957.

- Sequential playback
- UI
  image
- doesn't stop when you switch account
- pause others when new one starts

TODO:
- [x] Clean up, separate stuff into commits
- [x] Address TODOs
- [x] Test again, fix bugs.
- [ ] Add tests?
  Not sure if we can record voice messages in Playwright.

This is based on https://github.com/deltachat/deltachat-desktop/pull/6062.

The overall code structure and approach, and functionality, are ready for review.

### DIFF
--- a/packages/frontend/src/contexts/NextVoiceMessagePlayerContext.tsx
+++ b/packages/frontend/src/contexts/NextVoiceMessagePlayerContext.tsx
@@ -40,7 +40,13 @@ export function NextVoiceMessagePlayerProvider({
 }: React.PropsWithChildren<{}>) {
   const mediaPlayerMutexCtx = useContext(MediaPlayerMutexContext)
 
-  const [currMessage, setCurrMessage] = useState<null | AudioMessageInfo>(null)
+  const [currMessage_, setCurrMessage] = useState<null | AudioMessageInfo>(null)
+  const currMessage =
+    // Check in case the global player has started playing something else,
+    // i.e. not a voice message.
+    currMessage_ != null && mediaPlayerMutexCtx.currentSrc === currMessage_.src
+      ? currMessage_
+      : null
 
   const { nextMessage, nextMessageIsLoading } =
     useGetNextVoiceMessage(currMessage)


WofWca:
@nicodh please take a glance and let me know if you see a reason not to proceed with the current implementation.

nicodh:
> @nicodh please take a glance and let me know if you see a reason not to proceed with the current implementation.

I added some comments in https://github.com/deltachat/deltachat-desktop/tree/wofwca/global-voice-message-player-1

nicodh:
2 minor issues on a first test: 
- it seems jumping forward on the global player doesn't get synced with the MutedAudioPlayer
- the UI still needs some fine tuning - but that can be done later

image


Concerning the tests. I found this stackoverflow post https://stackoverflow.com/questions/76781679/how-can-i-send-audio-to-fake-mic-input-using-playwright-on-chromium - maybe that works?

Tbd: if I play an audio from the Apps & Media Gallery View - the global player is half visible below the dialog. What is expected if the overlay is closed: should the player stop or continue?

packages/frontend/src/contexts/NextVoiceMesagePlayerContext.tsx
```diff
@@ -0,0 +1,219 @@
...
+  accountId: number
+  chatId: number
+  messageId: number
+  src: string
+}
+
+type ContextValue = {
+  setCurrMessage: (newMsg: AudioMessageInfo | null) => void
+}
+export const NextVoiceMesagePlayerContext = createContext<ContextValue>({
```
nicodh: There are some names (including the filename) with a typo: mesage => message

nicodh:
A comment in the NextVoiceMessagePlayerContext would be nice. What does it do? What logic is behind the message  queue etc.

The feature to continuously play all voice/audio messages as long as the player is not stopped needs some considerations/discussions imo

At least for me it was kind of unexpected

WofWca:
> it seems jumping forward on the global player doesn't get synced with the MutedAudioPlayer

Yes, there is still a lot that can still be done. This also goes for `volume`, `muted`, `playbackRate`. I think the way it is now is fine for a start.

> What is expected if the overlay is closed: should the player stop or continue?

I don't know. I'd say that it should continue. In Telegram it does continue.

> A comment in the NextVoiceMessagePlayerContext would be nice. What does it do?

Right. I sometimes forget to document things.

> The feature to continuously play all voice/audio messages as long as the player is not stopped needs some considerations/discussions imo

What is there to discuss? Whether it should exist? It was already requested in https://github.com/deltachat/deltachat-desktop/issues/5957, and other messengers do this already. WhatsApp, Telegram, Signal. Both mobile and desktop.

WofWca:
I added some comments, and factored out some logic into `useGetNextVoiceMessage`, to separate "pure" code from effectful code.

nicodh:
> What is there to discuss? Whether it should exist? It was already requested in #5957,

Ah ok. I oversaw "starting the next voice message automatically" in the collapsed item in the description

packages/frontend/src/components/AudioPlayer/index.tsx
```diff
@@ -18,20 +18,35 @@ type Props = React.AudioHTMLAttributes<HTMLAudioElement> & {
...
+     * of us calling `play()`.
+     */
+    onPlayNonProgrammatic?: React.ReactEventHandler<HTMLMediaElement>
+  }) {
   const ref = useRef<HTMLAudioElement>(null)
 
   const mediaPlayerMutexCtx = useContext(MediaPlayerMutexContext)
 
+  const ignoreOnePlayEvent = useRef(false)
+  const ignoreOnePauseEvent = useRef(false)
```
nicodh, line 38: Would be good to add a comment here to make obvious it's not just a typo (OnePlay instead OnPlay) but should avoid reacting on an event that is triggered by another event handler - at least that is how I understand it.

WofWca: Well I'd say both of the names work.

nicodh: Yes the name is fine I just say a comment would help here to understand what is the reason to have them and/or when this is false or true

WofWca: The prop description is enough is it not?

https://github.com/deltachat/deltachat-desktop/blob/995d0a3a8695ff045131471e4da5854c13770e32/packages/frontend/src/components/AudioPlayer/index.tsx#L26-L31

WofWca:
Rebased.

WofWca:
> not sure what you mean with "overlay is closed" here.

I think what's meant is "the Gallery dialog is closed".

WofWca:
> in-chat-player and global-player seem to share UI, which is great. but i would also use the same background tone etc. eg. play/pause hover is visible only in in-chat-player, so i would copy from there for now.

The background color is simply partly transparent, that's why it's happening. It's better to pick a concrete color then.

Note that we're using the native Chromium audio player, so we're limited in what we can do. I think it's possible to manipulate its Shadow DOM, but we have to re-check if things work on Chromium upgrades.

r10s:
> Note that we're using the native Chromium audio player, so we're limited in what we can do. I think it's possible to manipulate its Shadow DOM, but we have to re-check if things work on Chromium upgrades.

No, i would not insist on some colors when using other colors makes maintenance easier.

so, it makes totally sense to stay with the defaults there, at least for now.

WofWca:
Also regarding the "too cluttering" issue: I think most of that is better to solve by simply expanding the player to also below the account list sidebar.

Edit: I will make an MR that improves styles a little.

PR #6076, opened 2026-03-02
feat: add autostart functionality

### LINKED ISSUE #2518: Option to setup autostart for Deltachat

nicodh:
resolves #2518

It seems this does not work for Windows appx packages. We could consider to use https://github.com/felixrieseberg/electron-winstore-auto-launch which is quite old but in the Issues it is reported that it works at least with Windows 11

I'm not sure how this works in SNAP, maybe I will open an issue there.

### DIFF
--- a/packages/target-tauri/runtime-tauri/runtime.ts
+++ b/packages/target-tauri/runtime-tauri/runtime.ts
@@ -166,6 +166,7 @@ class TauriRuntime implements Runtime {
       ...deprecated,
       bounds: {}, // managed by tauri_plugin_window_state plugin
       HTMLEmailWindowBounds: undefined, // managed by tauri_plugin_window_state plugin
+      autostartElectron: false, // not needed in tauri version
     } satisfies Partial<DesktopSettingsType>
 
     const frontendAndTauri = {


packages/target-electron/src/autostart.ts
```diff
@@ -0,0 +1,171 @@
...
+    if (loginItemSettings.openAtLogin !== enable) {
+      app.setLoginItemSettings({ openAtLogin: enable })
+      log.info(`Autostart ${enable ? 'enabled' : 'disabled'}`)
+    }
+  } else if (currentPlatform === 'linux') {
+    if (process.env.SNAP) {
+      log.warn('Autostart not supported inside Snap')
+      return
+    }
+    const autostartFile = getLinuxAutostartFilePath()
```
nicodh, line 128: tbd: what would be expected here: if the settings have autostart enabled but I start a different app than the one that created an autostart entry previously (AppImage instead of deb for example) - should we A) just replace the file autostart/deltachat-desktop.desktop with the one needed for AppImage? Or should we B) show "autostart not enabled" in the Settings when we find the exec line to be a different from getLinuxExecPath for the current app?

A) makes sense if it is unexpected that someone installs a package just as an additional package (only testers would do that imo) 
B) makes sense if you don't want to override any settings if you just start another package of an existing app.

Note that while the accounts & settings of AppImage, Flatpak & deb would reside in different places the generated autostart file would be in the same place.

So to simplify things we could just check if autostart is enabled and if so check if an autostart file exists and if the content of the exec string is equal to  getLinuxExecPath. Then we would get rid of all the parsing stuff.

If someone wants to have a custom autostart file (like without the --minimized parameter) the settings should be set to disable autostart and create a custom autostart file for that

nicodh:
Just tested the setup.exe from preview artifacts on Windows:

In Apps->Autostart I see: (the german terms say "Registrypath" and "No effect"
grafik

in
Computer\HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Run\
I see
electron.app.DeltaChat-DevBuild: C:\Users\user2\AppData\Local\Temp\3AOfYPqzaIh6AiVIOA6oQrpVbQJ\DeltaChat-DevBuild.exe

edit:
But that might be an effect of the fact that the app needs 2 steps to go through the Windows Defender control so it doesn't say much about the real behaviour.

But it turned out that there is no "easy" way to add the "official" Deltachat App to autostart in Windows, since it is not listed when opening the "Add" sign in Apps->Start apps which is really bad.

I added addautolaunchextension:true to the appx build configuration. https://www.electron.build/appx.html#addautolaunchextension Let's see if that helps.

packages/target-electron/src/autostart.ts
```diff
@@ -0,0 +1,135 @@
...
+    const legacyFile = join(getLinuxAutostartDir(), 'deltachat-desktop.desktop')
+    try {
+      if (enable) {
+        await mkdir(getLinuxAutostartDir(), { recursive: true })
+        await writeFile(autostartFile, getLinuxDesktopFileContent(), 'utf-8')
```
WofWca: Shouldn't we delete the files on uninstall?

packages/target-electron/src/autostart.ts
```diff
@@ -0,0 +1,135 @@
...
+import { existsSync } from 'fs'
+
+import { getLogger } from '../../shared/logger.js'
+import { appx } from './isAppx.js'
+import { AutostartState } from '../../shared/shared-types.js'
+
+const log = getLogger('main/autostart')
+
+function getLinuxPackageType(): 'other' | 'appimage' | 'flatpak' {
+  if (process.env.APPIMAGE) return 'appimage'
```
WofWca, line 14:
```suggestion
  // https://docs.appimage.org/packaging-guide/environment-variables.html#environment-variables
  if (process.env.APPIMAGE) return 'appimage'
```

packages/target-electron/src/autostart.ts
```diff
@@ -0,0 +1,135 @@
...
+}
+
+function getLinuxAutostartFilePath(): string {
+  const packageType = getLinuxPackageType()
+  const suffix = packageType !== 'other' ? `-${packageType}` : ''
+  return join(getLinuxAutostartDir(), `deltachat-desktop${suffix}.desktop`)
+}
+
+function getLinuxExecPath(): string {
+  // When running as an AppImage, APPIMAGE env var is set to the .AppImage file path
```
WofWca, line 38:
```suggestion
  // When running as an AppImage, APPIMAGE env var is set to the .AppImage file path
  // https://docs.appimage.org/packaging-guide/environment-variables.html#environment-variables
```

packages/target-electron/src/autostart.ts
```diff
@@ -0,0 +1,135 @@
...
+  // In Flatpak, $XDG_CONFIG_HOME is redirected to the app sandbox even with
+  // --filesystem=host. Detect Flatpak and use $HOME directly so the autostart
+  // file ends up where the desktop environment can find it.
+  // see https://docs.flatpak.org/en/latest/conventions.html#xdg-base-directories
+  const isInsideFlatpak = Boolean(process.env.FLATPAK_ID)
+  const configHome = isInsideFlatpak
+    ? join(app.getPath('home'), '.config')
+    : process.env.XDG_CONFIG_HOME || join(app.getPath('home'), '.config')
+  return join(configHome, 'autostart')
+}
```
WofWca, line 29: Have you looked at https://github.com/flatpak/xdg-desktop-portal/blob/main/data/org.freedesktop.impl.portal.Background.xml, `EnableAutostart`.
Sounds like a more proper way.
Issue: https://github.com/flatpak/flatpak/issues/118.

Is it not suitable for us? Why?

In Tauri we also do it with "portals": https://github.com/deltachat/deltachat-desktop/pull/5094

packages/target-electron/src/autostart.ts
```diff
@@ -0,0 +1,135 @@
...
+  const autostartFile = getLinuxAutostartFilePath()
+  return existsSync(autostartFile)
+}
+
+function getLinuxDesktopFileContent(): string {
+  return `[Desktop Entry]
+Type=Application
+Name=DeltaChat
+Comment=Delta Chat email-based messenger
+Exec=${escapeDesktopExecArg(getLinuxExecPath())} --minimized
```
WofWca: As per https://specifications.freedesktop.org/desktop-entry/latest/exec-variables.html

> If no full path is provided the executable is looked up in the $PATH environment variable used by the desktop environment.

Don't we expect Delta Chat executable to be in PATH? This would allow us to simplify or get rid of `getLinuxExecPath`?

packages/target-electron/src/autostart.ts
```diff
@@ -0,0 +1,135 @@
...
+function getLinuxAutostartRegisteredState(): boolean {
+  const autostartFile = getLinuxAutostartFilePath()
+  return existsSync(autostartFile)
+}
+
+function getLinuxDesktopFileContent(): string {
+  return `[Desktop Entry]
+Type=Application
+Name=DeltaChat
+Comment=Delta Chat email-based messenger
```
WofWca:
```suggestion
Name=Delta Chat
Comment=Delta Chat decentralized private messenger
```

packages/target-electron/src/autostart.ts
```diff
@@ -0,0 +1,135 @@
...
+      } else {
+        if (existsSync(autostartFile)) {
+          await rm(autostartFile)
+          log.info(`Autostart disabled: removed ${autostartFile}`)
+        }
+      }
+      // Clean up legacy autostart file from before per-package-type naming
+      if (existsSync(legacyFile) && legacyFile !== autostartFile) {
+        await rm(legacyFile)
+      }
```
WofWca: Umm is this code written only for those who have tested this MR?

packages/target-electron/src/autostart.ts
```diff
@@ -0,0 +1,135 @@
...
+  const currentPlatform = platform()
+
+  if (currentPlatform === 'darwin' || currentPlatform === 'win32') {
+    if (appx) {
+      log.warn('Autostart not supported for Windows Store (APPX) builds')
+      return
+    }
+    const loginItemSettings = app.getLoginItemSettings()
+    if (loginItemSettings.openAtLogin !== enable) {
+      app.setLoginItemSettings({ openAtLogin: enable })
```
WofWca:
```suggestion
      app.setLoginItemSettings({ openAtLogin: enable, args: ['--minimized'] })
```

(I haven't tested if this works, but wouldn't hurt to add?)

packages/target-electron/src/autostart.ts
```diff
@@ -0,0 +1,135 @@
...
+          log.info(`Autostart disabled: removed ${autostartFile}`)
+        }
+      }
+      // Clean up legacy autostart file from before per-package-type naming
+      if (existsSync(legacyFile) && legacyFile !== autostartFile) {
+        await rm(legacyFile)
+      }
+    } catch (error) {
+      log.error(`Failed to ${enable ? 'enable' : 'disable'} autostart`, error)
+    }
```
WofWca: IDK if this try-catch is much helpful? Maybe errors should be handled by the caller of `applyAutostart`.

packages/target-electron/src/autostart.ts
```diff
@@ -0,0 +1,135 @@
+import { app } from 'electron'
+import { platform } from 'os'
+import { writeFile, rm, mkdir } from 'fs/promises'
+import { join } from 'path'
+import { existsSync } from 'fs'
```
WofWca: We can use `async` version.

packages/target-electron/src/autostart.ts
nicodh: If it worked for flatpak with Tauri that is definitely the way to go! You are welcome to apply that in this PR the same way you did it in Tauri.

WofWca: Well in Tauri we do that with a Rust lib, i.e. an abstraction, as far as I understand. This is the code: https://github.com/deltachat/deltachat-desktop/commit/36417743f4705c530ca5d268dc9c00c472303617#diff-98f72aea8508e418957d62ed0db9f279a7908f0de1cf0162c680a258eb952c39R48-R50

nicodh: In the comment above the autostart section it says:
 This is deprecated! New versions of xdg-desktop-portal will not call this method.

And I didn't find any tauri flatpak at all. So what exactly is your suggestion here?

WofWca: Well it's not that I know much about Flatpak. Just pointing out that they seem to be supporting autostart stuff so we don't have to hack around the sandbox and stuff. I don't know how exactly to use their API.

Apparently this is the new documentation:
https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Background.html

Related issues:
- https://github.com/ente-io/ente/issues/9008
- https://github.com/hkdb/aerion/issues/33#issuecomment-4045033121

> And I didn't find any tauri flatpak at all.

I believe it worked, but it was rejected by Flathub as a "duplicate app".

packages/target-electron/src/autostart.ts
nicodh: Not so easy, documentation is not very helpful. I found this: https://www.electron.build/app-builder-lib.interface.linuxtargetspecificoptions#afterRemove so maybe we can add an afterRemove script for debian but let's put that in a followup. There are already enough "unpredictable" changes in this PR

packages/target-electron/src/autostart.ts
nicodh: So you want to add that doc link twice?

packages/target-electron/src/autostart.ts
nicodh: So I did some reserach on this. It seems we would have to talk to gdbus or install a npm dependency dbus-next. But even then there are limitations since teh Portal can not be queried to get the current state for our UI.

And since we set the filesystem to Host anyway https://github.com/flathub/chat.delta.desktop/blob/b8ec1f6fc20d4854452a30be3d80326d4768c948/chat.delta.desktop.yml#L22 we can write to ~/.config/autostart/

I also had a look at https://github.com/flatpak/xdg-desktop-portal/issues/1445 . So xdg-desktop-portal version 4 is rolling out https://repology.org/project/xdg-desktop-portal/versions shows

So somewhen in the next month we should maybe start to find a Portal solution for the autostart but for now I would also like to do that in a followup

WofWca: > So somewhen in the next month we should maybe start to find a Portal solution for the autostart but for now I would also like to do that in a followup

Alright, but maybe we should then remove autostart support for Flatpak in this MR, to avoid future migration headaches? Finding and deleting old files and stuff, detecting whether autostart was enabled.

packages/target-electron/src/autostart.ts
WofWca: Yes. It's worth adding in both of the places. I don't have an idea on how to DRY this.

packages/target-electron/src/autostart.ts
nicodh: Well if we have the choice between: no autostart feature for flatpak at all or maybe some orphaned autostart entries if we sometime change the flatpak autostart behaviour somewhen in the future I would go for the first option

packages/target-electron/src/autostart.ts
nicodh: I'm not sure if that is the better approach. It might happen for someone who let's say has an ApplicationImage and  a debian package install - it is not clear to which the Deltachat executable in $PATH would belong.

Here we "know" the current process is the one running the current instance

packages/target-electron/src/autostart.ts
nicodh: I don't agree. The other comment is just 15 lines above and the doc is just about why it is expected that env.APPIMAGE exists and holds the path

packages/target-electron/src/autostart.ts
WofWca: > if we sometime change

But you said "somewhen in the next month". Is it worth it to introduce this implementation only for a month?

WofWca: BTW the dbus setup has been discussed in https://github.com/deltachat/deltachat-desktop/issues/2518 a little.

nicodh: It should say the next month**s** - I have no concrete plan to address that issue yet. Do you?

packages/target-electron/src/autostart.ts
```diff
@@ -0,0 +1,127 @@
...
+
+function getLinuxAutostartDir(): string {
+  // In Flatpak, $XDG_CONFIG_HOME is redirected to the app sandbox even with
+  // --filesystem=host. Detect Flatpak and use $HOME directly so the autostart
+  // file ends up where the desktop environment can find it.
+  // see https://docs.flatpak.org/en/latest/conventions.html#xdg-base-directories
+  const isInsideFlatpak = Boolean(process.env.FLATPAK_ID)
+  const configHome = isInsideFlatpak
+    ? join(app.getPath('home'), '.config')
+    : process.env.XDG_CONFIG_HOME || join(app.getPath('home'), '.config')
```
WofWca, line 27:
```suggestion
    ? app.getPath('appData')
    : process.env.XDG_CONFIG_HOME || app.getPath('appData')
```

https://www.electronjs.org/docs/latest/api/app

    appData Per-user application data directory, which by default points to:

    %APPDATA% on Windows
    $XDG_CONFIG_HOME or ~/.config on Linux
    ~/Library/Application Support on macOS

packages/target-electron/src/autostart.ts
```diff
@@ -0,0 +1,134 @@
...
+      return
+    }
+    const autostartFile = getLinuxAutostartFilePath()
+    if (enable) {
+      await mkdir(getLinuxAutostartDir(), { recursive: true })
+      await writeFile(autostartFile, getLinuxDesktopFileContent(), 'utf-8')
+      log.info(`Autostart enabled: created ${autostartFile}`)
+    } else {
+      try {
+        await access(autostartFile)
```
WofWca:
```suggestion
```

Since this is wrapped in try-catch, `access` seems unnecessary.

packages/target-electron/src/autostart.ts
WofWca: Alright, if that's the case then let's try this approach.

packages/target-electron/src/autostart.ts
```diff
@@ -0,0 +1,134 @@
...
+    return false
+  }
+}
+
+function getLinuxDesktopFileContent(): string {
+  return `[Desktop Entry]
+Type=Application
+Name=Delta Chat
+Comment=Delta Chat decentralized private messenger
+Exec=${escapeDesktopExecArg(getLinuxExecPath())} --minimized
```
WofWca: FTR `KeePassXC` uses `flatpak run org.keepassxc.KeePassXC`. I don't know if that's done over the portal thing that we discussed.

packages/target-electron/src/autostart.ts
```diff
@@ -0,0 +1,134 @@
...
+
+// see https://specifications.freedesktop.org/desktop-entry/latest/exec-variables.html
+function escapeDesktopExecArg(arg: string): string {
+  const escaped = arg
+    .replace(/\\/g, '\\\\')
+    .replace(/"/g, '\\"')
+    .replace(/\$/g, '\\$')
+    .replace(/`/g, '\\`')
+  return `"${escaped}"`
+}
```
WofWca, line 58: The linked documentation seems to talk more about arguments and not the executable program, i.e. file path. I'm not sure if this escape function is supposed to work for paths?
If the point of this function is to support file paths with these characters in them then it doesn't work, based on my tests. I see this error:

```log
systemd-xdg-autostart-generator: /home/wofwca/.config/autostart/deltachat-desktop-appimage.desktop:5: Undefined escape sequence \$.
systemd-xdg-autostart-generator: /home/wofwca/.config/autostart/deltachat-desktop-appimage.desktop:5: Failed to parse file: Invalid argument
systemd-xdg-autostart-generator: Failed to parse /home/wofwca/.config/autostart/deltachat-desktop-appimage.desktop, ignoring it
```

For a file named

```
deltachat-desktop.add-autostart-feature.AppImage test$\`
```

WofWca (approved):
Nice overall!

I tested this with AppImage. Turning it on and off works, and does have effect, i.e. it auto-starts.
On Windows I tested toggling it on and off but not restarting. It works (with https://github.com/deltachat/deltachat-desktop/pull/6127).
I did not test macOS, Flatpak and Snap.

`--minimized` doesn't work, but that's a separate issue, I will take a look.

packages/target-electron/src/autostart.ts
nicodh: I think this is more about inconsistency in electron: appData should not go to config home but does in electron.

According to https://specifications.freedesktop.org/basedir/latest/ appData should go to $XDG_STATE_HOME
not to $XDG_CONFIG_HOME
"There is a single base directory relative to which user-specific state data should be written. This directory is defined by the environment variable $XDG_STATE_HOME."

So I would not rely on the appData path when we want to get the path to the autostart config

packages/target-electron/src/autostart.ts
```diff
@@ -0,0 +1,132 @@
...
+    }
+    const autostartFile = getLinuxAutostartFilePath()
+    if (enable) {
+      await mkdir(getLinuxAutostartDir(), { recursive: true })
+      await writeFile(autostartFile, getLinuxDesktopFileContent(), 'utf-8')
+      log.info(`Autostart enabled: created ${autostartFile}`)
+    } else {
+      await access(autostartFile)
+      await rm(autostartFile)
+      log.info(`Autostart disabled: removed ${autostartFile}`)
```
WofWca: When autostart is disabled on Linux, this code runs unconditionally on app startup, and this will throw every time because there is no such file, resulting in a "Failed to disable autostart" log entry every time.

nicodh: That's why it was in a try catch with a silent catch - what you complained about above. I just wasn't aware anymore - after moving the overall try catch one level up

WofWca: I complained about `access` not being useful and not about try-catch. https://github.com/deltachat/deltachat-desktop/pull/6076#discussion_r2951884738

nicodh: But maybe we could just skip applyAutostart on startup completely, so per default no autostart is configured and only if a user turns it on we set a file?

WofWca: That would be a bigger change because the stored `autostart` boolean would be still true by default, out of sync with reality.

WofWca: But indeed it would be nice not to run this unnecessarily.

Regarding whether it should be enabled by default: I think it should. But maybe it's better to check what the common practice is.

Or actually, given that this is the first version of this feature, maybe it ~~doesn't~~ does make sense to disable it by default.

WofWca: So yeah, I am for not doing this on startup, together with changing the default to `false`. But we would then also need a migration to set it to `false`, because it's already present as `true` in people's configs.

packages/target-electron/src/autostart.ts
nicodh: But there is also ~/.local/share/flatpak/exports/bin/cht.delta.desktop as direct path to the binary, so I would expect that's what process.execPath would report (but not sure about that)

WofWca: Yes, probably. What I'm saying is that `flatpak run` might be a better way to launch Flatpak apps.

nicodh:
I successfully tested that our build process is still working and that the Autostart for the created dmg file works on MacOS Tahoe 26.3

Issue #5525, opened 2025-10-07
chat does not always scroll down on new call messages


r10s:
when the chat is already scrolled down, and a call is placed, the chat should scroll down, so that the bubble representing the call becomes visible 

currently, the chat is not scrolled down and only the "scroll down" button is shown

nicodh:
Maybe related to https://github.com/deltachat/deltachat-desktop/issues/4329 ?

nicodh:
SOlved with https://github.com/deltachat/deltachat-desktop/pull/6077

PR #6077, opened 2026-03-03
Scroll to new message

### LINKED ISSUE #4329: We don't scroll to some newly arrived messages
- Delta Chat Version: 1.48.0
- Expected behavior: No matter what message shows up in a chat, we should scroll to it (as long as we're scrolled to the bottom, of course)
- Actual behavior: The scroll position remains where it was and you have to scroll to the new message
    1. Open a chat with someone
    2. Send a `.xdc` (webxdc) app.
    3. Open that app and open the dev tools for the app.
    4. Type
        ```javascript
        webxdc.sendUpdate({
            payload: null,
            info: 'test',
        }, '')
        ```
- Screenshots:

    https://github.com/user-attachments/assets/ae4c30f9-813f-4817-9206-28c9f08dfe89

- Logs:

The above reproduction steps also work for when the info message is triggered by other chat members.

As was discussed in a chat today, some messages do not trigger the [`IncomingMsg`](https://github.com/deltachat/deltachat-core-rust/blob/61fd0d400fd2dbd40ba9494d8d8a65a0c5749d9f/src/events/payload.rs#L110-L120) event. Then only trigger [`MsgsChanged`](https://github.com/deltachat/deltachat-core-rust/blob/61fd0d400fd2dbd40ba9494d8d8a65a0c5749d9f/src/events/payload.rs#L72-L84) But currently the `IncomingMsg` event is what we're listening on to scroll to new messages:

https://github.com/deltachat/deltachat-desktop/blob/c28d796ce5ba98d84905537c3b8514635f3680a2/packages/frontend/src/stores/messagelist.ts#L65-L73

https://github.com/deltachat/deltachat-desktop/blob/c28d796ce5ba98d84905537c3b8514635f3680a2/packages/frontend/src/stores/chat/chat_view_reducer.ts#L92-L106

So, as was suggested by @link2xt and @r10s, we probably need to rewrite the code that updates `viewState`, to update `viewState` based on actual changes to the array of messages, and not just events, to catch all such cases.

Related older issues: https://github.com/deltachat/deltachat-desktop/issues/1783, https://github.com/deltachat/deltachat-desktop/pull/1793,

nicodh:
Fixes #4329 and #5525

Some new messages just trigger an MsgChanged event and not IncomingMessage. Info messages triggered by current user actions like (renaming group, changing disappearing messages settings etc.) and also the IncomingCall message only triggers MsgChanged.

This PR applies the same behaviour as if that were incoming messages i.e. scroll to them and mark them as read if the related chat is opened.

Some weird behaviour: testing in various groups on main it turned out that the wrong behaviour (not scrolling to these messages but only show the jumpDown button) is not always reproducable. Some groups show that behaviour some not!

### DIFF
--- a/packages/frontend/src/components/message/MessageList.tsx
+++ b/packages/frontend/src/components/message/MessageList.tsx
@@ -524,11 +524,6 @@ export default function MessageList({
             behavior: 'smooth',
           })
           pendingProgrammaticSmoothScrollTo.current = scrollTo
-          // Mark all messages as noticed. This is needed for messages that are
-          // not picked up by the IntersectionObserver (e.g. info messages
-          // from the current user's own actions such as changing the group
-          // name, whose state is not DC_STATE_IN_FRESH).
-          markChatAsSeen(accountId, chat.id)
 
           // Smooth scroll duration is not defined by the spec:
           // https://drafts.csswg.org/cssom-view/#scrolling:
@@ -583,7 +578,6 @@ export default function MessageList({
     }, 0)
   }, [
     chat,
-    accountId,
     onScroll,
     unlockScroll,
     viewState,


packages/frontend/src/components/message/MessageList.tsx
```diff
@@ -529,6 +533,11 @@ export default function MessageList({
             behavior: 'smooth',
           })
           pendingProgrammaticSmoothScrollTo.current = scrollTo
+          // Mark all messages as noticed. This is needed for messages that are
+          // not picked up by the IntersectionObserver (e.g. info messages
+          // from the current user's own actions such as changing the group
+          // name, whose state is not DC_STATE_IN_FRESH).
+          markChatAsSeen(accountId, chat.id)
```
WofWca: It doesn't make sense to have this only in the `scrollTo.ifClose === true` branch. If you scroll manually to such a message then it will also not get marked as read.

Also I'm not understanding why we need to `markChatAsSeen` for such messages. They're not shown as "unread", so why do we have to do this?

Also see this related discussion https://github.com/chatmail/core/issues/7884. It seems that what we want is to just mark individual messages as read even when they're our own.

packages/frontend/src/stores/messagelist.ts
```diff
@@ -817,13 +762,83 @@ export class MessageListStore extends Store<MessageListState> {
...
+    // check if there is an intersection
+    if (indexStart !== -1 && messageListItems[indexStart + 1]) {
+      indexStart = indexStart + 1
+    }
+
+    // if index start is not the end, set the end to the last item
+    const indexEnd =
+      indexStart !== messageListItems.length - 1
+        ? messageListItems.length - 1
+        : indexStart
```
WofWca, line 813: Moving code while changing it at the same time makes it harder to review.

packages/frontend/src/components/message/MessageList.tsx
nicodh: I tested again. This is/was related to IncomingCall messages. But it seems the badge stays only unread when step debugging (probably a timing issue)

So I reverted the commit completely

WofWca (approved):
Oh, and I forgot to review the third change, which is the fix for the "jump down" button flashing. Yes, this also looks good, and works, based on testing. It was a really annoying bug.

As to the "main" change of this MR: as already said, it looks good. I also re-tested it again. Works for "disappearing messages", calls. Both self-sent and peer-sent.

PR #6083, opened 2026-03-04
fix: UI state of global player resetting

WofWca:


WofWca:
Yes, just the last commit. GitHub sucks at rebasing.

PR #6085, opened 2026-03-04
feat: use new api func createQrSvg

### LINKED ISSUE #5969: better scannable "Add Second Device" QR code

nicodh:
Resolves #5969

Needs core with https://github.com/chatmail/core/pull/7949 

Tested with local core and above mentioned PR

This is how it looks like now:

image

edit: removed invite qr code screenshot as it it not modified anymore in the pr

### DIFF
--- a/packages/frontend/src/components/dialogs/SetupMultiDevice/SendBackupDialog.tsx
+++ b/packages/frontend/src/components/dialogs/SetupMultiDevice/SendBackupDialog.tsx
@@ -135,7 +135,7 @@ export function SendBackupDialog({ onClose }: DialogProps) {
       onClose={cancel}
       width={500}
     >
-      <DialogHeader title={tx('multidevice_title')} />
+      <DialogHeader title={tx('multidevice_title')} onClose={cancel} />
       {!inProgress && (
         <>
           <DialogBody>
@@ -144,10 +144,7 @@ export function SendBackupDialog({ onClose }: DialogProps) {
             </DialogContent>
           </DialogBody>
           <DialogFooter>
-            <FooterActions align='spaceBetween'>
-              <FooterActionButton onClick={cancel}>
-                {tx('cancel')}
-              </FooterActionButton>
+            <FooterActions>
               <FooterActionButton
                 styling='primary'
                 onClick={startNetworkedTransfer}
@@ -187,21 +184,16 @@ export function SendBackupDialog({ onClose }: DialogProps) {
           </DialogBody>
           <DialogFooter>
             <FooterActions align='spaceBetween'>
-              <span className={styles.buttonGroup}>
-                <FooterActionButton
-                  onClick={() => runtime.openHelpWindow('multiclient')}
-                >
-                  {tx('troubleshooting')}
-                </FooterActionButton>
-                {stage === 'awaiting_scan' && svgUrl && qrContent && (
-                  <FooterActionButton onClick={copyQrToClipboard}>
-                    {tx('global_menu_edit_copy_desktop')}
-                  </FooterActionButton>
-                )}
-              </span>
-              <FooterActionButton onClick={cancel}>
-                {tx('cancel')}
+              <FooterActionButton
+                onClick={() => runtime.openHelpWindow('multiclient')}
+              >
+                {tx('troubleshooting')}
               </FooterActionButton>
+              {stage === 'awaiting_scan' && svgUrl && qrContent && (
+                <FooterActionButton onClick={copyQrToClipboard}>
+                  {tx('global_menu_edit_copy_desktop')}
+                </FooterActionButton>
+              )}
             </FooterActions>
           </DialogFooter>
         </>
--- a/packages/frontend/src/components/dialogs/SetupMultiDevice/styles.module.scss
+++ b/packages/frontend/src/components/dialogs/SetupMultiDevice/styles.module.scss
@@ -63,10 +63,6 @@
   margin-inline-start: 10px;
 }
 
-.buttonGroup {
-  display: flex;
-}
-
 .receiveSteps {
   padding: 4px 22px;
 }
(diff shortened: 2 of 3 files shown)


r10s:
fine with "Add Second Device".

for the "QR invite code", all other UI still use the old approach, if we want to switch here as well (as svg in practise creates problems and does not solve them, it is not really nice fitting into existing UI) we probably want to add the avatar manually and also the text.

so, to close https://github.com/deltachat/deltachat-desktop/issues/5969, i'd postpone reworking "QR invite code" and do that in a separate issue. this needs some discussion before.

nicodh:
Ok, I reverted all changes for the QR invite codes.

Only the second device qr code is affected now

packages/frontend/src/components/dialogs/SetupMultiDevice/SendBackupDialog.tsx
```diff
@@ -179,6 +180,7 @@ export function SendBackupDialog({ onClose }: DialogProps) {
                     )}
                   </SendBackupMainProgress>
                 </SendBackupMain>
+                {stage !== 'transferring' && <SendBackupSteps />}
```
WofWca, line 180: Why has 1e6d1de872b4f6be0f2bd6b35049749547f82ec8 been reverted (partially)?

nicodh: It seemed to me that it's better to show the steps below the qr code image. Since the qr code is the most important content

packages/frontend/src/components/dialogs/SetupMultiDevice/styles.module.scss
```diff
@@ -55,7 +55,7 @@
 
 .qrCode {
   width: 100%;
-  max-height: 60vh;
+  max-height: 45vh;
```
WofWca, line 58: FTR this particular change makes QR scannability worse and not better. (it makes the QR smaller if window height is > 600px).

packages/frontend/src/components/dialogs/SetupMultiDevice/SendBackupDialog.tsx
```diff
@@ -85,8 +85,10 @@ export function SendBackupDialog({ onClose }: DialogProps) {
       setError(null)
       setStage('preparing')
       const transfer = BackendRemote.rpc.provideBackup(accountId)
-      setQrSvg(await BackendRemote.rpc.getBackupQrSvg(accountId))
-      setQrContent(await BackendRemote.rpc.getBackupQr(accountId))
+      const qrCode = await BackendRemote.rpc.getBackupQr(accountId)
+      const qrCodeSvg = await BackendRemote.rpc.createQrSvg(qrCode)
```
WofWca, line 89: Also note that with the border removed, the "very small window" case which we used to support since 1e6d1de872b4f6be0f2bd6b35049749547f82ec8 (https://github.com/deltachat/deltachat-desktop/pull/6011) is no longer supported: the QR code doesn't fit.
But maybe that situation is not realistic, and we need a different approach to solve it anyway.

packages/frontend/src/components/dialogs/SetupMultiDevice/SendBackupDialog.tsx
```diff
@@ -161,13 +163,12 @@ export function SendBackupDialog({ onClose }: DialogProps) {
...
             <DialogContent>
               <SendBackup>
-                {stage !== 'transferring' && <SendBackupSteps />}
                 <SendBackupMain>
                   {stage === 'awaiting_scan' && svgUrl && qrContent && (
                     <img className={styles.qrCode} src={svgUrl} />
                   )}
                   <SendBackupMainProgress
-                    style={stage === 'transferring' ? { width: '100%' } : {}}
+                    style={stage === 'transferring' ? { width: '90%' } : {}}
```
WofWca, line 168: This change is irrelevant, right?

I like 100% more, but OK.

Image

packages/frontend/src/components/dialogs/SetupMultiDevice/styles.module.scss
nicodh: Yes but the idea was: while we need 60vH for small screens we don't need it for bigger screens and you will see more content on bigger screens if the qr code is smaller. It is a result of my testing with different screen sizes. We want to have no scroll bar for as many scenarios as possible.

packages/frontend/src/components/dialogs/SetupMultiDevice/SendBackupDialog.tsx
nicodh: You have a screen to show what you mean with "the qr code doesn't fit"? I tried various scenarios and the smallest screen I still can see the full qr code was something 300 x 500px if I recall right

WofWca: See the screenshot below "Very small window" in https://github.com/deltachat/deltachat-desktop/pull/6011. At that screen size on this MR the QR no longer fits.

I will submit an MR where I remove these fat buttons to give the QR code some more breathing space.

packages/frontend/src/components/dialogs/SetupMultiDevice/styles.module.scss
WofWca: Perhaps perhaps. My thinking was that the bigger the QR is the better.

PR #6089, opened 2026-03-07
Add channel tests

nicodh:


packages/e2e-tests/tests/qrcode-tests.spec.ts
```diff
@@ -309,11 +308,141 @@ test('wrong qr code for onboarding shows error message', async ({
...
+  const userAProfile = await getProfile(page, userA.id, true)
+  const { address, password } = userAProfile
+  if (!password) {
+    throw new Error(`Profile ${userA.name} has no password!`)
+  }
+  const newUsername = userA.name + '2'
+
+  await page.evaluate(
+    ([addr, pass]) => {
+      navigator.clipboard.writeText(
```
WofWca, line 396: Note that there is a linter warning here.

nicodh:
> Didn't look too closely, but more tests == good.
> 
> Also too much reliance on internal IDs and class names instead of human-readable text and labels for my liking, but we've talked about that already.

I will probably add a clickButton helper function that just takes the name and applies it to getByRole('button', {name}) in a followup

PR #6096, opened 2026-03-10
fix: do not cover video controls

### LINKED ISSUE #5209: video controls obstructed by message metatdata (time stamp and status)
When watching a video, fullscreen mode cannot be activated, and the video buttons are obscured by the time stamp and encryption indicator.

macOS 15.5
DC-Desktop 1.59.1 (Tauri)

Image

nicodh:
Resolves #5209

Before:

image

After

image

### DIFF
--- a/packages/frontend/src/components/attachment/messageAttachment.tsx
+++ b/packages/frontend/src/components/attachment/messageAttachment.tsx
@@ -172,9 +172,7 @@ export default function Attachment({
     }
     // the native fullscreen option is better right now so we don't need to open our own one
     return (
-      // content-below is always needed even if there is no text,
-      // because videos controls should not be overlapped by meta data
-      <div className={classNames('message-attachment-media', 'content-below')}>
+      <div className={classNames('message-attachment-media')}>
         <video
           className='attachment-content video-content'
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/attachment/messageAttachment.tsx
```diff
@@ -172,12 +172,9 @@ export default function Attachment({
...
     return (
-      <div
-        className={classNames(
-          'message-attachment-media',
-          withCaption ? 'content-below' : null
-        )}
-      >
+      // content-below is always needed even if there is no text,
+      // because videos controls should not be overlapped by meta data
+      <div className={classNames('message-attachment-media', 'content-below')}>
```
WofWca: Maybe it'd be simpler to adjust the CSS selector instead, by adding `.video-only` to it.

nicodh: What do you mean?  .video-only is used to adapt the whole message bubble background color.

WofWca: Yes, I understand. I mean we shouldn't apply the "contentbelow" class, and instead use the  "video only" class in CSS

PR #6098, opened 2026-03-10
fix: set min width for message body

### LINKED ISSUE #6090: Reaction overlapping with user avatar
instead of being shown under  the message it can happen that it overflows for stickers (when date label becomes long after some days)

Image

nicodh:
Avoid messages with less than 150 px width since the meta data might overlap the avatar

resolves #6090

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7098,9 +7098,9 @@ dependencies = [
 
 [[package]]
 name = "quinn-proto"
-version = "0.11.13"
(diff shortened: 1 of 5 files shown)


nicodh:
> Not really a proper fix, because the date can get even longer, so IDK if it's enough to close the issue, but this improves the situation.

I also tried longer date but no realistic length of that date covered the avatar with that width

Issue #6081, opened 2026-03-04
Tapping a call bubble in chat should answer/start call


r10s:
when not already in a call, tapping a call bubble (or the call icon inside), should answer the call (if not stale yet) or start an outgoing call

see https://github.com/deltachat/deltachat-ios/issues/3001

WofWca:
After #6099 only "or start an outgoing call" part of this issue is left to do.

And, well, maybe the issue with multiple call windows (see the MR's code TODO).

nicodh:
@r10s  Concerning the multiple call windows:
What is expected to happen if a call for a chat is already open and I click on a call message in another chat? Show a message like: "There is already a call open, you can only run one call at the same time"?
Or maybe ask to close the current call and open a new one?

WofWca:
Is there actually a good reason not to let people have multiple calls?
Edit: I mean with different people.

r10s:
- when tapping a bubble with a not-yet-stale-unaccepted-incoming call, accept the call

- when tapping a bubble with a contact we already have a call, the existing call window shall be raised (put to foreground)

- when tapping a bubble with a contact we do not have an ongoing call with, start the call - not matter if there is another call - we already allow multiple calls in other ways

WofWca:
> when tapping a bubble with a not-yet-stale-unaccepted-incoming call, accept the call

Currently we only open the "incoming call" window, where the user needs to press the green round button to accept the call. Is this what you meant?

r10s:
> Currently we only open the "incoming call" window, where the user needs to press the green round button to accept the call. Is this what you meant?

yes, that is good enough as well for now. i mainly wanted to make sure, that this does not result in another outgoing call :)

PR #6099, opened 2026-03-10
feat: click on call message to open call window

WofWca:
- **refactor: fix `IconButton` `className` prop**
- **refactor: named params for `openCallWindow`**
- **refactor: unify `startWithCameraEnabled`**
- **refactor: support inherited color for `<Icon>`**
- **refactor: `<IconButton>` to take `<button>` props**
- **refactor: turn "phone" icon into actual element**
- **feat: click on call message to open call window**

So far only when the call state is "alerting".
This is useful for when the caller is not among "who can call me"
(https://github.com/deltachat/deltachat-desktop/pull/6044).
Not perfect so far, but covers the use case.

Partially addresses
https://github.com/deltachat/deltachat-desktop/issues/6081.

### DIFF
--- a/packages/frontend/scss/_global.scss
+++ b/packages/frontend/scss/_global.scss
@@ -75,6 +75,9 @@ button {
   font-size: inherit;
   user-select: none;
 }
+button:disabled {
+  cursor: default;
+}
 
 a {
   color: var(--globalLinkColor);


packages/frontend/src/components/message/Message.tsx
```diff
@@ -1243,3 +1253,60 @@ function WebxdcMessageContent({
...
+      : undefined
+  // TODO fix: don't open a second window if one is already open
+  // for this message (this should be done in the main process).
+  const onClick = onClickParams
+    ? () => runtime.openIncomingVideoCallWindow(onClickParams)
+    : undefined
+
+  return (
+    <IconButton
+      aria-label='📞'
```
nicodh, line 1301: This is a strange label - doesn't make sense imo since it's not visible but should be readable

WofWca: What is not visible?
This label represents exactly what is on the screen, so sighted and blind people see the same.

WofWca:
> It is hard to detect that the phone icon is clickable when the rest of the message is not

Does that need to change? The use case is that you have changed "who can call me" and this get an "incoming call" message in the message list but the window doesn't pop up. I think whoever wants to accept such a call will try clicking the "phone" icon.
Also Telegram has this call message bubble UI.

The other potential use case is starting a new call, but we already have a dedicated button for that. We don't need to make this little button discoverable.

WofWca:
> 2. The cursor should not show a pointer when the click would have no effect anymore (when the call is finished)

I agree, I guess I'll change that.

> 3. clicking on an outgoing call message opens a new windows - there is a TODO but is that planned for this PR?

I think it's OK as is, production-ready. This MR covers a pretty rare use case I would say, and the "happy path" works.

packages/frontend/src/components/message/Message.tsx
nicodh: >  blind people see the same
??
you mean you expect a screenreader to translate that emoji to "phone" ? Maybe - I don't know...

WofWca: Yes, screen readers are supposed to read emojis. Mine reads "telephone receiver button".

PR #6102, opened 2026-03-11
chore: upgrade Electron to 41

WofWca:
We don't seem to be affected by breaking changes
https://github.com/electron/website/pull/1030.

Regarding "PDFs no longer create a separate WebContents":
we still disable PDF in WebXDC, disabling mime sniffing.
And overall, I have not investigated this, but it sounds
like a positive security change.

Regarding "Updated Cookie Change Cause
in the Cookie 'changed' Event":
we don't utilize cookies.

I have tested this a little. Works well.
Disabling `IsolateSandboxedIframes` still seems to be required
to make the "connectivity" dialog work.

### DIFF
--- a/packages/target-electron/package.json
+++ b/packages/target-electron/package.json
@@ -81,7 +81,7 @@
     "application-config": "^3.0.0",
     "chai": "^5.1.1",
     "debounce": "^1.2.0",
-    "electron": "^40.6.0",
+    "electron": "^41.0.0",
     "electron-builder": "^26.7.0",
     "esbuild": "^0.25.0",
     "mocha": "^10.7.0",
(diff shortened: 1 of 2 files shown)


nicodh:
Are you sure we want to upgrade to a 16h old release?

WofWca:
As usual, I think it's better to have latest Chromium security patches, see
previous discussions:
- https://github.com/deltachat/deltachat-desktop/pull/5970#issuecomment-3801127463
- https://github.com/deltachat/deltachat-desktop/issues/5816

Plus we still have some time until release.
Plus I think that Electron's "stable" channel can be considered stable enough, they don't seem to be breaking things often.

nicodh (approved):
Even when this might delay the next arch release (they are still waiting for electron release 40) it makes sense to stay upp-to-date with chrome updates

PR #6103, opened 2026-03-11
chore: rename `.vscode` files to `...template...`

WofWca:
Otherwise customizing these files, such as adding a build task,
is a pain, because git tracks changes in them.
VSCode doesn't yet seem to have
a one-size-fits-all solution for this:
- https://github.com/microsoft/vscode/issues/40233.
- https://github.com/microsoft/vscode/issues/15909.
- https://github.com/microsoft/vscode/issues/116740.


WofWca:
Well I'd say that everything beyond `package.json -> scripts` is a matter of personalization. Sure you might find other's settings useful but yep, not at a cost of having to always use the same settings as everyone.

PR #6105, opened 2026-03-11
chore: upgrade `electron-builder` & deps

WofWca:
What I did is first bumped its version in package.json,
followed by `pnpm install`.

Then `pnpm remove --recursive electron-builder`
and `pnpm add --save-dev electron-builder`.
The main point of this is to get rid of packages
that depend on tar@6 which has vulnerabilities.

Note that this adds the `electron-winstaller` dependency,
which has a build script which we did not add
to `onlyBuiltDependencies`.
This might break the production build. Let's test?

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -444,7 +444,7 @@ importers:
         version: 40.6.0
       electron-builder:
         specifier: ^26.8.1
-        version: 26.8.1(electron-builder-squirrel-windows@24.13.3)
+        version: 26.8.1(electron-builder-squirrel-windows@26.8.1)
       esbuild:
         specifier: ^0.25.0
         version: 0.25.8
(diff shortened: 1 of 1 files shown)


WofWca:
The things is that it might not be a minor upgrade actually. See MR description. This added the dependency `electron-winstaller`, which seems to me like a big change actually.
If something breaks, we can try approving its install script. It's coming from the Electron org, so it should be fine security-wise (we already have Electron itself approved).

WofWca:
Sorry perhaps I wasn't clear. The first commit didn't add that dependency, but the second one did.

WofWca:
So yep, since you, @nicodh, are in charge of releases for the most part, you merge at your discretion.
All I wanted is to get rid of audit issues.

nicodh:
Ok, I suggest to wait with merging since it might happen that we have to make a minor release soon if the Apple store needs a new release to start the review process from scratch

WofWca:
The point of this was to upgrade dependencies, which has been done in https://github.com/deltachat/deltachat-desktop/pull/6114.
Either way this needs to be redone.

Issue #5978, opened 2026-01-27
channels: show view count


r10s:
when a core with the [new api `get_message_read_receipt_count()`](https://github.com/chatmail/core/pull/7732) is released, we want to show the view count for channel owners, as folllows:

<img width="400" src="https://github.com/user-attachments/assets/68f66e55-ed06-42dc-9796-450cd061a191" />

the view count should be shown unconditionally, so also if 0.

the ticks shall not be shown, only pending/error state.

in the chatlist, neither view count nor ticks are shown, only pending/error state

counterpart of https://github.com/deltachat/deltachat-android/issues/4176 and https://github.com/deltachat/deltachat-ios/issues/2990

PR #6107, opened 2026-03-11
feat: add view count to own channel messages

WofWca:
- **feat: show view count on own channel posts**
- **improvement: hide delivery status for channels**

image

image

Closes https://github.com/deltachat/deltachat-desktop/issues/5978.

TODO:
- [x] Better icon


nicodh:
> better icon

What about https://fonts.google.com/icons?selected=Material+Symbols+Outlined:visibility:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=eye&icon.size=24&icon.color=%231f1f1f

These are the icons we use else where

WofWca:
New icon:

image

WofWca:
Fixed tests, hopefully improved accessibility.

WofWca:
Well IDK. I would say that this is a problem with accessibility and not a problem with tests. If we had some kind of tooltip for the icon or the `status` element then we could select by that.

WofWca:
Also do we want this in the upcoming release? Should I merge now?

PR #6108, opened 2026-03-11
improvement: hide view count while sending

WofWca:


nicodh (commented):
Why is this one liner in a separate commit?

WofWca:
Because this is not specified in the issue about "view count", so I split this change off so as to not block the base MR with potential discussions about this change.

nicodh:
Needs rebase

nicodh (approved):
Reviewed by reading

PR #6109, opened 2026-03-12
Render markdown in messages

darichey:
Add an off-by-default experimental option to render markdown in messages.

Uses react-markdown and supports headings, bold, italic, inline code, code blocks, blockquotes, and lists. Integrates existing linkify parsing for emails, hashtags, and bot commands.

Screenshot 2026-03-11 220916

### DIFF
--- a/packages/frontend/package.json
+++ b/packages/frontend/package.json
@@ -42,6 +42,7 @@
     "path-browserify": "^1.0.1",
     "react": "^19.2.0",
     "react-dom": "^19.2.0",
+    "react-markdown": "^10.1.0",
     "react-string-replace": "^1.1.1",
     "react-virtualized-auto-sizer": "^1.0.26",
     "react-window": "^1.8.11",
--- a/packages/frontend/scss/message/_message.scss
+++ b/packages/frontend/scss/message/_message.scss
@@ -634,10 +634,105 @@ $info-text-max-width: 400px;
   }
 }
 
+// Markdown container: override pre-wrap from .text so block elements
+// don't get extra whitespace from newlines in the rendered HTML
+.markdown-content {
+  white-space: normal;
+}
+
 .quoted-text,
 .msg-body > .text {
   a {
     text-decoration: underline;
+  }
+
+  // Make markdown headings more compact
+  h1,
+  h2,
+  h3,
+  h4,
+  h5,
+  h6 {
+    margin: 0;
+    line-height: 1.2;
+    user-select: text;
+  }
+
+  h1 {
+    font-size: 1.5em;
+  }
+
+  h2 {
+    font-size: 1.25em;
+  }
+
+  h3 {
+    font-size: 1.1em;
+  }
+
+  h4,
+  h5,
+  h6 {
+    font-size: 1em;
+  }
+
+  code {
+    font-family: monospace;
+    background-color: var(--messageCodeBg, rgba(0, 0, 0, 0.08));
+    padding: 2px 4px;
+    border-radius: 3px;
+    font-size: 0.9em;
+  }
+
+  pre {
+    margin: 4px 0;
+    padding: 8px 12px;
+    background-color: var(--messageCodeBg, rgba(0, 0, 0, 0.08));
+    border-radius: 6px;
+    overflow-x: auto;
+    max-width: 100%;
+
+    code {
+      background-color: transparent;
+      padding: 0;
+      border-radius: 0;
+      font-size: 0.85em;
+      white-space: pre;
+    }
+  }
+
+  p {
+    margin: 0;
+  }
+
+  blockquote {
+    margin: 2px 0;
+    padding-inline-start: 12px;
+    border-inline-start: 3px solid var(--messageText, currentColor);
+    opacity: 0.85;
+  }
+
+  ul,
+  ol {
+    margin: 2px 0;
+    padding-inline-start: 1.2em;
+    white-space: normal;
+  }
+
+  ul {
+    list-style-type: disc;
+  }
+
+  ol {
+    list-style-type: decimal;
+  }
+
+  li {
+    margin: 0;
+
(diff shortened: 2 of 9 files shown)


nicodh:
Thanks for your commit. But we had markup support in desktop before and removed it in https://github.com/deltachat/deltachat-desktop/pull/5502

The main reason: it doesn't make sense to have markup support in one client but not in others (the mobile apps). So the only way to get it in would be to have a solution for all clients.

We want to support it in future but it's currently no priority since we have other more important tasks.

r10s:
plus, while in general, some text formatting is wanted, it is unclear if this is markdown at the end.

might totally be, that the user does not see anything of the underlying format, but select text and then eg. "bold" from the context menu.

markdown is arguably more for advanced user, while delta aims to target non-advanced user.


i am moving this PR to "resurrection" therefore, where we collect ideas, that we may want to refer later to.

if you want to help, which is very welcome,   please pick one of the open issues, these are things agreed on and already triaged as a priority

rcano:
> might totally be, that the user does not see anything of the underlying format, but select text and then eg. "bold" from the context menu.
> 
> markdown is arguably more for advanced user, while delta aims to target non-advanced user.

Without trying to be rude, I'd like to counter argue to this: this feature, included in this manner, is to _visualize_ a message that you are getting in anyway. People used to markdown chat are sending markdown in their chats. We are used to it from discord, jira, reddit, slack and others, and none of these people are "advanced" other than having a job that requires a computer.
Furthermore the way it was implemented here, it's a self enabled, local option to visualize the message. It doesn't change anything for anybody (it doesn't enshrine a rich text format in the transport itself), and given that deltachat is built on the concept of trusted contacts that you add, it makes sense that they will chat with you in a way that makes sense to you.

> plus, while in general, some text formatting is wanted, it is unclear if this is markdown at the end.

and this wouldn't in anyway change that. It's basically the equivalent of a file browser having the feature of previewing a file-type.

darichey:
> The main reason: it doesn't make sense to have markup support in one client but not in others (the mobile apps). So the only way to get it in would be to have a solution for all clients.

I was aware of the previous support and understood the _main_ reasons to drop it, as listed in #5502, were entirely about the old implementation's technical debt (unmaintained library, link bugs, wasm complexity), not about the choice of markdown as a format or being limited to a subset of clients.

The approach taken here is extremely minimal and uses a well-maintained, battle-tested library. I won't claim that it is bug-free, but this implementation has a much smaller, easier-to-understand footprint than the previous one.

I understand the desire for consistency across clients, but I'd gently push back on the reasoning here. Getting support in all clients is a valid reason to block graduating the feature from experimental to stable, but the only path to having this on all clients is to land it on the first one.

Are you open to reconsidering?

link2xt:
We likely don't want to end up with markdown on the wire. It is underspecified, so it will then be very difficult to get the same formatting on all clients once you introduce markdown in one of the clients.

Email already has HTML as a standard for markup, it is good for on-the-wire format. UIs need plain text with spans such as https://core.telegram.org/api/entities
https://github.com/deltachat/message-parser is similar to that, but it outputs AST that is ok for rendering but not editing, so also not exactly what is wanted e.g. for Android to have an UI where user can select part of the text and mark it as bold instead of typing `*`.

rcano:
> We likely don't want to end up with markdown on the wire. It is underspecified, so it will then be very difficult to get the same formatting on all clients once you introduce markdown in one of the clients

Where does this come from? I fail to see how it's connected to the replies here. Who is talking about markdown on the wire?

link2xt:
> Who is talking about markdown on the wire?

PR is rendering the text that is sent in emails as markdown.

rcano:
> > Who is talking about markdown on the wire?
> 
> PR is rendering the text that is sent in emails as markdown.

That's a weird way to put it, sounds even misleading to me. A message is received, a message that happens to contain text that we collectively call markdown, and the optional feature renders it according to the markdown rules (or react-markdown rules to be more specific).
This is cosmetic feature. Has nothing to do with transport.

codiflow:
Totally agree with @rcano and @darichey here. And thanks for @darichey for this PR – a custom-built version with your PR is working just fine and makes my every day live with DC Desktop SO much easier as it increases readability of my messages 😎

At the end of February 2026 I was referring to this feature in [this thread](https://support.delta.chat/t/thoughts-roadmap-to-make-delta-chat-attractive-for-ordinary-users-rich-text-image-compression-quality/4808#p-21204-markup-in-messages-2) too and I really wonder why the DC team is not prioritizing this feature in any way as there seems to be a significant demand since years (!). And your PR shows that it can be achieved with a small code footprint. 🙏

Don't want to argue with the devs as it's their software and their choice what they implement. We all are free to build our custom versions. But as Arcane Chat for Android has this feature as well (and I prefer Arcane Chat over DC on Android because it implements features which increase usability) it would be great to have this as an opt-in in the Desktop version as well. And as it is an "experimental feature" by now it is available for those who explicitly want it and not for the masses.

From my point of view removing / denying an intermediate-PR which brings in clarity and readability to messages (especially with the code tags) in favor of a later-to-discuss "perfect" solution (HTML or whatever) doesn't make any sense to me. Especially because this PR does not endanger security in any kind or messes aroundwith the core library.

Would really prefer that the devs reconsider their decision from March 🙏

For those who want to build their own version from this PR to have a feature-rich desktop client, here's a simple tutorial:

- Clone the patched version from @darichey : `git clone -b render-markdown https://github.com/darichey/deltachat-desktop`
- Create a patch from the last commit: `git diff HEAD~1 > markdown-support.patch`
- Checkout newest version from DC Desktop: `git clone https://github.com/deltachat/deltachat-desktop && git checkout v2.49.1`
- Apply patch: `git apply /path/to/markdown-support.patch`
- Follow official build steps from https://github.com/deltachat/deltachat-desktop#from-source-

Issue #6080, opened 2026-03-04
"mentions" are not enabled by default


r10s:
when creating a new profile, we want "notifications" as well as "mentions" being enabled - this is what android/ios are doing and which seems to be the gist also on desktop, cmp https://github.com/deltachat/deltachat-desktop/pull/4633 , however, it does not work.

instead, mentions are disabled after creating a new profile.

tested on mac 2.43.0 as well as on main

WofWca:
It is enabled by default.

https://github.com/deltachat/deltachat-desktop/blob/47c235c97c1cccb998d38914638d387f96bf6651/packages/shared/state.ts#L32

https://github.com/deltachat/deltachat-desktop/pull/4633.

The thing is that it's a "desktop setting" so it affects all profiles. It's incorrect for it to be in the "current profile" section. It was like that right from the start https://github.com/deltachat/deltachat-desktop/issues/4538.

r10s:
ah, i see.

then the issue is that it is not a per-profile settings, which probably makes some sense, also wrt consistency with android/iOS.

WofWca:
We can have per-account map as we used to do already

https://github.com/deltachat/deltachat-desktop/blob/47c235c97c1cccb998d38914638d387f96bf6651/packages/shared/shared-types.d.ts#L51-L52

but it feels like by this point we should have moved the whole "mentions" logic to Core.

PR #6110, opened 2026-03-12
fix: make "mentions" setting per-account

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/6080.

Note that we do _not_ perform a migration
to apply the current value to all existing profiles.
This is on purpose, because in the current release version
the "Mentions" setting toggle is incorrectly placed
in the "Current Profile" (and not "All Profiles")
notification settings.
This means that some people might have disabled it
thinking that they only disabled it for one account,
without realizing that they actually disabled mentions
on all accounts.
It is not nice, but the best thing to do seems to be
to forcefully enable mentions on all accounts.
People will disable it then where needed.
Wrongly enabling mentions is better than wrongly disabling them.

I have tested this.

### DIFF
--- a/packages/frontend/src/system-integration/notifications.ts
+++ b/packages/frontend/src/system-integration/notifications.ts
@@ -350,6 +350,9 @@ async function flushNotifications(accountId: number) {
   let isMentionsEnabled: undefined | boolean
   async function getIsMentionsEnabled(): Promise<boolean> {
     if (isMentionsEnabled == undefined) {
+      // Note that we're not using `SettingsStoreInstance`
+      // because it only relevant for currently selected account
+      // whereas notifications can come for any accounts.
       const str =
         (await BackendRemote.rpc.getConfig(accountId, 'ui.mentions_enabled')) ??
         mentionsEnabledDefaultVal


packages/frontend/src/system-integration/notifications.ts
```diff
@@ -345,6 +347,17 @@ async function flushNotifications(accountId: number) {
     .filter(e => e.muted)
     .map(e => e.id)
 
+  let isMentionsEnabled: undefined | boolean
+  async function getIsMentionsEnabled(): Promise<boolean> {
+    if (isMentionsEnabled == undefined) {
+      const str =
+        (await BackendRemote.rpc.getConfig(accountId, 'ui.mentions_enabled')) ??
```
nicodh, line 357: is there a reason to get the value from rpc instead settings store, other than the fact that we want to get rid of the settings store anyway in future?

packages/shared/shared-types.d.ts
```diff
@@ -37,7 +37,10 @@ export interface DesktopSettingsType {
   locale: string | null
   notifications: boolean
   showNotificationContent: boolean
-  isMentionsEnabled: boolean
+  /**
+   * @deprecated moved to core, per-account `ui.mentions_enabled`.
+   */
+  isMentionsEnabled?: boolean
```
nicodh, line 43: What is the idea of marking as deprecated instead just removing it?

This is not a public API but only consumed inside the code we have under control.

The other deprecated annotations show there is no value in keeping such values.

I think we should remove it from existing json config files as well

nicodh:
My comments are no blockers but should be considered nevertheless

packages/frontend/src/system-integration/notifications.ts
WofWca: `SettingsStore` stores the settings for the current account only, but notifications can come for any account. I actually implemented it with `SettingsStore` first until I started testing it.

Maybe it's worth adding a comment then.

packages/shared/shared-types.d.ts
WofWca: See https://github.com/deltachat/deltachat-desktop/pull/5786#discussion_r2584015053. The property is still present on the object.

packages/frontend/src/system-integration/notifications.ts
```diff
@@ -345,6 +347,17 @@ async function flushNotifications(accountId: number) {
     .filter(e => e.muted)
     .map(e => e.id)
 
+  let isMentionsEnabled: undefined | boolean
+  async function getIsMentionsEnabled(): Promise<boolean> {
+    if (isMentionsEnabled == undefined) {
+      const str =
```
WofWca, line 356:
```suggestion
      // Note that we're not using `SettingsStoreInstance`
      // because it only relevant for currently selected account
      // whereas notifications can come for any accounts.
      const str =
```

packages/frontend/src/system-integration/notifications.ts
WofWca: https://github.com/deltachat/deltachat-desktop/pull/6110#discussion_r2940346192

packages/shared/shared-types.d.ts
nicodh: Ok, maybe we should start removing the old keys soon (in another PR)

PR #6113, opened 2026-03-12
fix: app picker: refresh on connectivity change

WofWca:
Follow-up to
d249fa5e098d4f8d35b8ec9885be4af33652e501
(https://github.com/deltachat/deltachat-desktop/pull/5445).

A side-effect of this is that we reload icons
on almost every `ConnectivityChanged`.


WofWca:
This is probably not important enough for the downside of reloading icons. It's fine for the user to simply reopen the app picker.

PR #6114, opened 2026-03-14
Update core 2.45.0

nicodh:
Since we have new stock strings I also updated the translations

To avoid having to delete empty files manually I added a task to our find-complete-translations  script to remove these files. It turned out, that hebrew is listed as language although it has no translations which doesn't make sense so we removed it for now as long as there are no translations available.

### DIFF
--- a/bin/find_complete_translations.sh
+++ b/bin/find_complete_translations.sh
@@ -1,13 +1,12 @@
 #!/bin/bash
 
 # checks if there are any new translations which reached the threshold to be included in the language selection
-# deletes files which are below the threshold
 
 threshold=150
 
 code=$(cat <<EOF
 
-const {readFileSync, unlinkSync} = require('fs');
+const {readFileSync} = require('fs');
 
 process.stdin.on('data', (d) => {
     let list = d.toString().split('\n').map(l=>{
@@ -18,12 +17,6 @@ process.stdin.on('data', (d) => {
     // remove non lang file entries
     list.pop()
     list.pop()
-    for (const [n, l] of list) {
-        if (l && n < $threshold) {
-            unlinkSync('_locales/' + l + '.xml')
-            console.log('deleted _locales/' + l + '.xml (only ' + n + ' lines)')
-        }
-    }
     // filter list
     list = list.sort(([n1], [n2])=> n1-n2)
     list = list.filter(([n])=> n >= $threshold)
@@ -37,7 +30,6 @@ process.stdin.on('data', (d) => {
             console.log(l + ' is not in languagelist, despite having enough lines, maybe it is new')
         }
     }
-
 })
 
 EOF
--- a/.tx/config
+++ b/.tx/config
@@ -5,8 +5,7 @@ host = https://app.transifex.com
 file_filter            = _locales/<lang>.xml
 source_file            = _locales/en.xml
 type                   = ANDROID
-minimum_perc           = 0
+minimum_perc           = 20
 resource_name          = strings.xml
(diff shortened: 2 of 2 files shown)


pnpm-lock.yaml
```diff
@@ -6282,6 +5960,18 @@ snapshots:
 
   electron-to-chromium@1.5.267: {}
 
+  electron-winstaller@5.4.0:
```
WofWca, line 5963: So are we fine with doing this upgrade after all? We talked about this in https://github.com/deltachat/deltachat-desktop/pull/6105.

bin/find_complete_translations.sh
WofWca, line 3: The name of the file and this commend (and maybe documentation?) need to be updated.

bin/find_complete_translations.sh
```diff
@@ -17,6 +17,12 @@ process.stdin.on('data', (d) => {
     // remove non lang file entries
     list.pop()
     list.pop()
+    for (const [n, l] of list) {
+        if (l && n < $threshold) {
```
WofWca: And are we OK with deleting files that actually do contain some strings? E.g. `nah`, `tg`.
And overall - I don't know much about the packaging process, but what's the point of _deleting_ the files? Isn't this supposed to be only about maintaining the contents of the `_languages.json` file?

nicodh: Well there were more empty files that were downloaded on translations update and I deleted them manually. I didn't know that there were other empty files that were committed (including one that even was listed in _languages.json

So either we go with this script and don't add unused files to the repository or we add all empty files just like they are downloaded by the update script. I decided to go with the first variant.

pnpm-lock.yaml
nicodh: I thought the issue with #6105 was if we want to upgrade electron-builder from 26.7 to 26.8. It's a bit weird that  electron-builder-squirrel-windows was still on 24.x while electron-builder was on 26.x.

I didn't look deeper into that but just tested if it works (when I updated electron-builder 26.x). So you expect that this change affects the build process even if electron-builder stays on 26.7 ?

I will have a look on the build machine. And if it causes trouble I still can downgrade again, so let's not block the core upgrade

WofWca: > So you expect that this change affects the build process even if electron-builder stays on 26.7 ?

yes. See the MR that I mentioned:

> Note that this adds the `electron-winstaller` dependency, which has a build script which we did not add to `onlyBuiltDependencies`.

bin/find_complete_translations.sh
WofWca: > And are we OK with deleting files that actually do contain some strings?

I thought that the fact that a language is not in `_languages.json` simply means that it will not be shown in app stores, but then I realized that this is about whether you can pick it in the language selector.

But still - what's the point of deleting the files? And can't we just configure the behavior in Transifex CLI config?

https://github.com/deltachat/deltachat-desktop/blob/7423988b5760f0211ce77875a2855fed48cad541/.tx/config#L1-L12

bin/find_complete_translations.sh
WofWca: No longer relevant

bin/find_complete_translations.sh
WofWca: Done.

WofWca (approved):

WofWca:
Maybe it's worth doing another `tx pull` so as to sync the xml files again.

PR #6124, opened 2026-03-17
fix: user specific tmp dir if needed

### LINKED ISSUE #5479: Invalid temp folder permissions on non-primary user on linux
- Operating System (Linux/Mac/Windows/iOS/Android): Linux - nixos
- Delta Chat Version: 2.11.0
- Expected behavior: Dropping in files, pasting in files from clipboard, sending webxdc apps from built-in store (and possibly also other features using temp files) work.
- Actual behavior: nothing happens in these cases, developer console shows permission denied error.
   1. Install delta chat system wide on linux -> not flatpak.
   2. use delta chat on the first user: drop in a file and send it to confirm that it works
   3. create a second account
   4. open delta chat on the second account and also try to drop in a file to send it.
   5. observe that nothing happens and there is a permission error in the log
- Logs: TODO

The first user of delta chat temp files creates and owns the directory for dc temp files  meaning that other users can't use it.
Also I haven't checked if this also affects mac and windows, they have different temporary file paths so could be that they are unaffected.

### solution ideas:
 
A. maybe we should create a tmp directory per system user?
B not have a `/tmp/[deltachat tmp folder name]/[filename]/[file]` directory, but instead just /tmp/[deltachat tmp folder name]-[filename]/[file]`.
    - but that would make the cleanup of all old temp files (that were not deleted for whatever reason) more difficult (instead of clear folder it would mean look for all folders with the delta chat prefix)
 C store temporary files in the blob folder of the use, where core cleans them up automatically after one hour
    - may not be a good idea for some use cases like copying temp files out for opening them in an external program
    - may make problems on windows: say you are choosing an avatar in desktop but don't save/apply it for an hour, then core tries to delete it, but since it is shown to the user in electron, electron may still hold the lock on the file and the deletion would fail
D. something else?

nicodh:
resolves #5479

### DIFF
--- a/packages/target-electron/src/application-constants.ts
+++ b/packages/target-electron/src/application-constants.ts
@@ -68,7 +68,12 @@ export function getCustomThemesPath() {
 // this is used for temporary files (because core expects file paths, can not accept blobs directly yet)
 // used when sending file from webxdc and when pasting a file from clipboard
 export function getDraftTempDir() {
-  return join(app.getPath('temp'), 'chat.delta.desktop-draft')
+  // On POSIX (Linux, macOS) the system temp dir is shared between users
+  // (e.g. /tmp), so we append the numeric UID to avoid permission conflicts
+  // if multiple users use the same tmp dir
+  const uid = process.getuid?.()
+  const suffix = uid !== undefined ? `-${uid}` : ''
+  return join(app.getPath('temp'), `chat.delta.desktop-draft${suffix}`)
 }
 
 export const supportedURISchemes = [


packages/target-electron/src/application-constants.ts
```diff
@@ -68,7 +68,11 @@ export function getCustomThemesPath() {
 // this is used for temporary files (because core expects file paths, can not accept blobs directly yet)
 // used when sending file from webxdc and when pasting a file from clipboard
 export function getDraftTempDir() {
-  return join(app.getPath('temp'), 'chat.delta.desktop-draft')
+  // On POSIX (Linux, macOS) the system temp dir is shared between users
+  // (e.g. /tmp), so we append the numeric UID to avoid permission conflicts.
```
WofWca: Would be nice to clarify what permission conflicts we're talking about? As far as I got it, this is about multiple users running Delta Chat.

nicodh: It's described in the issue, but yes I can add a note about that

packages/target-electron/src/application-constants.ts
```diff
@@ -68,7 +68,12 @@ export function getCustomThemesPath() {
 // this is used for temporary files (because core expects file paths, can not accept blobs directly yet)
 // used when sending file from webxdc and when pasting a file from clipboard
 export function getDraftTempDir() {
-  return join(app.getPath('temp'), 'chat.delta.desktop-draft')
+  // On POSIX (Linux, macOS) the system temp dir is shared between users
+  // (e.g. /tmp), so we append the numeric UID to avoid permission conflicts
+  // if multiple users use the same tmp dir
```
WofWca, line 73:
```suggestion
  // of multiple users using the same tmp dir
```

Otherwise it sounds like we expect multiple users to use the same dir.

PR #6125, opened 2026-03-17
improvement: remove Cancel button from Add Device

WofWca:
Move it to the dialog header, as a "cross" (close) button.

This gives more space to the QR code itself.

Also this makes the dialog look closer to how it does on Android.


WofWca:
Fixed tests.

PR #6133, opened 2026-03-19
chore: update translations

nicodh:


bin/setup-apparmor-dev.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
```
WofWca: Unrelated file

_locales/az.xml
```diff
@@ -201,7 +207,9 @@
         <item quantity="one">%d yeni messaj</item>
         <item quantity="other">%d yeni messaj</item>
     </plurals>
-    <string name="chat_new_group_hint">Başqalarının bu qrup daxilində cavab verməsinə imkan verən ilk mesajı yaradın. \n\n• Bütün üzvlər Delta Chat istifadə etsələr daha yaxşıdır. \n\n• İlk mesajı çatdırmaq bir müddət ala bilər.</string>
+    <string name="chat_new_group_hint">Başqalarının bu qrup daxilində cavab verməsinə imkan verən ilk mesajı
+        yaradın. \n\n• Bütün üzvlər Delta Chat istifadə etsələr daha yaxşıdır. \n\n• İlk mesajı
+        çatdırmaq bir müddət ala bilər.</string>
```
WofWca, line 212: Have you checked whether these added line breaks are stripped off in production?

nicodh: No. But these line breaks are not introduced by this PR

WofWca: Do you see any other strings with such line breaks on current `main`? I do not. `        (?!<)` search in `_locales` gives 0 results.

nicodh: No but we never touched the translations. That must happen in transifex.

r10s: the linebreaks were added accidentally when adding autoformatting in android, see recent chats and https://github.com/deltachat/deltachat-android/pull/4296#issuecomment-4089406548

for the english source language, they're removed meanwhile again (as they are shown to translators which is annoying, as meaningless, also they cause retranslations as english source is used as key)

the translations, however, can have additional linebreaks - but i assume, desktop ignores them as well as android and uses only explicit `\n`

WofWca: I was talking about the actual line breaks that this MR adds to this file, and not the `\n`s that were already present.

WofWca: And I just tested. They are _not_ stripped off in production.

image

(I manually modified the "stickers" string).

Most of the time they are not actually visible though, but we might have elements that preserve white-space.

image

WofWca: For instance, device messages are problematic. This is `device_talk_welcome_message2` with line breaks added by me for testing:

image

@r10s Android is probably affected by this issue as well, I don't think whitespace would be ignored for stock strings.

nicodh: But we should not block the whole translation update just because there are small errors in one translation file, so I don't understand why you consider that a blocker (not approving)

nicodh:
Ok, looking again at https://github.com/deltachat/deltachat-android/pull/4296 I see that the strings are modified after downloading them. I assumed the strings were fixed on transifex and the only error existed in one language.

So I will add a similar step in desktop

WofWca:
> I see that the strings are modified after downloading them

I'm not sure where you get that from? The MR you reference changes the _source_ file, which I don't think gets downloaded from Transifex.

WofWca:
Either way I think the whitespace issues need to be fixed upstream because they are not just about formatting.

nicodh:
Ok, I added an additional step in the removeJunk function in xmlToJson.

So the JSON looks good now and afaik only the json is used

nicodh:
Newer translations are in https://github.com/deltachat/deltachat-desktop/pull/6137 to avoid another rebase

PR #6137, opened 2026-03-19
Unpublish transport

nicodh:
- shows a button to unpublish a transport instead deleting it
- shows a hint "hidden from contacts" in the transportlist

image

### DIFF
--- a/packages/frontend/src/components/dialogs/Transports/index.tsx
+++ b/packages/frontend/src/components/dialogs/Transports/index.tsx
@@ -1,10 +1,16 @@
 import React, { useCallback, useEffect, useState } from 'react'
 import { DialogProps } from '../../../contexts/DialogContext'
-import Dialog, { DialogBody, DialogHeader, DialogFooter } from '../../Dialog'
+import Dialog, {
+  DialogBody,
+  DialogHeader,
+  DialogFooter,
+  FooterActions,
+  FooterActionButton,
+  DialogContent,
+} from '../../Dialog'
 import { BackendRemote, onDCEvent } from '../../../backend-com'
 
 import useTranslationFunction from '../../../hooks/useTranslationFunction'
-import useConfirmationDialog from '../../../hooks/dialog/useConfirmationDialog'
 import useAlertDialog from '../../../hooks/dialog/useAlertDialog'
 import BasicQrScanner from '../BasicScanner'
 import EditAccountAndPasswordDialog from '../EditAccountAndPasswordDialog'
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/dialogs/Transports/index.tsx
```diff
@@ -237,3 +255,60 @@ export default function TransportsDialog(
...
+        <FooterActions align='spaceBetween'>
+          <FooterActionButton onClick={onClose}>
+            {tx('cancel')}
+          </FooterActionButton>
+          <FooterActionButton onClick={hideFromContacts}>
+            {tx('hide_from_contacts')}
+          </FooterActionButton>
+          <FooterActionButton styling='danger' onClick={removeTransport}>
+            {tx('remove_transport')}
+          </FooterActionButton>
```
WofWca, line 309: I'm a little unsure about the desired button order. I think we want "Hide from Contacts" to be the primary one? On Android the "Cancel" button is in the middle, so IDK what's that supposed to be:

Image

We have this "if in doubt" piece of text, which also points in the direction of "unpublish" having to be the primary button.

WofWca (approved):
Regarding the last commit of this MR: looks good and works, implements the UI that is also implemented in Android.
The other commit are a part of this MR: https://github.com/deltachat/deltachat-desktop/pull/6133.

I also tried this with a screen reader, and the new text is read OK.

packages/frontend/src/components/dialogs/Transports/index.tsx
nicodh: Yes I was also a bit unsure about that. Maybe @r10s has a opinion here?

r10s: it is good as it is currently on desktop. the selected action is "delete", so that should be the primary destructive action, with style as usualy (red and iirc not selectedable by enter).

the "hide from contacts" is just an alternative.

the button order on android is chosen also to not hit the wrong choice with a big thumb, this is not an issue on desktop.

r10s:
ah, a litte too late: can we have the string "Messages are received on all relays." (`transport_list_hint`) below the list? if it is too complex (as on android), however,  it should not block the release

see screenshot at https://github.com/deltachat/deltachat-ios/pull/3059 to get the idea

nicodh:
> ah, a litte too late: can we have the string "Messages are received on all relays." (`transport_list_hint`) below the list? if it is too complex (as on android), however, it should not block the release

Sure not a big thing.

nicodh:
@r10s any opinion about the button order? Since I anyway make a followup now

WofWca:
> any opinion about the button order?

Didn't he reply already? https://github.com/deltachat/deltachat-desktop/pull/6137#discussion_r2966115967

r10s:
great!

the button order is fine, see my comment above at https://github.com/deltachat/deltachat-desktop/pull/6137#discussion_r2966115967

nicodh:
Sorry I'm travelling all day with flaky internet

PR #6143, opened 2026-03-20
feat: show transport hint

nicodh:
image

### DIFF
--- a/packages/frontend/src/components/dialogs/Transports/styles.module.scss
+++ b/packages/frontend/src/components/dialogs/Transports/styles.module.scss
@@ -59,9 +59,3 @@ $iconSize: 1.3em;
 .addTransportButton {
   padding: 10px;
 }
-
-.transportHint {
-  font-size: 0.9em;
-  color: var(--textSecondary);
-  margin: 10px 10px 0 32px;
-}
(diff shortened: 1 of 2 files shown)


nicodh:
I also adapted the button to be more explicit

packages/frontend/src/components/dialogs/Transports/styles.module.scss
```diff
@@ -59,3 +59,9 @@ $iconSize: 1.3em;
 .addTransportButton {
   padding: 10px;
 }
+
+.transportHint {
+  font-size: 0.9em;
+  color: var(--textSecondary);
+  margin: 10px 10px 0 32px;
```
r10s: just use the same styling as for the "Disappearing Messages" and comparable hints.

```suggestion
  margin-top: 10px;
```

r10s:
> I also adapted the button to be more explicit

ah, i just wanted to file an issue for that: the button needs to be left. it is not smth as "OK" "Cancel". it should stay left

the new text is okay

packages/frontend/src/components/dialogs/Transports/styles.module.scss
nicodh: I used p tag now similar to other hints so we need no margings at all

PR #6145, opened 2026-03-21
Fixed wrong parameter and clarified some texts

codiflow:
In the most recent build of sass (1.98.0) theres no option names `--include-path`. I always get the message `Could not find an option named "--include-path".`

Had to use `--load-path` or `-I` instead.

### DIFF
--- a/docs/THEMES.md
+++ b/docs/THEMES.md
@@ -78,7 +78,7 @@ We use SCSS to create themes, because it allows us to save work by using its col
 4. compile the theme to CSS in watch mode.
 
 ```sh
-# For Windows:
+# General:
 pnpm sass --watch path/to/Deltachat/config/folder/custom-themes/my_theme.scss path/to/Deltachat/config/folder/custom-themes/my_theme.css -I packages/frontend/themes/ --no-source-map
 # For Linux:
 pnpm sass --watch ~/.config/DeltaChat/custom-themes/my_theme.scss ~/.config/DeltaChat/custom-themes/my_theme.css -I packages/frontend/themes/ --no-source-map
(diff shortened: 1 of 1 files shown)


docs/THEMES.md
```diff
@@ -109,12 +110,13 @@ Read the comments in the theme files for additional information.
 3. compile the theme to CSS in watch mode.
 
 ```sh
-sass -w path/to/Deltachat/config/folder/custom-themes/my_theme.scss path/to/Deltachat/config/folder/custom-themes/my_theme.css --include-path path/to/deltachat-desktop-git-folder/themes/
-# for linux:
-sass -w ~/.config/DeltaChat/custom-themes/my_theme.scss ~/.config/DeltaChat/custom-themes/my_theme.css --include-path themes/
+# For Windows:
```
WofWca: I don't think that this is necessarily "for Windows". This is simply a "generic" version.

codiflow: Thanks for approving the PR 🙏

Regarding the package `dart-sass` – I just searched for it in the Arch Linux / Manjaro package sources and could find it in the `extra` repository (so not even AUR but the official ones):

`$ yay -Ss dart-sass`
```
extra/dart-sass 1.97.3-1 (3.6 MiB 10.5 MiB)
    Sass makes CSS fun again
```

Don't know if the package is in the Debian/Ubuntu repositories too.

WofWca: Alright so about the command line thing, would you like to revert the "For Windows" thing? Or do you disagree?

codiflow: I just removed the "Windows" part and replaced it with "General". Hope that's ok.

I added it because during the phase of going through the tutorial it was confusing me. I started to add the first command and then I saw that there was a "linux specific" command below. I did not expect there being two lines for the same command when theres no "descriptive" comment before the first line. So I was thinking these are two separate commands.

Maybe this was just an issue for me but having a comment for each line makes it even more clear that it's the same command just for different OSs. 😉

WofWca: Sure, I am not claiming that the documentation is in its best state currently.

nicodh (approved):
Reviewed by reading

PR #6150, opened 2026-03-24
fix: video overlapping with text below

WofWca:
This partially reverts
ac5e5979e7f84a38ca18a74106278e67d62f4d6b
(https://github.com/deltachat/deltachat-desktop/pull/6096),
which introduced this issue.

What I meant in
[my suggestion](https://github.com/deltachat/deltachat-desktop/pull/6096#discussion_r2919693993)
on that MR is that we don't need to _change_ the condition
when `content-below` is applied, and not to completely remove it.

I have tested that the metadata still doesn't overlap
with video controls if there is no text,
and that the video doesn't overlap with text if there _is_ text.


WofWca:
Discussed in private, since this fixes a regression, it's probably better to merge this now, right before a release.

PR #6151, opened 2026-03-24
refactor: simplify video attachment CSS

WofWca:
This also slightly (`7px`) increases the margin
between the video and the metadata element.

Let's not be deeply overriding `margin-bottom`
(or rather introducing a padding to compensate for negative margin)
and instead define it in the same place
where we have the `.content-below` styles.

Follow-up to de291375d, ac5e5979e7f84a38ca18a74106278e67d62f4d6b.

### DIFF
--- a/packages/frontend/scss/message/_message-attachment.scss
+++ b/packages/frontend/scss/message/_message-attachment.scss
@@ -91,8 +91,12 @@
     }
   }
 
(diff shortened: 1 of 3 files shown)


WofWca:
I'm not sure what you mean? This MR doesn't have any effect on messages that have both a video and text.

PR #6153, opened 2026-03-24
feat: leave & delete group

### LINKED ISSUE #6104: leave groups/channels before deletion
after discussion at https://github.com/deltachat/deltachat-android/issues/4158, consent was that it is better ux to leave a group/channel before its deletion and to simplify the options.

implementation was done at https://github.com/deltachat/deltachat-android/pull/4262 and https://github.com/deltachat/deltachat-ios/pull/3048 - see the **screenshots** there to get an idea.

to do the same on desktop, the following is needed:

- create a `shallLeaveBeforeDelete` function, see implementations above for details

- in the chat's three-dot-menu, depending on `shallLeaveBeforeDelete`, show _either_ "Leave"  _or_ "Delete Chat" (when saying "Leave" i mean "Leave Group"  resp. "Leave Channel")

- "Leave" as well as "Delete Chat" shall the be last ones in the menu, before "Clear Chat"

- when tapping "Leave", the alert shall have the buttons "Leave", "Leave & Delete for Me" and "Cancel"

- when tapping "Delete Chat", the alert shall have the Option "Delete for Me"  and "Cancel"

- in the chatlist, where multi-selection is allowed, rename "Delete Chat" to a simple "Delete" (this fixes plural issue, when on that, do the same for "Pin" and "Archive"). no need to have "Leave" as an option there

- when selecting "Delete" _and_ a group or channel is selected, use the button label "Leave & Delete for Me". otherwise, use "Delete for Me" 

all strings are available in the translations as `menu_leave_group`, `menu_leave_channel`, `menu_leave_and_delete`, `menu_delete_chat`, `delete_for_me`

nicodh:
resolves #6104

### DIFF
--- a/packages/e2e-tests/tests/chat-context-menu.spec.ts
+++ b/packages/e2e-tests/tests/chat-context-menu.spec.ts
@@ -130,8 +130,8 @@ test.describe('Main View - 3-Dot Menu', () => {
       'Mute Notifications',
       'Archive Chat',
       'Block Contact',
-      'Delete Chat',
       'Clear Chat',
+      'Delete Chat',
     ])
 
     await closeMenu(page)
@@ -154,8 +154,8 @@ test.describe('Main View - 3-Dot Menu', () => {
       'Mute Notifications',
       'Archive Chat',
       'Clone Chat',
-      'Leave Group',
       'Clear Chat',
+      'Leave Group',
     ])
 
     await closeMenu(page)
@@ -398,7 +398,7 @@ test.describe('Chat List Context Menu - Single Selection', () => {
       'View Profile',
       'Encryption Info',
       'Clone Chat',
-      'Delete',
+      'Leave Group',
     ])
 
     await closeMenu(page)
--- a/packages/e2e-tests/tests/chat-context-menu.spec.ts
+++ b/packages/e2e-tests/tests/chat-context-menu.spec.ts
@@ -208,7 +208,7 @@ test.describe('Main View - 3-Dot Menu', () => {
       page.getByRole('menuitem', { name: 'Leave Group' })
     ).toBeVisible()
     await expect(
-      page.getByRole('menuitem', { name: 'Delete Chat' })
+      page.getByRole('menuitem', { name: 'Delete' })
     ).not.toBeVisible()
 
     // Leave the group (without deleting)
@@ -224,7 +224,7 @@ test.describe('Main View - 3-Dot Menu', () => {
       page.getByRole('menuitem', { name: 'Delete Chat' })
     ).toBeVisible()
     await expect(
-      page.getByRole('menuitem', { name: 'Leave Group' })
+      page.getByRole('menuitem', { name: 'Leave' })
     ).not.toBeVisible()
     await closeMenu(page)
 
@@ -390,7 +390,7 @@ test.describe('Chat List Context Menu - Single Selection', () => {
 
     await openChatListContextMenu(page, groupName)
 
-    // Chat list menu items for group chat (no Leave Group in chat list)
+    // Chat list menu items for group chat
     await expect(getMenuItems(page)).toHaveText([
       'Pin Chat',
       'Mute Notifications',
--- a/packages/e2e-tests/tests/chat-context-menu.spec.ts
+++ b/packages/e2e-tests/tests/chat-context-menu.spec.ts
@@ -130,8 +130,8 @@ test.describe('Main View - 3-Dot Menu', () => {
       'Mute Notifications',
       'Archive Chat',
       'Block Contact',
-      'Clear Chat',
       'Delete Chat',
+      'Clear Chat',
     ])
 
     await closeMenu(page)
@@ -148,7 +148,6 @@ test.describe('Main View - 3-Dot Menu', () => {
 
     await openMainViewThreeDotMenu(page)
 
-    // Main view menu items for group chat (encrypted, member)
     await expect(getMenuItems(page)).toHaveText([
       'Search in Chat',
       'Disappearing Messages',
@@ -157,10 +156,90 @@ test.describe('Main View - 3-Dot Menu', () => {
       'Clone Chat',
       'Leave Group',
       'Clear Chat',
-      'Delete Chat',
     ])
 
     await closeMenu(page)
+  })
+
+  test('Leave Group opens leave dialog with correct options', async () => {
+    const userA = existingProfiles[0]
+
+    await switchToProfile(page, userA.id)
+    await page
+      .locator('.chat-list .chat-list-item')
+      .filter({ hasText: groupName })
+      .click()
+
+    await openMainViewThreeDotMenu(page)
+    await page.getByRole('menuitem', { name: 'Leave Group' }).click()
+
+    // Verify leave dialog opens with the expected options
+    const dialog = page.getByRole('dialog')
+    await expect(dialog).toContainText('Are you sure you want to leave?')
+    await expect(dialog.getByRole('button', { name: 'Cancel' })).toBeVisible()
+    await expect(
+      dialog.getByRole('button', { name: 'Leave & Delete for Me' })
+    ).toBeVisible()
+    await expect(
+      dialog.getByRole('button', { name: 'Leave Group' })
+    ).toBeVisible()
+
+    // Cancel to preserve state
+    await dialog.getByRole('button', { name: 'Cancel' }).click()
+  })
+
+  test('after leaving group, Delete Chat replaces Leave Group', async () => {
+    const userA = existingProfiles[0]
+    const disposableGroupName = 'Leave Then Delete Group'
+
+    await switchToProfile(page, userA.id)
+
+    // Create a group
+    await page.locator('#new-chat-button').click()
+    await page.getByRole('button', { name: 'New Group' }).click()
+    await page
+      .getByRole('textbox', { name: 'Group Name' })
+      .fill(disposableGroupName)
+    await page.getByTestId('group-create-button').click()
+
+    // Verify the group is selected and Leave Group is shown
+    await openMainViewThreeDotMenu(page)
+    await expect(
+      page.getByRole('menuitem', { name: 'Leave Group' })
+    ).toBeVisible()
+    await expect(
+      page.getByRole('menuitem', { name: 'Delete Chat' })
+    ).not.toBeVisible()
+
+    // Leave the group (without deleting)
+    await page.getByRole('menuitem', { name: 'Leave Group' }).click()
+    await page
+      .getByRole('dialog')
+      .getByRole('button', { name: 'Leave Group' })
+      .click()
+
+    // After leaving, re-open menu: Delete Chat should replace Leave Group
+    await openMainViewThreeDotMenu(page)
+    await expect(
+      page.getByRole('menuitem', { name: 'Delete Chat' })
+    ).toBeVisible()
+    await expect(
+      page.getByRole('menuitem', { name: 'Leave Group' })
+    ).not.toBeVisible()
+    await closeMenu(page)
+
+    // delete the chat
+    await openMainViewThreeDotMenu(page)
+    await page.getByRole('menuitem', { name: 'Delete Chat' }).click()
+    await page
+      .getByRole('dialog')
+      .getByRole('button', { name: 'Delete' })
+      .click()
+    await expect(
+      page
+        .locator('.chat-list .chat-list-item')
+        .filter({ hasText: disposableGroupName })
+    ).not.toBeVisible()
   })
 
   test('search in chat focuses search input', async () => {
@@ -298,7 +377,7 @@ test.describe('Chat List Context Menu - Single Selection', () => {
       'View Profile',
       'Encryption Info',
       'Block Contact',
-      'Delete Chat',
+      'Delete',
     ])
 
     await closeMenu(page)
@@ -311,7 +390,7 @@ test.describe('Chat List Context Menu - Single Selection', () => {
 
     await openChatListContextMenu(page, groupName)
 
-    // Chat list menu items for group chat
+    // Chat list menu items for group chat (no Leave Group in chat list)
     await expect(getMenuItems(page)).toHaveText([
       'Pin Chat',
       'Mute Notifications',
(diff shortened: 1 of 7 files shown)


packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -503,35 +526,36 @@ export function useChatContextMenu(): {
...
+      !shouldLeaveBeforeDelete && {
+        label: isMainView ? tx('menu_delete_chat') : tx('delete'),
+        action: onDeleteChats,
+        danger: true,
+      },
       isMainView && {
         label: tx('clear_chat'),
         action: onClearChat,
         danger: true,
       },
```
r10s: "Clear Chat" should be before leave/delete

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -22,6 +22,28 @@ import { ActionEmitter, KeybindAction } from '../../keybindings'
...
+    Partial<Pick<ChatListItem, 'isSelfInGroup' | 'isContactRequest'>>
+): boolean {
+  if (chat.chatType === 'InBroadcast') {
+    return !!chat.isSelfInGroup
+  }
+  if (
+    chat.chatType === 'Group' &&
+    chat.isEncrypted &&
+    chat.isSelfInGroup &&
+    !chat.isContactRequest
```
r10s, line 42: on android/ios, we're using `type != single && isEncrypted && canSend() && !isOutBroadcast` - at least the  isContactRequest should be replaced by canSend, as there are more reasons for not being able to send

but maybe just use the same logic in general, it is tested and running on android/ios since a while

packages/frontend/src/components/dialogs/LeaveGroupDialog.tsx
```diff
@@ -0,0 +1,62 @@
...
+            onClick={() => handleResult('leave-and-delete')}
+          >
+            {tx('menu_leave_and_delete')}
+          </FooterActionButton>
+          <FooterActionButton
+            styling='danger'
+            onClick={() => handleResult('leave')}
+          >
+            {isGroup ? tx('menu_leave_group') : tx('menu_leave_channel')}
+          </FooterActionButton>
```
r10s, line 57: let's swap the two buttons it is a bit more logical and in the flow how one reads.

```suggestion
          <FooterActionButton
            styling='danger'
            onClick={() => handleResult('leave')}
          >
            {isGroup ? tx('menu_leave_group') : tx('menu_leave_channel')}
          </FooterActionButton>
          <FooterActionButton
            styling='danger'
            onClick={() => handleResult('leave-and-delete')}
          >
            {tx('menu_leave_and_delete')}
          </FooterActionButton>
```

r10s (approved):
some minors but in general this lgtm

packages/frontend/src/hooks/chat/useChatDialog.ts
```diff
@@ -150,20 +165,24 @@ export default function useChatDialog() {
...
+          openDialog(LeaveGroupDialog, {
+            isGroup,
+            cb: resolve,
+          })
+        }
+      )
 
-      if (hasUserConfirmed) {
-        BackendRemote.rpc.leaveGroup(accountId, chatId)
+      if (result !== 'cancel') {
```
WofWca:
```suggestion
      if (result === 'leave' || result === 'leave-and-delete') {
```

Let's be more careful

packages/frontend/src/hooks/chat/useChatDialog.ts
```diff
@@ -150,20 +165,24 @@ export default function useChatDialog() {
...
+        await BackendRemote.rpc.leaveGroup(accountId, chatId)
+      }
+      if (result === 'leave-and-delete') {
+        unselectChat()
+        await EffectfulBackendActions.deleteChat(accountId, chatId)
```
WofWca:
```suggestion
        BackendRemote.rpc.leaveGroup(accountId, chatId)
      }
      if (result === 'leave-and-delete') {
        unselectChat()
        EffectfulBackendActions.deleteChat(accountId, chatId)
```

No need to slow down the operation by `await`ing.

packages/frontend/src/hooks/chat/useChatDialog.ts
```diff
@@ -133,6 +145,9 @@ export default function useChatDialog() {
             if (selectedChatId === chat.id) {
               unselectChat()
             }
+            if (shallLeaveBeforeDelete(chat)) {
+              await BackendRemote.rpc.leaveGroup(accountId, chat.id)
```
WofWca, line 153: Here there is also no need to await

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -22,6 +22,28 @@ import { ActionEmitter, KeybindAction } from '../../keybindings'
 
 const log = getLogger('ChatListContextMenu')
 
+/**
+ * Returns true if the user should leave the group/channel before deleting it.
+ * When true, the menu shows "Leave" instead of "Delete Chat".
+ */
+export function shallLeaveBeforeDelete(
+  chat: Pick<ChatListItem, 'chatType' | 'isEncrypted'> &
+    Partial<Pick<ChatListItem, 'isSelfInGroup' | 'isContactRequest'>>
```
WofWca: It looks like the function should document in what circumstances these fields can or must or should be omitted. Or we simply should reconsider whether we need `Partial` here in the first place.
The same goes for the `openDeleteChatsDialog` function.

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -472,12 +496,11 @@ export function useChatContextMenu(): {
       !isOutBroadcast &&
       relatedChat.chatType !== 'InBroadcast'
 
-    const showLeaveGroupOption =
+    // In the main view, show either "Leave" or "Delete Chat", not both
+    const shouldLeaveBeforeDelete =
```
WofWca, line 502: The name is incorrect. As you can see, this variable can be `false` despite `shallLeaveBeforeDelete()` returning `true`. This is also reflected in the fact that for group chats in the resulting dialog there is only the "Leave & Delete" button and no "delete only".

This should be named `showLeaveInsteadOfDelete` or something like that.

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -22,6 +22,28 @@ import { ActionEmitter, KeybindAction } from '../../keybindings'
 
 const log = getLogger('ChatListContextMenu')
 
+/**
+ * Returns true if the user should leave the group/channel before deleting it.
+ * When true, the menu shows "Leave" instead of "Delete Chat".
+ */
+export function shallLeaveBeforeDelete(
```
WofWca: So IMU in the end what we get is: if it's possible to leave the chat, then only show "Leave" and "Leave and Delete". If it's not possible to leave the chat, then only show "Delete".
Otherwise there would be cases where "delete" and "leave" are both present separately.

Then I think things would be much simpler if we named this `canLeaveChat`.

Yes, I saw that it's hypothetically possible to leave unencrypted chats, but now we basically disallow it: https://github.com/deltachat/deltachat-android/issues/4158#issuecomment-3769663920.

This is also in line with https://github.com/deltachat/deltachat-android/issues/4158#issuecomment-3954905289

> you cannot delete a chat without leaving before

packages/frontend/src/components/chat/ChatContextMenu.tsx
nicodh: I used the same conditions now

nicodh: Just saw that we use that also for multiple chats which are of a different type and don't have canSend. So I guess I have to revert the cange

packages/frontend/src/hooks/chat/useChatDialog.ts
nicodh: I used an early return now

packages/frontend/src/components/chat/ChatContextMenu.tsx
nicodh: I removed Partial by changing the outdated types in openDeleteChatsDialog

packages/frontend/src/hooks/chat/useChatDialog.ts
nicodh: One could argue that it's safer to await async functions especially when called inside a loop.

But I think it's neither a big overhead to await nor a huge improvement not to await.
Tbh not really worth discussing imo

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -472,12 +497,10 @@ export function useChatContextMenu(): {
...
       relatedChat.chatType !== 'InBroadcast'
 
-    const showLeaveGroupOption =
-      relatedChat &&
-      isGroup &&
-      relatedChat.isEncrypted &&
-      relatedChat.selfInGroup &&
-      !relatedChat.isContactRequest
+    // If a single chat is selected (or in main view),
+    // show either "Leave" or "Delete Chat", not both
```
WofWca, line 501: The "or in main view" part seems to be no longer relevant.
And overall the comment doesn't make much sense to me (as well as the variable name, but I mentioned that in a separate thread). The second line is always true and not only when a single chat is selected.

packages/frontend/src/components/chat/ChatContextMenu.tsx
WofWca: Ideally we should have such logic in the Core. If we mess it up then it becomes impossible for the user to leave a chat (instead they can only delete the chat).

At the very least I think we should have explicit logic for each `chatType` (swith-case with exhaustivity check).

I have checked the `canSend` property. For chats of type `Group` it seems to be equivalent to `!chat.isContactRequest` when `isSelfInGroup`.

https://github.com/chatmail/core/blob/df665d9ba14b7afc13ad0dddd623bd9759901328/src/chat.rs#L1439-L1496

Maybe it's fine to keep the function as is because it's also duplicated on other clients, so hopefully we'll get notified when other clients change their implementations.

packages/e2e-tests/tests/chat-context-menu.spec.ts
```diff
@@ -148,21 +148,100 @@ test.describe('Main View - 3-Dot Menu', () => {
...
+      .fill(disposableGroupName)
+    await page.getByTestId('group-create-button').click()
+
+    // Verify the group is selected and Leave Group is shown
+    await openMainViewThreeDotMenu(page)
+    await expect(
+      page.getByRole('menuitem', { name: 'Leave Group' })
+    ).toBeVisible()
+    await expect(
+      page.getByRole('menuitem', { name: 'Delete Chat' })
```
WofWca:
```suggestion
      page.getByRole('menuitem', { name: 'Delete' })
```

So that it's safer from potential future rewordings.

packages/e2e-tests/tests/chat-context-menu.spec.ts
```diff
@@ -148,21 +148,100 @@ test.describe('Main View - 3-Dot Menu', () => {
...
+      .getByRole('button', { name: 'Leave Group' })
+      .click()
+
+    // After leaving, re-open menu: Delete Chat should replace Leave Group
+    await openMainViewThreeDotMenu(page)
+    await expect(
+      page.getByRole('menuitem', { name: 'Delete Chat' })
+    ).toBeVisible()
+    await expect(
+      page.getByRole('menuitem', { name: 'Leave Group' })
```
WofWca:
```suggestion
      page.getByRole('menuitem', { name: 'Leave' })
```

packages/e2e-tests/tests/chat-context-menu.spec.ts
```diff
@@ -311,7 +390,7 @@ test.describe('Chat List Context Menu - Single Selection', () => {
 
     await openChatListContextMenu(page, groupName)
 
-    // Chat list menu items for group chat
+    // Chat list menu items for group chat (no Leave Group in chat list)
```
WofWca:
```suggestion
    // Chat list menu items for group chat
```

e6088e5a34e4844f7518b419169520b332e4adbd made this comment no longer true.
But see https://github.com/deltachat/deltachat-desktop/pull/6153/changes#r3058153064 as an alternative solution.

WofWca: Although now the following point from the issue description is no longer obeyed.

> in the chatlist, where multi-selection is allowed, rename "Delete Chat" to a simple "Delete" (this fixes plural issue, when on that, do the same for "Pin" and "Archive"). no need to have "Leave" as an option there

packages/frontend/src/components/chat/ChatContextMenu.tsx
nicodh: This is a bit nitpick imo. shouldLeaveBeforeDelete is the reason why we showLeaveInsteadOfDelete and leave is only shown if the user can leave the group, otherwise (if he already left) it shows delete so the comment seems correct for me. The only case where we only show leave-and-delete is if multiple groups are selected since we need something to be applied for all groups.

nicodh: The only thing that could be improved: if all chats of a selection "canLeave" then it makes sense also to show leave instead delete and then the comment needs to be changed. But I don't think it's worth the effort...

packages/frontend/src/components/chat/ChatContextMenu.tsx
nicodh: See below

WofWca (approved):
Works, at least in basic cases. Covered by existing tests, and new ones have been nicely added.
As I said, I think it'd be better to have this logic in Core. But iOS and Android do it this way also.

Again, note that this doesn't address the 'rename "Delete Chat" to a simple "Delete"' point from the issue. Personally I like it the way it is in this MR right now.
Pin / Archive / Delete is also good.

The irrelevant changes, such as the rename of the `relatedChat` parameter of `buildEncryptionInfoMenuItem`, also seem fine.

r10s (approved):

PR #6156, opened 2026-03-25
feat: rework message info

### LINKED ISSUE #5994: get "message info" out of nerd mode
we want to get the message info dialog out of the nerdy debug view, doing things similar to whatsapp & co here, this means mainly showing the timestamps the message is processed. the layout should be the following:

```
**Message Info**                            ⋮  X

Sent                            2026-01-01 14:12

Received                        2026-01-01 14:30

✓✓ **Read By**

(A) Alice                       2026-01-01 15:12

(B) Bob                         2026-01-01 16:12

(C) Claire                      2026-01-01 17:12
```

this means for the current diialog:

- [ ] remove the debug field, to a menu entry in the three dot menu called "Developer"
- [ ] remove "Edited" flag (this is redundant and random, we also do not repeat other means, not need to be exhaustive here)
- [ ] remove messageID (if really needed, it should be added by core to the debug info)
- [ ] change order
- [ ] date should all be formatted the same way (currently it is not)
- [ ] remove grey background for the read receipt rows, all should look the same. the avatar rows should look like the reactions
- [ ] the lines with the read receipts should be tappable (as for reactions)
- [ ] use the same standard font sizes for all text but "Message Info" headline (which is the same font as elsewhere)
- [ ] "Read By" should be bold

the text itself

this issue is the outcome of several one to one discussions. showing or discussions about additional information as read receipts, edited state etc. is probably out of scope. we do not want to have the dialog clutters much, put provide otherwise missing means on point

once this is settled, file issue for android and iOS

nicodh:
Resolves #5994

Maybe better to review commit by commt since I refactored the MessageDetails from class to function first before introducing the new features.

Concerning the MessageID - should we show it nowhere at all or only in the developer view?

image

image

### DIFF
--- a/packages/frontend/src/components/dialogs/MessageDetail/styles.module.scss
+++ b/packages/frontend/src/components/dialogs/MessageDetail/styles.module.scss
@@ -1,5 +1,5 @@
 .dialogHeader {
-  padding-bottom: 0;
+  padding-bottom: 10px;
 }
 
 .doubleCheckmarkIcon {
@@ -14,7 +14,7 @@
 .formattedMessageInfo {
   margin: 10px;
   font-size: 0.95em;
-  padding: 5px 20px;
+  padding: 5px 0px;
 }
 
 .infoRow {
@@ -22,7 +22,7 @@
   justify-content: space-between;
   align-items: center;
   margin-bottom: 10px;
-  padding: 2px 0;
+  padding: 2px 20px;
 }
 
 .infoLabel {
@@ -39,16 +39,14 @@
   display: flex;
   align-items: center;
   gap: 5px;
-  margin-top: 10px;
-  margin-bottom: 5px;
+  margin-top: 20px;
 }
 
 .readReceiptFormatted {
   display: flex;
   justify-content: space-between;
   align-items: center;
-  margin-bottom: 8px;
-  width: 100%;
+  width: calc(100% - 20px);
   background: none;
   border: none;
   padding: 10px;
@@ -57,7 +55,7 @@
   font: inherit;
   color: inherit;
   text-align: start;
-
+  margin-left: 10px;
   &:hover {
     background-color: var(--chatListItemBgHover);
   }
--- a/packages/frontend/src/components/dialogs/MessageDetail/BasicMessageInfo.tsx
+++ b/packages/frontend/src/components/dialogs/MessageDetail/BasicMessageInfo.tsx
@@ -49,14 +49,14 @@ export function BasicMessageInfo(props: BasicMessageInfoProps) {
 
   // Format sent timestamp
   const sentTimestamp = moment(message.timestamp * 1000)
-  const sentFormatted = sentTimestamp.format('L LT')
+  const sentFormatted = sentTimestamp.format('lll')
 
   // Format received timestamp
   const receivedTimestamp = message.receivedTimestamp
     ? moment(message.receivedTimestamp * 1000)
     : null
   const receivedFormatted = receivedTimestamp
-    ? receivedTimestamp.format('L LT')
+    ? receivedTimestamp.format('lll')
     : null
 
   return (
@@ -120,7 +120,7 @@ function ReadReceiptFormatted(props: {
   }, [props.receipt.contactId, accountId])
 
   const time = moment(props.receipt.timestamp * 1000)
-  const timeFormatted = time.format('L LT')
+  const timeFormatted = time.format('lll')
 
   if (!contact) {
     return null
--- a/packages/frontend/themes/_themebase.scss
+++ b/packages/frontend/themes/_themebase.scss
@@ -232,10 +232,6 @@ $scrollbarTransparency: 0.34 !default;
   --htmlEmail-color: #{$textPrimary};
   --htmlEmail-border-color: #{changeContrast($bgPrimary, 20%)};
 
-  /** MessageDetails Colors **/
-  --MessageDetails-ReadReceipt-Bg: #{changeContrast($bgPrimary, 10%)};
-  --MessageDetails-ReadReceipt-Date: #{changeContrast($textPrimary, 30%)};
-
   /** generic reusable colors */
   --summary-text-color: #{changeContrast($textPrimary, 33%)};
   --textPrimary: #{$textPrimary};


packages/frontend/src/components/dialogs/MessageDetail/styles.module.scss
```diff
@@ -23,24 +23,87 @@
...
+
+.readReceiptFormatted {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  margin-bottom: 8px;
+  width: 100%;
+  background: none;
+  border: none;
+  padding: 10px;
```
nicodh, line 52: tbd: the padding is meant to show a nicer hover button on read receipt rows but it moves the read date a bit to the left which breaks the date alignments from above.

WofWca:
> Concerning the MessageID - should we show it nowhere at all or only in the developer view?

The issue says

>  remove messageID (if really needed, it should be added by core to the debug info)

nicodh:
> > Concerning the MessageID - should we show it nowhere at all or only in the developer view?
> 
> The issue says
> 
> > remove messageID (if really needed, it should be added by core to the debug info)

I know but would like to understand what's the background since we are in technical details level in the second dialog

packages/frontend/src/components/dialogs/MessageDetail/ReadReceipts.tsx
```diff
@@ -9,26 +9,154 @@ import moment from 'moment'
...
+            ))}
+        </>
+      )}
+    </div>
+  )
+}
+
+function ReadReceiptFormatted(props: {
+  receipt: T.MessageReadReceipt
+  index: number
```
WofWca:
```suggestion
```

Inused.

packages/frontend/src/components/dialogs/MessageDetail/MessageDetail.tsx
```diff
@@ -1,113 +1,80 @@
...
-              ) : null}
-              {this.props.messageId && (
-                <tr>
-                  <td className='module-message-detail__label'>messageId</td>
-                  <td>{this.props.messageId}</td>
-                </tr>
-              )}
-              {isEdited && (
-                <tr>
-                  <td className='module-message-detail__label'>
```
WofWca, line 99: Styles for these classes also need to be removed.

packages/frontend/src/components/dialogs/MessageDetail/ReadReceipts.tsx
```diff
@@ -9,26 +9,154 @@ import moment from 'moment'
...
+type FormattedMessageInfoProps = {
+  message: T.Message
+  info: string
+  messageId: number
+  onClose: () => void
+}
+
+export function FormattedMessageInfo(props: FormattedMessageInfoProps) {
+  const { message } = props
+  const _tx = useTranslationFunction()
```
WofWca:
```suggestion
  const tx = useTranslationFunction()
```

Why is it named like this?

packages/frontend/src/components/dialogs/MessageDetail/ReadReceipts.tsx
```diff
@@ -9,26 +9,154 @@ import moment from 'moment'
...
+type ReadReceiptsListProps = { messageId: number }
+
+type FormattedMessageInfoProps = {
+  message: T.Message
+  info: string
+  messageId: number
+  onClose: () => void
+}
+
+export function FormattedMessageInfo(props: FormattedMessageInfoProps) {
```
WofWca:
```suggestion
export function BasicMessageInfo(props: BasicMessageInfoProps) {
```

WofWca: I don't think it's appropriate to have this component in a file named `ReadReceipts`?

packages/frontend/src/components/dialogs/MessageDetail/ReadReceipts.tsx
```diff
@@ -9,26 +9,154 @@ import moment from 'moment'
...
+          aria-hidden={true}
+        />
+        <span className={styles.contactName}>{contact.displayName}</span>
+      </div>
+      <span className={styles.infoValue}>{timeFormatted}</span>
+    </button>
   )
+}
+
+export function ReadReceiptsList(props: ReadReceiptsListProps) {
```
WofWca: This component is unused.

packages/frontend/src/components/dialogs/MessageDetail/MessageDetail.tsx
```diff
@@ -1,113 +1,80 @@
...
+    const handleKeyDown = (ev: KeyboardEvent) => {
+      if (ev.code === 'Escape') {
+        ev.preventDefault()
+        ev.stopPropagation()
+        onCloseTechnicalDetails()
+      }
+    }
+    document.addEventListener('keydown', handleKeyDown)
+    return () => document.removeEventListener('keydown', handleKeyDown)
+  }, [showTechnicalDetails, onCloseTechnicalDetails])
```
WofWca: What's the idea behind this complex dialog handling? Why not just use the usual `openDialog()` for the "developer" dialog?

packages/frontend/src/components/dialogs/MessageDetail/MessageDetail.tsx
nicodh: Good finding!

packages/frontend/src/components/dialogs/MessageDetail/ReadReceipts.tsx
nicodh: I agree and renamed it

packages/frontend/src/components/dialogs/MessageDetail/ReadReceipts.tsx
nicodh: Removed it in a cleanup commit

packages/frontend/src/components/dialogs/MessageDetail/ReadReceipts.tsx
nicodh: Renamed it

packages/frontend/src/components/dialogs/MessageDetail/MessageDetail.tsx
nicodh: Added a separate dialog

packages/frontend/src/components/dialogs/MessageDetail/styles.module.scss
```diff
@@ -23,24 +27,87 @@
...
 }
 
-.ReadReceipt {
+.readReceipt {
   display: flex;
   align-items: center;
   margin-block: 5.8px;
 }
 
-.ReadReceiptContactLabel {
```
WofWca, line 32: Some of these styles are also unused now.

At this point since this is basically a complete rewrite it's easier to just remove all styles and only bring back the ones that are actually used.

nicodh: Ok, I  now checked all twice

r10s:
for the Message-ID: it is fine as-is. the debug-info should be printed as given from core; and the Message-ID is useful for debugging. sorry for being unclear in the original issue.

apart from that, we should tweak the following:

- add more space before/after the send/received area, maybe the double (green zigzag lines in the screenshot). the space between send/received is fine.

- date should all be formatted the same way as done in the bubbles - currently i have `Jun 30, 2025 2:48 PM` for absolute date in the bubble where message info shows `06/30/2025 2:48 PM`

- left column should be left-aligned correctly (left red line in the screenshot)

- dates should be right-aligned correctly as well (right red line in the screenshot)  

(probably sent/received should use the same padding as the contact cells, if the headlines have different paddings, that's probably okay)


Screenshot 2026-03-26 at 15 11 41

packages/frontend/src/components/dialogs/MessageDetail/styles.module.scss
```diff
@@ -1,46 +1,75 @@
...
   display: flex;
+  justify-content: space-between;
   align-items: center;
-  font-size: 1.15em;
+  margin-bottom: 10px;
+  padding: 2px 0;
 }
 
-.ReadReceiptBox {
-  background-color: var(--MessageDetails-ReadReceipt-Bg);
```
WofWca, line 12: These `--MessageDetails` variables are also unused now 😅

nicodh:
> * left column should be left-aligned correctly (left red line in the screenshot)
> * dates should be right-aligned correctly as well (right red line in the screenshot)

That was what I asked for in my first comment above.

> tbd: the padding is meant to show a nicer hover button on read receipt rows but it moves the read date a bit to the left which breaks the date alignments from above.

 It would result in a non padding button onHover.
image

instead


image

If that is wanted it's fine and I will align all left and right

packages/frontend/src/components/dialogs/MessageDetail/styles.module.scss
WofWca: This can be solved by removing the padding from the parent element (is it the dialog itself?)

WofWca: (And then applying the padding / margin to children individually)

WofWca: Or rather, to be precise, by applying this same amount of padding to the "sent" and "received" rows.

nicodh: Yes I also thought of that if both (alignement & button padding) is wanted

r10s:
can't we have the second and add the same padding around sent/received row? then also the additional space is there

PR #6162, opened 2026-03-25
refactor: remove unnecessary `processQr`

WofWca:
`const _test: false = !qr`, so the condition is never true.

The function has been introduced in
2f9e829954f3ef087cbbca7060d082cdb19aa6b9
(https://github.com/deltachat/deltachat-desktop/pull/3773).
It's not clear to me whether it has ever been useful.

The comments are also redundant IMO.
If they are needed, they should be added on `rpc.checkQr()`.

### DIFF
--- a/packages/frontend/src/backend/qr.ts
+++ b/packages/frontend/src/backend/qr.ts
@@ -1,5 +1,3 @@
-import { BackendRemote } from '../backend-com'
-
 import type { T } from '@deltachat/jsonrpc-client'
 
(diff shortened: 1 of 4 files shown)


packages/frontend/src/backend/qr.ts
```diff
@@ -16,27 +14,3 @@ export type QrWithUrl<Q = T.Qr> = {
   qr: Q
   url: string
 }
```
nicodh, line 16: SInce the file now only contains types it would make sense to rename it to qr-types.ts

PR #6167, opened 2026-03-28
Cleanup unused css

nicodh:
Found with a script and manually reviewed

### DIFF
--- a/packages/frontend/scss/_emoji.scss
+++ b/packages/frontend/scss/_emoji.scss
@@ -20,16 +20,6 @@
   font-size: 3em;
 }
 
-// we need these, or we'll make conversation items too big in the left-nav
-.conversations {
-  .emoji-container.small,
-  .emoji-container.medium,
-  .emoji-container.large,
-  .emoji-container.jumbo {
-    font-size: 1em;
-  }
-}
-
 // This is the emoji button in the composer
 button.emoji {
   width: 36px;
--- a/packages/frontend/scss/chat/_chat-list.scss
+++ b/packages/frontend/scss/chat/_chat-list.scss
@@ -23,25 +23,4 @@
     border-bottom: var(--cli-search-result-divider-border-width) solid
       var(--cli-search-result-divider-border);
   }
-
-  .search-in-chat-label {
-    display: flex;
-    height: 64px;
-    flex-direction: row;
-    padding: 0px 10px;
-    align-items: center;
-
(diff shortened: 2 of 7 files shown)


WofWca:
Do you mind sharing what script you used?

nicodh:
> Do you mind sharing what script you used?

No not at all:

```bash
#!/bin/bash
# Extracts CSS class names from SCSS files and checks if they're referenced in source code.
#
# Usage:
#   ./bin/find-unused-css-classes.sh           # show only unused classes
#   ./bin/find-unused-css-classes.sh --all     # show all classes with usage status

set -euo pipefail

SCSS_DIR="packages/frontend/scss"
SRC_DIRS="packages/frontend/src packages/target-electron/src packages/target-tauri/src packages/target-browser/src"
SHOW_ALL=false

if [[ "${1:-}" == "--all" ]]; then
  SHOW_ALL=true
fi

# Extract class names from SCSS files:
# - Match .class-name patterns (standard CSS class selectors)
# - Exclude lines that are comments
# - Deduplicate
classes=$(grep -rhoP '(?<![\w-])\.(?![\d])([a-zA-Z][\w-]*)' "$SCSS_DIR" \
  --include='*.scss' \
  | grep -v '//' \
  | sed 's/^\.//' \
  | sort -u)

unused_count=0
used_count=0
total=0

for class in $classes; do
  total=$((total + 1))

  # Search in TSX/JSX/TS/JS source files and also in SCSS (for @extend etc.)
  found=false
  for dir in $SRC_DIRS; do
    if [[ -d "$dir" ]] && grep -rql --include='*.tsx' --include='*.ts' --include='*.jsx' --include='*.js' --include='*.scss' "$class" "$dir" 2>/dev/null; then
      found=true
      break
    fi
  done

  if $found; then
    used_count=$((used_count + 1))
    if $SHOW_ALL; then
      echo "  USED: $class"
    fi
  else
    unused_count=$((unused_count + 1))
    # Find where it's defined
    locations=$(grep -rn "\.$class\b" "$SCSS_DIR" --include='*.scss' | head -3)
    echo "UNUSED: $class"
    echo "$locations" | sed 's/^/         /'
    echo ""
  fi
done

echo "---"
echo "Total classes: $total"
echo "Used: $used_count"
echo "Possibly unused: $unused_count"`

It's an AI generated script, so maybe there are better ones but it found the unused classes

PR #6170, opened 2026-03-29
fix: translate static aria label

### LINKED ISSUE #4611: Aria-Label missing for various buttons
- [ ] jump down button in message list (both modes should have different strings (jump back, jump/scroll down completely), translation strings also need to be create)
- [ ] verified chat/contact icon has no aria label
- [ ] accounts in account list should have account name & label in aria label -> maybe also connection state and if account is muted

nicodh:
Partly resolves #4611

- aria-description for Account Items

### DIFF
--- a/packages/frontend/src/components/AccountListSidebar/AccountItem.tsx
+++ b/packages/frontend/src/components/AccountListSidebar/AccountItem.tsx
@@ -269,6 +269,11 @@ export default function AccountItem({
         role='tab'
         aria-selected={isSelected}
         aria-busy={!account && accountFetch.loading}
+        aria-label={
+          account?.kind === 'Configured'
+            ? (account.displayName || undefined)
+            : undefined
+        }
         onClick={() => onSelectAccount(accountId)}
         onContextMenu={onContextMenu}
         aria-haspopup='menu'
@@ -316,7 +321,7 @@ export default function AccountItem({
         )}
         {muted && (
           <div
-            aria-label='Account notifications muted'
+            aria-label={tx('muted')}
             className={styles.accountMutedIconShadow}
           >
             <Icon className={styles.accountMutedIcon} icon='audio-muted' />
--- a/packages/frontend/src/components/AccountListSidebar/AccountItem.tsx
+++ b/packages/frontend/src/components/AccountListSidebar/AccountItem.tsx
@@ -271,8 +271,26 @@ export default function AccountItem({
         aria-busy={!account && accountFetch.loading}
         aria-label={
           account?.kind === 'Configured'
-            ? (account.displayName || undefined)
+            ? account.displayName || account.addr || undefined
             : undefined
+        }
+        aria-description={
+          [
+            account?.kind === 'Configured' && account.displayName
+              ? account.addr
+              : undefined,
+            unreadCount
+              ? tx('chat_n_new_messages', String(unreadCount), {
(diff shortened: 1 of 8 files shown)


nicodh:
For the verification icon: we have no translation for "verified" but it is also not clear if that is the wording we want

packages/frontend/src/components/Avatar/index.tsx
```diff
@@ -60,7 +60,11 @@ export function Avatar(props: {
   } = props
 
   const content = avatarPath ? (
-    <img className='content' src={runtime.transformBlobURL(avatarPath)} />
+    <img
+      className='content'
+      src={runtime.transformBlobURL(avatarPath)}
+      alt={displayName}
```
WofWca, line 66: Have you checked in which places this is this beneficial? That would only be places where the display name is not displayed otherwise. Maybe we should make the `aria-hidden` prop required and not optional to see this more clearly.

Edit: see `alt` guidelines: https://www.w3.org/WAI/tutorials/images/decision-tree/

packages/frontend/src/components/AccountListSidebar/AccountItem.tsx
```diff
@@ -269,6 +269,29 @@ export default function AccountItem({
...
+                })
+              : undefined,
+            muted ? tx('muted') : undefined,
+            bgSyncDisabled && !isSelected
+              ? tx('background_sync_disabled_explaination')
+              : undefined,
+          ]
+            .filter(Boolean)
+            .join(', ') || undefined
+        }
```
WofWca, line 294: So what this change does is it adds `aria-label` with account display name, and "moves" all the other content to `aria-description`.

This does IMO improve accessibility, but I think that this is not the right approach. I described my preferred one in the issue. We must properly use the hover tooltip.

Additionally, even if we don't use the tooltip, it would be simpler to separate the contents of the button into an element beside it, and use `aria-describedby`, instead of duplicating all the logic, manually defining overrides.
The button itself can then have `position: absolute; inset: 0;` to preserve the original look.

WofWca (commented):
The MR seems to only apply to a part of the changes.

Regarding the profile: this is a button with lots of content, and it probably should have been a button placed _beside_ the info that it wraps (and perhaps with `position: absolute; inset: 0;`), but I think it's OK to apply `aria-label` to it (basically overriding the contents for most screen readers), because all that content is available elsewhere, i.e. in the dialog that opens when you click the button.

packages/frontend/src/components/Avatar/index.tsx
WofWca: But actually this probably is a topic for another MR.

WofWca (approved):

packages/frontend/src/components/AccountListSidebar/AccountItem.tsx
nicodh: If we reuse tooltip for aria-describedby we would have to change the way the tooltip is built, since right now it's only filled with text if one hovers over the account button what would not be appropriate for screen reader.

So I would merge this as it is and improve stuff in a followup

WofWca: Yes, the tooltip stuff needs a big refactor.

But the "separate element" approach from my last paragraph would still work.

I think it's OK to merge as is.

PR #6182, opened 2026-04-03
feat: "Mark as Unread" menu item (and "as Read")

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/6181.

### DIFF
--- a/packages/e2e-tests/tests/chat-context-menu.spec.ts
+++ b/packages/e2e-tests/tests/chat-context-menu.spec.ts
@@ -372,6 +372,7 @@ test.describe('Chat List Context Menu - Single Selection', () => {
     // Chat list menu items for DM chat
     await expect(getMenuItems(page)).toHaveText([
       'Pin Chat',
+      'Unread',
       'Mute Notifications',
       'Archive Chat',
       'View Profile',
@@ -393,6 +394,7 @@ test.describe('Chat List Context Menu - Single Selection', () => {
     // Chat list menu items for group chat
     await expect(getMenuItems(page)).toHaveText([
       'Pin Chat',
+      'Unread',
       'Mute Notifications',
       'Archive Chat',
       'View Profile',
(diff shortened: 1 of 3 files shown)


WofWca:
I'm working on fixing tests. Otherwise this is ready.

WofWca:
- [x] Maybe we should also be unselecting the chats when marking them as unread

_locales/_untranslated_en.json
```diff
@@ -112,5 +112,8 @@
   },
   "messages_are_e2ee": {
     "message": "Messages are end-to-end encrypted."
+  },
+  "mark_as_unread": {
```
WofWca: FYI this string is to be added on the next translations pull

nicodh (changes requested):
Is it intended that there are no conditions at all wether or not a chat can be marked as unread? Even empty chats and selfChat ?The latter does not even show a fresh message counter that would make the change visible.

btw. Mute notifications also doesn't make sense on selfChat but that's another issue

WofWca:
> Even empty chats

Good point, but why not. E.g. you might want to create a group and set a "reminder" for yourself to write a message there later.

> and selfChat ?The latter does not even show a fresh message counter

Ah I see. It only works if you have a "saved message" from someone else in your self chat.
It's a Core issue I would say.
In Telegram it is possible to mark "saved messages" as unread.

> Mute notifications also doesn't make sense on selfChat but that's another issue

A use case for muting "saved messages" would be to combine it with "mark as unread" and then mute for some time. You'll have a "one new message" badge when the mute ends.
FYI Telegram doesn't have "mute" on "Saved messages".

nicodh:
needs rebase

_locales/_untranslated_en.json
WofWca: Hm actually there is only the short "Unread".

nicodh (approved):
Reviewed by reading & testing

Issue #6177, opened 2026-04-01
profile invisible after update on macOS


r10s:
this was reported at https://mastodon.social/@pancake@infosec.exchange/116329299237001520

not sure if the same issue,
but i could reproduce the following:

- install via .dmg  from delta.chat
- open app, create "Profile A"
- install or update via macos app store
- "Profile A" is no longer visible - instead you're offered to create a new one

this can be fixed to reinstall the .dmg from delta.chat - then you suddenly have "Profile A" back

installing first from app store or from other sources (homebrew) may result in similar issues.

we have a similar situation on android, however, it is far better there as you have _two_ app icons then - on macOS you only have one and it looks really as if the profile is lost.

locations used for the `accounts` dir:
macOS app store:
`database_dir=/Users/bpetersen/Library/Containers/chat.delta.desktop.electron/Data/Library/Application Support/DeltaChat/accounts`
installation from delta.chat:
`database_dir=/Users/bpetersen/Library/Application Support/DeltaChat/accounts`

not sure, what we can reasonably do about that. maybe we can use the same dir? that would require an migration. or show a warning? including a hint how to fix it?

related: https://github.com/deltachat/deltachat-desktop/issues/6176

nicodh:
It's more or less the same behaviour like in Windows. Data dir is either sandboxed (when installed from Store) or not if installed with dmg file (the latter is only possible by ignoring warnings from the OS).

While it is not possible to detect from the sandboxed version if there is already data in  Library/Application Support/DeltaChat it is possible the other way round. So if someone tries to upgrade by running Setup.exe when having a previous version installed from Store we could show a warning if
1. there are no configured accounts yet
2. there is an account folder inside /Users/username/Library/Application Support/DeltaChat/

nicodh:
I would suggest to not make any migrations from one path to the other as it is a one-way-path. We can only migrate a sandboxed version to a non sandboxed, but not the other way round

r10s:
agree for not doing a migration. this tends to cause far larger bugs  - where this one is mostly visible to advanced user.

but showing a warning, where possible, makes sense

WofWca:
I think that the best solution is simply not to publish blog posts before we have actually released a new version on all distribution channels.

r10s:
you cannot aways avoid delays and sync things perfectly. and doing blog post after everything is released has issues on its own, we had that in the past, where we need to explain and so on.

[...]

PR #6186, opened 2026-04-03
Warn about existing accounts & migration for macos data

nicodh:
Related to #6177

shows a warning if

- a previous Apple Store install exists and the user starts a dmg install
- a previous MS Store install exists and the user starts a Setup/Portable exe
- a previous Setup.exe install exists and the user starts a MS Store install

The added MacOS container-migration.plist should transfer the data into the sandbox container if a first time MacOS Store install is started and a previous install exists

This is how it will look like on macOS 15:

Screenshot 2026-04-09 at 13 51 32

This is how it will look on a (german)  windows 11 OS (text outdated)

screenshot-dc-desktop-windows

### DIFF
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -117,6 +117,6 @@
     "message": "Warning"
   },
   "data_found_other_installation_message": {
-    "message": "Your profiles from another Delta Chat installation are not accessible - different installations store their data in separate locations.\n\nTo keep the existing profiles:\n1. Exit now\n2. Open the other Delta Chat - you may need to reinstall it from %1$s\n3. Create a backup there at \"Settings / Chats\"\n4. Relaunch this version\n5. Restore the backup on the next page at \"I already have a profile\"\n\nContinue without your previous profiles?"
+    "message": "Your profiles from another Delta Chat installation are not accessible - different installations store their data in separate locations.\n\nTo keep existing profiles:\n1. Quit now\n2. Open the other Delta Chat - if needed, reinstall it from %1$s\n3. Back up each profile at \"Settings/Chats\"\n4. Relaunch this version\n5. Restore the backups using \"I already have a profile\"\n\nContinue without your previous profiles?"
   }
 }
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -113,10 +113,10 @@
   "messages_are_e2ee": {
     "message": "Messages are end-to-end encrypted."
   },
-  "data_found_other_installation_title": {
-    "message": "Data found from a previous %1$s installation"
+  "warning": {
+    "message": "Warning"
   },
   "data_found_other_installation_message": {
-    "message": "A previous %1$s installation left account data that this version will not see.\nWe recommend to avoid running different versions of Delta Chat simultaneously.\nTo keep your existing profiles, you can create a backup in the previous app before launching the new version and import it again."
+    "message": "Your profiles from another Delta Chat installation are not accessible - different installations store their data in separate locations.\n\nTo keep the existing profiles:\n1. Exit now\n2. Open the other Delta Chat - you may need to reinstall it from %1$s\n3. Create a backup there at \"Settings / Chats\"\n4. Relaunch this version\n5. Restore the backup on the next page at \"I already have a profile\"\n\nContinue without your previous profiles?"
   }
 }
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -112,5 +112,11 @@
   },
   "messages_are_e2ee": {
     "message": "Messages are end-to-end encrypted."
+  },
+  "data_found_other_installation_title": {
+    "message": "Data found from a previous %1$s installation"
+  },
+  "data_found_other_installation_message": {
+    "message": "A previous %1$s installation left account data that this version will not see.\nWe recommend to avoid running different versions of Delta Chat simultaneously.\nTo keep your existing profiles, you can create a backup in the previous app before launching the new version and import it again."
   }
 }
--- a/packages/target-electron/build/container-migration.plist
+++ b/packages/target-electron/build/container-migration.plist
@@ -4,7 +4,7 @@
 <dict>
 	<key>Move</key>
 	<array>
-		<string>${Application Support}/DeltaChat</string>
+		<string>${ApplicationSupport}/DeltaChat</string>
 	</array>
 </dict>
 </plist>
--- a/packages/target-electron/build/container-migration.plist
+++ b/packages/target-electron/build/container-migration.plist
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>Move</key>
+	<array>
+		<string>${Application Support}/DeltaChat</string>
+	</array>
+</dict>
+</plist>
--- a/packages/target-electron/src/index.ts
+++ b/packages/target-electron/src/index.ts
@@ -221,7 +221,7 @@ async function onReady([
         type: 'warning',
         title: tx('warning'),
         message: tx('data_found_other_installation_message', otherStoreName),
-        buttons: [tx('perm_continue'), tx('global_menu_file_quit_desktop')],
+        buttons: [tx('ok'), tx('cancel')],
         defaultId: 0,
         cancelId: 1,
       })
--- a/packages/target-electron/src/index.ts
+++ b/packages/target-electron/src/index.ts
@@ -6,7 +6,7 @@ import { app as rawApp, dialog, ipcMain, protocol, clipboard } from 'electron'
 import { BrowserWindow } from 'electron/main'
 import rc from './rc.js'
 import contextMenu from './electron-context-menu.js'
-import { initIsWindowsStorePackageVar } from './isAppx.js'
+import { appx, initIsWindowsStorePackageVar } from './isAppx.js'
 import { getHelpMenu } from './help_menu.js'
 import { initialisePowerMonitor } from './resume_from_sleep.js'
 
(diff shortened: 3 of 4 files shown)


packages/target-electron/src/index.ts
```diff
@@ -183,6 +183,53 @@ async function onReady([_appReady, _loadedState, _appx, _webxdc_cleanup]: [
...
+      const appxPath = join(
+        homedir(),
+        'AppData/Local/Packages/merlinux.DeltaChat_v2ry5hvxhdhyy/LocalCache/Local/DeltaChat/accounts'
+      )
+      const otherPath = appx ? normalPath : appxPath
+      if (existsSync(otherPath)) {
+        otherStoreName = appx ? 'non-store' : 'Microsoft Store'
+        otherAccountsPath = otherPath
+      }
+    }
```
WofWca, line 218: Shouldn't we also check the `normalPath` when starting a portable version then?

packages/target-electron/src/index.ts
```diff
@@ -183,6 +183,53 @@ async function onReady([_appReady, _loadedState, _appx, _webxdc_cleanup]: [
   acceptThemeCLI()
   setLanguage(DesktopSettings.state.locale || app.getLocale().split('-')[0]) // can consist of 2 strings like in en-GB
 
+  // Warn users if data exists from a different installation variant
+  // (e.g. Mac App Store vs DMG, or Windows Store APPX vs Setup.exe),
+  // but only when there are no accounts yet in the current location.
+  if (!existsSync(getAccountsPath())) {
```
WofWca: Doesn't have to be blocking.

Also we don't have to wait for `onReady` to check the existence of the dir (but we do have to wait for the `appx` to get initialized).

packages/target-electron/src/index.ts
WofWca, line 218: The MR description says

> a previous Setup.exe install exists and the user starts a MS Store install

However note that installing the MS Store version while already having the .exe version would make both of the versions use the .exe path. See https://github.com/deltachat/deltachat-desktop/issues/3959#issuecomment-2391376364.
So it's rather about the other way around: starting the .exe version while already having a MS Store version.

_locales/_untranslated_en.json
```diff
@@ -112,5 +112,11 @@
   },
   "messages_are_e2ee": {
     "message": "Messages are end-to-end encrypted."
+  },
+  "data_found_other_installation_title": {
+    "message": "Data found from a previous %1$s installation"
+  },
+  "data_found_other_installation_message": {
+    "message": "A previous %1$s installation left account data that this version will not see.\nWe recommend to avoid running different versions of Delta Chat simultaneously.\nTo keep your existing profiles, you can create a backup in the previous app before launching the new version and import it again."
```
WofWca: > you can create a backup in the previous app

The user most likely has uninstalled the old installation. We should then recommend them to install the previous distribution, otherwise it's not clear how they should open the previous app.

Also maybe we should be more clear about how to make a backup, i.e. "go to settings and click this and that for each of your accounts".

WofWca (commented):
Looks alright. The `index.ts` code doesn't look like it can break stuff.
I can't say much about the new `.plist` though, but it seems to resemble what's being described in https://developer.apple.com/documentation/security/migrating-your-app-s-files-to-its-app-sandbox-container. As long as we're gonna test this change I think it's fine.

r10s:
> I can't say much about the new .plist though, but it seems to resemble what's being described in https://developer.apple.com/documentation/security/migrating-your-app-s-files-to-its-app-sandbox-container. As long as we're gonna test this change I think it's fine.
> As long as we're gonna test this change I think it's fine.

it looks simple enough, but well :)

no idea how to test that before we're actually in the store

packages/target-electron/build/container-migration.plist
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>Move</key>
+	<array>
+		<string>${Application Support}/DeltaChat</string>
```
r10s: is the space correct here?

https://developer.apple.com/documentation/security/migrating-your-app-s-files-to-its-app-sandbox-container says it is `${ApplicationSupport}`

_locales/_untranslated_en.json
r10s: to avoid clutter and to make translations, let's strike the title.

i also tried to make the instructions clearer and more actionable, avoiding some pitfalls as (eg. at least on macOS, the `Delta Chat.dmg` overwrite each other by default)

```suggestion
  "data_found_other_installation_message": {
    "message": "Your profiles from another Delta Chat installation are not accessible - different installations store their data in separate locations.\n\nTo keep the existing profiles:\n1. Exit now\n2. Open the other Delta Chat - you may need to reinstall it from %1$s\n3. Create a backup there at \"Settings / Chats\"\n4. Relaunch this version\n5. Restore the backup on the next page at \"I already have a profile\"\n\nContinue without your previous profiles?"
```

packages/target-electron/src/index.ts
```diff
@@ -183,6 +183,53 @@ async function onReady([_appReady, _loadedState, _appx, _webxdc_cleanup]: [
...
+    } else if (process.platform === 'win32') {
+      const { homedir } = await import('os')
+      const normalPath = join(homedir(), 'AppData/Local/DeltaChat/accounts')
+      const appxPath = join(
+        homedir(),
+        'AppData/Local/Packages/merlinux.DeltaChat_v2ry5hvxhdhyy/LocalCache/Local/DeltaChat/accounts'
+      )
+      const otherPath = appx ? normalPath : appxPath
+      if (existsSync(otherPath)) {
+        otherStoreName = appx ? 'non-store' : 'Microsoft Store'
```
r10s:
```suggestion
        otherStoreName = appx ? 'get.delta.chat : 'Microsoft Store'
```

non-store is weird, if `get.delta.chat` are not always matching, suggestions are welcome. the string is only used as the "original installation source" now, see other suggestion

packages/target-electron/src/index.ts
```diff
@@ -183,6 +183,53 @@ async function onReady([_appReady, _loadedState, _appx, _webxdc_cleanup]: [
...
+    let otherAccountsPath: string | undefined
+
+    if (process.platform === 'darwin' && !process.mas) {
+      const { homedir } = await import('os')
+      const sandboxPath = join(
+        homedir(),
+        'Library/Containers/chat.delta.desktop.electron/Data/Library/Application Support/DeltaChat/accounts'
+      )
+      if (existsSync(sandboxPath)) {
+        otherStoreName = 'Mac App Store'
```
r10s: the system name is macOS, not Mac. however,  this is no worthful information as the user knows they're on macos - and if not, it is even better to name the thing to what it is:

```suggestion
        otherStoreName = 'App Store'
```

packages/target-electron/src/index.ts
nicodh: The warning is only displayed if there are no accounts. If the accounts from previous install are migrated/used the launch will run as expected.
But a comment should inform about that behaviour

nicodh:
> no idea how to test that before we're actually in the store

I could build a package build from this branch and upload as a testflight release - but  that uses another path probably anyway...

_locales/_untranslated_en.json
nicodh: I thinkl there should be a dialog title so I just used "Warning" now

r10s: i made a suggestion below, lets discuss there. multi-profile i would leave out, it adds noise to the many. if ppl do not get that implicitly, once ppl get how to transfer one, they will manage the others

packages/target-electron/src/index.ts
nicodh: I would sayn no, since someone who runs Deltachat as portable (which is intended to be run from an USB stick or similar) does that probably to avoid having data on the "normal" paths

WofWca: Then we also shouldn't be checking the `appxPath`? But we do.

packages/target-electron/src/index.ts
WofWca: I made a follow-up https://github.com/deltachat/deltachat-desktop/pull/6208

packages/target-electron/src/index.ts
nicodh: I consider the portable version as a special case. I.e. if you have a portable version and you plug it into a machine you would expect to run only that version and only with the profiles that are on the stick.

So maybe we should skip the whole dialog also if process.env.PORTABLE_EXECUTABLE_DIR exists?

_locales/_untranslated_en.json
```diff
@@ -112,5 +112,11 @@
   },
   "messages_are_e2ee": {
     "message": "Messages are end-to-end encrypted."
+  },
+  "warning": {
+    "message": "Warning"
+  },
+  "data_found_other_installation_message": {
+    "message": "Your profiles from another Delta Chat installation are not accessible - different installations store their data in separate locations.\n\nTo keep the existing profiles:\n1. Exit now\n2. Open the other Delta Chat - you may need to reinstall it from %1$s\n3. Create a backup there at \"Settings / Chats\"\n4. Relaunch this version\n5. Restore the backup on the next page at \"I already have a profile\"\n\nContinue without your previous profiles?"
```
nicodh: Why is there a question mark at the end? "Continue without your previous profiles?"

packages/target-electron/src/index.ts
WofWca: We only show the warning if the accounts folder is empty. I don't think that we need to forcefully disable this check for the portable version.

_locales/_untranslated_en.json
nicodh: You mean as question to be answered with "Ok"?

_locales/_untranslated_en.json
r10s: the title is ignored entirely on macOS, btw.

so also from that point of view, good to not put information there.

at many other places, we're note setting title (eg. `load_remote_content_ask` does not set a title) - so, might be electron uses the app name in this case. if so, we can still remove it alltogether, to avoid copy+paste issues in the future (when dev may think, title is shown and put meaningful things there)

but this is no blocker

_locales/_untranslated_en.json
r10s: the following makes the text a little shorter, while removing the plural/singular confusion and giving the user and idea how to fix the situation when having several backups.

the "on the next page" is removed, it was a bit weird for an app anyways, and explaining in detail where to find the button when restoring multiple profiles is too complex; ppl manage to do that anyways before, so this is only about giving some pointers, not a full walk-through

```suggestion
    "message": "Your profiles from another Delta Chat installation are not accessible - different installations store their data in separate locations.\n\nTo keep existing profiles:\n1. Quit now\n2. Open the other Delta Chat - if needed, reinstall it from %1$s\n3. Back up each profile at \"Settings/Chats\"\n4. Relaunch this version\n5. Restore the backups using \"I already have a profile\"\n\nContinue without your previous profiles?"
```

see initial post for a recent screenshot

packages/target-electron/src/index.ts
```diff
@@ -173,16 +173,65 @@ Also make sure you are not trying to run multiple instances of deltachat.`
...
+        otherAccountsPath = otherPath
+      }
+    }
+
+    if (otherStoreName && otherAccountsPath) {
+      const result = await dialog.showMessageBox({
+        type: 'warning',
+        title: tx('warning'),
+        message: tx('data_found_other_installation_message', otherStoreName),
+        buttons: [tx('ok'), tx('cancel')],
```
r10s: the buttons should better read quit / continue:

```suggestion
        buttons: [tx('perm_continue'), tx('global_menu_file_quit_desktop')],
```

(as macOS is strict on "quit" where others use both, "exit" and "quit", maybe even "leave", we stay with "quit" - also as we have a string for that :)

see initial post for a recent screenshot

r10s:
> Why is there a question mark at the end? "Continue without your previous profiles?"

because you can also "quit". so, there is a choice, even tho, not a choice ppl would like or expect.

to the question itself: if you "continue", you will not have your profiles, unless you have your backup ready as explained some lines above. so, it is a "Continue without your previous profile and, if you want to, import them afterwards?" but that is a bit long.

main purpose of the, maybe somehow redundant, question is, to show the impact. to make it really clear, you will start without profiles, so ppl do not see that as a bug

packages/target-electron/src/index.ts
```diff
@@ -1,500 +1,549 @@
...
+    })
+    contents.setWindowOpenHandler(_details => {
+      // prevent new windows from being created when clicking on links
+      return { action: 'deny' }
+    })
+  }
+
+  // Prevent webview tags from being created,
+  // if you need them make sure to read https://www.electronjs.org/docs/latest/tutorial/security#12-verify-webview-options-before-creation
+  // to not indroduce security risks
```
WofWca: Line breaks got messed up?

nicodh: Seems to have happened when accepting changes directly in github UI - fixed it already

nicodh:
I tested the cases on Windows & MacOS that a non store install shows the correct warning if a store install exists.

packages/target-electron/src/index.ts
WofWca: Doesn't seem so

image

nicodh: Strange, reformatted it locally. Now I checked the previous version out and reapplied the last changes

WofWca (approved):
I think we can say that this addresses the issue.

PR #6187, opened 2026-04-03
fix: don't scroll to active chat on order change

### LINKED ISSUE #6087: Unpinning the chat scrolls the chatlist to the selected chat
Version 2.43, Linux, Wayland

1. Pin some chats to the top
2. Select some chat a few pages below
3. Scroll up to pinned chats
4. Without selecting the chat, right-click on it, and click "unpin".
5. The chat gets unpinned, but the chatlist scrolls a few pages down to the selected chat. Now you want to unpin another chat, you need to scroll all the way back again.

WofWca:
- **refactor: `useEffectEvent` for scroll chat in view**
- **fix: don't scroll to active chat on order change**

Closes https://github.com/deltachat/deltachat-desktop/issues/6087.

Some more cases where we'd scroll:
1. Receive a message in the active chat
    such that it order in the chat list moves up.
2. Archive a chat that is above the current chat.

Instead let's only scroll on load
(or on switch between archived or normal)
and when switching between chats.

I tested those use cases, it works fine in them.

FTR the logic that I'm changing has been introduced in
https://github.com/deltachat/deltachat-desktop/pull/1659.

### DIFF
--- a/packages/frontend/src/components/chat/ChatList.tsx
+++ b/packages/frontend/src/components/chat/ChatList.tsx
@@ -250,6 +250,17 @@ export default function ChatList(props: {
     () => (activeChatId != null ? chatListIds.indexOf(activeChatId) : -1),
     [chatListIds, activeChatId]
   )
+  /**
+   * Can be `false` when `chatListIds` is still loading (it's empty then)
+   * or if we're switching between the "archived" and normal view.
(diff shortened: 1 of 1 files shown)


nicodh (approved):
Reviewed by reading & testing

PR #6190, opened 2026-04-03
fix: don't `marknoticedChat` on chat select & jump

WofWca:
This addresses
https://github.com/chatmail/core/issues/7965#issuecomment-4123439128.

Note that as of Core 2.48.0 this would re-introduce this bug
https://github.com/deltachat/deltachat-desktop/issues/3072.
Waiting for Core to do its thing so that this can be merged.


WofWca:
Not much has happened in Core on this front. Let's close for now. We're probably gonna re-state what needs to be done anyway when things move forward.

PR #6191, opened 2026-04-03
refactor: rename markChatAsSeen -> marknoticedChat

WofWca:
Core still makes a distinction between "seen" and "noticed"
so let's not be confusing and call conform
to the same naming convention.

TODO:
- [x] Merge the base into main


nicodh (approved):
Reviewed by reading
Just a renaming

PR #6193, opened 2026-04-03
feat: sticker groups, search, and inline suggestions

verncat:
Hello! 

I implement stickers functionality: https://support.delta.chat/t/stickers-groups-stickers-alias-an-emoji-for-quick-search-and-stickers-sync/5009

Organizes stickers into groups by subfolder name with a pack navigation bar, adds a search bar to the sticker picker (filters by alias/pack name/filename), and introduces inline sticker suggestions triggered by typing `:query` in the compose box.

This feature would make stickers in Delta Chat much more organized and convenient.

What do you think? Feedback and help are appreciated.
I think need some polish for styles...

### Changes in file structure
I use Signal-like sticker file structure.
Aliases are extracted from the filename convention `{index}.{text_name}+{emoji}.ext`.
```bash
$ tree $APP/accounts/$ACCOUNT/stickers/
stickers/
├── group1/
│   ├── 0.happy+😊.webp
│   ├── 1.sad_face+😢.webp
│   └── 2.heart+❤️.webp
└── MyAwesomeGroup/
    ├── 0.wave+👋.webp
    └── 1.thumbs_up+👍.webp
```

Old way also supported!

### Screenshots
image
<br>
image

### DIFF
--- a/packages/frontend/src/components/composer/EmojiAndStickerPicker.tsx
+++ b/packages/frontend/src/components/composer/EmojiAndStickerPicker.tsx
@@ -4,6 +4,8 @@ import React, {
   forwardRef,
   PropsWithChildren,
   useRef,
+  useMemo,
+  useCallback,
 } from 'react'
 import classNames from 'classnames'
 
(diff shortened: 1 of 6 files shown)


packages/frontend/src/components/composer/EmojiAndStickerPicker.tsx
```diff
@@ -92,12 +149,137 @@ function StickersListItem(props: { filePath: string; onClick: () => void }) {
...
+      </div>
+      <button
+        type='button'
+        className='sticker-folder-button'
+        onClick={onOpenStickerFolder}
+        title={tx('open_sticker_folder')}
+      >
+        <svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 20 20'>
+          <path d='M2 4a2 2 0 0 1 2-2h3.17a2 2 0 0 1 1.41.59l1.42 1.41h6a2 2 0 0 1 2 2v2h-2V6H9.83l-1.42-1.41-.29-.3-.29.3L6.41 6H4v10h7v2H4a2 2 0 0 1-2-2V4zm15 6v3h3v2h-3v3h-2v-3h-3v-2h3v-3h2z' />
+        </svg>
```
WofWca: What license is this icon under?

WofWca (commented):
Thanks!
Did you write this yourself or with AI?

packages/frontend/src/components/composer/EmojiAndStickerPicker.tsx
verncat: These icons were generated by AI. My mistake. I will replace all the icons with others.

r10s:
fist of all, thanks for thinking about stickers!

there are recent changes in the sticker area on android as well, after some internal discussions, it is clear that we want to get the desktop stickers out of the "nerd mode" as well, mainly meaning, that **the user should not think about internal folder structure** and thelike at all.

this means for desktop:
- remove the "do smth with some folders" hints and the option to open the folder
- when tapping a sticker, ask if it should be added to the sticker collection (similar to the context menu just now)
- add option to delete the sticker

after that, idea is on android as well as on desktop to have the option of stickers being grouped.

for this PR it means, that we do not want to have more dependencies to the folders, eg. do "add folder button" that opens the file explorer with some super-weird path, where ppl can mess up things.

_might_ be that the underlying structure is still folders, and nerds can do smth there, but this should not be exposed nor documented or we want to push ppl to figure around with that.

so, just now, this PR seems not to be mergable,  before adapting, we should be more clear, what to do with stickers.

i've created an issue for the points above, https://github.com/deltachat/deltachat-desktop/issues/6194, it was not done yet, as we were busy with other things :)

cc @adbenitez

verncat:
Okie, thank you for such a comprehensive answer about stickers ::)
I think yes, this is the best way for user-friendly

verncat:
I actively use DeltaChat in my daily life, and I'd really like to have a convenient way to interact with stickers.
Like, for example, Telegram.
I really enjoy using stickers, and I constantly use shortcuts like ":hi"

nicodh:
Closing since we won't go with the suggested folder structure but look for a more user friendly approach as mentioned above

PR #6198, opened 2026-04-06
fix: initialize `ConnectivityToast` immediately

WofWca:
Instead of only initializing when we get a `ConnectivityChanged`
event.

Otherwise when not connected we wouldn't show
the "Not Connected" toast e.g. after switching the account.

Related:
- https://github.com/deltachat/deltachat-desktop/issues/6196.


nicodh (approved):
Reviewed by reading

nicodh:
Is there a reason why you don't merge this?

WofWca:
Because of the upcoming release. Don't want to risk breaking stuff.

PR #6200, opened 2026-04-06
fix: show active chat as selected after archive

WofWca:
Reproduction steps:
1. Click on a chat (to make it active).
   We'll display it as selected in the chat list.
2. Go to "Archived Chats".
3. Go back to normal chat list view.

You'll see that the active chat is not displayed as selected,
i.e. there are 0 selected chats.

The change in `ChatList` makes sure that
when clicking on "Archived Chats" we don't unselect all chats.

The change in `useMultiselect.ts` makes sure that
clicking on a chat a second time doesn't change selection.
That is: the selection becomes "invalid"
when the active chat changes,
but clicking on the newly active chat again
would "validate" the selection (IOW we'd enter the multiselect mode)
so when the newly selected chat
gets removed from the list then the selection becomes empty.

### DIFF
--- a/packages/e2e-tests/tests/chat-list-multiselect.spec.ts
+++ b/packages/e2e-tests/tests/chat-list-multiselect.spec.ts
@@ -380,6 +380,38 @@ test('resets selection when the active chat changes', async () => {
   await expectSelectedChats([6])
 })
 
+test.describe("doesn't unselect active chat after switching to archive", () => {
+  test.beforeAll(async () => {
+    // Archive a chat
+    await getChat(7).click()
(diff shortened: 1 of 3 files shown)


nicodh (approved):
Reviewed by reading & testing

Note that a selection of multiple chats will get lost when switching to archived and back which is fine and probably expected

WofWca:
Yes, we have code that specifically does this:

https://github.com/deltachat/deltachat-desktop/blob/4b28a6ebb845051dfc6a549d2bf63e5ccd7d10c1/packages/frontend/src/components/chat/ChatList.tsx#L983-L1026

Issue #5935, opened 2026-01-13
fork `emoji-mart`, as the project was not updated in a long time.


Simon-Laux:
We need to start using our own fork as the original maintainers seem to have abandoned it:
- no updates to the emojis-standard, which had new versions meanwhile adding new emojis and removed the detailed family emojis.
- my bug fix PR was not merged and not even reviewed since over 2 years: https://github.com/missive/emoji-mart/pull/809
- there are many other open PRs, like upgrades to newer react versions and bug fixes: https://github.com/missive/emoji-mart/pulls

WofWca:
Is there not an existing fork already?

Note that OSes usually have their own built-in emoji picker, usually invoked with `Win + .`.

Simon-Laux:
> Is there not an existing fork already?

Haven't found one of first look. We only need to fix the bugs we have, we don't really need to deal with anything else.
I'd say we just fork and merge the bug fix PRS that we like.

> Note that OSes usually have their own built-in emoji picker

Quality of those varies greatly and most users don't know about them, also emojis in there look different from in our app on Windows and on Linux.

r10s:
this seems to be solved for now by https://delta.chat/en/help#troubleshooting

nicodh:
> this seems to be solved for now by https://delta.chat/en/help#troubleshooting

Hmm, wrong link? It's solved with https://github.com/deltachat/deltachat-desktop/pull/6201

PR #6201, opened 2026-04-07
Replace emoji mart

nicodh:
This replaces the dependency to emoji-mart with a pinned version of a fork of the origin library. I checked the updates on https://github.com/viaduck/emoji-mart-awesome and consider them useful.

Future updates should be checked carefully.


WofWca (commented):
I think you mentioned a fork by the Element messenger devs? This one is not that?

> Future updates should be checked carefully.

Because this fork is not used (and therefore not checked) by many people?

nicodh:
The other fork was by session https://github.com/missive/emoji-mart/compare/main...session-foundation:session-emoji-mart:main but it seems less maintained and has no npm package.

I think to add it as a pinned version should be fine. We could check for changes and if anything happens we don't want or like we would have to look for another solution

packages/frontend/src/components/EmojiPicker/index.tsx
```diff
@@ -1,21 +1,31 @@
...
+  id: string
+  name: string
+  native: string
+  unified: string
+  shortcodes: string
+  skin?: number
+  keywords?: string[]
+  aliases?: string[]
+  emoticons?: string[]
+}
```
WofWca, line 21: Maybe add a link to where this is copy-pasted from?

packages/frontend/package.json
```diff
@@ -26,13 +26,13 @@
     "@deltachat-desktop/runtime-interface": "link:../runtime",
     "@deltachat-desktop/shared": "link:../shared",
     "@deltachat/jsonrpc-client": "catalog:",
+    "@emoji-mart-awesome/react": "^1.0.0",
```
WofWca: You mentioned that the version is pinned, but this is only the case for the `emoji-mart-awesome` package and not this one.

But anyway I think there is no strong need to pin the version in package.json, because one has to manually update pnpm-lock.json anyway. But I'm fine with either.

WofWca (commented):
Looks like this is related to https://github.com/deltachat/deltachat-desktop/issues/5935.

I have looked through the fork's changes on GitHub and found nothing malicious there.
But I didn't check whether the build matches the one published on NPM.

I also didn't check the Tauri version, didn't verify the CSP hash value.

Testing the reaction picker and the compoer's emoji picker returned no issues.

This also preserves the original "frequently picked" ones.

But: which of the fork's changes are actually important to us? Is it just about the skin tones?

packages/frontend/package.json
nicodh: Ok, I pinned now both, @emoji-mart-awesome/react and @emoji-mart/data

packages/frontend/src/components/EmojiPicker/index.tsx
nicodh: It is derived from https://github.com/missive/emoji-mart#-get-emoji-data-from-native. I didn't find a type export

nicodh:
> But: which of the fork's changes are actually important to us? Is it just about the skin tones?

The main point is the support for react 19 but I also consider the other changes useful as said above

WofWca:
> The main point is the support for react 19

But the one that we have now also works fine with React 19.

The fork that we're switching to also doesn't seem like it's that much maintained, so I don't see this as a significant improvement.

nicodh:
> The fork that we're switching to also doesn't seem like it's that much maintained, so I don't see this as a significant improvement.

So what would you propose? The one we use is not maintained since years.

WofWca:
Well it's an improvement nonetheless so I say we should merge this.

Only that I think that the separators are not supposed to look like this:

image

WofWca:
Here is the code that I suspect is responsible:

https://github.com/deltachat/deltachat-desktop/blob/2de7ce30502dace5e8d122df287418f600c9528a/packages/frontend/src/components/EmojiPicker/styles.module.scss#L1-L33

nicodh:
> Here is the code that I suspect is responsible:
> 
> https://github.com/deltachat/deltachat-desktop/blob/2de7ce30502dace5e8d122df287418f600c9528a/packages/frontend/src/components/EmojiPicker/styles.module.scss#L1-L33

I didn't find an easy way to override the styles so I used a <style> component to be appended to the picker. But it works...

packages/frontend/src/components/EmojiPicker/index.tsx
```diff
@@ -34,11 +44,26 @@ export default function EmojiPicker({
...
   }
 
+  const wrapperRef = useRef<HTMLDivElement>(null)
+
+  useEffect(() => {
+    const picker = wrapperRef.current?.querySelector('em-emoji-picker')
+    const shadowRoot = picker?.shadowRoot
+    if (!shadowRoot) return
+    const style = document.createElement('style')
+    style.textContent = '.category > .sticky { background: none !important; }'
```
WofWca: I think this makes things worse

Image

packages/frontend/src/components/EmojiPicker/index.tsx
```diff
@@ -34,11 +44,26 @@ export default function EmojiPicker({
...
+    shadowRoot.appendChild(style)
+    return () => {
+      shadowRoot.removeChild(style)
+    }
+  }, [])
```
WofWca, line 60: OK, looks like it's actually intended that it looks this way

https://github.com/missive/emoji-mart/compare/main...viaduck:emoji-mart-awesome:viaduck-v3#diff-ca441c0466b535006bab8ef9c4b06d06da13c420111682a47f3b593585b5a9d6R382

Either way I don't think we should be trying to deeply override its styles.

nicodh: I changed the bg color to the theme background.

> Either way I don't think we should be trying to deeply override its styles.

you have a better approach?

WofWca: If it's portant then we should add a CSS var upstream to make this configurable.
If not then we should leave it as is.

WofWca (approved):
Works.

PR #6202, opened 2026-04-07
feat: add inline reply on notifications (Win, Mac)

WofWca:
- **refactor: move `unknownErrorToString` to `shared`**
- **feat: add inline reply on notifications (Win, Mac)**

image

Electron recently added support for these on Windows,
so now I think it's worth implementing.

Platforms that don't support this are not affected:
they would simply not have the "reply" UI on the notification.

This is also compatible with earlier Electron versions,
namely the one that is used on the community FreeBSD distribution,
where it's Electron v39 IIRC.

Note that we already have inline replies on Tauri.


nicodh (approved):
Reviewed by reading & testing (on Windows)

Except notifications.ts all other changed files are only affected by the unknownErrorToString move

PR #6203, opened 2026-04-07
improvement: HTML viewer: move Load Remote Images

### LINKED ISSUE #4344: "Show Full Message…" window has "Show remote images" option even when there are no remote images
- Operating System (Linux/Mac/Windows/iOS/Android): Debian GNU/Linux trixie (testing)
- Delta Chat Version: 1.48.0
- Expected behavior: only present "Show Remote Images" checkbox when a message actually has remote images
- Actual behavior:  "Show Remote Images" checkbox appears in every large message.
- Steps to reproduce the problem:  Send a long message with no remote images or other resources to an account with deltachat-desktop.  click "Show Full Message …".  look in the upper right corner of the new window.

Having the "Show remote images…" option present even when there are no remote images simply encourages people to get used to checking that box just to see whether the message they get will change.  this is unhealthy conditioning.

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/4344.

The point is to make it more "hidden" so that people don't click it
just to see what it does.

Although the original issue is about the option being displayed
even when there are no remote images,
this commit improves the situation with that.

Note that this MR doesn't affect Tauri. In Tauri things remain as is.

### DIFF
--- a/packages/target-electron/src/windows/html_email.ts
+++ b/packages/target-electron/src/windows/html_email.ts
@@ -255,7 +255,7 @@ export function openHtmlEmailWindow(
         label: tx('load_remote_content'),
         checked: loadRemoteContent,
         click() {
-          update_restrictions(null, !loadRemoteContent)
+          update_restrictions(!loadRemoteContent)
         },
       }),
       always_show: () => ({
@@ -270,7 +270,7 @@ export function openHtmlEmailWindow(
             HTMLEmailAlwaysLoadRemoteContent: newValue,
           })
           // apply change
-          update_restrictions(null, newValue, true)
+          update_restrictions(newValue, true)
         },
       }),
       dont_ask: () => ({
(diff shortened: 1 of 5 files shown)


nicodh (approved):
Reviewed by reading & testing

Works as expected.

The 3dot menu is very small, but that's out of scope of this PR

image

r10s (approved):
it fixes the issue, can be merged as-is therefore

however, as this now always requires on additional tap, the "show warning" and "always load" option and their counterplay becomes even more weird and hard to understand for everyone who is not a nerd or a dev. the original idea was "i want to toggle that super quick, with one tap!!!!", it is still one tap less, but well, it is not worth having three menu options for that part.

in a subsequent PR, let's do the same as on android/iOS here, so always show a dialog and offer the options "Never", "Once", "Always". this would make things consistent, and remove some additional questions the user has.

this will slow down the "tapping" incentive even more

PR #6206, opened 2026-04-08
simplify call wording

r10s:
it is clear by position of the bubbles if it is incoming/outgoing, for summaries it is not important enough information, see https://github.com/deltachat/deltachat-android/pull/4263 for more discussions.


WofWca (approved):
I didn't assess whether this is good in itself, but this is in line with DC Android.

Also I have checked that these are the only places where `outgoing_video_call` and `incoming_audio_call` are used in DC Desktop,

PR #6207, opened 2026-04-09
feat: add button to clear all relay storage

link2xt:
This is a PR for testing https://github.com/chatmail/core/pull/8105

Not meant as a final UI, I have just put the button somewhere so I can click it and see if it works, currently looks like this:
image

It successfully wiped all my storage, now quota is reset to 0.


r10s:
with https://github.com/chatmail/relay/pull/929 merged, the main reason for that function is removed

PR #6208, opened 2026-04-09
refactor: simplify startup accounts path check

WofWca:
It doesn't make sense to wait for `initIsWindowsStorePackageVar`
prior to checking the existence of the path.

This also improves performance a little,
thanks to using the async version.


nicodh (approved):
While I agree that it is cleaner not to wait for initIsWindowsStorePackageVar to detect the accountsPath I doubt if the performance improvement is measurable

Issue #6210, opened 2026-04-09
crash on using the app


r10s:
the last days, i got random crashes, not sure if they are related to any action. i cannot reproduce it. but this time, i saved the log locally :)

[crash.txt](https://github.com/user-attachments/files/26607815/crash.txt)

i cannot remember having crashes like that before, but also here, not sure :)

version 2.48
macOS 15.7.3

WofWca:
Looks like a Chromium crash, but maybe I'm just not qualified to read the log?

First thing to try is upgrading Electron I guess.

nicodh:
Memory Tag 253 48.0G
Memory Tag 255 1.3T 

The virtual memory tags are absurd high

nicodh:
Maybe related to https://github.com/electron/electron/issues/49462 ?

There are the same high MemoryTags 48G & 1.3T

r10s:
> First thing to try is upgrading Electron I guess.

this is also what LLM suggested :)

> The virtual memory tags are absurd high

the app was definetely not running for a long time, maybe 10 minutes

WofWca:
Oops, closed accidentally

r10s:
another one, maybe same:

https://gist.github.com/r10s/bccf09cadcb227d25605ed1dd6259d15

Raiden-GH:
Version 2.49.0 crashes frequently on my system. I have the impression that it often happens shortly after launching the app, just as a new message arrives. But I need to monitor this further.

macOS 26.4.1
DC-Desktop 2.49.0 Electron-Version (Apple Silicon)

nicodh:
Is it the store install or the dmg?

r10s:
for me, it is dmg installation

Raiden-GH:
It is the dmg installation.

WofWca:
Looks similar to https://github.com/electron/electron/issues/50247 or https://github.com/electron/electron/issues/50393 but not sure. Those seem also Mac-specific.

WofWca:
We also got a Delta Chat log with an entry triggering `unhandledrejection`: "Unhandled Rejection: undefined".

https://github.com/deltachat/deltachat-desktop/blob/e75c0d712323734fd2ee2e8ad97c78e0bc915375/packages/frontend/src/App.tsx#L124-L130

It's the last entry in the log. The previous ones look normal

Raiden-GH:
I can now reliably reproduce the crash.

[...]

PR #6211, opened 2026-04-09
chore: upgrade Electron

WofWca:
This might fix
https://github.com/deltachat/deltachat-desktop/issues/6210.

### DIFF
--- a/packages/target-electron/package.json
+++ b/packages/target-electron/package.json
@@ -81,7 +81,7 @@
     "application-config": "^3.0.0",
     "chai": "^5.1.1",
     "debounce": "^1.2.0",
-    "electron": "^41.1.0",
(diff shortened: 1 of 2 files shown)


WofWca:
I tested it. It works (i.e. doesn't break horribly, but IDK if it fixes the issue).

nicodh:
> This might fix #6210.

Just a guess or were there related fixes?

WofWca:
There are some fixes for crashes in the change log, but I don't think they are directly relevant.

So yep, basically just a guess.

nicodh:
It's always a bit risky to upgrade shortly(?) before a bugfix release - but if there are crashes now anyway it might be worth

nicodh (approved):
Reviewed by reading &"smoke" testing

PR #6212, opened 2026-04-10
fix: add bg for quotes in video msg

nicodh:
Resolves #6192 

image


WofWca (approved):
Looks better, I'd say fixes the issue. Nice that it's after all possible to reuse the existing styles.

PR #6213, opened 2026-04-10
refactor: DRY `AppStoreUrl`

WofWca:
This was the intent of the commit that introduced this variable
but I guess I forgot to actually use it.


nicodh (approved):
Reviewed by reading

PR #6215, opened 2026-04-10
feat: custom app picker URL setting

WofWca:
- **refactor: DRY `AppStoreUrl`**
- **refactor: App Picker: don't export `AppStoreUrl`**
- **refactor: make `DeltaInput` support more props**
- **refactor: narrower `setDesktopSetting` types**
- **refactor: setDesktopSetting: allow `undefined` val**
- **feat: setting to set app picker URL**

image

For testing one can use https://wofwca.github.io/webxdc_website/apps/.

Related: https://github.com/webxdc/website/issues/135.

### DIFF
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -64,5 +64,8 @@
   },
   "paste_in_qr_scanner": {
     "message": "Click to paste it in QR scanner"
+  },
+  "webxdc_store_url_explain_2_desktop": {
+    "message": "This is different from the same feature on Android and iOS.\nThis must be the path containing the xdcget-lock.json file."
   }
 }
--- a/packages/e2e-tests/tests/basic-tests.spec.ts
+++ b/packages/e2e-tests/tests/basic-tests.spec.ts
@@ -420,6 +420,34 @@ test('add app from picker to chat', async () => {
   expect(finalAppIconsCount).toBeGreaterThan(initialAppIconsCount)
 })
 
+test('custom app picker URL', async () => {
+  // It's the default URL but still let's check if things work.
+  // Change casing to check if changing the setting value works
+  const newUrl = 'HTTPS://APPS.TESTRUN.ORG/'
+
+  await page.getByRole('button', { name: 'Settings' }).click()
+  await page.getByRole('button', { name: 'Advanced' }).click()
+  const settingsDialog = page
+    .getByRole('dialog')
+    .filter({ hasText: 'Advanced' })
+
+  await expect(settingsDialog).not.toContainText(newUrl, { ignoreCase: false })
+  await page.getByRole('button', { name: 'App Picker URL' }).click()
+  await page.getByRole('dialog').last().getByRole('textbox').fill(newUrl)
+  await page.keyboard.press('Enter')
+  await expect(settingsDialog).toContainText(newUrl, { ignoreCase: false })
+
+  await page.keyboard.press('Escape')
+  await page.keyboard.press('Escape')
+
+  await page.getByRole('button', { name: 'Attach' }).click()
+  await page.getByRole('menuitem', { name: 'App' }).click()
+  await page.getByRole('button', { name: 'Poll' }).first().click()
+  await page.getByRole('button', { name: 'Add to Chat' }).click()
+  await page.getByRole('button', { name: 'Send', exact: true }).click()
+  await expect(page.locator(`.message`).last()).toContainText('Poll')
+})
+
 test('recent apps context menu', async () => {
   await page
     .getByTestId('last-used-apps')
--- a/packages/e2e-tests/tests/basic-tests.spec.ts
+++ b/packages/e2e-tests/tests/basic-tests.spec.ts
@@ -11,6 +11,7 @@ import {
   reloadPage,
   sendMessage,
   test,
+  selectChat,
 } from '../playwright-helper'
 
 /**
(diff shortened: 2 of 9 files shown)


WofWca:
Thinking this over: there might be a bit of a problem in the fact that this feature is not exactly 1-1 with the respective Android feature. The thing is that Android expects that the URL will load a website (index.html), and Desktop with this MR expects that there is gonna be `./xdc-lock.json` file at that URL. Which is only the case for https://wofwca.github.io/webxdc_website/apps/.

Maybe one could say that this is a problem of the store websites. Or we should make the distinction with the Android's feature clearer.

r10s:
iirc, it was mentioned elsewhere: we need to first fix the main-website to allow using the exact same url on both, android and desktop. this is very likely easier than to deal with frustrated user, confusion and additional support questions

WofWca:
Rebased

WofWca:
OK, I amended the last commit, added an extra string to the dialog. See the new screenshot in the MR description.

WofWca:
> iirc, it was mentioned elsewhere: we need to first fix the main-website

To clarify, I heard that later in private discussions it was concluded that this feature is still useful.

Despite the fact that I think it's better to proceed with my plan https://github.com/webxdc/website/issues/135 rather than making users change the URL, we anyway should not have unchangeable hard-coded URLs (which this MR helps with).

_locales/_untranslated_en.json
```diff
@@ -64,5 +64,8 @@
   },
   "paste_in_qr_scanner": {
     "message": "Click to paste it in QR scanner"
+  },
+  "webxdc_store_url_explain_2_desktop": {
+    "message": "This is different from the same feature on Android and iOS.\nThis must the path containing the xdcget-lock.json file."
```
nicodh:
```suggestion
    "message": "This is different from the same feature on Android and iOS.\nThis must be the path containing the xdcget-lock.json file."
```
Wouldn't it be better to validate the path by checking that there is an xdcget-lock.json file provided? Then we don't need to mention it in the hint but just show an error in case it misses:
"This is not a valid app picker URL. No xdcget-lock.json was found in that path!"

packages/shared/shared-types.d.ts
```diff
@@ -86,6 +89,12 @@ export interface DesktopSettingsType {
   autostart: boolean
   /** whether to start Electron with system on supported platforms */
   autostartElectron: boolean
+  /**
+   * URL relative to which to fetch the `xdc-lock.json` file.
```
nicodh:
```suggestion
   * URL relative to which to fetch the `xdcget-lock.json` file.
```

packages/frontend/src/components/AppPicker/index.tsx
```diff
@@ -134,7 +146,13 @@ export function AppPicker({ onAppSelected }: Props) {
       let count = 0
       for (const app of apps) {
         BackendRemote.rpc
-          .getHttpResponse(selectedAccountId(), AppStoreUrl + app.icon_relname)
+          .getHttpResponse(
+            selectedAccountId(),
+            // Note that here we're still using the default store URL.
+            // This is because we don't expect the custom store to host icons.
+            // We only expect the `xdc-lock.json` file from it.
```
nicodh:
```suggestion
            // We only expect the `xdcget-lock.json` file from it.
```

So what is the idea to handle the icons then? This means the app picker url is not intended to provide other or additional apps but just an other url? Is it reasonable to list apps without icons?

packages/e2e-tests/tests/basic-tests.spec.ts
```diff
@@ -421,6 +422,39 @@ test('add app from picker to chat', async () => {
...
 
+test('custom app picker URL', async () => {
+  const userA = existingProfiles[0]
+  const userB = existingProfiles[1]
+  await switchToProfile(page, userA.id)
+  await selectChat(page, userB.name)
+
+  // It's the default URL but still let's check if things work.
+  // Change casing to check if changing the setting value works.
+  const newUrl = 'HTTPS://APPS.testRUN.ORG/'
```
nicodh, line 435: It's not clear what exactly we test here. The fact that casing is ignored when a custom URL is provided? Should the URL fail the appStoreUrl === defaultAppStoreBaseUrl condition in [AppPicker/index.tsx#375 ](https://github.com/deltachat/deltachat-desktop/pull/6215/changes#diff-0503af9120be813fec4e65f02c25da3cb71bdc11930120e2a3dabd6e8ff3d206R374) so the URL is used as "fake" custom URL?

At least this comment should then say somethinglike: the different casing should trick the settings store to consider it a new URL while the network just ignores the case.

Don't we want to normalize case when adding a new URL?

_locales/_untranslated_en.json
WofWca: That'd be nice, but this is an advanced feature anyway, which is also subject to change. So I'd say this can go in a follow-up.

packages/frontend/src/components/AppPicker/index.tsx
WofWca: > So what is the idea to handle the icons then?

WDYM?

> This means the app picker url is not intended to provide other or additional apps but just an other url?

As of now, yes, this is the main idea.

> Is it reasonable to list apps without icons?

This is indeed not great, but it is better than not providing anything at all.

packages/e2e-tests/tests/basic-tests.spec.ts
WofWca: > Don't we want to normalize case when adding a new URL?

What would be the purpose? I don't think we need extra logic. Especially given that URL paths are actually case-sensitive in general, i.e. observable by the server.

WofWca:
Fixed comments and the string, as suggested.

nicodh (approved):
Reviewed by reading & testing

PR #6216, opened 2026-04-10
style: round outline for settings button

WofWca:
And also make sure that it's centered inside the outline,
by setting `line-height: 0`.

image

### DIFF
--- a/packages/frontend/src/components/AccountListSidebar/styles.module.scss
+++ b/packages/frontend/src/components/AccountListSidebar/styles.module.scss
@@ -199,6 +199,8 @@
 
   .settingsButton {
     background-color: transparent;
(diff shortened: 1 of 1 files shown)


r10s:
what is the reason for that change?

WofWca:
The reason is that it looks better than what we have now:

image

r10s:
i will have a closer look these days. what i always disliked is that it looked too disabled

r10s:
yip, the settings icon should be focused inside the focus rectangle.

but the focus rectangle should be the same as for other buttons on the sidebar, so not round.

if one wants to improve things here, it makes sense to have the settings icon a bit lighter, roughly the same white level as the other icons

PR #6217, opened 2026-04-10
improvement: better app picker fetch error msg

WofWca:
Especially now that it's possible to set a custom app store URL,
let's make it clearer what's going on.

TODO:
- [x] Merge the base, up to the last commit


nicodh (approved):
Reviewed by reading

I'm not sure if the current approach split everything in multiple very small PRs is so useful. Here I see the should be merged into branch 3aa9689d19, so I have to search which PR is bound to that branch - turns out to be a draft (which I would not review)

WofWca:
Well I think this MR is useful for both current main and that draft MR. So the base of this commit is not too important.
But yep I should have done something with this MR after marking the base one as draft. I forgot.

PR #6219, opened 2026-04-11
fix: use correct filename when dragging a file out

### LINKED ISSUE #5108: drag file out: filename is hash
The filename is the hash from the blob db instead of the original file.
Affects both electron and tauri edition.

We may need to copy it to a temporary directory with it's real name before starting the drag out action.
(though there is a slight question of when we can delete the temporary file again.)

nicodh:
resolves #5108 partly (only for electron not for tauri!)

### DIFF
--- a/packages/frontend/src/components/attachment/Attachment.tsx
+++ b/packages/frontend/src/components/attachment/Attachment.tsx
@@ -74,8 +74,8 @@ export function getExtension({
 }
 
 export function dragAttachmentOut(
-  file: MessageTypeAttachmentSubset['file'],
-  fileName: string,
+  file: string,
+  fileName: string | null,
   dragEvent: React.DragEvent<HTMLElement>
 ) {
   dragEvent.preventDefault()
--- a/packages/frontend/src/components/attachment/Attachment.tsx
+++ b/packages/frontend/src/components/attachment/Attachment.tsx
@@ -74,7 +74,7 @@ export function getExtension({
 }
 
 export function dragAttachmentOut(
-  file: string,
+  file: MessageTypeAttachmentSubset['file'] & string,
   fileName: string | null,
   dragEvent: React.DragEvent<HTMLElement>
 ) {
--- a/packages/frontend/src/components/attachment/Attachment.tsx
+++ b/packages/frontend/src/components/attachment/Attachment.tsx
@@ -76,7 +76,7 @@ export function getExtension({
 export function dragAttachmentOut(
   file: MessageTypeAttachmentSubset['file'],
   fileName: string,
-  dragEvent: React.DragEvent<HTMLDivElement>
+  dragEvent: React.DragEvent<HTMLElement>
 ) {
   dragEvent.preventDefault()
 
--- a/packages/frontend/src/components/attachment/Attachment.tsx
+++ b/packages/frontend/src/components/attachment/Attachment.tsx
@@ -75,10 +75,12 @@ export function getExtension({
 
 export function dragAttachmentOut(
   file: MessageTypeAttachmentSubset['file'],
+  fileName: string,
   dragEvent: React.DragEvent<HTMLDivElement>
 ) {
   dragEvent.preventDefault()
-  if (file) runtime.onDragFileOut(file)
+
+  if (file) runtime.onDragFileOut(file, fileName)
 }
 
 export type MessageTypeAttachmentSubset = Pick<
--- a/packages/runtime/runtime.ts
+++ b/packages/runtime/runtime.ts
@@ -182,7 +182,7 @@ export interface Runtime {
    * only support this if you have a real implementation for `isDroppedFileFromOutside`
    * file is the name of the hashed file and realName is the original name of the file
    */
-  onDragFileOut(file: string, realName: string): void
+  onDragFileOut(file: string, realName: string | null): void
   /** Set drag listener to handle drag and drop events */
   setDropListener(onDrop: DropListener | null): void
   /** guard function that checks if it is a file from `onDragFileOut`, if so it denies the drop.
--- a/packages/runtime/runtime.ts
+++ b/packages/runtime/runtime.ts
@@ -178,8 +178,11 @@ export interface Runtime {
   } | null>
   saveBackgroundImage(file: string, isDefaultPicture: boolean): Promise<string>
 
-  /** only support this if you have a real implementation for `isDroppedFileFromOutside`  */
-  onDragFileOut(file: string): void
+  /**
+   * only support this if you have a real implementation for `isDroppedFileFromOutside`
+   * file is the name of the hashed file and realName is the original name of the file
+   */
+  onDragFileOut(file: string, realName: string): void
   /** Set drag listener to handle drag and drop events */
   setDropListener(onDrop: DropListener | null): void
   /** guard function that checks if it is a file from `onDragFileOut`, if so it denies the drop.
--- a/packages/target-electron/runtime-electron/runtime.ts
+++ b/packages/target-electron/runtime-electron/runtime.ts
@@ -138,7 +138,7 @@ class ElectronRuntime implements Runtime {
   onShowDialog:
     | ((kind: 'about' | 'keybindings' | 'settings') => void)
     | undefined
-  onDragFileOut(file: string, realName: string): void {
+  onDragFileOut(file: string, realName: string | null): void {
     ipcBackend.send('ondragstart', file, realName)
   }
   isDroppedFileFromOutside(file: string): boolean {
--- a/packages/target-electron/runtime-electron/runtime.ts
+++ b/packages/target-electron/runtime-electron/runtime.ts
@@ -138,8 +138,8 @@ class ElectronRuntime implements Runtime {
   onShowDialog:
     | ((kind: 'about' | 'keybindings' | 'settings') => void)
     | undefined
-  onDragFileOut(file: string): void {
-    ipcBackend.send('ondragstart', file)
+  onDragFileOut(file: string, realName: string): void {
+    ipcBackend.send('ondragstart', file, realName)
   }
   isDroppedFileFromOutside(file: string): boolean {
     // ".sqlite-blobs" is the old folder name that could still be there in old accounts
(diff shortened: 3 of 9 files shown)


packages/target-electron/src/ipc.ts
```diff
@@ -111,7 +111,14 @@ export async function init(cwd: string, logHandler: LogHandler) {
...
   )
 
-  ipcMain.on('ondragstart', (event, filePath) => {
+  ipcMain.on('ondragstart', (event, filePath, realName) => {
+    let tmpFilePath = filePath
+    if (realName && realName !== '') {
+      const tmpDir = join(getDraftTempDir(), `drag-${Date.now()}`)
+      mkdirSync(tmpDir, { recursive: true })
+      tmpFilePath = join(tmpDir, basename(realName))
+      copyFileSync(filePath, tmpFilePath)
```
WofWca: Why do we need to use the sync FS API? Async is as easy to use but it doesn't block the whole process.

packages/target-electron/src/ipc.ts
```diff
@@ -111,7 +111,14 @@ export async function init(cwd: string, logHandler: LogHandler) {
     logHandler.log(channel, level, stacktrace, ...args)
   )
 
-  ipcMain.on('ondragstart', (event, filePath) => {
+  ipcMain.on('ondragstart', (event, filePath, realName) => {
+    let tmpFilePath = filePath
+    if (realName && realName !== '') {
+      const tmpDir = join(getDraftTempDir(), `drag-${Date.now()}`)
```
WofWca: The comment on `getDraftTempDir` is now outdated, and maybe the name of the function itself.

packages/target-electron/runtime-electron/runtime.ts
```diff
@@ -138,8 +138,8 @@ class ElectronRuntime implements Runtime {
   onShowDialog:
     | ((kind: 'about' | 'keybindings' | 'settings') => void)
     | undefined
-  onDragFileOut(file: string): void {
-    ipcBackend.send('ondragstart', file)
+  onDragFileOut(file: string, realName: string): void {
```
WofWca:
```suggestion
  onDragFileOut(file: string, realName: string | null): void {
```

Otherwise the signature doesn't match.

packages/target-electron/src/ipc.ts
WofWca: The file is not deleted until after reboot or app restart, right?

packages/frontend/src/components/attachment/Attachment.tsx
```diff
@@ -74,11 +74,13 @@ export function getExtension({
 }
 
 export function dragAttachmentOut(
-  file: MessageTypeAttachmentSubset['file'],
-  dragEvent: React.DragEvent<HTMLDivElement>
+  file: string,
```
WofWca:
```suggestion
  file: MessageTypeAttachmentSubset['file'] & string,
```

WofWca (commented):
Security-wise looks OK, the file name is wrapped in `basename`.
The Tauri version will keep working as is as far as I can see.

packages/frontend/src/components/attachment/Attachment.tsx
nicodh: What is the benefit of using that type definition?

packages/target-electron/src/ipc.ts
nicodh: Yes, I don't consider that critical

packages/target-electron/src/ipc.ts
nicodh: You mean the comment should mention the additional use case?

"and maybe the name of the function itself." what do you mean?

packages/target-electron/src/ipc.ts
nicodh: Since we need the file to be copied before the user "drops" the file

packages/target-electron/src/ipc.ts
WofWca: 1. Yes
2. The name of the function references `draft`, but this use case has nothing to do with drafts.

packages/target-electron/src/ipc.ts
WofWca: Does it really help though? This doesn't prevent the user from moving the mouse and releasing the drag. Am I misunderstanding?

packages/frontend/src/components/attachment/Attachment.tsx
WofWca: This makes it clear what you're supposed to provide as the argument. Which might clarify that this is the file path, and not the base64 contents of the file or something.

packages/target-electron/src/ipc.ts
WofWca: Hmm so you brought up that the file needs to be copied before the "drop" event happens. Which seems like a big thing. Copying a file is not instantaneous especially on HDDs, we might not be able to afford it for such quick operations as "drag and drop".

Would it be possible to first set the original file path, and then, when we have copied the file, set it again but to the new file? i.e. calling `startDrag` two times I guess.

nicodh: No - the move can't be interrupted of course - but onDragEnd would not be fired before onDragStart is finished since the latter blocks main. But tbh I'm not really sure

packages/target-electron/src/ipc.ts
nicodh: Ok, I will fix it

packages/target-electron/src/ipc.ts
WofWca: Also have you considered [`fs.link()`](https://nodejs.org/api/fs.html#fspromiseslinkexistingpath-newpath)?

nicodh: But then we would have to check for filesystem support. So your concern is, that the drag operation freezes the UI for a notable time?

WofWca: Yes, and that the drag-and-drop operation is just overall slowed down.

nicodh: We could use linkSync with a fallback to copyFileSync if not supported. I just played around a bit. With a 500mb file I have like half a second not visible drag file icon. But that is already much bigger then we expect to handle

WofWca: I'm not so sure about this. On some OSes the `/tmp` size is quite limited, around ~4 GB. If you drag and drop a 100 MB file 10 times you've just filled 1/4 of all the available space. With people not rebooting their machines (and utilizing sleep instead) this can be a problem.

WofWca: > With a 500mb file I have like half a second not visible drag file icon

I assume you have an SSD and it's not otherwise busy.

nicodh: Yes but 500mb is far beyond the expected range.
Anyway, I can use link with a fallback to copy since that's not much effort

nicodh: But we don't have a cleanup in place for the draft usages of the tmp dir. So this is not only about this use case.

I don't know if it makes sense to consider extreme edge cases (4GB, 10x drag a 100MB file, never reboot)
That would mean we have to listen for the dragend event which happens outside of electron

nicodh: Other solution would be remove the tmp dir after 30 seconds. If the app is closed before the tmp dir will be anyway cleaned up on next start

WofWca: > But we don't have a cleanup in place for the draft usages of the tmp dir. 

I don't think so? We have and utilize `removeTempFile`.

> 30 seconds

Too little probably. They might drag-and-drop the file to e.g. a file picker of a browser, and we'd have to keep the file until they have submitted the form

Anyways the `fs.link` thing might greatly improve the situation.

nicodh: > I don't think so? We have and utilize removeTempFile.

voice messages for example create a draft that is not deleted after sending

> They might drag-and-drop the file to e.g. a file picker of a browser, and we'd have to keep the file until they have submitted the form

This is even a more constructed edge case than above. There is no need to always cover all possible use cases in all thinkable scenarios. Not if they are not security relevant. So even if you keep the file 8 hours, what if the user submits the form a day later? etc.

packages/target-electron/src/application-constants.ts
```diff
@@ -66,8 +66,8 @@ export function getCustomThemesPath() {
 }
 
 // this is used for temporary files (because core expects file paths, can not accept blobs directly yet)
```
WofWca: This line is still outdated.

packages/target-electron/src/application-constants.ts
WofWca: The dir name is outdated, this is not specific to drafts

packages/runtime/runtime.ts
```diff
@@ -178,8 +178,11 @@ export interface Runtime {
   } | null>
   saveBackgroundImage(file: string, isDefaultPicture: boolean): Promise<string>
 
-  /** only support this if you have a real implementation for `isDroppedFileFromOutside`  */
-  onDragFileOut(file: string): void
+  /**
+   * only support this if you have a real implementation for `isDroppedFileFromOutside`
+   * file is the name of the hashed file and realName is the original name of the file
```
WofWca, line 183:
```suggestion
   * file is the path to the file (with its name being a hash if it's in the blob dir)
   * and realName is the original name of the file
```

WofWca (approved):
For the reasons that I outlined I think this is not worth it.
Probably better to fix this in Core or add Electron API to set file name.
Also as we've talked, the new tmp path pattern not covered by the `isDroppedFileFromOutside`, so that code is broken now, and it's noticeable on Windows.
I did not check whether we need to handle more errors beside `EXDEV`. For clarity, `EXDEV` means "oldpath and newpath are not on the same mounted filesystem", and it's is expected to happen on Linux, because IMU `/tmp` is tmpfs, which is different from the regular fs.

Otherwise drag-and-drop seems to work on my Linux and Windows machines, at least in the most basic case. The file name is preserved, so this fixes the issue.

nicodh:
> For the reasons that I outlined I think this is not worth it.

You mean fix the issue is  not worth it?

WofWca:
Well fixing the issue would be nice, but without the downsides of this MR.

PR #6221, opened 2026-04-11
fix: better prevent `chatId` leaking between accs

### LINKED ISSUE #5993: Chat React state leaking between accounts (`ChatProvider` is not behind `key={accountId}`)
Related:
- https://github.com/deltachat/deltachat-desktop/pull/5992
- https://github.com/deltachat/deltachat-desktop/pull/3789#issuecomment-3707004972

Even after https://github.com/deltachat/deltachat-desktop/pull/5992 there are still some races.

I think one of them is that `selectChat` gets captured and then executed inside of an async function whose execution might get delayed, and by that time we might have switched the account already. But, since the function has captured the `accountId` variable, the `if (accountId !== nextAccountId) {` doesn't really help.

https://github.com/deltachat/deltachat-desktop/blob/4ffc6b03ebc402db470811151c31d2b5656d08d9/packages/frontend/src/contexts/ChatContext.tsx#L95-L156

`selectLastChat` is one place where this could be happening. Just imagine that there is a `await new Promise(r => setTimeout(r, 10_000))` after `await getLastChatId()`.

https://github.com/deltachat/deltachat-desktop/blob/454925d87bbd52aab9c4c47c63543103a7540dfd/packages/frontend/src/hooks/chat/useSelectLastChat.ts#L21-L34

The result is that `setChatId` gets executed, but that `chatId` might have been one obtained from the old account's `getLastChatId(accountId)`.

I believe that we after all should strive for putting `key={accountId}` on `ChatProvider`, as I mentioned in https://github.com/deltachat/deltachat-desktop/pull/3789#issuecomment-3707004972, instead of trying to re-implement React's `key` feature inside of it.
Although I might be wrong and `ChatProvider` is really one special piece that we should implement manually.

Anyways, the question is: how do we handle the fact that a captured `selectChat` might get executed with a big delay? Or is the question "how do we make sure to never execute it if it's outdated?"?

WofWca:
I think we can say that
this closes https://github.com/deltachat/deltachat-desktop/issues/5993.
I have tried switching accounts rapidly with this and have not managed
to trigger any errors in the console,
besides the occasional
"Tried to jumpToMessage, but messageListItems is empty".

Swithcing between accounts rapidly does indeed print the new
"but we've already switched the account" log from time to time.

I am pretty sure that this shouldn't break things.
I was worried that there might be cases where we actually _want_
to have the `selectChat` closure stored and invoked asynchronously,
such as for "switch to another account, then select a chat with this ID
in that account", but we utilize `saveLastChatId()`
followed by `selectAccount()` in such cases,
e.g. in the notification click handler.
And the function itself already has a guard check
for whether the target account ID is the same as the account ID
that we had selected when the closure was created.

I tested that the notification click handler
does properly switch accounts and select the target chat
and highlight the target message.
Cross-account message forwarding also works fine,
also cancelling the forward.

I am not sure about why the React linter stopped complaining about
`unselectChatRef.current = unselectChat`.
I had to remove the "ignore" comment there, otherwise it complains about
"Unused eslint-disable directive".

I am TBH not happy that I have to write this.
It feels like React is missing API for this,
something like a "weaker" `key=`.
Or maybe I'm bad at looking for a library that does this.
Or maybe I'm not seeing a simpler solution.

### DIFF
--- a/packages/frontend/src/ScreenController.tsx
+++ b/packages/frontend/src/ScreenController.tsx
@@ -173,6 +173,14 @@ export default class ScreenController extends Component {
       return
     }
 
+    // Now that we automatically invalidate `chatId` when `accountId` changes
+    // in `ChatContext`, one might think that it's not necessary
+    // to explicitly `unselectChat()` here.
+    // But apparently without this we still get `chatId` leaking between
+    // different `accountId`s, namely when loading recent app icons
+    // while rapidly switching accounts.
+    // Maybe this has to do with us using `window.__selectedAccountId`
+    // in some places and the `accountId` prop in others.
     this.unselectChatRef.current?.()
 
     const previousAccountId = this.selectedAccountId
--- a/packages/frontend/src/ScreenController.tsx
+++ b/packages/frontend/src/ScreenController.tsx
@@ -173,7 +173,7 @@ export default class ScreenController extends Component {
       return
     }
 
-    // Now that we automatically invalidate `chatId` when `accountId` changes
+    // Since we automatically invalidate `chatId` when `accountId` changes
     // in `ChatContext`, one might think that it's not necessary
     // to explicitly `unselectChat()` here.
     // But apparently without this we still get `chatId` leaking between
--- a/packages/frontend/src/contexts/ChatContext.tsx
+++ b/packages/frontend/src/contexts/ChatContext.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useRef, useState } from 'react'
+import React, { useCallback, useEffect, useRef, useMemo, useState } from 'react'
 
 import { ActionEmitter, KeybindAction } from '../keybindings'
 import { marknoticedChat, saveLastChatId } from '../backend/chat'
@@ -56,7 +56,45 @@ export const ChatProvider = ({
   accountId,
   unselectChatRef,
 }: PropsWithChildren<Props>) => {
-  const [chatId, setChatId] = useState<number | undefined>()
+  const sessionId = useMemo(
+    () => Symbol(),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [accountId]
+  )
+  const sessionIdRef = useRef(sessionId)
+  sessionIdRef.current = sessionId
+
+  // When the `accountId` changes, we invalidate `chatId`,
+  // and the `setChatId` function.
+  //
+  // We usually utilize `key={accountId}` for this,
+  // but applying `key=` to a React context provider
+  // makes _all_ its descendants (as opposed just the components
+  // that depend on the context) re-render, including `AccountListSidebar`.
+  const [chatId_, setChatId_] = useState<
+    undefined | { sessionId: symbol; chatId: number }
+  >(undefined)
+  const chatId =
+    chatId_ == undefined || chatId_.sessionId !== sessionId
+      ? undefined
+      : chatId_.chatId
+  const setChatId = useCallback(
+    (newChatId: number | undefined) => {
+      if (sessionId !== sessionIdRef.current) {
+        // This may happen when `selectChat` gets called
+        // from `Promise.then` or `setTimeout` and such.
+        log.info(
+          "Called setChatId, but we've already switched the account after the closure was created, ignoring"
+        )
+        return
+      }
+      setChatId_(
+        newChatId == undefined ? undefined : { chatId: newChatId, sessionId }
+      )
+    },
+    [sessionId]
+  )
+
   useEffect(() => {
     window.__selectedChatId = chatId
   }, [chatId])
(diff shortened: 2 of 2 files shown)


nicodh:
Why do we need to introduce a new sessionId?

Couldn't we just use an accountIdRef instead? That would be clearer and better to understand.

```javascript
const accountIdRef = useRef(accountId)
  useLayoutEffect(() => {
    accountIdRef.current = accountId
  }, [accountId])
  `

**Comment by nicodh** [comment]
It would be very helpful to have a test for this

**Comment by WofWca** [comment]
> It would be very helpful to have a test for this

Not sure how one would write such a test? We don't even have repro steps in the issue.
But we do have tests that have to do with switching between chats, so that's something.

**Comment by WofWca** [comment]
> Why do we need to introduce a new sessionId?

Good point. I though about that, and probably should have written about this in the description or comments.
Reasons:
- if we switch to another account and then back, we probably should have the stuff still invalidated. Relying on `accountId` would "restore" the previous state, which is not the case for `sessionId`.
- We already have multiple other places where we utilize this pattern, so perhaps the readability is not that different?

**Comment by nicodh** [comment]
> * if we switch to another account and then back, we probably should have the stuff still invalidated. Relying on `accountId` would "restore" the previous state, which is not the case for `sessionId`.

So we have a race condition that is almost not reproducable (switching very fast between accounts) and try to prevent an even more specific case (switching very fast forth and back between 2 accounts) of that race condition? Idk if that is worth to introduce more complexity

> We already have multiple other places where we utilize this pattern

also using the name sessionId?

In general I would like to evaluate again the much simpler approach (using key=accountId) in useChat. Chatlist and messageList have to be rerendered anyway when switching accounts. If the only cost would be an unneccessary rerender of the account list that seems not too expensive....

**Comment by WofWca** [comment]
> also using the name sessionId?

We have `selectionId_`, `thisMessageListIntanceId`, `fetchId`.

> Idk if that is worth to introduce more complexity

I would argue that having stuff depend on `accountId` is more complexity. With `sessionId` you are sure that when `accountId` changes then stuff is invalidated, whereas with `accountId` you have to keep in mind that old state can be retained and come back.

> If the only cost would be an unneccessary rerender of the account list that seems not too expensive....

The thing is that it blinks and has to do a full reload, and the scroll position gets lost. It doesn't look good at all.
I sure would also like to utilize `key` but that doesn't seem possible without a bigger refactor.

**Comment by WofWca** [comment]
> In general I would like to evaluate again the much simpler approach (using key=accountId) in useChat

I just had an idea: move the `T.FullChat` out of `ChatProvider`, to some level below, and only keep `chatId` in `ChatProvider`. I'll try to see if we have obstacles for it.

**Comment by WofWca** [comment]
OK, I realized that that idea would not eliminate the need for the code that invalidates `chatId`. It would simply split `ChatProvider`.

File: packages/frontend/src/contexts/ChatContext.tsx [hunk 1]
```diff
@@ -80,10 +118,9 @@ export const ChatProvider = ({
     // Make sure that the chat belongs to the current account,
     // to prevent data leaking between accounts, and weird race-y bugs.
     //
-    // We usually utilize `key={accountId}` for this,
-    // but applying `key=` to a React context provider
-    // makes _all_ its descendants (as opposed just the components
-    // that depend on the context) re-render, including `AccountListSidebar`.
+    // I think now that we have `sessionId` this check is no longer needed,
+    // because that basically ensures that `chatId` is always valid
+    // for the current `accountId`?
```
nicodh, line 122: If it is no longer needed we should remove it

packages/frontend/src/ScreenController.tsx
```diff
@@ -173,6 +173,14 @@ export default class ScreenController extends Component {
...
     }
 
+    // Now that we automatically invalidate `chatId` when `accountId` changes
+    // in `ChatContext`, one might think that it's not necessary
+    // to explicitly `unselectChat()` here.
+    // But apparently without this we still get `chatId` leaking between
+    // different `accountId`s, namely when loading recent app icons
+    // while rapidly switching accounts.
+    // Maybe this has to do with us using `window.__selectedAccountId`
+    // in some places and the `accountId` prop in others.
```
nicodh: This comment makes only sense in the context of the current PR. For future reads it should be sufficient to say something like


```suggestion
    // This is needed as without this we still get `chatId` leaking between
    // different `accountId`s, namely when loading recent app icons
    // while rapidly switching accounts.
    // Maybe this has to do with us using `window.__selectedAccountId`
    // in some places and the `accountId` prop in others
```

nicodh (approved):
Let's merge this if there is no better solution. At least by reading it seems to solve the issue

packages/frontend/src/ScreenController.tsx
WofWca: Perhaps yes, this could be worded less "temporally", but the point is that `unselectChat()` is still needed _despite the fact_ that we automatically invalidate `chatId`.

packages/frontend/src/contexts/ChatContext.tsx
WofWca: I am not sure if we don't, plus it's not too much code.
I suggest to keep it until it gets in the way (or until someone needs to touch this code again).

packages/frontend/src/ScreenController.tsx
```diff
@@ -173,6 +173,14 @@ export default class ScreenController extends Component {
       return
     }
 
+    // Now that we automatically invalidate `chatId` when `accountId` changes
```
WofWca:
```suggestion
    // Since we automatically invalidate `chatId` when `accountId` changes
```

WofWca:
Replaced "Now that" with "since" and rebased.

PR #6223, opened 2026-04-11
feat: `Ctrl + Alt + PageDown` to switch accounts

WofWca:
Why not `Ctrl + Shift + PageDown`? Because it feels like
an "alternative action" on the message list
where we have `Ctrl + PageDown` to switch between chats.
Whereas "Alt" is a little more awkward,
so it feels further from the action of switching between chats.

Closes https://support.delta.chat/t/can-you-provide-a-shortcut-key-to-switch-between-different-accounts/3259?u=wofwca.

image

### DIFF
--- a/docs/KEYBINDINGS.md
+++ b/docs/KEYBINDINGS.md
@@ -2,6 +2,8 @@
 
 > If you are on mac replace alt with the option key.
 
+Switch between accounts: `Ctrl + Alt + PageUp`, `Ctrl + Alt + PageDown`
+
 Switch between chats: `alt + arrow down`, `alt + arrow up`,
 `Ctrl + PageDown`, `Ctrl + PageUp`,
 `Ctrl + Tab`, `Ctrl + Shift + Tab`
--- a/packages/e2e-tests/tests/keyboard.spec.ts
+++ b/packages/e2e-tests/tests/keyboard.spec.ts
@@ -13,13 +13,23 @@ test.describe.configure({
 
 expect.configure({ timeout: 5_000 })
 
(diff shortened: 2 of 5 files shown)


WofWca:
Should also add tests for this.

nicodh (commented):
Reviewed by reading &  testing

Needs rebase!

nicodh (approved):
Needs rebase

_locales/_untranslated_en.json
```diff
@@ -124,5 +124,8 @@
   },
   "remove_attachment": {
     "message": "Remove Attachment"
+  },
+  "switch_between_accounts": {
+    "message": "Switch between Accounts"
```
r10s: we are not saying "Account" but "Profile" since some time, so let's use the existing string "Switch Profile". the string id for that is funnily `switch_account`.

WofWca: Ah right. Didn't pay attention to the wording.

Oh nice. I missed that string, since I was looking in the vicinity of `switch_between_chats`.

WofWca:
Changed the string.
I also moved the section to below "Switch between chats" in the shortcuts dialog, that feels better. See screenshot in the MR description.

WofWca:
I have a concern with the fact that with this it's possible to switch accounts while a dialog is open. Can cause bugs with our reliance on `window.__selectedAccountId`.
I think we should fix that first.

WofWca:
I submitted https://github.com/deltachat/deltachat-desktop/pull/6413.
When that is merged, this will need to be rebased, and adding tests would be nice.

WofWca:
Rebased, added tests.

PR #6225, opened 2026-04-12
fix: add missing aria-labels for various buttons

### LINKED ISSUE #4611: Aria-Label missing for various buttons

Deepak8858:
This PR addresses some of the missing aria-labels identified in #4611.

### Changes:
- Added `alt` label to `InlineVerifiedIcon` using the `verified_by_unknown` (Introduced) translation.
- Updated `JumpDownButton` to use different `aria-label` strings depending on whether it is in "scroll to bottom" or "jump back" mode.

Fixes #4611

### DIFF
--- a/packages/frontend/src/components/VerifiedIcon.tsx
+++ b/packages/frontend/src/components/VerifiedIcon.tsx
@@ -1,7 +1,14 @@
 import React, { CSSProperties } from 'react'
+import useTranslationFunction from '../hooks/useTranslationFunction'
 
 export function InlineVerifiedIcon({ style }: { style?: CSSProperties }) {
+  const tx = useTranslationFunction()
   return (
-    <img className='verified-icon' src='./images/verified.svg' style={style} />
+    <img
+      className='verified-icon'
+      src='./images/verified.svg'
+      style={style}
+      alt={tx('verified_by_unknown')}
+    />
   )
 }
--- a/packages/frontend/src/components/message/MessageList.tsx
+++ b/packages/frontend/src/components/message/MessageList.tsx
@@ -1071,7 +1071,7 @@ function JumpDownButton({
           }}
           // Technically this is not always "to bottom",
           // but perhaps it's good enough.
-          aria-label={tx('menu_scroll_to_bottom')}
+          aria-label={stackIsEmpty ? tx('menu_scroll_to_bottom') : tx('back')}
         >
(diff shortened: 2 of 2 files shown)


packages/frontend/src/components/message/MessageList.tsx
```diff
@@ -1071,7 +1071,7 @@ function JumpDownButton({
           }}
           // Technically this is not always "to bottom",
```
nicodh: The comment needs to be updated with this PR

packages/frontend/src/components/VerifiedIcon.tsx
```diff
@@ -1,7 +1,14 @@
...
 
 export function InlineVerifiedIcon({ style }: { style?: CSSProperties }) {
+  const tx = useTranslationFunction()
   return (
-    <img className='verified-icon' src='./images/verified.svg' style={style} />
+    <img
+      className='verified-icon'
+      src='./images/verified.svg'
+      style={style}
+      alt={tx('verified_by_unknown')}
```
nicodh, line 11: We should be aware that  the translation key is not translated literally (just "introduced" instead of "introduced by unknown") and should be adapted since otherwise it could happen that the alt says "Introduced by unknown" while the Text says "Verified by user x".
But that is not in the scope of this PR

packages/frontend/src/components/message/MessageList.tsx
```diff
@@ -1071,7 +1071,7 @@ function JumpDownButton({
...
           // but perhaps it's good enough.
```
nicodh, line 1073:
```suggestion
```
Comment is not valid anymore

nicodh (changes requested):
Reviewed by reading & testing
Thanks for the contribution!

If the outdated comment is removed it's fine to be merged!
Note: as mentioned in the first PR description this fixes not the whole issue so we have to reopen the issue after merging.

nicodh:
After some internal discussion: is it even helpful to add an alt attribute to the icon image? The text it contains follows the icon anyway so maybe it makes more sense to set aria-hidden="true" since the icon is more a visual addon while the text is sufficient as information.

nicodh:
closed in favour of #6266

PR #6226, opened 2026-04-12
fix: don't close dialog when closing menu with Esc

WofWca:
Affected places:
- "New Chat" dialog
- "View Contact" dialog, the three-dot menu
- Maybe more

What seems to have been happening is that pressing Escape
would trigger `closeCallback` which would close the context menu dialog.
Which would then cause the keydown event to be handled
by the currently open dialog.
What this commit does is it makes the context menu dialog
only close once using the native behavior of `<dialog>` elements.

The manual Escape handling has been written
before we started utilizing the `<dialog>` element for the context menu.
Now we can simplify things.


nicodh (approved):
Reviewed by reading & testing

PR #6227, opened 2026-04-12
refactor: remove scroll prevention code for menu

WofWca:
Basically this reverts 07bc1d956d4f1f6939e2e4fa2ca8ddb58a0c1e94.

In 0e0d0b80d598802ebb972c8d6c05b593bd896f24
(https://github.com/deltachat/deltachat-desktop/pull/4006)
we started using `<dialog>` for the context menu,
so now we don't need to worry about scrolling prevention,
because when the dialog is open, the rest of the page content
becomes inert.


nicodh (approved):
Reviewed by reading

PR #6228, opened 2026-04-12
refactor: remove `window.__refetchChatlist`

WofWca:
Follow-up to 5c770b9689113efb10a2ce7c12a08e24dacb6061
(https://github.com/deltachat/deltachat-desktop/pull/3268).
That MR removed the assignment to the variable,
making it always `undefined`,
but apparently mistakenly didn't remove the variable altogether.

Also remove the comment which apparently referenced the fact
that we had to reload the chat list when doing certain actions,
such as deleting a chat.
It was introduced in e969ce4bca25a10980e90ca54df437ed6018518c
(https://github.com/deltachat/deltachat-desktop/pull/2874),
before we had the core event.

### DIFF
--- a/packages/frontend/src/backend-com.ts
+++ b/packages/frontend/src/backend-com.ts
@@ -25,10 +25,8 @@ export namespace EffectfulBackendActions {
     runtime.deleteWebxdcAccountData(account_id)
   }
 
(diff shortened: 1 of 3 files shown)


nicodh (approved):
Reviewed by reading

PR #6229, opened 2026-04-12
chore: `npx playwright` -> `pnpm exec playwright`

WofWca:
I believe `npx playwright` will use a version
different from the one installed with `pnpm install`.
This can cause Playwright to install browser versions
that are different from the ones that are required
by the `pnpm install` version of Playwright,
so in the end one will have to install both.

I suspect that this could also reduce the difference between CI testing env and local testing env.


nicodh (approved):
Reviewed by reading - tests are fine

PR #6230, opened 2026-04-12
chore: don't require dev to copy `_env` file

WofWca:
So that it's easier for them to set up the project to run E2E tests.


nicodh:
I don't think it's worth to change here anything.
Copy a _env file to .env is a very common practice and I think it makes sense to make the user aware that the tests rely on a specific (external) server.

WofWca:
> Copy a _env file to .env is a very common practice

Sure it is, but does stuff have to not work without this ritual being performed?

> it makes sense to make the user aware that the tests rely on a specific (external) server

This sure is nice, but do we want to do it at a cost of making the dev experience more painful?
I mean we also tell the devs to do `pnpm exec playwright install --with-deps`, but we don't make them specify the server from which the browsers are downloaded.

nicodh:
I think it is to be expected that running an install command always involves external servers. And I don't consider copy/rename a file a "painful" experience. It's just a way to gently force the developer to have a look at the .env file

WofWca:
We should revert https://github.com/deltachat/deltachat-desktop/commit/37351dea37e2860e3c60ccf6afcd4a63f018e6a2 (https://github.com/deltachat/deltachat-desktop/pull/5972) then.

PR #6231, opened 2026-04-13
Sticker folder rewrite

nicodh:
resolves #6194

### DIFF
--- a/packages/frontend/src/components/composer/EmojiAndStickerPicker.tsx
+++ b/packages/frontend/src/components/composer/EmojiAndStickerPicker.tsx
@@ -11,7 +11,6 @@ import { BackendRemote } from '../../backend-com'
 import { selectedAccountId } from '../../ScreenController'
 import { runtime } from '@deltachat-desktop/runtime-interface'
 import EmojiPicker from '../EmojiPicker'
-import Button from '../Button'
 import useTranslationFunction from '../../hooks/useTranslationFunction'
 import useMessage from '../../hooks/chat/useMessage'
 
@@ -115,12 +114,6 @@ export const StickerPicker = ({
 }) => {
   const tx = useTranslationFunction()
 
-  const onOpenStickerFolder = async () => {
-    const folder =
-      await BackendRemote.rpc.miscGetStickerFolder(selectedAccountId())
-    runtime.openPath(folder)
-  }
-
   const stickerPackNames = Object.keys(stickers)
 
   return (
@@ -130,35 +123,20 @@ export const StickerPicker = ({
       aria-labelledby={labelledBy}
       className='sticker-picker'
     >
-      {stickerPackNames.length > 0 ? (
-        <>
-          <div className='sticker-container'>
-            {stickerPackNames.map(name => (
-              <DisplayedStickerPack
-                chatId={chatId}
-                key={name}
-                stickerPackName={name}
-                stickerPackImages={stickers[name]}
-                setShowEmojiPicker={setShowEmojiPicker}
-              />
-            ))}
-          </div>
-          <div className='sticker-actions-container'>
-            <Button onClick={onOpenStickerFolder}>
-              {tx('open_sticker_folder')}
-            </Button>
-          </div>
-        </>
-      ) : (
-        <div className='sticker-container'>
-          <div className='no-stickers'>
-            <p className='description'>{tx('add_stickers_instructions')}</p>
-            <Button onClick={onOpenStickerFolder}>
-              {tx('open_sticker_folder')}
-            </Button>
-          </div>
-        </div>
-      )}
+      <div className='sticker-container'>
+        {stickerPackNames.map(name => (
+          <DisplayedStickerPack
+            chatId={chatId}
+            key={name}
+            stickerPackName={name}
+            stickerPackImages={stickers[name]}
+            setShowEmojiPicker={setShowEmojiPicker}
+          />
+        ))}
+      </div>
+      <div className='sticker-hint'>
+        <p className='description'>{tx('sticker_picker_empty_hint')}</p>
+      </div>
     </div>
   )
 }
--- a/packages/frontend/src/components/composer/EmojiAndStickerPicker.tsx
+++ b/packages/frontend/src/components/composer/EmojiAndStickerPicker.tsx
@@ -106,7 +106,7 @@ function StickersListItem(props: {
         {
           label: tx('menu_copy_image_to_clipboard'),
           action: () => {
-            runtime.writeClipboardImage(filePath.replace('file://', ''))
+            runtime.writeClipboardImage(filePath)
           },
         },
         { type: 'separator' },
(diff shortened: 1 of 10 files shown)


packages/target-electron/src/ipc.ts
```diff
@@ -514,3 +516,16 @@ async function removeTempFile(path: string) {
   }
   await rm(path)
 }
+
+async function deleteSticker(stickerPath: string) {
+  const resolved = normalize(stickerPath)
+  const accountsPath = getAccountsPath()
+  if (!resolved.startsWith(accountsPath + sep)) {
```
WofWca:
```suggestion
  if (
    !resolved.startsWith(accountsPath + sep) ||
    !resolved.includes('stickers')
  ) {
```

Ideally we should parse the path and check it by its components properly, or utilize `miscGetStickerFolder(accountId)`. But this suggestion makes it way safer already.

packages/frontend/src/components/composer/EmojiAndStickerPicker.tsx
```diff
@@ -79,16 +85,56 @@ const DisplayedStickerPack = ({
...
+          },
+        },
+        { type: 'separator' },
+        {
+          label: tx('delete'),
+          danger: true,
+          action: async () => {
+            const confirmed = await openConfirmationDialog({
+              message: tx('ask_delete_sticker'),
+              isConfirmDanger: true,
```
WofWca, line 120:
```suggestion
              confirmLabel: tx('delete'),
              isConfirmDanger: true,
```

In line with Android, and is more clear.

packages/frontend/src/components/composer/EmojiAndStickerPicker.tsx
```diff
@@ -130,35 +172,21 @@ export const StickerPicker = ({
...
+            stickerPackName={name}
+            stickerPackImages={stickers[name]}
+            setShowEmojiPicker={setShowEmojiPicker}
+            onStickerDeleted={onStickerDeleted}
+          />
+        ))}
+      </div>
+      <div className='sticker-hint'>
+        <p className='description'>{tx('sticker_picker_empty_hint')}</p>
+      </div>
```
WofWca: Why has the `tickerPackNames.length > 0` logic been removed? I don't think it makes sense to always display the hint. Neither does Android display it when the list is not empty.

Image

packages/frontend/src/components/composer/EmojiAndStickerPicker.tsx
```diff
@@ -79,16 +85,56 @@ const DisplayedStickerPack = ({
...
+    })
+  }
+
   return (
     <button
       type='button'
       ref={ref}
       className={'sticker ' + rovingTabindex.className}
       onClick={onClick}
+      onContextMenu={onContextMenu}
```
WofWca, line 138:
```suggestion
      onContextMenu={onContextMenu}
      aria-haspopup='menu'
```

We do it (amost?) always when we have a custom context menu.

packages/runtime/runtime.ts
```diff
@@ -92,6 +92,9 @@ export interface Runtime {
   downloadFile(pathToSource: string, filename: string): Promise<void>
   transformBlobURL(blob: string): string
   transformStickerURL(sticker_path: string): string
+  /** Moves a sticker file to the system trash. The path must be an absolute
+   * filesystem path (without `file://`) inside the user's sticker folder. */
```
```suggestion
   * filesystem path inside the user's sticker folder. */
```

It's implied I'd say.

packages/frontend/src/components/composer/EmojiAndStickerPicker.tsx
```diff
@@ -79,16 +85,56 @@ const DisplayedStickerPack = ({
...
+  const onContextMenu = (ev: React.MouseEvent) => {
+    ev.preventDefault()
+    openContextMenu({
+      x: ev.clientX,
+      y: ev.clientY,
+      items: [
+        {
+          label: tx('menu_copy_image_to_clipboard'),
+          action: () => {
+            runtime.writeClipboardImage(filePath.replace('file://', ''))
```
WofWca:
```suggestion
            runtime.writeClipboardImage(filePath)
```

It's a file path already anyway. On top it's not enough to strip off `file://` to convert a URL to a path for one because it could include percent-encoded characters (e.g. space).
I tested it and it works.

packages/frontend/src/components/composer/EmojiAndStickerPicker.tsx
```diff
@@ -79,16 +85,56 @@ const DisplayedStickerPack = ({
...
+        {
+          label: tx('delete'),
+          danger: true,
+          action: async () => {
+            const confirmed = await openConfirmationDialog({
+              message: tx('ask_delete_sticker'),
+              isConfirmDanger: true,
+            })
+            if (confirmed) {
+              await runtime.deleteSticker(filePath.replace('file://', ''))
```
WofWca:
```suggestion
              await runtime.deleteSticker(filePath)
```

See comment above, https://github.com/deltachat/deltachat-desktop/pull/6231/changes#r3147052946

packages/frontend/src/components/composer/EmojiAndStickerPicker.tsx
```diff
@@ -130,35 +172,21 @@ export const StickerPicker = ({
...
+            key={name}
+            stickerPackName={name}
+            stickerPackImages={stickers[name]}
+            setShowEmojiPicker={setShowEmojiPicker}
+            onStickerDeleted={onStickerDeleted}
+          />
+        ))}
+      </div>
+      <div className='sticker-hint'>
+        <p className='description'>{tx('sticker_picker_empty_hint')}</p>
```
WofWca:
```suggestion
        <p>{tx('sticker_picker_empty_hint')}</p>
```

Unused class name.

WofWca (approved):
Works overall, seems to close issue.

Looks like the change from `click` to `mousedown` will not break that behavior for touch screens.
I tested keyboard accessibility and it remains fine.

The "copy image to clipboard" is not in the issue, and I'm not sure if it makes sense for stickers which are usually animated (and copying an animated image seems to only copy one frame of it to the clipboard), but it might be useful.

FYI it's possible to add the same sticker multiple times (if it's in a different message), but this bug is also present on Android, and this perhaps should be solved in Core anyway.

Interesting that that `trashItem` is used instead of `rm`, I think it's nice.

packages/frontend/src/components/composer/EmojiAndStickerPicker.tsx
nicodh: It still makes sense to show the hint to remind the user how it works, but if Android doesn't show it let's remove it in desktop too

WofWca: OK yep "doesn't make sense" is too strong of a wording, but I think this takes space unnecessarily. I don't think that it's hard to doscover this "tap to add" functionality anyway. This is the first thing you'd do when you see a sticker that you want to add.

PR #6234, opened 2026-04-13
chore: prepare release 2.49.0

nicodh:


WofWca (approved):
Just a version bump + a changelog update.
Other clients are preparing releases with this Core version, so should we I guess.
I didn't look at the changelog too closely.

PR #6235, opened 2026-04-13
test: don't require TLS cert to run tests

WofWca:
- **refactor: rename `sslserver` to `server`**
- **refactor: `target-browser`: support generic URLs**
- **refactor: tests: factor out `loadEnv()`**
- **test: don't require TLS cert to run tests**
- **test: auto-create `data` dir**

This removes the necessity for developers to install Rust
and `rustls-cert-gen`,
by adding the `DC_FRONTEND_NO_TLS` env var,
which makes the web server serve plain HTTP instead of HTTPS.

When this and https://github.com/deltachat/deltachat-desktop/pull/6230 get merged all you'll have to do to run tests will be `pnpm install` and `pnpm exec playwright install --with-deps`.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -16,64 +16,6 @@ on:
       - 'README_ASSETS/**'
       - 'README.md'
 
-env: # dummy certificates for testing
-  PRIVATE_CERTIFICATE_KEY: |
-    -----BEGIN PRIVATE KEY-----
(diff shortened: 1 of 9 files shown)


nicodh (approved):
Reviewed by reading & testing

This is a useful improvement!

TODO: the README in target-browser should be updated and mention that it is now optional to install rust generate self signed certificates

WofWca:
> the README in target-browser should be updated

I have considered that and thought that target-browser is more user oriented, so maybe It's not good to recommend disabling TLS.
Maybe it actually is, but either way I decided not to do it in this MR at least. Here I only updated the E2E test instructions.

PR #6241, opened 2026-04-14
fix: add better aria-label to close button

nicodh:
related to #4611


WofWca (commented):
I'm not sure if this is really an improvement if we're not translating the strings. People can at least sort of infer the meaning of "close" given the respective section label, as a sighted user would understand the meaning of a cross. But if they don't speak English the button is gonna be a mystery to them.

nicodh:
>  if we're not translating the strings

why do you assume we won't translate the strings?

WofWca (approved):
Because you just added the strings to `_untranslated_en.json`, and strings tend to hang around there indefinitely without getting translated.
If that's not the case then looks good.

The CRLF fix is also good.

nicodh:
> Because you just added the strings to `_untranslated_en.json`, 


But there is no other way to add strings to translations. The actual translations happens outside of this project that's why I didn't understand your comment.

Issue #6243, opened 2026-04-14
App icons in top bar: Transparency and hover text


breisa:
- Operating System (Linux): Arch
- Delta Chat Version: `v2.35.0`
- Expected behavior: 
  App icons in the bar above the chat should allow transparency, as in all other places where they are rendered in the app. On hovering, the full, potentially customized, app title/name should be shown instead of just the name of the app. 
- Actual behavior: Transparent parts are displayed as white and the hover text shows only the short app name.
- Steps to reproduce the problem:
  Send and app like [BetterList](https://github.com/breisa/betterlist/releases/download/v1.0.3/betterlist.xdc) to chat. Change the default title. Observe the issues.
- Screenshots:
  <img width="209" height="148" alt="Image" src="https://github.com/user-attachments/assets/489012e3-2a80-48fb-94c3-a7ceb4d145a7" />
  <img width="387" height="219" alt="Image" src="https://github.com/user-attachments/assets/ad47f2f5-3be5-4e56-8aa0-5fb0bcd9fa92" />
  <img width="248" height="311" alt="Image" src="https://github.com/user-attachments/assets/cac6d918-6380-444b-97ae-71708eb913e3" />

- Logs: -

I would appreciate it if you consider fixing these, thank you.

WofWca:
Hi!

Transparency in the icon is discouraged by the WebXDC spec: https://webxdc.org/docs/spec/format.html#icon-files, so we don't try to support that in DC Desktop.

Regarding the title: good idea, I will check if it's easy to do.

breisa:
Thanks, I did overlook that when I checked the spec. For the sake of consistency, it might make sense to enable transparency in the app bar anyway, especially since even non-transparent app icons seem to render with white edges because of this:

<img width="121" height="90" alt="Image" src="https://github.com/user-attachments/assets/e04c6f18-77c2-4bbf-855c-590058cc9cb0" />

r10s:
the white edges are a bug then. probably the background need to be have same corner radius than icon

do not add transparency.

it is discouraged to not make icon look differently or bad on some platform or even within the same platform. background should be white, as is currently. if that looks bad, it looks same bad everywhere and app author can iterate

PR #6244, opened 2026-04-14
style: fix "recent apps" white borders

WofWca:
The `background-color` line has been introduced in
3bc21041bfbafea2da9c297adcdffa28528e8480
(https://github.com/deltachat/deltachat-desktop/pull/5149).
The apparent intent of it is to replace transparency with white color.
However, this causes issues with `border-radius`.
The browser tried to blend the icon color with white,
and not with the color of the chat title bar,
resulting in annoying white "glow".

Note that we don't have this `background-color` thing
in the gallery or the message list either.

See https://github.com/deltachat/deltachat-desktop/issues/6243#issuecomment-4243735849.

### DIFF
--- a/packages/frontend/src/components/screens/MainScreen/styles.module.scss
+++ b/packages/frontend/src/components/screens/MainScreen/styles.module.scss
@@ -149,7 +149,6 @@
     height: 25px;
     cursor: pointer;
     border-radius: 3px;
-    background-color: #fff;
   }
 }
 


nicodh (approved):

r10s:
so, the glow cannot be fixed? 

~~if in doubt, corner-radius can be removed, it does not make much of a difference at that size.~~
EDIT: corner-radius is not used, just bg-color set results in glow


> Note that we don't have this background-color thing
in the gallery or the message list either.

this is a bug as well then, i indeed see some apps using transparency in their icons, which does not fit nicely at other places, eg by having inconsistent look. incentive should be to nudge app devs to not use transparency, so host apps can rely on that, add shadow, whatnot. 

app devs often do not test all platforms, therefore the strictness here, forcing a unified look

bg-transparency is unwanted, i commented also in the issue. icon consistency will degrade otherwise

r10s:
otoh, maybe more pragmatic approach is to let transpareny render as is. i just checked ios, it is similar as desktop

and having the unwanted glow for icons following the spec is questionable indeed

but i general, we should not to anything to make transparent icons look good. they are still discouraged, and it is still better to not have then, for a unified look

r10s (approved):

WofWca:
> so, the glow cannot be fixed?

Probably can, at least by inserting a smaller absolutely positioned element behind the image.
Or by waiting for Chromium to fix this (if it's a bug).

PR #6249, opened 2026-04-15
refactor: `useRef` -> `useMemo` for Observer

WofWca:
I believe this doens't affect behavior,
because the `onUnreadMessageInView` is already re-created
when its dependencies, namely `chat.id` and `accountId`, get changed,
and that's because we have the `MessageList` behind `key={chatId}`.

This should make it harder to introduce such bugs
where the observer closure captures a value which might become outdated.
Also simplifies things a bit, makes the observer non-nullable.

I have tested that the observer only gets created once after selecting a chat,
and does not get re-created when scrolling or jumping between messages
or placing reactions on messages.

### DIFF
--- a/packages/frontend/src/components/message/MessageList.tsx
+++ b/packages/frontend/src/components/message/MessageList.tsx
@@ -2,7 +2,6 @@ import React, {
   useRef,
   useCallback,
   useLayoutEffect,
-  RefObject,
   useEffect,
   useState,
   useMemo,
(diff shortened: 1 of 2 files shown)


nicodh (approved):
Reviewed by reading

While I agree that this is an improvement code wise I want to point out that there is always a certain risk to introduce new bugs with a change that does not resolve a bug or problem. As long as there are bugs to resolve they should be prioritized to mere refactorings except when needed to introduce new features etc.

PR #6250, opened 2026-04-15
build: fix extraResources plist path

nicodh:
- add some build related comments

(and removes CRLF line breaks from afterPack hook file)


packages/target-electron/build/afterPackHook.mjs
```diff
@@ -1,243 +1,232 @@
...
+import { env } from 'process'
+
+/**
+ * Electron-Builder afterPack Hook
+ *
+ * This file is an afterPack lifecycle hook for electron-builder. It runs automatically
+ * after the application has been packaged into an .app (macOS), executable directory
+ * (Windows/Linux), but before final distribution files (.dmg, .exe, .AppImage) are created.
+ *
+ * Configured in gen-electron-builder-config.js under "build.afterPack" property.
```
WofWca, line 17: Redundant paragraph? One can just look at the name, or search for where the file is referenced.

packages/target-electron/build/afterSignHook.cjs
```diff
@@ -1,16 +1,45 @@
...
+ *
+ * Notarization is Apple's automated security scanning process. After you code-sign your
+ * macOS app, you upload it to Apple's notary service, which:
+ * - Scans the app for malicious content
+ * - Checks that it's properly signed
+ * - Issues a "ticket" if the app passes
+ *
+ * Apps downloaded from the internet MUST be notarized or users will see "malicious software"
+ * warnings and may not be able to open the app at all. Notarization proves the app comes
+ * from an identified developer and hasn't been tampered with.
```
WofWca, line 27: Is this the place for describing what notarization is?

packages/target-electron/build/afterSignHook.cjs
```diff
@@ -1,16 +1,45 @@
...
+ * (.dmg, .pkg) are created.
+ *
+ * Configured in gen-electron-builder-config.js under "build.afterSign" property.
+ *
+ * Build Process Timeline (macOS):
+ * 1. electron-builder packages the app (afterPack hook runs)
+ * 2. App is code-signed with developer certificate
+ * 3. afterSign hook runs (this file)
+ * 4. App is notarized by Apple's servers (happens in this hook)
+ * 5. electron-builder creates final .dmg/.pkg installers
```
WofWca, line 17: Did you write these comments yourself? Are they useful or are they just describing how Electron-builder works?

WofWca (approved):
Not really my area of expertise.
According to the name of the MR as far as I can see the only relevant change is the `plist` name change.
The other stuff: formatting, removing old commented out code, adding comments, looks fine.

packages/target-electron/build/afterPackHook.mjs
nicodh: Of course you can search for it, but the build process happen inside the desktop-builder so it is not too easy which parts are configured/used inside the builder and which inside desktop. I just had to deal with the build process and I decided more documentation makes sense especially since we have to look into that part only rarely

packages/target-electron/build/afterSignHook.cjs
nicodh: It's the place where notarizing takes place so why not?

packages/target-electron/build/afterSignHook.cjs
nicodh: > Are they useful or are they just describing how Electron-builder works?

They are useful because they describes how electron-builder works. I just had to deal with signing errors and had to go through all the steps to find the place where what fails and that's why I decided next time I woud like to have it documented

PR #6255, opened 2026-04-15
fix: chat list switching between archive by itself

### LINKED ISSUE #6253: incoming messages alter selection in chatlist
since recently, incoming messages may alter the selection of the chatlist.

eg. 
- have a chat from the archive selected, eg. TESTCHAT
- navigate to another chat, by hitting the "back button" above the archive chats
- a new message arrives in TESTCHAT

-> chatlist changes back archive unexpectedly

this is unexpected.

the app should not play around with the selection the user did in the chatlist, esp. when it is only moments before. it can happen any time, however, happens more often when downloading messages.

iirc, there was some recent change in 2.48 or so, but i cannot find the exact chat

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/6253.

This bug has been introduced in
0bddb68a127f8626d81e8215c899a0c7bd94e8ce
(https://github.com/deltachat/deltachat-desktop/pull/5824).
The error is that `chatNoLinger` is `undefined` while
the chat is getting reloaded (`chatFetch.refresh`),
so `useHasChanged2()` returns `true` when the chat finishes refreshing,
whereas we only want to perform the "is archived?" check
on the initial load of the chat.

I have confirmed that this fixes the issue.

Edit: this could also be related to the archive switching bug that @ralphtheninja reported (see also https://github.com/deltachat/deltachat-desktop/issues/5993)

### DIFF
--- a/packages/frontend/src/contexts/ChatContext.tsx
+++ b/packages/frontend/src/contexts/ChatContext.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useState } from 'react'
+import React, { useCallback, useEffect, useRef, useState } from 'react'
 
 import { ActionEmitter, KeybindAction } from '../keybindings'
 import { markChatAsSeen, saveLastChatId } from '../backend/chat'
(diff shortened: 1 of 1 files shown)


packages/frontend/src/contexts/ChatContext.tsx
```diff
@@ -166,7 +166,17 @@ export const ChatProvider = ({
...
+  if (useHasChanged2(chatId) && chatId != undefined) {
+    // eslint-disable-next-line react-hooks/refs
+    isArchivedChecked.current = false
+  }
+  if (
+    // eslint-disable-next-line react-hooks/refs
+    !isArchivedChecked.current &&
+    chatNoLinger != undefined
+  ) {
+    isArchivedChecked.current = true
```
nicodh: Why would something like this  not work? I tested and it fixed the issue


```suggestion
  const lastArchivedCheckChatId = useRef<number | undefined>(undefined)
  if (
    chatNoLinger != undefined &&
    // eslint-disable-next-line react-hooks/refs
    chatId !== lastArchivedCheckChatId.current
  ) {
    // eslint-disable-next-line react-hooks/refs
    lastArchivedCheckChatId.current = chatId
```

WofWca: At a glance looks like it could work, but how is it better?

nicodh: Simpler to read, no need for useHasChanged2

nicodh (changes requested):
See inline comment

packages/frontend/src/contexts/ChatContext.tsx
WofWca: Hm, maybe maybe.

nicodh (approved):

PR #6257, opened 2026-04-15
refactor: replace `findIndex()` with `some()`

WofWca:


r10s:
i am wondering about this and comparable PR:

this seem not to solve any current issue. but has the risk of introducing some. one would need to review carefully. etc. usually we do refactorings only when code changes are needed in that area for other reasons.

in general, these unrelated refactorings are not wanted. and in special, not during isssue/pr cleanup phase we're currently aiming for.

WofWca:
OK.
I thought that cleanup was also about code cleanup and not only about closing issues.

nicodh (approved):

PR #6264, opened 2026-04-16
fix: refocus input when send message button is pressed

account1009:


packages/frontend/src/components/composer/Composer.tsx
```diff
@@ -281,6 +281,9 @@ const Composer = forwardRef<
             // TODO fix: hypothetically by this point the user
             // could have started typing (and even have sent!)
             // a new message already, so this would override it on the backend.
+            setTimeout(() => {
+              regularMessageInputRef.current?.focus()
+            })
```
WofWca: The comment above applies to `removeDraft`, but now it looks like it applies to the `setTimeout`.

WofWca (commented):
Thanks!

Though I think it makes more sense to have the code in the "click" handler of the button somewhere. More narrowly-scoped, and there is not gonna be a delay in focusing.

account1009:
@WofWca I've moved the focus at the beginning of composerSendMessage to reduce lag. Do you think that the suggestion is fine now?

WofWca (commented):
I would prefer to have the code even closer to the button itself. Currently this code would get executed also on Ctrl + Enter, and it's not obvious what the code even does.
Plus it's not executed when editing a message.

I can finish this myself if you'd like.

account1009:
I've not touched the part related to the edit because #5713 already added the fix for edit mode: 
    // Focus the regular message input after editing
    setTimeout(() => {
      regularMessageInputRef.current?.focus()
    })
do you think that we have to remove that fix by @trufae and add something similar but before the computation in both cases (edit, normal message) ?

WofWca:
Good point.
Although the difference of that code is that we actually _want_ to run it even on control enter.

OK I guess then it doesn't matter that much. I think it's good enough to just add a comment saying what this code is for (i.e. To handle touch / mouse click).

WofWca:
> this code

I mean the code added in this PR.

account1009:
okok @WofWca , I've added a comment as you suggested. I hope that now the fix is fine. If anything else is missing feel free to change it as you prefer

packages/frontend/src/components/composer/Composer.tsx
```diff
@@ -222,14 +222,17 @@ const Composer = forwardRef<
     messageEditing.isEditingModeActive || draftIsLoading
       ? null
       : async () => {
+          // Focus message input to handle touch / mouse click
```
WofWca:
```suggestion
          // Focus message input in case the message was sent
          // with the "Send" button touch / mouse click and not Ctrl + Enter.
```

WofWca (commented):
Thanks for bearing with me!

I have tested this for a bit, also with a screen reader. I haven't found any downsides.

If you still have some patience and want to earn bonus internet points, you can add tests for this. I think somewhere in [this file](https://github.com/deltachat/deltachat-desktop/blob/31b687d11fa8d0d37d1dd5307fdccd783d2a68c0/packages/e2e-tests/tests/composer.spec.ts) would fit.

account1009:
> you can add tests for this

I've added tests with the last commit. Let me know if everything is fine now or propose improvements if anything else is missing

account1009:
I close this because it is derived my main branch so I can't propose other pull requests for other issues.
Anyway this PR looks stucked I don't know why

WofWca:
I'll come back to this, don't worry.

account1009:
I had to destroy my forked main branch because I wanted it synced with deltachat/deltachat-desktop so I've recreated another PR this time from a branch. The fact is that I was providing patches instead of PR for other issues because I made the mistake to propose this PR from the main branch. Sorry

WofWca:
That's fine, we'll figure it out 🙂

PR #6266, opened 2026-04-17
fix: dynamic aria-label for jump scroll button

nicodh:


WofWca (commented):
I don't think that this is good. "Back" could be interpreted in lots of ways in this context. With "scroll to bottom" it's at least clear that this only has to do with scrolling, and, as the comment says, it's good enough, despite the fact that sometimes it's not really all the way to the bottom.

nicodh:
Idk we could say "back to last position" instead although I would say that is how "back"  is interpreted anyway here - "scroll to bottom" in most scenarios scrolls to the very bottom so reading a different label after a jump tp message would help to distinguish

nicodh:
What I don't understand: the issue is from 7th of Februar2025 "jump down button in message list (both modes should have different strings (jump back, jump/scroll down completely))" and you commented it but didn't disagree. And now when the issue is fixed you argue it doesn't make sense.

It would be more efficient to discuss issues before they get fixed

WofWca:
You mean this one https://github.com/deltachat/deltachat-desktop/issues/4611. Well yes, I agree with the idea of having different labels, but it doesn't mean that I am OK with changing the label to any string whatsoever. "Back" is not good.

> "scroll to bottom" in most scenarios scrolls to the very bottom so reading a different label after a jump tp message would help to distinguish

Good point. Though I'm afraid it won't work here, because the button is not displayed unconditionally, and is missing most of the time. So I don't think it's that easily recognizable "oh it's the same button, just with a different label".
Imagine if it wasn't a round button with an arrow pointing down in the corner but instead just an unstyled `<button>` right above the composer section. That's how blind people see the app

And especially that nothing (besides scrolling, which does not in itself gets announced by screen readers) actually happens when one presses the button. This makes it almost practically impossible to figure out what the button does.

_locales/_untranslated_en.json
```diff
@@ -124,5 +124,8 @@
   },
   "remove_attachment": {
     "message": "Remove Attachment"
+  },
+  "back_to_last_message": {
+    "message": "Back to last message"
```
WofWca, line 129:
```suggestion
    "message": "Scroll back to last message"
```

More clarity for blind people who can't see the effect of this button.

WofWca (approved):

nicodh:
After some internal discussion we decided not to add this and try to make a cleanup with all accessibility issues: bundle the issues somehow together and try to find someone who is really affected and can give feedback what is useful and what not

Issue #6209, opened 2026-04-09
Ctrl + Space for multiselect doesn't work on Linux


WofWca:
- Operating System (Linux/Mac/Windows/iOS/Android): Linux
- Delta Chat Version: 2.48.0
- Expected behavior: Ctrl + Space toggles selection of chat list items
- Actual behavior:
- Steps to reproduce the problem:
- Screenshots:
- Logs:

https://github.com/deltachat/deltachat-desktop/blob/a4cbe0806328a79af1135ae464486c57bf40af48/packages/frontend/src/hooks/useMultiselect.ts#L233-L258

On Windows it works, but not on Linux.

I believe we have to also be listening for `kwydown`. But we can't remove `onClick` listener because we need it for mouse clicks. Then I guess we have to introduce some logic to only run the function once even if both of the events fired.

WofWca:
Oh, and I think adding the onkeydown event handler will make us better prepared for adding multi-select to the message list. Because in the message list, the selectable elements are not buttons. They don't receive the on-click event when you press spacebar.

nicodh:
Ctrl + Space hast no keybinding ?? What do you expect?

WofWca:
> Ctrl + Space hast no keybinding ?? 

Not sure what you mean?

> What do you expect?

As I wrote, Ctrl + Space should select / unselect the currently focused chat. Roughly as Shift + Space, but not contiguously.

But I just realized that we have tests for this. Maybe this issue is only reproducible in Electron? Or only on my OS?

https://github.com/deltachat/deltachat-desktop/blob/c343bff00c53553d2036cf21eca0cc47af74bc16/packages/e2e-tests/tests/chat-list-multiselect.spec.ts#L114-L129

WofWca:
> Maybe this issue is only reproducible in Electron? Or only on my OS?

Ahhh I see. It's because the tests use `ControlOrMeta`. Indeed on Linux it's also possible to select with "Windows (a.k.a. Meta) + Space". But Ctrl should also work I think.

WofWca:
Hmmmm I tried to change `ControlOrMeta` to `Control` in https://github.com/deltachat/deltachat-desktop/pull/6247, but the tests still pass, both in CI and on my Linux system, with `pnpm e2e`.
Also Playwright docs on https://playwright.dev/docs/api/class-keyboard#keyboard-press say:
> ControlOrMeta resolves to Control on Windows and Linux and to Meta on macOS.

[...]

PR #6268, opened 2026-04-17
feat: a way to select multiple messages

### LINKED ISSUE #2892: can´t delete several pics at once

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/2892.

image

In the end the functionality is quite conservative, one will not notice that we support multiselect until they Ctrl + Click or Shift + Click.

TODO:
- [x] Address TODOs.
- [x] Tests.
  https://github.com/deltachat/deltachat-desktop/pull/6361
- [x] Clean up the code.
- [x] Merge the base MRs (containing the first few commits) so that this one's easier to review.
  - https://github.com/deltachat/deltachat-desktop/pull/6284
  - https://github.com/deltachat/deltachat-desktop/pull/6285
- [ ] Consider adding a safeguard (maybe a confirmation dialog) if more than 100 or 1000 messages are selected.
  Telegram simply won't let you select more than 100 messages.
  https://github.com/deltachat/deltachat-desktop/issues/2892#issuecomment-4406316861
  Edit: ~~https://github.com/deltachat/deltachat-desktop/pull/6366~~
  Edit 2: https://github.com/deltachat/deltachat-desktop/pull/6369
- [ ] (maybe later) add more context menu items, such as "Save / Unsave Messages", "download attachment" and whatever else Android and iOS already support.
- [x] Consider squashing some commits, especially the ones made in follow-up MRs.

### DIFF
--- a/packages/frontend/scss/message/_message-list.scss
+++ b/packages/frontend/scss/message/_message-list.scss
@@ -184,5 +184,21 @@
         -moz-animation: none;
       }
     }
+
+    // If you Shift+Click some messages then their text
+    // also gets selected, which looks pretty ugly.
+    // However, some people might still want to select text this way,
+    // so let's keep a really faint color for them.
+    //
+    // We apply this to the whole `<ol>` and not `<li>`
+    // because otherwise daymarkers are not affected by this
+    // (because they're not selectable).
+    &:has(.multiselectable-message[aria-selected='true'])::selection {
+      background-color: color-mix(
+        in srgb,
+        var(--messageHightlightColor) 5%,
+        transparent
+      );
+    }
   }
 }
--- a/packages/frontend/scss/message/_message-list.scss
+++ b/packages/frontend/scss/message/_message-list.scss
@@ -174,6 +174,15 @@
           background-color: transparent;
         }
       }
+
+      &:has(.multiselectable-message[aria-selected='true']) {
+        background-color: var(--messageHightlightColor);
+
+        // Don't play the color fade animation in case the message is both
+        // `.highlight`ed and selected.
+        -webkit-animation: none;
+        -moz-animation: none;
+      }
     }
   }
 }
--- a/packages/frontend/scss/message/_message-list.scss
+++ b/packages/frontend/scss/message/_message-list.scss
@@ -105,6 +105,22 @@
     &::-webkit-scrollbar-track {
       background: transparent;
     }
+
+    // If you Shift+Click some messages then their text
+    // also gets selected, which looks pretty ugly.
+    // However, some people might still want to select text this way,
+    // so let's keep a really faint color for them.
+    //
+    // We apply this to the whole `<ol>` and not `<li>`
+    // because otherwise daymarkers are not affected by this
+    // (because they're not selectable).
+    &.multiselected-one-or-more::selection {
+      background-color: color-mix(
+        in srgb,
+        var(--messageHightlightColor) 5%,
+        transparent
+      );
+    }
   }
 
   ol {
--- a/packages/frontend/scss/message/_message-list.scss
+++ b/packages/frontend/scss/message/_message-list.scss
@@ -114,7 +114,12 @@
     // We apply this to the whole `<ol>` and not `<li>`
     // because otherwise daymarkers are not affected by this
     // (because they're not selectable).
-    &.multiselected-one-or-more::selection {
+    //
+    // Only apply this if two or more are selected because if the user
+    // tries to simply select text in a message with Shift + Click
+    // then that will select the message,
+    // in which case we should not faint the color.
+    &.multiselected-two-or-more::selection {
       background-color: color-mix(
         in srgb,
         var(--messageHightlightColor) 5%,
--- a/packages/frontend/src/components/dialogs/ConfirmDeleteMessage.tsx
+++ b/packages/frontend/src/components/dialogs/ConfirmDeleteMessage.tsx
@@ -1,4 +1,4 @@
-import React, { useContext } from 'react'
+import React, { useCallback, useContext } from 'react'
 
 import Dialog, {
   DialogBody,
@@ -13,10 +13,17 @@ import type { DialogProps } from '../../contexts/DialogContext'
 import { BackendRemote, Type } from '../../backend-com'
 import { C } from '@deltachat/jsonrpc-client'
 import { ScreenContext } from '../../contexts/ScreenContext'
+import { useFetch } from '../../hooks/useFetch'
+import { unknownErrorToString } from '@deltachat-desktop/shared/unknownErrorToString'
 
 export type Props = {
   accountId: number
-  msg: Type.Message
+  messageIds: Array<Type.Message['id']>
+  /**
+   * If some messages from {@linkcode messageIds} are already loaded,
+   * provide them here for better performance.
+   */
+  loadedMessages: { [id: Type.Message['id']]: Type.Message | undefined }
   chat: Type.FullChat
 } & DialogProps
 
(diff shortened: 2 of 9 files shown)


WofWca:
OK I think I got the basics of how this will work. Please give your feedback if you have any, otherwise I'm proceeding.

packages/frontend/scss/message/_message-list.scss
```diff
@@ -174,6 +174,29 @@
...
+      // TODO only display the selection when messagelist has focus?
+
+      // As of writing we only expect messages themselves
+      // to have `aria-selected` attribute,
+      // so it's OK to use such a generic selector.
+      // &:has(.roving-tabindex[aria-selected='true'])
+      &:has(.multiselectable-message[aria-selected='true'])
+      // &:has(.roving-tabindex:focus)
+      {
+        // TODO ahhh, we can't use `outline`, because it's used
```
nicodh: what does ahhh mean?

packages/frontend/src/components/chat/ChatListItemRow.tsx
```diff
@@ -126,6 +126,24 @@ export const ChatListItemRowChat = React.memo<{
   // eslint-disable-next-line react-hooks/refs
   onChatClick2Ref.current = onChatClick2
 
+  const multiselectOnKeyDown = multiselect?.onKeyDown
```
nicodh, line 129: Why is this needed? Is the chalist affected at all?

packages/frontend/scss/message/_message-list.scss
WofWca: ah
interjection
—used to express delight, relief, regret, or contempt

In this case regret.

packages/frontend/src/components/chat/ChatListItemRow.tsx
WofWca: Why is what needed? The `onKeyDown` listener in the chat list?
It's to fix https://github.com/deltachat/deltachat-desktop/issues/6209, see commit history.
I'll split it off into a separate MR for easier review when I'm sure that everything's working. Unless you want to see all these changes in one MR.

packages/frontend/src/hooks/useMultiselect.ts
```diff
@@ -242,26 +285,76 @@ export function useMultiselect<T>(
...
+        //
+        // My thinking was that we don't want to display messages as selected
+        // if only a single message is selected.
+        // But if we don't then it means that we also don't want
+        // to suddenly 2 messages get displayed as selected
+        // if the user clicked just one.
+        //
+        // Maybe just take a look at how WhatsApp does this? Telegram,
+        // maybe Signal (but those might be bad at this).
+        //
```
nicodh: Signal adds a new column with checkboxes  to the message list and a bar at the bottom showing the number of selected messages and 2 icons for deleting & forwarding
And while the column is visible you can click on messages to add them or use Ctrl + click etc.
But the click to open that column is Ctrl+click

I would say it is fine if you can start a multi select only with Ctrl + click and Shift + click only works if there is a selected message already

packages/frontend/src/components/chat/ChatListItemRow.tsx
WofWca: As promised, https://github.com/deltachat/deltachat-desktop/pull/6285 created and merged.

WofWca:
OK, I got the code and the functionality. The remaining TODOs are in the issue description, but they can be done in separate MRs IMO (possibly before this gets merged into main).
Ready for everyone's review.

packages/frontend/src/components/message/MessageList.tsx
```diff
@@ -952,19 +962,25 @@ function MessageLoadingError({
...
+    ref
+  )
 
   return (
     <div className='info-message' id={String(messageId.msg_id)}>
       <div
         ref={ref}
-        className={'bubble ' + rovingTabindex.className}
+        className={
+          'bubble multiselectable-message' + focusAndMultiselect.className
```
nicodh:
```suggestion
          'bubble multiselectable-message ' + focusAndMultiselect.className
```
I assume the classnames should not be concatenated?

packages/frontend/src/components/message/MessageList.tsx
```diff
@@ -977,19 +993,25 @@ function MessageLoading({
...
```
nicodh: same missing space here

packages/frontend/src/components/dialogs/ConfirmDeleteMessage.tsx
```diff
@@ -13,30 +13,83 @@ import type { DialogProps } from '../../contexts/DialogContext'
...
-  const { accountId, msg, chat, onClose } = props
+  const { accountId, chat, onClose, messageIds, loadedMessages } = props
 
-  const deleteForAllPossible =
-    msg.sender.id === C.DC_CONTACT_ID_SELF &&
-    !msg.isInfo &&
-    chat.canSend &&
-    !chat.isSelfTalk &&
-    chat.isEncrypted
+  const deleteForAllPossibleF = useFetch(
```
nicodh: If the last F stands for Fetch I would prefer to write it out

packages/frontend/src/components/dialogs/ConfirmDeleteMessage.tsx
```diff
@@ -13,30 +13,83 @@ import type { DialogProps } from '../../contexts/DialogContext'
...
 
 export type Props = {
   accountId: number
-  msg: Type.Message
+  messageIds: Array<Type.Message['id']>
+  /**
+   * If some messages from {@linkcode messageIds} are already loaded,
+   * provide them here for better performance.
+   */
+  loadedMessages: { [id: Type.Message['id']]: Type.Message | undefined }
```
nicodh, line 26: why is this plural while in fact the object only holds one message? (the one that opened the context menu)
Skip loading one message should be not noticable as "better performance" but ok, if we have it already I would use it. But mainly for the case where we have only one msg ID which is probably the most common use case.

Or is there a use case where we have multiple loaded messages here?

nicodh (commented):
Works as expected. Even the scenario that a message is deleted while being selected is covered! 👍

There is a small issue though. If you select an additional message with Shift + click it often also selects the text of the intermediate messages.
If that is easy to fix it would be nice.

Image

packages/frontend/src/components/dialogs/ConfirmDeleteMessage.tsx
WofWca: I thought about reducing this to "only one message" but I though that the code and the API don't actually get that much simpler with that.
The idea is also to keep an ability to provide the whole `messageCache` from `messagelist.ts` when / if it somehow becomes available, as a prop or through context.

> Or is there a use case where we have multiple loaded messages here?

No, there is no such case right now.

WofWca:
> If you select an additional message with Shift + click it often also selects the text of the intermediate messages.

I pushed a commit that makes the selected text color really faint.

nicodh (approved):

Note: the reactions menu of selected messages is still clickable and adds a reaction (but only for the message it's activated on) - but might be that's ok since it's not clear what would be expected

WofWca:
Rebased.

r10s:
very nice to get things moved forward here! thanks a lot for the PR:

some things during a first test:

- it seems to be okay to really need to tap a bubble, but not a row. however:

- [x] it is confusing how to "deselect everything": this should be possible by a simple tap outside somewhere, as usual. it is weird that one has to tap another bubble for that
  https://github.com/deltachat/deltachat-desktop/pull/6399

- [x] the number of selected messages should be shown in the title. that way, the user can track and esitimate the impact, esp. when accidentally holding "shift" - and not only the three messages on the screen are selected.

- an additional confirmation "> 100" is not needed  then

- [ ] when tapping the "..." beside a message (instead of right tap), everything gets visually deselected - but it seems still to work. "..." should not deselect messages.
  https://github.com/deltachat/deltachat-desktop/pull/6682

- [ ] the "emoji" button should be hidden in multi-select mode, it does not work, or it is unclear how it should work. just hide

- [x] "Escape" key should probably exit the multi-select mode
  https://github.com/deltachat/deltachat-desktop/pull/6403

r10s:
> the number of selected messages should be shown in the title. that way, the user can track and esitimate the impact, esp. when accidentally holding "shift" - and not only the three messages on the screen are selected.

this is how it should look like:

Screenshot 2026-05-11 at 16 31 37

removal of avatar/title/subtitle shows clearly that we're in another mode now, and puts focus on the selection. also, this is a similar place as android/iOS (ftr, chat multi-select needs that as well, but that is another issue)

r10s:
ftr: i am fine, to merge this PR (without #6366, however) , and to do subsequent improvements in follow ups

r10s:
another point:

- when tapping an images in multi-select mode, one wants to select that message, not to enlarge the image. for images without text, it is currently impossible to select them

WofWca:
> for images without text, it is currently impossible to select them

Not quite. One can tap a little to the side of the bubble, where the buttons are.
But yep, that's a problem.
I'm thinking of solving this by applying the [`inert`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/inert) attribute. But accessibility will need to be considered/
Edit: to address this I opened
- https://github.com/deltachat/deltachat-desktop/pull/6371.
- https://github.com/deltachat/deltachat-desktop/pull/6395.

WofWca:
From this point on I will stop directly modifying this MR and only do so through (approved) MRs, so that there is no need to re-review this MR.

WofWca:
Rebased, and also adjusted the profile creation part of tests, to be in line with changes to the main branch.

WofWca:
OK, we have almost-ready MRs for stuff that I consider important to finish this feature. I've written down the remaining TODOs in the relevant issue. I think it's a good time to merge.

PR #6273, opened 2026-04-17
refactoring: share log handler

### LINKED ISSUE #6233: DC hang after disk full, need app restart to recover
- Operating System (Linux/Mac/Windows/iOS/Android): Linux (Fedora Workstation 43, x86_64)
- Delta Chat Version: 2.48
- Expected behavior: DC should report "not enough permanent storage/disk space to store new messages, will recover once this issue is resolved". For existing, old messages it should be able to display these, no matter what disk free status ist.
- Actual behavior: DC threw an unexpected error. After freeing up disk space, the GUI did not recover - I could click several chat partners or groups on the left hand side, but the display on the right (chat content) did not update any longer at all. Closing the DC GUI app was not enough to fix this. I had to kill "invisible" processes, and then restart DC to get a working chat again. Only users above base level can find "invible processes" (not being shown in the task bar), thus is has a UX impact.
- Steps to reproduce the problem: Run a "wget -mnp ftp.pigwa.net" to mirror retro Atari ST 68k software collections on your machine. Have only 300 GB available space. The wget will fill it over night. Have DC runnning at the same time. In the morning it reports the unexpected error, and the GUI hangs.
- Screenshots:
- Logs:

nicodh:
~~maybe resolves #6233~~

Errors should anyway be handled


WofWca:
So how would listening to an event affect behavior? I mean I'm for better error messages but I doubt that this will help with <https://github.com/deltachat/deltachat-desktop/issues/6233>?

nicodh:
> So how would listening to an event affect behavior? I mean I'm for better error messages but I doubt that this will help with #6233?

Oops - the important lines were missing. Somehow the editor didn't save and I didn't check twice

packages/target-electron/src/log-handler.ts
```diff
@@ -50,8 +54,13 @@ export function createLogHandler() {
...
       )
-      if (stream.writable) {
-        stream.write(`${line.join('\t')}\n`)
+      if (stream.writable && !stream.destroyed) {
+        try {
+          stream.write(`${line.join('\t')}\n`)
+        } catch (_err) {
+          // Silently ignore write errors to prevent app freeze
+          // Error is already logged via stream error handler
+        }
```
WofWca: Documentation says that if an error occurs then a callback gets invoked, which I assume means that `write()` doesn't throw then. If so then this code doesn't do anything, and I think it's worth clarifying it in the comments. But just in case I think it makes sense to wrap log stuff into `try-catch`. But then it'd be better to wrap the whole function and not just this part.

https://nodejs.org/api/stream.html#class-streamwritable

packages/target-electron/src/log-handler.ts
```diff
@@ -50,8 +54,13 @@ export function createLogHandler() {
       line = line.concat(
         [stacktrace, ...args].map(value => JSON.stringify(value))
       )
-      if (stream.writable) {
-        stream.write(`${line.join('\t')}\n`)
+      if (stream.writable && !stream.destroyed) {
+        try {
+          stream.write(`${line.join('\t')}\n`)
+        } catch (_err) {
+          // Silently ignore write errors to prevent app freeze
```
WofWca:
```suggestion
          // Silently ignore write errors
          // so that the app can keeps running if logging fails
```

WofWca (approved):

nicodh:
I changed the error handling to a more appropriate way using drain and backpressure from write (suggested by AI and researched by me)
BUT:
when trying to test the error handling by making the logfile read only, it turns out, that this does NOT freeze the app at all. It just shows an error:

image

But the app keeps working. So our assumption that this could be the reason was wrong. Probably the error is caused when trying to write into the database or blobdir.

tbd: should we keep this PR? It supresses the error which is not an improvement, but isome refactorings could be useful (like sharing the log handler)

packages/shared/log-handler.ts
```diff
@@ -47,18 +58,24 @@ export function createLogHandler() {
...
+          try {
+            return JSON.stringify(value)
+          } catch {
+            return '[unserializable]'
+          }
         })
+      )
+      if (streamErrored || draining || !stream.writable || stream.destroyed) {
+        // Drop message: stream is broken, buffer is full, or
+        // stream was closed. Prevents unbounded memory growth
```
WofWca: We should write about dropped entries to the log, otherwise it might be getting logs that are very confusing.

WofWca: The default `highWaterMark` is 16kB. Sounds a little too little if all we want is to "prevent unbounded memory growth". Have you tested it and checked whether the watermark gets hit during normal operation?

WofWca (commented):
> tbd: should we keep this PR?

If you're willing to finish it then I'd say yes.

> So our assumption that this could be the reason was wrong. 

I guess it depends on what actually happens on disk full - does `write()` fail or does it just keep buffering.

But indeed might be better to handle the "disk full" condition in Core first, if it is not.

packages/shared/log-handler.ts
nicodh: We have the time stamp so any gap is easy to detect

nicodh:
I adjusted the PR now: the error is shwon to the user as before but the refactoring kept (only one log-handler, better stream handling)

packages/shared/log-handler.ts
nicodh: No I didn't test. Seems not worth.

> Sounds a little too little 
There is no condition based on highWaterMark it's about buffering
See https://nodejs.org/api/stream.html#buffering 

"While the total size of the internal write buffer is below the threshold set by highWaterMark, calls to writable.write() will return true"

 and the unbounded memory growth would happen if we would pass more log messages to a stream that can't write anymore.

What would do you want to express/suggest here?

WofWca: Are you assuming that the gap will last for hours? What if its duration is comparable to the time between log entries, e.g. one second?

WofWca: > What would do you want to express/suggest here?

That it's possible that we'd be dropping log entries very frequently, especially on bursts, such as the ones that happen on `maybeNetwork`.

> No I didn't test. Seems not worth.

If so then let's increase `highWaterMark` to a megabyte or two, if the concern is unbounded memory growth.

nicodh: As said I think it's not worth here to add more stuff so I will revert to the simple handling we had before and just keep the refactoring (using a shared log handler)

packages/target-electron/src/index.ts
```diff
@@ -104,9 +104,19 @@ mkdirSync(getLogsPath(), { recursive: true })
...
-import { cleanupLogFolder, createLogHandler } from './log-handler.js'
-const logHandler = createLogHandler()
-import { getLogger, setLogHandler } from '../../shared/logger.js'
+import {
+  cleanupLogFolder,
+  createLogHandler,
+} from '@deltachat-desktop/shared/log-handler.js'
+const logHandler = createLogHandler(getLogsPath(), {
+  onWriteError: (err, logFilePath) => {
+    throw new Error(
```
WofWca, line 114:
```suggestion
    // Trigger our `uncaughtException` listener.
    throw new Error(
```

WofWca (approved):
Looks good.

I have checked that the removed file had been copy-pasted, with only the `LOGS_DIR` part being different, i.e. this MR removes duplicate code.
The extra `try-catch` for JSON.stringify looks good.
The LOGS_DIR changes don't seem like they change behavior.
The unrelated import path changes are also fine.

I didn't try to trigger an error, but logging to a file seems to work as usual on Electron.

PR #6274, opened 2026-04-17
fix: prevent crash on MacOS when closing notify twice

### LINKED ISSUE #6210: crash on using the app

nicodh:
fixes #6210

Closing a already closd notification crashes the app immediately on MacOS

Besides reverting d78646d83c this adds a WeakSet to keep references to closed notifications (without preventing them from being garbage collected)

### DIFF
--- a/packages/target-electron/src/notifications.ts
+++ b/packages/target-electron/src/notifications.ts
@@ -111,6 +111,12 @@ const notifications: {
 } = {}
 
 /**
+ * Track closed notifications to prevent calling close() on
+ * already-closed notifications, which crashes app immediately on macOS.
+ */
+const closedNotifications = new WeakSet<Notification>()
+
+/**
  * triggers creation of a notification, adds appropriate
  * callbacks and shows it via electron Notification API
  *
@@ -133,23 +139,23 @@ function showNotification(_event: IpcMainInvokeEvent, data: DcNotification) {
         notifications[accountId]?.[chatId]?.[data.messageId]?.filter(
           n => n !== notify
         ) || []
-      notify.close()
+      if (!closedNotifications.has(notify)) {
+        closedNotifications.add(notify)
+        notify.close()
+      }
     })
     notify.on('close', () => {
+      // Mark as closed to prevent calling close() again later
+      closedNotifications.add(notify)
       // on Window and Linux this can be triggered by system time out
-      // when the message is moved to notification center so only close
-      // the notification on this event on Mac
+      // when the message is moved to notification center so only clear
+      // the notification tracking on Mac
       if (isMac) {
         notifications[accountId][chatId][data.messageId] =
           notifications[accountId]?.[chatId]?.[data.messageId]?.filter(
             n => n !== notify
           ) || []
-        // But now that we've removed the references to the notification,
-        // make sure that it really is closed.
-        notify.close()
       }
-      // eslint-disable-next-line no-console
-      console.log('Notification close event triggered', notify)
     })
     notify.on('reply', async e => {
       // See the Android's implementation:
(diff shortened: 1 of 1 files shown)


WofWca (commented):
Good job figuring this out!

However, doesn't this break notification clearing on non-mac systems? Currently even after the `close` event we still keep the reference to the Notificaion. That is for `close()`ing it later. But with this MR if a `close` event fires then we'll never `close()` the notification again.
I think we should only run the new check on Mac.

nicodh:
> if a close event fires then we'll never close() the notification again

ok, my focus was on MacOS where the notifications is "really" closed but looking at the documentation it says " If a notification is in the Action Center after the initial close event is emitted, a call to notification.close() will remove the notification from the action center but the close event will not be emitted again."

But the comment says:
  // on Window and Linux this can be triggered by system time out
  // when the message is moved to notification center so only close
  // the notification on this event on Mac

So the solution would be to add the notification to the closedNotifications only on MacOS.

But if the notification was closed by system timeout on non MacOS why should we remove it from the Action center by calling close on it. The user might not have seen it yet. It seems not possible to distinguish if the close event was triggered by manual closing it (then we should call notify.close())  or by timeout (then we should only close it in clearNotificationsForMessage/Chat etc.)

WofWca:
> why should we remove it from the Action center by calling close on it. The user might not have seen it yet.

We don't want to remove it immediately. We want to remove it when its message gets read from another device. I'm not sure where you're coming from with this?

WofWca:
>  I think we should only run the new check on Mac.

Wait, so isn't the crash caused by the infinite loop of `close()` ing and the `close` event firing? I think then it'd be simpler to just revert my commit and that's it. No need to introduce the `WeakSet`.

nicodh:
Ok, I would suggest to only add the closed notification to the WeakSet on Mac to prevent crashes in any case.

packages/target-electron/src/notifications.ts
```diff
@@ -133,23 +139,27 @@ function showNotification(_event: IpcMainInvokeEvent, data: DcNotification) {
         notifications[accountId]?.[chatId]?.[data.messageId]?.filter(
           n => n !== notify
         ) || []
-      notify.close()
+      if (!closedNotifications.has(notify)) {
+        if (isMac) {
+          closedNotifications.add(notify)
+        }
+        notify.close()
+      }
```
WofWca, line 147: Maybe better to extract this into an utility function called `closeNotification` at this point.

packages/target-electron/src/notifications.ts
```diff
@@ -110,6 +110,12 @@ const notifications: {
   [accountId: number]: { [chatId: number]: { [msgId: number]: Notification[] } }
 } = {}
 
+/**
+ * Track closed notifications on Mac to prevent calling close() on
+ * already-closed notifications, which crashes app immediately on macOS.
```
WofWca, line 115:
```suggestion
 * already-closed notifications, which crashes app immediately on macOS.
 *
 * Or maybe it only used to crash simply because of an infinite loop
 * of us calling `close()` inside the `close` event handler. Not sure.
```

packages/target-electron/src/notifications.ts
```diff
@@ -133,23 +139,27 @@ function showNotification(_event: IpcMainInvokeEvent, data: DcNotification) {
...
+          closedNotifications.add(notify)
+        }
+        notify.close()
+      }
     })
     notify.on('close', () => {
+      // Mark as closed to prevent calling close() again later
+      if (isMac) {
+        closedNotifications.add(notify)
+      }
```
WofWca, line 153: There is already an `if (isMac)` check a line below, this can be moved inside of there.

WofWca (approved):
The Windows bug is fixed now and it seems that only macOS is affected by this change, as far as I can tell.
I am also fine with the removal of the log entry.

PR #6276, opened 2026-04-18
refactoring: follow up notification change

nicodh:
Just extract close function


WofWca (approved):
Looks a little better. The behavior is not affected, as far as I can tell by looking at the code.

Issue #3738, opened 2024-03-27
Sending a message sometimes becomes slow on chats with many messages / sometimes draft is not removed


hpk42:
Those of you who are with me in some chat groups will have seen that i sometimes sent "double messages". 
This happens because when i hit "CTRL-return" or click the send button in a group chat with many messages, it takes 3-10 seconds until the message is sent, and then i click a second time because i wrongly think i hadn't clicked it.  Sometimes, the draft-message doesn't even go away. 
In chats with only few messages, i do not experience the problem. It's smooth/swift there. 
Sometimes it seems to be even fast in chats with many messages. 

It's all the same database, so my best guess it that something in the sending-logic (changing a draft message to a sent one) might iterate over all messages in a chat.  But whatever the guess, it would be great if someone more knowledgable of the code can take a look into what happens when the send-button is clicked. 

![image](https://github.com/deltachat/deltachat-desktop/assets/73579/2148e51f-7385-434e-96b8-993790186289)

Simon-Laux:
> It's all the same database, so my best guess it that something in the sending-logic (changing a draft message to a sent one) might iterate over all messages in a chat.

if so its core being slow with sending as the code waits for core.

Anyways this code area is in high need of rewriting. see https://github.com/deltachat/deltachat-desktop/milestone/9

hpk42:
On Wed, Mar 27, 2024 at 05:53 -0700, Simon Laux wrote:

> > It's all the same database, so my best guess it that something in the sending-logic (changing a draft message to a sent one) might iterate over all messages in a chat.
> if so its core being slow with sending as the code waits for core.

I find this unlikely as on Android i never have this problem. 
If i had in on both i'd file it directly on core.

Simon-Laux:
so on android you never have delays while sending?
I have a hard time imagining that desktop ui code has delays that depend on message count in the chat as unless you scrolled to load messages only the most recent 20 or so messages are even loaded and displayed in the ui

hpk42:
On Wed, Mar 27, 2024 at 10:04 -0700, Simon Laux wrote:
> so on android you never have delays while sending?

no, i don't have delays there. 

> I have a hard time imagining that desktop ui code has delays that depend on message count in the chat as unless you scrolled to load messages only the most recent 20 or so messages are even loaded and displayed in the ui

maybe it is something else and not the number of messages in a chat
because sometimes it's also fast in large chats
and the app is sometimes slow at other occassions eg
[...]

PR #6278, opened 2026-04-19
fix: enable sticker replies

### LINKED ISSUE #2998: Sending a sticker as answer (with a quote) is not possible from desktop
if reply is set, but the draft is otherwise empty (no image/file/text), then the quote should be attached to the sticker and sent out.

nicodh:
resolves #2998

### DIFF
--- a/packages/frontend/src/components/composer/EmojiAndStickerPicker.tsx
+++ b/packages/frontend/src/components/composer/EmojiAndStickerPicker.tsx
@@ -17,6 +17,7 @@ import useTranslationFunction from '../../hooks/useTranslationFunction'
 import useMessage from '../../hooks/chat/useMessage'
 import { ContextMenuContext } from '../../contexts/ContextMenuContext'
 import useConfirmationDialog from '../../hooks/dialog/useConfirmationDialog'
+import type { DraftObject } from '../../hooks/chat/useDraft'
 
 import styles from './styles.module.scss'
 
@@ -32,6 +33,8 @@ type Props = {
   chatId: number
   setShowEmojiPicker: (enabled: boolean) => void
   onStickerDeleted: () => void
+  draftQuote?: DraftObject['quote']
+  removeQuote?: () => void
 }
 
 const DisplayedStickerPack = ({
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/composer/EmojiAndStickerPicker.tsx
```diff
@@ -28,30 +29,49 @@ type Props = {
...
     const stickerPath = fileName.replace('file://', '')
-    BackendRemote.rpc.sendSticker(accountId, chatId, stickerPath).then(msgId =>
+
+    // If there's a quote, send the sticker with the quote
+    if (draftQuote && draftQuote.kind === 'WithMessage') {
+      await sendMessage(accountId, chatId, {
+        file: stickerPath,
+        viewtype: 'Sticker',
+        quotedMessageId: draftQuote.messageId,
+      })
```
WofWca: Do you have a justification why it's OK to skip `msg.force_sticker();`?
Otherwise we need to add the `quote_message_id` param to the `send_sticker` function.

https://github.com/chatmail/core/blob/8b58b16cb528e29a7f69f1337a3889333fa413c7/deltachat-jsonrpc/src/api.rs#L2362-L2378

link2xt: There is a logic that may turn Viewtype::Sticker into non-sticker if it looks like a screenshot:
https://github.com/chatmail/core/blob/d069e75cd8cbd2ac12d3817429fa584174e86175/src/message.rs#L2324-L2330

IIRC was introduced for the case when an image was dragged from gallery into the chat on iOS, resulting in sending screenshots as stickers.

link2xt: After some discussion with @r10s, plan is to remove sticker heuristics and `force_sticker()` in the core, then fix the problems in each UI that accidentally sends screenshots as stickers: https://github.com/chatmail/core/pull/8162

nicodh: So for now we can leave this PR as it is. Although if you have images as sticker they will end up to be transformed by core to and sent as normal image message. After core is updated we could adapt what is needed

nicodh: I tested this with current core and the viewType Sticker in the sent message is preserved now so no adaptions needed

WofWca: Kind of looks like `send_sticker` is not needed at all then.

packages/frontend/src/components/composer/EmojiAndStickerPicker.tsx
```diff
@@ -257,6 +292,8 @@ export const EmojiAndStickerPicker = forwardRef<
           chatId={chatId}
           stickers={stickers}
           setShowEmojiPicker={setShowEmojiPicker}
+          draftQuote={draftQuote}
+          removeQuote={removeQuote}
```
WofWca: Feels like we shouldn't be passing such functions so deeply to `StickerPicker`. Instead handle "sticker click" somewhere higher.

packages/frontend/src/components/composer/EmojiAndStickerPicker.tsx
```diff
@@ -28,30 +29,49 @@ type Props = {
   stickerPackImages: string[]
   chatId: number
   setShowEmojiPicker: (enabled: boolean) => void
+  draftQuote?: DraftObject['quote']
+  removeQuote?: () => void
```
WofWca:
```suggestion
  draftQuote: DraftObject['quote']
  removeQuote: () => void
```

I'd say these shouldn't be optional.
It still compiles then.

WofWca (approved):
Works alright. I though about race conditions like https://github.com/deltachat/deltachat-desktop/issues/3738 https://github.com/chatmail/issue-stargate/issues/13. I couldn't come up with anything too bad. The quote might linger around until the message gets sent, but that alone doesn't enable the "double send" bug.

packages/frontend/src/components/composer/EmojiAndStickerPicker.tsx
nicodh: I agree and changed it to be handled inside composer

packages/frontend/src/components/composer/EmojiAndStickerPicker.tsx
nicodh: Is obsolete with last commit

WofWca (approved):
Tested again, still works. The code looks better.

PR #6279, opened 2026-04-19
fix: switch account in all SelectChat dialogs

### LINKED ISSUE #6270: Switch profile is missing in send to chat dialog from webxdc

nicodh:
resolves #6270


packages/frontend/src/components/dialogs/SelectChat/index.tsx
```diff
@@ -17,28 +24,77 @@ import { RovingTabindexProvider } from '../../../contexts/RovingTabindex'
...
   const tx = useTranslationFunction()
+  const currentAccountId = selectedAccountId()
+  const [targetAccountId, setTargetAccountId] = useState(currentAccountId)
+  const { openDialog } = useDialog()
+
+  const accountId = props.enableAccountSwitch
+    ? targetAccountId
+    : currentAccountId
+
+  const configuredAccountsFetch = useRpcFetch(getConfiguredAccounts, [])
```
WofWca: FYI the regular `useFetch` can be used, but it doesn't matter much I think,

packages/frontend/src/components/dialogs/ForwardMessage/index.tsx
```diff
@@ -158,36 +128,13 @@ export default function ForwardMessage(props: ForwardMessageProps) {
...
-
   return (
     <SelectChat
       headerTitle={tx('forward_to')}
       onChatClick={onChatClick}
       onClose={onClose}
       listFlags={C.DC_GCL_FOR_FORWARDING | C.DC_GCL_NO_SPECIALS}
-      accountId={targetAccountId}
-      accountSwitch={accountSwitch}
-      key={targetAccountId}
```
WofWca, line 219: This re-introduces the bug that we discussed in https://github.com/deltachat/deltachat-desktop/pull/5916#discussion_r2678699256.

packages/frontend/src/components/dialogs/SelectChat/index.tsx
```diff
@@ -17,28 +24,77 @@ import { RovingTabindexProvider } from '../../../contexts/RovingTabindex'
 type Props = {
   onClose: DialogProps['onClose']
   headerTitle: string
-  onChatClick: (chatId: number) => void
+  onChatClick: (chatId: number, accountId: number) => void
```
WofWca: I think we should put `accountId` as the first argument, because otherwise it's pretty easy for the user of this component to handle the `chatId` argument but ignore the `accountId` argument.
I can see that this is done this way for use cases where `accountId` actually needs to be ignored because `enableAccountSwitch == false`. I suggest to solve this by doing some TypeScript magic where the type of the `onChatClick` function depends on whether `enableAccountSwitch` is `true`.

WofWca (changes requested):

packages/frontend/src/components/dialogs/ForwardMessage/index.tsx
nicodh: So maybe it's time then to make useLogicVirtualChatList account switch aware. I added a commit that resets the chat cache on account switch

WofWca: Why not just wrap the lower part of the `SelectChat` dialog in `key={accountId}` again?

WofWca: (but yep, then the state of the lower part (`useChatList` and stuff) would also need be in a separate component)

nicodh: Ok I splitted the Select Chat in 2 components now

packages/frontend/src/components/dialogs/SelectChat/index.tsx
```diff
@@ -17,28 +24,107 @@ import { RovingTabindexProvider } from '../../../contexts/RovingTabindex'
...
+  const hasMultipleAccounts = props.enableAccountSwitch
+    ? configuredAccountsFetch.result?.ok
+      ? configuredAccountsFetch.result.value.length > 1
+      : false
+    : false
+
+  const accountFetch = useRpcFetch(BackendRemote.rpc.getAccountInfo, [
+    accountId,
+  ])
+  const accountInfo = accountFetch.lingeringResult?.ok
```
WofWca:
```suggestion
  const configuredAccountsFetch = useRpcFetch(
    getConfiguredAccounts,
    props.enableAccountSwitch ? [] : null
  )
  const hasMultipleAccounts = props.enableAccountSwitch
    ? configuredAccountsFetch?.result?.ok
      ? configuredAccountsFetch.result.value.length > 1
      : false
    : false

  const accountFetch = useRpcFetch(
    BackendRemote.rpc.getAccountInfo,
    props.enableAccountSwitch ? [accountId] : null
  )
  const accountInfo = accountFetch?.lingeringResult?.ok
```

For performance don't fetch the stuff when `!props.enableAccountSwitch`.
I have tested this.

Also I think this can be avoided by moving this stuff into another component which only gets rendered when `props.enableAccountSwitch`.

packages/frontend/src/components/dialogs/ViewProfile/menu.tsx
```diff
@@ -203,24 +203,31 @@ function ShareProfileDialog(
   const { onClose, onParentClose, contact } = props
 
   const tx = useTranslationFunction()
-  const accountId = selectedAccountId()
+  const currentAccountId = selectedAccountId()
   const createDraftMessage = useCreateDraftMessage()
 
-  const onChatClick = async (chatId: number) => {
+  const onChatClick = async (targetAccountId: number, chatId: number) => {
+    // Close dialogs before createDraftMessage because it may switch accounts
```
WofWca: The comment makes it sound like we want to close the dialogs _only if_ we switch accounts. But in reality we want to close it regardless, right?

I'd try to clarify this comment, make it about _the order_ and not the fact that we close the dialogs at all.

Same goes for other 3 instances of such a comment.

packages/frontend/src/components/dialogs/WebxdcSendToChat/index.tsx
```diff
@@ -19,10 +18,9 @@ export default function WebxdcSaveToChatDialog(props: Props) {
   const { onClose, messageText, file } = props
 
   const tx = useTranslationFunction()
-  const accountId = selectedAccountId()
   const createDraftMessage = useCreateDraftMessage()
 
-  const onChatClick = async (chatId: number) => {
+  const onChatClick = async (chatId: number, accountId: number) => {
```
WofWca:
```suggestion
  const onChatClick = async (accountId: number, chatId: number) => {
```

Previously I suggested to put `accountId` as the first argument. It would be extra safe to use named parameters (`{ targetAccountId: number, chatId: number }`) to avoid such bugs.

packages/frontend/src/hooks/chat/useCreateDraftMesssage.ts
```diff
@@ -36,7 +37,22 @@ export default function useCreateDraftMessage() {
 
   return useCallback<CreateDraftMessage>(
     async (accountId, chatId, messageText, file) => {
-      selectChat(accountId, chatId)
+      const isCrossAccount = accountId !== selectedAccountId()
+
+      if (isCrossAccount) {
+        try {
+          await saveLastChatId(accountId, chatId)
```
WofWca, line 44: FYI we have other instances of such code and one of them has a comment explaining why `saveLastChatId` is needed. Maybe we ought to clean that up a little some day.

packages/frontend/src/components/dialogs/WebxdcSendToChat/index.tsx
WofWca: Also maybe better to rename to `targetAccountId` to be extra clear that this could be different from the currently active account.
Same for other occurrences.

WofWca (changes requested):
Works nicely! (apart from WebXDC, see my comment)
The `key={accountId}` bug is gone, the refactor that did it looks nice.

The CSS changes also seem OK, didn't look to closely.

It's cool that so little adjustment was needed to make it work.

Edit: only the WebXDC bug comment is really blocking the merge IMO.

WofWca (approved):

PR #6281, opened 2026-04-19
fix: improve 3-dot menu size in html mails

nicodh:
new | old

image


WofWca (approved):
Looks better to me. Code the code is OK.
Not sure if we can call this a fix though.

PR #6282, opened 2026-04-20
fix: always use core viewType for compatibility

### LINKED ISSUE #6049: Desktop ignores viewtype and displays SVG as an image even though it should have a File viewtype from the core
Reported here:
https://support.delta.chat/t/support-rendering-of-svg-on-mobile-goodfirstissue/4760

This results in expectation that SVG will be rendered for receivers, but receivers on mobile cannot see the image, it will appear as a file.

nicodh:
resolves #6049

As mentioned in the comment for MEDIA_VIEW_TYPES: the compatibilty with other clients now has higher priority to avoid one client sending media that the other clients won't be able to run/see for the cost that not all media types that could be displayed will be displayed

### DIFF
--- a/packages/frontend/src/components/attachment/Attachment.tsx
+++ b/packages/frontend/src/components/attachment/Attachment.tsx
@@ -2,60 +2,30 @@ import mimeTypes from 'mime-types'
 import { Type } from '../../backend-com'
 import { runtime } from '@deltachat-desktop/runtime-interface'
 
-/* Section - Data Copied in part from Signal */
-// Supported media types in google chrome
-// See: https://en.wikipedia.org/wiki/Comparison_of_web_browsers#Image_format_support
-const SUPPORTED_IMAGE_MIME_TYPES = Object.freeze([
-  'image/bmp',
-  'image/gif',
-  'image/jpeg',
-  'image/jpg',
-  'image/svg+xml',
-  'image/webp',
-  'image/x-xbitmap',
-  // ICO
-  'image/vnd.microsoft.icon',
-  'image/ico',
-  'image/icon',
-  'image/x-icon',
-  // PNG
-  'image/apng',
-  'image/png',
-  // for opening avatars
-  'image/x',
-])
-// See: https://www.chromium.org/audio-video
-const SUPPORTED_VIDEO_MIME_TYPES = Object.freeze([
-  'video/mp4',
-  'video/ogg',
-  'video/webm',
-  'video/quicktime',
-])
-/* EndSection - Data Copied in part from Signal */
-
-export function isImage(filemime: string | null) {
-  return SUPPORTED_IMAGE_MIME_TYPES.includes(filemime || '')
+export function isImage(viewType: Type.Viewtype | null) {
+  return viewType === 'Image' || viewType === 'Gif'
 }
 
 export function hasAttachment(attachment: MessageTypeAttachmentSubset | null) {
   return attachment && attachment.file
 }
 
-export function isVideo(filemime: string | null) {
-  return SUPPORTED_VIDEO_MIME_TYPES.includes(filemime || '')
+// note that we rely on the viewTypes from core here to make sure all Deltachat clients
+// display the same attachment types in the same way - although this could result in not
+// showing some attachments even if we could show them (like svg for example)
+// see guess_msgtype_from_path_suffix in https://github.com/chatmail/core/blob/main/src/message.rs
+const MEDIA_VIEW_TYPES = ['Image', 'Gif', 'Video', 'Audio', 'Voice'] as const
+
+export function isDisplayableByFullscreenMedia(viewType: Type.Viewtype | null) {
+  return MEDIA_VIEW_TYPES.includes(
+    viewType as (typeof MEDIA_VIEW_TYPES)[number]
+  )
 }
 
-export function isAudio(filemime: string | null) {
-  if (!filemime) return false
-  return filemime.startsWith('audio/')
-}
-
-export function isDisplayableByFullscreenMedia(filemime: string | null) {
-  return isImage(filemime) || isAudio(filemime) || isVideo(filemime)
-}
-
-export function isGenericAttachment(filemime: string | null) {
-  return !(isImage(filemime) || isVideo(filemime) || isAudio(filemime))
+export function isGenericAttachment(viewType: Type.Viewtype | null) {
+  return !MEDIA_VIEW_TYPES.includes(
+    viewType as (typeof MEDIA_VIEW_TYPES)[number]
+  )
 }
 
 export function getExtension({
--- a/packages/frontend/src/components/attachment/Attachment.tsx
+++ b/packages/frontend/src/components/attachment/Attachment.tsx
@@ -10,9 +10,9 @@ export function hasAttachment(attachment: MessageTypeAttachmentSubset | null) {
   return attachment && attachment.file
 }
 
-// note that we rely on the viewTypes from core here to make sure all Deltachat clients
+// note that we rely on the viewTypes from core here to make sure all Delta Chat clients
 // display the same attachment types in the same way - although this could result in not
-// showing some attachments even if we could show them (like svg for example)
+// showing some attachments as files even if we could show them as media (like svg for example)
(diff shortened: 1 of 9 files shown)


packages/frontend/src/components/attachment/galleryAttachment.tsx
```diff
@@ -53,7 +47,7 @@ const contextMenuFactory = (
   openDialog: OpenDialog,
   jumpToMessage: JumpToMessage
 ) => {
-  const showCopyImage = message.viewType === 'Image'
+  const showCopyImage = isImage(message.viewType)
```
nicodh: This now also includes viewType GIF I don't know why that was excluded before

packages/frontend/src/components/attachment/Attachment.tsx
```diff
@@ -2,60 +2,30 @@ import mimeTypes from 'mime-types'
...
 
-export function isVideo(filemime: string | null) {
-  return SUPPORTED_VIDEO_MIME_TYPES.includes(filemime || '')
-}
-
-export function isAudio(filemime: string | null) {
-  if (!filemime) return false
-  return filemime.startsWith('audio/')
-}
+// note that we rely on the viewTypes from core here to make sure all Deltachat clients
```
WofWca:
```suggestion
// note that we rely on the viewTypes from core here to make sure all Delta Chat clients
```

packages/frontend/src/components/attachment/Attachment.tsx
```diff
@@ -2,60 +2,30 @@ import mimeTypes from 'mime-types'
...
+// display the same attachment types in the same way - although this could result in not
+// showing some attachments even if we could show them (like svg for example)
```
WofWca:
```suggestion
// display the same attachment types in the same way - although this could result in
// showing some attachments as files even if we could show them as media (like svg for example)
```

Otherwise sounds like we don't show the attachment at all.

packages/frontend/src/components/attachment/galleryAttachment.tsx
WofWca: Have you checked if it works though? Does it copy the whole GIF with all its frames and not just one frame?

This is how our "copy image" currently works.

https://github.com/deltachat/deltachat-desktop/blob/228c4bb7c8ef93016ceca90167542b24107a51b5/packages/target-electron/src/ipc.ts#L376-L378

Docs state https://www.electronjs.org/docs/latest/api/native-image#nativeimagecreatefrompathpath

> from an image file (e.g., PNG or JPEG)

packages/frontend/src/components/attachment/galleryAttachment.tsx
```diff
@@ -235,7 +229,7 @@ export function ImageAttachment({
       accountId
     )
     const { file, fileMime } = message
-    const hasSupportedFormat = isImage(fileMime)
+    const hasSupportedFormat = isImage(message.viewType)
```
WofWca: Does this check make sense any more?
Same for video and audio.

packages/frontend/src/components/attachment/messageAttachment.tsx
```diff
@@ -308,7 +306,7 @@ export function DraftAttachment({
         />
       </div>
     )
-  } else if (isVideo(attachment.fileMime)) {
+  } else if (attachment.viewType === 'Video') {
```
WofWca, line 300: With this change when attaching a video (either with "attach" menu or drag and drop) the video is not previewable and is displayed as a file. Should we fix that somehow?

WofWca (commented):
Tested draft images, voice messages, tested GIFs in the gallery and message list for a bit, tested notification avatars and images. Seems to work alright on the surface level (except for draft videos, see comment).
I can see that previously `isImage` would also return `true` for stickers, and I have checked the changes for whether they are OK in this regard (they seem to be).

Edit: I've also checked that `fileMime` is no longer used when deciding how to display an attachment, so yes, this addresses the issue.

packages/frontend/src/components/attachment/messageAttachment.tsx
nicodh: After discussing with @r10s we decided to go for a similar behaviour as on Android: show a preview for video, audio, vCards & webxdc (not supported yet) even when these files are loaded from the "File" menu.

We have to rely on the extension since there is no viewType right after selecting a file and add it to a draft. The final viewType will anyway be calculated by core but for the preview we have to go with the extension.

packages/frontend/src/components/attachment/galleryAttachment.tsx
nicodh: Removed it

packages/frontend/src/components/attachment/galleryAttachment.tsx
```diff
@@ -477,13 +460,6 @@ export function AudioAttachment({
               })
             }
           />
-        ) : (
-          <div>
-            {window.static_translate(
-              'cannot_display_unsuported_file_type',
```
WofWca, line 488: Can we remove the string now? Or is it used on Android or iOS?

WofWca (commented):
So what the removal of `hasSupportedFormat` does is it basically makes us always display `<video>` `<audio>` or `` purely based on what Core returns, and to never show the "broken media" square.
Since there is still always a way to export such attachments and view them outside of Delta Chat, I think it's fine.

Related:
- https://github.com/deltachat/deltachat-desktop/issues/4663

packages/frontend/src/components/composer/menuAttachment.tsx
```diff
@@ -86,7 +90,18 @@ export default function MenuAttachment({
 
     if (files.length === 1) {
       setLastPath(dirname(files[0]))
-      addFileToDraft(files[0], basename(files[0]), 'File')
+      // for Audio, Video & Vcard set a ViewType that has a preview here
+      let viewType = 'File' as T.Viewtype
+      if (AUDIO_EXTENSIONS.some(ext => files[0].toLowerCase().endsWith(ext))) {
+        viewType = 'Audio'
+      } else if (
+        VIDEO_EXTENSIONS.some(ext => files[0].toLowerCase().endsWith(ext))
```
WofWca: A file named `aaaapng` will be recognized as an image file.
See other code that doesn't have this bug:

https://github.com/deltachat/deltachat-desktop/blob/b883cd6958eee40769a4e454f9dd97d65efd6635/packages/frontend/src/components/message/MessageListAndComposer.tsx#L93

packages/frontend/src/components/attachment/messageAttachment.tsx
WofWca: > for the preview we have to go with the extension.

"Have to" because Core currently doesn't support this or because there are reasons for Core not to support this?

Anyways, I think this is good enough.

packages/frontend/src/components/composer/menuAttachment.tsx
WofWca:
```suggestion
      if (AUDIO_EXTENSIONS.some(ext => files[0].toLowerCase().endsWith('.' + ext))) {
        viewType = 'Audio'
      } else if (
        VIDEO_EXTENSIONS.some(ext => files[0].toLowerCase().endsWith('.' + ext))
```

packages/frontend/src/components/attachment/galleryAttachment.tsx
nicodh: We don't remove translation files - has to happen in transifex

packages/frontend/src/components/attachment/galleryAttachment.tsx
nicodh: Just tested it and it doesn't work, so I ecluded Gifs from copyImage

packages/frontend/src/components/attachment/galleryAttachment.tsx
WofWca: Sure, but it has to be done nonetheless.

WofWca (approved):
I found no more issues with this.
Even if something's broken it shouldn't break too badly because downloading an attachment is always an option. Let's move on.
Closes the issue.

The only question left is the removal of the string.

PR #6283, opened 2026-04-20
Show channel description

### LINKED ISSUE #6280: Channel description not displayed
I've noticed that the channel descriptions aren't displayed for recipients. While it's possible to enter a description on the sender's side, and a system message appears in the chat history on both sides indicating that the description has been changed, the description itself isn't visible to the recipient.

macOS 26.4.1 
DC-Desktop 2.49.1 Electron Version (Apple Silicon)

nicodh:
resolves #6280

### DIFF
--- a/packages/frontend/src/components/dialogs/MailingListProfile.tsx
+++ b/packages/frontend/src/components/dialogs/MailingListProfile.tsx
@@ -7,6 +7,8 @@ import { type T } from '@deltachat/jsonrpc-client'
 import { DialogProps } from '../../contexts/DialogContext'
 import ProfileInfoHeader from '../ProfileInfoHeader'
 import { shouldDisableClickForFullscreen } from '../Avatar'
+import { BackendRemote } from '../../backend-com'
+import { useRpcFetch } from '../../hooks/useFetch'
 
 /**
  * This dialog is used to display the profile of a mailing list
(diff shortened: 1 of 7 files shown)


WofWca (commented):
Ah, so we added descriptions for outgoing channels but didn't add them for incoming ones #6026. That's funny.

This is better than nothing, however, the description doesn't linkify links and emails and the usual stuff that we linkify in messages and contact descriptions.
I have checked that we linkify stuff on Android.
This is the code that we use for contacts, where it works:

https://github.com/deltachat/deltachat-desktop/blob/31b687d11fa8d0d37d1dd5307fdccd783d2a68c0/packages/frontend/src/components/dialogs/ViewProfile/index.tsx#L266-L277

It's beginning to look like we need to DRY this stuff, move all these descriptions to inside of `ProfileInfoHeader`.

But, as I said, this is better than nothing. Do you think we should merge as is and then do the rest in follow-ups or would you like to do it in this MR? If it's the former then I'll proceed with review.

Edit: BTW yes, we also don't linkify stuff in group descriptions, but we should.

packages/frontend/src/components/dialogs/MailingListProfile.tsx
```diff
@@ -18,15 +20,24 @@ import { shouldDisableClickForFullscreen } from '../Avatar'
...
 
   const tx = useTranslationFunction()
 
+  const descriptionFetch = useRpcFetch(
+    BackendRemote.rpc.getChatDescription,
+    chat.chatType === 'InBroadcast' ? [accountId, chat.id] : null
+  )
+  const groupDescription = descriptionFetch?.result?.ok
+    ? descriptionFetch.result.value
+    : null
```
WofWca, line 39: Always happy to see my child, `useRpcFetch`, succeeding.

nicodh:
Good point. I will add that to this PR

WofWca (approved):
This fixes the bug, and the other bug with not linkifying group descriptions.

The introduced `closeAllDialogs` is to close the "profile" dialog when a `mailto:` link gets clicked, which makes sense. For comparison, we have `closeProfileDialog` in `BotCommandSuggestion`.

Clicking a bot command in a channel or a group description does nothing.

The usage of `parseAndRenderMessage` instead of `MessageBody` is a thing that is different from the existing code that we have for rendering contact descriptions, but I guess it's fine.

The presence of tests is nice.

PR #6285, opened 2026-04-20
fix: chat list: Ctrl + Space multiselect on Linux

### LINKED ISSUE #6209: Ctrl + Space for multiselect doesn't work on Linux

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/6209.

Apparently in Chromium on Linux the "click" event doesn't fire
for the Ctrl + Space combination.
Might be because there is such a native shortcut in Chrome.

It's not entirely clear though why the E2E tests are passing then.

Regardless, we want to be handling the `keydown` event anyway
in order for multiselect to be able to work
with non-`<button>` elements,
because such elements don't fire the `click` event on Space press.
We'll need this when we add multiselect to the message list
(https://github.com/deltachat/deltachat-desktop/pull/6268).

### DIFF
--- a/packages/frontend/src/components/chat/ChatListItem.tsx
+++ b/packages/frontend/src/components/chat/ChatListItem.tsx
@@ -141,11 +141,13 @@ export const PlaceholderChatListItem = React.memo(
 
 function ChatListItemArchiveLink({
   onClick,
+  onKeyDown,
   onFocus,
   chatListItem,
   ...rest
 }: {
   onClick: (event: React.MouseEvent) => void
+  onKeyDown?: React.KeyboardEventHandler
   onFocus?: (event: React.FocusEvent) => void
   chatListItem: Type.ChatListItemFetchResult & {
     kind: 'ArchiveLink'
@@ -179,7 +181,10 @@ function ChatListItemArchiveLink({
       {...rest}
       tabIndex={tabIndex}
       onClick={onClick}
-      onKeyDown={tabindexOnKeydown}
+      onKeyDown={e => {
+        onKeyDown?.(e)
+        tabindexOnKeydown(e)
+      }}
       onFocus={e => {
         tabindexSetAsActiveElement()
         onFocus?.(e)
@@ -204,6 +209,7 @@ function ChatListItemArchiveLink({
 function ChatListItemError({
   chatListItem,
   onClick,
+  onKeyDown,
   roleTab,
   onFocus,
   isSelected,
@@ -214,6 +220,7 @@ function ChatListItemError({
     kind: 'Error'
   }
   onClick: (event: React.MouseEvent) => void
+  onKeyDown?: React.KeyboardEventHandler
   onFocus?: (event: React.FocusEvent) => void
   onContextMenu?: (
     event: React.MouseEvent<HTMLButtonElement, MouseEvent>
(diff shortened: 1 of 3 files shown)


nicodh:
As already mentioned in the issue comment. I would not expect Ctrl + space or Shift + space to have any keybinding and it seems to be a shortcut that has different effects in different programs.

On Linux I see no effect for Ctrl + Space nor in main neither in this PR
It is also not in the list of the keybindings dialog.
 
What exactly do you want to achieve?

WofWca:
> I would not expect Ctrl + space or Shift + space to have any keybinding

So how would you do multiselect using only the keyboard? Besides Shift + ArrowUp.

> and it seems to be a shortcut that has different effects in different programs.

As the comments on `useMultiselect` suggests, the chat list widget's shortcuts are supposed to be aligned with the Listbox pattern https://www.w3.org/WAI/ARIA/apg/patterns/listbox/.

> Control + Space Changes the selection state of the focused option.
> Shift + Space (Optional): Selects contiguous items from the most recently selected item to the focused item.

https://github.com/deltachat/deltachat-desktop/blob/c18b2341282ed31c033b906254eb1d5aad5f3df0/packages/frontend/src/hooks/useMultiselect.ts#L18-L19

> On Linux I see no effect for Ctrl + Space nor in main neither in this PR

By this you mean that Ctrl + Space doesn't change selection or that this MR doesn't change the behavior of Ctrl + Space?

>  It is also not in the list of the keybindings dialog.

I don't think it needs to be, for the same reasons that we don't list "ArrowDown to scroll down or to select next item" or "Shift + ArrowUp to select text up": it's intuitive for people who do work with multiselectable widgets with keyboard-only.

WofWca:
Rebased

nicodh:
When the focus is on a chatlist item Ctrl + space selects it and hitting it again unselects it. That is for me the case in main as it is with this PR.

On Ubuntu/Gnome

WofWca:
Hmm weird. I tested on Ubuntu and Arch, both KDE and Gnome, on two different keyboards and the bug is there on main. I also tried a Windows 11 VM on the same host machine, and there is no bug.
Also tried with German keyboard layout.

I'm not sure why it's different for us. Maybe I'd need to try to find a different machine to try this on.

WofWca:
I got someone else who's on Linux to try this and for them neither Ctrl + Space nor Meta + Space works.
Not sure why Meta + Space doesn't work for them (but it doesn't need to because we only need that for Mac).
(edit: ah I tried Meta + Space on my machine on Gnome and it also doesn't work).
They confirmed that Shift + ArrowDown works for them.

WofWca:
I also confirmed that this MR fixes it on Gnome and KDE.

So IDK if we need to be getting to the bottom of this. Some machines are affected, some are not. I don't know why, but this MR doesn't introduce any weird stuff.

nicodh (approved):
Reviewed by reading &b testing

I can't reproduce the issue and so the fix but it doesn't break anything any if it solves issue on other platforms it's fine

WofWca:
> I got someone else who's on Linux to try this

They reported back that the bug is now fixed for them on latest main.

PR #6289, opened 2026-04-22
translate some untranslated strings

r10s:
draft, 
- some strings already exist under correct name, they are removed from `untranslated`
- some strings already exist, but need another name
- some strings require being pulled after https://github.com/deltachat/deltachat-android/pull/4372 is merged

### DIFF
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -1,7 +1,4 @@
 {
-  "no_account_selected": {
-    "message": "No profile selected"
-  },
   "pref_background_sync_disabled": {
     "message": "Only Synchronize Selected Profile"
   },
@@ -20,32 +17,11 @@
   "camera_access_failed": {
     "message": "Could not access video camera. You might want to check your video permissions or if the camera is used already by another program"
   },
-  "search_in": {
-    "message": "Search in %1$s"
-  },
-  "read_by": {
-    "message": "Read by"
-  },
-  "join_group": {
-    "message": "Join Group"
-  },
   "explain_desktop_browser_disabled_tray_pref": {
     "message": "Tray icon does exist in the web version, install the full version from https://get.delta.chat"
   },
-  "react_more_emojis": {
-    "message": "More emojis…"
-  },
-  "contact_is_blocked": {
-    "message": "This contact is blocked"
-  },
   "multidevice_connection_takes_too_long": {
     "message": "This takes longer than expected 🐌 make sure both devices are on the same Wi-Fi or network and check troubleshooting"
-  },
-  "search_result_for_x": {
-    "message": "Result for \"%s\""
-  },
-  "voice_send": {
-    "message": "Record a voice message"
   },
   "stop_recording": {
     "message": "Finish recording"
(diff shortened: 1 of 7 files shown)


nicodh (approved):
Reviewed by reading

packages/frontend/src/components/dialogs/CreateChat/index.tsx
```diff
@@ -850,21 +850,20 @@ export const ChatSettingsSetNameAndProfileImage = ({
...
     case GroupType.BROADCAST_LIST:
       inputLabel = tx('name_desktop')
-      missingNameErrorText = tx('please_enter_channel_name')
+      missingNameErrorText = tx('please_enter_chat_name')
       break
     default: {
       const _assert: never = groupType
       inputLabel = tx('group_name')
-      missingNameErrorText = tx('group_please_enter_group_name')
+      missingNameErrorText = tx('please_enter_chat_name')
```
WofWca, line 866: The label is always the same now, so this can be simplified now. But it's fine, can be done later, maybe when /if we implement the native validation with `required`,

packages/frontend/src/components/dialogs/ViewGroup.tsx
```diff
@@ -614,9 +614,7 @@ export function EditGroupNameDialog({
                 marginBottom: '-18px',
               }}
             >
-              {!isBroadcast
-                ? tx('group_please_enter_group_name')
-                : tx('please_enter_channel_name')}
+              {!tx('please_enter_chat_name')}
```
WofWca, line 617:
```suggestion
              {tx('please_enter_chat_name')}
```

_locales/_untranslated_en.json
```diff
@@ -20,33 +17,12 @@
...
-  "read_by": {
-    "message": "Read by"
-  },
-  "join_group": {
-    "message": "Join Group"
-  },
   "explain_desktop_browser_disabled_tray_pref": {
     "message": "Tray icon does exist in the web version, install the full version from https://get.delta.chat"
   },
-  "react_more_emojis": {
```
WofWca, line 35: This string has not been added in https://github.com/deltachat/deltachat-android/commit/af49018911ff296a11337a7c945dc14cb8364d89

WofWca (commented):
I wonder why our CI is green? This is not supposed to pass `pnpm check`.

nicodh:
> I wonder why our CI is green? This is not supposed to pass pnpm check.

Strange - I will fix the issues now

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -276,7 +276,7 @@ function buildViewEditMenuItems(
     },
     // Edit Channel
     relatedChat.chatType === 'OutBroadcast' && {
-      label: tx('edit_channel'),
+      label: tx('menu_view_profile'),
```
nicodh, line 279: If we use this as replacement for edit_channel the we can (and have to) simplify https://github.com/deltachat/deltachat-desktop/blob/15ff5058bf017e96e23d12f78167d1b6e804f3b7/packages/frontend/src/components/screens/MainScreen/MainScreen.tsx#L435

PR #6291, opened 2026-04-22
Button cleanup

### LINKED ISSUE #5887: unify button styles

_Originally posted by @r10s in https://github.com/deltachat/deltachat-desktop/pull/5882#discussion_r2627738406_

nicodh:
fixes #5887
Set consistent button heights, use flex layout for dialog footer buttons

Before 

image

image

After

image

image

### DIFF
--- a/packages/frontend/src/components/Dialog/styles.module.scss
+++ b/packages/frontend/src/components/Dialog/styles.module.scss
@@ -168,5 +168,4 @@ $paddingVertical: 20px;
   min-width: 100px;
   flex: 0 1 auto;
   margin: 0;
-  white-space: nowrap;
 }
--- a/packages/frontend/src/components/Dialog/styles.module.scss
+++ b/packages/frontend/src/components/Dialog/styles.module.scss
@@ -143,9 +143,8 @@ $paddingVertical: 20px;
 
 .footerActions {
   display: flex;
-  @media (width <= variables.$dialog-breakpoint) {
-    flex-direction: column;
-  }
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/Dialog/styles.module.scss
```diff
@@ -171,14 +170,7 @@ $paddingVertical: 20px;
...
-    margin-inline-start: 20px;
-  }
-  @media (width <= variables.$default-dialog-width) {
-    & + & {
-      margin-inline-start: 10px;
-    }
-  }
+  flex: 0 1 auto;
+  margin: 0;
+  white-space: nowrap;
```
WofWca:
```suggestion
```

This seems to be a residual line added during the development of this MR? Without this the dialogs look the same in most cases. Except when the button text would overflow the screen. Then I think we should wrap inside the button, which this line prevents us from doing.

WofWca (commented):
The fact that we allow for wrapping is good overall I think, as well as the transition to `gap`. However, I'd say that for narrow screens we should keep `flex-direction: column;`, i.e. not remove that first `@media` query. Otherwise the buttons don't take the full width. Worse, the button order is really awkward in e.g. the "delete or hide transport" dialog on narrower screens.

WofWca:
I rebased the MR, hopefully that's fine?

WofWca (approved):
As I said, I'm not super happy with how buttons look on < 500px, but it's not too important I guess.

image

Good otherwise.

PR #6292, opened 2026-04-22
fix: use dialog buttons in about dialog

### LINKED ISSUE #5887: unify button styles

nicodh:
This is part of a fix #5887 the final fix is in #6291


WofWca (approved):
Not a big deal, but I guess this addresses the issue.
Tested on short screens, I would say there is enough space.

r10s:
the issue is not the appearance, but the fact that we have different styles for the same thing at all.

that resulted in confusion, and then appearance issues in the past. making adaptions, including acceasibility, harder etc

PR #6297, opened 2026-04-23
fix: add counter for screenreaders

### LINKED ISSUE #5368: a11y: Consider adding new messages counter before the actual message when labelling tabs in the conversations page tab list
- Operating System: all
- Delta Chat Version: 2.8.0
- Expected behavior: When using deltachat with a screen reader the tabs in the conversations page tab list have an accessible label. Which looks like this: <chat name> <relative time> <message preview> <%d new messages>. Please, I would like you to reconsider tweaking this so number of new messages is added right after the chat name, before the relative time and message preview.
- Actual behavior: Number of new messages is reported at the end when using up and down arrow keys to navigate in the list of conversations.
* Open deltachat with a screen reader running
* Use tab or shift+tab to navigate until you find list of conversations.
* Use up and down arrow key to navigate and realize the number of new messages is reported at the end of the tab being selected. When there is a longer message to preview it is not very practical to listen to the whole long message before hearing how many new messages there are in the selected conversation.

nicodh:
fixes #5368

I decided to go with the redundant variant (having 2 function calls per chatlist item) since there was nice way to get the same counter element into the header (for screen readers) and to the end of chat-list-item-message with pure css

### DIFF
--- a/packages/frontend/src/components/chat/ChatListItem.tsx
+++ b/packages/frontend/src/components/chat/ChatListItem.tsx
@@ -12,25 +12,35 @@ import { useContextMenuWithActiveState } from '../ContextMenu'
 import { selectedAccountId } from '../../ScreenController'
 import { parseAndRenderMessage } from '../message/MessageParser'
 import { useRovingTabindex } from '../../contexts/RovingTabindex'
-import useTranslationFunction from '../../hooks/useTranslationFunction'
 import { shouldHideDeliveryStatus } from '../message/messageFunctions'
 
 const log = getLogger('renderer/chatlist/item')
 
-function FreshMessageCounter({ counter }: { counter: number }) {
-  const tx = useTranslationFunction()
-
+function FreshMessageCounter({
+  counter,
+  visible,
+}: {
+  counter: number
+  visible: boolean
+}) {
   if (counter === 0) return null
-  return (
-    <span
-      className='fresh-message-counter'
-      aria-label={tx('chat_n_new_messages', String(counter), {
-        quantity: counter,
-      })}
-    >
-      {counter}
-    </span>
-  )
+  const tx = window.static_translate
+  if (visible) {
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/chat/ChatListItem.tsx
```diff
@@ -12,25 +12,35 @@ import { useContextMenuWithActiveState } from '../ContextMenu'
...
-    <span
-      className='fresh-message-counter'
-      aria-label={tx('chat_n_new_messages', String(counter), {
-        quantity: counter,
-      })}
-    >
-      {counter}
-    </span>
-  )
+  const tx = window.static_translate
```
WofWca: What is the reason to go from `useTranslationFunction` to `window.static_translate`?

packages/frontend/src/components/chat/ChatListItem.tsx
```diff
@@ -42,13 +52,18 @@ function Header({
...
+  ChatListItemType,
+  'lastUpdated' | 'name' | 'isPinned' | 'isMuted' | 'freshMessageCounter'
+>) {
   const tx = window.static_translate
   return (
     <div className='header'>
       <div className='name'>
         <span>
           <span className='truncated'>{name}</span>
+          <FreshMessageCounter counter={freshMessageCounter} visible={false} />
```
WofWca, line 67: Nitpick: this should not be inside an element with class name "name".

WofWca (approved):
Tested with Orca reader. Works great, fixes the issue.
The code makes sense.
Much better!

packages/frontend/src/components/chat/ChatListItem.tsx
nicodh: Just to avoid having different ways to use tx, but considering again I decided to use the dynamic variant throughout

Issue #6293, opened 2026-04-22
chat creation issues


r10s:
- during channel creation, it should be possible to set an image
- for group/channel creation should be possible to set a description
- when not entering a channel name, there is no padding between error message and button

WofWca:
I think the latter should be solved by using native error messages by applying the `required` prop. Although I believe that does not allow for custom styles.

nicodh:
closed with #6314

PR #6298, opened 2026-04-23
fix: accessible "chat name is required" error msg

WofWca:
Before:

image

After:
image
image

The error appears when the user tries to submit the form.

Note that the error message is a bit more generic,
compared to "Please enter a name.".

This incidentally addresses the following point from
https://github.com/deltachat/deltachat-desktop/issues/6293
> when not entering a channel name, there is no padding
> between error message and button

Because this basically replaces the error message
with a native Chromium one.

Note that the "edit group" dialog is not affected. It requires us to add a `<form>` first. I'll do that in another MR.

### DIFF
--- a/packages/e2e-tests/playwright-helper.ts
+++ b/packages/e2e-tests/playwright-helper.ts
@@ -534,6 +534,11 @@ export const createGroupChat = async (
   await expect(chatUserB).toBeVisible()
   await page.locator('#new-chat-button').click()
   await page.locator('#newgroup button').click()
+
+  // Name is required
+  await page.locator('.group-name-input').press('Enter')
+  await expect(page.locator('.group-name-input:invalid')).toBeVisible()
+
   await page.locator('.group-name-input').fill(groupName)
   await page.locator('#addmember button').click()
   const addMemberDialog = page.getByTestId('add-member-dialog')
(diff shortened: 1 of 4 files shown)


nicodh (commented):
On my local instance I get this, when the languae is english:

Image

So it seems Chrome (Electron?)  gets the language independant of the selected language

nicodh:
So you plan to fix that? Or was does 👍 mean? 😁

WofWca:
The way I see it is that Electron or to be more precise, Chromium, follows the system language.

I made one search to find if there is an Electron API to set the Chromium language. I couldn't find it.

Maybe such a desync is not super nice, but if there is no trivial way to change that, I suggest to do nothing about this.

r10s:
the tooltip looks a bit strange to the layout otherwise used. we do not use that kind of design anywhere else. i know, it is what electron provides, but this knowledge does not matter :)

ppl looking closely _will_ complain about that language glitch - we know that from android - and we need to deal with the feedback then.

EDIT: otoh, this is not shown often, so maybe okay. discussed already enough :)  at least it removes code, and is more semantic. i would not veto here :)

nicodh (changes requested):
I agrre with @r10s and consider also the language spit should be avoided. So maybe look for a way to achieve accessibilty without changing the UI

WofWca:
Actually, there is a way to change the language, with [`--locale`](https://github.com/electron/electron/blob/main/docs/api/command-line-switches.md#--lang).

PR #6299, opened 2026-04-23
chore: fix missing labels

nicodh:
- temporary readd the missing lables to _untranslated

### DIFF
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -55,5 +55,11 @@
   },
   "who_can_call_me_toggle": {
     "message": "Show call screen on incoming calls"
+  },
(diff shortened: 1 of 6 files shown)


WofWca (approved):
This is a follow-up to https://github.com/deltachat/deltachat-desktop/pull/6289. I believe this adds all the strings that are needed, because otherwise `pnpm check` would error our (and it doesn't, I tested). That MR didn't change `target-electron/` files where we don't have this type checking, so it's all good.

I did not check the .xml file changes.

Also note that this MR doesn't address the `!tx('...')` error, but I propose to remove that piece of code anyway, see https://github.com/deltachat/deltachat-desktop/pull/6300.

PR #6300, opened 2026-04-23
fix: Edit Group dialog a11y and small bugs

WofWca:
- **feat: allow to submit "Edit Group" with "Enter"**
- **refactor: make `DeltaInput` support more props**
- **fix: broken "Group name is required" error msg**

Hide whitespace for easier review.

image

- The error message was invisible.
  Edit: ah I just realized that this is only the case since https://github.com/deltachat/deltachat-desktop/pull/6289.
  
  In the release version of DC it looks OK:
  image

- The error message was not accessible.
- Let's address the TODO comment:
  can't submit the form if the user emptied the group name input.
- Make it possible to submit the form with "Enter".

TODO:
- [ ] When / if this and https://github.com/deltachat/deltachat-desktop/pull/6298 gets approved and merged then we can remove the `please_enter_chat_name` string which was recently added. https://github.com/deltachat/deltachat-android/blob/af49018911ff296a11337a7c945dc14cb8364d89/src/main/res/values/strings.xml#L548

Related:
- https://github.com/deltachat/deltachat-desktop/pull/6298.
- 125fdd62a45c841f3cae3d48cb81262cf302fd82
  (https://github.com/deltachat/deltachat-desktop/pull/5805).

### DIFF
--- a/packages/e2e-tests/tests/group-tests.spec.ts
+++ b/packages/e2e-tests/tests/group-tests.spec.ts
@@ -397,7 +397,15 @@ test('Edit group profile from context menu and rename group', async () => {
   await page.getByTestId('edit-group').click({ force: true })
   // open edit group profile dialog
   await page.getByTestId('view-group-dialog-header-edit').click()
+
+  // Group name is required
+  await expect(page.locator('#groupname:invalid')).not.toBeVisible()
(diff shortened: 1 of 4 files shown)


nicodh (changes requested):
See #6298

PR #6301, opened 2026-04-23
improvement: error dialog: pretty-print err stack

WofWca:
With JSON.stringify all the line breaks are displayed
not as line breaks but as `\n`,
making the error barely readable.


nicodh (approved):
reviewed by reading

PR #6302, opened 2026-04-23
improvement: on msg error click show `msg.error`

### LINKED ISSUE #6290: show error when tapping the error icon of a message
when tapping the error icon of a message, the text returned by [dc_msg_get_error()](https://c.delta.chat/classdc__msg__t.html#ae5914bf190af647c43676e9b35bad9cd) resp. the corresponding jsonrpcs should be shown in a simple alert.

primary button right should be "OK"

the text should be selectable, so parts can be copied to the clipboard

WofWca:
...instead of opening the "message details" dialog,
which has the error hidden behind "three-dot menu -> Developer",
since 08199e00767f82ce860ad93b10006d700aa1f68a
(https://github.com/deltachat/deltachat-desktop/pull/6156).

Closes https://github.com/deltachat/deltachat-desktop/issues/6290.

image

### DIFF
--- a/packages/frontend/src/components/dialogs/AlertDialog.tsx
+++ b/packages/frontend/src/components/dialogs/AlertDialog.tsx
@@ -16,6 +16,10 @@ export type Props = {
   message: string
   okBtnLabel?: string
   dataTestid?: string
+  dialogComponentProps?: Omit<
+    Parameters<typeof Dialog>[0],
+    'dataTestid' | 'onClose'
+  >
 } & DialogProps
 
 export default function AlertDialog({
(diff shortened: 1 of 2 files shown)


nicodh (changes requested):
While using AlertDialog makes sense we should enable to pass a custom width to AlertDialog since the default 350 px width are annoying for a dialog with that much text. When testing I found maybe 600px appropriate.

Otherwise everything looks fine and works as expected

WofWca:
I have pushed a change to increase the width. However, 600px is not too good because at certain screen widths this causes overflow. Hopefully this is good enough for now?

nicodh (approved):

PR #6304, opened 2026-04-26
improvement: a11y: `alt=""` where img is redundant

WofWca:
Disclaimer: In light of the recent
"let's not work endlessly on accessibility" discussion:
nobody has asked me to make this change.
I'm making this contribution as an individual
and not as a Delta Chat team member.

See https://www.w3.org/WAI/tutorials/images/decision-tree/
These are either decorative or redundant.


r10s (approved):
thanks a lot for cleanup!

nicodh (approved):

PR #6307, opened 2026-04-27
ci: add zizmor

link2xt:
[zizmor](https://github.com/zizmorcore/zizmor) is a linter for github actions workflows, looking for common mistakes and insecure defaults (e.g all workflows having access to github token with pretty much all permissions by default, test workflows can create and delete releases etc.).

Started by copying workflow and configuration
from https://github.com/chatmail/core/

### DIFF
--- a/.github/workflows/basic-tests.yml
+++ b/.github/workflows/basic-tests.yml
@@ -11,13 +11,18 @@ on:
       - '.prettierrc.yml'
       - '.vscode/**'
 
+permissions: {}
+
 jobs:
   check:
     name: Code Validation
     runs-on: ubuntu-latest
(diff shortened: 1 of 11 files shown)


link2xt:
Warnings about archived repos can be ignored for now, they should not show up on PRs. Will be visible on Security tab once PR is merged, then we can decide what to do about them. Most important is that this will prevent bugs like introducing cache into release build workflows or substitution bugs, or some NPM dependency using github token available in the workflow because of default permissions (now disabled in all workflows with `permissions: {}`).

nicodh (approved):
Thanks for bringing that to desktop!

PR #6308, opened 2026-04-28
fix zizmor template expansion issues

nicodh:
Fix all issues listed in https://github.com/deltachat/deltachat-desktop/security/code-scanning

see https://github.com/deltachat/deltachat-desktop/security/code-scanning?query=is%3Aopen+branch%3Afix-zizmor-notes

### DIFF
--- a/.github/workflows/build-preview.yml
+++ b/.github/workflows/build-preview.yml
@@ -63,10 +63,12 @@ jobs:
       - name: LINUX renaming / upload Preperation
         if: runner.os == 'Linux'
         working-directory: packages/target-electron
+        env:
+          BRANCH: ${{ steps.prepare.outputs.branch }}
         run: |
           cd dist
           mkdir preview
-          mv *.AppImage preview/deltachat-desktop.${{ steps.prepare.outputs.branch }}.AppImage
+          mv *.AppImage preview/deltachat-desktop.$BRANCH.AppImage
           ls preview
           cd ..
       - name: MAC electron builder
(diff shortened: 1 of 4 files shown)


.github/workflows/build-preview.yml
```diff
@@ -85,27 +87,33 @@ jobs:
       - name: MAC renaming / upload Preperation
         if: runner.os == 'macOS'
         working-directory: packages/target-electron
+        env:
+          BRANCH: ${{ steps.prepare.outputs.branch }}
```
WofWca: In order to not define this var for each step one can move the assignment to the job level (one up).

.github/workflows/rust-tauri-lint.yml
```diff
@@ -56,21 +51,19 @@ jobs:
             ~/.cargo/git/db/
             target/
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-      - uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af # v1.0.7
+      - uses: actions-rust-lang/setup-rust-toolchain@2b1f5e9b395427c92ee4e3331786ca3c37afe2d7 # v1.16.0
```
WofWca: Why use a different one? Why not `dtolnay/rust-toolchain` again?

.github/workflows/rust-tauri-lint.yml
```diff
@@ -56,21 +51,19 @@ jobs:
...
           mkdir packages/target-tauri/html-dist
           touch packages/target-tauri/html-dist/tauri_main.html
-      - uses: actions-rs/clippy-check@b5b5f21f4797c02da247df37026fcd0a5024aa4d # v1.0.7
+      - name: Run clippy
         env:
           VERSION_INFO_GIT_REF: ${{ github.ref }}
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-          args: --workspace --tests --examples
+        run: cargo clippy --workspace --tests --examples
```
WofWca, line 66: I believe this will not add annotations on GitHub, like the action would?
I'm OK with this though, not a big deal.

WofWca (commented):
This fixes straight up injections, which is great!
Plus also addresses some other "action dependency" warnings.

I didn't check if the "upload preview to delta.chat" steps and Tauri workflows still work, but upload to GitHub artifacts does work (apart from a small thing that broke in https://github.com/deltachat/deltachat-desktop/pull/6307).

Side note: our usage of such env vars e.g. as an argument to `mv` is still concerning (looks like path traversal and stuff can happen).

WofWca (approved):

PR #6309, opened 2026-04-28
chore: avoid pnpm ignored build warnings

nicodh:
To get rid of the:
"Warning 
Ignored build scripts: electron-winstaller@5.4.0.                           
Run "pnpm approve-builds" to pick which dependencies should be allowed to run scripts."

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,6 @@
       ]
     },
     "onlyBuiltDependencies": [
-      "@parcel/watcher",
       "electron",
       "esbuild"
     ],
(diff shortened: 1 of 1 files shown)


WofWca (commented):
For `electron-winstaller` this makes sense.
Regarding `@parcel/wathcer`: have you checked if `pnpm watch` still works when changing SCSS files?

nicodh:
> For `electron-winstaller` this makes sense. Regarding `@parcel/wathcer`: have you checked if `pnpm watch` still works when changing SCSS files?

Yes it still works. There are several prebuilt binaries shipped as optional dependencies

WofWca (approved):
I have confirmed that on Linux `pnpm watch:electron` prints new log messages when one modifies a file in `packages/frontend/scss`.
I did not test on Windows or Mac.

Otherhwise also good.

PR #6313, opened 2026-04-29
Adapt settings

### LINKED ISSUE #6333: remove legacy options
the following legacy options are no longer needed as its core settings have no meaning any longer:
- "Move automatically to DeltaChat Folder"
- "Only Fetch from DeltaChat Folder"
- "Show Classic Emails"
- "Delete Messages from Server" (cave: "Delete Messages from Device" will stay as-is!)

nicodh:
Update settings

~~Add a configuration for imap folder. Not sure what would be an appropriate wording here~~

edit: 
resolves #6333
- imap folder UI is removed

 - [x]  removed "Move automatically to DeltaChat Folder"
 - [x]  removed "Only Fetch from DeltaChat Folder"
 - [x]  removed "Show Classic Emails"
 - [x]  removed "Delete Messages from Server"
 - [x] add "force unencrypted"


r10s:
let's simply go for "IMAP Folder" also for the far more important stuff we do not say what is used when it is "Default".

the whole dialog is not meant to be understandable without lots of background. we should also not point new ppl there nor make it outstanding otherwise

"IMAP Folder" is only meant to allow existing user to add a new relay while continue listening on the old one for a while. 

there were lots of discussions about if that is needed at all, but at the end we decided to go for it :)

r10s:
converted to draft, see recent discussion in release coordination, changes probably needed:
- imap folder should not be exposed in the UI
- remove delete_server_after option
- add "show unencrypted" option (name is not settled)

i suggest to postpone this PR until we have a core

r10s:
ftr, and as discussed in one-to-one, we do not want to have the "force encrypted" option that much in the face. it does not work for most user, and we will otherwise get pushback that one can that easily lower security standard (tho it will not work :). 

only saying, it is no blocker for this PR, and it is not yet fully decided where it will go (maybe to "advanced / relays / three dot menu" or "advanced / relay / edit" tho this is more per-relay)

Simon-Laux:
This is needed to run current core main branch with desktop, without this branch or removing code in working copy it does not run correctly.

r10s:
> ftr, and as discussed in one-to-one, we do not want to have the "force encrypted" option that much in the face. it does not work for most user, and we will otherwise get pushback that one can that easily lower security standard (tho it will not work :).


the option shall go to "Settings / Relays / Edit", see updated https://github.com/chatmail/core/pull/8220 and https://github.com/deltachat/deltachat-android/pull/4431

nicodh:
closed in favour of #6409 and #6410

link2xt:
#6409 did not add "force encryption" setting, still need to add it somewhere.

PR #6317, opened 2026-04-29
fix: Log dialog overflowing screen

WofWca:
The problem was that if the screen width is more than 500px
(`$default-dialog-width`) but is less than `width` prop
of the Log dialog (which is 800) then the dialog would be 800px wide.

This also should fix it for all future dialogs,
and IMO also simplifies the code a little.

Related:
- 68ac03bc9da067d6a65bddd5f244b4f98221d032
  (https://github.com/deltachat/deltachat-desktop/pull/4806).


nicodh (approved):
Reviewed by reading & testing

PR #6324, opened 2026-05-02
fix: detect images dropped with uppercase extensions (e.g. .JPG)

### LINKED ISSUE #6323: dragging an image to the main window sends it as file sometimes
since two years, since https://github.com/deltachat/deltachat-desktop/pull/4278 , dragging an image to the main window used to send the image as DC_MSG_IMAGE - so compressed, without original filename or metadata.

recent release (macOS, 2.49), however, sends it unexpectedly as DC_MSG_FILE again.

/me was just sending >5000k instead of 300k :)

0xdeadd:
Closes #6323.

## What

The drag-and-drop handler in `MessageListAndComposer.tsx` compares `file.ext` against a lowercase `IMAGE_EXTENSIONS` list. If a user drops a file with an uppercase extension like \`photo.JPG\`, the \`includes\` check fails and the file is sent as a raw \`DC_MSG_FILE\` instead of a compressed \`DC_MSG_IMAGE\`.

## Why

Per [@WofWca's diagnosis on the issue](https://github.com/deltachat/deltachat-desktop/issues/6323), the comparison is case-sensitive. Most operating systems treat file extensions case-insensitively for typing/identification purposes, so \`.JPG\`, \`.Jpg\`, and \`.jpg\` should all be detected as images.

## Fix

Lowercase \`file.ext\` before the \`includes\` check:

\`\`\`diff
 function isImage(file: ParsedPath) {
-  return IMAGE_EXTENSIONS.map(ext => '.' + ext).includes(file.ext)
+  return IMAGE_EXTENSIONS.map(ext => '.' + ext).includes(file.ext.toLowerCase())
 }
\`\`\`

The \`IMAGE_EXTENSIONS\` constant in \`packages/shared/constants.ts\` is already lowercase (\`['jpg', 'jpeg', 'png', 'apng', 'gif', 'webp']\`), so normalizing the input side is the minimal fix.

### DIFF
--- a/packages/frontend/src/components/message/MessageListAndComposer.tsx
+++ b/packages/frontend/src/components/message/MessageListAndComposer.tsx
@@ -90,7 +90,7 @@ export function getBackgroundImageStyle(
 }
 
 function isImage(file: ParsedPath) {
-  return IMAGE_EXTENSIONS.map(ext => '.' + ext).includes(file.ext)
+  return IMAGE_EXTENSIONS.map(ext => '.' + ext).includes(file.ext.toLowerCase())
 }
 
 export default function MessageListAndComposer({ accountId, chat }: Props) {


nicodh (approved):
Reviewed by reading.

Thanks for the contribution!

PR #6325, opened 2026-05-02
Improve audio player UI

nafirdev:
## Summary

This PR improves the audio player UI to make it more compact, minimal, and theme-aware.

## Changes

- Updated the audio player layout to better match chat voice-message bubbles
- Moved play/pause controls into the cover area
- Made the player more compact
- Improved styling to inherit colors from the surrounding message theme
- Added metadata extraction support for audio duration handling
- Updated the forced-muted audio player behavior

## Notes

The component now avoids fixed background/text colors where possible and uses theme-aware styling so it works better across light and dark message bubbles.

### DIFF
--- a/packages/frontend/src/components/AudioPlayer/extractAudioMetadata.ts
+++ b/packages/frontend/src/components/AudioPlayer/extractAudioMetadata.ts
@@ -1,0 +1,256 @@
+export type ExtractedAudioMetadata = {
+  title?: string;
+  artist?: string;
+  coverUrl?: string;
+};
+
+function cleanText(value?: string): string | undefined {
+  const cleaned = value?.replace(/\0/g, '').replace(/\s+/g, ' ').trim();
+  return cleaned || undefined;
+}
+
+function readSyncSafeInt(bytes: Uint8Array, offset: number): number {
+  return (
+    ((bytes[offset] & 0x7f) << 21) |
+    ((bytes[offset + 1] & 0x7f) << 14) |
+    ((bytes[offset + 2] & 0x7f) << 7) |
+    (bytes[offset + 3] & 0x7f)
+  );
+}
+
+function readUInt32(bytes: Uint8Array, offset: number): number {
+  return (
+    (bytes[offset] << 24) |
+    (bytes[offset + 1] << 16) |
+    (bytes[offset + 2] << 8) |
+    bytes[offset + 3]
+  ) >>> 0;
+}
+
+function decodeLatin1(bytes: Uint8Array): string {
+  let result = '';
+
+  for (let i = 0; i < bytes.length; i += 1) {
+    result += String.fromCharCode(bytes[i]);
+  }
+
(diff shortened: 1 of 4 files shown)


packages/frontend/src/components/AudioPlayer/styles.module.scss
```diff
@@ -1,11 +1,442 @@
...
-
-  // Overwrite user agent styles to make it look a tiny bit better but the real
-  // solution would be using a custom player
-  &::-webkit-media-controls-panel,
-  &::-webkit-media-controls-enclosure {
-    background-color: rgba(200, 200, 200, 0.3);
+  /*
+    Theme-aware variables:
+    بهتره والد/بابِل پیام این‌ها رو ست کنه.
+    اگر ست نشدن، از رنگ متن ارث‌بری می‌کنیم تا توی تم سفید/تیره بهتر کار کنه.
```
WofWca, line 5: Not many of whoever works with Delta Chat can understand this.

packages/frontend/src/components/AudioPlayer/extractAudioMetadata.ts
```diff
@@ -0,0 +1,256 @@
...
+  const blob = new Blob([imageBytes], { type: mimeType });
+
+  return URL.createObjectURL(blob);
+}
+
+function isValidFrameId(frameId: string): boolean {
+  return /^[A-Z0-9]{4}$/.test(frameId);
+}
+
+export async function extractAudioMetadata(src: string): Promise<ExtractedAudioMetadata> {
```
WofWca, line 144: We probably don't want to be maintaining a metadata decoder in our code base. I'm sure someone else has already implemented it (e.g. the Chromium team).

WofWca (commented):
Thanks for taking some time to work on Delta Chat.
Unfortunately this seems like a bit too much code for the benefits outlined in the description, and overall the MR seems underbaked (see e.g. non-English comments and strings). I am afraid trying to get this MR to a mergeable state is not going to be productive.

packages/frontend/src/components/AudioPlayer/styles.module.scss
nafirdev: Sorry about that, this was accidentally left from my local testing. I’m still getting used to the contribution workflow, and I’ll make sure all comments/descriptions are written in English or removed if unnecessary.

packages/frontend/src/components/AudioPlayer/extractAudioMetadata.ts
nafirdev: Thanks for the feedback, that makes sense. I added this only to extract the audio duration/metadata for the player UI, but I agree that maintaining a custom metadata decoder in the codebase is not ideal.

I’ll look for a simpler solution using existing browser/Chromium capabilities, or remove this part if it’s not needed for the UI change.

PR #6329, opened 2026-05-04
fix: mark visible messages as read after a programmatic scroll completes

### LINKED ISSUE #6050: DC_EVENT_MSGS_NOTICED is not always emitted when reading a chat on desktop
- Operating System (Linux/Mac/Windows/iOS/Android): MacOS
- Delta Chat Version: v2.35.0
- Expected behavior: When a chat is read it should propagate the noticed state to all devices 
- Actual behavior: DC_EVENT_MSGS_NOTICED is not always received on the iOS side
  - Receive a message
  - Open chat
  - No DC_EVENT_MSGS_NOTICED received on other devices

nicodh:
resolves #6050 

`onUnreadMessageInView` skips processing when `scheduler.isLocked('scroll')` is true or when the window does not have focus. Messages rendered into the viewport during one of those windows would have their intersection callbacks dropped, so they stayed unread until the user scrolled or the window was re-focused.

This commit re-initializes the Observers after unlockScroll() which causes the IntersectionObservers to deliver an entry on the next task with the current intersection state, which is handled by `onUnreadMessageInView`.

### DIFF
--- a/packages/frontend/src/components/message/MessageList.tsx
+++ b/packages/frontend/src/components/message/MessageList.tsx
@@ -32,43 +32,6 @@ import {
 } from '../../contexts/RovingTabindex'
 import { marknoticedChat } from '../../backend/chat'
 
-const onWindowFocus = (accountId: number) => {
-  log.debug('window focused')
-  const messageElements: HTMLElement[] = Array.prototype.slice.call(
-    document.querySelectorAll('#message-list .message-observer-bottom')
-  )
-
-  const visibleElements = messageElements.filter(el => {
-    const rect = el.getBoundingClientRect()
-    return (
-      rect.top >= 0 &&
-      rect.left >= 0 &&
-      rect.bottom <=
-        (window.innerHeight || document.documentElement.clientHeight) &&
-      rect.right <= (window.innerWidth || document.documentElement.clientWidth)
-    )
-  })
-
-  const messageIdsToMarkAsRead = visibleElements
-    .map(el =>
-      el.dataset.messageid ? Number.parseInt(el.dataset.messageid) : undefined
-    )
-    .filter(id => id != undefined)
-
-  if (messageIdsToMarkAsRead.length !== 0) {
-    log.debug(
-      `window was focused: marking ${messageIdsToMarkAsRead.length} visible messages as read`,
-      messageIdsToMarkAsRead
-    )
-    // FYI we also listen for `MsgsNoticed` event
-    // to update the badge counter,
-    // so `.then(debouncedUpdateBadgeCounter)` is probably not necessary.
-    BackendRemote.rpc
-      .markseenMsgs(accountId, messageIdsToMarkAsRead)
-      .then(throttledUpdateBadgeCounter)
-  }
-}
-
 /**
  * Returns a "live" version of `FullChat.freshMessageCounter`.
  * When the `chat` reference updates, we consider it to be the most up-to-date
@@ -228,11 +191,27 @@ export default function MessageList({
     })
   }, [onUnreadMessageInView])
 
+  // re-initializes the observer on all messages that are currently in the viewport
+  // and unread, so that they get marked as read if they are still in the viewport after
+  // the lock is released or the window regains focus. This is needed since on some
+  // conditions (e.g. window not focused, scroll lock) the `onUnreadMessageInView`
+  // callback returns early without marking messages as read, even if they are in view.
+  const reArmAllUnreadObservations = useCallback(() => {
+    const elements = document.querySelectorAll(
+      '#message-list .message-observer-bottom'
+    )
+    for (const el of elements) {
+      // Calling `unobserve()` followed by `observe()` schedules a fresh entry
+      // for the next IntersectionObserver delivery with the current state.
+      unreadMessageInViewIntersectionObserver.unobserve(el)
+      unreadMessageInViewIntersectionObserver.observe(el)
+    }
+  }, [unreadMessageInViewIntersectionObserver])
+
   useEffect(() => {
-    const onFocus = onWindowFocus.bind(null, accountId)
-    window.addEventListener('focus', onFocus)
-    return () => window.removeEventListener('focus', onFocus)
-  }, [accountId])
+    window.addEventListener('focus', reArmAllUnreadObservations)
+    return () => window.removeEventListener('focus', reArmAllUnreadObservations)
+  }, [reArmAllUnreadObservations])
 
   useEffect(() => {
     return () => {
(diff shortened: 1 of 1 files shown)


nicodh:
I tried to make an e2e test for the scenario described in https://github.com/deltachat/deltachat-desktop/issues/6050#issuecomment-4368719710 but the test was too flaky so I didn't add it.

It would be nice to simplify the scroll section in MessageList. Do we really want to support "other" browsers just for the case Tauri might be used some when? (see comments about webkit)

packages/frontend/src/components/message/MessageList.tsx
```diff
@@ -228,11 +191,27 @@ export default function MessageList({
     })
   }, [onUnreadMessageInView])
 
+  // re-initializes the observer on all messages that are currently in the viewport
+  // and unread, so that they get marked as read if they are still in the viewport after
+  // the lock is released or the window regains focus. This is needed since on some
+  // conditions (e.g. window not focused, scroll lock) the `onUnreadMessageInView`
+  // callback returns early without marking messages as read, even if they are in view.
```
WofWca: Our line width limit is 80.

packages/frontend/src/components/message/MessageList.tsx
```diff
@@ -228,11 +191,27 @@ export default function MessageList({
...
+  const reArmAllUnreadObservations = useCallback(() => {
```
```suggestion
  const markVisibleMessagesAsRead = useCallback(() => {
```

And TBH I would also completely remove the comment above this function. It makes things look way more comlpex than they really are. The fact that we call this on window focus is obvious from the code just below; the "early return" thing is mentioned inside the other newly added comment in `setTimeout`.

packages/frontend/src/components/message/MessageList.tsx
```diff
@@ -574,6 +553,10 @@ export default function MessageList({
         // let's invoke `onScroll`, e.g. to load more messages if we're close
         // to top / bottom
         onScroll()
+        // By the time this nested `setTimeout` runs, the `setTimeout(scheduler.unlock, 0)`
+        // queued by `unlockScroll()` above has already fired, so re-arming the
+        // observer now schedules a fresh delivery with the lock released.
+        reArmAllUnreadObservations()
```
WofWca: So the thing with "wait for lock to be released" is simply because there is an early return inside the observer callback? Otherwise we could mark all visible messages as read basically as soon as the message list is loaded? And we wouldn't need this if we used the existing `onWindowFocus()`, right? The comment could be simplified then it seems. Just say "because callback will do nothing if scroll is locked"?

packages/frontend/src/components/message/MessageList.tsx
```diff
@@ -228,11 +191,27 @@ export default function MessageList({
...
+  // the lock is released or the window regains focus. This is needed since on some
+  // conditions (e.g. window not focused, scroll lock) the `onUnreadMessageInView`
+  // callback returns early without marking messages as read, even if they are in view.
+  const reArmAllUnreadObservations = useCallback(() => {
+    const elements = document.querySelectorAll(
+      '#message-list .message-observer-bottom'
+    )
+    for (const el of elements) {
+      // Calling `unobserve()` followed by `observe()` schedules a fresh entry
+      // for the next IntersectionObserver delivery with the current state.
```
```suggestion
      // Calling `unobserve()` followed by `observe()`
      // will invoke the observer callback for this element.
```

WofWca (commented):
Good that you found a way to reproduce the bug! It seems quite bad, I wonder how long it existed for??

So the way I see it, the functional part of this MR is just

```diff
diff --git a/packages/frontend/src/components/message/MessageList.tsx b/packages/frontend/src/components/message/MessageList.tsx
index 5889ad12b..9a8af76ca 100644
--- a/packages/frontend/src/components/message/MessageList.tsx
+++ b/packages/frontend/src/components/message/MessageList.tsx
@@ -574,9 +574,11 @@ export default function MessageList({
         // let's invoke `onScroll`, e.g. to load more messages if we're close
         // to top / bottom
         onScroll()
+        onWindowFocus(accountId)
       }, 0)
     }, 0)
   }, [
+    accountId,
     chat,
     onScroll,
     unlockScroll,
```

plus a refactor where `onWindowFocus` is replaced with the existing `IntersectionObserver` callback, right?

WofWca:
And yes, this seems to fix the bug, according to my testing.

WofWca:
> It would be nice to simplify the scroll section in MessageList.

The thing is that I'm not sure if that code really is unnecessary even on Chromium.

nicodh:
> plus a refactor where `onWindowFocus` is replaced with the existing `IntersectionObserver` callback, right?

yes, where the refactoring is done by introducing a new function to re initialize the observer

packages/frontend/src/components/message/MessageList.tsx
WofWca: Or rather this seems to be simply about "when loading finished", i.e. basically what the MR title says.

packages/frontend/src/components/message/MessageList.tsx
nicodh: No I added this comment since it is not obvious just from looking at the code why it is needed. The fact that the callback is skipped under certain conditions and that unobserve/observe is needed to activate the observer again should be mentioned explicitly. 

The whole scroll handling has too few explaining comments imo.

And I don't agree on renaming the function to markVisibleMessagesAsRead since the name should reflect somehow that the unreadMessageInViewIntersectionObserver is first initialized inside the MessageWrapper  

But reArm could also be misinterpreted maybe - so let's say reActivate or refreshUnreadObservations ?

nicodh: I decided to go with reInvokeAllUnreadObservations

WofWca: > markVisibleMessagesAsRead since the name should reflect somehow that the unreadMessageInViewIntersectionObserver is first initialized inside the MessageWrapper

By "initialized" you mean that `observe()` gets called? I don't see why it matters.

nicodh: Since in the next refactoring someone might think this is needed, we already called observer() in MessageWrapper

WofWca: Ah now I see. But I don't think that. IDK, to me it doesn't seem like the rename would cause that. It's obvious from the usage of the function and the comments around that usage.

WofWca (approved):
I think that the code is much simpler than the comments and the function names make it seem, but I won't insist on changing that. Future generations can read my comments here hopefully.

Tested again: the bug is indeed fixed.
I don't think this can cause much to break. Invoking the observer callback some extra times won't hurt.

The removal of somewhat duplicate code is also nice.

PR #6330, opened 2026-05-04
fix: show appropriate buttons when qr code is pasted

### LINKED ISSUE #6294: Join channel shows wrong label
If I paste an invite code from a channel in the global search I see only "New Contact" instead "Join Channel" (which is what I see for group invite codes)

nicodh:
resolves #6294

image

image

### DIFF
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -61,5 +61,11 @@
   },
   "edit_channel": {
     "message": "Edit Channel"
+  },
+  "qr_link_pasted": {
+    "message": "QR link detected"
+  },
+  "paste_in_qr_scanner": {
+    "message": "Click to paste it in QR scanner"
   }
 }
--- a/packages/frontend/src/components/chat/ChatList.tsx
+++ b/packages/frontend/src/components/chat/ChatList.tsx
@@ -361,6 +361,13 @@ export default function ChatList(props: {
   // Render --------------------
   const tx = useTranslationFunction()
 
+  let showChatResults = true
+
+  if (chatListIds.length === 0 && queryStr && isInviteLink(queryStr)) {
+    // Don't show "0 chats" when the query is an invite link
+    showChatResults = false
+  }
+
   if (queryChatId) {
     return (
       <SearchInChatResults
@@ -379,7 +386,7 @@ export default function ChatList(props: {
       <AutoSizer disableWidth>
         {({ height }) => (
           <>
-            {isSearchActive && (
+            {isSearchActive && showChatResults && (
               <div
                 id='search-result-divider-chats'
                 className='search-result-divider'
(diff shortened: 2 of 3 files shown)


_locales/_untranslated_en.json
```diff
@@ -61,5 +61,11 @@
   },
   "edit_channel": {
     "message": "Edit Channel"
+  },
+  "qr_link_pasted": {
+    "message": "QR link detected"
+  },
+  "paste_in_qr_scanner": {
+    "message": "Click to paste it in QR scanner"
```
WofWca, line 69: Kinda confusing, too technical? The user has a link here and not a QR, they might have no idea what QR we're talking about and why they'd want to paste it anywhere. Maybe just "Open Link" with no subtitle will work? Or less "scary" "Check link". Or maybe even just "Link".

packages/frontend/src/components/chat/ChatList.tsx
```diff
@@ -361,6 +361,13 @@ export default function ChatList(props: {
   // Render --------------------
   const tx = useTranslationFunction()
 
+  let showChatResults = true
+
+  if (chatListIds.length === 0 && queryStr && isInviteLink(queryStr)) {
+    // Don't show "0 chats" when the query is an invite link
```
WofWca, line 367: So based on the code there can be a situation where we hide the "0 chats" divider but still keep the "0 messages" and "0 contacts" dividers, or vice versa.
Maybe things can be simplified, i.e. we could have the button rendered inside of `ChatList` directly and not inside of `ContactAndMessageSearchResults`?

WofWca (commented):
It turns out that we only show this buttons if it's an invite link, i.e. we don't show it for 'backup', 'account' etc, so basically the only part of this MR that affects production with current Core is the added handling of channel invite links.

https://github.com/deltachat/deltachat-desktop/blob/c2b4bdada95f7666ba423ee08bb829ec25bc1ba8/packages/frontend/src/components/chat/ChatList.tsx#L629-L630

Edit: but this seems to work overall. "Enter" key still works. And the removal of those dividers is nice IMO.

_locales/_untranslated_en.json
nicodh: Well if someone has such a link the recommendation would be to paste it manually into the QR scanner. If it is misunderstood and pasted here we show that hint.So I would assume someone who doesn't know what to do with such a link might end pasting it here.

The last case will only happen if a user pastes a invite link to an own group/channel invitelink 

Since that is a very rare case anyway I don't think more is here to be considered.

packages/frontend/src/components/chat/ChatList.tsx
nicodh: Not sure if that "simplifies" things. We still would have to hide these zero results only if query is an invite link.

I think it's fine as it is

_locales/_untranslated_en.json
WofWca: Pasting a link into a QR scanner makes little sense.

WofWca (approved):
I would adjust the wording. Otherwise looks good.
The topic of handling more types of QRs is for another MR anyway, I just wanted to point this out.

_locales/_untranslated_en.json
nicodh: The QR Scanner is a dialog that has a Paste button - so that is what is described here

WofWca: I understand. My point still stands.

PR #6334, opened 2026-05-04
fix: resize reactions after image load

### LINKED ISSUE #5858: Reactions are collapsed even though it appears as there would be enough space for them visually
Image

saw this a few times already.

version: 295c9ad79f7ac80b5fd8840aa29ddb1500fdf6a1

nicodh:
resolves #5858 

The included reactions e2e test is also a general test to make sure reactions work as expected


packages/e2e-tests/tests/reactions-image-collapse.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+      const imgs = document.querySelectorAll<HTMLImageElement>(
+        '.message.outgoing .msg-body img.attachment-content'
+      )
+      if (imgs.length === 0) return false
+      const img = imgs[imgs.length - 1]
+      return img.complete && img.naturalWidth > 0
+    },
+    undefined,
+    { timeout: 15_000 }
+  )
```
WofWca, line 209: It would be nice to verify that the image is initially "collapsed" so that we know that the `route()` delay code is working.

WofWca (approved):
The code looks good, as well as the refactor. It makes sense to use `ResizeObserver` instead of the window resize observer, as we're already doing in other places.

Edit: I, however, haven't checked if this really fixes the issue completely.

packages/e2e-tests/tests/reactions-image-collapse.spec.ts
WofWca: Better yet, verify that the reactions are initially collapsed.

nicodh: added a test for the inital image width and the collapsed reactions. Also tested that without the route delay these tests fail

PR #6335, opened 2026-05-04
feat: calls & channels are not experimental anymore

### LINKED ISSUE #6332: get "channel" and "calls" out of experimental
"channels" and "calls" should be available by default, they're no longer experimental and do not need a dedicated option any longer

nicodh:
resolves #6332

### DIFF
--- a/packages/e2e-tests/tests/group-tests.spec.ts
+++ b/packages/e2e-tests/tests/group-tests.spec.ts
@@ -585,18 +585,6 @@ test('create channel and add members', async ({ browserName }) => {
 
   await switchToProfile(page, userA.id)
 
-  // Enable the experimental Channels feature in settings (if not already enabled)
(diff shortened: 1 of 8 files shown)


WofWca:
Wouldn't this be the time to also remove the "Eruda" debugger thing from calls-webapp?

WofWca (approved):
I have checked that these old settings are no longer referenced anywhere, smoke-tested channels and calls. The channels are already covered by automatic tests (noice).
The CI failure is unrelated to the changes.

Overall I think it's time to un-experimental calls (even though they're still not in a perfect state IMO). Can't say much for channels, I don't use them much. But I heard they're being used by some people, so this means they're at least not horribly broken.
I couldn't recall any blocking TODOs for calls.

The Eruda thing can be addressed separately, and I'd say it isn't blocking.

PR #6336, opened 2026-05-04
fix: hide apps for channels

### LINKED ISSUE #6331: do not show app picker for channels
most app do not work today in channels

adding a .xdc via file should continue to work, to allow experiments

the corresponding tab in the gallery should stay as is

nicodh:
resolves #6331


WofWca (approved):
I'd prefer a more declarative approach, but yep, this should work (didn't test).

PR #5811, opened 2025-12-04
refactor: apply `type="button"` to all buttons

WofWca:
We do not utilize `<form>`s at all currently (well, just once),
so this is basically a no-brainer change.

This allows us to introduce `<form>` elements fearlessly,
ensuring that no button will act as a "submit" button
without our intent, see e.g. this recent MR:
https://github.com/deltachat/deltachat-desktop/pull/5805.

I have checked that prior to all the changes `pnpm check:lint`
returned 84 `react/button-has-type` errors and after my changes
`git diff | grep "type='button'" | wc -l` returned also 84.
I have also skimmed through all the changes in ~2 minutes to see that they look sane-ish.

Edit: I will rebase https://github.com/deltachat/deltachat-desktop/pull/5805 on top of main when this MR is merged if needed.

PR #6338, opened 2026-05-05
fix: add `type="button"`: accidental form submission

WofWca:
Follow-up to 10912f1978edd90416a5a8461cd457b5e91c7a8d
(https://github.com/deltachat/deltachat-desktop/pull/5811).

This resulted in a bug at least in the "Create Group" dialog
where clicking on the avatar would submit the form.

Unfortunately the `react/button-has-type` linter rule
can't handle such cases where a dynamic tag is used.


nicodh (approved):
Reviewed by reading

PR #6340, opened 2026-05-06
streamline notification settings

r10s:
this PR:

- uses correct translation for "Calls" option
- moves most interesting options atop, advanced options down
- all options are now same "uncheck to get less disturbance"
- very first option is now to disable/mute notifications for the profile; even if it is not called "mute" any longer, but got a clear name, it clearly shows the relationship with "mute" used elsewhere, and is no longer hidden in some other option
- streamlines options with other clients
- removes the current/all profile notion [^1]

## before / after:

[^1]: i known, this has some fans, some even want to have separated settings dialogs or use that notion also elsewhere. but most non-advanced user just do not care, so that adds questions, and does not answer much. as a rule of thumb, most options are per-profile, or should be. the ones who are not usually do not raise this question

### DIFF
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -53,9 +53,6 @@
   "prevent_dialog_spam_checkbox": {
     "message": "Don't allow the app to open this dialog again"
   },
-  "who_can_call_me_toggle": {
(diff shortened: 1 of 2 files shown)


nicodh (approved):
Reviewed by reading & testing

PR #6341, opened 2026-05-06
Improve crash behaviour

### LINKED ISSUE #3592: Renderer stops working and shows black screen after some time
In the console I see
```
Error sending from webFrameMain:  Error: Render frame was disposed before WebFrameMain could be accessed
    at WebFrameMain.send (node:electron/js2c/browser_init:2:94617)
    at WebContents.send (node:electron/js2c/browser_init:2:79767)
    at Object.send (/home/user/src/deltachat/deltachat-desktop/src/main/windows/main.ts:172:24)
    at AccountManager.<anonymous> (/home/user/src/deltachat/deltachat-desktop/src/main/deltachat/controller.ts:99:18)
```

Menu is still displayed, I can click File->Settings to get more errors like this.

This I select Quit and it shutdowns cleanly:
```
523.8s [i]main/index: Starting app shutdown process
523.8s [D]main/index: mainWindow.window.on('close')
523.9s [i]main/state: Saving state to /home/user/.config/DeltaChat/config.json
524.4s [i]main/index: Quitting now. Bye.
```

This is on Linux with Wayland (river wayland compositor).

nicodh:
closes #3592

Since we can't really fix the reason for the crash we can at least improve the behaviour: first reload the window hoping that will restore the app (although we loose state) . If it's not possible to reload show a dialog that a crash occured and (hopefully) a useful error message.

image

### DIFF
--- a/packages/target-electron/src/windows/main.ts
+++ b/packages/target-electron/src/windows/main.ts
@@ -129,6 +129,7 @@ export function init(options: { hidden: boolean }) {
     // crashed the main process
     setImmediate(() => {
       if (mainWindow.isDestroyed()) {
+        log.warn('skipping renderer reload: its window is destroyed')
         return
       }
       try {
--- a/packages/target-electron/src/windows/main.ts
+++ b/packages/target-electron/src/windows/main.ts
@@ -34,6 +34,11 @@ type ExtendedBrowserWindow = BrowserWindow & {
 
 export let window: ExtendedBrowserWindow | null = null
 
+let rendererGone = false
+let lastRendererReloadAttempt = 0
+let crashDialogOpen = false
+const RENDERER_RELOAD_COOLDOWN_MS = 30_000
+
 export function init(options: { hidden: boolean }) {
   if (window) {
     return window.show()
@@ -99,6 +104,43 @@ export function init(options: { hidden: boolean }) {
     // Prevent drag-and-drop from navigating the Electron window, which can happen
     // before our drag-and-drop handlers have been initialized.
     e.preventDefault()
+  })
+
+  // Recover from a dead renderer (like OOM-killed) by reloading once.
+  // Without this the BrowserWindow stays up but its frame is gone, so every
+  // subsequent webContents.send throws "Render frame was disposed".
+  // See https://github.com/deltachat/deltachat-desktop/issues/3592
+  mainWindow.webContents.on('render-process-gone', (_event, details) => {
+    log.error('renderer process gone', details)
+    if (details.reason === 'clean-exit') {
+      // Normal shutdown
+      return
+    }
+    rendererGone = true
+    const now = Date.now()
+    if (now - lastRendererReloadAttempt < RENDERER_RELOAD_COOLDOWN_MS) {
+      log.warn('renderer process gone again within cooldown, prompting user')
+      promptUserAfterRendererCrash(mainWindow, details)
+      return
+    }
+    lastRendererReloadAttempt = now
+    log.info('attempting one-shot reload to recover renderer')
+    // Defer the reload: without the defer webContents.reload()
+    // crashed the main process
+    setImmediate(() => {
+      if (mainWindow.isDestroyed()) {
+        return
+      }
+      try {
+        mainWindow.webContents.reload()
+      } catch (error) {
+        log.error('failed to reload renderer:', error)
+      }
+    })
+  })
+
+  mainWindow.webContents.on('did-finish-load', () => {
+    rendererGone = false
   })
 
   const saveBounds = debounce(() => {
@@ -227,6 +269,47 @@ export function hide() {
   window?.hide()
 }
 
+async function promptUserAfterRendererCrash(
+  win: BrowserWindow,
+  details: electron.RenderProcessGoneDetails
+) {
+  if (crashDialogOpen) {
+    return
+  }
+  crashDialogOpen = true
+  try {
+    const { response } = await electron.dialog.showMessageBox(win, {
+      type: 'warning',
+      title: 'Delta Chat had a problem',
+      message:
+        'The window stopped responding and could not recover automatically. Reload to try again, or quit the app.',
+      detail: `Reason: ${details.reason} (exit code ${details.exitCode})`,
+      buttons: ['Reload', 'Quit'],
+      defaultId: 0,
+      cancelId: 1,
+    })
+    if (response === 1) {
+      electron.app.quit()
+      return
+    }
+    // Reset cooldown so the next crash gets one more automatic recovery
+    // attempt before re-prompting.
+    lastRendererReloadAttempt = 0
+    if (win.isDestroyed()) {
+      return
+    }
+    try {
+      win.webContents.reload()
+    } catch (error) {
+      log.error('failed to reload renderer after user prompt:', error)
+    }
+  } catch (error) {
+    log.error('failed to show renderer-crash dialog:', error)
+  } finally {
+    crashDialogOpen = false
+  }
+}
+
 export function send(channel: string, ...args: any[]) {
   if (!window) {
     log.warn("window not defined, can't send ipc to renderer")
@@ -234,6 +317,13 @@ export function send(channel: string, ...args: any[]) {
   }
   if (window.isDestroyed()) {
     log.info('window is destroyed. not sending message', args)
+    return
+  }
+  if (rendererGone) {
+    // Drop quietly until the reload finishes (or, if reload was suppressed by
+    // the cooldown, until the user restarts the app). Otherwise every menu
+    // click would throw "Render frame was disposed" again.
+    log.info('renderer process is gone, not sending message', { channel })
     return
   }
   try {
@@ -335,7 +425,7 @@ export function toggleDevTools() {
 }
 
 export function chooseLanguage(locale: string) {
-  window?.webContents.send('chooseLanguage', locale)
+  send('chooseLanguage', locale)
 }
 
 export function setZoomFactor(factor: number) {
--- a/packages/target-electron/src/windows/main.ts
+++ b/packages/target-electron/src/windows/main.ts
@@ -282,7 +282,7 @@ async function promptUserAfterRendererCrash(
       type: 'warning',
       title: 'Delta Chat had a problem',
       message:
-        'The window stopped responding and could not recover automatically. Reload to try again, or quit the app.',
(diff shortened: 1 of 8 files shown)


packages/target-electron/src/windows/main.ts
```diff
@@ -101,6 +106,43 @@ export function init(options: { hidden: boolean }) {
     e.preventDefault()
   })
```
nicodh, line 108: For testing insert this here and trigger a crash with Ctrl+Shift+F12


```suggestion
    // TEMP: devmode-only accelerator to force a renderer crash for testing the
  // render-process-gone recovery path. Press Ctrl+Shift+F12 with the window
  // focused. Remove before merging.
  if (process.argv.includes('--devmode')) {
    mainWindow.webContents.on('before-input-event', (_event, input) => {
      if (
        input.type === 'keyDown' &&
        input.control &&
        input.shift &&
        input.key === 'F12'
      ) {
        log.warn('DEBUG: forcefully crashing renderer for testing')
        mainWindow.webContents.forcefullyCrashRenderer()
      }
    })
  }
```

packages/target-electron/src/windows/main.ts
```diff
@@ -227,6 +269,47 @@ export function hide() {
...
+  }
+  crashDialogOpen = true
+  try {
+    const { response } = await electron.dialog.showMessageBox(win, {
+      type: 'warning',
+      title: 'Delta Chat had a problem',
+      message:
+        'The window stopped responding and could not recover automatically. Reload to try again, or quit the app.',
+      detail: `Reason: ${details.reason} (exit code ${details.exitCode})`,
+      buttons: ['Reload', 'Quit'],
```
WofWca, line 288: At least "Quit" can be internationalized.

packages/target-electron/src/windows/main.ts
```diff
@@ -227,6 +269,47 @@ export function hide() {
...
+  if (crashDialogOpen) {
+    return
+  }
+  crashDialogOpen = true
+  try {
+    const { response } = await electron.dialog.showMessageBox(win, {
+      type: 'warning',
+      title: 'Delta Chat had a problem',
+      message:
+        'The window stopped responding and could not recover automatically. Reload to try again, or quit the app.',
```
WofWca:
```suggestion
        'The window stopped responding and could not recover automatically.',
```

This is redundant. Buttons are already saying this.

WofWca (commented):
Sounds like a reasonable workaround.

Have you investigated what other apps do? Maybe something common.

nicodh:
> Sounds like a reasonable workaround.
> 
> Have you investigated what other apps do? Maybe something common.

No I haven't investigated.

packages/target-electron/src/windows/main.ts
nicodh: I think it's ok to let these rare error messages to be english only. People are used to that and a mix of languages might look like an error itself

packages/target-electron/src/windows/main.ts
```diff
@@ -101,6 +106,43 @@ export function init(options: { hidden: boolean }) {
...
+  // Without this the BrowserWindow stays up but its frame is gone, so every
+  // subsequent webContents.send throws "Render frame was disposed".
+  // See https://github.com/deltachat/deltachat-desktop/issues/3592
+  mainWindow.webContents.on('render-process-gone', (_event, details) => {
+    log.error('renderer process gone', details)
+    if (details.reason === 'clean-exit') {
+      // Normal shutdown
+      return
+    }
+    rendererGone = true
```
WofWca, line 119:
```suggestion
    if (details.reason === 'clean-exit') {
      // Normal shutdown
      return
    }
    log.error('renderer process gone', details)
    rendererGone = true
```

If `clean-exit` then this is not an error I guess? Move the long entry.

packages/target-electron/src/windows/main.ts
```diff
@@ -101,6 +106,43 @@ export function init(options: { hidden: boolean }) {
...
+      promptUserAfterRendererCrash(mainWindow, details)
+      return
+    }
+    lastRendererReloadAttempt = now
+    log.info('attempting one-shot reload to recover renderer')
+    // Defer the reload: without the defer webContents.reload()
+    // crashed the main process
+    setImmediate(() => {
+      if (mainWindow.isDestroyed()) {
+        return
```
WofWca, line 133:
```suggestion
        log.warn('skipping renderer reload: its window is destroyed')
        return
```

When do we expect this to happen? Or is this basically just defensive coding? If so, let's add a log entry?

packages/target-electron/src/windows/main.ts
```diff
@@ -101,6 +106,43 @@ export function init(options: { hidden: boolean }) {
...
+    }
+    rendererGone = true
+    const now = Date.now()
+    if (now - lastRendererReloadAttempt < RENDERER_RELOAD_COOLDOWN_MS) {
+      log.warn('renderer process gone again within cooldown, prompting user')
+      promptUserAfterRendererCrash(mainWindow, details)
+      return
+    }
+    lastRendererReloadAttempt = now
+    log.info('attempting one-shot reload to recover renderer')
```
WofWca, line 127: Are we sure that we want to always attempt automatic restart, even if the reason is "`memory-eviction` - Process proactively terminated to prevent a future out-of-memory (OOM) situation"?
https://www.electronjs.org/docs/latest/api/app#event-child-process-gone
Maybe it's better to assume by default that there is a good reason for the process to be gone?
I'm inclined to think that showing a notification and quitting would be more appropriate at least in this case?

I also considered simply ignoring the "renderer gone" and only attempting restart when the window gets focused, but an issue here would be that the user might think that the app is still working whereas it actually doesn't even show notifications for incoming messages (because the data flows through the renderer process).

WofWca (commented):
Tested this for a bit by killing the process via task manager. Seems to work as intended. Though I haven't tried OOM.
This doesn't really solve the root cause (and IDK if it's solvable), but probably enough to close the issue.

packages/target-electron/src/windows/main.ts
nicodh: Well we are at least in a flaky state here: the main window should be closed before the event render-process-gone is emitted. At least in my tests I don't see this error log message.

```
16.4s [i]main/index: Starting app shutdown process
16.4s [D]main/index: mainWindow.window.on('close')
16.4s [i]main/state: Saving state to /home/ndh/projekte/deltachat/data/config.json
16.4s [i]/mainWindow: window is destroyed. not sending message [
  '{"jsonrpc":"2.0","id":267,"result":[{"contextId":0,"event":{"kind":"Info","msg":"src/accounts.rs:358: Stopping IO for all accounts."}},{"contextId":3,"event":{"kind":"Info","msg":"src/scheduler.rs:109: stopping IO"}},{"contextId":3,"event":{"kind":"ConnectivityChanged"}},{"contextId":3,"event":{"kind":"Info","msg":"src/imap/idle.rs:81: \\"INBOX\\": Received interrupt, stopping IDLE."}},{"contextId":3,"event":{"kind":"Info","msg":"src/scheduler.rs:747: Shutting down SMTP loop."}},{"contextId":3,"event":{"kind":"Info","msg":"src/imap/idle.rs:81: \\"INBOX\\": Received interrupt, stopping IDLE."}},{"contextId":3,"event":{"kind":"Info","msg":"src/scheduler.rs:409: Transport 1: Shutting down inbox loop."}},{"contextId":3,"event":{"kind":"Info","msg":"src/scheduler.rs:409: Transport 2: Shutting down inbox loop."}}]}'
]
16.4s [i]core/event: 0 src/accounts.rs:358: Stopping IO for all accounts.
16.4s [i]core/event: 3 src/scheduler.rs:109: stopping IO
16.4s [D]core/event: 3 ConnectivityChanged { kind: 'ConnectivityChanged' }
16.4s [i]core/event: 3 src/imap/idle.rs:81: "INBOX": Received interrupt, stopping IDLE.
16.4s [i]core/event: 3 src/scheduler.rs:747: Shutting down SMTP loop.
16.4s [i]core/event: 3 src/imap/idle.rs:81: "INBOX": Received interrupt, stopping IDLE.
16.4s [i]core/event: 3 src/scheduler.rs:409: Transport 1: Shutting down inbox loop.
16.4s [i]core/event: 3 src/scheduler.rs:409: Transport 2: Shutting down inbox loop.
16.9s [i]main/index: Quitting now. Bye.

```

packages/target-electron/src/windows/main.ts
nicodh: I would say there is a lot of speculation why which error happens. In my experience Chrome is often using a lot of memory and a restart reduces that. So I think trying to restart at least once has a good chance to recover the app.

WofWca: > I would say there is a lot of speculation why which error happens.

Well at least it's quite possible that the process can get killed due to OOM? In fact that's the best and most likely explanation we have for the bug that we're trying to solve? If we're getting OOM-killed then it's an OOM situation and we probably shouldn't be acting as if Delta Chat is the most important process on the system. A fresh reload is also not light on the CPU and disk.

WofWca (approved):
Apart from the fact that I think we shouldn't be always auto-restarting, looks OK.

packages/target-electron/src/windows/main.ts
```diff
@@ -227,6 +270,47 @@ export function hide() {
...
+async function promptUserAfterRendererCrash(
+  win: BrowserWindow,
+  details: electron.RenderProcessGoneDetails
+) {
+  if (crashDialogOpen) {
+    return
+  }
+  crashDialogOpen = true
+  try {
+    const { response } = await electron.dialog.showMessageBox(win, {
```
WofWca, line 282: Also I'm not sure if popping up a message box out of the blue is nice. Maybe, again, a notification would be better.

nicodh: Well we are in a scenario where the app shows a black screen - notification or dialog - what is your point?

packages/target-electron/src/windows/main.ts
nicodh: As said above: I would assume a renderer restart to be similar to a Chrome restart and that in my experience often reduces the memory consumption significantly

packages/target-electron/src/windows/main.ts
WofWca: The app could be in the background. `showMessageBox` steals focus, but a notification doesn't.

PR #6342, opened 2026-05-06
chore: remove obsolete eslint directives

nicodh:
Avoid eslint warnings


r10s (approved):
came exactly over that today  - and was wondering how my setup could be broken :)

PR #6349, opened 2026-05-07
fix: align profile name in settings

### LINKED ISSUE #6343: settings name/bio should be left-aligned
the name and the bio excerpt (first line is fine) should be left-aligned to the avatar. the padding is fine.

this is the layout we're also using at similar places, eg. chatlist or contact list.

take care, to not change alignment in profile (maybe this is related somehow)

nicodh:
resolves #6343


WofWca (approved):
Better!

PR #6350, opened 2026-05-07
test: reduce flakiness in composer.spec

nicodh:
The composer tests often fail. Locally even more often than in the CI.

These changes seem to reduce that flakiness (at least in my local tests).

Note: the changes were strongly based on AI recommendations

### DIFF
--- a/packages/e2e-tests/tests/composer.spec.ts
+++ b/packages/e2e-tests/tests/composer.spec.ts
@@ -216,7 +216,15 @@ test.describe('draft', () => {
       .getByRole('button', { name: 'Add attachment' })
       .click()
     await page.getByRole('menuitem', { name: 'Contact' }).click()
-    await page.getByRole('dialog').getByRole('button', { name: 'Me' }).click()
+    const dialog = page.getByRole('dialog')
+    // Filter by current accounts display name so only the
+    // self-contact ("Me") remains.
+    await dialog.locator('input').fill('Alice')
+    const selfContactItem = dialog
+      .getByRole('listitem')
+      .filter({ hasText: 'Me' })
+    await expect(selfContactItem).toBeVisible()
+    await selfContactItem.getByRole('button').click()
   }
   async function testDraftHasFile() {
     const myName = 'Alice'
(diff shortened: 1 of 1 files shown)


packages/e2e-tests/tests/composer.spec.ts
```diff
@@ -579,6 +586,7 @@ test.describe('Ctrl + Up shortcut', () => {
   })
 
   test('removes quote on Ctrl + Down', async () => {
+    await textarea.focus()
```
WofWca, line 589: Doesn't make much sense to only do this in one test and not the others?

packages/e2e-tests/tests/composer.spec.ts
```diff
@@ -292,18 +300,17 @@ test.describe('draft', () => {
...
-    // Make sure that we have actually switched the chat
-    // and loaded its draft.
-    await textarea.fill('foo')
-    await currChat!.click()
-    await expect(textarea).toBeEmpty()
+    await otherChatTab.click()
+    // Wait for the other chat to fully load before going back.
+    await expect(textarea).not.toBeDisabled()
+    await currChatHandle!.click()
+    await expect(composerSection).toHaveText('')
```
WofWca, line 313: Do you know what this change actually does?

packages/e2e-tests/tests/composer.spec.ts
```diff
@@ -216,7 +216,15 @@ test.describe('draft', () => {
...
       .click()
     await page.getByRole('menuitem', { name: 'Contact' }).click()
-    await page.getByRole('dialog').getByRole('button', { name: 'Me' }).click()
+    const dialog = page.getByRole('dialog')
+    // Filter by current accounts display name so only the
+    // self-contact ("Me") remains.
+    await dialog.locator('input').fill('Alice')
+    const selfContactItem = dialog
+      .getByRole('listitem')
+      .filter({ hasText: 'Me' })
```
WofWca, line 225: Do you know if this is actually useful?

WofWca (approved):
I don't see how this can help except maybe for that `focus()` but if you say that it indeed helps then alright.

packages/e2e-tests/tests/composer.spec.ts
nicodh: That was the most often failing test and that fixed it. So it made sense to add it.

packages/e2e-tests/tests/composer.spec.ts
nicodh: "Previously this used `textarea.fill('foo')` as a probe, but that is racy: React's debounced draft save can fire after the chat switch and write 'foo' into the original chat's draft."

WofWca: I don't see what's racy about this.

nicodh: I started this just as a side effect since I stumbled again into the flakiness of these tests https://github.com/deltachat/deltachat-desktop/actions/runs/25478542258/job/74757399965#step:6:1139

It might make sense to dive deeper into this and find out why it was flaky but for now I would like to either apply these changes or skip the composer tests

packages/e2e-tests/tests/composer.spec.ts
nicodh: As said above: this fixed another failure. I didn't keep the reports but I remember there was an error. We could try to remove it again and then test often enough to make sure it's not flaky, but I would say it doesn't harm

packages/e2e-tests/tests/composer.spec.ts
WofWca: Yep, I'm sorry about writing them so flaky.

I already approved the MR. If it really does help then let's merge. Whoever needs to change that code later can read our discussion here.

PR #6353, opened 2026-05-07
fix: contiguous multiselect: handle item removal

WofWca:
This changes a couple of things:
1. When the "from" item has been removed,
   Shift + Click will only select the "to" item.
2. Handle the absence of the "to" item before handling the absence
   of the "from" item (move that piece of code).
3. When the "to" item is absent, return and do nothing.

This will be useful for the upcoming message multiselect,
where item deletion and contiguous selection
is perhaps more common than that for chats.

### DIFF
--- a/packages/e2e-tests/tests/chat-list-multiselect.spec.ts
+++ b/packages/e2e-tests/tests/chat-list-multiselect.spec.ts
@@ -194,7 +194,7 @@ test.describe('Shift + Click', () => {
     await expectSelectedChats([9, 8, 7])
   })
 
-  test('handles selection start chat getting removed', async () => {
+  test("doesn't break if the item at the start of the selection gets removed", async () => {
     const chatName = 'Chat to be removed'
     await createDummyChat(page, chatName)
     const chat = chatList.getByRole('tab', { name: chatName })
--- a/packages/e2e-tests/tests/chat-list-multiselect.spec.ts
+++ b/packages/e2e-tests/tests/chat-list-multiselect.spec.ts
@@ -8,6 +8,7 @@ import {
   reloadPage,
   test,
   createNDummyChats,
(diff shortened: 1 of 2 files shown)


packages/e2e-tests/tests/chat-list-multiselect.spec.ts
```diff
@@ -192,6 +193,25 @@ test.describe('Shift + Click', () => {
     // This is the topmost chat: do nothing.
     await expectSelectedChats([9, 8, 7])
   })
+
+  test('handles selection start chat getting removed', async () => {
```
nicodh: Not clear what is meant here. 
Seems to be Handle deleting the first of selected chats but wording is more like removing a chat from selection ??

nicodh (approved):
Reviewed by reading & testing

Works as expected, deleting a selected chat from other device does not break anything.

Test should be renamed imo

packages/e2e-tests/tests/chat-list-multiselect.spec.ts
WofWca: Yep I couldn't come up with a less terrible description when I wrote it. Maybe this?

```suggestion
  test('can handle chat at the selection start getting deleted', async () => {
```

nicodh: "selection start getting deleted" is still a bit ambigous: does it mean at the "selection start" or "start getting deleted" which would be avoided with "the first of selected chats getting deleted"

But both wordings are ok, so choose what you like

WofWca:
```suggestion
  test("doesn't break if the item at the start of the selection gets removed", async () => {
```

PR #6354, opened 2026-05-07
chore: pin pnpm version with packageManager

nicodh:
To avoid failures in CI like today after a new pnpm release.

Means also we have to upgrade manually from time to time

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,5 +87,6 @@
     "typescript-eslint": "^8.51.0",
     "xml-js": "^1.6.11",
     "yaml": "^2.8.3"
-  }
+  },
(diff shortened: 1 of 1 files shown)


WofWca (approved):
Would be better to use `devEngines.packageManager` https://pnpm.io/package_json#devenginespackagemanager instead of this legacy field. I opened a WIP MR for that: https://github.com/deltachat/deltachat-desktop/pull/6355.
Or simply upgrading to support pnpm v11, as long as it's fine by our FreeBSD distribution downstream.

While that is in the works, let's fix CI with this MR.

PR #6356, opened 2026-05-08
fix: truncate author name in quotes

### LINKED ISSUE #5901: style: should truncate author name in quoted msg
- Delta Chat Version:
- Expected behavior: the name doesn't take too much space
- Actual behavior: it does
  1. Set a long alias for a contact
  2. Reply to their message
- Screenshots:
  
  Image

- Logs:

We already do the same in groups for the main message (not quotes). We should do the same for quotes. I tried simply applying `white-space: nowrap`, but it causes the messages to stretch and cause horizontal scroll, see #5161. Maybe this can be easily done when that issue is closed.

https://github.com/deltachat/deltachat-desktop/blob/9fabfe67ca6f10ec54e811cba23c1a6a74157c84/packages/frontend/scss/message/_message.scss#L582-L586

https://github.com/deltachat/deltachat-desktop/blob/9fabfe67ca6f10ec54e811cba23c1a6a74157c84/packages/frontend/scss/message/_message.scss#L660-L679

https://github.com/deltachat/deltachat-desktop/blob/10912f1978edd90416a5a8461cd457b5e91c7a8d/packages/frontend/src/components/message/Message.tsx#L980-L1013

Related:
- https://github.com/deltachat/deltachat-desktop/pull/2747

nicodh:
resolves #5901

I see no horizontal scrolls

image

### DIFF
--- a/packages/frontend/scss/message/_message.scss
+++ b/packages/frontend/scss/message/_message.scss
@@ -558,6 +558,10 @@ $info-text-max-width: 400px;
     text-overflow: ellipsis;
   }
 
+  .quote-author {
+    white-space: nowrap;
+  }
+
   .quoted-text {
     font-size: 12px;
     line-height: 12px;


WofWca (approved):
Looks like this should work. I tried this at different screen ~~heights~~ widths. Works both in messages and in the composer.

PR #6358, opened 2026-05-08
fix: adapt width to new MacOS traffic lights

### LINKED ISSUE #5898: Traffic lights in macOS shifted
I noticed that the macOS traffic lights are no longer centered in the profile switcher and the green light protrudes slightly into the search field.

macOS 26.2
DC-Desktop 2.35.0

Image

nicodh:
resolves #5898
Account list sidebar is now 8px wider on MacOS

image

### DIFF
--- a/packages/frontend/src/components/AccountListSidebar/AccountListSidebar.tsx
+++ b/packages/frontend/src/components/AccountListSidebar/AccountListSidebar.tsx
@@ -148,9 +148,13 @@ export default function AccountListSidebar({
     return <div></div>
   }
 
+  const isMac = runtime.getRuntimeInfo().isMac
+
   return (
-    <div className={styles.accountListSidebar}>
-      {runtime.getRuntimeInfo().isMac && !smallScreenMode && (
+    <div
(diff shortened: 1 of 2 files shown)


r10s (changes requested):
i do not have macos26 here, so i cannot really test as well.

but in general, the new layout change looks a bit clumsy, as we now lots of space left/right - but atop things are very close together. also first not sure about the much fatter "selection" indicator, but once one get used to it, it is fine.

also, that we now have different layouts for mac/non-mac will cause trouble in the future.

i suggest the following:

- make the bar a little bit wider on all OS - where does the 8px come from? maybe also 4px or 6px are sufficient?
- see the fat selection indicator as a feature, iirc, ppl were complaining here and there that the selection is not visible enough
- hover on the selected profile should not make the selection indicator _smaller_ - this was also an issue before, but now it becomes far more visible
- set padding _atop_ of the first icon to the same as left/right

again, do not do that special for macos, do that for all systems. that way maintainance is easier, esp. as few of us have macos

nicodh:
closed in favor of #6380

PR #6359, opened 2026-05-08
test: more reliable multiselect tests

WofWca:
The problem with `toContainText()` is that
`expect([1,2,3]).toContainText([1,3])` passes.


nicodh (approved):
Reviewed by reading

PR #6360, opened 2026-05-08
fix: settings: align bio to start

WofWca:
It's not aligned if the name is longer than bio
(or the "Bio" placeholder).
For `<button>`s `text-align: center;` is the default,
that is what causes it.

Follow-up to https://github.com/deltachat/deltachat-desktop/pull/6349.


nicodh (approved):
Reviewed by reading

PR #6362, opened 2026-05-08
chore: bump required Node version in package.json

WofWca:
To make pnpm stop warning about unsupported engine.
I have tested `pnpm dev` on Node 26. Works fine.


nicodh (approved):
reviewed by reading

PR #6363, opened 2026-05-09
fix: HTML viewer: show sent time instead of receive time

### LINKED ISSUE #6344: Show full message in Saved Messages displays epoch 0 timestamp
- Summary:
When expanding a very long text message in Saved Messages by clicking “show full message”, the displayed timestamp changes to epoch 0 (1970-01-01 or similar) instead of the correct message time. The timestamp remains correct when the message is collapsed, and in the chat list.

Mac

- Delta Chat Version:
v2.48.0

- Expected behavior:
The timestamp should remain unchanged and display the correct message time, whether collapsed or expanded.

- Actual behavior:
After expanding, the timestamp incorrectly changes to epoch 0. Collapsing restores the correct time.

1. Open DeltaChat and go to Saved Messages (chat with yourself).
2. Send a very long text message (long enough to be truncated with a “show full message” button).
3. Observe the timestamp - it shows the correct sent time initially.
4. Tap/click “show full message” to expand the entire text.
5. Look at the timestamp again - it now shows epoch 0 (e.g., Thursday, January 1, 1970 [timezone] AM).

- Screenshots:
Image

- Logs:

- Additional Observations
1. Timestamp is correct in chat list view.
2. Timestamp is correct in message info/details panel.
3. Bug occurs only in Saved Messages (not in chats with other contacts).
4. Bug occurs only while expanded; collapsing fixes it.

nicodh:
resolves #6344

### DIFF
--- a/packages/frontend/src/components/message/messageFunctions.ts
+++ b/packages/frontend/src/components/message/messageFunctions.ts
@@ -96,9 +96,14 @@ export async function openMessageHTML(messageId: number) {
     sender: { displayName },
     subject,
     chatId,
+    fromId,
+    timestamp,
     receivedTimestamp,
   } = await BackendRemote.rpc.getMessage(accountId, messageId)
-  const receiveTime = moment(receivedTimestamp * 1000).format('LLLL')
+  const isFromSelf = fromId === C.DC_CONTACT_ID_SELF
+  const receiveTime = moment(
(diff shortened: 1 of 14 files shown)


r10s (changes requested):
see comment in the issue: just always show sent-timestamp

r10s (approved):

PR #6364, opened 2026-05-09
Fix double group image

nicodh:
The duplicate code was inserted using githubs "apply suggestion" function.

The second commit fixes a missing margin in the ImageCropper dialog


WofWca (approved):
The first is trivial.
The latter is good. Also on short screens this margin would go inside the scrollable area, not taking space unnecessarily, also good.

Image

PR #6366, opened 2026-05-11
improvement: extra confirm if selected > 100 msgs

WofWca:
This might not be the best solution UX-wise,
but it's better than nothing.

image


r10s (changes requested):
this additional alert is not helpful, even if we improve the wording.

instead, we should improve the original PR:

- show the impact of the number of selected messages in the title during selection already. this is what android/iOS are doing. 



that way, the user sees at any point, how many messages are selected, esp. after scrolling, when the selected messages are out of view, if shift-tap works etc.

as another iteration, some time later with experience, we can consider to iterate. but i really doubt that is needed, we have multi-selection on other platforms since years.

WofWca:
Replied in https://github.com/deltachat/deltachat-desktop/issues/2892#issuecomment-4422987687.

PR #6369, opened 2026-05-12
improvement: show num msgs in "Forward?" dialog

WofWca:
~~We don't have a string specifically for this,
like we do for "delete N messages?",
so let's reuse the existing "N selected" for now.~~

Supersedes https://github.com/deltachat/deltachat-desktop/pull/6366.

### DIFF
--- a/packages/e2e-tests/tests/forward-message.spec.ts
+++ b/packages/e2e-tests/tests/forward-message.spec.ts
@@ -138,9 +138,7 @@ test('forward message to another chat (same account, with confirmation)', async 
   // Confirmation dialog should appear
   const confirmDialog = page.getByTestId('confirm-dialog')
   await expect(confirmDialog).toBeVisible()
-  await expect(confirmDialog).toContainText(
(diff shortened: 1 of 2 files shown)


WofWca:
I will make an MR to add the string.

WofWca:
> I will make an MR to add the string.

Actually, I'm not sure if we need so many strings? That is a pluralized "Forward N messages?", plus a pluralized "Forward N messages to M chats?". Only for English that would be 4 strings to translate.
Maybe this MR's approach is good enough.

r10s:
i added a nice string to https://github.com/deltachat/deltachat-android/pull/4427 , it replaces the old forward string, so that's fine, esp. as this is one if the strings shown more often to the user

nicodh:
Since the translations are available so maybe import them with this PR?

nicodh (changes requested):
If this uses the new 'ask_forward_messages' string for the translation it could be merged

nicodh (approved):
Reviewed by reading & testing

PR #6370, opened 2026-05-12
refactor: remove unused `top: -121px;` CSS

WofWca:
The element has `position: static;` so it has no effect.
And it's anyway not good to specify position like this.
This line is lingering around since over 7 years,
I have not investigated when it stopped having an effect.

image


nicodh (approved):
Reviewed by reading

PR #6371, opened 2026-05-12
fix: Ctrl+Click on msg image to multiselect

WofWca:
Make the interactive elements inside a message inert (unclickable)
when multiselect "mode" is active.

I had to change `getMessage` in tests
because originally it was pointing to the message text,
and clicking it without `force` would not work in Playwrtight,
so I had to select some element outside of `inert`.

The CSS change should not break anything,
because `.author-avatar` is only used as "message author avatar",
and not e.g. also in VCards.

WIP because I want to take time to consider alternative solutions. But yep, this works.
- [ ] Hmmmm, but https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/inert#accessibility_concerns says
  > hidden from assistive technologies as they are excluded from the accessibility tree
  
  I believe we don't want this, because we want to read the message contents when it gets focused.
  Edit: just confirmed with Orca reader. It only announces "opens menu" when you switch between messages if at least one message is selected.
- [ ] If we go for this then the selection text color CSS is not needed?

### DIFF
--- a/packages/e2e-tests/tests/message-list-multiselect.spec.ts
+++ b/packages/e2e-tests/tests/message-list-multiselect.spec.ts
@@ -1,4 +1,5 @@
 import { expect, type Page } from '@playwright/test'
+import path from 'path'
 
 import {
   createProfiles,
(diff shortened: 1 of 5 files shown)


WofWca:
Superseded by https://github.com/deltachat/deltachat-desktop/pull/6395.

PR #6372, opened 2026-05-12
fix: small screen, closing main window should unselect chat

account1009:
Small screen: closing main window should unselect chat (and go back to chat list)

### DIFF
--- a/packages/frontend/src/components/RuntimeAdapter.tsx
+++ b/packages/frontend/src/components/RuntimeAdapter.tsx
@@ -29,7 +29,7 @@ export default function RuntimeAdapter({
 }: PropsWithChildren<Props>) {
   const processQr = useProcessQr()
   const { jumpToMessage } = useMessage()
-  const { selectChat } = useChat()
+  const { selectChat, unselectChat } = useChat()
 
   const { closeDialog, openDialog, closeAllDialogs } = useDialog()
   const openSendToDialogId = useRef<string | undefined>(undefined)
@@ -73,6 +73,10 @@ export default function RuntimeAdapter({
         }
       }
     )
+
+    runtime.onHide = () => {
+      unselectChat()
+    }
 
     runtime.onShowDialog = kind => {
       if (kind === 'about') {
--- a/packages/frontend/src/components/RuntimeAdapter.tsx
+++ b/packages/frontend/src/components/RuntimeAdapter.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef } from 'react'
+import React, { useContext, useEffect, useRef } from 'react'
 
 import useMessage from '../hooks/chat/useMessage'
 import useProcessQr from '../hooks/useProcessQr'
@@ -10,6 +10,7 @@ import { ActionEmitter, KeybindAction } from '../keybindings'
 import useDialog from '../hooks/dialog/useDialog'
 import WebxdcSaveToChatDialog from './dialogs/WebxdcSendToChat'
 import { saveLastChatId } from '../backend/chat'
+import { ScreenContext } from '../contexts/ScreenContext'
 import useChat from '../hooks/chat/useChat'
 import SettingsStoreInstance from '../stores/settings'
 import { SCAN_CONTEXT_TYPE } from '../hooks/useProcessQr'
@@ -29,6 +30,7 @@ export default function RuntimeAdapter({
 }: PropsWithChildren<Props>) {
   const processQr = useProcessQr()
   const { jumpToMessage } = useMessage()
+  const { smallScreenMode } = useContext(ScreenContext)
   const { selectChat, unselectChat } = useChat()
 
   const { closeDialog, openDialog, closeAllDialogs } = useDialog()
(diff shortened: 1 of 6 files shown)


packages/frontend/src/components/RuntimeAdapter.tsx
```diff
@@ -74,6 +74,10 @@ export default function RuntimeAdapter({
       }
     )
 
+    runtime.onHide = () => {
+      unselectChat()
```
nicodh: This should be wrapped into a if(smallScreenMode) condition (provided by useContext(ScreenContext)) since otherwise it would be unexpected that a previously selected chat is suddenly unselected

packages/target-electron/runtime-electron/runtime.ts
```diff
@@ -148,6 +148,7 @@ class ElectronRuntime implements Runtime {
     return !forbiddenPathRegEx.test(file.replaceAll('\\', '/'))
   }
   onThemeUpdate: (() => void) | undefined
+  onHide: (() => void) | undefined
```
nicodh, line 151: If a property is added to runtime the interface needs to be extended and all other runtime implementations need to implement onHide too. Even if they don't provide a handler. See other runtimes as example with not implemented comments

nicodh (changes requested):
With the requested changes this should be fine.

Don't forget to run pnpm check before comitting. It will show you (most) errors/warnings that the CI check would also show.

packages/target-electron/runtime-electron/runtime.ts
account1009: Yes, noticed now and added. I was not running pnpm check to detect these issues

packages/frontend/src/components/RuntimeAdapter.tsx
account1009: I've fixed that part now

r10s:
what exactly does "closing main window" mean? 

eg. when i exit the app, one may also expect to open the same chat on restart, not always the chatlist.

not against the pr, but i lack understanding :)

nicodh:
> what exactly does "closing main window" mean?
> 
> eg. when i exit the app, one may also expect to open the same chat on restart, not always the chatlist.
> 
> not against the pr, but i lack understanding :)

Exit or minimize are both meant here. Yes the expectation might as well be keeping the last selected chat open. I guess it's about the fact that no chatlist is visible then and to see new messages  needs always an additional click first.

I don't use small screens a lot so I can't say what the best practice would be...

account1009:
I've been using deltachat-desktop on a small screen daily in the last 3 weeks so I've proposed this pull request for a common concrete scenario that is happening often.

When I'm interested about a specific conversation I just open another app keeping deltachat not as the main focused window. In this case I explicitly want to keep the conversation opened because maybe I'm switch to another software only temporary.

But in other cases I close the deltachat window. That I reopen deltachat (maybe one hour later). In these cases I don't even remember what was the last conversation. I reopen deltachat to write in another chat or to check if there are unread messages from any of my contacts. In very rare cases I'm interested to stay in the last opened conversation.

I mean: I already have the possibility to keep the deltachat window open (but not in the foreground) if I want to return to the previous conversation. If I close the window it is because I don't care about the last conversation

I hope that now the concrete case is clearer. Let me know if you need additional info :)

r10s:
for minimize one should not change navigation. that is unexpected.

but one can be sceptical also for exit  - as it adds code and compexity, and we also otherwise "keep the view", esp on small devices with low memory, one may exit delta chat often - and expect delta to be opened as before. many ppl also do not differ between minimize and close window

one can surely argue for both, tho. if it is about getting aware of new messages: that is an issue at any point. we can show an "unread badge" on the "back" arrow to fix that unconditional

account1009:
Maybe that part that I've proposed is also what I see tagged around as "catching up with other clients".
On android apps such as deltachat or arcanechat the behavior that I've described is the default one.

I hope that I'm not off topic but I think that deltachat android or arcanechat can be considered "smallscreen" apps

account1009:
> for minimize one should not change navigation. that is unexpected.

For this part I suspect that we are talking about something different and there is a misunderstanding: for example in Gnome Linux if I right click on a window title in the menu with the different actions there is Hide and Close. I'm not talking about the Hide button behavior where is fine to keep the current conversation open but I'm talking about the Close button behavior.

Probably the names that I've used in the code generate confusion (such as 'deltachat-hidden' or onHide)

More info to give the idea:
Menu File » Quit (fine as it is now)
Window manager » Hide (fine as it is now)
Window manager » Close  <----- I'm talking and proposed a change about this behavior

nicodh:
> Menu File » Quit (fine as it is now)
> Window manager » Hide (fine as it is now)
> Window manager » Close <----- I'm talking and proposed a change about this behavior


So if I understand it right: closing the window via Window manager (by clicking on the close button on the window bar) should behave same as closing the app from menu. Ok, but why did you name the new runtime function "onHide"? I think that caused the confusion. So with your change the Window Manager close  function triggers the runtime onHide function??

In general: we try to have as few runtime functions as possible, so in the described scenario it's just a question of how you close the app. If you would use a shortcut like Ctrl+q it would close the app in the expected way.

So to sum it up: we would add a new runtime function for the case that
1. we are on Small Screen
2. the user wants to close the app only by clicking on the Window Close button
3. the user wants to have a fresh state when opening the app again

This seems a very special scenario and there are easy workarounds to achieve the same behaviour (like a shortcut or one click more) so I think we should not extend runtime for that.

PR #6373, opened 2026-05-12
fix: qr code video is flipped for back camera

### LINKED ISSUE #6269: The qr code video is flipped for back camera
The qr code video is flipped horizontally to have a correct behaviour for front cameras

The problem is that when selecting a back camera (instead of a front camera) the horizontal flip has not to be performed

account1009:
Closes https://github.com/deltachat/deltachat-desktop/issues/6269.

The qr code video is flipped horizontally to have a correct behaviour for front cameras

The problem is that when selecting a back camera (instead of a front camera) the horizontal flip has not to be performed

### DIFF
--- a/packages/frontend/src/components/QrReader/index.tsx
+++ b/packages/frontend/src/components/QrReader/index.tsx
@@ -381,13 +381,13 @@ export const QrReader = forwardRef<QrCodeScanRef, Props>(
           activeStream = stream
 
           const settings = videoTracks[0].getSettings()
-          let flipped = true;
-          if (settings.facingMode != undefined) {
-            flipped = !(settings.facingMode == 'environment')
+          let isFacingUser = true
+          if (settings.facingMode !== undefined) {
+            isFacingUser = settings.facingMode === 'user'
           } else {
-            flipped = !(videoTracks[0].label.toLowerCase().includes('back'))
+            isFacingUser = !(videoTracks[0].label.toLowerCase().includes('back'))
           }
-          setFlipped(flipped)
+          setFlipped(isFacingUser)
 
           if (settings.deviceId) {
             setDeviceId(settings.deviceId)
--- a/packages/frontend/src/components/QrReader/index.tsx
+++ b/packages/frontend/src/components/QrReader/index.tsx
@@ -68,7 +68,7 @@ export const QrReader = forwardRef<QrCodeScanRef, Props>(
     const [videoDevices, setVideoDevices] = useState<MediaDeviceInfo[]>([])
     const [deviceId, setDeviceId] = useState<string | undefined>(undefined)
     const [processingFile, setProcessingFile] = useState(false)
-
+    const [flipped, setFlipped] = useState(false)
     const workerRef = useRef<Worker | null>(null)
     const workerClipBoardRef = useRef<Worker | null>(null)
     // Only create the worker once per component instance.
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/QrReader/index.tsx
```diff
@@ -381,6 +381,14 @@ export const QrReader = forwardRef<QrCodeScanRef, Props>(
           activeStream = stream
 
           const settings = videoTracks[0].getSettings()
+          let flipped = true;
```
nicodh: we should now "shadow" the existing state variable with the same name. Better give it a different name here

packages/frontend/src/components/QrReader/index.tsx
```diff
@@ -381,6 +381,14 @@ export const QrReader = forwardRef<QrCodeScanRef, Props>(
...
+          if (settings.facingMode != undefined) {
+            flipped = !(settings.facingMode == 'environment')
```
nicodh: we use strict comparison operators (=== vs. ==) in this project. Eslint will complain here an on similar places

packages/frontend/src/components/QrReader/index.tsx
nicodh, line 330: Is it ok that facingMode here always defaults to 'user' or shouldn't that also be adapted now?

packages/frontend/src/components/QrReader/index.tsx
nicodh: Also note: we skip trailing semicolons as a coding guideline

nicodh (changes requested):
In general this seems a valid approach although I can't test since I have no device with 2 cameras.

Some changes are needed...

packages/frontend/src/components/QrReader/index.tsx
account1009: I've updated with isFacingUser as a variable. Do you mean something like that?

packages/frontend/src/components/QrReader/index.tsx
```diff
@@ -381,6 +381,14 @@ export const QrReader = forwardRef<QrCodeScanRef, Props>(
           activeStream = stream
 
           const settings = videoTracks[0].getSettings()
+          let isFacingUser = true
+          if (settings.facingMode !== undefined) {
+            isFacingUser = settings.facingMode === 'user'
```
WofWca:
```suggestion
            isFacingUser = settings.facingMode !== 'environment'
```

'user' is not the only user-facing mode, apparently.

https://developer.mozilla.org/en-US/docs/Web/API/MediaTrackSettings/facingMode#videofacingmodeenum

packages/frontend/src/components/QrReader/index.tsx
```diff
@@ -68,7 +68,7 @@ export const QrReader = forwardRef<QrCodeScanRef, Props>(
     const [videoDevices, setVideoDevices] = useState<MediaDeviceInfo[]>([])
     const [deviceId, setDeviceId] = useState<string | undefined>(undefined)
     const [processingFile, setProcessingFile] = useState(false)
-
+    const [flipped, setFlipped] = useState(false)
```
WofWca, line 71:
```suggestion
    const [flipped, setFlipped] = useState(false)

```

packages/frontend/src/components/QrReader/index.tsx
WofWca: Good point, IDK firstly why we're requesting a user-facing camera instead environment-facing, secondly why we're requesting `deviceId` and `facingMode` at the same time. This has been like this from the start e2b23d328c1c1f409b35451b89661fe6a13c1613. Maybe someone just forgot to clean up IDK.

But if @account1009 tested this MR and it works then I think it's a topic for a separate MR.

WofWca (approved):
Looks simple.

Edit: But I haven't tested with an environment-facing camera.

packages/frontend/src/components/QrReader/index.tsx
WofWca: Indeed this is buggy. `deviceId` does not take priority for me, and selecting another the camera doesn't work sometimes, because `getUserMedia` returns the camera with another `deviceId`.

account1009: If you are talking about the menu selection of the camera that does not change the camera... I've noticed the same but I've not tried to solve to keep this PR on a single topic. Maybe in the future in another PR I'll check that part to try to understand what is going on

PR #6375, opened 2026-05-12
React zoom pan pinch update

### LINKED ISSUE #6367: Zooming images is very slow
I've noticed that zooming images using the touchpad is very slow. It takes several attempts before the image is enlarged enough. This wasn't always the case. Unfortunately I don't know in which version this change occurred.

macOS 26.4.1
DC-Desktop 2.49.1

nicodh:
resolves #6367

In my tests the updated library pinches much faster then before

the attribute wheelPanning in
panning={{
     wheelPanning: true,
}}

seems to have been removed but I couldn't detect any changes except the improvement of pinching

#public-preview

### DIFF
--- a/packages/frontend/package.json
+++ b/packages/frontend/package.json
@@ -46,7 +46,7 @@
     "react-virtualized-auto-sizer": "^1.0.26",
     "react-window": "^1.8.11",
     "react-window-infinite-loader": "^1.0.10",
-    "react-zoom-pan-pinch": "^3.7.0",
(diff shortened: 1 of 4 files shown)


nicodh:
Seems to fix the issue according to the user feedback

r10s (approved):
i cannot see a difference on my macOS 15 - maybe it gets improved with macos 26.

but it is also not worse as before, code is removed, and we updated the dependency - so makes sense, esp. if issue reporter reports an improvement.

thanks a lot for taking care!

PR #6376, opened 2026-05-13
feat: click on call messages opens/focus call window

### LINKED ISSUE #6081: Tapping a call bubble in chat should answer/start call

nicodh:
resolves #6081 

tbd: currently only a click on the phone icon triggers that behaviour - should we extend that to the whole message bubble as stated in the issue?

### DIFF
--- a/packages/frontend/src/components/message/Message.tsx
+++ b/packages/frontend/src/components/message/Message.tsx
@@ -1285,31 +1285,40 @@ function CallIconButton({
   const refresh = useEffectEvent(callInfoFetch.refresh)
   useEffect(() => {
     return onDCEvent(accountId, 'MsgsChanged', event => {
+      // MsgsChanged event is fired when the call state changes
       if (event.msgId !== messageId) {
         return
       }
+      // update the call info
       refresh()
     })
   }, [accountId, messageId])
 
-  const onClickParams:
-    | undefined
-    | Parameters<typeof runtime.openIncomingVideoCallWindow>[0] =
-    callInfoFetch.result?.ok &&
-    callInfoFetch.result.value.state.kind === 'Alerting'
-      ? {
-          accountId,
-          chatId,
-          callMessageId: messageId,
-          callerWebrtcOffer: callInfoFetch.result.value.sdpOffer,
-          startWithCameraEnabled: callInfoFetch.result.value.hasVideo,
-        }
-      : undefined
-  // TODO fix: don't open a second window if one is already open
-  // for this message (this should be done in the main process).
-  const onClick = onClickParams
-    ? () => runtime.openIncomingVideoCallWindow(onClickParams)
+  const callInfo = callInfoFetch.result?.ok
+    ? callInfoFetch.result.value
     : undefined
+
+  const callWindowParams = callInfo
+    ? {
+        accountId,
+        chatId,
+        callMessageId: messageId,
+        callerWebrtcOffer: callInfo.sdpOffer,
+        startWithCameraEnabled: callInfo.hasVideo,
+      }
+    : undefined
+
+  const onClick =
+    callInfo == undefined
+      ? undefined
+      : callInfo.state.kind === 'Alerting' || callInfo.state.kind === 'Active'
+        ? // Focus the existing window (if any) or open the incoming call dialog.
+          () => runtime.openIncomingVideoCallWindow(callWindowParams!)
+        : // Terminated state (Completed, Missed, Declined, Canceled): start a new outgoing call.
+          () =>
+            runtime.startOutgoingVideoCall(accountId, chatId, {
+              startWithCameraEnabled: callInfo.hasVideo,
+            })
 
   return (
     <IconButton
--- a/packages/target-electron/src/windows/video-call.ts
+++ b/packages/target-electron/src/windows/video-call.ts
@@ -22,12 +22,33 @@ import { setContentProtection } from '../content-protection'
 
 const log = getLogger('windows/video-call')
 
+/**
+ * Tracks open call windows by "accountId-chatId" key.
+ * Used to focus an existing call window when the user clicks a call bubble.
+ */
+const openCallWindows = new Map<string, BrowserWindow>()
+
+function callWindowKey(accountId: number, chatId: number): string {
+  return `${accountId}-${chatId}`
+}
+
 export function startOutgoingVideoCall(
   accountId: number,
   chatId: number,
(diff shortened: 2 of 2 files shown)


packages/frontend/src/components/message/Message.tsx
```diff
@@ -1285,32 +1285,41 @@ function CallIconButton({
   const refresh = useEffectEvent(callInfoFetch.refresh)
   useEffect(() => {
     return onDCEvent(accountId, 'MsgsChanged', event => {
+      // MsgsChanged event is fired when the call state changes
       if (event.msgId !== messageId) {
         return
       }
+      // update the call info
```
WofWca, line 1292: Better to rename the `refresh` function then, if such clarification is necessary?

packages/frontend/src/components/message/Message.tsx
```diff
@@ -1285,32 +1285,41 @@ function CallIconButton({
...
+        startWithCameraEnabled: callInfo.hasVideo,
+      }
+    : undefined
+
+  const onClick =
+    callInfo == undefined
+      ? undefined
+      : callInfo.state.kind === 'Alerting' || callInfo.state.kind === 'Active'
+        ? // Focus the existing window (if any) or open the incoming call dialog.
+          () => runtime.openIncomingVideoCallWindow(callWindowParams!)
```
WofWca, line 1316: The runtime method name (as well as the name of the function in `video-call.ts`) is outdated now.

WofWca, line 1316: The current behavior makes no sense if you've started or accepted this call on another device.
Clicking the button pops up the "incoming call" window (with ringing and stuff). Closing the window will end the call on all devices.

packages/frontend/src/components/message/Message.tsx
```diff
@@ -1285,32 +1285,41 @@ function CallIconButton({
...
+        callMessageId: messageId,
+        callerWebrtcOffer: callInfo.sdpOffer,
+        startWithCameraEnabled: callInfo.hasVideo,
+      }
+    : undefined
+
+  const onClick =
+    callInfo == undefined
+      ? undefined
+      : callInfo.state.kind === 'Alerting' || callInfo.state.kind === 'Active'
```
WofWca, line 1314: We are checking the state of a single call message here. If another bubble gets pressed then that will try to start another call.

packages/target-electron/src/windows/video-call.ts
```diff
@@ -137,6 +158,17 @@ export function openIncomingVideoCallWindow({
...
+  // foreground instead of opening a second window.
+  const existingWin = openCallWindows.get(callWindowKey(accountId, chatId))
+  if (existingWin && !existingWin.isDestroyed()) {
+    if (existingWin.isMinimized()) {
+      existingWin.restore()
+    }
+    existingWin.focus()
+    return
+  }
+
```
WofWca, line 171: Not sure what to do about this, but something I took note of. `openIncomingVideoCallWindow` can get called not only as a result of a button press but also from the `IncomingCall` event listener. So if there is an ongoing call in this chat and another call in this chat gets started by the remote peer, then this will simply focus the original call's window. Maybe that's something that's supposed to be solved in Core, IDK.

WofWca (approved):
Works, for the most trivial cases.

But looks like there is a ton of stuff to consider before we can really achieve the "only one call per chat" goal (without the "call ends by itself" bugs). Maybe worth syncing up with the Android and iOS implementations, or considering moving some of the logic to Core (possibly leading to API changes).
Personal opinion: https://github.com/deltachat/deltachat-desktop/issues/6081 is not that important of a feature to uncover all these edge cases before we've figured out how things should actually work.

> tbd: currently only a click on the phone icon triggers that behaviour - should we extend that to the whole message bubble as stated in the issue?

On Android taping anywhere on the bubble starts the call.
On Telegram only the "phone" icon is clickable.
I think the latter is better, because one probably doesn't expect the whole bubble to be a button.
Also because TBH I don't like how easy it is to accidentally start a call. On Android there isn't even a delay during which you can cancel the call without sending the "incoming call" message to the remote party.

packages/frontend/src/components/message/Message.tsx
nicodh: I'm not sure if we want to support all multi device scenarios here - clicking on call messages in an existing chat while having a call within the same chat on another device is a very rare use case I would say.

But we could of course decide not to open a new window if the status is ringing or active and there is no active call window on desktop

packages/frontend/src/components/message/Message.tsx
nicodh: Not if there is already a call window opened for the same account and chat. Then that will get the focus.

But again: if the call is open on another device that would start a new call but that is also true in the current implementation: if you click on the phone button while having a call with the same user on another device a new call would be started.

So I would argue, this PR improves that at least for calls on the same device

packages/frontend/src/components/message/Message.tsx
nicodh: It's more related to the fact that we should update the call info when we received a message changed event

packages/frontend/src/components/message/Message.tsx
nicodh: You mean since it would just focus to an existing call window if one exists?

I would say that fallback is needed and missing anyway. Right now if you click on the phone icon repeatedly you open lots of new call windows

packages/target-electron/src/windows/video-call.ts
nicodh: > So if there is an ongoing call in this chat and another call in this chat gets started by the remote peer, then this will simply focus the original call's window

But what else would you expect to happen in that case? Btw. with this PR that would not be possible anymore, only if the peer starts the call from a different device than the ongoing call

packages/frontend/src/components/message/Message.tsx
WofWca: I meant that this is not only about _incoming_ calls but also outgoing.
Ah but now I see that on current main it's also not quite right, because the button is clickable also for outgoing calls that are "Alerting" (edit: but that's not intentional).

nicodh:
I will merge this now as it improves the current situation already.

We would need a rpc.getCallInfoForChat to detect if there is an ongoing call for the chat on another device

PR #6379, opened 2026-05-13
fix: show rtl text in correct direction in mixed texts

### LINKED ISSUE #6345: Persian (RTL) text renders incorrectly and inconsistently across operating systems in DeltaChat
- Description:
DeltaChat does not display Persian phrases properly. The rendering behavior varies depending on the operating system. In some cases, characters appear disjointed or in the wrong order. In other cases, the text direction (right-to-left) is ignored, causing punctuation and spaces to appear at the wrong end of the string.

Mac, iOS, Android
(and probably Linux and Windows as well - wasn't able to test)

- Delta Chat Version:
iOS: v2.48.0
Mac: v2.48.0
Android: v2.49.0

- Expected behavior:
All Persian text should render correctly in right-to-left order. Mixed English-Persian sentences should maintain proper bidirectional (bidi) ordering according to Unicode standards. The same message should appear identically across all operating systems where DeltaChat is supported.

- Actual behavior:
The rendering of Persian text is inconsistent. On some OS versions, words and phrases appear separated or reversed. On others, mixed English-Persian lines show incorrect text direction, with punctuation or spaces appearing at the left edge instead of the right.

1. Open DeltaChat on multiple different operating systems (e.g., Android and Windows, or iOS and Linux).
2. Send a message containing Persian+English text from either system to another.
3. Observe the display of the Persian+English phrase on both the sending and receiving devices.

- Screenshots:
Image
Image
Image

- Logs:
- Additional notes:
-- The same Persian text displays correctly in other messaging apps (e.g., Telegram, WhatsApp, Signal) on the same devices.
-- The issue appears OS-dependent rather than solely DeltaChat version dependent.

- Sample combinations to test and compare:

```
1. English only
The quick brown fox jumps over the lazy dog near the riverbank.

2. Persian only
روباه قهوه‌ای سریع از روی سگ تنبل پرید و کنار رودخانه فرود آمد.

3. English with one Persian phrase in the middle
The design process requires توجه به جزئیات for a successful outcome.

4. Persian with one English phrase in the middle
فرآیند طراحی نیاز به attention to detail برای نتیجه موفق دارد.

5. English with multiple Persian phrases in the middle, begin and end with English
We believe that همکاری تیمی and ارتباط شفاف are essential, yet نوآوری مداوم remains the key driver.

6. Persian with multiple English phrases in the middle, begin and end with Persian
به نظر ما teamwork و clear communication ضروری هستند، اما continuous innovation همچنان محرک اصلی باقی می‌ماند.

7. Alternating words (Eng-Persian-Eng-Persian)
This این simple یک mixed جمله pattern الگو works.

8. Persian sentence starting with an English word
Hello به همکاران جدید خوش آمدید. امروز روی پروژه اصلی کار می‌کنیم.

9. English sentence starting with a Persian word
سلام everyone, welcome to the new team. Today we will work on the main project.

10. Fully mixed in short phrases (no clear dominant language)
کاربر باید click کند و then وارد صفحه new شود. بعد از آن، save را بزند.

11. Persian with English inside parentheses
مراحل اجرا (implementation steps) شامل سه فاز (design, testing, deployment) می‌باشد.

12. English with Persian inside parentheses
The final review (بررسی نهایی) will take place on Friday (جمعه) next week.

13. One language as subject, the other as rest of sentence
مدیریت پروژه is the most critical factor for success in agile environments.

14. Persian quote inside English sentence
He kept saying, «فردا کار را انجام می‌دهم», but nothing ever got done.

15. English quote inside Persian sentence
همیشه به خود می‌گفت "I'll do it tomorrow" اما هیچ‌وقت کاری انجام نمی‌شد.
```

nicodh:
resolves #6345 

image

image

### DIFF
--- a/packages/e2e-tests/tests/group-tests.spec.ts
+++ b/packages/e2e-tests/tests/group-tests.spec.ts
@@ -368,12 +368,6 @@ test('Withdraw group invite link', async ({ browserName }) => {
       .getByRole('listitem')
       .filter({ hasText: 'Member Me removed by Alice' })
   ).toBeVisible()
-  await expect(
-    page
-      .getByRole('list', { name: 'Messages' })
-      .getByRole('listitem')
-      .filter({ hasText: groupInviteMessage })
-  ).toBeVisible()
   // the group chat opens but composer should not be available
   // timeout to make sure (or more probable) that this is not
   // just due to a delayed response or a timing issue
--- a/packages/e2e-tests/tests/group-tests.spec.ts
+++ b/packages/e2e-tests/tests/group-tests.spec.ts
@@ -368,6 +368,12 @@ test('Withdraw group invite link', async ({ browserName }) => {
       .getByRole('listitem')
       .filter({ hasText: 'Member Me removed by Alice' })
   ).toBeVisible()
+  await expect(
+    page
+      .getByRole('list', { name: 'Messages' })
+      .getByRole('listitem')
+      .filter({ hasText: groupInviteMessage })
+  ).toBeVisible()
   // the group chat opens but composer should not be available
   // timeout to make sure (or more probable) that this is not
   // just due to a delayed response or a timing issue
--- a/packages/frontend/scss/chat/_chat-list-item.scss
+++ b/packages/frontend/scss/chat/_chat-list-item.scss
@@ -200,6 +200,7 @@
       .text {
         display: inline-block;
         max-width: fit-content;
+        unicode-bidi: plaintext;
       }
     }
   }
--- a/packages/target-electron/src/windows/html_email.ts
+++ b/packages/target-electron/src/windows/html_email.ts
@@ -512,9 +512,6 @@ function makeBrowserView(
   sandboxedView.webContents.insertCSS(`:root {
       color: black;
       background-color: white;
-    }
-    body {
-      unicode-bidi: plaintext;
     }`)
 
   const openLink = (url: string) => {
--- a/packages/target-electron/src/windows/html_email.ts
+++ b/packages/target-electron/src/windows/html_email.ts
@@ -512,6 +512,9 @@ function makeBrowserView(
   sandboxedView.webContents.insertCSS(`:root {
       color: black;
       background-color: white;
+    }
+    body {
+      unicode-bidi: plaintext;
     }`)
 
   const openLink = (url: string) => {
--- a/packages/frontend/scss/message/_message.scss
+++ b/packages/frontend/scss/message/_message.scss
@@ -110,6 +110,7 @@ $info-text-max-width: 400px;
         font-size: 14px;
         line-height: 18px;
         text-align: start;
+        unicode-bidi: plaintext;
 
         overflow-wrap: break-word;
         word-wrap: break-word;
@@ -564,6 +565,7 @@ $info-text-max-width: 400px;
     font-size: 12px;
     line-height: 12px;
     margin: 0;
+    unicode-bidi: plaintext;
     max-height: 40px;
     overflow: hidden;
     line-clamp: 3;
--- a/packages/target-electron/static/electron_html_email_view/electron_html_email_view.css
+++ b/packages/target-electron/static/electron_html_email_view/electron_html_email_view.css
@@ -31,10 +31,12 @@ div.header > .title p {
 
 #subject {
   font-size: 1em;
+  unicode-bidi: plaintext;
 }
 
 div.header > .title > p.details {
   font-size: 0.8em;
+  unicode-bidi: plaintext;
 }
 
 div.header > div.network-toggle {


packages/e2e-tests/tests/group-tests.spec.ts
```diff
@@ -368,12 +368,6 @@ test('Withdraw group invite link', async ({ browserName }) => {
       .getByRole('listitem')
       .filter({ hasText: 'Member Me removed by Alice' })
   ).toBeVisible()
-  await expect(
```
WofWca, line 371: Why do we not expect to see this message? Isn't it a bug that it's not visible? Or maybe it's simply because the user scans the invitation link before they actually receive the "you have been removed from the group" message, in which case the test actually doesn't test what it's supposed to test?

packages/frontend/scss/chat/_chat-list-item.scss
```diff
@@ -200,6 +200,7 @@
       .text {
         display: inline-block;
         max-width: fit-content;
+        unicode-bidi: plaintext;
```
WofWca, line 203: Does it make sense to apply this to elements that are supposed to occupy just one line? Does this actually fix something?
Same for the HTML email viewer header changes.

packages/target-electron/src/windows/html_email.ts
```diff
@@ -512,6 +512,9 @@ function makeBrowserView(
   sandboxedView.webContents.insertCSS(`:root {
       color: black;
       background-color: white;
+    }
+    body {
+      unicode-bidi: plaintext;
```
WofWca: I'm pretty sure that this should be done in Core instead? Because we don't want to be injecting CSS if this is actually an HTML message and not just a long plaintext message.

WofWca (commented):
OK, I took a look at how Telegram does this, and it looks like they take the same approach. Except apparently they also apply `dir="auto"` to such user-generated content.
https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/unicode-bidi#plaintext sounds applicable.

> This keyword makes the elements directionality calculated without considering its parent bidirectional state or the value of the [direction](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/direction) property.

packages/e2e-tests/tests/group-tests.spec.ts
nicodh: >  Isn't it a bug that it's not visible? 

It's not a bug it's a timing issue. In local tests I never experience that error.

And it's obviously not related to the changes of this PR. So I wanted to remove a flakiness.

I have no idea how to make these tests more reliable. The groupInviteMessage should be created by core and not depend on the peer to answer.

packages/frontend/scss/chat/_chat-list-item.scss
nicodh: I just wanted to be on the safe side. What if the text is mixed? And I can't imagine how this could harm here

packages/target-electron/src/windows/html_email.ts
nicodh: Well if it turns out that solves the problem we still can move that part to core.

In Android these mixed direction texts are displayed as expected btw

WofWca: I'm saying that this can break actual HTML messages.

packages/e2e-tests/tests/group-tests.spec.ts
WofWca: I'll submit an MR.

packages/target-electron/src/windows/html_email.ts
nicodh: Can you tell a bit more?
- how can it break HTML messages?
- how can we (would core) avoid it?
- as this obviously fixes the related Issue how else could we fix it?

WofWca: 1. HTML messages are basically messages that already include all the necessary markup and CSS, including all this directionality stuff. We should not be mutating HTML messages' contents, i.e. by injecting this CSS.
2. By injecting this CSS only to the "long truncated plaintext messages", returning that from `rpc.getMessageHtml()`, but not to the messages that are inherently HTML (non-plaintext).
3. By moving this logic to Core. The CSS should be returned from `rpc.getMessageHtml()`. Maybe I'm not seeing an obstacle here but at least this is how it looks like it should be.

Alternative: the Core can expose API that lets us know whether a message is an actual HTML message or a truncated plaintext message. Based on this we decide whether to inject this CSS.

packages/frontend/scss/chat/_chat-list-item.scss
WofWca: > I just wanted to be on the safe side.

Alright, that makes sense. In this case this CSS applies only to the text but not the whole line "author: message".

But in case of the HTML viewer header I think we should apply this CSS to a more narrow scope, i.e. only to the "sender" element, because the date should always be on the right for LTR languages.
But fine, not a big issue.

packages/target-electron/src/windows/html_email.ts
nicodh: > HTML messages are basically messages that already include all the necessary markup and CSS, including all this directionality stuff. We should not be mutating HTML messages' contents, i.e. by injecting this CSS.

I agree but I don't see how this breaks HTML? I would assume the plaintext value applies only to plaintext.

If you know more let me know. We can still add a core fix but as I said I don't see how this hurts.

If there is a real reason we could remove that rule and wait for the core fix

WofWca: By "break" here I mean "change the way the message looks". We should not change the way the message looks.

WofWca: See the closely related MR: https://github.com/chatmail/core/pull/8227, which added `dir="auto"`.

nicodh: I removed the style. I don't think it's worth that long discussion. Let's get it fixed in core

packages/e2e-tests/tests/group-tests.spec.ts
nicodh: I readded the expect statement

WofWca (approved):
We should probably apply this to other user-generated content, such as bio and group description, so this probably doesn't completely close the issue, but is a big improvement in this regard still.

PR #6380, opened 2026-05-14
fix: adapt accountlist sidebar

### LINKED ISSUE #5898: Traffic lights in macOS shifted

nicodh:
Apply suggestions from https://github.com/deltachat/deltachat-desktop/pull/6358#pullrequestreview-4264876403

resolves #5898 

Before

image

After

image

MacOS Tahoe

image

### DIFF
--- a/packages/frontend/src/components/AccountListSidebar/styles.module.scss
+++ b/packages/frontend/src/components/AccountListSidebar/styles.module.scss
@@ -41,8 +41,9 @@
   background-color: var(--accountSidebarBg);
   display: flex;
   flex-direction: column;
-  min-width: 68px;
+  min-width: 76px;
   overflow: hidden;
+  padding-top: 4px;
 
   .accountListNav {
     flex-grow: 1;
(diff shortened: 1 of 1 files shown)


Simon-Laux:
image

The indicator needs to be moved a bit to the left so that you don't see it curve in on the other side. (alternatively no border radius on the left side if having it stick out more was an active decision.)

r10s (approved):
lgtm, much better, esp. that the hover does not get _smaller_ on hovering.

one nitpick: leflt/right of the avatars we have 14px - top/bottom, we have 16px. i looks more coherent when using 14px everywhere. i would have made a code suggestion, but i cannot find that easiy, the following is a gimp mockup:

left: 14px/16px, right: suggestion of 14px everywhere:

Image &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;

PR #6381, opened 2026-05-14
test: wait for join message

nicodh:
Avoid failures like in #6345


nicodh:
closed in favour of #6382

PR #6383, opened 2026-05-14
chore: upgrade Electron

WofWca:
Includes a bunch of Chromium security fixes.

I smoke-tested it. Works.


nicodh (approved):
Makes sense to stay up-to-date with Chrome these days

PR #6385, opened 2026-05-16
fix: set default height smaller then screen height

### LINKED ISSUE #5534: Default window size on Linux can be too large for screen, even when a manual resize can make it fit
On Linux with a KDE Wayland desktop, the window opens up too large for the screen if the screen is below a certain size, even though the window can manually be resized to easily fit.

- Operating System (Linux/Mac/Windows/iOS/Android): Linux Wayland
- Delta Chat Version: Delta Chat Desktop 2.11.1 (git: flathub) Delta Chat Core v2.11.0
- Expected behavior: window fits
- Actual behavior: the bottom can be cut off needlessly like this:
  
  Image

- Screenshots:
- Logs:

nicodh:
Resolves #5534


WofWca (approved):
Looks like this should make things better and resolve the issue, though didn't test.

PR #6386, opened 2026-05-16
feat: show dialog for load image settings

### LINKED ISSUE #6267: Copy HTML email viewer menu UI from Android
See https://github.com/deltachat/deltachat-desktop/pull/6203#pullrequestreview-4116180667.

nicodh:
resolves #6267

### DIFF
--- a/packages/target-electron/src/windows/html_email.ts
+++ b/packages/target-electron/src/windows/html_email.ts
@@ -244,68 +244,68 @@ export function openHtmlEmailWindow(
   )
   let context_menu_handle = createContextMenu(window, sandboxedView.webContents)
 
-  window.webContents.ipc.handle('html-view:more-menu', (_ev, { x, y }) => {
-    const menu = electron.Menu.buildFromTemplate([
-      {
-        id: 'load_remote_content',
-        label: tx('load_remote_content'),
-        click() {
-          const isAlways =
-            !isContactRequest &&
-            DesktopSettings.state.HTMLEmailAlwaysLoadRemoteContent
-          const isOnce = loadRemoteContent && !isAlways
-          const isNever = !loadRemoteContent
-          const mark = (active: boolean, label: string) =>
-            (active ? '✅ ' : '') + label
-          // needed to close the dialog if the window is closed while it's open
-          const abortController = new AbortController()
-          const onClose = () => abortController.abort()
-          window.once('close', onClose)
-          dialog
-            .showMessageBox(window, {
-              message: tx('load_remote_content_ask'),
-              buttons: [
-                mark(isNever, tx('never')),
-                mark(isOnce, tx('once')),
-                mark(isAlways, tx('always')),
-              ],
-              type: 'none',
-              icon: '',
-              defaultId: isAlways ? 2 : isOnce ? 1 : 0,
-              cancelId: 0,
-              signal: abortController.signal,
+  window.webContents.ipc.handle('html_email:get_menu_labels', () => ({
+    load_remote_content: tx('load_remote_content'),
+  }))
+
+  window.webContents.ipc.handle('html-view:load-remote-content', () => {
+    const isAlways =
+      !isContactRequest &&
+      DesktopSettings.state.HTMLEmailAlwaysLoadRemoteContent
+    const isOnce = loadRemoteContent && !isAlways
+    const isNever = !loadRemoteContent
+    const mark = (active: boolean, label: string) =>
+      (active ? '✅ ' : '') + label
+    // needed to close the dialog if the window is closed while it's open
+    const abortController = new AbortController()
+    const onClose = () => abortController.abort()
+    window.once('close', onClose)
+    const buttons = [mark(isNever, tx('never')), mark(isOnce, tx('once'))]
+    if (!isContactRequest) {
+      buttons.push(mark(isAlways, tx('always')))
+    }
+    dialog
+      .showMessageBox(window, {
+        message: tx('load_remote_content_ask'),
+        buttons: buttons,
+        type: 'none',
+        icon: '',
+        defaultId: isAlways ? 2 : isOnce ? 1 : 0,
+        cancelId: 0,
+        signal: abortController.signal,
+      })
+      .then(({ response }) => {
+        window.off('close', onClose)
+        if (response === 0) {
+          // Never: disable for this session, clear persistent setting
+          if (DesktopSettings.state.HTMLEmailAlwaysLoadRemoteContent) {
+            DesktopSettings.update({
+              HTMLEmailAlwaysLoadRemoteContent: false,
             })
-            .then(({ response }) => {
-              window.off('close', onClose)
-              if (response === 0) {
-                // Never: disable for this session, clear persistent setting
-                if (DesktopSettings.state.HTMLEmailAlwaysLoadRemoteContent) {
-                  DesktopSettings.update({
-                    HTMLEmailAlwaysLoadRemoteContent: false,
-                  })
-                }
-                update_restrictions(false)
-              } else if (response === 1) {
-                // Once: enable for this session only
-                update_restrictions(true)
-              } else if (response === 2) {
-                // Always: enable + persist (only for non-contact requests)
-                if (!isContactRequest) {
-                  DesktopSettings.update({
-                    HTMLEmailAlwaysLoadRemoteContent: true,
-                  })
-                }
-                update_restrictions(true)
-              }
+          }
+          update_restrictions(false)
+        } else if (response === 1) {
+          // Once: enable for this session only, clear persistent setting
+          if (DesktopSettings.state.HTMLEmailAlwaysLoadRemoteContent) {
+            DesktopSettings.update({
+              HTMLEmailAlwaysLoadRemoteContent: false,
             })
-            .catch(() => {
-              // dialog was closed because the window was closed (AbortError)
-              window.off('close', onClose)
+          }
+          update_restrictions(true)
+        } else if (response === 2) {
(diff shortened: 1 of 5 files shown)


r10s:
- for contact requests, the must be no "Always" option, default is always "Never" for contact request

- the menu is misplaced (see video). maybe we can also have the menu style used in the main window?

- when selecting "Always", and directly afterwards "Once", "Always" stays selected (see video)

otherwise a great improvement, making handling of these options much easier! thanks a lot!

https://github.com/user-attachments/assets/d2ae8ae8-43d9-4a18-8bed-1c383cf01030

nicodh:
> * for contact requests, the must be no "Always" option, default is always "Never" for contact request

fixed

> * the menu is misplaced (see video). maybe we can also have the menu style used in the main window?

I changed from electrons built in menu to a custom now that looks more similar. It's a bit tricky since the window is very separated from the window logic we have for the main view, and the wrapper on top of the email window has very little height. But should be ok now hopefully

> * when selecting "Always", and directly afterwards "Once", "Always" stays selected (see video)

fixed

r10s (approved):
reviewed by testing, lgtm now, thanks a lot!

packages/target-electron/static/electron_html_email_view/electron_html_email_view.html
```diff
@@ -26,7 +26,14 @@
...
       <div class="network-toggle">
-        <button id="toggle_network_more_button">⋮</button>
+        <button
+          id="toggle_network_more_button"
+          aria-haspopup="menu"
+          aria-expanded="false"
+        >⋮</button>
+        <div id="more-menu-dropdown" class="more-menu-dropdown" hidden role="menu">
+          <button id="load-remote-content-btn" class="more-menu-item" role="menuitem"></button>
+        </div>
```
WofWca, line 36: Was it necessary to transition to a web implementation of the menu instead of keeping using the native one?
Because this one doesn't implement `role="menu"` entirely well. E.g. no label, no auto-closing.
https://www.w3.org/WAI/ARIA/apg/patterns/menubar/

packages/target-electron/src/windows/html_email.ts
```diff
@@ -244,62 +243,68 @@ export function openHtmlEmailWindow(
...
     }
-    menu.popup({ window, x, y })
+    dialog
+      .showMessageBox(window, {
+        message: tx('load_remote_content_ask'),
+        buttons: buttons,
+        type: 'none',
+        icon: '',
+        defaultId: isAlways ? 2 : isOnce ? 1 : 0,
+        cancelId: 0,
```
WofWca: It doesn't make sense to change settings if the dialog was simply dismissed with "escape".

packages/target-electron/src/windows/html_email.ts
```diff
@@ -244,62 +243,68 @@ export function openHtmlEmailWindow(
...
+
+  window.webContents.ipc.handle('html-view:load-remote-content', () => {
+    const isAlways =
+      !isContactRequest &&
+      DesktopSettings.state.HTMLEmailAlwaysLoadRemoteContent
+    const isOnce = loadRemoteContent && !isAlways
+    const isNever = !loadRemoteContent
+    const mark = (active: boolean, label: string) =>
+      (active ? '✅ ' : '') + label
+    // needed to close the dialog if the window is closed while it's open
```
WofWca:
```suggestion
    // needed to close the dialog if the window is closed
    // while the dialog is open
```

But fine either way.

packages/target-electron/src/windows/html_email.ts
WofWca: Perhaps this can work for now.

```suggestion
        cancelId: isAlways ? 2 : isOnce ? 1 : 0,
```

packages/target-electron/src/windows/html_email.ts
```diff
@@ -244,62 +243,68 @@ export function openHtmlEmailWindow(
...
+              HTMLEmailAlwaysLoadRemoteContent: true,
+            })
+          }
+          update_restrictions(true)
+        }
+      })
+      .catch(() => {
+        // dialog was closed because the window was closed (AbortError)
+        window.off('close', onClose)
+      })
```
WofWca, line 308: `finally` would be simpler, but IDK if we need an explicit `off` at all.
But fine either way.

packages/target-electron/src/windows/html_email.ts
WofWca: Ah no, this won't work, because the stored setting could be "always" but the "Always" button would not be displayed if this is a contact request.

packages/target-electron/src/windows/html_email.ts
```diff
@@ -244,62 +243,68 @@ export function openHtmlEmailWindow(
...
+  window.webContents.ipc.handle('html_email:get_menu_labels', () => ({
+    load_remote_content: tx('load_remote_content'),
+  }))
+
+  window.webContents.ipc.handle('html-view:load-remote-content', () => {
+    const isAlways =
+      !isContactRequest &&
+      DesktopSettings.state.HTMLEmailAlwaysLoadRemoteContent
+    const isOnce = loadRemoteContent && !isAlways
+    const isNever = !loadRemoteContent
```
WofWca: It would be simpler to have just one value to represent which radio button is the selected one it seems.
But fine either way.

WofWca (commented):
Ah so this was a radio button all along.

Android implementation, for reference: https://github.com/deltachat/deltachat-android/blob/a3a8b3581f82456bb7fe3342485cef4593c31315/src/main/java/org/thoughtcrime/securesms/FullMsgActivity.java#L119-L189
It also displays "no" label instead of "Never" on the "never" button. And overall the logic seems to be different. E.g. there is also the `blockLoadingRemote` (a.k.a. "is contact request") check inside the button click handler. IDK if that needs to be addressed in order to consider the issue closed.

packages/target-electron/static/electron_html_email_view/electron_html_email_view.html
nicodh: If you look at my first commit I started with the native one, but switched to the styled menu after feedback from r10s. Or which "native" menu you meant?

WofWca: Ah I didn't read that, sorry.
OK, let's roll with it.

packages/target-electron/src/windows/html_email.ts
```diff
@@ -244,62 +243,67 @@ export function openHtmlEmailWindow(
...
+
+  window.webContents.ipc.handle('html-view:load-remote-content', () => {
+    const currentState =
+      !isContactRequest &&
+      DesktopSettings.state.HTMLEmailAlwaysLoadRemoteContent
+        ? 'always'
+        : loadRemoteContent
+          ? 'once'
+          : 'never'
+    const possibleStates = isContactRequest
```
WofWca:
```suggestion
    const possibleStates: Array<'never' | 'once' | 'always'> = isContactRequest
```

packages/target-electron/src/windows/html_email.ts
```diff
@@ -244,62 +243,67 @@ export function openHtmlEmailWindow(
...
-      }),
-      dont_ask: () => ({
-        id: 'show_warning',
-        type: 'checkbox',
-        label: tx('show_warning'),
-        checked: DesktopSettings.state.HTMLEmailAskForRemoteLoadingConfirmation,
-        click() {
-          DesktopSettings.update({
-            HTMLEmailAskForRemoteLoadingConfirmation:
-              !DesktopSettings.state.HTMLEmailAskForRemoteLoadingConfirmation,
```
WofWca, line 285: This setting is now unused on the Electron version. Might be worth adding a comment about that.
(Still used on Tauri though, as well as the strings).

packages/target-electron/src/windows/html_email.ts
```diff
@@ -244,62 +243,67 @@ export function openHtmlEmailWindow(
...
+          update_restrictions(false)
+        } else if (selected === 'once') {
+          if (DesktopSettings.state.HTMLEmailAlwaysLoadRemoteContent) {
+            DesktopSettings.update({
+              HTMLEmailAlwaysLoadRemoteContent: false,
+            })
+          }
+          update_restrictions(true)
+        } else {
+          // 'always' — only selectable for non-contact requests
```
WofWca, line 300:
```suggestion
```

Better move to the actual logic or remove.
But fine either way.

WofWca, line 300: Prevent future bugs.

```suggestion
          const _assert: 'always' = selected
          // 'always' — only selectable for non-contact requests
```

packages/target-electron/src/windows/html_email.ts
```diff
@@ -244,62 +243,67 @@ export function openHtmlEmailWindow(
...
+    window.once('close', onClose)
+    const buttons = possibleStates.map(
+      state => (state === currentState ? '✅ ' : '') + tx(state)
+    )
+    dialog
+      .showMessageBox(window, {
+        message: tx('load_remote_content_ask'),
+        buttons,
+        type: 'none',
+        icon: '',
```
WofWca, line 275: What's the purpose of removing the icon, and why not use 'question'?

```suggestion
        type: 'question',
```

But fine either way.

WofWca (approved):
I have checked that as of now `currentIndex` cannot be `-1`.
Also tested this for a little (but didn't test with contact requests). Seems to behave closer to how it works on Android, same for the code,
Edit: but still not entirely so (see my previous comment), so you decide if it's good enough.

PR #6392, opened 2026-05-17
fix: keep focus on old element on dialog close

WofWca:
- **refactor: factor out `focusMessageInput` fn**
- **fix: keep focus on old element on dialog close**

...instead of always focusing the composer.
Example,
1. Focus the "Settings" button (big gear icon).
2. Press Space to open the Settings dialog.
3. Press Escape.

One would expect that the settings button would get focused again
but instead we focus the composer.
https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/
also says:
> When a dialog closes, focus returns
> to the element that invoked the dialog unless...

The main change here is the `useEffect` in `MessageListAndComposer.tsx`.
The problem was that the effect focuses the composer,
but the effect's dependency, `onMouseUp`,
changes every time `hasOpenDialogs` changes.

The rest of the commit are additions of `composer.focus()`
after adding attachments in various ways,
e.g. via the WebXDC app picker or "Attach -> File".

Related:
- https://github.com/deltachat/deltachat-desktop/issues/4590.
- https://github.com/deltachat/deltachat-desktop/issues/4128.

This bug has apparently been introduced in the monster commit
584b59c1149223d04d8497f862955015248c77b7
(https://github.com/deltachat/deltachat-desktop/pull/3512).
Before that the effect had 0 dependencies, as written in
26be632de53d0b8c8f49ac1c7d51eff4aa2ccc24
(https://github.com/deltachat/deltachat-desktop/pull/2052).

Note that this MR doesn't fix _all_ the cases of focus ending up in weird places.

### DIFF
--- a/packages/frontend/src/hooks/chat/useDraft.ts
+++ b/packages/frontend/src/hooks/chat/useDraft.ts
@@ -310,12 +310,6 @@ export function useDraft(
         return
       }
 
-      // This is most likely not needed anymore
-      // as we try to focus the composer in more concrete places
-      // where `addFileToDraft()` is used.
-      // Related: https://github.com/deltachat/deltachat-desktop/issues/4590.
-      inputRef.current?.focus()
-
       const newDraftState: typeof draftState = {
         ...draftState,
         file,
@@ -330,7 +324,7 @@ export function useDraft(
       debouncedSaveAndRefetchDraft?.clear()
       return saveAndRefetchDraft(newDraftState)
     },
-    [draftState, inputRef, saveAndRefetchDraft, debouncedSaveAndRefetchDraft]
+    [draftState, saveAndRefetchDraft, debouncedSaveAndRefetchDraft]
   )
 
   const quoteMessage = useMemo(
--- a/packages/frontend/src/hooks/chat/useDraft.ts
+++ b/packages/frontend/src/hooks/chat/useDraft.ts
@@ -310,7 +310,12 @@ export function useDraft(
         return
       }
 
+      // This is most likely not needed anymore
+      // as we try to focus the composer in more concrete places
+      // where `addFileToDraft()` is used.
+      // Related: https://github.com/deltachat/deltachat-desktop/issues/4590.
       inputRef.current?.focus()
+
       const newDraftState: typeof draftState = {
         ...draftState,
         file,
--- a/packages/e2e-tests/tests/basic-tests.spec.ts
+++ b/packages/e2e-tests/tests/basic-tests.spec.ts
@@ -398,6 +398,7 @@ test('add app from picker to chat', async () => {
   await page.getByTestId('add-app-to-chat').click()
   const appDraft = page.locator('.attachment-quote-section .text-part')
   await expect(appDraft).toContainText(appName)
+  await expect(page.locator('.create-or-edit-message-input')).toBeFocused()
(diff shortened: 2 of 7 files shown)


packages/frontend/src/hooks/chat/useDraft.ts
```diff
@@ -309,7 +309,13 @@ export function useDraft(
       if (debouncedSaveAndRefetchDraft == null || saveAndRefetchDraft == null) {
         return
       }
+
+      // This is most likely not needed anymore
+      // as we try to focus the composer in more concrete places
```
nicodh: I just tested if this line can be removed and it turned out it is needed when drag dropping a file while the focus being somewhere else. Just an edge case but I would suggest to remove the misleading comment.

nicodh (changes requested):
Reviewed by testing & reading

Lgtm but see my comment...

packages/frontend/src/hooks/chat/useDraft.ts
WofWca: Are you against the idea of removing this line in general? Or is this only about the comments being misleading?

nicodh: No I just stumbled over "likely not needed anymore" and wanted to check if we can remove it or not. It turned out not (without additional changes) - so either remove the comment that says "likely not needed" or make the line really obsolete and remove it :-)

WofWca: I would say that in case of drag-and-drop it's OK not to focus the composer. The main point of all these `composer.focus()` is to _return_ focus to the composer, e.g. after going to "Add Attachment -> File". But drag-and-drop does not involve having to move focus out of the composer anyway.

nicodh: But then the line is indeed not needed anymore. If there is a reason to keep it, I would name it in the comment. In the current state the comment does not really add any value imo

WofWca: > If there is a reason to keep it, I would name it in the comment. In the current state the comment does not really add any value imo

Maybe there is a good reason, maybe there is not. I don't know. I have checked a little, but maybe I missed something. My idea is to keep the removal of that line for a follow up or for when it causes problems or otherwise gets in the way.

> In the current state the comment does not really add any value imo

It makes it clear that the line probably doesn't do much and can be removed if needed.

WofWca: If my last comment doesn't convince you, I'll remove it.

nicodh: My point is: you just dived into the focus behaviour and came to the conclusion it is not needed. If we don't remove it at some later point someone else would have to check again if it is really needed so why not do it right now.

 > I have checked a little, but maybe I missed something.

Well and if it is not much harm and can be readded later. Just to avoid this ambiguous state... ('most likely not needed')

WofWca: Removed it, especially now that we have https://github.com/deltachat/deltachat-desktop/pull/6451.

WofWca:
Rebased

nicodh (approved):
Reviewed by reading & testing

PR #6396, opened 2026-05-18
fix: reload settings after importing backup

WofWca:
This basically reverts commits
75aa4864555ad17ad92a7a64b0d5f3881590fb39
(https://github.com/deltachat/deltachat-desktop/pull/6111)
which apparently introduced this bug.

Additionally this reverts
d8f588236bd98dec5013f2dd468e5d2075761ac6
(https://github.com/deltachat/deltachat-desktop/pull/6184)
which only partially fixed the bug,
i.e. only for the "instant onboarding" case.

### DIFF
--- a/packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
+++ b/packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
@@ -14,7 +14,9 @@ import ChatList from '../../chat/ChatList'
 import { Avatar } from '../../Avatar'
 import ConnectivityToast from '../../ConnectivityToast'
 import MailingListProfile from '../../dialogs/MailingListProfile'
-import { useSettingsStore } from '../../../stores/settings'
+import SettingsStoreInstance, {
+  useSettingsStore,
+} from '../../../stores/settings'
 import { BackendRemote, onDCEvent, Type } from '../../../backend-com'
 import Button from '../../Button'
 import Icon from '../../Icon'
@@ -233,6 +235,22 @@ export default function MainScreen({ accountId }: Props) {
       lastUsedAppsFetch?.refresh()
     })
   }, [accountId, lastUsedAppsFetch])
+
+  // There is another `load()` in `ScreenController.selectAccount()`,
+  // but that is not enough because we also need to reload settings
+  // after an unconfigured account becomes a configured one,
+  // i.e. after login or backup import,
+  // which correlates with `MainScreen` becoming rendered.
+  useEffect(() => {
+    if (
+      SettingsStoreInstance.state?.accountId === accountId &&
+      SettingsStoreInstance.state?.settings.configured_addr
+    ) {
+      log.debug('account is already configured, skipping settings reload')
+      return
+    }
+    SettingsStoreInstance.effect.load()
+  }, [accountId])
 
   const isSearchActive = queryStr.length > 0 || queryChatId !== null
   const showArchivedChats = !isSearchActive && archivedChatsSelected
--- a/packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
+++ b/packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
@@ -241,6 +241,10 @@ export default function MainScreen({ accountId }: Props) {
   // after an unconfigured account becomes a configured one,
   // i.e. after login or backup import,
   // which correlates with `MainScreen` becoming rendered.
+  //
+  // TODO perf: this causes settings to load twice
+  // if they are still loading as a result of
+  // `ScreenController.selectAccount()`.
   useEffect(() => {
     if (
       SettingsStoreInstance.state?.accountId === accountId &&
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
```diff
@@ -234,6 +236,22 @@ export default function MainScreen({ accountId }: Props) {
...
+  useEffect(() => {
+    if (
+      SettingsStoreInstance.state?.accountId === accountId &&
+      SettingsStoreInstance.state?.settings.configured_addr
+    ) {
+      log.debug('account is already configured, skipping settings reload')
+      return
+    }
+    SettingsStoreInstance.effect.load()
+  }, [accountId])
```
nicodh, line 257: This results in a redundant duplicate load() of the settings on each regular account switch. I think that can be avoided by always calling SettingsStoreInstance.effect.load() in Screencontroller.selectAccount()  (not only if accountId !== this.selectedAccountId) since then unconfigured/configured accounts will always reload the settings.

That would also avoid that Main Screen as a view component needs to care about reload settings which seems a cleaner approach

nicodh (changes requested):

packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
WofWca: > This results in a redundant duplicate load()

Yes, that's unfortunate.

> I think that can be avoided by

I don't think so. For example we don't call `selectAccount` a second time after "instant onboarding" is done. This is only the case for `AccountSetupScreen`. Nor does it make sense, based on the name on the function, to call it a second time with the same `accountId`.

https://github.com/deltachat/deltachat-desktop/blob/aad179f7e33eb59e73da3bb12433919ce03e8295/packages/frontend/src/ScreenController.tsx#L144-L169

https://github.com/deltachat/deltachat-desktop/blob/1cf34d36c53898526cb0c7bd4152d8ab544a9d55/packages/frontend/src/components/screens/AccountSetupScreen.tsx#L22-L42

nicodh: > For example we don't call selectAccount a second time after "instant onboarding" is done. 

You mean theline that was removed in this PR? So we could/should restore that call

> Nor does it make sense, based on the name on the function, to call it a second time with the same accountId.

I would say that is the result of the account ID architecture we have: assigning an ID before configuring an account. So adding a comment on the second call why it's needed should clarify that. And finally the account setup/instant onboarding is the rare use case compared to the regular account switch.

WofWca: > You mean theline that was removed in this PR? 

No I mean you're saying that we can simply call `settings.load()` inside of `selectAccount` and nowhere else. I am saying that we can't because we need to reload the settings after the account has been configured but `selectAccount` doesn't always get called when an account transitions from unconfigured to configured.

> So adding a comment on the second call why it's needed should clarify that.

I think it's simpler to directly reload the settings instead of calling `selectAccount` for the sole purpose of its side-effect of reloading settings.

nicodh: > I think it's simpler to directly reload the settings instead of calling selectAccount for the sole purpose of its side-effect of reloading settings

Then go with that. I just want to point out that we should avoid to introduce again the redundant loading of the settings twice on each account switch.

WofWca: I would prefer having things more declarative, i.e. have just one or two places where we call `load()`, instead of adding imperative `load()`s to 5 or so places and hoping that we don't forget it next time. Or do you think that the latter is better?

For the former, one can consider refactoring things into a single `useEffect` that depends on the `screen` and `accountId`.

packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
```diff
@@ -234,6 +236,22 @@ export default function MainScreen({ accountId }: Props) {
     })
   }, [accountId, lastUsedAppsFetch])
 
+  // There is another `load()` in `ScreenController.selectAccount()`,
+  // but that is not enough because we also need to reload settings
+  // after an unconfigured account becomes a configured one,
+  // i.e. after login or backup import,
+  // which correlates with `MainScreen` becoming rendered.
```
WofWca, line 243:
```suggestion
  // which correlates with `MainScreen` becoming rendered.
  //
  // TODO perf: this causes settings to load twice
  // if they are still loading as a result of
  // `ScreenController.selectAccount()`.
```

nicodh (approved):
Let's merge this to get the fix in the next release and refactor soon to get a cleaner fix for the next release without redundant load calls

WofWca:
What about the TODO comment? Shall I add it?

WofWca:
Rebased

nicodh:
> What about the TODO comment? Shall I add it?

Yes, makes sense.

PR #6398, opened 2026-05-18
test: make some tests offline: import acc backup

WofWca:
Currently in order to run tests we're creating accounts
on a special Chatmail server dedicated for running tests.
That, however, is not required for some tests
as they don't require network at all.

So let's make the profile from a backup.

This also improves test performance,
and, obviously, tests the "import backup" workflow.

This additionally improves performance by reusing the main page
in the `beforeAll` hooks.

If one is concerned with the contents of the .tar file, you may use some SQLite explorer software to inspect the database. It's not super big. Almost all of the 45 tables are empty.

TODO:
- [x] Merge the base into main, otherwise the tests fail because of the bug that the base MR fixes.


nicodh (approved):
Reviewed by reading & testing

I doubt if it makes sense to remove the images and the device messages chat. It's just some kb and might be useful to also have a check that importing images from backup works. Also so count of initial chats is the "expected" number.

WofWca:
I have thought about having a more sizeable account for tests, i.e. with some hundreds of chats and thousands of messages, to have a more "normal"-looking account to do tests on, and to reduce the setup that we have to do for each test, such as creating dummy chats for the "chat multiselect tests".
But that can be done later, as a separate option. And maybe to be stored in a different branch or repo. Our repo is already embarrassingly heavy.

WofWca:
Rebased.

PR #6402, opened 2026-05-19
fix: Shift + Tab sometimes selecting chats

WofWca:
Reproduction:
1. Activate (open) a chat.
2. Move focus to that chat list item.
3. Press ArrowDown a couple of times (so that the focused chat
   is not the active chat).
4. Press Tab (to move focus ouf of the chat list).
5. Press Shift + Tab (to move the focus back in the chat list).

This will unexpectedly perform a contiguous selection
between the active chat and the focused chat.
This commit fixes that.

This is also gonna be useful for the message list multiselect,
where this bug is more annoying.


nicodh (approved):
Reviewed by reading & testing

puh - that's really an edge case. But since not too many lines are added, ok to fix it :-)

WofWca:
> puh - that's really an edge case.

Not really for the message list. There it would be happening basically all the time.

PR #6404, opened 2026-05-20
fix: Settings: keep focus after going back a level

WofWca:
Reproduction:
1. Open Settings.
2. Click on "Chats".
3. Press Escape.

The focus will get reset, now you have to tab through items above
to get to "Chats" again.

Note that this commit changes how things look:
now we open the sub-levels in a new dialog,
which darkens the background further, and the main dialog
could be visible behind the new one.

This also basically reverts
f979e9fa62be4f4085d24111996039faa3f0b5a5
(https://github.com/deltachat/deltachat-desktop/pull/4702),
which was made to (partially) address
https://github.com/deltachat/deltachat-desktop/issues/4128.

Related:
- https://github.com/deltachat/deltachat-desktop/pull/6392.

### DIFF
--- a/packages/frontend/src/components/Settings/Settings.tsx
+++ b/packages/frontend/src/components/Settings/Settings.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react'
+import React, { useEffect, useCallback } from 'react'
 
 import { useSettingsStore } from '../../stores/settings'
 import { SendBackupDialog } from '../dialogs/SetupMultiDevice'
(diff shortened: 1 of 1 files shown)


nicodh:
Hmm, I'm not sure if we want to pile dialogs up like that 🤔

image


Is the only reason, that the focus gets lost in the parent dialog or are there other reasons? @r10s do you have an opinion UI wise?

WofWca:
> Is the only reason, that the focus gets lost in the parent dialog or are there other reasons?

Yes, this is the only reason that I'm aware of.

r10s:
we definetely want to pile the dialogs, the added opacity is a bit doom style, but well

nicodh (approved):
Reviewed by reading & testing

Can be merged after rebase imo

WofWca:
So @r10s is fine with this?

I had ideas of adding opacity (not `display: none`!) to the underlying dialog when a "child" dialog gets opened, but that's some extra complexity. I would say this can be done later, in follow-ups, if needed.

WofWca:
Got green light from @nicodh  to merge this.

PR #6405, opened 2026-05-21
fix: Escape in Create Group goes back to New Chat

WofWca:
Partially addresses
https://github.com/deltachat/deltachat-desktop/issues/4128.

Note that this changes how things look a bit:
now the "Create Group" / "Create Channel" dialog is displayed
on top of the "New Chat" dialog, instead of basically replacing it.

When group / channel creation finishes, we still close
both of the dialogs.

Hide whitespace for easier review

This is mostly AI-generated, with adjustments from me. Let me know if you'd like me to rewrite this by hand.
The prompts were:
> In `CreateChat` move the sub-sections into their own dialogs.
> Replace `setViewMode` with `onDone` which we'll call
> when chat creation is done, which will close the parent dialog.

> There is no need to have thin wrappers like
> `CreateBroadcastListDialog`.
> Remove them and do the changes on `CreateBroadcastList` directly.
> Let's also keep the name `CreateBroadcastList` for now,
> don't append `Dialog`.

Related:
- https://github.com/deltachat/deltachat-desktop/pull/6404

### DIFF
--- a/packages/e2e-tests/playwright-helper.ts
+++ b/packages/e2e-tests/playwright-helper.ts
@@ -545,7 +545,7 @@ export const createGroupChat = async (
   await page.getByPlaceholder('Description').fill('Test group description')
   await page.locator('#addmember button').click()
   const addMemberDialog = page.getByTestId('add-member-dialog')
-  await page
+  await addMemberDialog
     .locator('.contact-list-item')
     .filter({ hasText: member.name })
     .click()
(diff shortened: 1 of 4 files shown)


nicodh (approved):
Reviewed by reading & testing

Note: this PR fixes the issue for Create group/Create channel as the title says. The issue, that esc/cancel after "New contact" also closes the NewChat dialog is still left

WofWca:
Good point. I added this to my TODO comment on that issue.

PR #6406, opened 2026-05-21
fix: use new "empty sticker folder" string, improve style

WofWca:
Seems like this was an unrelated change introduced by mistake in
2e2c7e6a308f6bf7969e29e6633e120d3fee34b7
(https://github.com/deltachat/deltachat-desktop/pull/6278),
was probably supposed to be a follow-up to
https://github.com/deltachat/deltachat-desktop/pull/6231#discussion_r3146871822.

The old string is deprecated:
https://github.com/deltachat/deltachat-desktop/blob/7a654e06ee533261b53315ceee2ae6b57f4dbeb8/_locales/en.xml#L184-L190

Also improve the font size and center the text, and remove an unused class name.

Before:
image
After:
image


nicodh (approved):
Reviewed by reading & testing

PR #6409, opened 2026-05-22
chore: upgrade core to 2.50.0

link2xt:


WofWca:
This closes https://github.com/deltachat/deltachat-desktop/issues/6333
and supersedes https://github.com/deltachat/deltachat-desktop/pull/6313 ? Or not?

WofWca:
And a better question: Would Delta Chat be in a releasible state if this gets merged, or are further adjustments needed?

nicodh (approved):
Reviewed by testing & reading.

Thanks for contributing to desktop :-)

link2xt:
> And a better question: Would Delta Chat be in a releasible state if this gets merged, or are further adjustments needed?

Still need to add "force encryption" setting. https://github.com/deltachat/deltachat-desktop/pull/6313/changes was adding "force encryption" into advanced relay settings.

This setting affects all relays, but @r10s said in some chat "let's leave it there and call it "force encryption for all relays" or so". So the idea is to still put this setting somewhere into advanced relay settings like #6313 did.

nicodh:
Well we have both options now. Putting it in relay settings (Like Android) or in Advanced settings.
Will be merged after it's decided what's considered best.

PR #6410, opened 2026-05-23
feat: add option to force encryption

nicodh:


nicodh:
See the related PR in Android https://github.com/deltachat/deltachat-android/pull/4431

packages/frontend/src/stores/settings.ts
```diff
@@ -132,6 +134,10 @@ class SettingsStore extends Store<SettingsStoreState | null> {
         settings['ui.mentions_enabled'] = mentionsEnabledDefaultVal
       }
 
+      if (settings['force_encryption'] == null) {
+        settings['force_encryption'] = '1'
+      }
+
```
WofWca: What's the benefit of this? Core already provides a default value

https://github.com/chatmail/core/blob/480248a1682f4a32a164c837d6d6aef5569edf8d/src/config.rs#L474-L479

I tested this and Core returns `'1'` even for fresh unconfigured accounts. Let's not add extra logic that can potentially get out of sync.

```suggestion
```

packages/frontend/src/components/LoginForm.tsx
```diff
@@ -42,6 +44,8 @@ export default function LoginForm({
   const [providerInfo, setProviderInfo] = useState<
     Type.ProviderInfo | undefined
   >()
+  const settingsStore = useSettingsStore()[0]
+  const forceEncryption = settingsStore?.settings['force_encryption'] !== '0'
```
WofWca: I think it's better to falsely say that encryption is disabled than it is to falsely say that it's enabled.

```suggestion
  const forceEncryption = settingsStore?.settings['force_encryption'] === '1'
```

packages/frontend/src/components/LoginForm.tsx
```diff
@@ -260,7 +264,22 @@ export default function LoginForm({
               <option value={Socket.starttls}>STARTTLS</option>
               <option value={Socket.plain}>{tx('off')}</option>
             </DeltaSelect>
-
+            <div className='delta-form-group delta-switch'>
```
WofWca: Quite misleading placement of the setting, but yep I see that Android's doing the same.
Do we want toggling the switch to take effect immediately or only when the form gets submitted?

WofWca (commented):
Works.

packages/frontend/src/components/LoginForm.tsx
nicodh: Hmm, good question. The "cancel" button somehow implies "discard my changes" so maybe an unexpected behaviour if we change it immediately. On the other hand this settings is anyway an "expert" feature that should be only changed by people who know what they are doing. And of course "Login"  is also a bit strange for saving a value that is not related to the login. But that is all a result of the mentioned "misplacement" of this setting.

r10s: - if it is not an unreasonable effort, we should apply the changes only when "OK" is hit, we also changed that on android at https://github.com/deltachat/deltachat-android/pull/4431/changes/40faa3d43e64207a12acaf26e3bdf5891e6fa33c

- in any case, however, it should be the really _last_ option in the "more settings", so after "certificate checks"

- for clarity: no need to add it to "untranslated", there is the existing string `enforce_e2ee` saying "Enforce Encryption for All Relays", which is clearer. position was discussed a lot, it is fine, if ppl do not find the option without explicitly knowing it exist by other means

nicodh: Ok, I changed the behaviour. The setting is not applied when the user cancels the form.

The translation still waits for #6416 to be approved. Then the untranslated entry will be removed.

packages/frontend/scss/login/_login.scss
```diff
@@ -152,6 +152,25 @@ div.delta-form-group {
     }
   }
 
+  &.delta-switch {
+    overflow: visible;
+
```
WofWca:
```suggestion
```

This doesn't do anything, right? It's the default value anyway.

packages/frontend/src/components/dialogs/EditAccountAndPasswordDialog.tsx
```diff
@@ -57,11 +57,15 @@ function EditAccountInner({
...
   addr?: string
 }) {
+  const settingsStore = useSettingsStore()[0]
   const [initialSettings, setInitialAccountSettings] =
     useState<Credentials>(defaultCredentials())
 
   const [accountSettings, setAccountSettings] =
     useState<Credentials>(defaultCredentials())
+  const [forceEncryption, setForceEncryption] = useState<boolean>(
+    settingsStore?.settings['force_encryption'] !== '0'
```
WofWca:
```suggestion
    settingsStore?.settings['force_encryption'] === '1'
```

We already changed this in one place, so let's be consistent? Especially here where the `settingsStore` can be nullish. See https://github.com/deltachat/deltachat-desktop/pull/6410#discussion_r3305095058

packages/frontend/src/components/screens/AccountSetupScreen.tsx
```diff
@@ -32,25 +33,43 @@ export default function AccountSetupScreen({
...
-            openDialog(AlertDialog, {
-              message: error,
-              cb: () => setPromptDialogId(null),
-            })
-          ),
-      }),
-    [accountId, openDialog, selectAccount, credentials]
+  const settingsStore = useSettingsStore()[0]
+  const [forceEncryption, setForceEncryption] = useState<boolean>(
+    settingsStore?.settings['force_encryption'] !== '0'
```
WofWca:
```suggestion
    settingsStore?.settings['force_encryption'] === '1'
```

See https://github.com/deltachat/deltachat-desktop/pull/6410#discussion_r3305095058

packages/frontend/src/components/LoginForm.tsx
```diff
@@ -275,6 +283,26 @@ export default function LoginForm({
...
+                  disabled={settingsStore == null}
+                  onChange={() => {
+                    if (setForceEncryption) {
+                      setForceEncryption(!forceEncryption)
+                      return
+                    }
+                    SettingsStoreInstance.effect.setCoreSetting(
+                      'force_encryption',
+                      forceEncryption ? '0' : '1'
+                    )
```
WofWca: This is basically unused now, right?

WofWca (approved):
Tested again, both of the places where `LoginForm` is used. Works. I didn't test the "error" path though.
The code looks alright. The setting only gets saved on success.

PR #6411, opened 2026-05-23
chore: avoid eslint warning

nicodh:


packages/frontend/src/components/Settings/Autodelete.tsx
```diff
@@ -70,10 +70,9 @@ export default function Autodelete({
       title: tx('autodel_device_title'),
       onSave: async (_seconds: string) => {
         const seconds = Number(_seconds)
-        const fromServer = false
         const estimateCount = await BackendRemote.rpc.estimateAutoDeletionCount(
           accountId,
-          fromServer,
+          false,
```
WofWca: There is no warning here though? And I'm pretty sure it was intentional. Let's keep a comment at least then


```suggestion
          false, // fromServer
```

WofWca (approved):
Oops, I missed that one, sorry.

PR #6412, opened 2026-05-25
refactor: remove unnecessary check

WofWca:
The wrapped component used to take `settingsStore` as a prop,
and that was the reason for the check.
Now it doesn't, so the check is not needed.


nicodh (approved):
Reviewed by reading

PR #6413, opened 2026-05-25
fix: disable keyboard shortcuts when have dialog

WofWca:
For example, don't open multiple "New Chat" dialogs,
don't switch chats while a "View Chat" dialog is open.

However this commit is primarily for
https://github.com/deltachat/deltachat-desktop/pull/6223
where this is much more important.

### DIFF
--- a/packages/frontend/src/keybindings.ts
+++ b/packages/frontend/src/keybindings.ts
@@ -127,12 +127,30 @@ export function matchesNonLetterShortcut(
   return !isPrintableAscii && ev.code === code
 }
 
+function haveOpenDialogs(): boolean {
+  return document.querySelector('dialog:modal') != null
+}
+
 export function keyDownEvent2Action(
   ev: KeyboardEvent
 ): KeybindAction | undefined {
   const { isMac } = runtime.getRuntimeInfo()
 
-  if (window.__contextMenuActive) {
+  // If a dialog is open, disable shortcuts.
+  // This also applies to a context menu being open.
+  //
+  // Perhaps there are some shortcuts that we don't strictly need to disable
+  // when a dialog is open, such as `Settings_Open` and `Debug_MaybeNetwork`,
+  // but it's probably not important to support that,
+  // so let's not complicate things yet.
+  if (
+    (ev.target instanceof HTMLElement && ev.target.closest('dialog') != null) ||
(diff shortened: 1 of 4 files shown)


packages/frontend/src/keybindings.ts
```diff
@@ -127,12 +127,30 @@ export function matchesNonLetterShortcut(
...
   ev: KeyboardEvent
 ): KeybindAction | undefined {
   const { isMac } = runtime.getRuntimeInfo()
 
-  if (window.__contextMenuActive) {
+  // If a dialog is open, disable shortcuts.
+  // This also applies to a context menu being open.
+  //
+  // Perhaps there are some shortcuts that we don't strictly need to disable
+  // when a dialog is open, such as `Settings_Open` and `Debug_MaybeNetwork`,
```
nicodh, line 143: In my tests it turned out that Ctrl + , still opens the settings while a dialog is open. Maybe since it's also referenced in the top menu.

nicodh (approved):
Reviewed by reading & testing

While it might be irritating that an open context menu blocks a createChat dialog etc. from being opened, we definitely should block to open multiple same dialogs at the same time

WofWca:
Rebased.

packages/frontend/src/keybindings.ts
WofWca: Hm yep. We don't have a check here

https://github.com/deltachat/deltachat-desktop/blob/454925d87bbd52aab9c4c47c63543103a7540dfd/packages/frontend/src/components/RuntimeAdapter.tsx#L77-L85

https://github.com/deltachat/deltachat-desktop/blob/f810410a57608ce7f842cda640813d2ca0eb3718/packages/frontend/src/contexts/KeybindingsContext.tsx#L25-L34

WofWca:
> it might be irritating that an open context menu blocks a createChat dialog etc. from being opened

This MR doesn't change that aspect though. It's already the case on current main.

packages/frontend/src/keybindings.ts
WofWca: Hm yep it's this code

https://github.com/deltachat/deltachat-desktop/blob/454925d87bbd52aab9c4c47c63543103a7540dfd/packages/frontend/src/components/RuntimeAdapter.tsx#L77-L85

But I'd say this is for another MR.

PR #6416, opened 2026-05-26
chore: add burmese language

nicodh:
Translation updates also affect #6410


WofWca (approved):
So this also updates other languages, besides adding Burmese.

I only looked at hr `_languages.json` file and confirmed that the language code is correct.

PR #6417, opened 2026-05-26
chore: move ws 8.x to catalog

nicodh:
The failing test in https://github.com/deltachat/deltachat-desktop/pull/6408 revealed the fact that browser and electron use different versions of ws. So let's see if we can't use ws 8.x throughout.

### DIFF
--- a/packages/target-browser/package.json
+++ b/packages/target-browser/package.json
@@ -25,13 +25,13 @@
     "express-session": "^1.18.2",
     "node-localstorage": "^3.0.5",
     "resolve-path": "^1.4.0",
-    "ws": "~8.18.0"
+    "ws": "catalog:"
   },
   "devDependencies": {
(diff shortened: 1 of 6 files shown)


WofWca (commented):
So what's the point of this? These places are not inter-dependent it seems?

nicodh:
It's an upgrade of ws for electron and as said just to have the same version all over the project

WofWca (approved):
Alright, an upgrade is good, and having the same (or different) versions between Electron and target-browser doesn't really matter as of now I guess.

Simon-Laux:
related: https://github.com/chatmail/yerpc/pull/74 removes the dependency on websocket package in yerpc, so when we decide to publish a new version of yerpc, we will have less trouble with websocket npm packages.

PR #6420, opened 2026-05-27
chore: upgrade Electron

WofWca:
Some memory safety fixes
https://releases.electronjs.org/release/compare/v41.6.1/v41.7.1.


nicodh (approved):
Makes sense. Seems not to introduce any severe changes for us

PR #6421, opened 2026-05-27
test: add multi-Delta-Chat-instance tests

WofWca:
- **refactor: test: factor out `resolvePath()`**
- **refactor: test: simplify `DC_ACCOUNTS_DIR`**
- **test: add `PRIVATE_CERTIFICATE_PATH` support**
- **refactor: test: move some consts to another file**
- **test: support `DATA_DIR` env var**
- **test: add multi-Delta-Chat-instance tests**

test: add multi-Delta-Chat-instance tests

This starts multiple Delta Chat instances,
available on different ports,
each having their own data directory (for accounts and stuff).

And add tests for "Add Second Device".

This will also allow us to test things
that we previously weren't able to test,
due to the fact that we can only have one page open
(see `error_other_client_stole_dc_connection`).
For example, we can now add tests
for receiving a message while a chat is open.
Also this might allow to simplify other multi-account tests,
because now explicitly switching between accounts is not required.

Those developers who have an existing `e2e-tests/.env` file
should remove `DC_ACCOUNTS_DIR` from it,
because the accounts dir is now specified per-instance.

The initial version of this MR has been AI-generated,
then within a couple of hours I rewrote ~75% of it.
The idea itself, i.e. using multiple ports and data dirs, is mine.
The `test()` is also 100% mine.
The prompt was something along the lines of
> Add support for running multiple app instances
> on separate ports in `e2e-tests`.
> Keep existing tests as is.

TODO:
- [x] Merge the base https://github.com/deltachat/deltachat-desktop/pull/6398 into main, such that only 6 commits remain.

### DIFF
--- a/packages/e2e-tests/playwright.config.ts
+++ b/packages/e2e-tests/playwright.config.ts
@@ -1,8 +1,14 @@
+import path from 'node:path'
 import { defineConfig, devices } from '@playwright/test'
 
 import { loadEnv } from './load-env'
 import type { TestOptions } from './playwright-helper'
-import { baseURL, DC_FRONTEND_NO_TLS } from './playwright-helper'
+import {
+  DC_FRONTEND_NO_TLS,
+  NUM_APP_INSTANCES,
+  instanceBaseURL,
+  instancePort,
+} from './playwright-helper'
 
 loadEnv()
 
@@ -33,7 +39,8 @@ export default defineConfig<TestOptions>({
   /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
   use: {
     /* Base URL to use in actions like `await page.goto('/')`. */
-    baseURL: baseURL,
+    // Instance 0 is used as the default instance for single-instance tests.
+    baseURL: instanceBaseURL(0),
 
     /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
     trace: 'on-first-retry',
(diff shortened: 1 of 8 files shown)


packages/e2e-tests/playwright.config.ts
```diff
@@ -39,7 +39,8 @@ export default defineConfig<TestOptions>({
   /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
   use: {
     /* Base URL to use in actions like `await page.goto('/')`. */
-    baseURL: baseURL,
+    // Instance 0 is used as the default instance for singlie-instance tests.
```
nicodh:
```suggestion
    // Instance 0 is used as the default instance for single-instance tests.
```

packages/e2e-tests/tests/add-second-device.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+
+test.beforeAll(async ({ browser }) => {
+  ;[{ context: contextA, page: pageA }, { context: contextB, page: pageB }] =
+    await Promise.all([
+      openInstancePage(browser, 0),
+      openInstancePage(browser, 1),
+    ])
+
+  await importDummyProfileFromBackup(pageA)
+})
```
nicodh, line 28: I think we should add a paragraph to the E2E-TESTING.md about how to test/simulate 2 different devices or point to the beforeAll function of this test.

nicodh (approved):

Very nice to have this available now for multi-device testing!

packages/e2e-tests/tests/add-second-device.spec.ts
WofWca: I fear that that this might be just duplicate info that might get outdated.
But maybe it's worth it.
I'll add it.

WofWca: Done, take a look.

PR #6424, opened 2026-05-29
feat: show imap folder if set

### LINKED ISSUE #6422: Add "IMAP folder" field to advanced relay settings
IMAP login params have a "folder" setting since 2.50, right after the "port" setting.

It is a replacement for old `only_fetch_mvbox` setting, for migrated users it may be set to `DeltaChat`.

It should be displayed, possibly as a disabled field, and resubmitted when user reconfigures the relay so the relay does not get reconfigured to use INBOX folder accidentally just because user wants to change the password.

nicodh:
Resolves #6422 

image

Field is only shown if imap_folder is not empty.

tbd: 
1. do we need the hint?
2. should we add a translation for "IMAP Folder"?

### DIFF
--- a/packages/frontend/src/components/LoginForm.tsx
+++ b/packages/frontend/src/components/LoginForm.tsx
@@ -210,15 +210,15 @@ export default function LoginForm({
             />
             {imapFolder !== null && (
               <>
-                <p>{tx('pref_imap_folder_warn_disable_defaults')}</p>
                 <DeltaInput
                   key='imapFolder'
                   id='imapFolder'
                   placeholder={tx('automatic')}
-                  label={tx('imap_folder')}
+                  label='IMAP Folder'
                   type='text'
                   value={imapFolder}
                   onChange={handleCredentialsChange}
+                  disabled
                 />
               </>
             )}
--- a/packages/frontend/src/components/LoginForm.tsx
+++ b/packages/frontend/src/components/LoginForm.tsx
@@ -115,6 +115,7 @@ export default function LoginForm({
     password,
     imapServer,
     imapPort,
+    imapFolder,
     imapSecurity,
     certificateChecks,
     smtpUser,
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/LoginForm.tsx
```diff
@@ -207,6 +208,20 @@ export default function LoginForm({
...
+                <p>{tx('pref_imap_folder_warn_disable_defaults')}</p>
+                <DeltaInput
+                  key='imapFolder'
+                  id='imapFolder'
+                  placeholder={tx('automatic')}
+                  label={tx('imap_folder')}
+                  type='text'
+                  value={imapFolder}
+                  onChange={handleCredentialsChange}
+                />
```
WofWca, line 222: Doesn't the issue say that it needs to be always disabled?

WofWca (commented):
Looks like the "resubmitted" part is addressed, because we always pass the entire "credentials" object (didn't check thoroughly yet though).

packages/frontend/src/components/LoginForm.tsx
WofWca:
```suggestion
                <DeltaInput
                  key='imapFolder'
                  id='imapFolder'
                  disabled
                  placeholder={tx('automatic')}
                  label={tx('imap_folder')}
                  type='text'
                  value={imapFolder}
                  onChange={handleCredentialsChange}
                />
```

r10s:
> do we need the hint?

no.

the string is already deprecated.

> should we add a translation for "IMAP Folder"?

no.

just hardcode the string as an android . the future of that option is unclear, and it is for very few ppl only. giving strings to translators that cannot be seen in normal UI and will otherwise raise question does not make sense

packages/frontend/src/components/LoginForm.tsx
r10s: > Doesn't the issue that that it needs to be always disabled?

yes, the folder must not be editable. we do not want to raise new issues

WofWca (approved):
I have verified that the value of the setting gets passed to Core here.

https://github.com/deltachat/deltachat-desktop/blob/820deb5fcb0b0f350fce1041122dddad60196948/packages/frontend/src/components/dialogs/ConfigureProgressDialog.tsx#L104-L107

And that the field is disabled, and that it does get displayed when the value is non-null.

PR #6425, opened 2026-05-30
chore: upgrade core to 2.51.0

nicodh:


WofWca (approved):
Just a version bump.
I heard we can't release 2.50 and 2.51 is required, so yep.
There are no breaking changes that need to be addressed in the changelog https://github.com/chatmail/core/blob/main/CHANGELOG.md#2510---2026-05-29

PR #6426, opened 2026-05-30
test: add "export then import backup" test

WofWca:


packages/e2e-tests/tests/backup-tests.spec.ts
```diff
@@ -65,6 +69,87 @@ test('import backup from file', async () => {
...
+        await deleteSelectedProfile(page)
+      })
+    )
+    // await contextA.close()
+    await contextB.close()
+  })
+
+  test('export', async () => {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const page = undefined // Shadow the original var
```
nicodh: What is the reason for this line?
page is not used anyway?

WofWca:
```suggestion
    const page = undefined // Make sure not to use it accidentally.
```

(I'll adjust the other occurrence)

nicodh: I don't agree to add a unused variable to avoid possible future mistakes. 
Imo it's enough to add a comment line in the beginning if you consider that a real risk but not add code

WofWca: Umm why not? It's better when you can see the mistake immediately rather than after something fails and after you find the comment? Code is better than a comment.

nicodh: Where to start where to end? There are a lot of "possible" future mistakes...

And it's inside a test that would probably fail if something wrong is added

WofWca: I don't know what's so bad about two lines of code that prevent misuse (why do we need TypeScript and ESLint then), but whatever, I removed it.

nicodh (approved):
Reviewed by reading

PR #6429, opened 2026-06-01
fix: wait for leaving before deleting a group

### LINKED ISSUE #6428: Leave and delete group chat sometimes not working
See https://github.com/chatmail/core/issues/8214

Reproducable in 2.49.1 and 2.50.0

Seems the group chat is sometimes deleted before the user leaves it

nicodh:
resolves #6428

### DIFF
--- a/packages/e2e-tests/tests/chat-context-menu.spec.ts
+++ b/packages/e2e-tests/tests/chat-context-menu.spec.ts
@@ -481,23 +481,36 @@ test.describe('Chat List Context Menu - Single Selection', () => {
 test.describe('Removal actions', () => {
   test('delete chat from main view closes the chat', async () => {
     const userA = existingProfiles[0]
+    const userB = existingProfiles[1]
 
     await switchToProfile(page, userA.id)
 
     // Create group to delete
+    const leaveGroupName = 'Leave Notify Test Group'
     await page.locator('#new-chat-button').click()
     await page.getByRole('button', { name: 'New Group' }).click()
+    await page.getByRole('textbox', { name: 'Group Name' }).fill(leaveGroupName)
+    await page.locator('#addmember button').click()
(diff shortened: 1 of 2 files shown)


nicodh:
@WofWca I had the awaits in my initial commit but you asked to remove them (and I gave in too easily) https://github.com/deltachat/deltachat-desktop/pull/6153#discussion_r3051373882

Do you remember why? Both functions return promises and the leave group won't be completed if the delete chat call is faster.

WofWca:
Sorry about that.
The reason was better performance. It did not occur to me to check if Core doesn't have such a race.

WofWca (approved):
I have verified the issue. Indeed, upon "leave and delete" sometimes there is an error in the console saying "Chat#123 not found" (or something along the lines).

The test looks sensible at a glance.
I don't see how the new code can break anything, so it probably needs to get merged now, right before the release.
The code is also in line with the other occurrence of `leaveChat()`.

link2xt:
Does it make sense to enable no-floating-promises lint for the frontend?

Copy-pasting this section, but replacing `e2e-tests` with `frontend` results in ~250 warnings from `pnpm check:lint`:
https://github.com/deltachat/deltachat-desktop/blob/9c2ba8a88682bddeebd9428c2d0315f8a1e44b00/eslint.config.js#L48-L62

There is still a way to `void` the promise if you explicitly want to ignore it, but should catch ignoring the promise accidentally: https://typescript-eslint.io/rules/no-floating-promises/

IIRC we already had similar problems in the past in the build scripts.

nicodh:
> Does it make sense to enable no-floating-promises lint for the frontend?

Yes maybe, although it would not have protected us in that case. We could start with adding a lot of void operators to all calls that definitely need no handling like open Dialogs, Context menus  etc.

PR #6430, opened 2026-06-02
fix: warn before sending large image

### LINKED ISSUE #5053: Message input lag / typing slowness
MacOS 1.56.0

This has been reported by various people but I found a reproducible test case today.

I attached this image file to a chat. It was not compressed further. Having this image in any of your chats or saved messages will impact the Desktop client's performance by causing the Rederer helper to eat 100% CPU in bursts

Image

nicodh:
fixes #5053 

Current behaviour:

Large-Image

Desktop shows a core error if the image is sent as Image (in the old deprecated "user feedback" way) and renders the image only if it is sent as File.

The reason is in core:

1. core runs out of memory when img reader decodes the image https://github.com/chatmail/core/blob/main/src/blob.rs#L360 which results in the fallback to viewType File
2. when the image is sent as file core converts it to an image based on the mime type and "converts" it to viewType Image but does NOT run imgreader.decode() 

In the latter case desktop displays the uncompressed image which causes delays in Chrome.

So the fix in desktop is:  show a warning when an Image is too large to be rendered in the UI.

image

And don't render images that are larger than LARGE_IMAGE_PIXEL_THRESHOLD anymore.

In Android the large image results in a grey empty "image like" message. So the same behaviour from core results in a viewType Image that the webview can't display.

android-large-image

I also checked how Signal treats large images: it turned out, the images are resized already on the client side (see https://github.com/signalapp/Signal-Desktop/blob/main/ts/util/Attachment.std.ts#L51 and https://github.com/signalapp/Signal-Desktop/blob/main/ts/util/Attachment.std.ts#L235 ) 

So it is not possible at all to send large images. Neither as Image nor as File.

### DIFF
--- a/packages/frontend/src/hooks/chat/useDraft.ts
+++ b/packages/frontend/src/hooks/chat/useDraft.ts
@@ -10,6 +10,7 @@ import { T } from '@deltachat/jsonrpc-client'
 
 import { getLogger } from '../../../../shared/logger'
 import { BackendRemote, Type } from '../../backend-com'
+import { LARGE_IMAGE_PIXEL_THRESHOLD } from '../../components/attachment/Attachment'
 import { MessageTypeAttachmentSubset } from '../../components/attachment/Attachment'
 import { KeybindAction } from '../../keybindings'
 import useMessage from './useMessage'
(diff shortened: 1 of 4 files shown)


packages/frontend/src/hooks/chat/useDraft.ts
```diff
@@ -23,6 +24,35 @@ import useAlertDialog from '../dialog/useAlertDialog'
...
+    img.onload = () =>
+      done({ width: img.naturalWidth, height: img.naturalHeight })
+    img.onerror = () => done(null)
+    // Build a file:// URL.  On Unix the path already starts with '/' so we
+    // get exactly three slashes: file:// + /path = file:///path.
+    // On Windows paths look like C:/... so we prepend an extra slash.
+    const normalized = filePath.replace(/\\/g, '/')
+    img.src = normalized.startsWith('/')
+      ? `file://${normalized}`
+      : `file:///${normalized}`
```
WofWca, line 52: Is there a need to introduce a new way of constructing a value for the `src` attribute instead of utilizing what we already have e.g. here?

https://github.com/deltachat/deltachat-desktop/blob/8768f68f5befeab9f6a9fc7f28f35b08de241a2a/packages/frontend/src/components/attachment/messageAttachment.tsx#L157

I know that one comes from Core and the other from Electron, but aren't they formatted the same way, i.e. both are just an absolute path?

WofWca (commented):
The change to `messageAttachment.ts` seems good to me. Such mega-large messages are rare (in this decade at least), and it's OK to let them open it in another program.

But about the rest: firstly it sounds like it would be better to fix this in Core, or at least wait for the resolution in case it can't be thoroughly fixed.

Otherwise: do we actually want this warning dialog, UX-wise? I mean what is the user even supposed to do with that information? They probably already know that the image is large, and the fact that it's not going to get rendered as an image can be discovered when they actually send the image.
It also means that we would need to add the same UI to Android and iOS, because if one sends such an image from Android then the image would also be displayed as a file on Desktop?

nicodh:
> But about the rest: firstly it sounds like it would be better to fix this in Core, or at least wait for the resolution in case it can't be thoroughly fixed.

Yes the final fix should happen in core of course

> Otherwise: do we actually want this warning dialog, UX-wise? I mean what is the user even supposed to do with that information? They probably already know that the image is large, and the fact that it's not going to get rendered as an image can be discovered when they actually send the image. 

That's the point: the image is visible in the draft but not anymore after sending it. So after the fix in core we could propose to send it as file instead image.

We could decide to convert inherently such images to viewType file and show it as such in the draft. Then we don't need a warning anymore.

WofWca:
> the image is visible in the draft but not anymore after sending it

I see. That also sounds like something that we should do in Core. If not then, again, it means that we'd need to re-implement this on Android, iOS.

nicodh:
Closing as it is not clear if we want to show such a warning at all

PR #6432, opened 2026-06-02
chore: update nix flake, tauri and tauri plugins

Simon-Laux:
- **update nix flake and electron dependency of nix devShell (was outdated)**
- **move tauri and plugin versions together and be more explicit about the versions**
- **update tauri and plugins**

Reason for this is that I wanted to install/test the nix package for tauri (https://github.com/NixOS/nixpkgs/pull/398998) and it did not work because of version miss-matches. I hope pinning the versions will prevent the problem in the future.

I did a smoke test on linux, it worked fine. Though maybe we should also do a quick test for dragging the titlebar on macOS, changelog mentioned some changes in that area.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -14,6 +14,10 @@
     nixpkgs,
     rust-overlay,
   }: let
+    inherit (nixpkgs) lib;
+    electronVersionSpec = (lib.importJSON ./packages/target-electron/package.json).devDependencies.electron;
+    electronVersion = lib.removePrefix "^" electronVersionSpec;
+
     supportedSystems = ["x86_64-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin"];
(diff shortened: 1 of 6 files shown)


flake.nix
```diff
@@ -45,8 +45,8 @@
           gobject-introspection
           cargo
           cargo-tauri
-          nodejs_22 # what electron uses
-          electron_37 # electron version should be same as defined in packages/target-electron/package.json
+          nodejs_22 # what project defines
+          electron_41 # electron version should be same as defined in packages/target-electron/package.json
```
dotlambda: You can use `lib.importJSON` and then parse the version and use `pkgs."electron_{lib.versions.major version}"` here.

nicodh (approved):
Reviewed by reading

As this change doesn't affect production software it can be merged imo

PR #6433, opened 2026-06-03
use electron version from package.json

dotlambda:
see https://github.com/deltachat/deltachat-desktop/pull/6432#pullrequestreview-4414295641
cc @Simon-Laux


Simon-Laux (approved):

dotlambda:
Btw the same could be done for nodejs.

PR #6434, opened 2026-06-03
Build on prepare release branches

nicodh:
This should trigger the release build artefacts...

### DIFF
--- a/.github/workflows/build-tauri-release.yml
+++ b/.github/workflows/build-tauri-release.yml
@@ -2,8 +2,8 @@ name: Build Tauri Release
 
 on:
   push:
-    tags:
-      - 'v*'
+    branches:
+      - '*prepare-release*'
 
 permissions: {}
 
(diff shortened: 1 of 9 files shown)


.github/workflows/build-tauri-release.yml
```diff
@@ -49,11 +49,17 @@ jobs:
           npm i -g pnpm
           pnpm install
 
+      - name: set version ref
+        shell: bash
+        run: |
+          VERSION=$(jq -r .version package.json)
+          echo "VERSION_INFO_GIT_REF=v${VERSION}" >> "$GITHUB_ENV"
```
WofWca, line 56: Note that in other places we take a slightly different approach.

https://github.com/deltachat/deltachat-desktop/blob/06cdcd64d733592acea47d1e2fdcf0087d758687/.github/workflows/build-preview.yml#L41-L45

But this also seems fine. GitHub docs show something similar:
https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#environment-files

WofWca (approved):
Seems OK. I think updating the var names is not necessary, because this is basically saying "this is gonna be the git ref when we push the tag".

Maybe worth adding some comments about this, why in some places we use `${{ github.ref }}` and here we use the `package.json` field.

Edit: also I think it's still safe to trigger on branch instead of a tag, firstly because outside contributors can't push branches to the main repo, secondly because all that the workflow is doing is simply create artifacts.

nicodh:
Closing as these changes were force pushed to the pre release branch

PR #6435, opened 2026-06-03
fix: do not render large images

nicodh:
This is related to #6430 and just avoids the rendering of large images

No warning is shown to the user if he uploads such an image but the image will not be displayed in the UI anymore. The OOM error in core should be fixed there.

### DIFF
--- a/packages/frontend/src/hooks/chat/useDraft.ts
+++ b/packages/frontend/src/hooks/chat/useDraft.ts
@@ -309,6 +309,7 @@ export function useDraft(
       if (debouncedSaveAndRefetchDraft == null || saveAndRefetchDraft == null) {
         return
       }
+
       inputRef.current?.focus()
       const newDraftState: typeof draftState = {
         ...draftState,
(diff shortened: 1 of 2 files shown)


packages/frontend/src/hooks/chat/useDraft.ts
```diff
@@ -309,6 +309,7 @@ export function useDraft(
       if (debouncedSaveAndRefetchDraft == null || saveAndRefetchDraft == null) {
         return
       }
+
```
WofWca, line 312:
```suggestion
```

WofWca (approved):
As I said in the previous MR, it's fine (and probably good) to render such images as files because then the user can just open it externally.

The removal of `else` does not affect behavior, it's only the `if`.

Although now that I'm testing again, having the image in the chat or in the draft area doesn't cause that much lag, so maybe Chromium got better at handling this and the problem is no longer relevant? I'll comment on the issue.

PR #6437, opened 2026-06-04
fix: avoid static hard coded lang attribute in html tag

### LINKED ISSUE #6436: Screen reader stick to english language
While most of the label are translated to the correct language when switching language, some screen reader still use an english speaking voice and name generic elements in english, no matter what language the user selected

nicodh:
fixes #6436

### DIFF
--- a/packages/frontend/src/App.tsx
+++ b/packages/frontend/src/App.tsx
@@ -69,6 +69,7 @@ function I18nContextWrapper({ children }: { children: React.ReactElement }) {
     setLocaleData(localeData)
     moment.locale(localeData.locale)
     updateCoreStrings()
+    document.documentElement.lang = localeData.locale.replace('_', '-')
   }
 
   useLayoutEffect(() => {
--- a/packages/frontend/static/main.html
+++ b/packages/frontend/static/main.html
@@ -1,5 +1,5 @@
 <!doctype html>
-<html lang="en">
+<html>
   <head>
     <meta charset="utf-8" />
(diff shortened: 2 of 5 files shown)


packages/frontend/src/App.tsx
```diff
@@ -69,6 +69,7 @@ function I18nContextWrapper({ children }: { children: React.ReactElement }) {
     setLocaleData(localeData)
     moment.locale(localeData.locale)
     updateCoreStrings()
+    document.documentElement.lang = localeData.locale.replace('_', '-')
```
WofWca, line 72: `locale` and `localeData.locale` are the same? Then we don't need to `await getLocaleData()`.
But not a big deal.

WofWca (approved):
Makes sense, but I didn't check whether this actually fixes the issue. I also didn't double-check though if the `replace('-', '_')` part is correct, but we already have it elsewhere:

https://github.com/deltachat/deltachat-desktop/blob/62a682cce0653f38330c4459778df41d47a99fc8/packages/shared/localize.ts#L31

This doesn't affect Tauri and Browser, but can be done in follow-ups.

MarcoZehe:
I can confirm that this fixes the issue for me. I've tried the build on the Mac, and VoiceOver now uses a German voice for all the text which is in German. It used to use the English voice because DeltaChat told it that the UI was in English due to the hard-coded language attribute. Thanks for fixing this so quickly!

r10s (approved):
makes a lot of sense! thanks a taking care!

PR #6441, opened 2026-06-07
dev: force local links to jsonrpc client

nicodh:
It turned out that running pnpm add link in each project resulted in an inconsistent pnpm.lock file. So running pnpm check:types failed.

Since `pnpm add` includes a `pnpm install` the already "resolved" references in lockfile were not updated anymore. So now we replace the references manually in each package.json and then run `pnpm install --force`


nicodh:
Rebased!
For testing:
run
`python ./bin/link_core/build_and_link_local_core.py <path-to-local-core>`
 
and check types

WofWca (approved):
I didn't get what the problem was, but the resulting code looks OK, and seems to work.

nicodh:
> I didn't get what the problem was, but the resulting code looks OK, and seems to work.

after 
`python ./bin/link_core/build_and_link_local_core.py <path-to-local-core>`

running pnpm `check:types` failed since some references to the remote jsonrpc lib were left

with these changes all references are replaced

PR #6443, opened 2026-06-08
feature: command palette

nicodh:
This PR adds a new command palette dialog to desktop as proposed in https://support.delta.chat/t/command-palette/2390

The basic idea is to have better user experience for switching accounts & chats, search in current or other accounts and get direct access to many options that are typically hidden in context menues or settings.

For now it only supports search in accounts or chats (with an optional unread filter) and switching accounts or jumping to chats or messages.

Most of the UI concept is taken from github command palette (can be enabled in settings->feature preview). It also has a breadcrumb showing the current scope and the listed items/options offer a jump-to (enter) or a tab-into (tab) action.  

 Since there are tons of possible feature here let's start with basic features first.
 
 My idea for next steps are:
 
 1. add the command mode (triggered by typing > or opening the dialog with Ctrl + P)
 2. the command mode shows the first items from "createChat" dialog plus some of the related context menu actions (depending on scope)
 3. entering @ shows a list with all contacts where enter would create a 1:1 chat or jump to an existing chat, tab would search only in messages from that contact (needs a new core function searchBySender) 
 4. adding a ? would show some help text explaining the usage and different modes

Current state:

The palette is opened in account scope. Entering a query searches the whole account

initial view

The account crumb can be deleted with backspace then we are in root scope. Enter switches to the selected account, tab narrows the search to that account (without switching the account)

switch-account
 
 
 In Chat scope the breadcrumb shows the account and the chat to search in:
 
 
serach-in-chat


nicodh:
Concerning AI usage I would like to mention: I used AI for specific tasks like reviewing my code, asking for specific solutions (like how to improve accessibility) but always double checked proposed solutions and mostly modified them

packages/frontend/src/components/dialogs/CommandPalette/usePaletteSearch.ts
```diff
@@ -0,0 +1,309 @@
...
+        })
+      }
+    }
+
+    // Debounce so we don't fire a search on every keystroke.
+    const timeout = setTimeout(run, 200)
+    return () => {
+      cancelled = true
+      clearTimeout(timeout)
+    }
```
WofWca, line 338: I think that throttle would be nicer. Faster. We already utilize it for other searches.

packages/frontend/src/components/dialogs/CommandPalette/usePaletteSearch.ts
```diff
@@ -0,0 +1,309 @@
...
+              )
+              actionsRef.current.close()
+            },
+          })
+        }
+      } catch (error) {
+        log.error('failed to search contacts', error)
+      }
+      return items
+    }
```
WofWca, line 227: I think it'd be better to handle errors outside of the function instead of wrapping its entire body.

```javascript
.catch(e => {
  console.log('failed foo:', e)
  return [];
}
```

Same goes for all other usages of `try` in this MR.

packages/frontend/src/components/dialogs/CommandPalette/index.tsx
```diff
@@ -0,0 +1,437 @@
...
+            <Icon icon='clear' size={18} />
+          </button>
+        )}
+      </div>
+      <div className={styles.list} ref={listRef}>
+        {showNoResults && (
+          <div className={styles.empty} role='status'>
+            {tx('search_no_result_for_x', query)}
+          </div>
+        )}
```
WofWca:
```suggestion
        <div role='status' style={{ display: 'contents' }}>
          {showNoResults && (
            <div className={styles.empty}>
              {tx('search_no_result_for_x', query)}
            </div>
          )}
        </div>
```

`aria-live` only announces when the _contents_ changes and not when the element with `aria-live` itself appears.

packages/frontend/src/components/dialogs/CommandPalette/index.tsx
```diff
@@ -0,0 +1,437 @@
...
+      popScope()
+    }
+  }
+
+  const placeholder =
+    scope === 'root'
+      ? tx('switch_account')
+      : scope === 'chat'
+        ? tx('search_in_chat')
+        : tx('command_palette_placeholder')
```
WofWca: Do we really need a new string?

```suggestion
        : tx('search_explain')
```

https://github.com/deltachat/deltachat-desktop/blob/84d413329393c589a7e9e4e43b3ff7d502b5f92b/_locales/en.xml#L557

packages/frontend/src/components/dialogs/CommandPalette/index.tsx
```diff
@@ -0,0 +1,437 @@
...
+    accountName = accountInfo.displayName || accountInfo.addr || ''
+    avatarPath = accountInfo.profileImage || undefined
+    accountColor = accountInfo.color || undefined
+  }
+  const chatName = scopedChat?.name ?? ''
+
+  const actions: PaletteActions = useMemo(
+    () => ({
+      selectChat: async (accId, cId) => {
+        if (accId === accountId) {
```
WofWca: Since we're dealing with cross-account stuff, which is quite a sensitive thing IMO, it would be nice to at least have more explicit names. e.g. `selectedAccountId` (or `activeAccountId`) and `targetAccountId`.

packages/frontend/src/components/dialogs/CommandPalette/index.tsx
```diff
@@ -0,0 +1,437 @@
...
+    entries: { item: PaletteItem; index: number }[]
+  }[] = []
+  items.forEach((item, index) => {
+    const last = groups[groups.length - 1]
+    if (last && last.section === item.section) {
+      last.entries.push({ item, index })
+    } else {
+      groups.push({ section: item.section, entries: [{ item, index }] })
+    }
+  })
```
WofWca: Wrap in `useMemo`.

```suggestion
  const groups = useMemo(() => {
    const groups: {
      section: PaletteSection
      entries: { item: PaletteItem; index: number }[]
    }[] = []
    items.forEach((item, index) => {
      const last = groups[groups.length - 1]
      if (last && last.section === item.section) {
        last.entries.push({ item, index })
      } else {
        groups.push({ section: item.section, entries: [{ item, index }] })
      }
    })
    return groups
  }, [items])
```

packages/frontend/src/components/dialogs/CommandPalette/types.ts
```diff
@@ -0,0 +1,38 @@
+export type PaletteSection = 'accounts' | 'chats' | 'contacts' | 'messages'
+
+export type PaletteScope = 'root' | 'account' | 'chat'
+
+/** Enough info to scope the palette to (and label the crumb for) an account. */
+export type PaletteAccountRef = {
```
WofWca: This is just a partial version of the account type? The name sounds like it's somehow related to React refs.

```suggestion
export type AccountPartial = {
```

packages/frontend/src/keybindings.ts
```diff
@@ -170,6 +171,8 @@ export function keyDownEvent2Action(
       return KeybindAction.ChatList_FocusSearchInput
     } else if ((ev.metaKey || ev.ctrlKey) && matchesLetterShortcut(ev, 'n')) {
       return KeybindAction.NewChat_Open
+    } else if ((ev.metaKey || ev.ctrlKey) && matchesLetterShortcut(ev, 'k')) {
```
WofWca, line 217: Why `Ctrl + K` and not `Ctrl + P`? Isn't the latter more common? Also what the forum post suggests. Then `Ctrl + Shift + P` for commands.

packages/frontend/src/components/dialogs/CommandPalette/usePaletteItems.ts
```diff
@@ -0,0 +1,309 @@
...
+
+export function usePaletteItems({
+  accountId,
+  scope,
+  chatId,
+  query,
+  actions,
+}: UsePaletteItemsParams): { items: PaletteItem[]; isLoading: boolean } {
+  // `forKey` records which (scope, account, chat, query) the items belong to.
+  const currentKey = `${scope}\n${accountId}\n${chatId ?? ''}\n${query.trim()}`
```
WofWca: MR description says

> (with an optional unread filter)

Apparently this has been plant but has not been implemented?

WofWca (approved):
I think it's OK to merge, at least as an experiment. I recognize that this adds a maintenance burden. But I think we can afford this, as long as we
1. Treat this as basically an extension (like a browser extension) which is purely additive and does not do anything that can't be done otherwise.
2. Keep the code compartmentalized (which it is now IMO).
3. Keep the code quality high enough to be sure that this doesn't break anything and doesn't have significant bugs (like sending a message to a wrong account or something).
4. Don't feel obligated to maintain this feature (e.g. review pull requests for it).
5. Are going to be okay with removing this if there is a need.

However what I think we should be prioritize instead is improving keyboard accessibility for the stuff that we already have.
In the end I think it is achievable to have UX as smooth and as fast as one can achieve with a command palette widget if we polish (or rethink) our existing stuff.
For example, the search that we already have is quite a lot similar to what this PR introduces.
And typing a command name can be replaced with typing to focus (or activate) a menu item in the context menu, like it is on native context menus in most OSes and browsers
Another thing that needs work is to figure out when it's okay to clear the search query.
Another thing is improving focus management in the chat list: one thing that comes to mind is having the currently active chat as the one initially focused.
Well, basically we need to adopt more of the combobox pattern practices to our existing search.

I understand that changing existing stuff is usually met with more resistance that just writing new stuff, but to circumvent that we can utilize the "experimental" toggles, like e.g. we had for the "Ctrl + Up to quote" shortcut.

The new code doesn't use any effects for backend functions, so it will not accidentally send a message somewhere.
And the code looks alright otherwise, but I didn't review super thoroughly.

I did not search the internet to check for copyright violations, but looks like this MR doesn't introduce this much significantly complex and context-independent code.

packages/frontend/src/components/dialogs/CommandPalette/usePaletteSearch.ts
nicodh: But wouldn't that trigger more unneeded requests? Where do we use it in other searches?

packages/frontend/src/components/dialogs/CommandPalette/usePaletteSearch.ts
nicodh: Ok, changed it

packages/frontend/src/components/dialogs/CommandPalette/index.tsx
nicodh: Ok, I oversaw that. Removed it

packages/frontend/src/components/dialogs/CommandPalette/index.tsx
nicodh: I renamed the on to currentAccountId and the other to scopeAccountId

packages/frontend/src/components/dialogs/CommandPalette/index.tsx
nicodh: Makes sense

packages/frontend/src/components/dialogs/CommandPalette/usePaletteItems.ts
nicodh: Yes PR is on it's way

packages/frontend/src/keybindings.ts
nicodh: I thought using Ctrl+P to open the palette directly in command mode, but that can be discussed of course

packages/frontend/src/components/dialogs/CommandPalette/index.tsx
nicodh: Applied your suggestion

packages/frontend/src/components/dialogs/CommandPalette/types.ts
nicodh: Agreed

packages/frontend/src/components/dialogs/CommandPalette/usePaletteSearch.ts
WofWca: That would trigger more requests, but I wouldn't say that they're unneeded. It's nicer to get more immediate feedback IMO, instead of always having to wait extra 200ms for the results to appear. Here is our chat list search code:

https://github.com/deltachat/deltachat-desktop/blob/4b28a6ebb845051dfc6a549d2bf63e5ccd7d10c1/packages/frontend/src/components/chat/ChatListHelpers.tsx#L172-L194

But maybe it's a matter of taste IDK.

packages/frontend/src/keybindings.ts
WofWca: OK, fine either way, can be changed later.

WofWca (approved):
Tested again, still works.

r10s (commented):
i agree, this should be treated as an experiment, and we should not put resources into that nerd-mode that would be missing otherwise.

everything possible in that "command palette" should also be possible otherwise, so that we do not push normal users to that mode.

for nerds it is nice, tho :)

some things during testing, non-exhaustive, non-blocking:

- when opening, on mac, there is always the string `>` in the search input field. it should be empty
- let's remove the tiny icons before "commands", "chats", "switch profile" etc. headlines - they do add clutter only, are not used otherwise (switch profile is even out connectivity icon), and are by far not self-explaining. there is no point in having icons everywhere. and having less things make maintenance easier
- the searchline should always be same height - currently height changes when an avatar is added. let's make the avatars in the seachline smaller therefore
- profile avatars are square otherwise
- for the "hover/select" color, do not invent a new grey tone, but use the existing gray from hovering in  the chatlist

link2xt:
> when opening, on mac, there is always the string > in the search input field. it should be empty

This is when you open with Ctrl-P. Ctrl-K opens quick search, and then typing ">" turns the search into a command palette. Removing ">" switches back to search.

PR #6445, opened 2026-06-09
Command palette followup

nicodh:
This PR adds commands to the command palette :-) and a basic filter (only for unread chats so far)

Some notes:

1. I added only few commands to not overload this PR (more can be added easily later)

2. I did 2 slight changes on code outside the CommandPalette, replacing the (customized) archive icon, we used only as avatar so far - with the default material icon and adapted the style for that avatar and allow to pass an initialAction to the createChat dialog.

3. to avoid complexity command mode is not possible for other account scopes than the current active account

4. I added an alias (:unread) for the existing is:unread filter for 2 reasons: no need to type 10 characters for a filter and maybe later use the colon as a token for a filter mode which could be extended and show possible filter in the list so no typing is needed anymore.

5. When canceling some of the command triggered dialogs (the createChat 'new...' items) the original "parent" dialog is opened but I consider that ok

Command mode in chat scope

image

Unread filter active

image


nicodh:
Just as an outlook:

1. I would like to get at least one more feature in: the token @ should list all contacts with an option to create/jump-to a chat with that contact or search messages from that contact (needs a core function)
2. I do not plan to extend this command palette "endlessly"
3. I agree with the restrictions mentioned by @WofWca in the first PR, although I consider it a difference if you have to memorize many Shortcuts to switch chats, accounts, search etc. or just one to open the palette and then continue "intuitively"

link2xt:
It seems the easiest way to access chat commands is:
1. Opening the search with Ctrl-K
2. Finding the chat
3. Pressing "Tab" to enter it
4. Typing ">"

The last step is not discoverable, I only found that it is possible from reading the PR description.

And with the limitation of not being able to access the commands for chats in the profile that is not currently selected, maybe a better approach would be to add commands like "delete chat" to the command palette that opens with Ctrl-P and apply these commands to the active chat? If i already have the chat open and want to e.g. archive it, currently i need to find it again instead of pressing Ctrl-P and typing "Archive".

packages/frontend/src/components/dialogs/CreateChat/index.tsx
```diff
@@ -86,11 +86,22 @@ const enum GroupType {
 
 type ViewMode = 'main_' | GroupType
 
-export default function CreateChat(props: DialogProps) {
+/** Lets callers (e.g. the command palette) open CreateChat in a specific view. */
+export type CreateChatInitialAction = 'new-group' | 'new-channel'
```
WofWca: Can just reuse the `GroupType` type, no need for ternaries then

Edit: ah, this will have to be rewritten anyway. We removed `setViewMode` recently.

packages/frontend/src/components/dialogs/CommandPalette/index.tsx
```diff
@@ -333,6 +411,24 @@ export default function CommandPalette({ onClose }: Props) {
             <span className={styles.crumbSep}>/</span>
           </span>
         )}
+        {scope === 'account' && filter === 'unread' && (
+          <span className={styles.filterCrumb}>
+            <span className={styles.crumbLabel}>
+              {tx('command_palette_filter_unread')}
```
WofWca: And remove the added string.

```suggestion
              {tx('search_unread')}
```

https://github.com/deltachat/deltachat-desktop/blob/84d413329393c589a7e9e4e43b3ff7d502b5f92b/_locales/en.xml#L560-L561

packages/frontend/src/components/chat/ChatListItem.tsx
```diff
@@ -205,8 +205,8 @@ function ChatListItemArchiveLink({
         isContextMenuActive ? 'context-menu-active' : ''
       }`}
     >
-      <div className='avatar'>
-        <img className='content' src='./images/icons/icon-archive.svg' />
+      <div className='avatar archive-avatar' aria-hidden={true}>
```
WofWca: Let's either not add irrelevant changes, or utilize `alt=""` instead.

WofWca: The icon is invisible on dark theme. Consider utilizing `currentColor`, as in the original icon.

Image

packages/frontend/src/components/dialogs/CommandPalette/usePaletteCommands.ts
```diff
@@ -0,0 +1,193 @@
...
+  const chatIdToFetch = scopedChat?.id
+  const entries = useMemo(
+    () => (chatIdToFetch != null ? [chatIdToFetch] : null),
+    [chatIdToFetch]
+  )
+  const chatItemFetch = useRpcFetch(
+    BackendRemote.rpc.getChatlistItemsByEntries,
+    entries != null ? [accountId, entries] : null
+  )
+  const chatItemResult = chatItemFetch?.lingeringResult
```
WofWca: Using `lingeringResult` instead of the regular `result` has a potential to run the actions on the wrong chat. Especially important for "delete".

packages/frontend/src/components/dialogs/CommandPalette/usePaletteCommands.ts
```diff
@@ -0,0 +1,193 @@
...
+
+import type { PaletteItem } from './types'
+import Settings from '../../Settings'
+
+export type UsePaletteCommandsParams = {
+  accountId: number
+  /** The chat in the breadcrumb, if any — enables chat-specific commands. */
+  scopedChat: { id: number; name: string } | null
+  /** The chat currently open in the main view, so delete can unselect it. */
+  selectedChatId: number | null
```
WofWca: Since the introduction of multiselect we've termed this `activeChatId`. It would be nice to use the same terminology here not just for consistency but for distinguishing between "selected" as in "selected to perform actions on in the command palette" or as in "the chat in the main chat view".

WofWca (commented):
I only smoke-tested this. The code looks fine overall.

packages/frontend/src/components/dialogs/CreateChat/index.tsx
nicodh: Is refactored

packages/frontend/src/components/chat/ChatListItem.tsx
nicodh: I used Icon component now

nicodh:
>  maybe a better approach would be to add commands like "delete chat" to the command palette that opens with Ctrl-P and apply these commands to the active chat? If i already have the chat open and want to e.g. archive it, currently i need to find it again instead of pressing Ctrl-P and typing "Archive".

Now the palette opens in command mode with the active chat already selected

WofWca (approved):
The archive icon looks much brighter than before (and than on Android), but I think it's alright

Image

Image

I also played around with commands for a bit. Seems to work alright.
The code outside of `CommandPalette/` is almost untouched by this MR.

PR #6446, opened 2026-06-09
chore: upgrade core to 2.52.0

nicodh:


WofWca (approved):
Just a bump (but didn't check `Cargo.lock`). Looks like there are no breaking changes for us to address https://github.com/chatmail/core/blob/main/CHANGELOG.md#2520---2026-06-09.

PR #6448, opened 2026-06-09
fix: undismissable context menu in fullscreen media

WofWca:
Unclosable for keyboard users without activating a menu item.

1. Click on a photo in a chat to start viewing it in full screen.
2. Tab-focus the image element.
3. Invoke the context menu.
4. Press Escape.

Nothing will happen, whereas the context menu should close.

The problem is `preventDefault()` prevents the default action
which is to close the context menu dialog.

This partially reverts 6ce4d78f30cce9dfabe8288876f690c8a134ec42
(https://github.com/deltachat/deltachat-desktop/pull/5191),
which introduced this hack.
The hack became a bug apparently after
0795308eaeac0e3f9084339695926cdbf8857dab
(https://github.com/deltachat/deltachat-desktop/pull/6226).

Related:
- https://github.com/deltachat/deltachat-desktop/issues/4128.

### DIFF
--- a/packages/frontend/src/components/dialogs/FullscreenMedia.tsx
+++ b/packages/frontend/src/components/dialogs/FullscreenMedia.tsx
@@ -13,7 +13,7 @@ import { useInitEffect } from '../helpers/hooks'
 import { BackendRemote, onDCEvent, Type } from '../../backend-com'
 import { selectedAccountId } from '../../ScreenController'
 import useTranslationFunction from '../../hooks/useTranslationFunction'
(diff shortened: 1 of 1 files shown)


nicodh (approved):
Reviewed by reading & testing

The title is a bit misleading. The context menu is closable but not with just pressing escape. Clicking on any space outside the menu closes it. But of course it's better when it is also closable with escape

WofWca:
Sorry, forgot to say that I meant keyboard-only users. I edited the MR description.

PR #6449, opened 2026-06-09
refactor: FullscreenMedia no global event listener

WofWca:
The comment also appears to be no longer relevant.
It appears to refer to a hypothetical shortcut handler function
which handles all shortcuts in all dialogs.
It's just simpler to have them scoped per component.

I have tested this. Still works.

TODO:
- [x] Merge the base.


nicodh (approved):
Reviewed by reading & testing

PR #6451, opened 2026-06-09
fix: focus composer when user tries to type

### LINKED ISSUE #6246: Esc should focus the input bar again
When the focus is somewhere else (e.g. in the search), and I press Escape, it would be nice if the focus can go back to the input bar.

WofWca:
This is a more accessible and common-practice way to do what
8fe1381b017eec203ae5df0da5de3188129da378
(https://github.com/deltachat/deltachat-desktop/pull/2007)
did.
We sort of reverted that MR for the most part, as part of
https://github.com/deltachat/deltachat-desktop/issues/4590
("a11y: stop constantly re-focusing composer"),
which was necessary for accessibility
but was, unfortunately, mostly annoying to pointer device users.

This commit should basically fix the "composer not focused" issues
once and for all (or for 95% at least).
And this also allows us to be more liberal in focuing other stuff,
e.g. focusing a message after a click on a notification.

I have tested this with Orca Reader and NVDA,
specifically with those one-key commands that they have,
such as "u" to go to the next link in Orca Reader.
Everything works fine. I confirmed that pressing "u"
does not move the focus to the composer and does not type "u" in it.
Same for NVDA.
This, of course, changes once one enters the focus mode
("Insert + A" on Orca Reader).

Closes https://support.delta.chat/t/always-focus-the-message-field-in-a-chat/4271?u=wofwca.
I would say that this also
closes https://github.com/deltachat/deltachat-desktop/issues/6246.

This is quite a significant change, so we'd probably want to do quite some testing of this.

TODO:
- [x] Merge the base, https://github.com/deltachat/deltachat-desktop/pull/6392.
  This MR does not really depend on that one
  but they simply change the same code, the `onMouseUp` part.
- [ ] (can do later) add tests

### DIFF
--- a/packages/frontend/src/keybindings.ts
+++ b/packages/frontend/src/keybindings.ts
@@ -126,6 +126,20 @@ export function matchesNonLetterShortcut(
   const isPrintableAscii = ev.key.length === 1 && ev.key >= ' ' && ev.key <= '~'
   return !isPrintableAscii && ev.code === code
 }
+const regexIsPrintable = /\S| /
+function isPrintableCharacter(str: KeyboardEvent['key']): boolean {
+  // From https://www.w3.org/WAI/ARIA/apg/patterns/combobox/
+  return str.length === 1 && regexIsPrintable.test(str)
+}
+
+function isInput(el: HTMLElement): boolean {
+  return (
+    el.tagName === 'TEXTAREA' ||
+    el.tagName === 'INPUT' ||
+    el.tagName === 'SELECT' ||
+    el.isContentEditable
+  )
+}
 
 function haveOpenDialogs(): boolean {
   return document.querySelector('dialog:modal') != null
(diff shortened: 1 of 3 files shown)


packages/frontend/src/keybindings.ts
```diff
@@ -126,6 +126,20 @@ export function matchesNonLetterShortcut(
   const isPrintableAscii = ev.key.length === 1 && ev.key >= ' ' && ev.key <= '~'
   return !isPrintableAscii && ev.code === code
 }
+const regexIsPrintable = /\S| /
+function isPrintableCharacter(str: KeyboardEvent['key']): boolean {
+  // From https://www.w3.org/WAI/ARIA/apg/patterns/combobox/
+  return str.length === 1 && regexIsPrintable.test(str) != null
```
nicodh: regexIsPrintable.test(str)  is never null since it always returns a boolean, so the regex is basically only str.length === 1 (which is fine according to my research)

https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test

nicodh (approved):
Reviewed by reading & testing and doing some own research.

Good approach to solve this!! 

Maybe some feedback from who is using screenreaders in everyday life would be nice

packages/frontend/src/keybindings.ts
WofWca: Oopise, you're right. I forgot that when migrating from `.match()` to `.test()`. Fixed it.
I think that it makes sense to keep it while the W3 docs also have it. At least until we know for certain that it's not needed.

MarcoZehe:
I gave the test build a spin on MacOS with VoiceOver, and so far haven't found any problems. VoiceOver has a quick navigation mode, too, but this needs to be enabled specifically with VoiceOver+Q. In that mode, keys that are assigned execute the VoiceOver command and are not passed through, so they work as expected. Others that are not assigned, do set focus to the input field. That is OK, since VoiceOver decides what to trap and what not. If quick nav mode is off, all keys like letters etc. set focus to the input field. I can still tab through the UI, and I can arrow through the list of chats and list of messages in a chat just fine without losing focus. And I look forward to the Escape key restoring focus to the element when closing popups. Right now, focus lands somewhere unpredictable, even in a release build, so that fix will also improve keyboard accessibility significantly. Thanks for all the work!

PR #6454, opened 2026-06-10
feat: show icon when proxy is enabled

### LINKED ISSUE #5509: no hint that a proxy is enabled, no easy way to switch proxies

Image

Image

nicodh:
New (clickable) Icon when proxy is enabled. Shown in green when connected, otherwise orange. 

resolves #5509

image

tbd: 
1. showing the icon left from qr code button above the chats has the problem, that that icon is "converted" to a reset button when the search field input is not empty. The current solution has the problem that it's only visible when a chat is selected, but that's ok imo
2. the current solution does not show an empty shield when proxy is enabled but no proxy is set, but an orange icon since the app will keep trying to connect if proxy is enabled

Icon is also visible in connectivity toast, but not clickable. Click on the toast opens the connectivity dialog which is readonly.

image


packages/frontend/src/components/Icon/styles.module.scss
```diff
@@ -32,6 +32,14 @@
       background-color: #bf0000;
     }
   }
+
+  &.success {
+    background-color: #34c759;
+  }
+
+  &.warning {
+    background-color: #ff9500;
```
WofWca, line 41: According to https://github.com/deltachat/deltachat-desktop/pull/4259#issuecomment-2438446558, we don't want to "hard-code" colors.

packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
```diff
@@ -587,6 +617,21 @@ function ChatNavButtons({
         chat.contactIds.some(id => id > C.DC_CONTACT_ID_LAST_SPECIAL) && (
           <CallButton chat={chat} />
         )}
+      {proxyEnabled && (
+        <Button
```
WofWca, line 621: This is not in line with Android, and the original issue seems to say that the icon is supposed to be hidden if (and, I assume, _only if_) the list of proxies is empty.

https://github.com/deltachat/deltachat-android/blob/20fbb0e1ee95dc101234039f2e1794dbe3a96850/src/main/java/org/thoughtcrime/securesms/ConversationListActivity.java#L461-L469

Same goes for the "Connectivity" toast.

packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
```diff
@@ -550,6 +552,34 @@ function ChatNavButtons({
...
+    }
+    return onDCEvent(accountId, 'ConnectivityChanged', refreshConnectivity)
+  }, [accountId, proxyEnabled, refreshConnectivity])
+
+  // When a proxy is enabled, show an icon in the chat navbar
+  // indicating the current connectivity status (green/orange)
+  const proxyConnected =
+    connectivityFetch?.lingeringResult?.ok === true &&
+    connectivityFetch.lingeringResult.value >= C.DC_CONNECTIVITY_WORKING
+
```
WofWca, line 582: It'd be nicer for performance to have this in a global state (e.g. context), shared with the existing "connectivity" toast. We reload connectivity on every alt-tab.
But fine, maybe not too bad.

packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
```diff
@@ -587,6 +617,21 @@ function ChatNavButtons({
...
+      {proxyEnabled && (
+        <Button
+          onClick={openProxyDialog}
+          aria-label={tx('proxy_settings')}
+          title={tx('proxy_settings')}
+          className='navbar-button'
+          styling='borderless'
+        >
+          <Icon
+            coloring={proxyConnected ? 'success' : 'warning'}
```
WofWca, line 629: Note that this is also yellow when `DC_CONNECTIVITY_CONNECTING`. This is [in line](https://github.com/chatmail/core/blob/bd6c9908e455be7704e7fde101d7394caa18dc76/deltachat-jsonrpc/src/api.rs#L2080-L2083) with the "Connectivity" dialog, but IDK if this is how it should be in the "regular" UI.

packages/frontend/src/stores/settings.ts
```diff
@@ -23,6 +23,7 @@ export interface SettingsStoreState {
       is_chatmail: '0' | '1'
       who_can_call_me: WhoCanCallMe
       'ui.mentions_enabled': '0' | '1'
+      proxy_enabled: '0' | '1'
```
WofWca, line 26:
```suggestion
      proxy_enabled: Proxy
```

(this will need some other adjustments)

WofWca (approved):
Tested for a bit. Seems to work as intended. The addition of the icon to the toast is nice IMO.

The irrelevant changes are also good.

packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
WofWca: Ah OK, on Android they also show a yellow dot against the avatar when "connecting".

r10s (commented):
did not build, only by the screenshots:

- the colored icon is not our UI-language, let's do what we and others are doing here, leave it monochrome. this is also consistent with android/ios. for "not connected" we can remove the checkmark. the idea of the issue is about showning the "there is some proxy set" in general, and this is sufficiently done.

- nice to have the icon also beside "connecting"

- i am wondering why the icon is not shown atop of the chatlist. unless there is a good reason, let's move it there - for inner and outer consistency - eg. to other delta chats and eg. to telegram

nicodh:
> * i am wondering why the icon is not shown atop of the chatlist. unless there is a good reason, let's move it there - for inner and outer consistency - eg. to other delta chats and eg. to telegram

Where there? Above the chatlist we have the input field for the search and the qr code scan button which switches to be a cross icon for removing the search query if the user starts typing. So should we narrow  the search input field to show the proxy icon?

r10s:
> So should we narrow the search input field to show the proxy icon?

yes, this is what eg.  telegram is doing as well. it is still wide enough.

the icon also makes a little more sense there, as it is arguably nothing belonging to a single chat

nicodh:
Closed in favour of #6522

PR #6456, opened 2026-06-11
feat: add three-dot menu and member search to group profile

### LINKED ISSUE #6180: add three-dot-menu to group/channel profile
while the "contact profile" has a three dot menu, group/channels profiles have not yet.

the new three dot menu shall contain the following items:

```
Edit
---
Encryption Info
Clone Chat
```

for channels and read-only groups some items may be  hidden.

the three dot menu should be at the place of the current "Edit pencil", which will be removed (edit is more important for contact profiles, and has no separate icon there as well)

the "Clone Chat" and "Encryption Info" items from the chatlist's context menu can be removed then ("Clone Chat" in the chat menu should stay for now)

reason for this change is to streamline the chatlist menu, to stream line options inside desktop, and to be more consistent with other platforms, cmp https://github.com/deltachat/deltachat-pages/pull/1312

nicodh:
resolves #6180 

- move some options from chatlist menu to group/channel profile
- ~~add a search to filter member list~~
- add Ctrl+F shortcut to open the search

This also replaces the (outdated) PR #5697

### DIFF
--- a/packages/e2e-tests/tests/chat-context-menu.spec.ts
+++ b/packages/e2e-tests/tests/chat-context-menu.spec.ts
@@ -376,7 +376,6 @@ test.describe('Chat List Context Menu - Single Selection', () => {
       'Mute Notifications',
       'Archive Chat',
       'View Profile',
-      'Encryption Info',
       'Block Contact',
       'Delete',
     ])
@@ -398,8 +397,6 @@ test.describe('Chat List Context Menu - Single Selection', () => {
       'Mute Notifications',
       'Archive Chat',
       'View Profile',
-      'Encryption Info',
-      'Clone Chat',
       'Leave Group',
     ])
 
@@ -452,12 +449,14 @@ test.describe('Chat List Context Menu - Single Selection', () => {
     await switchToProfile(page, userA.id)
 
     await openChatListContextMenu(page, userB.name)
-    await page.getByRole('menuitem', { name: 'Encryption Info' }).click()
+    await page.getByRole('menuitem', { name: 'View Profile' }).click()
+    await page.locator('#view-profile-menu').click()
+    await page.getByTestId('encryption-info').click()
 
     // Verify encryption info dialog opens
-    await expect(page.getByRole('dialog')).toBeVisible()
-    await expect(page.getByRole('dialog')).toContainText('Encryption')
+    await expect(page.getByRole('dialog').last()).toContainText('Encryption')
 
+    await page.keyboard.press('Escape')
     await page.keyboard.press('Escape')
   })
 
@@ -467,13 +466,16 @@ test.describe('Chat List Context Menu - Single Selection', () => {
     await switchToProfile(page, userA.id)
 
     await openChatListContextMenu(page, groupName)
-    await page.getByRole('menuitem', { name: 'Clone Chat' }).click()
+    await page.getByRole('menuitem', { name: 'View Profile' }).click()
+    await page.getByTestId('view-group-menu').click()
+    await page.getByTestId('clone-chat').click()
 
-    // Verify clone chat dialog opens
-    await expect(page.getByRole('dialog')).toBeVisible()
-    // The clone dialog should show create group interface
+    // Verify clone chat dialog opens,
     await expect(page.locator('.group-name-input')).toBeVisible()
+    // it should show the create group interface
 
+    // close clone dialog and group profile dialog
+    await page.keyboard.press('Escape')
     await page.keyboard.press('Escape')
   })
 })
--- a/packages/e2e-tests/tests/chat-context-menu.spec.ts
+++ b/packages/e2e-tests/tests/chat-context-menu.spec.ts
@@ -472,7 +472,6 @@ test.describe('Chat List Context Menu - Single Selection', () => {
 
     // Verify clone chat dialog opens,
     await expect(page.locator('.group-name-input')).toBeVisible()
-    // it should show the create group interface
 
     // close clone dialog and group profile dialog
     await page.keyboard.press('Escape')
--- a/packages/e2e-tests/tests/unencrypted-group-tests.spec.ts
+++ b/packages/e2e-tests/tests/unencrypted-group-tests.spec.ts
@@ -220,7 +220,7 @@ test('check group dialog for unencrypted group has appropriate entries', async (
   await expect(dialog.locator('#showqrcode')).not.toBeVisible({ timeout: 1 })
 
   await dialog.getByTestId('view-group-menu').click()
-  // The three-dot menu should not show have an "Edit" entry for unencrypted groups.
+  // The three-dot menu should not have an "Edit" entry for unencrypted groups.
   await expect(page.getByTestId('view-group-edit')).not.toBeVisible({
     timeout: 1,
   })
--- a/packages/e2e-tests/tests/unencrypted-group-tests.spec.ts
+++ b/packages/e2e-tests/tests/unencrypted-group-tests.spec.ts
@@ -215,12 +215,17 @@ test('check group dialog for unencrypted group has appropriate entries', async (
   ).not.toBeVisible({ timeout: 1 })
   await expect(dialog.locator('#addmember')).not.toBeVisible({ timeout: 1 })
   await expect(
-    dialog.getByTestId('view-group-dialog-header-edit')
-  ).not.toBeVisible({ timeout: 1 })
-  await expect(
     dialog.getByRole('button', { name: 'QR Invite Code' })
   ).not.toBeVisible({ timeout: 1 })
   await expect(dialog.locator('#showqrcode')).not.toBeVisible({ timeout: 1 })
+
+  await dialog.getByTestId('view-group-menu').click()
+  // The three-dot menu should not show have an "Edit" entry for unencrypted groups.
+  await expect(page.getByTestId('view-group-edit')).not.toBeVisible({
+    timeout: 1,
+  })
+  await expect(page.getByTestId('encryption-info')).toBeVisible()
+  await page.keyboard.press('Escape')
 
   await page.getByTestId('view-group-dialog-header-close').click()
 })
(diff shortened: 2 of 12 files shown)


r10s (commented):
some comments, no blockers:

- if there is no result, there should be a line as `No results found for "xxx"` (`search_no_result_for_x`)

- there is always the headline "past members", even if there are no results below (not sure if it makes sense to search over past members in general, this may add confusion, and they're questionable in general. but as long as we show them i am fine with including them to search. but it that adds complexity, i would not hesitate to drop)

nicodh:
Ok, I implemented these 2 changes (no results message and hide past members if empty)

packages/e2e-tests/tests/chat-context-menu.spec.ts
```diff
@@ -467,13 +466,16 @@ test.describe('Chat List Context Menu - Single Selection', () => {
...
+    await page.getByRole('menuitem', { name: 'View Profile' }).click()
+    await page.getByTestId('view-group-menu').click()
+    await page.getByTestId('clone-chat').click()
 
-    // Verify clone chat dialog opens
-    await expect(page.getByRole('dialog')).toBeVisible()
-    // The clone dialog should show create group interface
+    // Verify clone chat dialog opens,
     await expect(page.locator('.group-name-input')).toBeVisible()
+    // it should show the create group interface
```
WofWca:
```suggestion
```

Not sure which line this comment is referring to at this point.
But maybe it's just me.

packages/e2e-tests/tests/unencrypted-group-tests.spec.ts
```diff
@@ -214,14 +214,19 @@ test('check group dialog for unencrypted group has appropriate entries', async (
...
-  await expect(
-    dialog.getByTestId('view-group-dialog-header-edit')
-  ).not.toBeVisible({ timeout: 1 })
   await expect(
     dialog.getByRole('button', { name: 'QR Invite Code' })
   ).not.toBeVisible({ timeout: 1 })
   await expect(dialog.locator('#showqrcode')).not.toBeVisible({ timeout: 1 })
 
+  await dialog.getByTestId('view-group-menu').click()
+  // The three-dot menu should not show have an "Edit" entry for unencrypted groups.
```
WofWca:
```suggestion
  // The three-dot menu should not have an "Edit" entry for unencrypted groups.
```

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -540,9 +505,9 @@ export function useChatContextMenu(): {
       archive,
       { type: 'separator' },
       ...(!isMainView ? viewEditMenuItems : []),
-      !isMainView && encryptionInfoItem,
```
WofWca, line 543: This makes it impossible to check encryption info for incoming channels, unlike it is on Android. Also for `Mailinglist`. Refer to the following code:

https://github.com/deltachat/deltachat-desktop/blob/36764980cbc36fdf04ba1e2c921d14b319302d0c/packages/frontend/src/components/screens/MainScreen/MainScreen.tsx#L429-L442

packages/frontend/src/components/dialogs/ViewGroup/index.tsx
```diff
@@ -285,6 +292,42 @@ function ViewGroupInner(
     setGroupImage,
   } = useGroup(accountId, chat)
 
+  const [showMemberFilter, setShowMemberFilter] = useState(false)
+
+  // Open the member filter with Ctrl+F (Cmd+F on macOS)
+  // The global keybinding handler is disabled while a dialog is open
+  useEffect(() => {
+    const onKeyDown = (ev: KeyboardEvent) => {
+      // copied from
```
WofWca:
```suggestion
      // copied from `keybindings.ts`
```

packages/frontend/src/components/dialogs/ViewGroup/index.tsx
```diff
@@ -285,6 +292,42 @@ function ViewGroupInner(
...
+      }
+
+      // to be specific, only react when this
+      // dialog is the topmost modal dialog
+      const dialogEl =
+        groupMemberContactListWrapperRef.current?.closest('dialog')
+      const modals = document.querySelectorAll('dialog:modal')
+      if (!dialogEl || modals[modals.length - 1] !== dialogEl) {
+        return
+      }
```
WofWca, line 344: Why not just attach the listener to the dialog element itself?

packages/frontend/src/components/dialogs/ViewGroup/ViewGroupMenu.tsx
```diff
@@ -0,0 +1,75 @@
...
+      dataTestid: 'view-group-edit',
+    },
+    !showMemberFilter && {
+      label: tx('search'),
+      action: onShowMemberFilter,
+      dataTestid: 'show-member-filter',
+    },
+    {
+      type: 'separator',
+    },
```
WofWca, line 44: There might be 0 items above the separator.
But not a big deal.

packages/frontend/src/components/dialogs/ViewGroup/ViewGroupMenu.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const { openDialog } = useDialog()
+  const tx = useTranslationFunction()
+
+  const menu: (ContextMenuItem | false)[] = [
+    allowEdit && {
+      label: tx('global_menu_edit_desktop'),
+      action: onClickEdit,
+      dataTestid: 'view-group-edit',
+    },
+    !showMemberFilter && {
```
WofWca, line 37:
```suggestion
    {
```

IDK if we need this logic.
But then it would be nice to properly focus the input and not only when it gets mounted.
Not too important.

packages/frontend/src/components/dialogs/ViewGroup/index.tsx
```diff
@@ -285,6 +292,42 @@ function ViewGroupInner(
     setGroupImage,
   } = useGroup(accountId, chat)
 
+  const [showMemberFilter, setShowMemberFilter] = useState(false)
```
WofWca, line 314: Minor: it would be nice to make it impossible to enter invalid state, i.e. when `memberFilter !== '' && showMemberFilter === true`, since we rely on that somewhat.

packages/frontend/src/components/dialogs/ViewGroup/index.tsx
```diff
@@ -414,6 +480,24 @@ function ViewGroupInner(
...
+                setMemberFilter('')
+              }}
+            />
+          </div>
+        )}
+        {filterContacts(groupContacts).length === 0 && memberFilter !== '' && (
+          <div className='group-member-filter-no-result'>
+            {tx('search_no_result_for_x', memberFilter)}
+          </div>
+        )}
```
WofWca, line 521: FYI this is not accessible i.e. search results are not announced, but the same goes for other search occurrences, so maybe this is for another MR.

WofWca (commented):
Works overall and looks good, but removes the "Encryption Info" for channels, which doesn't seem intentional.

packages/frontend/src/components/dialogs/ViewGroup/index.tsx
nicodh: I tried that, but the dialog has no focus so the key handler does not apply.

packages/frontend/src/components/chat/ChatContextMenu.tsx
nicodh: Good catch! I added the encryption info to the main menu just for Channels & Mailinglists

packages/frontend/src/components/dialogs/ViewGroup/index.tsx
WofWca: Ah, it's this `blur()` thing comes back to bite us again...

https://github.com/deltachat/deltachat-desktop/blob/9181dc76087ec86a4e642ea41ccf4b54089c07d7/packages/frontend/src/keybindings.ts#L164-L168

WofWca (approved):
In my reading this doesn't fully address the issue as it's currently written, i.e. the issue doesn't seem to imply that the "Encryption info" item is supposed to be placed differently for incoming and outgoing channels. I leave this for someone else's judgement.

Otherwise looks good. Will spare some suffering to people managing big groups.

r10s (commented):
another thing i have overseen above:

the quick changing of the whole dialog height on typing is single character is unexpected and annoying. it is hard to follow the input line by that.

and the fast flickering of background / dialog it is also an issue when it comes to accessibility, visual sensitivity, epilepsy.

let's not change the dialog height on typing (as eg. in contact search)

nicodh:
> another thing i have overseen above:
> 
> the quick changing of the whole dialog height on typing is single character is unexpected and annoying. it is hard to follow the input line by that.
> 
> and the fast flickering of background / dialog it is also an issue when it comes to accessibility, visual sensitivity, epilepsy.
> 
> let's not change the dialog height on typing (as eg. in contact search)

Then we have 3 options:

1. set the dialog height to a fix height no matter how many members are listed (like contacts list)
2.  set the dialog height to a fix height like 60% if few members 90% if many
3. "remember" the initial dynamic height as a reference and keep that when filtering/searching

Concerning the encryption info inChannels/Mailinglists: I wanted to avoid to add a 3 dot menu for these profiles that just would contain on item (encryption info) - which is anyway a bit useless on mailing lists

r10s:
if not too complex, 3. sounds good and flexible to me. for 1. and 2. that may look less nice for few members, or with short or large description

for the three dot menu with one item only: that'd be totally fine, so if we consider "encryption info" for channels useful, it should be at the same, consistent place

WofWca (approved):
The three-dot menu thing has been addressed.
I didn't re-test it though, E2E tests do it.

nicodh:
So after talking with @r10s I decided to set a fix maximum height when searching and align the dialog to top (like @WofWca suggested)

packages/frontend/src/components/dialogs/ViewGroup/styles.module.scss
```diff
@@ -1,6 +1,11 @@
 $borderRadius: 10px;
 $closeButtonSize: 30px;
 
+.topAlignedDialog {
+  margin-block-start: 25px;
+  margin-block-end: auto;
+}
+
```
WofWca, line 8: Minor: I think it'd be better to have this in the dialog stylesheet: it might come in handy in other cases, i.e. other places where we have dynamically-sizeable dialogs.

WofWca (approved):
Better.
Re-tested and confirmed that the dialog is indeed aligned to the top, i.e. reducing layout shifts.

PR #6457, opened 2026-06-11
fix: don't faint text select color if <2 msgs selected

WofWca:
- **refactor: msg multiselect: `:has()` -> class name**
- **fix: faint text select color if <2 msgs selected**

Before / after:

image image

### DIFF
--- a/packages/frontend/scss/message/_message-list.scss
+++ b/packages/frontend/scss/message/_message-list.scss
@@ -105,6 +105,22 @@
     &::-webkit-scrollbar-track {
       background: transparent;
     }
+
+    // If you Shift+Click some messages then their text
+    // also gets selected, which looks pretty ugly.
+    // However, some people might still want to select text this way,
+    // so let's keep a really faint color for them.
+    //
+    // We apply this to the whole `<ol>` and not `<li>`
+    // because otherwise daymarkers are not affected by this
+    // (because they're not selectable).
+    &.multiselected-one-or-more::selection {
+      background-color: color-mix(
+        in srgb,
+        var(--messageHightlightColor) 5%,
+        transparent
+      );
+    }
   }
 
   ol {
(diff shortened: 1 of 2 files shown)


nicodh:
Hmm, I wonder if there would be a better solution. I'm not sure if a user after an attempt to select a long text inside a message would still continue if he sees the message as a whole got selected on the Shift+Click.

Is there a way to detect if there is already some text (and only text) selected and if so just don't select the whole message? Only if not I would say this workaround is acceptable. :-)

WofWca:
> I'm not sure if a user after an attempt to select a long text inside a message would still continue if he sees the message as a whole got selected on the Shift+Click.

Continue with what?

> Is there a way to detect if there is already some text (and only text) selected and if so just don't select the whole message?

I don't think that there can be a simple solution. It's anyway possible to first regular-click somewhere in text and then Shift+Click, and that will select text. Also one can regular-click on one place and then Shift-Click on the same place which will not change text selection. Also it's possible to first select some text and then Shift+Click on selection start, and that will unselect all text. All these are text selection operations.

One thing we could do is not change selection on Shift + Click if it resulted in only one item getting selected.
But this doesn't contradict the idea of this MR, which is simply to be more conservative in when to apply this faint color hack.

nicodh (approved):
Reviewed by reading & testing

Although there is some downside, this improves the "select text" situation with no doubt

WofWca:
What's the downside that this MR introduces?

nicodh:
> What's the downside that this MR introduces?

What I mentioned in my comment. The fact that visually we have 2 selects at the same time the selected text and the message as a whole. But as I said, this is better then not seeing that the text is selected.

WofWca:
Ah I see. Makes sense. Then maybe my idea would work there.
Let's keep that for a follow-up I suppose.

PR #6458, opened 2026-06-11
feat: show num selected messages

WofWca:
image
image

This is different from how we display it on Delta Chat android and iOS
where we display the number of selected messages
as a replacement to the chat name.
But I think that this is better because such display is more common
and thus more familiar to users on desktop.
Namely this design is inspired by file explorers
on Windows and some Linux OSes, which have a similarly positioned
and styled status bar.

The left border color could used some more work, but the idea is there.

Addresses https://github.com/deltachat/deltachat-desktop/pull/6268#issuecomment-4421685886.

### DIFF
--- a/packages/frontend/scss/message/_message-list.scss
+++ b/packages/frontend/scss/message/_message-list.scss
@@ -206,4 +206,18 @@
       }
     }
   }
+
+  .num-selected-messages-status {
+    position: absolute;
+    bottom: 0;
+    inset-inline-start: 0;
+    border-start-end-radius: 0.25rem;
+    $borderColor: color-mix(in srgb, var(--contextMenuBorder) 50%, transparent);
+    border-block-start: 1px solid $borderColor;
+    border-inline-end: 1px solid $borderColor;
+
+    background: var(--composerBg);
+    padding: 0.5rem 0.625rem;
+    line-height: 1;
+  }
 }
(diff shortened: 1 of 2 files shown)


nicodh:
Hmm, this is a bit hidden - such a small field at the very left bottom corner. I didn't find it on your screenshots on first sight.

What about a position top 30px, inline-start 44% and some box shadow to make sure there is contrast to the background?

just a raw example:
image

WofWca:
I think what you describe is not as big of a problem as it might seem, because the element is dynamic: it appears and disappears as you "enter" and "exit" the multiselect "mode", and as you change selection. On top of the fact that the element itself is not essential.

What you propose is also fine by me, but I prefer my way.
I can change that aspect if this means that then the MR will be good to merge.

nicodh (approved):
Reviewed by reading & testing

I leave it to you to adapt the styling and location of the counter in this PR or in a followup

WofWca:
Alright, I think that having an ugly counter is better than not having one at all, so let's merge now.
I'll work on the new implementation (as we previously discussed, replacing either the composer or the chat title) later.
We can also consider omitting the "added message multiselect" from the CHANGELOG. Not just for the counter but overall for a more "phased rollout" if one can call it that.

r10s:
i talked with @nicodh some days ago - i was fine with merging current state.

for iteration: the counter should not add cluttering, and be very visible. at best we have less, the user has to process visually. so the overlay suggestion is not the best option.

instead

1. hide the message input bar and show actions and counter there.

2. or show the counter instead of avatar/title/subtitle.

both options are widely used in other apps, and also on delta mobile. 

the first option seems not so far from
what we have currently - bigger and at position of input bar would di

PR #6462, opened 2026-06-11
fix: QR code scanner in cut on the right when using a vertical camera

### LINKED ISSUE #6461: QR code scanner in cut on the right when using a vertical camera
The issue is really common on mobile linux distributions such as Mobian , postmarketOS or FuriOS because these devices have a vertical camera instead of an horizontal camera. (the device that I've used for the test is FuriPhone FLX1s)

Image

As you can check on the screenshot... the camera is cut on the right when I try to scan a qr code on the "Add a Second Device" or "SCAN QR CODE" sections

The problem is the grey bar on the right

I've a solution on a Pull Request arriving soon :)

Enrico

account1009:
Closes https://github.com/deltachat/deltachat-desktop/issues/6461.

This PR makes the qr scanner responsive and valid for both horizontal cameras and vertical cameras.
The previous code was working only with horizontal cameras while vertical cameras were cut as you can check on the screenshot of this issue: https://github.com/deltachat/deltachat-desktop/issues/6461

I've tested the solution on a device with horizontal camera (M2 Ashai Linux) and a device with vertical camera (FuriPhone FLX1s Linux Phone) and everything is fine :)

Enrico

### DIFF
--- a/packages/frontend/src/components/QrReader/styles.module.scss
+++ b/packages/frontend/src/components/QrReader/styles.module.scss
@@ -6,7 +6,9 @@
 
 .qrReaderVideo {
   display: none;
-  height: 375px;
+  height: 100%;
+  width: 100%;
+  object-fit: cover;
   left: 0;
   position: absolute;
   right: 0;


packages/frontend/src/components/QrReader/styles.module.scss
```diff
@@ -6,7 +6,9 @@
 
 .qrReaderVideo {
   display: none;
-  height: 375px;
+  height: 100%;
+  width: 100%;
+  object-fit: cover;
```
WofWca, line 11:
```suggestion
  object-fit: contain;
```

WofWca (commented):
This works, but IDK about `cover`. `cover` doesn't show the full video, some parts are left out. I think it's better to just center the view (which is what my suggestion does).

packages/frontend/src/components/QrReader/styles.module.scss
account1009: hi @WofWca I'm not convinced about your suggestion. I think that is a downgrade
In both vertical and horizontal camera cases it generates grey bars (left-right or top-bottom).

The problem is also that for vertical case the grey bars generated with your solution are ever covering the square that suggests you a safe zone where you have to put the qr.

So the message would be contradictory: the safe zone is inside the suggested square but at the same time inside the square there is a grey unusable area

So I still think that the `object-fit: cover;` is the best way to solve the problem

WofWca: Well IDK, the "safe zone" I believe is purely decorative. So I'd say we should prefer to show what's actually happening.
But OTOH QRs are anyways square, so it's probably OK to cut out the edges of the camera's video. Then it becomes only a matter of where the QR is: at the edge of the video or at the center.

WofWca (approved):

nicodh (approved):
Reviewed by reading

PR #6463, opened 2026-06-12
fix: message list call btns: respect `tabindex`

WofWca:
Otherwise if you have many calls in the message list
then the message list becomes a keyboard trap.


nicodh (approved):
Reviewed by reading

PR #6466, opened 2026-06-12
fix: don't send whitespace-only message

WofWca:
We have the same check on Android:
https://github.com/deltachat/deltachat-android/blob/d5f627be501dc9ad1d314b62b36708f8afa2242b/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java#L1653-L1668
(though Android also shows a toast, which I think is too much).

When it comes to limiting what users can do,
there needs to be a good reason for it.
It's perhaps funny to confuse your friends
with a goofy-looking empty message, but it's only funny once.
A more common, in my eyes, use case is when the user is just fiddling
with the message input not knowing what to write and types in
just spaces or just newlines and presses "send"
expecting that this will not send the message,
like it is on basically all other messengers,
and on other Delta Chat clients.

### DIFF
--- a/packages/e2e-tests/tests/composer.spec.ts
+++ b/packages/e2e-tests/tests/composer.spec.ts
@@ -103,6 +103,44 @@ async function typeText(text: string) {
   await expect(textarea).toHaveText(text)
 }
 
+test("doesn't send empty message", async () => {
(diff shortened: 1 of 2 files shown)


nicodh (approved):
Reviewed by reading

e2e failure seems not related

PR #6473, opened 2026-06-15
fix: preserve line breaks in channel dialog descr, don't center text

### LINKED ISSUE #6470: Channel description on subscriber side without line break
I noticed that a line break in the channel description on the subscriber page is only displayed as a space. Probably this is because the text is always centered there?

macOS 26.5.1
DC-Desktop 2.52.0 (Electron)

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/6470.

We used to also center group descriptions, but we stopped doing that in
https://github.com/deltachat/deltachat-desktop/commit/12305f4b45d3926d7678f0185c5ffaea38f04667
(https://github.com/deltachat/deltachat-desktop/pull/6168).

image

### DIFF
--- a/packages/frontend/src/components/ProfileInfoHeader/styles.module.scss
+++ b/packages/frontend/src/components/ProfileInfoHeader/styles.module.scss
@@ -21,6 +21,12 @@
   user-select: text;
 }
 
+.description {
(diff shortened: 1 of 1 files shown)


nicodh (approved):
Reviewed by reading

test failure is not related

PR #6475, opened 2026-06-15
fix: misc fixes for jsonrpc-client types, message rendering, and chat scheduler

PanzerPeter:
- **fix: use exported T namespace instead of deep jsonrpc-client type imports**
- **perf: avoid recreating Intl.Segmenter on every message render**
- **fix: run chat scheduler queued effects in FIFO order**

### DIFF
--- a/packages/frontend/src/components/message/MessageBody.tsx
+++ b/packages/frontend/src/components/message/MessageBody.tsx
@@ -11,6 +11,17 @@ const MAX_BYTE_SIZE_OF_EMOJI = 10 /* 10 is maybe already too generous? */
 const MAX_STRING_LENGTH_FOR_BIG_EMOJI =
   MAX_BIG_EMOJI_COUNT * MAX_BYTE_SIZE_OF_EMOJI
 
+// Created once at module load instead of on every message render:
+// constructing an `Intl.Segmenter` is relatively expensive, and
+// `countEmojisIfOnlyContainsEmoji` runs for every (short) message.
+// `granularity: 'grapheme'` is locale-independent, so the locale is irrelevant.
+// See `avatarInitial.ts` for the same pattern.
+const graphemeSegmenter = new Intl.Segmenter(undefined, {
+  // Split the input into segments at grapheme cluster
+  // (user-perceived character) boundaries
+  granularity: 'grapheme',
+})
+
 /**
  * if a message contains only emojis and is not too long,
  * we display the emojis bigger as usual inline emojis
@@ -50,12 +61,7 @@ function countEmojisIfOnlyContainsEmoji(str: string): number | null {
     if (firstToken.t === 'EMOJI') {
       // use Intl.Segmenter to split grapheme clusters (emoji + skin tone modifier etc)
       // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter/Segmenter
-      const segmenter = new (Intl as any).Segmenter('en', {
-        // Split the input into segments at grapheme cluster
-        // (user-perceived character) boundaries
-        granularity: 'grapheme',
-      })
-      return [...segmenter.segment(firstToken.v)].length
+      return [...graphemeSegmenter.segment(firstToken.v)].length
     }
   }
   return null
--- a/packages/frontend/src/components/message/MessageBody.tsx
+++ b/packages/frontend/src/components/message/MessageBody.tsx
@@ -11,10 +11,6 @@ const MAX_BYTE_SIZE_OF_EMOJI = 10 /* 10 is maybe already too generous? */
 const MAX_STRING_LENGTH_FOR_BIG_EMOJI =
   MAX_BIG_EMOJI_COUNT * MAX_BYTE_SIZE_OF_EMOJI
 
-// Created once at module load instead of on every message render:
-// constructing an `Intl.Segmenter` is relatively expensive, and
-// `countEmojisIfOnlyContainsEmoji` runs for every (short) message.
-// `granularity: 'grapheme'` is locale-independent, so the locale is irrelevant.
 // See `avatarInitial.ts` for the same pattern.
 const graphemeSegmenter = new Intl.Segmenter(undefined, {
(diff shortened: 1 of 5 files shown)


packages/frontend/tsconfig.json
```diff
@@ -9,7 +10,7 @@
     "jsx": "react",
     "allowJs": false,
     "esModuleInterop": true,
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
WofWca: What's the point of this?

packages/frontend/src/stores/chat/chat_scheduler.ts
```diff
@@ -80,7 +80,10 @@ export class ChatStoreScheduler {
         return
       }
 
-      const effect = this.effectQueue.pop()
+      // Use `shift()` (not `pop()`) so effects run in FIFO order,
+      // i.e. the order they were queued in, as documented for
+      // `queuedEffect`/`lockedQueuedEffect`. `pop()` would run them LIFO.
```
WofWca:
```suggestion
```

This should be a commit message IMO.

packages/frontend/src/stores/chat/chat_scheduler.ts
```diff
@@ -80,7 +80,10 @@ export class ChatStoreScheduler {
...
+      const effect = this.effectQueue.shift()
```
WofWca, line 84: Also FTR there was another commit like this by a different author.
https://github.com/lambadalambda/deltachat-desktop/commit/0c04a530d07a31fd791534b9ccac2aabde53d53a

They claim that it fixes some crash.

packages/frontend/src/components/message/MessageBody.tsx
```diff
@@ -11,6 +11,17 @@ const MAX_BYTE_SIZE_OF_EMOJI = 10 /* 10 is maybe already too generous? */
 const MAX_STRING_LENGTH_FOR_BIG_EMOJI =
   MAX_BIG_EMOJI_COUNT * MAX_BYTE_SIZE_OF_EMOJI
 
+// Created once at module load instead of on every message render:
+// constructing an `Intl.Segmenter` is relatively expensive, and
+// `countEmojisIfOnlyContainsEmoji` runs for every (short) message.
+// `granularity: 'grapheme'` is locale-independent, so the locale is irrelevant.
```
WofWca: This is also rather fitting for a commit message and not a comment.

```suggestion
```

packages/frontend/src/components/message/MessageBody.tsx
```diff
@@ -11,6 +11,17 @@ const MAX_BYTE_SIZE_OF_EMOJI = 10 /* 10 is maybe already too generous? */
 const MAX_STRING_LENGTH_FOR_BIG_EMOJI =
   MAX_BIG_EMOJI_COUNT * MAX_BYTE_SIZE_OF_EMOJI
 
+// Created once at module load instead of on every message render:
+// constructing an `Intl.Segmenter` is relatively expensive, and
```
WofWca: Also just curious, do you know how expensive it is and what makes it expensive?

WofWca (commented):
Hi and welcome, and thanks for the MR!

Overall looks good.

The "types" commit I think is not a `fix` but a `refactor`.

packages/frontend/src/components/message/MessageBody.tsx
PanzerPeter: hey, sorry for the late response, I just noticed it by now. I answer the other comments shortly aswell
instantiating Intl objects forces the JS engine to load heavy locale data and internal Unicode lookup tables. Doing that on every single message render would tank scrolling performance, so caching it at the module level avoids that hit.

packages/frontend/src/stores/chat/chat_scheduler.ts
PanzerPeter: yeah I agree with this one too, I'll do better in the future

packages/frontend/tsconfig.json
PanzerPeter: "bundler" is TypeScript's modern resolution strategy for tools like Vite or Webpack. Unlike the old "node" setting, it properly understands things like the exports field in modern package.json files and matches how our actual build tools resolve imports as I know.

packages/frontend/src/stores/chat/chat_scheduler.ts
PanzerPeter: Good catch I edited it in my next commit

packages/frontend/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "compilerOptions": {
     "outDir": "../../tsc-dist/renderer",
+    "rootDir": "..",
```
WofWca: How about `rootDir`? Why is `"rootDirs": [".", "../shared"],` but this is a dir up?

There is another approach I take in https://github.com/deltachat/deltachat-desktop/pull/6487. Maybe it's better to just drop this change from the MR, unless there is a good reason behind it.

PanzerPeter: Ah, good catch! Yeah, no real reason for that being there. I've removed the tsconfig changes (moduleResolution / rootDir) since they're unrelated to the type-import fix, and check:types is still passing fine. Agreed that we should tackle the shared-import stuff properly in #6487. Thanks!

WofWca (approved):
Thanks! Looks good now.

Are you OK with a squash-merge or would you like to clean up commit history?

PanzerPeter:
sure, I'm fine with a squash-merge. But if you want I can clean up the history

WofWca:
OK, let's squash then.

PanzerPeter:
Hey, just a quick question, is everything alright or I anything else I have to do on my end? as I know  the E2E failures are pre-existing and unrelated to my changes

WofWca:
Hey
We're waiting for someone else's review.

nicodh (approved):
Reviewed by reading

nicodh:
Thanks and sorry for the delay in the review!

PR #6480, opened 2026-06-17
fix: remove is chatmail

### LINKED ISSUE #5737: bots with mailto: link in html response not working
some bots like www gateway bots (bots to fetch preview/snapshots of websites and send them to the users as html messages) use mailto links in html messages so commands inside the html can be forwarded back to the bot chat, when such a mailto is clicked in android, the key contact for the given address is used as the chat where to put the draft to, that already works well, desktop is instead rejecting to use "query contact id by addr" api and showing dialog:

Image

### LINKED ISSUE #6459: remove all usages of "is_chatmail"
"is_chatmail" was needed mainly for decision whether  to show "New Email" or "Add Contact Manually" and corresponding dialogs.

this should be replaced by checking "force_encryption"

all other usages should not be needed as well.

this also fixes the issue that "New Email" is availabe when onboarding on classic relays (as "force_encryption" is off by default) https://github.com/chatmail/core/issues/8327

cmp https://github.com/deltachat/deltachat-ios/pull/3142https://github.com/deltachat/deltachat-android/pull/4480

nicodh:
resolves #6459 , resolves #5737

### DIFF
--- a/packages/frontend/src/components/chat/ChatList.tsx
+++ b/packages/frontend/src/components/chat/ChatList.tsx
@@ -604,7 +604,7 @@ function ContactAndMessageSearchResults({
   const tabindexWrapperElementMessages = useRef<HTMLDivElement>(null)
 
   const settingsStore = useSettingsStore()[0]
-  const isChatmail = settingsStore?.settings.is_chatmail === '1'
+  const forceEncryption = settingsStore?.settings.force_encryption !== '0'
 
   const addContactOnClick = async () => {
     if (!queryStrIsValidEmail || !queryStr) return
@@ -697,7 +697,7 @@ function ContactAndMessageSearchResults({
           >
             {ChatListItemRowContact}
           </ChatListPart>
-          {!isChatmail &&
+          {!forceEncryption &&
             contactIds.length === 0 &&
             chatListSearchResultsIsEmpty &&
             queryStrIsValidEmail && (
--- a/packages/frontend/src/components/dialogs/EditAccountAndPasswordDialog.tsx
+++ b/packages/frontend/src/components/dialogs/EditAccountAndPasswordDialog.tsx
@@ -36,13 +36,15 @@ export default function EditAccountAndPasswordDialog({
   const tx = useTranslationFunction()
 
   const settingsStore = useSettingsStore()[0]
-  const isChatmail = settingsStore?.settings.is_chatmail === '1'
+  const forceEncryption = settingsStore?.settings.force_encryption !== '0'
 
   return (
     <Dialog canOutsideClickClose={false} onClose={onClose}>
       <DialogHeader
         title={
-          isChatmail ? tx('edit_transport') : tx('manual_account_setup_option')
+          forceEncryption
+            ? tx('edit_transport')
+            : tx('manual_account_setup_option')
         }
       />
       <EditAccountInner {...{ onClose, addr }} />
--- a/packages/frontend/src/components/dialogs/EditAccountAndPasswordDialog.tsx
+++ b/packages/frontend/src/components/dialogs/EditAccountAndPasswordDialog.tsx
@@ -42,6 +42,8 @@ export default function EditAccountAndPasswordDialog({
     <Dialog canOutsideClickClose={false} onClose={onClose}>
       <DialogHeader
         title={
+          // forceEncryption is false if account is not
+          // configured yet or the setting is set to false
           forceEncryption
             ? tx('edit_transport')
             : tx('manual_account_setup_option')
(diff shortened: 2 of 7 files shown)


packages/frontend/src/hooks/useProcessQr.ts
```diff
@@ -109,7 +109,7 @@ export default function useProcessQR() {
   const { secureJoin } = useSecureJoin()
 
   const settingsStore = useSettingsStore()[0]
-  const isChatmail = settingsStore?.settings.is_chatmail === '1'
+  const forceEncryption = settingsStore?.settings.force_encryption !== '0'
```
WofWca:
```suggestion
```

The linter is right.

packages/frontend/src/components/chat/ChatList.tsx
```diff
@@ -604,7 +604,7 @@ function ContactAndMessageSearchResults({
   const tabindexWrapperElementMessages = useRef<HTMLDivElement>(null)
 
   const settingsStore = useSettingsStore()[0]
-  const isChatmail = settingsStore?.settings.is_chatmail === '1'
+  const forceEncryption = settingsStore?.settings.force_encryption !== '0'
```
WofWca, line 607:
```suggestion
  const forceEncryption = settingsStore?.settings.force_encryption === '1'
```

See https://github.com/deltachat/deltachat-desktop/pull/6410#discussion_r3305095058.

Although here this only affects whether the "add contact" option is shown, I think my suggestion is still better, for consistency and future-proofing.
Same for other instances
Although not critical.

packages/frontend/src/components/dialogs/EditAccountAndPasswordDialog.tsx
```diff
@@ -36,13 +36,15 @@ export default function EditAccountAndPasswordDialog({
...
+  const forceEncryption = settingsStore?.settings.force_encryption !== '0'
 
   return (
     <Dialog canOutsideClickClose={false} onClose={onClose}>
       <DialogHeader
         title={
-          isChatmail ? tx('edit_transport') : tx('manual_account_setup_option')
+          forceEncryption
+            ? tx('edit_transport')
+            : tx('manual_account_setup_option')
```
WofWca, line 49: I don't see the logic behind this condition, but maybe it's just me.
I think the original idea was to only show the "manual setup" string on the initial setup screen, and not afterwards in the "relays" dialog. https://github.com/deltachat/deltachat-desktop/issues/5687.

WofWca (commented):
This is not a refactor.

I confirmed that this removes all references to `is_chatmail`.

I confirmed that clicking a mailto link opens the existing key-contact, i.e. this basically reverts https://github.com/deltachat/deltachat-desktop/issues/5620.

But, based on my understanding of the various recent `mailto:` MRs and issues, such as https://github.com/deltachat/deltachat-android/pull/4482, it should not be possible to create a plain email chat when `force_encryption == 1`. But it is. You decide if this is a topic for a separate MR.

packages/frontend/src/components/dialogs/EditAccountAndPasswordDialog.tsx
nicodh: I just changed the settings key here so the logic was not part of this PR

nicodh:
> But, based on my understanding of the various recent `mailto:` MRs and issues, such as [deltachat/deltachat-android#4482](https://github.com/deltachat/deltachat-android/pull/4482), it should not be possible to create a plain email chat when `force_encryption != 1`. But it is. You decide if this is a topic for a separate MR.

There was some for and back so now I agreed with r10s to always use the InvalidUnencryptedMailDialog dialog if force_encrypted is set

packages/frontend/src/hooks/chat/useCreateChatByEmail.ts
```diff
@@ -24,19 +26,23 @@ export type CreateChatByEmail = (
 export default function useCreateChatByEmail(): CreateChatByEmail {
   const tx = useTranslationFunction()
   const { openDialog } = useDialog()
+  const settingsStore = useSettingsStore()[0]
+  const isChatmail = settingsStore?.settings.force_encryption === '1'
```
WofWca:
```suggestion
  const forceEncryption = settingsStore?.settings.force_encryption === '1'
```

Is this intentional?

WofWca (approved):
I did not participate in the discussion, but the code looks alright.

packages/frontend/src/hooks/chat/useCreateChatByEmail.ts
nicodh: No, I changed it

packages/frontend/src/components/dialogs/EditAccountAndPasswordDialog.tsx
nicodh: Looking again at it: the headline should show "Use classic email as relay" if a user selected that option in the onboarding process. When editing we use forceEncryption to decide the headline. I agree that this could be amiguous: if a user wants to  create a transport for an existing chatmail account or if uses a classic email but sets forceEncryption to true the headline could be misleading, but maybe these rare case in a adavanced user setting are acceptyble

PR #6481, opened 2026-06-17
fix: add bottom margin

### LINKED ISSUE #6471: Log screen: No space between text field and buttons
I noticed that there's no space between the "Copy to Clipboard" and "Share" buttons in the log screen. 
Of course, this isn't a bug, but it looks kind of "wrong"?

macOS 26.5.1
DC-Desktop 2.52.0 (Electron)

Image

### LINKED ISSUE #6472: Profile transfer screen: Not enough space between transfer bar and bottom of the window
I've noticed that the space between the transfer bar and the bottom of the window in the profile transfer screen is very small. This makes the window look like it's cut off at the bottom. Of course, this isn't a bug, but I think it looks "wrong"?

macOS 26.5.1
DC-Desktop 2.52.0 (Electron)

Image

nicodh:
resolves #6472 , resolves #6471

### DIFF
--- a/packages/frontend/src/components/dialogs/Log/styles.module.scss
+++ b/packages/frontend/src/components/dialogs/Log/styles.module.scss
@@ -6,6 +6,7 @@
   display: flex;
   flex-direction: column;
   width: 100%;
+  margin-bottom: 20px;
 
   textarea {
     font-family: 'monospace';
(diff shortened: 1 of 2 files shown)


WofWca (approved):
I think that the referenced issues are indicative of a bigger problem going on with our dialog layout (e.g. the following "Delete chats?" dialog doesn't look great either):

Image

, but if we say that it's enough to only address those specific cases to consider the issues closed then it does indeed look like this solves them.

PR #6486, opened 2026-06-17
chore: upgrade ESLint, fix new errors

WofWca:
Also needed to bump the `target` in `tsconfig.json`,
because otherwise TypeScript complains that `Error` constructor
has only one argument, so I can't fix
the `preserve-caught-error` warning.

This also resolves a couple of `pnpm audit` warnings.

I only ran `pnpm check:lint` to verify that things work. I didn't go through the whole migration guide. https://eslint.org/docs/latest/use/migrate-to-10.0.0

TODO:
- [ ] Merge the base.
- [ ] Hmmm https://github.com/jsx-eslint/eslint-plugin-react/issues/3977 doesn't support v10, but things seem to work fine? IDK.


WofWca:
OK, if a dependency doesn't support it then IDK if we want to merge.

PR #6489, opened 2026-06-17
chore: remove `target` from most `tsconfig.json`s

WofWca:
Since TypeScript v6, it defaults to current-year ES version:
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html#simple-default-changes

This should improve performance a little,
thanks to not polyfilling stuff.

I only kept the value for target-electron (which is run in Node).
The rest basically always runs in close-to-latest Chromium.
The only target we actually support in production is Electron,
and we keep it up to date somewhat, so yep.

I smoke-tested this.

TODO:
- [x] Merge the base

### DIFF
--- a/packages/e2e-tests/tsconfig.json
+++ b/packages/e2e-tests/tsconfig.json
@@ -3,7 +3,6 @@
     "noEmit": true,
     "sourceMap": true,
     "module": "nodenext",
-    "target": "ES2022",
     "allowJs": false,
     "esModuleInterop": true
   },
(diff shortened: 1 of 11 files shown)


nicodh (approved):
Reviewed by reading

This might cause problems in tauri since we can't know what browser version is running. But on the other hand tauri is experimental so we don't need to care...

WofWca:
> Tauri

IMO the idea with Tauri is rather to expect the WebView version to be _ahead_ of what we ourselves can package.

But yep, good point. It's still possible to have an old version.

PR #6490, opened 2026-06-18
perf: JSON-RPC IPC: `.invoke` -> `.send`

WofWca:
[`invoke` docs](https://www.electronjs.org/docs/latest/api/ipc-renderer#ipcrendererinvokechannel-args)
state:
> If you do not need a response to the message,
> consider using ipcRenderer.send.

I haven't performed measurements, but this should be better
at least in theory.


nicodh (approved):
Reviewed by reading

Not sure if there is a measurable impact though, but also no harm/risk

PR #6491, opened 2026-06-19
fix: avoid horizontal scrollbar

### LINKED ISSUE #6483: Saved messages have horizontal scroll bar
If there are saved messages in the "Saved messages" chat there is a horizontal scroll bar in the message list. At least if the messages have enough text to get the full message width and the window width is below 900px.

Probably caused by the extended shortcut menu which includes the "original message" button.

Image

nicodh:
resolves #6483  

Mainly 2 changes: 
1. reserve space for the "show in chat" button for saved messages
2. add overflow-wrap rule for .download and info messages

Long author name in an info message with missing overflow-wrap

image

Note: the shortcut menu in "Saved Messages" needs a fix width of 110px - together with avatar and margins that means the width of the message could shrink down to 146px width before the small screen brakpoint hides the chatlist. Maybe acceptable since it only applies to a rare use case (only saved messages and screen width > 720 & < 900. To avoid that we would have to adapt the break points or hide the shortcut menu buttons for reactions and context menu.

Window width 721px
image

### DIFF
--- a/packages/frontend/scss/dialogs/_qr-code.scss
+++ b/packages/frontend/scss/dialogs/_qr-code.scss
@@ -55,7 +55,7 @@
 
 .copy-content-preview {
   background-color: var(--buttonSecondaryBackground);
-  word-wrap: break-word;
+  overflow-wrap: break-word;
   word-break: break-all;
   padding: 10px;
 }
--- a/packages/frontend/scss/message/_message.scss
+++ b/packages/frontend/scss/message/_message.scss
@@ -122,8 +122,6 @@ $info-text-max-width: 400px;
         unicode-bidi: plaintext;
 
         overflow-wrap: break-word;
-        word-wrap: break-word;
-        word-break: break-word;
         white-space: pre-wrap;
         cursor: default;
 
@@ -215,6 +213,8 @@ $info-text-max-width: 400px;
   }
 
   .download {
+    overflow-wrap: break-word;
+
     span.failed {
       color: var(--colorDanger);
       font-weight: bold;
@@ -479,6 +479,7 @@ $info-text-max-width: 400px;
     opacity: 0.9;
     color: var(--infoMessageBubbleText);
     outline-offset: 4px;
+    overflow-wrap: break-word;
 
     .status-icon.sending {
       background-color: var(--infoMessageBubbleText);
--- a/packages/frontend/scss/message/_message.scss
+++ b/packages/frontend/scss/message/_message.scss
@@ -24,7 +24,7 @@ $info-text-max-width: 400px;
   // Messages in the "Saved Messages" might
   // have an extra "show in chat" button
   &.has-original-msg-button {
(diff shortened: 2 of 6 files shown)


packages/frontend/scss/dialogs/_qr-code.scss
```diff
@@ -55,7 +55,7 @@
 
 .copy-content-preview {
   background-color: var(--buttonSecondaryBackground);
-  word-wrap: break-word;
+  overflow-wrap: break-word;
```
WofWca, line 58: This is only a refactor?

packages/frontend/scss/message/_message.scss
```diff
@@ -206,6 +213,8 @@ $info-text-max-width: 400px;
   }
 
   .download {
+    overflow-wrap: break-word;
```
WofWca, line 216: This fixes this particular bug, but it would be better to fix the underlying issue. Something in the direction of adding `.text` class name to the element, or DRYing the code further.
I'm didn't check but I suspect that before we had pre-messages, the message text was simply something like "big message - click to download". Nowadays the whole text is visible, and it's only the attachment that needs to be downloaded.

https://github.com/deltachat/deltachat-desktop/blob/7a6b62a2410bd43296c987a795cdf4519a2f0dfe/packages/frontend/src/components/message/Message.tsx#L920-L945

Maybe for a separate MR.

packages/frontend/src/components/ShortcutMenu/styles.module.scss
```diff
@@ -4,10 +4,10 @@ $transitionDuration: 50ms;
 .shortcutMenu {
   align-items: center;
   display: flex;
-  flex: 0 0 $shortcutMenuWidth;
+  flex: 0 0 var(--shortcut-menu-width, #{$shortcutMenuWidth});
```
WofWca, line 7:
```suggestion
  flex: 0 0 var(--shortcut-menu-width, #{$defaultShortcutMenuWidth});
```

WofWca (approved):
Should be good for the majority of cases, and not too bad for the mentioned case with really narrow message bubbles, so good overall.

We need to consider setting the width limit not on the bubble-and-buttons container but on the message bubble. Changing the breakpoint is not needed then.

https://github.com/deltachat/deltachat-desktop/blob/78f45d02209f7a53ff29f806f9d47bc5c8399123/packages/frontend/scss/message/_message.scss#L22

Also maybe removing the "react" button in saved messages would indeed be good, since firstly reacting to saved messages is not very useful (unless we're gonna implement Telegram-style "saved message classification"), secondly is still possible through the context menu.

packages/frontend/scss/dialogs/_qr-code.scss
nicodh: yes, word-wrap is legacy alias for overflow-wrap https://www.w3.org/TR/css-text-3/#overflow-wrap-property

PR #6492, opened 2026-06-19
perf: build: enable tree-shaking for main process

WofWca:
It's not clear why it was disabled,
but things seem to work fine when I try it.

### DIFF
--- a/packages/target-electron/bin/build.js
+++ b/packages/target-electron/bin/build.js
@@ -22,7 +22,6 @@ await build({
     'isomorphic-ws',
   ],
   entryPoints: ['src/index.ts'],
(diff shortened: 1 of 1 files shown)


nicodh:
I'm also not sure why it was disabled but there might be some risk that our builds. The result is just 2.9 kB smaller then without tree shaking so I'm not sure if we need it.

Was there a specific reason to re-enable it besides "no reason to disable it"

WofWca:
> Was there a specific reason to re-enable it besides "no reason to disable it"

No.

But I'm afraid that the alternative is to keep it disabled until the end of the days of this code.

PR #6494, opened 2026-06-19
test: split up a "Forward Message" test

WofWca:
So that it's easier to see which exact part failed.
This is helpful if the test failed due to an app picker network error.


WofWca:
The current test failure is a good example:

```
  2 failed
    [chatmail] › tests/basic-tests.spec.ts:364:1 › add app from picker to chat ─────────────────────
    [chatmail] › tests/forward-message.spec.ts:257:1 › forward message with .xdc file attachment ───
  3 flaky
```

nicodh (approved):
Reviewed by reading

PR #6496, opened 2026-06-19
refactor: factor out `ChatView` from `MainScreen`

WofWca:
- **refactor: rename `MessageListView` -> `ChatView`**
- **refactor: factor out `ChatView` from `MainScreen`**

This is in preparation for moving the multiselect counter to the chat view header.
But is also nice overall IMO.

`MainScreen` got quite bloated for its name,
so it's time to move things like "chat section subheader"
to a more specific file.

This is not expected to affect behavior.
This commit does not change HTML layout or JS logic.
It's mostly about moving things around.

The most non-trivial change here is perhaps that the `<section>`
is now referred to from 2 SCSS files, by two class names:
`chatAndNavbar` and `chatView`.
I also updated the comments where `chatAndNavbar` was referenced.

I also skimmed through the SCSS files, to check for things like
outdated comments referencing something that is now in the other file.

This will also allow us to more easily put the whole chat view
behind a React `key`.


packages/frontend/src/components/screens/MainScreen/styles.module.scss
nicodh: outdated...

nicodh (approved):
Reviewed by reading and testing!

Useful refactoring!

Needs rebase

packages/frontend/src/components/screens/MainScreen/styles.module.scss
WofWca: Oops... I though I was careful.

```suggestion
  // See below, `.chatView`'s flex-grow.
```

PR #6508, opened 2026-06-25
chore: upgrade electron to 42.5

nicodh:
see https://www.electronjs.org/docs/latest/breaking-changes#planned-breaking-api-changes-420

The only change that somehow affects us is "macOS notifications now use UNNotification API" which means you have no notifications on unsigned packages which will only be relevant for developers

### DIFF
--- a/packages/target-electron/package.json
+++ b/packages/target-electron/package.json
@@ -81,7 +81,7 @@
     "application-config": "^3.0.0",
     "chai": "^5.1.1",
     "debounce": "^1.2.0",
-    "electron": "^41.7.1",
+    "electron": "^42.5.0",
     "electron-builder": "^26.7.0",
     "esbuild": "^0.25.0",
     "mocha": "^10.7.0",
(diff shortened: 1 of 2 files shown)


WofWca (approved):
I went through https://www.electronjs.org/blog/electron-42-0#breaking-changes.

- `clearStorageData` is used, but not `quotas`, so all good.
- `ELECTRON_SKIP_BINARY_DOWNLOAD` unused
- `createFromNamedImage` unused
- Offscreen rendering is unused
- `UNNotification` - mentioned in the MR's description.

FYI this should also fix the "Electron failed to install" issue for devs.

WofWca:
Regarding FreeBSD: I heard that they're also moving to v42, so it's OK for us to upgrade, and to start using Electron v42's new features.

WofWca (approved):
The lock file looks OK.

WofWca (approved):

PR #6474, opened 2026-06-15
fix: unify profile description code

WofWca:
- **refactor: unify profile description code**
- **fix: unify profile description code more**

As I suggested in https://github.com/deltachat/deltachat-desktop/pull/6283#pullrequestreview-4156563131.

Fixes https://github.com/deltachat/deltachat-desktop/issues/6482.

This is AI-generated, see commit messages.

TODO:
- [x] Merge the base.

PR #6509, opened 2026-06-25
test: fix group description test

nicodh:
The class group-profile-description was removed in #6474 but the tests were not adapted


WofWca (approved):
Right. I didn't pay enough attention. Thanks!

PR #6510, opened 2026-06-25
feat: add min height for dialogs

nicodh:
- freeze the min height to avoid too frequent resizing
- release it again when resizing is wanted

The idea is to freeze the dynamic dialog height when resizing is unwanted by setting the new prop minHeight in dialogs 

See https://github.com/deltachat/deltachat-desktop/pull/6456#issuecomment-4793117466

This could also be used for the contact list, where search/filtering is also happening

### DIFF
--- a/packages/e2e-tests/tests/chat-context-menu.spec.ts
+++ b/packages/e2e-tests/tests/chat-context-menu.spec.ts
@@ -376,7 +376,6 @@ test.describe('Chat List Context Menu - Single Selection', () => {
       'Mute Notifications',
       'Archive Chat',
       'View Profile',
-      'Encryption Info',
       'Block Contact',
       'Delete',
     ])
@@ -398,8 +397,6 @@ test.describe('Chat List Context Menu - Single Selection', () => {
       'Mute Notifications',
       'Archive Chat',
       'View Profile',
-      'Encryption Info',
-      'Clone Chat',
       'Leave Group',
     ])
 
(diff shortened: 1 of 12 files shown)


WofWca (commented):
Introducing a bunch of JS for this is really not great IMO. Also bug-prone.

Have you considered solving the problem with just CSS? I see two options:
- add a prop to make the dialog aligned to the top instead of the center
- add a prop to make the dialog grow to its max size, regardless of the size of the contents.

WofWca:
Ah. I looked at the referenced comment again, it's been discussed there.

WofWca:
But my "alignment" idea hasn't been brought up. Maybe that would be good enough?

nicodh:
> But my "alignment" idea hasn't been brought up. Maybe that would be good enough?

How would that solve the problem? What we want to achieve is:
1. for small groups the dialog should not be longer than needed
2. otherwise the dialog should be 90vh high to use all space
3. the dialog should adapt when adding/removing members
4. the dialog should not adapt when filtering/searching

If you consider the current approach "too complex" - an alternative would be set the height to 90vh via css always when the search is active. That would be a minimal change and still achieve all points by the "cost" of showing a longer than needed  dialog in search mode for skmall groups

WofWca:
Checks all marks except the last one. Is this not good enough?

[Screencast_20260626.webm](https://github.com/user-attachments/assets/9b5ac6be-2a10-4fb7-bd4d-8b2e277f3684)

PR #6514, opened 2026-06-25
test: if basic-tests fails, don't fail other tests

WofWca:
What usually happens is that if "edit message" test runs
but some test after that fails, then accounts won't be deleted.
Instead they will be reused by other tests.
Then "clear chat removes messages" will also fail,
because it expects 2 messages in the chat,
but the chat would contain messages from the `basic-tests`.


nicodh (approved):
Assuming the failing tests here are fixed with #6511

WofWca:
Wrong assumption!
I fixed it now.

nicodh (approved):

PR #6518, opened 2026-06-26
chore: upgrade to pnpm v11

nicodh:
Needed some adaptions in package.json/pnpm-workspace

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,36 +38,6 @@
     "node": ">=22 <27",
     "pnpm": ">=9.6.0"
   },
-  "pnpm": {
-    "supportedArchitectures": {
-      "os": [
-        "darwin",
-        "win32",
-        "linux"
-      ],
-      "cpu": [
-        "x64",
-        "ia32",
-        "arm64"
-      ]
-    },
-    "onlyBuiltDependencies": [
-      "electron",
-      "esbuild"
-    ],
-    "overrides": {
-      "stylus": "npm:empty-npm-package@1.0.0",
-      "esbuild-plugin-inline-worker>esbuild": "^0.25.0",
-      "minimatch@>=3.0.0 <3.1.4": "^3.1.4",
-      "minimatch@>=5.0.0 <5.1.8": "^5.1.8",
-      "minimatch@>=9.0.0 <9.0.7": "^9.0.7",
-      "minimatch@>=10.0.0 <10.2.3": "^10.2.3"
-    },
-    "ignoredBuiltDependencies": [
-      "@parcel/watcher",
-      "electron-winstaller"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -38,36 +38,6 @@
...
-      ]
-    },
-    "onlyBuiltDependencies": [
-      "electron",
-      "esbuild"
-    ],
-    "overrides": {
-      "stylus": "npm:empty-npm-package@1.0.0",
-      "esbuild-plugin-inline-worker>esbuild": "^0.25.0",
-      "minimatch@>=3.0.0 <3.1.4": "^3.1.4",
```
nicodh, line 61: these minimatch overrides seemed to be not needed anymore (removing them resolved to the same packages as with the overrides)

pnpm-workspace.yaml
```diff
@@ -1,6 +1,25 @@
 packages:
   - packages/*
 
+supportedArchitectures:
```
nicodh, line 4: moved to pnpm-workdpace from the package.json "pnpm" field, which pnpm v11 no longer reads

see https://pnpm.io/migration

pnpm-workspace.yaml
```diff
@@ -1,6 +1,25 @@
...
+    - linux
+  cpu:
+    - x64
+    - ia32
+    - arm64
+
+allowBuilds:
+  esbuild: true
+  '@parcel/watcher': false
+  electron-winstaller: false
```
WofWca, line 17: Is it required for the pnpm@11 migration to execute these packages' install scripts?

nicodh: No, they were in ignoredBuiltDependencies before

WofWca: Soo, why not keep them there?

nicodh: https://pnpm.io/migration

"Consolidates build-dependency settings into allowBuilds. onlyBuiltDependencies, neverBuiltDependencies, ignoredBuiltDependencies, and onlyBuiltDependenciesFile are merged into a single allowBuilds map ({ name: true | false })."

WofWca: Ahhhh, I'm a dummy. I didn't see that this is a map and these are `false`.

WofWca (approved):
Tested with

```bash
rm **/node_modules/ -r
pnpm install
pnpm dev
```

Works.

I also tried reverting to `"packageManager": "pnpm@10.33.2"`. Then things still seem to work, so I guess this is also backwards-compatible? Good for us.

I also went through the migration guide. I only noticed that we have one script named "clean". So this is also an improvement for us probably.

PR #6521, opened 2026-06-27
fix: better contrast for full media buttons

### LINKED ISSUE #6519: buttons in image view partly difficult tp read
the used background/foreground combination of buttons in the image view is partly difficult to read or not readable at all. see the "back" button on the following random photo of a balcony i was sitting on:

Image

let's use the battle-proofed transparency background use elsewhere here as well:

```
background with alpha: #55444444
foreground:            #FFFFFF
```

this is used for info messages, markers, also on desktop, buttons on android and more.

it is the same color combination for dark mode and light mode.

other buttons in the enlarge image view have the potential same issue, let's use the same color there as well - this also solves teh issue that ppl regard eg. "close" as being disabled (as too grey) - we're getting reports from that here and there. as the button will be more recognisable then, we can use the same icon size than elsewhere

nicodh:
fixes #6519

### DIFF
--- a/packages/frontend/scss/gallery/_attachment-view.scss
+++ b/packages/frontend/scss/gallery/_attachment-view.scss
@@ -20,7 +20,7 @@
 .btn-wrapper {
   float: right;
   position: absolute;
-  background-color: rgba(0, 0, 0, 0.75);
+  background-color: var(--fullScreenMediaButtons);
   border-radius: 3px;
   display: flex;
   gap: 5px;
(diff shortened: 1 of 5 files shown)


packages/frontend/scss/gallery/_attachment-view.scss
```diff
@@ -20,7 +20,7 @@
 .btn-wrapper {
   float: right;
   position: absolute;
-  background-color: rgba(0, 0, 0, 0.75);
+  background-color: var(--fullScreenMediaButtons);
   border-radius: 3px;
   display: flex;
   gap: 5px;
```
r10s: let's unify gap with border and the gap used eg. in profiles.
 
apart from unification, this looks more coherent.

```suggestion
  gap: 8px;
```

r10s (approved):
very nice much better.

reviewed by testing, and also adding a little suggestion below

r10s (approved):

PR #6522, opened 2026-06-28
feat: show proxy icon

### LINKED ISSUE #5509: no hint that a proxy is enabled, no easy way to switch proxies

nicodh:
resolves https://github.com/deltachat/deltachat-desktop/issues/5509

Shows an icon above the chatlist after the qr button 

- extracts ChatListHeader to own component

When connected:

image

When not connected:

image


packages/frontend/src/components/screens/MainScreen/styles.module.scss
```diff
@@ -34,7 +34,7 @@
     @include navbar.navbar;
     display: flex;
     align-items: center;
-    padding: 0px 10px;
+    padding-inline-start: 10px;
```
WofWca: Minor: this got uglier when ~~there are 0 proxies~~ the proxy is disabled. Sounds like it's better to apply margins to the buttons.

Image

packages/frontend/src/components/screens/MainScreen/ChatListHeader.tsx
```diff
@@ -0,0 +1,113 @@
...
+      return
+    }
+    return onDCEvent(accountId, 'ConnectivityChanged', refreshConnectivity)
+  }, [accountId, proxyEnabled, refreshConnectivity])
+
+  // When a proxy is enabled, show an icon in the chat navbar
+  // indicating the current connectivity status (green/orange)
+  const proxyConnected =
+    connectivityFetch?.lingeringResult?.ok === true &&
+    connectivityFetch.lingeringResult.value >= C.DC_CONNECTIVITY_WORKING
```
WofWca, line 64: Minor: unhandled error and "loading" state.

packages/frontend/src/components/screens/MainScreen/ChatListHeader.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const refreshConnectivity = connectivityFetch?.refresh
+  useEffect(() => {
+    if (!proxyEnabled || !refreshConnectivity || accountId == undefined) {
+      return
+    }
+    return onDCEvent(accountId, 'ConnectivityChanged', refreshConnectivity)
+  }, [accountId, proxyEnabled, refreshConnectivity])
+
+  // When a proxy is enabled, show an icon in the chat navbar
+  // indicating the current connectivity status (green/orange)
```
WofWca:
```suggestion
```

"green/orange" is outdated, and the comment is weirdly placed + IDK if it's useful.

packages/frontend/src/components/screens/MainScreen/ChatListHeader.tsx
```diff
@@ -0,0 +1,113 @@
...
+              aria-label={tx('proxy_settings')}
+              title={tx('proxy_settings')}
+              styling='borderless'
+            >
+              <Icon
+                coloring='navbar'
+                icon={proxyConnected ? 'proxy' : 'proxy-not-connected'}
+                size={18}
+              ></Icon>
+            </Button>
```
WofWca, line 107: See https://github.com/deltachat/deltachat-desktop/pull/6454/changes#r3434965028

WofWca (approved):
I did not assess the UI/UX aspect of this MR (see my comments on the issue instead). Otherwise looks good.

packages/frontend/src/components/screens/MainScreen/styles.module.scss
nicodh: Fixed it but will make a followup for more consistent button styling

packages/frontend/src/components/screens/MainScreen/ChatListHeader.tsx
nicodh: In our UI it is (should) not be possible to enable proxy without having at least one proxy in the list

WofWca: Yes, but does this contradict what I said?

nicodh: I read your comment " the icon is supposed to be hidden if (and, I assume, only if) the list of proxies is empty." - if proxy_enabled is true but proxy list is empty hide it which in our case should not be possible.

WofWca: Anyway the behavior in this MR is not in line with Android and with the issue description.

nicodh: Ok, sorry after reading twice "this issue is only about adding the icon, which should show "enabled" (checkmark in the shield) or "not enabled" (only shield), if no proxies are set (list is empty), the icon should be hidden. the latter is also the default."

I see: it is not so much about an empty list and proxy_enabled but about having items in the list but  proxy_enabled set to false. So we should show an empty icon if there is a proxy URL configured even if proxy is generally disabled.

I will check with @r10s and post a followup PR for that

PR #6524, opened 2026-06-29
fix: set contextIsolation to true

nicodh:
- add preload script to the bundler
- add ipc channel whitelist


packages/target-electron/package.json
```diff
@@ -36,8 +36,9 @@
     "build:backend": "node ./bin/build.js",
     "build:calls-webapp": "node ../../bin/copy.js ./node_modules/@deltachat/calls-webapp/dist ./html-dist/calls-webapp",
     "build:runtime-impl": "pnpm esbuild --format=esm --bundle --minify --keep-names --sourcemap --outdir=./html-dist runtime-electron/runtime.ts",
-    "build:compose-frontend": "node ../../bin/copy.js ../frontend/html-dist ./html-dist && node ../../bin/copy.js ./static ./html-dist && pnpm build:runtime-impl",
-    "watch:compose-frontend": "node ../../bin/copy.js ../frontend/html-dist ./html-dist -w & node ../../bin/copy.js ./static ./html-dist -w & pnpm build:runtime-impl --watch",
+    "build:preload": "pnpm esbuild --platform=node --format=cjs --bundle --minify --keep-names --sourcemap --external:electron --outdir=./html-dist runtime-electron/preload.ts",
```
WofWca, line 39: Why `cjs` instead of `esm`?

packages/target-electron/runtime-electron/ipc-channels.ts
```diff
@@ -0,0 +1,89 @@
...
+ * `send` etc. these lists should restrict an untrusted renderer
+ * to call arbitrary ipc channels that aren't meant for it
+ */
+export const INVOKE_CHANNELS = Object.freeze([
+  'app.copyFileToInternalTmpDir',
+  'app.deleteSticker',
+  'app.removeTempFile',
+  'app.setBadgeCountAndTrayIconIndicator',
+  'app.writeTempFile',
+  'app.writeTempFileFromBase64',
```
WofWca, line 12: For later:
As I said in private, it would be nicer to make the entire `Runtime` API be exposed with `exposeInMainWorld`, but some methods, such as `createDeltaChatConnection` cannot be passed across the bridge currently. So that's for later.

https://github.com/deltachat/deltachat-desktop/blob/32d32154c786040fdc645e89e29fc3394ad351af/packages/runtime/runtime.ts#L39-L41

WofWca (approved):
Great, this should improve security against front-end compromise a little,
though there is still a lot that can be done.

I only tested with `pnpm dev`.

The types ensure that we haven't missed any channel names in the white list, which is good.

I also reviewed the irrelevant changes:
- The move of the `.startsWith('http:')` check to the main process looks good, more secure against front-end compromise.
- The rename of `ClickOnNotification` is fine, more consistent.
- `writeClipboardToTempFile` was indeed unused.
- 'show' also seems to be unused. At least there is no `'show'` in our code base (apart from the window event listener).

packages/target-electron/package.json
nicodh: https://www.electronjs.org/docs/latest/tutorial/esm#sandboxed-preload-scripts-cant-use-esm-imports

PR #6527, opened 2026-06-29
Button refactoring

nicodh:
- replace SearchInputButton with IconButton
- replace navbar-button (ChatView) with IconButton
- replace 3 dot menu button with 90° rotated "more" icon with vert_more icon


packages/frontend/src/components/Icon/styles.module.scss
```diff
@@ -39,3 +39,18 @@
...
+  border-radius: 10px;
+  height: 24px;
+  min-width: 24px;
+  padding-block: 0;
+  padding-inline: 1px 0;
+
+  &:hover {
+    background-color: var(--navBarButtonHover);
+  }
+}
```
WofWca: According to the class name and the prop description, everything except `border-radius` and `background-color` needs to be removed.

I think such stuff needs to be declared in another component that wraps `IconButton` (or `.scss` file). The `IconButton` itself then only needs to accept `className` and `iconClassName`.
Otherwise this just bloats the `IconButton` with stuff that is only ever used inside the chat list header. I don't think that this is a good refactor.

packages/frontend/src/components/screens/MainScreen/styles.module.scss
```diff
@@ -36,6 +36,10 @@
     align-items: center;
     padding: 0px 10px;
 
+    > button {
```
WofWca, line 39: minor: I'd prefer to be more specific. e.g. `.chatListHeaderButton {`

WofWca (approved):
The MR description and commit messages are missing the following points:
- Split up `SearchInput`: move "scan QR" buttons out of it.
- Add `data-no-drag-region` for the "Proxy" button.
- Make the "Proxy" button style (hover, specifically) match the style of "QR" and "clear search" buttons.

PR #6533, opened 2026-07-07
refactor: bin scripts cleanup

nicodh:
This refactoring started with finding an obsolete script (topEvil.sh) and the need to run some local scripts without polluting git workdir

It creates some subfolders (build, tools, translations) to give the 14 scripts in the bin dir some more structure and removes the lib dir.

I updated all path references and tested:
pnpm update:translations
pnpm prepare-release
pnpm check

I will test on the build machine that the build process still runs without errors

I also removed some outdated entries in .gitignore


nicodh:
Added a commit to fix the failing tauri on windows build:
https://github.com/deltachat/deltachat-desktop/pull/6533/changes/4bf2f4ca05d975c82a8e4cd7dd2ce1c8c6d7ad07

pnpm-workspace.yaml
```diff
@@ -1,6 +1,17 @@
...
+# Run package.json scripts through pnpm's POSIX shell emulator so inline env
+# vars (e.g. `NODE_ENV=production ...`) work cross-platform, incl. Windows CI.
+# (pnpm 11 no longer reads `shell-emulator` from .npmrc — it must live here.)
+shellEmulator: true
+
+# Shorten node_modules/.pnpm/<module> dir names so Windows doesn't hit the
+# 260-char path limit (nsis packaging breaks on the long allowOnlyOneInstallerInstance.nsh
+# path). The `pnpm pack:patch-node-modules` script needs the stdio package folder
+# names unchanged; longest there is 49 chars, and 70 leaves headroom without truncating those.
+virtualStoreDirMaxLength: 70
```
WofWca, line 13: This is AI-generated, right? Is this line necessary? The value is 60 on Windows anyway

https://pnpm.io/settings#virtualstoredirmaxlength

WofWca: Ah I see, this has been copy-pasted from here.

https://github.com/deltachat/deltachat-desktop/blob/4d8f77c73e76bdcc157f81d6a68c664aaa61d17a/.npmrc#L4-L10

nicodh: This is just copy pasted from .npmrc

pnpm 11 does not read these values from .npmrc anymore:

https://pnpm.io/settings

I just copy pasted both settings from the existing settings.

So: no it is not AI "generated" but the comment was a bit reworded and I considered it useful since the meaning didn't change.

The config itself was added  in https://github.com/deltachat/deltachat-desktop/commit/4d8f77c73e76bdcc157f81d6a68c664aaa61d17a

And it says: "the longest folder name there is 49 chars, but nsis processing breaks on allowOnlyOneInstallerInstance.nsh path
 and it seems to be fixed by setting this to 70"

So removing it should go into another PR if you want to do that...

bin/build/update-changelog.sh
```diff
@@ -2,12 +2,12 @@
 # Helper script to update changelog and footer links
```
WofWca, line 2: Minor: `update-changelog.sh` doesn't sound like it belongs to `build`?

bin/tools/setup-apparmor-dev.sh
```diff
@@ -11,7 +11,7 @@
 set -euo pipefail
```
WofWca, line 11: Minor: `bin/tools`, IDK if this is a good name. One can say that everything inside of `bin` is "tools", so it might be simpler to just have those right in `bin/`.

WofWca (approved):
Most of the changes make sense.
`topEvil` can be removed I think. We have removed Blueprint for the most part.
The `shellEmulator` change also looks good. Basically a follow-up to the `pnpm` migration.

I have checked that this MR doesn't touch production code and doesn't introduce or remove any logic (besides the `topEvil` script).
But I haven't tested any of the scripts, and did not verify that no things like docs and comments got outdated. If something breaks, that will basically only block the development workflow or release.

bin/build/update-changelog.sh
nicodh: Yes, I also moved release related stuff to build - scripts needed to "build" a release not only to "build" in the sense of compiling

bin/tools/setup-apparmor-dev.sh
nicodh: tools is of course the generic name for almost anything, but I meant it as "not fitting into other categories"

PR #6534, opened 2026-07-07
fix: show appropriate proxy icon

### LINKED ISSUE #5509: no hint that a proxy is enabled, no easy way to switch proxies

nicodh:
closes #5509

Now we always show an icon if there are proxy URLS configured. Independant of the proxyEnabled settings

I keep the icon at the right side o the qr button since the qr button is replaced with the clear button when a search query is in the search input field.

I also kept the different icons showing proxy connected or not connected as they were already there :-)


packages/frontend/src/components/Icon/index.tsx
```diff
@@ -36,7 +36,8 @@ export type IconName =
   | 'person-filled'
   | 'phone'
   | 'plus'
-  | 'proxy'
```
WofWca, line 39: The file also needs to be deleted.

WofWca (approved):
Seems to work.
Makes the behavior in line with Android.
The code looks good (apart from see my comment).

The connectivity toast doesn't need to be updated I think (i.e. it's OK as is in this MR).

PR #6535, opened 2026-07-07
fix: improve chatlist rendering on many changes

### LINKED ISSUE #5650: Poor performance when downloading many messages
- Operating System (Linux/Mac/Windows/iOS/Android): Intel macOS 15.7.1 (24G231)
- Delta Chat Version: 2.22.0 (git: v2.22.0)
- Expected behavior: Delta Chat remains responsive and does not spin my fans up when downloading messages
- Actual behavior: Delta Chat spins my fans up and becomes laggy when downloading messages
   1. Receive 100–500 messages to an account without the client open
   2. Open the Delta Chat application
   3. Wait while it downloads the large number of messages
   4. Observe fans spinning up, high processor/graphics usage (did not open Activity Monitor fast enough to get numbers)
- Screenshots: N/A
- Logs: Available upon request, I'm tired and don't want to go through to redact 28 MB of log data right now.

### Additional Information
I noticed this when opening Delta Chat on my laptop after a day of not looking at it. The recent issue with the main Delta Chat discussion group (DC Community) created a significant number of messages.

I don't know exactly what the cause is, but I suspect that it might be somewhat alleviated by collecting sustained, rapid UI updates into small batches (maybe 8 or 16 events at a time). I hope that it is not being caused by decryption work—the messages were sent in a group, which I understand to be encrypted using a shared AES key, and my laptop CPU has AES-NI support.

If nobody else is interested in looking at this, I can poke around with the profiling tools to see what I can find. It'll have to be in several weeks from now though; I already have too many projects on my plate.

nicodh:
Fixes #5650

In my tests (with ~1000 unread messages) that already avoided the ui freeze.

Collecting ChatChanged events and throttling the loadChats as a first step.
Skip unneccessary rerenders as a second step

We possibly could delay the reload even more by taking  network status Connectivity.WORKING in account as an additional check to wait longer with refreshing the chat list.

### DIFF
--- a/packages/frontend/src/components/chat/ChatContextMenu.tsx
+++ b/packages/frontend/src/components/chat/ChatContextMenu.tsx
@@ -326,13 +326,31 @@ export function useChatContextMenu(): {
     activeChatId: number | null
   ) => {
     // only if chatListItems contains a single chat, we need the full chat
+    const singleChatItem =
+      chatListItems.length === 1 ? chatListItems[0] : undefined
     const selectedChat =
-      chatListItems.length === 1
-        ? await BackendRemote.rpc.getFullChatById(
-            accountId,
-            chatListItems[0].id
-          )
+      singleChatItem !== undefined
+        ? await BackendRemote.rpc.getFullChatById(accountId, singleChatItem.id)
         : undefined
+    // The passed `chatListItems` come from the chat list's cache, which is
+    // updated with a throttle, so right after an action (e.g. pin/unpin)
+    // they may still hold the previous state for a moment.
+    if (selectedChat !== undefined) {
+      chatListItems = [fullChatToChatListItem(selectedChat)]
+    } else {
+      try {
+        const freshItems = await BackendRemote.rpc.getChatlistItemsByEntries(
+          accountId,
+          chatListItems.map(chat => chat.id)
+        )
+        chatListItems = chatListItems.map(chat => {
+          const freshItem = freshItems[chat.id]
+          return freshItem?.kind === 'ChatListItem' ? freshItem : chat
+        })
+      } catch (err) {
+        log.error('failed to fetch current chatlist items for menu', err)
+      }
+    }
     return openContextMenuInternalHandler(
       event,
       selectedChat,
--- a/packages/frontend/src/components/chat/ChatList.tsx
+++ b/packages/frontend/src/components/chat/ChatList.tsx
@@ -55,6 +55,7 @@ import { useSettingsStore } from '../../stores/settings'
 import { useMultiselect } from '../../hooks/useMultiselect'
 import { getLogger } from '@deltachat-desktop/shared/logger'
 import { useHasChanged2 } from '../../hooks/useHasChanged'
+import asyncThrottle from '@jcoreio/async-throttle'
 
 const log = getLogger('ChatList')
 const useMultiselectLog = getLogger('ChatListMultiselect')
@@ -816,23 +817,31 @@ export function useLogicVirtualChatList(
   useEffect(() => {
     const loadingChats = loadingChatsRef_.current
 
-    let debouncingChatlistItemRequests: { [chatid: number]: number } = {}
-
-    const updateChatListItem = async (chatId: number) => {
-      debouncingChatlistItemRequests[chatId] = 1
-      loadingChats.add(chatId)
-      const chats = await BackendRemote.rpc.getChatlistItemsByEntries(
-        accountId,
-        [chatId]
-      )
-      setChatCache(cache => ({ ...cache, ...chats }))
-      loadingChats.delete(chatId)
-      if (debouncingChatlistItemRequests[chatId] > 1) {
-        updateChatListItem(chatId)
-      } else {
-        debouncingChatlistItemRequests[chatId] = 0
+    // Batch `ChatlistItemChanged` events: collect the changed chat ids
+    // and update them with a single RPC call and a single `setChatCache`
+    // This reduces the number of re-renders bounded when many chats change
+    // in a short period, e.g. when a lot of messages arrive
+    const pendingChatIds = new Set<number>()
+    const flushPendingChatIds = asyncThrottle(async () => {
+      const chatIds = [...pendingChatIds]
+      pendingChatIds.clear()
+      if (chatIds.length === 0) {
+        return
       }
-    }
+      try {
+        const chats = await BackendRemote.rpc.getChatlistItemsByEntries(
+          accountId,
+          chatIds
+        )
+        setChatCache(cache => ({ ...cache, ...chats }))
+      } catch (err) {
+        log.error('failed to fetch changed chatlist items', err)
+      } finally {
+        for (const chatId of chatIds) {
+          loadingChats.delete(chatId)
+        }
+      }
+    }, 400)
 
     const removeListener = onDCEvent(
       accountId,
@@ -842,25 +851,19 @@ export function useLogicVirtualChatList(
           return
         }
         if (chatId !== null) {
-          if (
-            debouncingChatlistItemRequests[chatId] === undefined ||
-            debouncingChatlistItemRequests[chatId] === 0
-          ) {
-            updateChatListItem(chatId)
-          } else {
-            debouncingChatlistItemRequests[chatId] =
-              debouncingChatlistItemRequests[chatId] + 1
-          }
+          pendingChatIds.add(chatId)
+          loadingChats.add(chatId)
+          flushPendingChatIds.invokeIgnoreResult()
         } else {
-          // invalidate whole chatlist cache and reload everyhting that was visible before
+          // invalidate whole chatlist cache and reload everything that was visible before
+          pendingChatIds.clear()
           await reloadChats()
-          // reset debouncing
-          debouncingChatlistItemRequests = {}
         }
       }
     )
     return () => {
       removeListener()
+      flushPendingChatIds.cancel()
     }
   }, [accountId])
 
--- a/packages/frontend/src/components/chat/ChatListHelpers.tsx
+++ b/packages/frontend/src/components/chat/ChatListHelpers.tsx
@@ -4,6 +4,7 @@ import { debounce } from 'debounce'
 import { BackendRemote, onDCEvent } from '../../backend-com'
 import { selectedAccountId } from '../../ScreenController'
 import { useFetch } from '../../hooks/useFetch'
+import { useStableArray } from '../../hooks/useStableArray'
 import asyncThrottle from '@jcoreio/async-throttle'
 
 const log = getLogger('renderer/helpers/ChatList')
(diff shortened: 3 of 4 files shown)


packages/frontend/src/components/chat/ChatListHelpers.tsx
```diff
@@ -210,10 +211,16 @@ function useChatListSimple(
     )
   }
 
+  // Make sure we return the previous array as long as the IDs and their
+  // order does not change, avoiding a re-render of the whole chat list.
+  const chatListIds = useStableArray(
+    chatListFetch?.lingeringResult?.ok
+      ? chatListFetch.lingeringResult.value
+      : []
+  )
```
WofWca, line 220: Is it not possible to fix this bug in Core?

packages/frontend/src/components/chat/ChatList.tsx
```diff
@@ -816,23 +817,31 @@ export function useLogicVirtualChatList(
...
-      setChatCache(cache => ({ ...cache, ...chats }))
-      loadingChats.delete(chatId)
-      if (debouncingChatlistItemRequests[chatId] > 1) {
-        updateChatListItem(chatId)
-      } else {
-        debouncingChatlistItemRequests[chatId] = 0
+    // Batch `ChatlistItemChanged` events: collect the changed chat ids
+    // and update them with a single RPC call and a single `setChatCache`
+    // This reduces the number of re-renders bounded when many chats change
+    // in a short period, e.g. when a lot of messages arrive
```
WofWca, line 823: Minor: this sound like a commit message.

packages/frontend/src/components/chat/ChatList.tsx
```diff
@@ -816,23 +817,31 @@ export function useLogicVirtualChatList(
...
+    const pendingChatIds = new Set<number>()
+    const flushPendingChatIds = asyncThrottle(async () => {
```
WofWca, line 825: Minor: This should be named either `fetchPendingChats` or `throttledFetchPendingChats`. "flush" in the context of throttling means "stop throttling and invoke the underlying function now".

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -326,13 +326,31 @@ export function useChatContextMenu(): {
...
+        const freshItems = await BackendRemote.rpc.getChatlistItemsByEntries(
+          accountId,
+          chatListItems.map(chat => chat.id)
+        )
+        chatListItems = chatListItems.map(chat => {
+          const freshItem = freshItems[chat.id]
+          return freshItem?.kind === 'ChatListItem' ? freshItem : chat
+        })
+      } catch (err) {
+        log.error('failed to fetch current chatlist items for menu', err)
```
WofWca, line 351:
```suggestion
        throw new Error('failed to fetch current chatlist items for menu', {
          cause: err,
        })
```

No reason to consider the invocation of  `openChatListContextMenu` successful when this error occurs.

packages/frontend/src/components/chat/ChatContextMenu.tsx
```diff
@@ -326,13 +326,31 @@ export function useChatContextMenu(): {
...
-        ? await BackendRemote.rpc.getFullChatById(
-            accountId,
-            chatListItems[0].id
-          )
+      singleChatItem !== undefined
+        ? await BackendRemote.rpc.getFullChatById(accountId, singleChatItem.id)
         : undefined
+    // The passed `chatListItems` come from the chat list's cache, which is
+    // updated with a throttle, so right after an action (e.g. pin/unpin)
+    // they may still hold the previous state for a moment.
```
WofWca, line 337: The action happens on a _throttle_ and not a debounce, and even the throttle period is quite short. So it'd be better to adjust the tests instead of degrading performance.
Or at least reload only the chats that are actually pending a reload.
But I can live with it.

WofWca (commented):
I think that we can't say that this fixes the issue, because chat list reloading is not the only thing that keeps happening frequently as Core fetches the messages. The actual IMO will have to happen primarily in Core.

But yes, throttling and batching chat list reloading is a good thing.
Also good to get rid of this bug-prone-looking `debouncingChatlistItemRequests`.

packages/frontend/src/components/chat/ChatListHelpers.tsx
nicodh: Yes but that's another issue. I would also make sense to get something like an "xy new messages on the server" info up front to  make that visible for the user

packages/frontend/src/components/chat/ChatContextMenu.tsx
nicodh: > degrading performance

with one db call??

WofWca: Yes, one DB call. If the user's disk is busy, this can result in a several-second delay, as opposed to just always opening the menu instantly.

packages/frontend/src/components/chat/ChatListHelpers.tsx
WofWca: Sorry, I didn't get your point, I don't see how this has anything to do with UI.

packages/frontend/src/components/chat/ChatContextMenu.tsx
nicodh: It will just use the cached items as we do it right now. So no reason to throw an error

packages/frontend/src/components/chat/ChatContextMenu.tsx
nicodh: If one db call would cost several seconds we are not usable anyway

packages/frontend/src/components/chat/ChatContextMenu.tsx
WofWca: Ah. Sorry. I misread the code.

packages/frontend/src/components/chat/ChatListHelpers.tsx
nicodh: > Sorry, I didn't get your point

Same

Maybe you should be more explicit: what is "this" you are referring to?

WofWca: What the code appears to be doing is compensating the fact that Core fires the `ChatlistChanged` even when the list didn't actually change? Wouldn't that be a bug then, one that should be fixed in Core, instead of us doing workarounds?

nicodh: The problem is here: https://github.com/chatmail/core/blob/cac8c06c78dfb47341bd81c669f845948de76d5c/src/context.rs#L725 and here https://github.com/chatmail/core/blob/cac8c06c78dfb47341bd81c669f845948de76d5c/src/context.rs#L709

So the chatlistChanged event is emitted when a message changes. If we change that we would have to make sure no other client relies on that event being emitted here.

So for now I prefer to fix it in the client and maybe start a discussion what and how to change in core independant of this PR

packages/frontend/src/components/chat/ChatList.tsx
nicodh: I meant it in the sense: this is the function to flush the pending chat IDs

packages/frontend/src/components/chat/ChatList.tsx
nicodh: I think it makes sense: since the most common use case is that few events will be triggered it maybe get lost that there is a case where many events are triggered almost simultaneously

packages/frontend/src/components/chat/ChatListHelpers.tsx
nicodh: But what obviously is wrong in core, is the comment in https://github.com/chatmail/core/blob/cac8c06c78dfb47341bd81c669f845948de76d5c/src/events/chatlist_events.rs#L4

which claims the event is only triggered when the order or content of chat list changed

WofWca, line 220: Core bugs aside, do you mind specifying where exactly this saves on re-renders? `ChatList`'s render function is going to get invoked regardless of whether the return value of this hook is memoized or not, isn't it?

If one really wishes to optimize this place, then it needs to be done at least inside of the function passed to `useFetch`, so that it returns the reference to the same array if the array didn't change (though I didn't test this, not 100.00% sure if this will work).

WofWca: Ohhhh, so it fires on every message change. Damn. No wonder it's so sluggish. Good find!

> If we change that we would have to make sure no other client relies on that event being emitted here.

We wouldn't have to because the event is only supposed to be fired when the list of chat IDs changes (as you pointed out below).

> But what obviously is wrong in core, is the comment

IMO It's not the comment that is wrong but the fact that the event is fired too much (which technically is not a bug IMO, but still horrible for performance).

WofWca: > going to get invoked

I just tested it (in dev mode though), so I'm pretty sure it's the case.

nicodh: Hmm, react rendering is really a rabbit hole. I introduced useStableArray since I noticed that chatListIds is always a new array, even if the order and content didn't change. But you are right: the list is still re-rendered on any message change. It turned out, wrapping the RovingTabindexContext.Provider value into useMemo still re-renders the list but at least only the affected chat item is re-rendered then while all other rows are just skipped.

But let's not over optimize.

I tested locally and just the first of the three commits in this PR (throttling & batching) already reduced the load in the UI a lot. So maybe let's go with that for now and address the chatListChange event issue in core.

nicodh: Hmm, looking at the core: there are 18 emit_chatlist_changed calls. So obviously the event should have the name chatlist_maybe_changed.

And looking at the PR that introduced this and the original PR it seems another rabbit hole.
https://github.com/chatmail/core/pull/4476
https://github.com/chatmail/core/pull/2850

and in 2022(!) @r10s already warned that the new event needs some debouncing to avoid exactly what we are trying to solve here. " that would make sense only if that new DC_EVENT_CHATLIST_CHANGED is debounced and only be emitted when all messages and updates arrived, avoiding UIs being blocked by lots of updates." 
https://github.com/chatmail/core/pull/2850#issuecomment-1061744560

WofWca: > So obviously the event should have the name chatlist_maybe_changed

It's a common practice that these "SomethingChanged" Core events might be "false-positive". Same goes also for e.g. the stuff we do for the archive unread counter:

https://github.com/chatmail/core/blob/9ea576f182ea4659a63f20510a221e8eca476725/src/chat.rs#L5343-L5349

> needs some debouncing

We can also easily do debouncing (throttling) on the front-end. See the `asyncThrottle` usage across our codebase.
Edit: ah it's already throttled, never mind.

nicodh:
Closing in favour of https://github.com/deltachat/deltachat-desktop/pull/6538

PR #6536, opened 2026-07-08
fix: scroll message down after image is loaded or after remote edit

### LINKED ISSUE #4698: Chat scrolls up when a message gets edited (such that it's taller)

### LINKED ISSUE #6236: Chat doesn't always scroll down when receiving an image
- Delta Chat Version: 2.48.0
- Expected behavior: Chat should scroll down for new messages
- Actual behavior: The chat scrolls down when you get the "downloading" premessage, but then when the height is updated because the image is loaded it does not scroll down. 
- Steps to reproduce the problem: receive an image
- Screenshots:
- Logs:

nicodh:
fixes #6236 
fixes #4698

### DIFF
--- a/packages/frontend/src/stores/messagelist.ts
+++ b/packages/frontend/src/stores/messagelist.ts
@@ -341,7 +341,15 @@ export class MessageListStore extends Store<MessageListState> {
         return modifiedState
       }, 'unlockScroll')
     },
-    messageChanged: (message: Type.Message) => {
+    messageChanged: (
+      message: Type.Message,
+      /**
+       * Should be true if the change can make the message taller,
+       * as it could be the case for incoming image messages
+       * (when the download finishes) or for edited messages
+       */
+      scrollToBottomIfClose: boolean = false
+    ) => {
       const messageLoadResult: Type.MessageLoadResult = {
         kind: 'message',
         ...message,
@@ -353,6 +361,9 @@ export class MessageListStore extends Store<MessageListState> {
             ...state.messageCache,
             [message.id]: messageLoadResult,
           },
+          viewState: scrollToBottomIfClose
+            ? ChatViewReducer.scrollToBottomIfClose(state.viewState)
+            : state.viewState,
         }
         return modifiedState
       }, 'messageChanged')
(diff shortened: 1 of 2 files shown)


packages/frontend/src/stores/messagelist.ts
```diff
@@ -341,7 +341,15 @@ export class MessageListStore extends Store<MessageListState> {
...
     },
-    messageChanged: (message: Type.Message) => {
+    messageChanged: (
+      message: Type.Message,
+      /**
+       * Should be true if the change can make the message taller,
+       * as it could be the case for incoming image messages
+       * (when the download finishes) or for edited messages
+       */
+      scrollToBottomIfClose: boolean = false
```
WofWca, line 351:
```suggestion
      scrollToBottomIfClose: boolean
```

Minor: let's avoid accidental omission of the argument.

WofWca (approved):
Interesting approach!
The code looks conservative and tidy.
I tested both of the cases. I haven't managed to come up with how this could be a regression. Probably because we scroll to bottom only when already close to the bottom.

What I noticed though is that if the growing message is not the last one, then this smooth scroll IMO looks janky: first the messages get shifted to the bottom and then we smooth-scroll to them again.
But it's probably more often that it's the last message that changes in height, so it's probably OK. Or can be addressed in a follow-up.

Also for https://github.com/deltachat/deltachat-desktop/issues/6236 specifically I think it'd be nicer to take the approach I suggested, to further reduce layout shifts.
(for that we would also need to render "not downloaded" messages the same as downloaded).

PR #6537, opened 2026-07-08
fix: don't announce reactions on chat open/scroll

WofWca:
Also don't announce the changes on window resize.
Only announce when the actual list of reactions changes, as intended.

Reproducible at least with Orca reader. Just open a chat
that has some messages with reactions, and you'll hear it.
Then also scroll up and you'll hear it more.

I suspect that this has to do with the fact that `messageWidth`
is initially 0 for all messages
and only gets updated inside of `useEffect`,
resuling in all reactions being initially collapsed.

Relevant:
- https://github.com/deltachat/deltachat-desktop/pull/4885
- https://github.com/deltachat/deltachat-desktop/pull/5051

### DIFF
--- a/packages/frontend/src/components/Reactions/index.tsx
+++ b/packages/frontend/src/components/Reactions/index.tsx
@@ -60,11 +60,15 @@ export default function Reactions(props: Props) {
 
   return (
     <div className={styles.reactions}>
-      {reactions.slice(0, visibleEmojis).map(({ emoji, isFromSelf, count }) => {
+      {reactions.map(({ emoji, isFromSelf, count }, index) => {
         return (
           <span
             className={classNames(styles.emoji, {
               [styles.isFromSelf]: isFromSelf,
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/Reactions/index.tsx
```diff
@@ -60,11 +60,15 @@ export default function Reactions(props: Props) {
 
   return (
     <div className={styles.reactions}>
-      {reactions.slice(0, visibleEmojis).map(({ emoji, isFromSelf, count }) => {
+      {reactions.map(({ emoji, isFromSelf, count }, ind) => {
```
nicodh: I would pefer to name the index explicit here, saving 2 chars makes no sense imo

nicodh (approved):
Reviewed by reading

Sounds useful to avoid the announcements of reactions depending on their visibility

PR #6539, opened 2026-07-09
improvement: move msg select counter to header

WofWca:
image 
image

As requested in
- https://github.com/deltachat/deltachat-desktop/pull/6268#issuecomment-4421685886.
- https://github.com/deltachat/deltachat-desktop/pull/6458#issuecomment-4717197495.

Had to do a bunch of refactoring to make the message list state accessible to the chat header (lift state up).

FTR I don't think that this is a good change (see my comments on the linked issues/MRs).

### DIFF
--- a/packages/frontend/scss/message/_message-list.scss
+++ b/packages/frontend/scss/message/_message-list.scss
@@ -206,18 +206,4 @@
       }
     }
   }
-
-  .num-selected-messages-status {
-    position: absolute;
-    bottom: 0;
-    inset-inline-start: 0;
-    border-start-end-radius: 0.25rem;
-    $borderColor: color-mix(in srgb, var(--contextMenuBorder) 50%, transparent);
-    border-block-start: 1px solid $borderColor;
-    border-inline-end: 1px solid $borderColor;
-
-    background: var(--composerBg);
-    padding: 0.5rem 0.625rem;
-    line-height: 1;
-  }
 }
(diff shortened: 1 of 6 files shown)


nicodh:
> FTR I don't think that this is a good change (see my comments on the linked issues/MRs).

What do you mean here? The visualization or the fact that we have to lift the state?
Which comment are you referring to?

WofWca:
Well the state lifting in itself I guess is not too bad, but yes, generally the state should be as local as possible.
What I mean primarily is that this IMO unnecessarily shifts layout, obscures the chat title (which open the "View Profile" dialog), also leading to us having to introduce this `hidden` hack, for example. Layout-wise we get the "chat" section which starts with "5 messages selected", then the list of apps, the "call" button, some more stuff, and then the message list itself.
I think on Desktop, multiselect doesn't have to be an entire "mode" that you need to "enter and exit". Multiselect is just a natural part of how lists should work.

r10s:
> I think on Desktop, multiselect doesn't have to be an entire "mode" that you need to "enter and exit". Multiselect is just a natural part of how lists should work.

that was maybe true 20 years ago,, but if you look at how eg. telgram or signal work, that is pretty much a "mode". and it make some sense to focus on what is important when multiple things are selected

but anyways, this PR is very much of an improvement to make the state visible. it is totally fine to be merged as is. 

as written somewhere else, it also makes sense to hide the input bar and offering the "things to do there", this is what signal/telegram/etc are doing.

but not sure even then, if it makes sense to change this PR then, the title showing what is selected make a lot of sense. let's give it a try, it is an improvement, tho some may disagree. and: we can always iterate

nicodh (approved):
Reviewed by reading & testing

WofWca:
I re-checked the changes to see if they clash with e4dedf44f. Didn't do in-depth research but at least couldn't come up with anything by looking at the diff.

PR #6542, opened 2026-07-14
perf: avoid unneeded messagelist renders

nicodh:
We had several places where non memoized objects triggered unnecessary renderings. Although the effect in production mode is quite small it is remarkable in dev mode and avoids superfluous rendering without introducing significant new logic.

### DIFF
--- a/packages/frontend/src/components/message/MessageList.tsx
+++ b/packages/frontend/src/components/message/MessageList.tsx
@@ -980,16 +980,6 @@ export const MessageListInner = React.memo(
         </div>
       </>
     )
-  },
-  (prevProps, nextProps) => {
-    const areEqual: boolean =
-      prevProps.activeView === nextProps.activeView &&
-      prevProps.messageCache === nextProps.messageCache &&
-      prevProps.oldestFetchedMessageIndex ===
-        nextProps.oldestFetchedMessageIndex &&
-      prevProps.onScroll === nextProps.onScroll &&
-      prevProps.onWheel === nextProps.onWheel
-    return areEqual
   }
 )
 
--- a/packages/frontend/src/components/message/MessageList.tsx
+++ b/packages/frontend/src/components/message/MessageList.tsx
@@ -670,14 +670,17 @@ export default function MessageList({
     }
   }, [])
 
+  const messagesDisplayContextValue = useMemo(
+    () => ({
+      context: 'chat_messagelist' as const,
+      chatId: chat.id,
+      isDeviceChat: chat.isDeviceChat,
+    }),
+    [chat.id, chat.isDeviceChat]
+  )
+
   return (
-    <MessagesDisplayContext.Provider
-      value={{
-        context: 'chat_messagelist',
-        chatId: chat.id,
-        isDeviceChat: chat.isDeviceChat,
-      }}
-    >
+    <MessagesDisplayContext.Provider value={messagesDisplayContextValue}>
       <MessageListInner
         onScroll={onScroll}
         onScrollEnd={onScrollEnd}
@@ -749,11 +752,14 @@ export const MessageListInner = React.memo(
       loadMissingMessages,
     } = props
 
-    const conversationType: ConversationType = {
-      hasMultipleParticipants: chat.chatType !== 'Single',
-      isDeviceChat: chat.isDeviceChat as boolean,
-      chatType: chat.chatType,
-    }
+    const conversationType: ConversationType = useMemo(
+      () => ({
+        hasMultipleParticipants: chat.chatType !== 'Single',
+        isDeviceChat: chat.isDeviceChat as boolean,
+        chatType: chat.chatType,
+      }),
+      [chat.chatType, chat.isDeviceChat]
+    )
 
     useKeyBindingAction(KeybindAction.MessageList_PageUp, () => {
       if (messageListRef.current) {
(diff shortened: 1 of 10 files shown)


packages/frontend/src/stores/messagelist.ts
```diff
@@ -227,11 +227,35 @@ export class MessageListStore extends Store<MessageListState> {
     super(defaultState(), 'MessageListStore')
   }
 
+  /**
+   * Cache {@linkcode MessageListStore.activeView} and return the cached reference
+   * if nothing changed, so that it doesn't trigger `React.memo` to rerender.
+   */
```
WofWca: "return" is a term for functions, so this comment looks a bit misplaced.
Also consider simply saying something like "this is memoized", as the pattern looks similar to `useMemo`.

packages/frontend/src/contexts/RovingTabindex.tsx
```diff
@@ -255,15 +256,18 @@ export function RovingTabindexProvider({
     [activeElement, classNameOfTargetElements, direction, wrapperElementRef]
   )
 
+  const value: ContextValue = useMemo(
```
WofWca, line 259: about commit messages: consider using `perf:` as a prefix. `fix:` is for bugs rather, or potential bugs at least.

Also the MR title needs to be updated, after the new commits have been pushed.

packages/frontend/src/components/message/MessageList.tsx
```diff
@@ -281,17 +280,21 @@ export default function MessageList({
...
     const newShowJumpDownButton =
       !isNewestMessageLoaded ||
       distanceToBottom > maxScrollToBottomDistanceConsideredShort
     // Don't flash the button during a programmatic smooth scroll —
     // we already know we're scrolling to the bottom.
     if (pendingProgrammaticSmoothScrollTo.current === null) {
-      if (newShowJumpDownButton != showJumpDownButton) {
-        setShowJumpDownButton(newShowJumpDownButton)
-      }
+      // React bails out of re-rendering if the value didn't change.
```
```suggestion
```

packages/frontend/src/components/message/focusAndMultiselect.tsx
```diff
@@ -66,7 +66,11 @@ export function useMessageFocusAndMultiselectContextValue(props: {
     // even though all you did is click a link inside of it.
     { onNormalClick: 'unselectAll' }
   )
-  return { ...multiselect, resetSelection }
+  // Memoized because it is used as a context value.
```
WofWca, line 69:
```suggestion
```

This is implied by the function name.

packages/frontend/src/components/message/MessageWrapper.tsx
```diff
@@ -17,7 +17,11 @@ type RenderMessageProps = {
 
 const log = getLogger('renderer/message/MessageWrapper')
 
-export function MessageWrapper(props: RenderMessageProps) {
+// `React.memo`, so that a single changed message doesn't re-render
+// every other message in the list.
```
WofWca: IDK if this comment is any good. This memoization is not much different from any other, and it saves not only against a message changing but rather against the message list re-rendering overall.

packages/frontend/src/components/message/MessageWrapper.tsx
```diff
@@ -17,7 +17,11 @@ type RenderMessageProps = {
...
+export const MessageWrapper = React.memo(function MessageWrapper(
```
WofWca, line 20: Also the body of this component is not that big, so the memoization benefit is also not that big. But it's still OK to memoize.

WofWca (approved):
I was also a little annoyed by the fact that `useContext` causes a re-render, and I wasn't sure how to fix that, so this is great, thanks!
https://react.dev/reference/react/useContext#caveats
> React automatically re-renders all the children that use a particular context starting from the provider that receives a different value. The previous and the next values are compared with the [Object.is](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) comparison.

Regarding whether this actually affects performance in production: even if not, my approach is that it's still worth fixing, so that the next time someone works on performance, they know where _not_ to look.
And also it ensures that when the current bottlenecks get fixed, the stuff fixed in this MR doesn't become the new bottleneck.

I didn't check though that all these changes have an effect or if there is still some hidden always-re-created function or something.

The change to the tests looks sensible. 
The fixed "unchecked index access" is good.

Also I guess it's worth pointing out that the React Compiler is coming soon™️, so it will memoize all stuff automatically one day.
And IMO for the most part our bottleneck is the JSON-RPC requests. Either too many of them, or the Core being too slow, or the IPC, or the GC.

packages/frontend/src/components/message/focusAndMultiselect.tsx
nicodh: The comment is about why it's memoized

packages/frontend/src/stores/messagelist.ts
nicodh: return here refers to get activeView()

nicodh: Moved the placement

packages/frontend/src/components/message/MessageWrapper.tsx
nicodh: removed it

PR #6543, opened 2026-07-14
chore: parallelize build scripts

WofWca:
This reduces the run duration of `pnpm build:electron` by ~50%
on my machine. Although the time varies a lot,
so I didn't measure the exact times.
But this makes sense anyway.

See https://pnpm.io/cli/run#running-multiple-scripts
> You may run multiple scripts at the same time by using a regex instead of the script name.

I decided not to touch the `watch:` scripts because I wasn't sure
if they can handle this parallelization.

Also the `check` scripts can be parallelized this way,
but then the output will look a little different,
so I also omitted that for now.

Without the first two commits the last one would be broken.
The "themes" stuff would be a race, and the "source map support"
would fail due to a missing `html-dist` (also a race).

Note that devs will need to delete the now un-git-ignored `fallback-theme.css`.

I only tested the Electron version and E2E tests, but not Tauri.


packages/frontend/package.json
```diff
@@ -5,12 +5,15 @@
...
   "scripts": {
     "check:types": "tsc --noEmit --noUncheckedIndexedAccess false",
-    "build": "pnpm build:ts && pnpm build:themes && pnpm build:static && pnpm build:source-map-support && pnpm build:scss",
+    "build": "pnpm run \"/^build:.*/\"",
     "build:ts": "node ./bin/build-frontend-ts.mjs",
     "build:source-map-support": "node ./bin/build-frontend-source_map_support.js",
     "build:scss": "sass ./scss/manifest.scss ./html-dist/main.css --embed-source-map --style=compressed",
-    "build:static": "node ../../bin/tools/copy.js ../../static ./html-dist && node ../../bin/tools/copy.js ./static ./html-dist && node ../../bin/tools/copy.js ../../images ./html-dist/images",
-    "build:themes": "sass themes/light.scss ./html-dist/fallback-theme.css --style=compressed --embed-source-map && node ./bin/build-frontend-themes.js",
+    "build:static:1": "node ../../bin/tools/copy.js ../../static ./html-dist",
```
nicodh: build:static:root
build:static:frontend

would be better readable imo

nicodh (approved):

Very nice!

Issue #5308, opened 2025-07-20
option to disable/enable the menu bar (maybe disable by default)


Lilithmoi:
the menu bar is an old non-modern ui element that tends to be replaced by others, as it's generally not intuitive or even that nice looking. in delta chat's case, large portion of the menu bar's functionality is already provided elsewhere in the app - and if not then it should be.

the toggle to disable or enable the menu bar can be put in the appearance tab in settings as the option for showing tray icon is there too.

the only missing functionality from the menu bar would be in its view tab where the option to change the app's zoom level and language are. these can also be put in the appearance tab as those options are somewhat related. though at that point it would be better to simply call it "display" instead of "appearance"

WofWca:
<img width="233" height="62" alt="Image" src="https://github.com/user-attachments/assets/2e77b9ea-2fe3-4071-975b-d3e8a8635eb7" />

I agree, we should move stuff to regular UI IMO.
Though we now rely on zoom, with accelerators (shortcuts), IDK if we can keep but also remove the menu as well.

Electron also has an option to auto-hide the menu. IIRC I played around with it, but something was in the way, maybe https://github.com/electron/electron/pull/45930. Or maybe I'm misremembering.

Also see forum post with the same request: https://support.delta.chat/t/hide-menu-bar-linux-desktop/2032?u=wofwca.

Edit: BTW some other windows, such as webxdc, help window, HTML email viewer also have the menu bar, with different items sometimes.

WofWca:
Going now through the menu items again. These are worth paying attention to:
- "File -> Quit": already possible with the Tray icon.
  Unfortunately though the tray icon is not available on Gnome.
- "Edit": nothing that can't be done otherwise
- "View"
  - Float on top. I think this is not super useful, but maybe a little useful.
  - **LANGUAGE**
  - Developer stuff
- "Help"
  - "Report an issue" is not available elsewhere

I think we should
- [ ] move the "language" setting to the regular "settings",
  - [ ] A problem with this is that the "settings" button is not available on first launch when you have no accounts yet.
- [ ] then we can at least set the menu to auto-hide, at least by default. It will still be accessible but one has to discover it by pressing Alt.

r10s:
is auto-hide-menu also the default behavior for signal/telegram/other-messenger on _windows?_  (mac is always show anyways, and linux is diverse anyways)

WofWca:
On WhatsApp and Telegram for Windows there is no menu at all.
[...]

PR #6544, opened 2026-07-14
feat: add Hide Menu Bar setting

Bushikot:
## Summary

Adds a **"Hide Menu Bar"** setting that lets users auto-hide the native menu bar (File, Edit, View, Help). When hidden, it only appears when <kbd>Alt</kbd> is pressed (Linux/Windows).

On **macOS** the setting is a no-op — the native menu bar is always visible per platform convention. In Tauri, the menu item is not constructed at all (`#[cfg(not(target_os = "macos"))]`). In Electron, the menu item is hidden on macOS via `visible: process.platform !== 'darwin'`.

Default value is **`false`** (menu bar visible by default).

[Screencast From 2026-06-23 19-21-08.webm](https://github.com/user-attachments/assets/11b2ea24-f7c0-4d17-8764-90d9c41239c0)

### Electron
- `src/windows/main.ts` — `applyHideMenuBar()`, `toggleHideMenuBar()`, `isHideMenuBar()`
- `src/menu.ts` — checkbox in View menu (between Float on Top and Toggle Full Screen); hidden on macOS via `visible: process.platform !== 'darwin'`
- `src/ipc.ts` — IPC handler calls the single `applyHideMenuBar()` entry point
- Secondary windows (help `windows/help.ts`, html-email `windows/html_email.ts`, webxdc `deltachat/webxdc.ts`) apply the setting at creation time via `autoHideMenuBar: DesktopSettings.state.hideMenuBar`
- Setting passed to the main `BrowserWindow` at init via `hideMenuBar` option

#### Windows coverage (Electron)

| Window | Menu bar behaviour |
|---|---|
| main | `autoHideMenuBar` + `setMenuBarVisibility()` driven by the setting (live-updated on toggle) |
| help / html-email / webxdc | `autoHideMenuBar` applied at window creation |
| video-call | hardcoded `autoHideMenuBar: true` — see rationale below |

**Video call window** (`windows/video-call.ts`) keeps a hardcoded `autoHideMenuBar: true`. It is a transient, focused window that only lives for the duration of a call and ships its own minimal View-only menu (zoom / fullscreen / devtools). Forcing the menu bar to stay visible during a call (which is what `hideMenuBar: false` would otherwise do) would be distracting, and auto-hiding it on Alt is already the most aggressive useful behaviour — so the user setting is intentionally not applied there.

### Tauri
- Four menu definitions (main, help, html-email, webxdc) — each has `HideMenuBar` checkbox
- `settings.rs` — `get_hide_menu_bar()`, `apply_hide_menu_bar()`, `toggle_hide_menu_bar()`
- Uses `window.remove_menu()` / `window.set_menu()` (Tauri has no `autoHideMenuBar`)
- Called at startup in `lib.rs`
- `#[cfg(not(target_os = "macos"))]` on menu item and apply logic

### Frontend
- `DesktopSettingsSwitch` for `hideMenuBar` in Appearance settings (hidden in browser)
- `onDesktopSettingChanged` runtime hook — syncs backend-triggered changes to the UI

### Shared / Runtime
- `DesktopSettingsType.hideMenuBar: boolean` — default `false`
- `Runtime.onDesktopSettingChanged` interface — implemented for Electron & Tauri

## Notes
- macOS: menu item not constructed (Tauri) / hidden via `visible: false` (Electron); `applyHideMenuBar`/`apply_hide_menu_bar` are no-ops
- Tauri `inner_update_all` respects the hide state, so toggling from either the menu item or the Settings UI keeps the main window and all secondary windows (help, html-email, webxdc) consistent. `MenuManager::register_window` also respects the setting, so a secondary window opened while hiding is enabled never shows a menu bar in the first place.
- Electron secondary windows apply the setting when they are opened; an already-open secondary window is not live-updated when the setting is toggled afterwards (minor gap, matching the accepted Tauri behaviour). Closing and reopening the window picks up the new value.

## Testing
- Developed and manually tested on **Linux** only (Wayland), for Electron, Tauri & Browser targets (Firefox). MacOS and Windows builds were **not** manually run.
- Platform-specific paths are guarded: Tauri uses `#[cfg(not(target_os = "macos"))]` for menu item construction and `apply_hide_menu_bar`; Electron uses `process.platform === 'darwin'` guards in `applyHideMenuBar()` and `visible: process.platform !== 'darwin'` on the menu item.
- `cargo check`, `cargo clippy`, `pnpm check:types` and `pnpm check:lint`/`check:format` pass. Manual verification of the actual hide/show behaviour on macOS and Windows would be appreciated from reviewers.


packages/frontend/src/components/Settings/Appearance.tsx
```diff
@@ -132,6 +133,12 @@ export default function Appearance({
         label={tx('pref_use_system_ui_font')}
         callback={() => ThemeManager.refresh()}
       />
+      {!isBrowser && (
```
nicodh: This settings should also be hidden for MacOS, since it can't be applied.
 Use const { isMac } = runtime.getRuntimeInfo()

packages/target-tauri/src-tauri/src/state/menu_manager.rs
```diff
@@ -104,8 +115,23 @@ impl MenuManager {
...
 
-        // #[cfg(not(target_os = "macos"))]
-        win.set_menu(menu_generator_arc(app)?)?;
+        // On non-macOS, respect the "hide menu bar" setting already at
+        // window registration, so a secondary window opened while hiding is
+        // enabled does not flash (or permanently show) a menu bar until the
+        // next `update_all`.
+        #[cfg(not(target_os = "macos"))]
+        {
+            let menu = menu_generator_arc(app)?;
```
nicodh: No need to generate the menu bar if it is hidden, so maybe better move this into the condition together with win.set_menu(menu)?;

nicodh (commented):
LGTM

I tested this on Linux and Windows and it worked as expected!

Bushikot:
Hi @nicodh, thanks for the review! I'll address your comments and provide an update by the end of the week. Also, as a maintainer, what are your thoughts on setting the default for this new option to "true"? From my point of view, this menu looks a bit cluttered and redundant for casual users.

nicodh:
resolves #5308

packages/frontend/src/components/Settings/Appearance.tsx
r10s: wanted to write the same

packages/target-electron/runtime-electron/runtime.ts
```diff
@@ -447,6 +450,9 @@ class ElectronRuntime implements Runtime {
       this.onShowDialog?.('keybindings')
     )
     ipcBackend.on('showSettingsDialog', () => this.onShowDialog?.('settings'))
+    ipcBackend.on('desktop-setting-changed', (_ev, key, value) =>
+      this.onDesktopSettingChanged?.(key, value)
```
nicodh, line 457: the _ev argument should be passed here, since the  ipcRenderer strips the first argument again.
(https://github.com/deltachat/deltachat-desktop/blob/5f6efcded5087ea76d617e34fc07253c3123e954/packages/target-electron/runtime-electron/preload.ts#L35-L36)

with the result that an open Appearance settings dialog would not immediately sync with changes triggered from the view menu

This also means the signature of the runtime function onDesktopSettingChanged has to be changed

```typescript 

onDesktopSettingChanged:
    | ((
        ev: Event,
        key: keyof DesktopSettingsType,
        value: string | number | boolean
      ) => void)
    | undefined
```

Changing the key argument type to `keyof DesktopSettingsType` would also avoid the type cast in 

https://github.com/deltachat/deltachat-desktop/pull/6544/changes#diff-dd59c5e2954f0ee544433c4062f121387869fa70f166d16f39989a96298d8601R240

nicodh (changes requested):
After the changes I reviewed in depth now and found another issue (with small impact but worth to be fixed imo)

packages/target-electron/runtime-electron/runtime.ts
Bushikot: Addressed in dc6b095f9.
1. **`_ev` arg shift** (`runtime.ts:456`): `(_ev, key, value)` -> `(key, value)`.
2. **`key: keyof DesktopSettingsType`**: updated in the base `Runtime` interface and all three implementations (electron, tauri, browser). The `as any` in `stores/settings.ts:239` is gone.
**On `ev: Event` in your suggested signature** - I left it out. The preload strips `IpcRendererEvent` unconditionally for every channel (`preload.ts:35-36`):
```ts
// don't leak the IpcRendererEvent across the bridge
ipcRenderer.on(channel, (_ev, ...args) => callback(...args))
```
so `ipcBackend.on`'s callback type is `(...args: any[]) => void` - no event ever reaches a handler. Every other handler in the file confirms this (`(locale)`, `(url)`, `(data)`, `()` - none have an event parameter). An `ev: Event` first arg would always be `undefined`. That mismatch was the actual bug: the old `(_ev, key, value)` bound `key` to the value and left `value` as `undefined`, so the Appearance dialog never synced.
If you want the event forwarded, the preload bridge would need to be changed for all channels - let me know if that's the intent.

nicodh: No it's fine. It's better not to pass the to evt to onDesktopSettingChanged at all.

nicodh (approved):

Thank you for contributing to the project!!

packages/target-electron/src/windows/main.ts
```diff
@@ -415,6 +416,30 @@ export function isAlwaysOnTop() {
...
+}
+
+export function toggleHideMenuBar() {
+  if (!window) return
+  if (process.platform === 'darwin') return
+  const flag = !window.autoHideMenuBar
+  applyHideMenuBar(flag)
+  DesktopSettings.update({ hideMenuBar: flag })
+  // Notify frontend (Settings UI) about the change from View menu
+  window.webContents.send('desktop-setting-changed', 'hideMenuBar', flag)
```
WofWca:
```suggestion
  send('desktop-setting-changed', 'hideMenuBar', flag)
```

packages/runtime/runtime.ts
```diff
@@ -212,6 +212,12 @@ export interface Runtime {
     | undefined
   onResumeFromSleep: (() => void) | undefined
   onToggleNotifications: (() => void) | undefined
+  onDesktopSettingChanged:
+    | ((
+        key: keyof DesktopSettingsType,
+        value: string | number | boolean
+      ) => void)
+    | undefined
```
WofWca, line 220:
```suggestion
  onDesktopSettingChanged:
    | (<T extends keyof DesktopSettingsType>(
        key: T,
        value: DesktopSettingsType[T]
      ) => void)
    | undefined
```

Same for other occurrences.

WofWca (approved):
Tested it. Works as intended by the author.

Good:
- In the "help" window there is nothing that is important and is only accessible via the menu, so it's good to auto-hide.
- OK not to change `video-call.ts`, where it's always `true`.

Not so good:
- In the WebXDC window, the "Source Code" menu item is now harder to discover. Normal people probably don't use that, and the source code link can also be found in the app picker, but I'm not sure.
- When the menu is hidden like this, it might not be obvious that you can show it by pressing Alt, so people might have a hard time finding the "language" option
- On touch screens I'm pretty sure it's impossible to invoke the menu?
- On Tauri hiding the menu makes the "Language" setting completely inaccessible. I'd prefer to simply not change Tauri at all, especially looking at how much code it is.

Overall I think it'd be nicer not to introduce this setting, and instead go for what I suggested in https://github.com/deltachat/deltachat-desktop/issues/5308#issuecomment-4593663343.

But as long as we're gonna make follow-up changes, I think it's OK to merge as is.
Also I think we should not enable it by default. See my comment on the issue.

packages/runtime/runtime.ts
Bushikot: Hi @WofWca , thanks for your feedback!
I deliberately set the default for auto-hiding to false so that new users can easily find the menu on their first run. Since this feature is off by default, the issues you mentioned won't affect regular users. They will only apply to those who manually choose to hide the menu after launch.

- WebXDC: I will check what's available there, as I don't have strong opinion yet 
- Touch screens: You may hide/show menu via Setting -> Appearance -> Hide Menu Bar (check my video above)
- Language switch (all platforms and build targets): My intent in general is to move all the options uniquely available in the menu to the graphical Settings window. Then I guess it will be safe to make auto hide option enabled by default.

I will make sure to address this comment of yours:
> A problem with this is that the "settings" button is not available on first launch when you have no accounts yet.

@WofWca @nicodh please let me know if you want to keep changes granular and isolated, or it's better to proceed to menu items migration here as a part of this PR?

nicodh: I would say if the Tauri change is reverted this can be merged

The suggestion for type change and other stuff can be applied later

Bushikot:
Hi guys, Tauri changes have been removed completely following your request. I also made the UI flow for the "Hide menu bar" switch element more consistent by making it "disabled" (same as for "Show Tray Icon") instead of hiding it completely for targets where it's not applicable.

Screenshot From 2026-07-28 20-21-21

nicodh:
Thanks for contributing!! 🎉

PR #6547, opened 2026-07-15
chore: add desktopName to package.json

nicodh:
Based on the findings in https://github.com/Foundry376/Mailspring/pull/2657

### DIFF
--- a/packages/target-electron/package.json
+++ b/packages/target-electron/package.json
@@ -27,6 +27,7 @@
     "node": ">=20.0.0"
   },
   "productName": "DeltaChat",
+  "desktopName": "deltachat.desktop",
   "homepage": "https://delta.chat",
   "scripts": {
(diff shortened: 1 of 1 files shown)


WofWca (commented):
Soo, what is this actually supposed to do in practice? Is this not gonna break existing installations? We don't manually generate the `.desktop` file unlike Mailspring, so do we even need this? And are you sure that this is the right `package.json` to change? Because we have `name` also in the root one:

https://github.com/deltachat/deltachat-desktop/blob/6b62a8eb58809070ff144bc7cf53385baf0e4692/package.json#L3

In Electron docs `desktopName` is only mentioned [here](https://www.electronjs.org/docs/latest/api/base-window#winsetprogressbarprogress-options):
> ... desktopName field in package.json. By default, it will assume {app.name}.desktop.

nicodh:
CLosing in favour of #6559

PR #6548, opened 2026-07-15
test: use docker chatmail relay

nicodh:
This introduces the usage of a local chatmail relay running in a docker container with an IP address instead a domain. It should make our tests more independant of network conditions and external servers.

But it turned out, so far we can't use Instant Onboarding with an IP only relay since core does not support it (yet?).

The last commit adds a test file that holds all instant onboarding dependant tests, but needs an external chatmail server. I would suggest to keep these tests until there might be a solution in core. 

See https://github.com/chatmail/core/issues/8211

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -3,14 +3,14 @@ on:
   push:
     branches: [main]
     paths-ignore:
-      - 'docs/**'  # only trigger build if a file outside of /docs was changed
+      - 'docs/**' # only trigger build if a file outside of /docs was changed
       - '_locales/*.xml'
       - '.vscode/**'
       - 'README_ASSETS/**'
       - 'README.md'
   pull_request:
     paths-ignore:
-      - 'docs/**'  # only trigger build if a file outside of /docs was changed
+      - 'docs/**' # only trigger build if a file outside of /docs was changed
       - '_locales/*.xml'
       - '.vscode/**'
       - 'README_ASSETS/**'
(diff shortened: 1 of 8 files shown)


nicodh:
Changed the config to run the e2e tests with both relays (local & ci-chatmail) parallel

packages/e2e-tests/tests/instant-onboarding-group.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+ */
+
+test.describe.configure({
+  mode: 'serial',
+})
+
+test.skip(
+  () => isIpAddress(chatmailServerDomain),
+  'cannot onboard against a self-signed IP-only relay, see chatmail/core#8211'
+)
```
WofWca:
```suggestion
skipOnIpRelay()
```

packages/e2e-tests/tests/instant-onboarding-group.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+} from '../playwright-helper.js'
+
+/**
+ * Joining a group via instant onboarding needs the DCACCOUNT mechanism,
+ * which does not work against a self-signed IP-only relay
+ * (see `skipOnIpRelay`), so all tests that need it live in their own
+ * self-contained file. In CI, they are skipped for the local IP-based relay
+ * and run against the remote relay (see .github/workflows/e2e.yml).
+ */
+
```
WofWca, line 24:
```suggestion
```

This is also described in `skipOnIpRelay` docs.

Also I suspect that the comment is false because we have other tests which are skipped based on whether the relay is a Docker relay.

.github/workflows/e2e.yml
```diff
@@ -20,44 +20,102 @@ permissions: {}
...
+            sleep 10
+          done
+
+          docker logs chatmail
+          docker exec chatmail get-service-logs || true
+          exit 1
+      - name: Build browser target
+        run: pnpm build:browser
+      - name: Run e2e tests (${{ matrix.relay }} chatmail relay)
+        run: pnpm --filter=e2e-tests e2e
```
WofWca, line 84: Is the change from `pnpm e2e` to these 4 lines relevant to this MR?
Edit: ah I see, it was relevant until the last commit.

.github/workflows/e2e.yml
```diff
@@ -20,44 +20,102 @@ permissions: {}
...
       VERSION_INFO_GIT_REF: ${{ github.ref }}
-      DC_CHATMAIL_DOMAIN: ${{ vars.DC_CHATMAIL_DOMAIN }}
-      DC_MAIL_SERVER: ${{ vars.CI_MAIL_SERVER }}
-      DC_MAIL_SERVER_TOKEN: ${{ secrets.CI_MAIL_SERVER_TOKEN }}
-    timeout-minutes: 60
-    runs-on: ${{ matrix.os }}
+      DC_CHATMAIL_DOMAIN: ${{ matrix.chatmail_domain }}
+      DC_FRONTEND_NO_TLS: true
+      E2E_EXCLUDE_UNENCRYPTED: true
+    timeout-minutes: 75
```
WofWca: Does it often happen that running the tests takes more than 1 hour?

.github/workflows/e2e.yml
```diff
@@ -20,44 +20,102 @@ permissions: {}
...
       CI: true
       NODE_ENV: test
       VERSION_INFO_GIT_REF: ${{ github.ref }}
-      DC_CHATMAIL_DOMAIN: ${{ vars.DC_CHATMAIL_DOMAIN }}
-      DC_MAIL_SERVER: ${{ vars.CI_MAIL_SERVER }}
-      DC_MAIL_SERVER_TOKEN: ${{ secrets.CI_MAIL_SERVER_TOKEN }}
-    timeout-minutes: 60
-    runs-on: ${{ matrix.os }}
+      DC_CHATMAIL_DOMAIN: ${{ matrix.chatmail_domain }}
+      DC_FRONTEND_NO_TLS: true
```
WofWca: Was there a reason to move this from just the `pnpm e2e` step to the global scope?
Edit: ah, I see. It was needed until the last commit, now it's not.

.github/workflows/e2e.yml
```diff
@@ -20,44 +20,102 @@ permissions: {}
...
+      - name: Collect chatmail logs
+        if: ${{ !cancelled() && matrix.start_local_relay }}
+        run: |
+          mkdir -p chatmail-logs
+          docker logs chatmail > chatmail-logs/docker.log 2>&1 || true
+          docker exec chatmail get-service-logs > chatmail-logs/services.log 2>&1 || true
+      - name: Stop chatmail relay
+        if: ${{ always() && matrix.start_local_relay }}
+        working-directory: ./.ci/docker
+        run: docker compose down --volumes
```
WofWca:
```suggestion
```

What if we don't stop it?

packages/e2e-tests/playwright.config.ts
```diff
@@ -17,6 +17,10 @@ loadEnv()
  */
 export default defineConfig<TestOptions>({
   testDir: './tests',
+  testIgnore:
+    process.env.E2E_EXCLUDE_UNENCRYPTED === 'true'
+      ? '**/unencrypted-group-tests.spec.ts'
+      : undefined,
```
WofWca: This is irrelevant to this MR and doesn't do anything, right?

packages/e2e-tests/tests/chat-list-multiselect.spec.ts
```diff
@@ -189,6 +189,11 @@ test.describe('Shift + Click', () => {
       .getByRole('button', { name: 'Delete' })
       .click()
     await expectSelectedChats([])
+    // Wait until the chat is removed from the chat list (the update is
+    // throttled) before Shift + clicking: as long as the deleted chat is
+    // still in the list, it serves as the selection anchor and the click
+    // would range-select from the top of the list.
```
WofWca:
```suggestion
```

packages/e2e-tests/.gitignore
```diff
@@ -1,6 +1,7 @@
 node_modules/
 /test-results/
 /playwright-report/
+/playwright-report-public/
```
WofWca:
```suggestion
```

packages/e2e-tests/tests/group-tests.spec.ts
```diff
@@ -647,6 +580,18 @@ test('create channel and add members', async ({ browserName }) => {
     .filter({ hasText: channelName })
   await expect(channelChatItemB).toBeVisible()
 
+  // Wait for the securejoin handshake to finish: the resulting
+  // "You joined the channel." message must be seen in the (auto-opened)
+  // channel chat while userB is still selected. Otherwise it arrives
+  // after we switch to userA and counts as a second unread message,
+  // making the badge assertion below flaky
```
WofWca:
```suggestion
```

packages/e2e-tests/tests/instant-onboarding-group.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+  if (browserName.toLowerCase().indexOf('chrom') > -1) {
+    await page.context().grantPermissions(['clipboard-read', 'clipboard-write'])
+  }
+  const userA = getUser(0, existingProfiles)
+  const userB = getUser(1, existingProfiles)
+  const newUserName = userNames[2] ?? 'Chris'
+  const groupInviteMessage = `${userA.name} invited you to join this group`
+
+  // Establish contact between userA and userB and create the group
+  // that the new user is then invited to.
```
WofWca:
```suggestion
```

packages/e2e-tests/tests/instant-onboarding-group.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+  isIpAddress,
+} from '../playwright-helper.js'
+
+/**
+ * Joining a group via instant onboarding needs the DCACCOUNT mechanism,
+ * which does not work against a self-signed IP-only relay
+ * (see `skipOnIpRelay`), so all tests that need it live in their own
+ * self-contained file. In CI, they are skipped for the local IP-based relay
+ * and run against the remote relay (see .github/workflows/e2e.yml).
+ */
```
WofWca, line 23: > we have other tests which are skipped based on whether the relay is a Docker relay.

Ah and this also means that there is no need to move this test to a separate file?

WofWca (approved):
Great overall!

As a next step I think we'd want to make it easier to run this locally. Maybe by adding pnpm scripts.

I have checked that the test moved to the new file is unchanged, except for its setup (first few lines).

Regarding the irrelevant changes:
- the fix for the "doesn't break if the item at the start of the selection gets removed" test makes sense.
- "move pointer" makes sense.
- "Readd user to group" test changes seem like a refactor, but I didn't look.
- Removed whitespace before a comment in `e2e.yml` is OK.

FYI I also opened a couple of MRs to simplify this one.

packages/e2e-tests/tests/instant-onboarding-group.spec.ts
nicodh: Included the suggestion in a commit

.github/workflows/e2e.yml
nicodh: reverted the change

packages/e2e-tests/tests/chat-list-multiselect.spec.ts
nicodh: changed the comment

packages/e2e-tests/tests/instant-onboarding-group.spec.ts
nicodh: we can leave these tests in the previous tests since depending on skipOnIpRelay we would end having different numbers of profiles in the following tests after an instant onboarding test

packages/e2e-tests/tests/instant-onboarding-group.spec.ts
nicodh: changed the comment

PR #6549, opened 2026-07-15
test: fix: delete DC config after app picker URL test

### LINKED ISSUE #6515: dev: `pnpm e2e` a second time always fails at "custom app picker URL"
Introduced recently.
This has to do with the settings not getting cleared at test start / end.
I guess the temp solution is to clear them when that test has ran.

https://github.com/deltachat/deltachat-desktop/blob/8865e870447384feadf173beca201baf0903c4e2/packages/e2e-tests/tests/basic-tests.spec.ts#L434-L467

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/6515.

### DIFF
--- a/packages/e2e-tests/playwright-helper.ts
+++ b/packages/e2e-tests/playwright-helper.ts
@@ -6,6 +6,7 @@ import {
   Page,
 } from '@playwright/test'
 import path from 'path'
+import fs from 'fs/promises'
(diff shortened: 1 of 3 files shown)


WofWca:
Didn't work... I guess I'll just restore the setting inside the test.

PR #6550, opened 2026-07-15
test: don't fail "app picker URL" on second run

### LINKED ISSUE #6515: dev: `pnpm e2e` a second time always fails at "custom app picker URL"

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/6515.

Ideally I'd prefer to nuke the data folder when the tests have run,
but we don't have the facilities for that yet.

### DIFF
--- a/packages/e2e-tests/tests/basic-tests.spec.ts
+++ b/packages/e2e-tests/tests/basic-tests.spec.ts
@@ -433,40 +433,62 @@ test('add app from picker to chat', async () => {
   expect(finalAppIconsCount).toBeGreaterThan(initialAppIconsCount)
 })
 
-test('custom app picker URL', async () => {
-  const userA = existingProfiles[0]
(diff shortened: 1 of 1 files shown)


nicodh (approved):
This PR fixes the problem but I think it's time to add a reset data dir method soon. Andy maybe a clickButton helper for all the page.getByRole('button', { name: 'foo' }).click() calls

WofWca:
> reset data dir method soon

I had an idea of creating a new temp dir each time.

PR #6552, opened 2026-07-16
fix: add bottom margin to copy dialog

nicodh:
Margin was missing between buttons and content

image


WofWca (approved):
Didn't test but the code looks OK.

PR #6553, opened 2026-07-16
test: reactivate unencrypted tests

nicodh:
Maybe if we just use it for the unencrypted tests it will be reliable enough to use the mailserver?

### DIFF
--- a/packages/e2e-tests/playwright-helper.ts
+++ b/packages/e2e-tests/playwright-helper.ts
@@ -399,6 +399,35 @@ export async function getProfile(
   }
 }
 
+/**
+ * Toggle the "Enforce Encryption for All Relays" setting
+ * Unencrypted ("New Email") chats can only be created when it is disabled.
+ */
+export async function setForceEncryption(
+  page: Page,
+  accountId: string,
+  enabled: boolean
+): Promise<void> {
+  await page.getByTestId(`account-item-${accountId}`).click({ button: 'right' })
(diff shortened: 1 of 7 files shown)


packages/e2e-tests/playwright-helper.ts
```diff
@@ -313,6 +399,35 @@ export async function getProfile(
...
+  await page.getByTestId('open-transport-settings').click()
+  await page.getByLabel('Edit Relay').first().click()
+  await page.locator('#show-advanced-button').click()
+  const switchInput = page.getByRole('checkbox', {
+    name: 'Enforce Encryption for All Relays',
+  })
+  if ((await switchInput.isChecked()) !== enabled) {
+    await switchInput.press('Space')
+  }
+  await expect(switchInput).toBeChecked({ checked: enabled })
```
WofWca, line 423:
```suggestion
  await page
    .getByRole('checkbox', { name: 'Enforce Encryption for All Relays' })
    .setChecked(enabled)
```

have you tried this? (I am not sure though if this will work: docs seem to state that the element is clicked unconditionally).

WofWca (approved):
Assuming that this MR is only about the last 2 commits: I think we can try it.

packages/e2e-tests/playwright-helper.ts
nicodh: It didn't work but I didn't investigate further

PR #6555, opened 2026-07-17
fix: non-main chat list items: wrong aria-haspopup

WofWca:
For example, in the dialog that opens
when you're going to "share contact",
or in a contact's "Chats in Common" section.

`onContextMenu` being non-null adds `aria-haspopup="menu"`
on chat list items, so we must specify a nullish value
instead of a no-op function.


nicodh (approved):
Reviewed by reading

PR #6556, opened 2026-07-17
improvement: "delete transport" to unpublish it

### LINKED ISSUE #6554: fix: adapt transport handling to core changes
- always unpublish instead of delete
- remove question whether to unpublish or delete
- adapt list using list_transports to not show deleted/unpublished ones
- in ui, the action is simply called "delete"

`delete_transport()` must not be used by UIs anymore. Instead, `set_transport_unpublished()` must be called when a user clicks on "Remove".

list_transports() doesn't return unpublished relays anymore. UIs should use `list_transports()` rather than `list_transports_ex()`, because unpublished transports count as removed from the user point of view, and should not be shown in the relay list anymore.

WofWca:
"Remove Relay" now acts as "Unpublish",
and unpublished relays are hidden from the user.

Closes https://github.com/deltachat/deltachat-desktop/issues/6554.

[Screencast.webm](https://github.com/user-attachments/assets/6ed6ac54-c8e2-4226-b1e7-3c0b97d62687)

One needs the dev version of Core to fully test this. Otherwise transports still show up after the "deletion". But things work otherwise.

See relevant Core commits:
- https://github.com/chatmail/core/commit/6322758482bde034966f2ac769e28d7e3cf1d7de
- https://github.com/chatmail/core/commit/90ac0580431dbc92a3fdcbcbbeda6d9f44967325

### DIFF
--- a/packages/frontend/src/components/dialogs/Transports/index.tsx
+++ b/packages/frontend/src/components/dialogs/Transports/index.tsx
@@ -1,13 +1,6 @@
 import React, { useCallback, useEffect, useState } from 'react'
 import { DialogProps } from '../../../contexts/DialogContext'
-import Dialog, {
-  DialogBody,
-  DialogHeader,
-  DialogFooter,
-  FooterActions,
-  FooterActionButton,
-  DialogContent,
-} from '../../Dialog'
+import Dialog, { DialogBody, DialogHeader, DialogFooter } from '../../Dialog'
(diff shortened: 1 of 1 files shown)


nicodh (approved):
Reviewed by reading & testing

WofWca:
Merging now, I'll re-test immediately.

WofWca:
Re-tested, with Core 2.55.0. Things work as on the video.

PR #6562, opened 2026-07-20
chore: upgrade core to 2.55.0

nicodh:


WofWca (approved):
The addition of [`minimumReleaseAgeExclude`](https://pnpm.io/settings#minimumreleaseageexclude) for Core makes sense because upcoming Core releases are communicated in advance instead of us going "oh look a new release on NPM, let's upgrade", so it's OK to ignore that safeguard.

This also
- downgrades `@types/node` for some reason as a dependency for some packages
- upgrades `plist` to the next patch version
- Tauri: Upgrades `@tauri-apps/plugin-log`
- Tauri: upgrades a bunch of Rust stuff

CI is failing only for Tauri. I think it's fine to merge nonetheless, and fix later.

PR #6564, opened 2026-07-21
fix: AppImage no longer requires libfuse2

### LINKED ISSUE #6563: AppImage shows error: depends on EOL libfuse2 library
see https://www.electron.build/docs/appimage/#legacy-toolset-000--fuse2

nicodh:
fixes #6563
see https://www.electron.build/docs/appimage/#modern-toolset-103--static-runtime

### DIFF
--- a/packages/target-electron/build/gen-electron-builder-config.js
+++ b/packages/target-electron/build/gen-electron-builder-config.js
@@ -164,6 +164,12 @@ build['appImage'] = {
   artifactName: '${productName}-${version}-${arch}.${ext}',
 }
 
+// see https://www.electron.build/docs/appimage/#modern-toolset-103--static-runtime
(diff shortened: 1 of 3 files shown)


WofWca (approved):
Makes sense, to upgrade.
But do you have an idea how urgent this is? We can keep it for another release or not?

nicodh:
> Makes sense, to upgrade. But do you have an idea how urgent this is? We can keep it for another release or not?

see https://itsfoss.com/cant-run-appimage-ubuntu/ for example

WofWca:
That's from over a year ago so IMU this can wait.

nicodh:
> That's from over a year ago so IMU this can wait.

Wait for what?

WofWca:
Wait for until we've released 2.55.0

PR #6567, opened 2026-07-21
fix: Edit Group looping between old and new state, + form validation

### LINKED ISSUE #6561: when changing chat title _and_ chat description together, updates are not reliable
when editing a chat title together with the description (eg. in release group :), the latter is _not_ updated.

looking more closely, in a testing group, there is a wild flickering between old and new description - and then the old description wins :)

https://github.com/user-attachments/assets/6f55efb7-15d7-439c-a633-7a0ccff418d5

when only setting title or only setting description, things work reproducible.

tested on current main branch, macOS 15.7

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/6561.

See commit messages.
First one is required for the second one.

This fixes the bug, but I'd like to do more work on the "View Group" dialog in follow-ups. It's quite a mess. Edit: https://github.com/deltachat/deltachat-desktop/pull/6568.

TODO:
- [x] Merge the base into main

### DIFF
--- a/packages/frontend/src/components/dialogs/ViewGroup/index.tsx
+++ b/packages/frontend/src/components/dialogs/ViewGroup/index.tsx
@@ -384,11 +384,7 @@ function ViewGroupInner(
         groupDescription: string,
         groupImage: string | null
       ) => {
-        // TODO this check should be way earlier, you should not be able to "OK" the dialog if there is no group name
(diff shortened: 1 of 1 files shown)


nicodh (approved):
Reviewed by reading & testing

PR #6569, opened 2026-07-21
chore: upgrade core to 2.56.0

nicodh:


WofWca (approved):
Just an upgrade + a removal of the `oauth2` field from an object, which already doesn't do anything as far as I can see.

Issue #6073, opened 2026-03-01
Address of email contacts can't be copy-pasted


missytake:
- Operating System (Linux/Mac/Windows/iOS/Android): Xubuntu 24.04
- Delta Chat Version: 2.43.0
- Expected behavior: I can mark the email address of an email contact to copy-paste it.
- Actual behavior: The email address can't be selected.
- Steps to reproduce the problem: open a chat with an email contact, click on the email address in the contact title

This makes classic email usage much more annoying compared to 2.35.0, without adding a benefit. The contact view could simply continue to show the email address, as before - right now I can't see the email address at all in the contact view. Showing it in the contact view would probably be better than making it selectable in the chat title, as clicking on it also has the function of opening the contact view.

For encrypted contacts it would also be helpful to be able to mark the email address in the encryption info; and it would definitely be useful if the *fingerprint* could be marked and selected, so I can compare it to a fingerprint in a terminal.

r10s:
we do not want to have the email address in the contact profile, to simplify and unify layout, not even in unencrypted chats. it is complex enough. having it there sometimes adds confusion. the subtitle is a compromise already.

ref: https://github.com/deltachat/deltachat-desktop/issues/5938

for copy+paste, idk, if this is really a big thing worth targeting. i would be fine with a right-click on the subtitle and a "copy"
option there, tho. or, having "encryption info" copyable.

WofWca:
See https://github.com/deltachat/deltachat-desktop/pull/5941, https://github.com/deltachat/deltachat-desktop/issues/5938.

If we display the email in the profile header, why should we hide it in the "profile" dialog? The "profile" dialog is supposed to be a "more detailed" view of a profile, compared to the header, isn't it?

r10s:
well, here the consistency among profile rules. 

in general, showing the email address does harm. we want that as few as possible. also for classic mail. profiles are eg also show when tapping on group avatars and at other places

also, it is consistent with all other platforms now. that was hard enough. 

let's do not consider changing that again, or picking up that discussion again, raising questions about other cornercases.

as said, a right-click and select+copy in encryption info seems okay

PR #6570, opened 2026-07-22
fix disable Ctrl+a selecting all text

### LINKED ISSUE #6558: ctrl+a does weird select of all UI text
I enter multi-select mode, selecting one item, then I press ctrl+a to select all, instead this selects a mix of UI text including the palceholder of the input field, the "1 selected", "today" and some messages' text

even before the multi-select feature this ctrl+a behavior was problematic and I remember ending up in such weird selections but now it is even more likely to happen when people try ctrl+a which is common for "select all" when doing multi-selection

Image

nicodh:
fixes #6558

tbd: I added one exception: if a user has selected some text (with mouse down & drag or double click) Ctrl-a extends the selection (as expected I would say)

~~This also avoids the short "all text selected" rendering when selecting multiple messages~~

### DIFF
--- a/packages/frontend/src/components/message/focusAndMultiselect.tsx
+++ b/packages/frontend/src/components/message/focusAndMultiselect.tsx
@@ -89,6 +89,23 @@ export function useMessageFocusAndMultiselect(
       },
       [messageId, multiselect]
     ),
+    onMouseDown: useCallback(
+      (e: React.MouseEvent) => {
+        // Shift + Click extends the message selection. The browser would
+        // also extend the text selection until the re-render adds
+        // `multiselected-two-or-more` to the list and fades it out.
+        //
+        // Only when something is already selected: with an empty selection
+        // Shift + Click is still a plain "select this text" gesture.
+        if (e.shiftKey && multiselect.selectedItems.size >= 1) {
+          e.preventDefault()
+          // `preventDefault()` on `mousedown` also suppresses focus,
+          // which the roving tabindex needs.
+          elementRef.current?.focus()
+        }
+      },
+      [elementRef, multiselect]
+    ),
     onFocus: useCallback(() => {
       rovingTabindex.setAsActiveElement()
 
--- a/packages/frontend/src/components/message/focusAndMultiselect.tsx
+++ b/packages/frontend/src/components/message/focusAndMultiselect.tsx
@@ -89,23 +89,6 @@ export function useMessageFocusAndMultiselect(
       },
       [messageId, multiselect]
     ),
-    onMouseDown: useCallback(
-      (e: React.MouseEvent) => {
-        // Shift + Click extends the message selection. The browser would
-        // also extend the text selection until the re-render adds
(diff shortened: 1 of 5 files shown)


packages/frontend/src/components/message/focusAndMultiselect.tsx
```diff
@@ -89,6 +89,23 @@ export function useMessageFocusAndMultiselect(
...
+        // Shift + Click is still a plain "select this text" gesture.
+        if (e.shiftKey && multiselect.selectedItems.size >= 1) {
+          e.preventDefault()
+          // `preventDefault()` on `mousedown` also suppresses focus,
+          // which the roving tabindex needs.
+          elementRef.current?.focus()
+        }
+      },
+      [elementRef, multiselect]
+    ),
```
WofWca: A second Shift + Click on text doesn't do anything

packages/frontend/src/hooks/useSelectAllKeyboardShortcut.ts
```diff
@@ -0,0 +1,60 @@
...
+
+      const target = ev.target
+      if (
+        target instanceof HTMLInputElement ||
+        target instanceof HTMLTextAreaElement ||
+        (target instanceof HTMLElement && target.isContentEditable)
+      ) {
+        return
+      }
+      ev.preventDefault()
```
WofWca: Ctrl + A in the emoji picker search is broken

WofWca (commented):
As anticipated, too much code that is too flaky. I don't think we should proceed with this approach.

packages/frontend/src/components/message/focusAndMultiselect.tsx
nicodh: I removed the click+Shift handler since that fixes another problem than this PRs issue is about. So the second Shift+click event works as expected (and multi select with Shift+click has a blinking text selection again)

packages/frontend/src/hooks/useSelectAllKeyboardShortcut.ts
nicodh: Is fixed using composedPath which also works inside shadow DOM

nicodh:
In my opinion this PR improves the situation a lot.

I also replaced the existing Ctrl+a handler inApp.tsx which used a since long non existing evt.path property  with the useSelectAllKeyboardShortcut hook.

WofWca (approved):
I still think it's not worth it. Otherwise looks OK, didn't test.

WofWca:
This also applies to dialogs, including the "Encryption Info" dialog (see related https://github.com/deltachat/deltachat-desktop/issues/6073).

WofWca:
Also it's still possible to select input placeholders by click-dragging and Shift + Clicking, so I'm not sure if we can say that this MR fully closes the issue.

nicodh:
This doesn't cover all possible scenarios but improves the behavior a lot. In encryption dialogs you still can use shift click to select the long text so I consider this acceptable compromise

PR #6571, opened 2026-07-22
fix: autostart on nixos not working

### LINKED ISSUE #6431: Autostart does not work on nixos package
```
cat .config/autostart/deltachat-desktop.desktop
[Desktop Entry]
Type=Application
Name=Delta Chat
Comment=Delta Chat decentralized private messenger
Exec="/nix/store/4q0bhnj778nfz0v5c0zg0g5sp090bbm6-electron-unwrapped-41.2.0/libexec/electron/electron" -- --minimized
Hidden=false
NoDisplay=false
X-GNOME-Autostart-enabled=true
```
`Exec` points to a specific version of electron, which is wrong for 2 reasons:
- it should start delta chat not electron `Exec=deltachat -- --minimized`
- it should not pin to the exact version that set autostart, because that will cause unexpected things when you update the app.

<details>
<summary>For reference this is the file for tauri (that works), it calls deltachat-tauri directly, maybe it contains other useful things as well. (maybe `StartupNotify=false`, `Terminal=false`)
</summary>

```
cat .config/autostart/deltachat-tauri.desktop
[Desktop Entry]
Comment[en_US]=deltachat-tauristartup script
Comment=deltachat-tauristartup script
Exec=deltachat-tauri
GenericName[en_US]=
GenericName=
MimeType=
Name[en_US]=deltachat-tauri
Name=deltachat-tauri
Path=
StartupNotify=false
Terminal=false
TerminalOptions=
Type=Application
Version=1.0
X-KDE-SubstituteUID=false
X-KDE-Username=
```

</details>

Tested delta chat version: 2.49.1 (`/nix/store/5n5ig5jvvmd9skrmynhl1lw7m70nb62c-deltachat-desktop-2.49.1/bin/deltachat`)

nicodh:
fixes #6431

### DIFF
--- a/packages/target-electron/src/autostart.ts
+++ b/packages/target-electron/src/autostart.ts
@@ -34,9 +34,29 @@ function getLinuxAutostartFilePath(): string {
   return join(getLinuxAutostartDir(), `deltachat-desktop${suffix}.desktop`)
 }
 
-function getLinuxExecPath(): string {
-  // When running as an AppImage, APPIMAGE env var is set to the .AppImage file path
-  return process.env.APPIMAGE || process.execPath
+// Returns the command for the Exec= key of the autostart desktop entry,
+// without the trailing app arguments (`-- --minimized`).
+function getLinuxExecCommand(): string {
+  if (process.env.FLATPAK_ID) {
+    return `flatpak run ${escapeDesktopExecArg(process.env.FLATPAK_ID)}`
+  }
+  if (process.env.APPIMAGE) {
+    // When running as an AppImage, APPIMAGE env var is set to the
+    // .AppImage file path
+    return escapeDesktopExecArg(process.env.APPIMAGE)
+  }
+  if (process.execPath.startsWith('/nix/store/')) {
+    // On Nix/NixOS, process.execPath is a version-pinned path to the shared
+    // `electron` binary (e.g. /nix/store/…-electron-…/…/electron), not to
+    // Delta Chat. Pinning it would break autostart on every update so rely
+    // on the `deltachat` command
+    // see https://github.com/deltachat/deltachat-desktop/issues/6431
+    return 'deltachat'
+  }
+  // Regular installs (e.g. the .deb): process.execPath is the Delta Chat
+  // executable itself, and its absolute path is unambiguous even when several
+  // package formats are installed in parallel.
+  return escapeDesktopExecArg(process.execPath)
 }
 
 // see https://specifications.freedesktop.org/desktop-entry/latest/exec-variables.html
@@ -72,7 +92,9 @@ function getLinuxDesktopFileContent(): string {
 Type=Application
 Name=Delta Chat
 Comment=Delta Chat decentralized private messenger
-Exec=${escapeDesktopExecArg(getLinuxExecPath())} -- --minimized
+Exec=${getLinuxExecCommand()} -- --minimized
+Terminal=false
+StartupNotify=false
 Hidden=false
 NoDisplay=false
 X-GNOME-Autostart-enabled=true
--- a/packages/target-electron/src/autostart.ts
+++ b/packages/target-electron/src/autostart.ts
@@ -49,9 +49,10 @@ function getLinuxExecCommand(): string {
     // On Nix/NixOS, process.execPath is a version-pinned path to the shared
     // `electron` binary (e.g. /nix/store/…-electron-…/…/electron), not to
     // Delta Chat. Pinning it would break autostart on every update so rely
-    // on the `deltachat-desktop` command
+    // on the `deltachat` command
     // see https://github.com/deltachat/deltachat-desktop/issues/6431
-    return 'deltachat-desktop'
+    // and https://github.com/NixOS/nixpkgs/blob/a1016521f29c094127657220a52ed0b09d689919/pkgs/by-name/de/deltachat-desktop/package.nix#L129
+    return 'deltachat'
   }
   // Regular installs (e.g. the .deb): process.execPath is the Delta Chat
   // executable itself, and its absolute path is unambiguous even when several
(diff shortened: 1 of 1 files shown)


packages/target-electron/src/autostart.ts
```diff
@@ -34,9 +34,29 @@ function getLinuxAutostartFilePath(): string {
...
 
-function getLinuxExecPath(): string {
-  // When running as an AppImage, APPIMAGE env var is set to the .AppImage file path
-  return process.env.APPIMAGE || process.execPath
+// Returns the command for the Exec= key of the autostart desktop entry,
+// without the trailing app arguments (`-- --minimized`).
+function getLinuxExecCommand(): string {
+  if (process.env.FLATPAK_ID) {
+    return `flatpak run ${escapeDesktopExecArg(process.env.FLATPAK_ID)}`
+  }
```
WofWca, line 42: FTR I peeked at how KeePassXC actually sets `Exec=flatpak run...`. Turns out they're indeed using portals as one should and as I suspected in https://github.com/deltachat/deltachat-desktop/pull/6076#discussion_r2952005004:

https://github.com/keepassxreboot/keepassxc/blob/e888fec0703d0fb2e00faecfc9fd40249809634c/src/gui/osutils/nixutils/NixUtils.cpp#L146-L212

But doing thing this way (as in this MR) is probably still better than specifying a path.

packages/target-electron/src/autostart.ts
```diff
@@ -34,9 +34,29 @@ function getLinuxAutostartFilePath(): string {
...
 }
 
-function getLinuxExecPath(): string {
-  // When running as an AppImage, APPIMAGE env var is set to the .AppImage file path
-  return process.env.APPIMAGE || process.execPath
+// Returns the command for the Exec= key of the autostart desktop entry,
+// without the trailing app arguments (`-- --minimized`).
+function getLinuxExecCommand(): string {
+  if (process.env.FLATPAK_ID) {
+    return `flatpak run ${escapeDesktopExecArg(process.env.FLATPAK_ID)}`
```
WofWca: Minor: I know this attempts to preserve the behavior prior to this MR, but I suspect that the "escape" function is not needed especially here. See https://github.com/deltachat/deltachat-desktop/pull/6076#discussion_r2952414938.

packages/target-electron/src/autostart.ts
```diff
@@ -34,9 +34,29 @@ function getLinuxAutostartFilePath(): string {
...
+    // .AppImage file path
+    return escapeDesktopExecArg(process.env.APPIMAGE)
+  }
+  if (process.execPath.startsWith('/nix/store/')) {
+    // On Nix/NixOS, process.execPath is a version-pinned path to the shared
+    // `electron` binary (e.g. /nix/store/…-electron-…/…/electron), not to
+    // Delta Chat. Pinning it would break autostart on every update so rely
+    // on the `deltachat` command
+    // see https://github.com/deltachat/deltachat-desktop/issues/6431
+    return 'deltachat'
```
WofWca: Did you check that this is the right string? As opposed to something like `deltachat-desktop`. Searching for `(?<!@)deltachat(?!-)(?!_)(?!/)(?!\.)` in  our repo doesn't yield anything that looks related to the executable name.

packages/target-electron/src/autostart.ts
```diff
@@ -72,7 +92,9 @@ function getLinuxDesktopFileContent(): string {
 Type=Application
 Name=Delta Chat
 Comment=Delta Chat decentralized private messenger
-Exec=${escapeDesktopExecArg(getLinuxExecPath())} -- --minimized
+Exec=${getLinuxExecCommand()} -- --minimized
+Terminal=false
+StartupNotify=false
```
WofWca: Maybe it's just me, but what does `StartupNotify=false` do? Here are the docs https://specifications.freedesktop.org/startup-notification/latest/.

>   If false, it is KNOWN that the application does not work
  with startup notification at all (does not shown any window, breaks
  even when using StartupWMClass, etc.).

Does this apply to us?

WofWca (commented):
The general direction looks good, as far as I can understand.

`Terminal=false` also makes sense

packages/target-electron/src/autostart.ts
nicodh: This and Terminal=false were recommendations in the Issue description  I at least expected not to harm. But looking into the docs I assume we can remove them (although the keepassxc desktop file you referred to in your other comments has them)

packages/target-electron/src/autostart.ts
nicodh: Hmm, but what you suggested is the way we do it now or what do you mean?

WofWca: The proper way would be using portals I think. But this MR improves the situation.

WofWca: > but what you suggested

You mean what I suggested in this comment? We're not using portals, so no. Or am I missing something?

WofWca (approved):
Didn't test, but looks good.

packages/target-electron/src/autostart.ts
WofWca: Actually, `deltachat` might have been correct after all:

https://github.com/NixOS/nixpkgs/blob/04607e1165ac22c5fde6dcc54c9e0b3c0487c555/pkgs/by-name/de/deltachat-desktop/package.nix#L161

Edit: but I'm not sure.

WofWca (changes requested):
Revoking the approval, see https://github.com/deltachat/deltachat-desktop/pull/6571#discussion_r3721722918.

packages/target-electron/src/autostart.ts
WofWca: Maybe @dotlambda could confirm?

nicodh: So my initial assumption was right.

I was a bit confused since on my deb install the command is _deltachat-desktop_

WofWca (approved):
I'm still not 100% sure if `deltachat` as the command is right, but let's try I guess.

PR #6572, opened 2026-07-22
feat: group last used apps visually

### LINKED ISSUE #6531: Task: group recent apps visually to separate them from other icons
> ftr, what i wanted to have, but it was never a prio, is to put the app icons and the "all apps" also visually more together. but again, to avoid confusion. it one really accepts there is an reasonable attack vector, that is not sufficient. one would like to have a title and whatnot then.
> 
> Image
> 
> (the grouping by space would be nice unconditionally, the border, already not so sure, maybe okay, maybe it will look not optimal if there are no app, so one might want to hide it in that case, which then adds more complexity etc.)
> 
>  

 _Originally posted by @r10s in [#6530](https://github.com/deltachat/deltachat-desktop/issues/6530#issuecomment-4869444008)_

nicodh:
resolves #6531 

image

### DIFF
--- a/packages/frontend/src/components/ChatView/styles.module.scss
+++ b/packages/frontend/src/components/ChatView/styles.module.scss
@@ -73,6 +73,21 @@
       }
     }
 
+    .appsGroup {
+      display: flex;
+      align-items: center;
+      gap: 10px;
+      margin-inline-end: 24px;
+
+      border: 1px solid #999;
+      border-radius: 4px;
+      padding: 3px 4px;
+
+      .navbarButton {
+        margin-inline-end: 0;
+      }
+    }
+
     .messageMultiselectCounter {
       font-size: large;
       font-weight: bold;
@@ -84,7 +99,6 @@
 
 .webxdcIcons {
   line-height: 0;
-  margin-right: 12px;
   img.webxdcIcon {
     width: 25px;
     height: 25px;
--- a/packages/frontend/src/components/ChatView/styles.module.scss
+++ b/packages/frontend/src/components/ChatView/styles.module.scss
@@ -62,8 +62,8 @@
 
     .navbarButton {
       padding: 0;
-      width: 20px;
-      height: 20px;
+      width: 22px;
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/ChatView/styles.module.scss
```diff
@@ -73,6 +73,21 @@
...
 
+    .appsGroup {
+      display: flex;
+      align-items: center;
+      gap: 10px;
+      margin-inline-end: 24px;
+
+      border: 1px solid #999;
+      border-radius: 4px;
+      padding: 3px 4px;
```
WofWca: Symmetric padding should look better.

```suggestion
      padding: 3px;
```

WofWca (approved):
I think it looks ugly, but this implements the agreed upon solution.
The code looks good.

But note that this does nothing screen-reader-wise. If one wishes to fix that, consider making a `<section>` with `aria-label`.

r10s (changes requested):
> I think it looks ugly

yeah, the look is pretty dated because of the border.

let's get rid of that and only group by space, as it is done in these modern times. the border does not add much for the grouping but make it look dated

to make it appear even more like a group, i would additionally make the app icons a _little bit_ smaller and the "others icon" a _little bit_ larger. talking about 1~2 pixels only. 

finally, the space can also be there is there are no apps, so we always have two groups of icons. and switching between chats does not make the "others" icon jump that much (still a little bit as 📞 is not always present, but that is how it is then)

nicodh:
I removed the border, aligned the size of the icons and made the distance between app icons and map static.

@r10s : now all icons map, phone etc. are bigger than before, I'm not sure if that is what you intended. 

image

WofWca (approved):
Doesn't look ugly to me anymore.
The code also looks good.
Enlarging the icons is also good, they were quite hard to click (due to the lack of padding mostly).

r10s (approved):
looks great! thanks a lot!

PR #6573, opened 2026-07-22
chore: remove version in minimumReleaseAgeExclude

WofWca:
So that we don't have to bump these each time.
The policy applies to all versions.

See docs:
https://pnpm.io/settings#minimumreleaseageexclude


nicodh (approved):
Good finding!

PR #6576, opened 2026-07-22
update onboard help

r10s:
ran `./bin/help/create-local-help.sh`


WofWca (approved):
Didn't check, but this only affects the `static/help` dir.

PR #6580, opened 2026-07-23
fix: QR scanner camera select not working

### LINKED ISSUE #6540: Cannot switch to second camera for scanning QR code
- Operating System (Linux/Mac/Windows/iOS/Android): Linux Mint 22.2
- Delta Chat Version: v2.53.1 (installed via Flatpak)
- Expected behavior:
  - I have installed DC on a laptop, and after installation I selected "Add as second device" and saw the camera image for scanning the QR code. This was the image from built-in camera, but I wanted to use the image from an external USB camera (which I can move around more easily). The gear menu shows the second camera (nice!), but when selecting this entry there is a short spinner on grey screen, and then the built-in camera is used again. I have already restarted DC (exited it via tray icon), but that did not help.
  - btw. after plugging in the external camera I had to re-open the whole dialog to make the camera appear in gear menu. Not a big problem, but maybe difficult to discover.
- Actual behavior: see above.
  - use a laptop with a built-in camera.
  - attach a USB camera to the laptop.
  - install DC via Flatpak.
  - start DC and pick "Ich habe bereits ein Profil" (I have a profile).
  - pick "Als Zweitgerät hinzufügen".
  - open gear menu in lower right corner.
    - there should be two cameras listed, and the first one should have a checkmark.
  - pick the second camera.
  - desired behavior: video images from second camera are shown.
  - current behavior: video images from first camera are shown.
- Screenshots:
- Logs:

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/6540.

Turned out it was not as straightforward as removing `facingMode`.

### DIFF
--- a/packages/frontend/src/components/QrReader/index.tsx
+++ b/packages/frontend/src/components/QrReader/index.tsx
@@ -325,11 +325,6 @@ export const QrReader = forwardRef<QrCodeScanRef, Props>(
       }
 
       const startStream = async () => {
-        const videoConstraints: MediaTrackConstraints = {
-          deviceId,
-          facingMode: 'user',
-        }
-
         try {
           setReady(false)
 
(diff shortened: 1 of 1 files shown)


nicodh (approved):
Reviewed by reading & testing

PR #6585, opened 2026-07-23
feat: add (experimental) "Team Profiles"

### LINKED ISSUE #6520: allow to create experimental team profiles from UI
since https://github.com/chatmail/core/pull/7694 we have a flag to create so called "team profiles". 

as discussed with @hpk42 and @nicodh, we want to expose that to the UI, as a simple option with the name `create_team_profile` at "Add Profile / Create New Profile / Three dot menu" below "Use Proxy".

once the option is tapped, the dialog is update using the title `team_profile` and the hint `team_profile_hint`. the option "Create Team Profile" is gone then - if one changes their mind, one just hits back and "Create New Profile" again.

in the future, there might be more to tweak, but that is a good start.

the following strings should go to "untranslated":

```
create_team_profile = "Create Team Profile"
team_profile = "Team Profile (experimental)"
team_profile_hint = "Team profiles let multiple team members collaborate on messages and contacts together."
```

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/6520.
The Core config was introduced in
https://github.com/chatmail/core/pull/7694.
Reference implementation in Android:
https://github.com/deltachat/deltachat-android/pull/4543.

[team2.webm](https://github.com/user-attachments/assets/932acc55-8145-4507-a245-2cfb6c251cef)

image

image

image

### DIFF
--- a/_locales/_untranslated_en.json
+++ b/_locales/_untranslated_en.json
@@ -82,5 +82,20 @@
   },
   "webxdc_store_url_explain_2_desktop": {
     "message": "This is different from the same feature on Android and iOS.\nThis must be the path containing the xdcget-lock.json file."
+  },
+  "team": {
+    "message": "Team"
+  },
+  "create_team_profile": {
+    "message": "Create Team Profile"
+  },
+  "team_name": {
+    "message": "Team Name"
+  },
+  "team_profile": {
+    "message": "Team Profile (experimental)"
+  },
+  "team_profile_explain": {
+    "message": "A \"Team Profile\" is managed collectively by a group of people or an organization.\n\n⚠️ Team Profiles are experimental and may be changed or removed in the future."
   }
 }
--- a/packages/frontend/src/components/screens/WelcomeScreen/InstantOnboardingScreen.tsx
+++ b/packages/frontend/src/components/screens/WelcomeScreen/InstantOnboardingScreen.tsx
@@ -53,11 +53,11 @@ function buildContextMenu(
     isTeamProfile
       ? {
           // TODO fix a11y: the menu item needs to be `role="menuitemcheckbox"`.
-          label: '🗹 ' + tx('create_team_profile'),
+          label: '✓ ' + tx('create_team_profile'),
           action: () => setIsTeamProfile(false),
         }
       : {
-          label: '☐ ' + tx('create_team_profile'),
+          label: tx('create_team_profile'),
           action: () => {
             openDialog(ConfirmationDialog, {
               header: tx('create_team_profile'),
--- a/packages/frontend/src/components/screens/WelcomeScreen/InstantOnboardingScreen.tsx
+++ b/packages/frontend/src/components/screens/WelcomeScreen/InstantOnboardingScreen.tsx
@@ -24,6 +24,9 @@ import { OpenDialog } from '../../../contexts/DialogContext'
 import ProxyConfiguration from '../../dialogs/ProxyConfiguration'
 import { selectedAccountId } from '../../../ScreenController'
 import { TranslationKey } from '@deltachat-desktop/shared/translationKeyType'
+import SettingsStoreInstance from '../../../stores/settings'
+import type { ContextMenuItem } from '../../ContextMenu'
+import ConfirmationDialog from '../../dialogs/ConfirmationDialog'
 
 type Props = {
   onCancel: () => void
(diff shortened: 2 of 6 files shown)


_locales/_untranslated_en.json
```diff
@@ -82,5 +82,20 @@
...
+    "message": "Create Team Profile"
+  },
+  "team_name": {
+    "message": "Team Name"
+  },
+  "team_profile": {
+    "message": "Team Profile (experimental)"
+  },
+  "team_profile_explain": {
+    "message": "A \"Team Profile\" is managed collectively by a group of people or an organization.\n\n⚠️ Team Profiles are experimental and may be changed or removed in the future."
```
r10s, line 99: when pulling strings, these strings should be available without using `_untranslated_en.json`.

usually, we  pull strings in another PR (if not yet done because of release) and rebase this PR

WofWca: We haven't pulled them yet. The pulled strings will take priority over these ones

r10s:
very cool! thanks a lot for the PR!

some notes:

- the settings should already show, that this is a team profile, this is how it looks on android/iOS, both is fine for now, whatever is easier (but the non-team-profile view must not change):<br> 


- for ppl playing around, there should be an easier way to undo team profile creation:
   - show "Create Team Profile" menu item unconditionally
   - when in team profile mode, prepend the typical menu-like checkbox "✔ Create Team Profile" (are we using that elsewhere already? if so, use that style)
   - tapping the option again, will disable team profile mode

- also for ppl playing around, we want to show an alert before the team profile mode is actually enabled, that gives more focus  than having the text in the dialog among other things (which should stay as is, this is very nice!)
   - this is how that looks on android (and similar on ios):<br>

WofWca:
Force-pushed and edited MR description with new vid and screenshots.

The only point that is not addressed is the use of a checkbox. Our context menu lib doesn't have that.

r10s:
> The only point that is not addressed is the use of a checkbox. Our context menu lib doesn't have that.

i see, np. we should find a simple alternative then.

1. either concatenate a checkmark to the menu entry when selected ("✔ Create Team Profile" or "Create Team Profile ✔") (iOS uses the  checkmark, btw, but there it is supported by the menu system)

2. or, leave menu entry "Create Team Profile" as is, but show the same dialog again but with the button "Disable" instead of "Create Team Profile" (this is done on android, which does not have a checkmark in the menu system as well)

both are good enough - i think 1. is better and easier. but the existing "Back" in the menu is too confusing, as it lacks context:

Screenshot 2026-07-24 at 13 12 06</details>

WofWca:
> either concatenate a checkmark to the menu entry when selected

This is not good accessibility-wise because the menu item has to have the proper `menuitemcheckbox` role.

> 2

Sounds much more confusing than the supposedly context-less "Back". I'll choose 1 over it.

WofWca:
Also, to clarify, it's already possible to cancel team profile creation by pressing the existing "back" button on the left, or pressing Escape (closes the whole dialog).

r10s:
let's for for 1 then, "concatenate a checkmark to the menu entry when selected", this is also good enough accessibility wise.

and yes, i know, one can cancel by leaving the dialog and re-enter etc. but that is even worse, esp. on first app start

WofWca:
Force-pushed, uploaded a new video.

packages/frontend/src/components/screens/WelcomeScreen/InstantOnboardingScreen.tsx
```diff
@@ -45,6 +50,28 @@ function buildContextMenu(
...
       dataTestid: 'proxy-context-menu-item',
     },
+    isTeamProfile
+      ? {
+          // TODO fix a11y: the menu item needs to be `role="menuitemcheckbox"`.
+          label: '🗹 ' + tx('create_team_profile'),
+          action: () => setIsTeamProfile(false),
+        }
+      : {
+          label: '☐ ' + tx('create_team_profile'),
```
r10s: let's keep it simple here, see above

```suggestion
          label: '✓ ' + tx('create_team_profile'),
          action: () => setIsTeamProfile(false),
        }
      : {
          label: tx('create_team_profile'),
```

r10s (commented):
let's use the more standard checkbox i suggested, otherwise lgtm, thanks a lot.

background: the used characters (square with/without checkbox)  makes problems on macos or electron or font, but also in android studio and safari ...

also, when the item is _not enabled_, do not show a placeholder or so. only "Create Team Profile". we do not want raise questions for the many that will never use that

packages/frontend/src/components/screens/WelcomeScreen/InstantOnboardingScreen.tsx
WofWca: Feel free to apply, code-wise looks OK.

r10s (approved):
so, all looks good to me now :)

nicodh:
Any reason why this get's not merged?

WofWca:
Waiting for your approval

nicodh:
> Waiting for your approval

Oops - oversaw tat the approval is not from me 😅

nicodh (approved):

PR #6586, opened 2026-07-23
refactor: clarify that `useSecureJoin` is just fn

WofWca:
The fact that it's a hook makes it seem like it probably has state
and other returned values, whereas it's just one function.


nicodh (approved):
Reviewed by reading

PR #6589, opened 2026-07-24
chore: remove `bin/check-nodejs-version.js`

WofWca:
We have the `engines` field package.json, let's use that.


nicodh (approved):
Reviewed by reading

PR #6590, opened 2026-07-24
refactor: print stack trace on `useFetch` err

WofWca:
This has been brought up in
https://github.com/deltachat/deltachat-desktop/pull/5291#pullrequestreview-3019428286.
Turns out the stack is indeed somewhat useful after all,
at least in the dev tools (but not the log file):

image


nicodh (approved):
Reviewed by reading

PR #6591, opened 2026-07-24
improvement: translate some hard-coded strings

WofWca:


packages/frontend/src/components/message/MessageList.tsx
```diff
@@ -1019,7 +1020,7 @@ function MessageLoading({
         onKeyDown={focusAndMultiselect.onKeyDown}
         onFocus={focusAndMultiselect.onFocus}
       >
-        Loading Message {messageId.msg_id}
+        {tx('loading')} Message {messageId.msg_id}
```
r10s, line 1023: when is this shown?

- if it is user visible information, there should not be an id that is meaningless to the user. in this case use `tx('loading')` only

- if it is an error state, the existing english one good enough, for searching for errors even better

WofWca: This rarely shows up in the message list when the we erroneously render a message that has not loaded yet.

> there should not be an id that is meaningless to the user.

I'll make an MR to address this. Edit: https://github.com/deltachat/deltachat-desktop/pull/6592.

r10s (approved):

nicodh (approved):
Reviewed by reading

PR #6592, opened 2026-07-24
improvement: hide DB id of loading messages

WofWca:
Normally we should not render messages that are still loading,
but if it does happen, let's not show the ugly internals.
Still, log the message ID.

This also removes yet another untranslated string.

Addresses https://github.com/deltachat/deltachat-desktop/pull/6591#discussion_r3646092744.


r10s (approved):

nicodh (approved):
Reviewed by reading

PR #6594, opened 2026-07-25
chore: upgrade core to 2.57.0

nicodh:


WofWca (approved):
Just an upgrade (with an important fix)

(FTR I also noticed that I forgot to change Cargo.toml - the tauri-plugin-log get auto-bumped) - I'll make an MR.

PR #6603, opened 2026-08-02
fix: audio player overflow small screens

account1009:
The problem with the audio player on small screens is that:
- Message sent by others have the controls over the emoticons reaction icon so it is impossible to click that icon
- Message sent by myself add an horizontal scrollbar

The suggested pull request is to solve both issues :)

Enrico


WofWca (commented):
Why do these have to be in different files, I don't see it?

nicodh:
Would be nice to add some before-after screenshots

account1009:
@WofWca I respected the definitions on the original files

@nicodh yes I can share the before-after screenshots

Before Sender Point of View:
before_sender_point_of_view

Before Receiver Point of View:
before_receiver_point_of_view

After Sender Point of View:
after_sender_point_of_view

After Receiver Point of View:
after_receiver_point_of_view

packages/frontend/scss/_global.scss
```diff
@@ -216,6 +216,12 @@ audio {
   width: 300px;
 }
```
WofWca, line 218:
```suggestion
```

WofWca (approved):
Looking at your screenshots, it still overflows a bit. I think `max-width: 100%;` would do the job, and also wouldn't shrink the players from 300 to 255 if the screen is small but not as small as yours.
But I guess that can be done in follow-ups.
Thanks!

WofWca:
Regarding `100%`, FTR I just found this comment:

https://github.com/deltachat/deltachat-desktop/blob/aa07f3902e7f642935b9513908028a441fa8ec84/packages/frontend/src/components/AudioPlayer/styles.module.scss#L3

account1009:
@WofWca @nicodh do you have any news about this pull request? is it fine to merge it?

WofWca:
No news.
Fixing CI will move things forward a step.

account1009:
@WofWca now the prettier related thing is solved and CI is fine.

But I was waiting because "approved these changes" was already marked by you long time ago.

For a similar pull request I had to ask to @nicodh the merge

I don't know what is the procedure but I've written again because one month was passed :)

WofWca:
The procedure is to get a review from @nicodh

nicodh:
Thanks for contributing!

PR #6604, opened 2026-08-02
fix: camera now working on pipewire based linux distributions

account1009:
The camera was not working for Linux distributions that are using Pipewire

As a concrete case for example the camera was not working on Mobian and now I have the camera finally working on Deltachat :)

This conversation was useful to detect the cause of the issue: https://github.com/electron/electron/issues/45058

Enrico

### DIFF
--- a/packages/target-electron/src/index.ts
+++ b/packages/target-electron/src/index.ts
@@ -28,7 +28,10 @@ rawApp.commandLine.appendSwitch('host-resolver-rules', hostRules)
 
 rawApp.commandLine.appendSwitch('disable-features', 'IsolateSandboxedIframes')
 
-rawApp.commandLine.appendSwitch('enable-features', 'WebRtcPipeWireCamera,WebRtcPipeWireScreenCapturer')
+rawApp.commandLine.appendSwitch(
+  'enable-features',
+  'WebRtcPipeWireCamera,WebRtcPipeWireScreenCapturer'
+)
 
 if (rc['version'] === true || rc['v'] === true) {
   // eslint-disable-next-line no-console
--- a/packages/target-electron/src/index.ts
+++ b/packages/target-electron/src/index.ts
@@ -28,6 +28,8 @@ rawApp.commandLine.appendSwitch('host-resolver-rules', hostRules)
 
 rawApp.commandLine.appendSwitch('disable-features', 'IsolateSandboxedIframes')
 
+rawApp.commandLine.appendSwitch('enable-features', 'WebRtcPipeWireCamera,WebRtcPipeWireScreenCapturer')
+
 if (rc['version'] === true || rc['v'] === true) {
   // eslint-disable-next-line no-console
   console.info(BuildInfo.VERSION)


packages/target-electron/src/index.ts
```diff
@@ -28,6 +28,8 @@ rawApp.commandLine.appendSwitch('host-resolver-rules', hostRules)
 
 rawApp.commandLine.appendSwitch('disable-features', 'IsolateSandboxedIframes')
 
+rawApp.commandLine.appendSwitch('enable-features', 'WebRtcPipeWireCamera,WebRtcPipeWireScreenCapturer')
```
WofWca: Someone will need to run `pnpm fix:format`.

WofWca (commented):
Thanks for the MR!

Do you have an idea when these flags are going to get enabled by default? I would like to not fiddle with switches, and would prefer to just wait for them to stabilize upstream.

account1009:
@WofWca I think that we will have to wait many years before chromium remove Pipewire from experimental features because it looks that they don't care at all about Linux XD. They have the priority about fixing stuff on Chromium for ChromeOS ignoring Linux. Sorry but I'm upset about Chromium that added for example kinetic scroll for ChromeOS as the first feature (while it is less used than Linux) and the same kinetic feature after 10 years is still not reality for Linux :/

@nicodh I also add here the before-after screenshots to give an idea about the fix:

Before:
before

After:
after

Enrico

WofWca:
I'm sorry but I'm not convinced that this is a good enough reason to override Chromium's decision to have this behind a flag. Maybe better to push the decision upstream, all Linux Chromium users would be thankful.

account1009:
@WofWca I've checked the Chromium code and based on that metadata in this file
https://chromium.googlesource.com/chromium/src/+/main/chrome/browser/flag-metadata.json 

the webrtc-pipewire-camera will be enabled by default with chromium 160 version (so I think around December 2026)

This is the current metadata with the release future release version:
{
"name": "enable-webrtc-pipewire-camera",
"expiry_milestone": 160
}

The problem is that the release date continue to be postponed such as in this case: https://chromium-review.googlesource.com/c/chromium/src/+/6668416

I've not checked the date of the most ancient postpone because for now I've not cloned the Chromium repo

The metadata contains the owners emails for that webrtc-pipewire-camera part (3 mails) and there is also @markafoltz than is here on github. Maybe he can you help with the upstream process.

@markafoltz Why is the release date of the webrtc-pipewire-camera postponed year after year?

Do you need any help about the missing "any test coverage yet" mentioned as the cause of the postpone of 2025?
What do you suggest to unlock this release goal? Do you need specific tests upstream to verify that everything is working fine?
Do you have any idea to materialize step by step this goal?

Enrico

account1009:
@WofWca @nicodh  do you have any news about this pull request?
If we wait upstream solution (that can require months of years) can we can momentary use a downstream solution on deltachat to have camera working in the meanwhile?
If this temporary downstream approach is not fine could I activate to add patches on linux distribution level (such as in Arch, Alpine, etc) to fix the camera in the meanwhile?

Enrico

markafoltz:
1.  I'm not personally responsible for shipping this feature in Chromium; the other person listed as a contact in the flag  (Jan Grulich) would be the better point of contact.
2. I gather that this is an electron application.  The application developer should be able to enable this flag by default using electron APIs.
3. We won't remove the flag from Chromium as long as the feature is being maintained and has some path to shipping for Chrome users.

WofWca (approved):
Alright, a counter-argument would be that there wouldn't be a flag for this feature if it wasn't at least moderately good.
I think it's fine to go for this.

nicodh:
Thank you for your contribution!

PR #6607, opened 2026-08-04
chore: remove unused code

nicodh:
Most of the obsolete exports & functions were detected using knip https://knip.dev

### DIFF
--- a/packages/frontend/src/components/AudioPlayer/index.tsx
+++ b/packages/frontend/src/components/AudioPlayer/index.tsx
@@ -1,7 +1,5 @@
 import React, { useContext, useEffect, useRef } from 'react'
 
-import styles from './styles.module.scss'
-import classNames from 'classnames'
 import { getLogger } from '@deltachat-desktop/shared/logger'
 import { ForceMutedAudioPlayer } from './ForceMutedAudioPlayer'
 import {
@@ -111,23 +109,6 @@ export function AudioPlayer({
         }
         ignoreOnePauseEvent.current = false
       }}
-    />
-  )
-}
-
-export function AudioPlayerNonMutex({
-  ref,
-  className,
-  ...restProps
-}: Props & {
-  ref?: React.RefObject<HTMLAudioElement | null>
-}) {
-  return (
-    <audio
-      ref={ref}
-      controls
-      className={classNames(styles.audioPlayer, className)}
-      {...restProps}
     />
   )
 }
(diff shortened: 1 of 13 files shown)


packages/frontend/src/components/helpers/ClickableLink.tsx
```diff
@@ -28,26 +26,3 @@ export function ClickableLink({
     </a>
   )
 }
-
-/**
- * Same as {@link ClickableLink}, but does not require that
- * an account is selected, but the user of this component guarantees
- * that `href` is not a `mailto:` link.
- */
-export function ClickableNonMailtoLink({
```
WofWca, line 37: I'd keep this.

packages/frontend/src/components/AudioPlayer/index.tsx
```diff
@@ -114,20 +112,3 @@ export function AudioPlayer({
     />
   )
 }
-
-export function AudioPlayerNonMutex({
```
WofWca, line 118: This was unused from the very commit that introduced it. I think it's useful at least as documentation.

If the concern is "let's not maintain dead code", I suggest to simply add a comment that says that this is unused.

WofWca (commented):
`setBounds` and similar `main.ts` functions were introduced in c3957ccfbb0ae177fbe1f07a1f3ad96ba3027c2f, and TBH I haven't figured out if it was ever used.

I'd keep some. Removing the rest seems OK. Especially nice to see those one-line wrappers in `/backend` gone.

nicodh:
I don't see why we should keep unused code? You may add comments for documentation but why should it need code?
Especially the stuff in main - was introduced 6 years ago and apparently never used

`git log -S"mainWindow.setBounds" --all
`
or go through 
`git log -p -S".setBounds" --all`

to make sure.

I f you think the removal leads to missing comments/docs a followup would be welcome

WofWca:
> You may add comments for documentation but why should it need code?

Such unused code is not fundamentally different from comments.

> I f you think the removal leads to missing comments/docs a followup would be welcome

I don't know how to do it differently from what I suggested, so I guess that means that you insist.

WofWca (approved):

nicodh:
> Such unused code is not fundamentally different from comments.

I strongly disagree - code that is not used is always confusing, while noone would ever start reasoning why a  comment  exists

> I don't know how to do it differently from what I suggested, so I guess that means that you insist.

Add a comment instead of dead code

PR #6608, opened 2026-08-04
chore: remove unused functions

nicodh:
They were replaced during the time

### DIFF
--- a/packages/frontend/src/hooks/usePrevious.ts
+++ b/packages/frontend/src/hooks/usePrevious.ts
@@ -17,23 +17,3 @@ export function usePrevious(value: any) {
   // eslint-disable-next-line react-hooks/refs
   return ref.current
 }
-
-/**
- * Like {@link usePrevious}, but the returned value is from the previous
- * execution of this hook (i.e. previous render function execution),
- * and not from the previous actual render (not `useEffect`).
- *
- * It is this pattern:
- * https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes
- *
- * @returns `undefined` on initial run,
- * otherwise the `val` from the previous run of the hook.
- */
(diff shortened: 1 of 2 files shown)


packages/frontend/src/hooks/usePrevious.ts
```diff
@@ -17,23 +17,3 @@ export function usePrevious(value: any) {
...
- * execution of this hook (i.e. previous render function execution),
- * and not from the previous actual render (not `useEffect`).
- *
- * It is this pattern:
- * https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes
- *
- * @returns `undefined` on initial run,
- * otherwise the `val` from the previous run of the hook.
- */
-export function usePrevious2<T>(val: T): T | undefined {
```
WofWca, line 32: IDK about this. In a perfect code base we'd want to get rid of both `useHasChanged` and `useHasChanged2`, but I consider the latter a "better" one than `useHasChanged`, because it doesn't cause two re-renders (one original and one after `useEffect`).
So first I would replace all `useHasChanged` usages with `useHasChanged2` and then remove `useHasChanged`.

Again, to compromise, I suggest to add a comment that it's unused instead.

WofWca (commented):
`usePrevious2` I'd keep. The rest should be removed.

packages/frontend/src/hooks/usePrevious.ts
nicodh: Why are you referring to useHasChanged and useHasChanged2 here? Both don't use usePrevious2. If you would replace all usage of useHasChanged we could also remove usePrevious but that does not mean that the removal of usePrevious2 should not be removed now.

WofWca: Ah. I might have mixed them up.

WofWca (approved):

PR #6610, opened 2026-08-05
test: CI: replace a bash script with CLI switch

WofWca:
https://docs.docker.com/reference/cli/docker/compose/up/

> --wait		Wait for services to be running|healthy. Implies detached mode.

I'm not sure if `running|healthy` means "running and healthy" or "running or healthy", but in the commit history of https://github.com/deltachat/deltachat-desktop/pull/6548 I don't see if "wait for running without waiting for healthy" was tried - the very first commit introduced that bash script. And as we can see in this MR's CI, it works fine.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -70,29 +70,7 @@ jobs:
         run: |
           printf 'MAIL_DOMAIN=127.0.0.1\n' > .env
           docker pull "$CHATMAIL_IMAGE"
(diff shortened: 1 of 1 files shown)


nicodh:
I removed the bash script as suggested in this PR but at the same time changed the order a bit (starting the build and then wait for the container to be healthy) so this PR is not needed anymore
see 6c8472b69d3fa0956d

PR #6611, opened 2026-08-05
test: CI E2E: remove useless Docker repo checkout

WofWca:
We use the prebuild `ghcr.io/chatmail/docker:main` a few steps below.


WofWca:
OK we can't remove it, because we need the `docker-compose.yml` file I think.

PR #6618, opened 2026-08-06
refactor: remove fallback composer resizing code

WofWca:
`resizeTextareaAndComposer` is already unused in Electron.

All modern browsers support `field-sizing`,
including Firefox and Safari,
which was not the case back when we introduced the new approach:
de05e38515039feaf940295fc366066ed8baa4c8
(https://github.com/deltachat/deltachat-desktop/pull/5637).

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/field-sizing

I have tested this on Firefox but not Safari.

### DIFF
--- a/packages/frontend/scss/composer/_composer.scss
+++ b/packages/frontend/scss/composer/_composer.scss
@@ -154,16 +154,12 @@
       line-height: $line-height;
       margin-top: 8px;
       margin-bottom: 8px;
-      overflow-y: hidden;
       background-color: var(--composerBg);
       color: var(--composerText);
 
-      &.use-field-sizing-css-prop {
-        field-sizing: content;
-        max-height: calc(var(--maxLines) * $line-height);
-
-        overflow-y: auto;
-      }
+      field-sizing: content;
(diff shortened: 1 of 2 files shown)


WofWca:
We can consider postponing this until October, when Firefox ESR 153 is released which supports `field-sizing`:
https://whattrainisitnow.com/calendar/

nicodh:
> We can consider postponing this until October, when Firefox ESR 153 is released which supports `field-sizing`: https://whattrainisitnow.com/calendar/

We don't need to support Firefox so why should we wait?

WofWca:
There are some Delta Chat clients that run in a browser.
Of course we don't officially support them, but since this PR is only about removing some code, it can wait.

nicodh:
> ...but since this PR is only about removing some code, it can wait.

Removing code is always nice!! 😁

But ok, I'm fine with waiting some weeks...

nicodh (approved):
Reviewed by reading & testing

Good catch!

Issue #2722, opened 2022-04-29
make it possible to prepare webxdc while in draft


Simon-Laux:
useful for webxdc content that should be configured before sending like a poll.

Simon-Laux:
- [ ] hint for later: make sure that the webxdc window is closed when you remove the draft (I think easiest way would be to ensure that message deletion event is emitted when draft is cleared/removed - this might already be the case, but needs to be checked because it might be an issue)

WofWca:
- [x] The draft view will also need to be updated, as in https://github.com/deltachat/deltachat-desktop/issues/2871.

WofWca:
Implementation details: we will probably have to listen on "message changed" events to update the "summary" and the title.

r10s:
the title does not change, and the summary should be set to 'Tap "Send" to share' (`webxdc_draft_hint`) as ppl are otherwise [confused about if the xdc is actually sent or not](https://github.com/deltachat/deltachat-android/pull/3774). but even if not, there is not really a need to show summary, document name etc., there is only one user that knows about changes - so even if we have more space on desktop, i would not show it in draft mode to have a clear, uncluttered UI


<img width=320 src=https://github.com/user-attachments/assets/be0a2c96-c887-41bc-b2b0-92f67823de05>

PR #6624, opened 2026-08-07
fix: Ctrl + Up: incorrect quote order sometimes

WofWca:
The code assumes that `messageListItems` is always sorted by ID,
but it's not the case, i.e. in reality sometimes
a message with a lower ID occurs _after_ a message with a higher ID.
This is why we have `timestamp_sort` on messages:
https://github.com/chatmail/core/blob/b351b64e4ae78658f894f51412f89aca4e6ce822/src/chat.rs#L3195-L3202.

This is gonna be especially noticeable if/when
we start preserving draft message ID
(see https://github.com/deltachat/deltachat-desktop/issues/2722):
if you start drafting a message and receive one in the chat,
the received message will have a higher ID than the draft;
when the draft is finally sent,
it will occur _after_ the received message.


nicodh (approved):
Reviewed by reading

PR #6625, opened 2026-08-07
perf: Ctrl + Up: don't iterate over whole msg list

WofWca:
Remove the `messageListItems.filter()`.
Chat histories can contain thousands of messages.
Most of the time we actually need to only iterate once,
and sometimes skip one or two messages that can't be quoted.

This is basically a rewrite of the function using iterators.

### DIFF
--- a/packages/frontend/src/hooks/chat/useDraft.ts
+++ b/packages/frontend/src/hooks/chat/useDraft.ts
@@ -431,33 +431,53 @@ export function useDraft(
       })
     }
 
+    const step = upOrDown === KeybindAction.Composer_SelectReplyToUp ? -1 : +1
     /**
-     * filter all messageIds that can be replied to
+     * Iterate message list items up or down, starting from `fromInd`
+     * and skipping messages that definitely can't be replied to
      *
-     * see https://github.com/deltachat/deltachat-desktop/blob/77a1f88a351df49e5df38a14c3a1704a76ecdcb3/packages/frontend/src/components/message/Message.tsx#L274-L278
+     * Messages that are not loaded are not skipped,
+     * it's up to the caller how to handle them.
      */
-    const messageIds = messageListState.messageListItems
-      .filter((item): item is { kind: 'message'; msg_id: number } => {
-        if (item.kind !== 'message') {
-          return false
-        }
-        const id = item.msg_id
-
-        return (
-          messageListState.messageCache[id]?.kind === 'message' &&
-          messageListState.messageCache[id]?.isInfo === false
-        )
+    const makeIter = (fromInd: number) => {
+      return valuesFromInd({
+        arr: messageListState.messageListItems,
+        fromInd,
+        step,
       })
-      .map(({ msg_id }) => msg_id)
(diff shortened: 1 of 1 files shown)


packages/frontend/src/hooks/chat/useDraft.ts
```diff
@@ -431,33 +431,53 @@ export function useDraft(
...
-        if (item.kind !== 'message') {
-          return false
-        }
-        const id = item.msg_id
-
-        return (
-          messageListState.messageCache[id]?.kind === 'message' &&
-          messageListState.messageCache[id]?.isInfo === false
-        )
+    const makeIter = (fromInd: number) => {
```
nicodh:
```suggestion
    const quoteCandidatesFrom = (fromInd: number) => {
```
a more verbose name would be easier to read imo

packages/frontend/src/hooks/chat/useDraft.ts
```diff
@@ -721,3 +736,21 @@ function messageToQuote(
...
+ * Iterate from an arbitrary index, up or down.
+ */
+function* valuesFromInd<T>({
+  arr,
+  fromInd,
+  step,
+}: {
+  arr: T[]
+  fromInd: number
+  step: 1 | -1
```
nicodh:
```suggestion
  direction: 1 | -1
```
step reads as if it could be multiple

packages/frontend/src/hooks/chat/useDraft.ts
```diff
@@ -431,33 +431,53 @@ export function useDraft(
...
+        .map(msg_id => {
+          const fromCache = messageListState.messageCache[msg_id]
+          if (fromCache == undefined) {
+            return { kind: 'notLoaded', msg_id } as const
+          }
+
+          // see https://github.com/deltachat/deltachat-desktop/blob/77a1f88a351df49e5df38a14c3a1704a76ecdcb3/packages/frontend/src/components/message/Message.tsx#L274-L278
+          return fromCache.kind === 'message' && fromCache.isInfo === false
+            ? ({ kind: 'quotable', msg_id } as const)
+            : ({ kind: 'notQuotable', msg_id } as const)
```
nicodh, line 460: This reads a bit strange: we give it an explicit name just to filter it out in the next step? Couldn't we skip it immediately?

nicodh (approved):
IIUIC this also fixes some existing issues in the Ctrl + Up/Down behaviour so it should  be tagged as fix not perf

WofWca:
> IIUIC this also fixes some existing issues

Do you know which issues those are?

packages/frontend/src/hooks/chat/useDraft.ts
WofWca: I see your point, but then we would have to handle the "not loaded" case twice.

nicodh:
> > IIUIC this also fixes some existing issues
> 
> Do you know which issues those are?

Ok, I think I misunderstood how the loading was triggered before. So yes in behaviour there is no change

PR #6627, opened 2026-08-07
fix: "Edit" sometimes not moving cursor to the end

### LINKED ISSUE #5716: cursor placement in compose window for editing messages is unpredictable
On DeltaChat 2.22.0, when first editing a message, sometimes the text cursor is placed at the end of the compose window.  Other times it is placed at the beginning of the compose window.

It would be better if it were predictable.

From what i can tell, when i go from the normal compose window into the "Edit Message" view by choosing "Edit" from a message's context menu, the cursor is placed at the end.  But if i'm already in an "Edit Message" view, and then i choose "Edit" on a different message, the cursor gets placed at the beginning of the compose window.

I think it would be better if the cursor was always placed at he beginning of the compose window when choosing "Edit" on a message.

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/5716.

Let's move the cursor on the actual "enter edit message mode"
callback rather than by watching prop changes.

Also add a test for this. It doesn't cover the specific case
described in the issue, but at least it shows
that this functionality isn't completely broken.

### DIFF
--- a/packages/e2e-tests/tests/composer.spec.ts
+++ b/packages/e2e-tests/tests/composer.spec.ts
@@ -841,6 +841,20 @@ test.describe('edit message', () => {
     await page.keyboard.press('ArrowUp')
     await expect(textareaNonEdit()).toHaveText('\n\n\n\n\n')
     await expect(composerSection).not.toContainText('Edit Message')
+
+    await textareaNonEdit().clear()
+  })
+  test('text cursor is placed at the end', async () => {
+    await textarea.focus()
+    await page.keyboard.press('ArrowUp')
+    await expect(textareaEdit()).toHaveText('M 3')
+    await expect(textareaEdit()).toBeFocused()
+
+    await page.keyboard.press('Backspace')
+    await page.keyboard.press('Backspace')
+    await expect(textareaEdit()).toHaveText('M')
+
+    await page.keyboard.press('Escape')
   })
 })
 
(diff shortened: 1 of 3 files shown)


packages/e2e-tests/tests/composer.spec.ts
```diff
@@ -841,6 +841,20 @@ test.describe('edit message', () => {
     await page.keyboard.press('ArrowUp')
     await expect(textareaNonEdit()).toHaveText('\n\n\n\n\n')
     await expect(composerSection).not.toContainText('Edit Message')
+
+    await textareaNonEdit().clear()
+  })
+  test('text cursor is placed at the end', async () => {
+    await textarea.focus()
+    await page.keyboard.press('ArrowUp')
```
nicodh, line 849: I'm, not sure if that test covers the issue. When switching to edit mode by arrow up the cursor is always at the end. Only if I - while the edit is open - select "edit" on another messages context menu the cursor is (was) at the beginning of the text of that new messages text

nicodh (approved):
Reviewed by reading & testing

The test could/should be improved imo

packages/e2e-tests/tests/composer.spec.ts
WofWca: Yes, I mentioned this in the description. The test doesn't cover the specific issue that this MR is made to fix, but at least it proves that this MR doesn't break this feature overall.

nicodh: Ah - right. Read it once but started reviewing the files later and wasn't aware

Issue #6628, opened 2026-08-08
`example.com/fun_ftp_facts` link not clickable


WofWca:
- Operating System (Linux/Mac/Windows/iOS/Android): All
- Delta Chat Version: 2.57.0
- Expected behavior: Link recognized
- Actual behavior: Link not recognized
- Steps to reproduce the problem:
  1. Send `example.com/fun_ftp_facts` to a chat
- Screenshots:
- Logs:

From this report: https://support.delta.chat/t/react2shell-exploited-abused-on-deltachat/5624?u=wofwca.

Here URL construction fails, because the `ftp` part is recognized as `SLASH_SCHEME`.

https://github.com/deltachat/deltachat-desktop/blob/2357872a2115f91ede8bca5c0b1059941d855ea6/packages/frontend/src/components/message/MessageParser.tsx#L43-L50

Probably need to check only the first token? I didn't dive deep into the Linkify library docs.

r10s:
> Operating System (Linux/Mac/Windows/iOS/Android): All

ftr, android/iOS does not have the bug

WofWca:
> android/iOS does not have the bug

Our issue template is a bit weird then perhaps. We don't have iOS and Android builds of deltachat-desktop.

PR #6629, opened 2026-08-09
fix: display failed-to-parse URLs as text

WofWca:
Instead of `catch`ing the error one level above and falling back
to displaying the whole message's text as plain text.

Related: https://github.com/deltachat/deltachat-desktop/issues/6628.


nicodh (approved):
Reviewed by reading

Note: this improves the logging & error handling a bit, but doesn't fix the issue (comment is correctly saying "relates" instead "fixes")

PR #6630, opened 2026-08-09
refactor: `import 'linkifyjs'` -> `import type`

WofWca:


nicodh (approved):
Reviewed by reading

tbh I doubt if the benefit is worth an axplicit PR...

WofWca:
> tbh I doubt if the benefit is worth an axplicit PR...

This will save someone a few seconds the next time they'll need to find out where `linkifyjs` is actually used.

PR #6631, opened 2026-08-10
perf: message list: only load missing messages

WofWca:
...on incoming messages and on `__jumpToMessage`
(the latter also includes sending a message).
Originally we'd load all messages in `view`.

### DIFF
--- a/packages/frontend/src/stores/messagelist.ts
+++ b/packages/frontend/src/stores/messagelist.ts
@@ -546,7 +546,8 @@ export class MessageListStore extends Store<MessageListState> {
               this.accountId,
               messageListItems,
               oldestFetchedMessageListItemIndex,
-              newestFetchedMessageListItemIndex
+              newestFetchedMessageListItemIndex,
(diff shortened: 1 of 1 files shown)


nicodh (approved):
Reviewed by reading & testing (including some performance tests)

One big improvement is that we keep object identity in loadMessages  with 

`ret[id] = missing[id] ?? existingMessages[id]
`
which will prevent the MessageWrappers from rerendering messages that are in cache already

We really need unit tests for the messagelist store. There is quite some complexity and comments and use cases - hard to make sure that changes don't break existing stuff

WofWca:
> One big improvement is that we keep object identity in loadMessages with

Hm right, that's also a bit of a plus, I didn't think of it.

> We really need unit tests for the messagelist store.

I would prefer E2E tests. They'll also cover `MessageList.tsx`.
But both is good.

WofWca:
Just found out that this significantly improves performance in the "scroll chat way up" situation.
Previously scrolling beyond 100 messages up would slow things down terribly so that you constantly hit the top when scrolling.

Issue #6632, opened 2026-08-10
channel reactions


r10s:
next core will allow to react in channels. this requires some small changes in each UI. for desktop:

- [x] allow subscribers to use react in channels using only the five default reactions (👍, 👎, ❤️, 😂, 🙁) - the reaction option "..." to pick a random reaction must not be available in channels

- [x]  for channels, display reactions as usual. no need to filter by default or so, this will be done by the core as needed. i assume, this point is already just fine

- [x] don't notify reactions in outgoing-channels if channel is muted (do not consider them as mentions)

- [x] allow to mute/unmute an outgoing channel

- [x] reaction details need adaption as `reactions_by_contact` is not available for incoming broadcasts. as a minimum, show the reactions followed by the count, smth as  
`👍   3 reactions`  
`❤️   2 reactions`  
we can reuse the string `n_reactions` for that. this is what iOS is doing. android does much better here, showing pills that are tappable, that'd be nice as well, but if it delays closing this issue, better do the simple one first and the complex in another PR.

counterpart of https://github.com/deltachat/deltachat-ios/pull/3225 and https://github.com/deltachat/deltachat-android/pull/4560

details in the core PRs https://github.com/chatmail/core/pull/8450 and https://github.com/chatmail/core/pull/8496

kermoshina:
In Telegram the channel owner can pick their own set of reactions. Some prefer to not have binary 👍🏼👎🏼 and rather put some random weird stuff. A popular telegram channel on the topic of internet freedom uses for example just two réactions: 🐳 and 🌭. 

I guess this is harder to implement and we may want to limit ourselves with the 5 default ones. But I at least wanted to mention this

r10s:
"channel owner defined reactions" are known and planned, but will be the next step, see comment in core PR

ftr, in telegram, one needs to pay or upgrade channel somehow for many changes. leveling up. there are also paid reactions. crypto currency and premium is somehow related etc ... idk, and stopped wasting time trying to understand that. that is really ux from hell for nerds

PR #6635, opened 2026-08-11
feat: react to channel messages

WofWca:
Partially addresses
https://github.com/deltachat/deltachat-desktop/issues/6632.

What's left is the "Reactions" dialog for incoming channels,
which IMO can be kept for a follow-up.
Currently it simply displays either 0 or a single reaction from self.
However I'd say that this is not critical,
because reactions are already displayed inline on the message itself.
Although the list might get truncated on narrow screens.

### DIFF
--- a/packages/frontend/src/components/ReactionsBar/useReactionsBar.ts
+++ b/packages/frontend/src/components/ReactionsBar/useReactionsBar.ts
@@ -46,7 +46,14 @@ export default function useReactionsBar(): UseReactionsBar {
 
 /** Returns true if user should be able to send reactions to a message */
 export function showReactionsUi(message: T.Message, chat: T.FullChat): boolean {
-  return chat.canSend && !message.isInfo && message.originalMsgId === null
+  return (
+    (chat.canSend ||
+      (chat.chatType === 'InBroadcast' &&
+        chat.contactIds.includes(C.DC_CONTACT_ID_SELF) &&
+        !chat.isContactRequest)) &&
+    !message.isInfo &&
+    message.originalMsgId === null
+  )
 }
 
 function getMyReaction(reactions: T.Message['reactions']): string | undefined {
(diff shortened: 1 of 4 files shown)


WofWca:
- Replaced `NoLinger` with `WithLinger` because the former goes `undefined` while the chat is being reloaded...
- Fixed inverted test check.
- Hid the "React" button for incoming contact-request channels and for channels that you're not a member of.
  This logic should probably be a part of Core, but I guess it's not critical since an this would result in a reaction not getting sent.

packages/frontend/src/components/ReactionsBar/useReactionsBar.ts
```diff
@@ -46,7 +46,14 @@ export default function useReactionsBar(): UseReactionsBar {
 
 /** Returns true if user should be able to send reactions to a message */
 export function showReactionsUi(message: T.Message, chat: T.FullChat): boolean {
-  return chat.canSend && !message.isInfo && message.originalMsgId === null
+  return (
+    (chat.canSend ||
+      (chat.chatType === 'InBroadcast' &&
+        chat.contactIds.includes(C.DC_CONTACT_ID_SELF) &&
+        chat.isContactRequest)) &&
```
nicodh: This should be !chat.isContactRequest 
- as you wrote above -  don't show reactions on contact requests

nicodh (changes requested):
Except for the one condition this looks fine to me

I can provide a followup with the reaction dialog and some more e2e tests

packages/frontend/src/components/ReactionsBar/useReactionsBar.ts
WofWca:
```suggestion
        !chat.isContactRequest)) &&
```

Oops. Good thing we have tests.

nicodh (approved):

PR #6638, opened 2026-08-11
feat:  complete channel reactions

### LINKED ISSUE #6632: channel reactions

nicodh:
Closes #6632

- shows the appropriate reactions dialog (hiding who reacted to subscribers)
- adds e2e tests

### DIFF
--- a/packages/e2e-tests/tests/group-tests.spec.ts
+++ b/packages/e2e-tests/tests/group-tests.spec.ts
@@ -769,6 +769,7 @@ test.describe('channel reactions', () => {
     await selectChat(page, channelName)
     await sendMessage(page, channelName, messageText)
 
+    await expect(message()).toBeVisible()
     await expect(message()).not.toContainText('😂')
   })
 
@@ -789,6 +790,8 @@ test.describe('channel reactions', () => {
   test('react', async () => {
     await switchToProfile(page, userB.id)
     await selectChat(page, channelName)
+
+    await expect(message()).toBeVisible()
 
     // Subscribers can react although they can't send messages
     await expect(
@@ -822,8 +825,9 @@ test.describe('channel reactions', () => {
     )
     // Subscribers don't get to know who reacted with what, so the rows are
     // the accumulated counts and not the contacts, which would be clickable
-    await expect(rows.getByRole('button')).toHaveCount(0)
+    await expect(rows).toHaveCount(1)
     await expect(rows.filter({ hasText: '😂' })).toContainText('1 reaction')
+    await expect(rows.getByRole('button')).toHaveCount(0)
   })
 
   test('owner sees the reaction', async () => {
--- a/packages/e2e-tests/tests/group-tests.spec.ts
+++ b/packages/e2e-tests/tests/group-tests.spec.ts
@@ -1,4 +1,4 @@
-import { expect, type Page } from '@playwright/test'
+import { expect, type Locator, type Page } from '@playwright/test'
 
 import {
   groupName,
@@ -14,6 +14,7 @@ import {
   test,
   createGroupChat,
   createChat,
+  openReactionsBar,
   selectChat,
   sendMessage,
 } from '../playwright-helper.js'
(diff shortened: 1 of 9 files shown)


packages/frontend/src/components/ReactionsBar/ReactionsBar.tsx
```diff
@@ -54,13 +57,18 @@ export default function ReactionsBar({
...
-      await BackendRemote.rpc.sendReaction(accountId, messageId, [emoji])
+    try {
+      await BackendRemote.rpc.sendReaction(
+        accountId,
+        messageId,
+        emoji === myReaction ? [] : [emoji]
+      )
+    } catch (error) {
+      log.error('Failed to send reaction:', error)
+      void openAlertDialog({ message: unknownErrorToString(error) })
```
WofWca: We usually also utilize the `error_x` string. Plus it would be nice to add the `Failed to send reaction:` to the dialog.

packages/e2e-tests/tests/group-tests.spec.ts
```diff
@@ -783,10 +790,14 @@ test.describe('channel reactions', () => {
     await switchToProfile(page, userB.id)
     await selectChat(page, channelName)
 
-    await message().click({ button: 'right' })
-    await page.getByRole('menuitem', { name: 'React' }).click()
+    // Subscribers can react although they can't send messages
+    await expect(
+      page.locator('textarea.create-or-edit-message-input')
```
WofWca, line 798: Let's use a more generic locator for a `not` assertion.

```suggestion
      page.locator('textarea')
```

packages/e2e-tests/tests/group-tests.spec.ts
```diff
@@ -783,10 +790,14 @@ test.describe('channel reactions', () => {
...
+    ).toHaveCount(0)
```
WofWca, line 799: This assertion is placed such that it might pass while the chat is still loading. Not the case currently, but might break in the future. Better to run it at least when the message has already loaded.

packages/e2e-tests/tests/group-tests.spec.ts
```diff
@@ -795,19 +806,53 @@ test.describe('channel reactions', () => {
...
+    await switchToProfile(page, userB.id)
+    await selectChat(page, channelName)
+
+    const rows = (await openReactionsDialog(page, message())).getByRole(
+      'listitem'
+    )
+    // Subscribers don't get to know who reacted with what, so the rows are
+    // the accumulated counts and not the contacts, which would be clickable
+    await expect(rows.getByRole('button')).toHaveCount(0)
+    await expect(rows.filter({ hasText: '😂' })).toContainText('1 reaction')
```
WofWca, line 829: It's more reliable to assert `'listitem').toHaveText([...])`

Same for the "owner sees who reacted" test, that one seems vague.

WofWca (approved):
Tested it. Works. Closes the issue.

Image

I left a couple of nits.

This also adds error handling around `rpc.sendReaction`, which looks good.
The `openReactionsBar` looks alright.

packages/e2e-tests/tests/group-tests.spec.ts
```diff
@@ -762,6 +769,7 @@ test.describe('channel reactions', () => {
     await selectChat(page, channelName)
     await sendMessage(page, channelName, messageText)
 
+    await expect(message()).toBeVisible()
```
WofWca, line 772:
```suggestion
```

I'm pretty sure the `not.toContainText` assertion would fail anyway if the element is not visible.

WofWca (approved):

PR #6639, opened 2026-08-12
fix: video background in quotes

### LINKED ISSUE #6606: weird quoted message style when reply is a video
notice the name of the quote author being on the black background etc.
Image

this is how it looks in android:

Image

nicodh:
resolves #6606 

only the footer has now black background, not the whole message. (in fact it's only the attachment container which has an extended footer covering the lower message padding)


WofWca (approved):
I'm not a fan of negative margins (a wrapper element could probably do the job), but I guess it works for now.

Image

PR #6644, opened 2026-08-12
chore: translations update 2026-08

nicodh:


packages/frontend/src/components/dialogs/Transports/index.tsx
```diff
@@ -129,8 +129,7 @@ export default function TransportsDialog(
     async (transport: Transport) => {
       const userConfirmed = await openConfirmationDialog({
         confirmLabel: tx('remove_transport'),
-        // TODO should be a question? Overall might need a better label?
-        message: tx('remove_transport'),
+        message: tx('confirm_remove_relay_x', transport.addr.split('@')[1]),
```
WofWca: I think Android doesn't do this splitting?
Also note that there might be several addresses on the same server.

WofWca (approved):
- The removed strings from `_untranslated` are replaced with new strings
- The TODO in "Remove Relay?" is addressed

I didn't check the translation files themselves

WofWca (approved):

PR #6650, opened 2026-08-13
perf: load messages without loading full chat

### LINKED ISSUE #5235: perf: don't wait for full chat info to load before loading messages.
- Delta Chat Version: 1.59.2
  1. Click on a non-current chat in the chat list

Switching `chatId` is instant, but loading `T.FullChat` takes time.

https://github.com/deltachat/deltachat-desktop/blob/928d73a70ec991266e8b0ff274ce8194c9b18b03/packages/frontend/src/contexts/ChatContext.tsx#L72-L75

However, `MessageListAndComposer` relies only on `chatWithLinger` but not on `chatId`:

https://github.com/deltachat/deltachat-desktop/blob/3bc21041bfbafea2da9c297adcdffa28528e8480/packages/frontend/src/components/MessageListView.tsx#L13-L18

It would be nice to start loading messages as soon as we know the new `chatId`.

Note that I once wanted to do this, but it was not super trivial, because `MessageListAndComposer` really does rely on `T.FullChat` and not just `chatId`.

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/5235.

- `chatWithLinger` of `ChatContext` is never the old chat now.
  Instead it's `undefined` while loading.
  The old chat is available only as `oldChat`.
- `ChatViewInner` now needs to be able to handle
  `chatWithLinger == undefined`.
- `useMessageList` now takes `chatId`,
  which changes immediately as the new chat gets selected,
  instead of `chatWithLinger.id`,
  which is what the commit title is about.
  
Note that this introduces a bit of extra blinking,
namely in the chat header and possibly in the composer,
but I think this is fine, and maybe even good
as it provides more immediate feedback.

JSON-RPC logs now look like this:

```js
▲ [JSONRPC] { method: 'marknoticed_chat', ... }
▲ [JSONRPC] { method: 'set_config', ... }
▲ [JSONRPC] { method: 'get_first_unread_message_of_chat', ... }
▲ [JSONRPC] { method: 'get_message_list_items', ... }
▲ [JSONRPC] { method: 'get_full_chat_by_id', ... }
▲ [JSONRPC] { method: 'get_chat_media', ... }
▲ [JSONRPC] { method: 'get_messages', ... }
▲ [JSONRPC] { method: 'get_messages', ... }
▲ [JSONRPC] { method: 'get_contact', ... }
▲ [JSONRPC] { method: 'get_draft', ... }
▲ [JSONRPC] { method: 'get_messages', ... }
▲ [JSONRPC] { method: 'get_messages', ... }
```

We start loading messages even before we start loading the full chat.

I like how minimal this turned out to be, and glad that we're moving away from the lingering state. I think the introduced blinking due to "placeholder" state while the new chat is loading is fine. Maybe even good as it gives more immediate feedback when you select another chat.

### DIFF
--- a/packages/frontend/src/components/ChatView/ChatView.tsx
+++ b/packages/frontend/src/components/ChatView/ChatView.tsx
@@ -49,6 +49,7 @@ export function ChatView(
     className?: string
   }
 ) {
+  const tx = useTranslationFunction()
   const { chatWithLinger } = useChat()
   return (
     <section
@@ -76,7 +77,15 @@ export function ChatView(
           {/* Dummy header to reduce layout shifting
           when unselecting/selecting a chat. */}
           <nav className={styles.chatNavbar} data-tauri-drag-region></nav>
-          <NoChatSelected />
+          <NoChatSelected
+            messages={
+              <div className='info-message big' style={{ alignSelf: 'center' }}>
+                <div className='bubble'>
+                  {tx('no_chat_selected_suggestion_desktop')}
+                </div>
+              </div>
+            }
+          />
         </>
       )}
     </section>
--- a/packages/frontend/src/components/ChatView/ChatView.tsx
+++ b/packages/frontend/src/components/ChatView/ChatView.tsx
@@ -1,5 +1,5 @@
 import styles from './styles.module.scss'
-import React, { useCallback, useContext, useId, useMemo } from 'react'
+import React, { useCallback, useContext, useId, useMemo, useRef } from 'react'
 import { C } from '@deltachat/jsonrpc-client'
 
 import MessageListAndComposer from '../message/MessageListAndComposer'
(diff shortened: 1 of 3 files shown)


nicodh (commented):
While cleaning up the chatWithLinger / oldChat split in ChatContext seems useful to me, the performance improvement here should be proved by measuring imo. There are some "costs" that this PR introduces: fast switching between chats always triggers a new store and immediate message load which will be performed in core even when the result is not needed any more.

Another concern: why does MessageListAndComposerPlaceholder hold the "real" composer component with all it's props etc. although it is inert? Wouldn't be a sceleton more appropriate?

In general I would suggest to always make real measurements for performance optimization and always have a look if a change reduces complexity (and maybe loc in best case)

WofWca:
> why does MessageListAndComposerPlaceholder hold the "real" composer component

Making a skeleton would mean trying to mimic the real composer, with some of its logic, such as "hide if can't send".

> have a look if a change reduces complexity

This reduces complexity.

PR #6651, opened 2026-08-14
fix: upload previews in separate job

nicodh:
This avoids the requirement to install different rsync implementations on MacOS and Windows

#public-preview


missytake (approved):
Great :) that should make it much less likely to break

PR #6652, opened 2026-08-14
chore: upgrade core to 2.59.0

nicodh:


packages/frontend/src/stockStrings.ts
```diff
@@ -31,6 +31,10 @@ export async function updateCoreStrings() {
     | typeof C.DC_STR_PHASING_OUT
     | typeof C.DC_STR_MESSAGE_PINNED_BY_OTHER
     | typeof C.DC_STR_MESSAGE_PINNED_BY_YOU
+    | typeof C.DC_STR_ADD_YOU
+    | typeof C.DC_STR_ADD_YOU_BY
+    | typeof C.DC_STR_REMOVE_YOU
+    | typeof C.DC_STR_REMOVE_YOU_BY
```
WofWca, line 37: While these are not translated, maybe we can use the old strings for now?

    Expected substring: "Member Me removed by Alice"
    Received string:    "You were removed by Alice."

Hmm though substitution is a problem, maybe not worth it.

WofWca (approved):
Not super nice that these trains are untranslated, but it's probably not worth it to try to work around this.

Otherwise looks like a regular upgrade.
Regarding this support-color thing, I'm not entirely sure what it's about. I've also been struggling with it when upgrading.

PR #6654, opened 2026-08-15
remove: setting for webxdc devtools

nicodh:
wexdc dev tools are per default enabled in devmode but never in production

#public-preview

### DIFF
--- a/packages/frontend/src/components/Settings/ExperimentalFeatures.tsx
+++ b/packages/frontend/src/components/Settings/ExperimentalFeatures.tsx
@@ -90,18 +90,6 @@ export function ExperimentalFeatures() {
       >
         {tx('webxdc_store_url')}
       </SettingsSelector>
-      <DesktopSettingsSwitch
-        settingsKey='enableWebxdcDevTools'
-        label='Enable Webxdc Devtools'
-        // See https://delta.chat/en/2023-05-22-webxdc-security,
-        // "XDC-01-004 WP1: Data exfiltration via desktop app DevTools"
-        //
-        // Although thanks to another hardening measure this shouldn't be
-        // easy to pull off. Namely, direct internet access is sort of
-        // disabled for the Electron part of the app:
-        // 853b584251a5dacf60ebc616f7fb10edffb5c5e5/src/main/index.ts#L12-L21
-        description='Careful: opening developer tools on a malicious webxdc app could lead to the app getting access to the Internet'
-      />
     </>
   )
 }
--- a/packages/target-electron/package.json
+++ b/packages/target-electron/package.json
@@ -44,7 +44,6 @@
     "dev": "pnpm build && pnpm start",
     "dev-inspect": "pnpm build && pnpm start --inspect",
     "start": "NODE_OPTIONS=--enable-source-maps electron . --devmode --disable-http-cache --translation-watch",
-    "start:nodev": "NODE_OPTIONS=--enable-source-maps electron . --disable-http-cache --translation-watch",
     "clean": "rm -rf ./bundle_out ./html-dist ./dist",
     "migration-pretest": "esbuild --bundle --sourcemap --platform=node migration-tests/*.ts --outdir=migration-tests/compiled --external:mocha --external:chai --external:adm-zip --external:isomorphic-ws --external:@deltachat/jsonrpc-client --external:@deltachat/stdio-rpc-server --format=esm",
     "migration-test": "pnpm migration-pretest && NODE_OPTIONS=--enable-source-maps mocha 'migration-tests/compiled/*.js'",
--- a/packages/target-electron/package.json
+++ b/packages/target-electron/package.json
@@ -44,6 +44,7 @@
     "dev": "pnpm build && pnpm start",
     "dev-inspect": "pnpm build && pnpm start --inspect",
     "start": "NODE_OPTIONS=--enable-source-maps electron . --devmode --disable-http-cache --translation-watch",
+    "start:nodev": "NODE_OPTIONS=--enable-source-maps electron . --disable-http-cache --translation-watch",
     "clean": "rm -rf ./bundle_out ./html-dist ./dist",
     "migration-pretest": "esbuild --bundle --sourcemap --platform=node migration-tests/*.ts --outdir=migration-tests/compiled --external:mocha --external:chai --external:adm-zip --external:isomorphic-ws --external:@deltachat/jsonrpc-client --external:@deltachat/stdio-rpc-server --format=esm",
     "migration-test": "pnpm migration-pretest && NODE_OPTIONS=--enable-source-maps mocha 'migration-tests/compiled/*.js'",
--- a/packages/target-electron/src/deltachat/webxdc.ts
+++ b/packages/target-electron/src/deltachat/webxdc.ts
@@ -459,7 +459,7 @@ export default class DCWebxdc {
                     // Deliberately no `role: 'toggleDevTools'`: the role opens
                     // devtools directly, bypassing the confirmation below.
                     // Its shortcuts are bound to this item instead.
-                    click: () => {
+                    click: async () => {
                       const { webContents } = webxdcWindow
                       if (webContents.isDevToolsOpened()) {
                         webContents.closeDevTools()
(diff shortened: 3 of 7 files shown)


nicodh:
As discussed with @r10s

packages/target-electron/package.json
```diff
@@ -44,6 +44,7 @@
     "dev": "pnpm build && pnpm start",
     "dev-inspect": "pnpm build && pnpm start --inspect",
     "start": "NODE_OPTIONS=--enable-source-maps electron . --devmode --disable-http-cache --translation-watch",
+    "start:nodev": "NODE_OPTIONS=--enable-source-maps electron . --disable-http-cache --translation-watch",
```
WofWca:
```suggestion
    "start:no-dev": "NODE_OPTIONS=--enable-source-maps electron . --disable-http-cache --translation-watch",
```

Might sound like `node-v`.

packages/frontend/src/components/Settings/ExperimentalFeatures.tsx
```diff
@@ -90,18 +90,6 @@ export function ExperimentalFeatures() {
...
-      <DesktopSettingsSwitch
-        settingsKey='enableWebxdcDevTools'
-        label='Enable Webxdc Devtools'
-        // See https://delta.chat/en/2023-05-22-webxdc-security,
-        // "XDC-01-004 WP1: Data exfiltration via desktop app DevTools"
-        //
-        // Although thanks to another hardening measure this shouldn't be
-        // easy to pull off. Namely, direct internet access is sort of
-        // disabled for the Electron part of the app:
-        // 853b584251a5dacf60ebc616f7fb10edffb5c5e5/src/main/index.ts#L12-L21
```
WofWca, line 102: Not sure if it's good to remove this part and add an exclamation point instead. But I won't insist.

WofWca (commented):
Note that some WebXDC app developers rely on dev tools.

- [x] I'm thinking: maybe it would actually be fine to enable dev tools by default? Since, as the comment that this MR proposes to remove says, the devtools exploit is actually covered by another security layer.

Regarding the removal of the `before-input-event`: I am fine with it. We anyway have the Ctrl + Shift + I shortcut.
It might also improve performance, which is good.

This will need a corresponding MR on the WebXDC website:
https://github.com/webxdc/website/blob/4fe40f5c65bde28d9f8eb97a817e4e6d410613f6/src-docs/faq/debugging.md#L65-L99
Note that the command is `deltachat-desktop -- --devmode` and not `deltachat-desktop --devmode`.

WofWca (approved):
OK, I think it's fine to go for this. If one needs dev tools, they'll have to restart the app with `--devmode`. A bit more annoying than a toggle, but probably good enough.

The majority of users don't need to be bothered about this toggle at all.
OTOH maybe what we'd want is another level of "advanced"-ness in settings, but IDK. (Settings IMO need to be completely reorganized first).

Regarding simply always enabling dev tools even in prod: probably another topic. Still not sure if it's safe. But would be really nice, to further demonstrate that these really are "just web apps".

Please make sure that we don't forget to update the docs before merging this (see my comment above).

r10s:
hm. my idea was not to make it harder for devs and user,  but easier.

i think the current is bad eg for macos, or windows, it is not always clear there, how to pass args. doable, sure, but eg i would not know without research

my idea in fact was the following: 
- remove the settings option
- leave the menu entry
- when selecting, the menu entry, let the user confirm about the risk
- that confirmation is redone on every restart of delta (so, just a in-memory toggle). which is good, as we consider the question needed currently - and the real risk is not the thing in development by oneself, but a random 3rd party app, of unknown source, which one wants to inspect - long after having forgotten that one confirmed the risk at some point a year ago - or set some option in some startup line :)

if we later think, the confirmation is not needed, the better

WofWca:
I just remembered that if the option is to be removed then its string is also to be removed,

nicodh:
> * remove the settings option
> * leave the menu entry
> * when selecting, the menu entry, let the user confirm about the risk

Well if the devtools are always enabled the we need to catch all possible shortcuts to open them on all platforms to show the confirmation dialog

https://github.com/jaredwilli/devtools-cheatsheet

WofWca:
I think it should be possible to remove the native shortcut and to introduce a custom one which would programmatically open the dev tools.

nicodh:
> I think it should be possible to remove the native shortcut and to introduce a custom one which would programmatically open the dev tools.

You mean all 8 possible shortcuts that might be bound to that (according to https://developer.chrome.com/docs/devtools/shortcuts?hl=de#open ) . There is no onDevToolOpened event to listen to

WofWca:
Sorry, I meant the native menu item.

nicodh:
But the shortcuts also need to be catched

WofWca:
I'm not sure if I'm getting your point. If the native `role: 'toggleDevTools'` menu item is not present, the shortcut will just not work. And I just re-tested it. We don't need to be catching and preventing the shortcut manually or anything. Is this what you're talking about?

Simon-Laux:
`--devmode` is not something I would recommend to users, in devmode the app is slightly slower because all the logging and in the DevTools of the main window you have full jsonrpc access to core (`exp.rpc`).
I believe Bjorn's suggestion of "warning in confirmation dialog before opening them" is better, provided that is possible (probably needs the custom menu entry) and that it still solves the security issue.

r10s:
if we do not really have "Developer Tools" under our control, and cannot add a beforehand dialog on first access, i would also be fine with moving the switch from "Settings" dialog to the developer menu:

Screenshot 2026-08-18 at 15 25 08

for the main window, we're already adding custom items to "View / Developer", so maybe that's easier

packages/frontend/src/components/Settings/ExperimentalFeatures.tsx
nicodh: tbd: should we just use the existing text for a warning/confirmation dialog? Or just "Be aware that opening the devtools might carry some risks in malivious webxdc apps?"

nicodh:
Added a new solution (after discussing with @WofWca) 
The click on the menu item "Open devtools" triggers a dialog showing the warning about the possible risk. Once that is confirmed the devtools are opened and the settings is remembered for just that webxdc instance.

The same behaviour can be triggered with Ctrl+Shift+I which is the common shortcut to open the devtools

image

image

nicodh:
Note: the removal of the openWebxdcDevtools setting needs to be updated in tauri https://github.com/deltachat/deltachat-tauri/issues/25

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -430,29 +440,53 @@ export default class DCWebxdc {
...
+                        const confirmed =
+                          dialog.showMessageBoxSync(webxdcWindow, {
+                            type: 'warning',
+                            buttons: [tx('open'), tx('cancel')],
+                            defaultId: 1,
+                            cancelId: 1,
+                            title: tx('webxdc_devtools_dialog_title'),
+                            message: tx('webxdc_devtools_dialog_title'),
+                            detail: tx('webxdc_devtools_dialog_message'),
+                          }) === 0
```
WofWca: Any reason to use the sync version? It will block the main process.

```suggestion
                    click: async () => {
                      const { webContents } = webxdcWindow
                      if (webContents.isDevToolsOpened()) {
                        webContents.closeDevTools()
                        return
                      }
                      const instance = open_apps[appId]
                      if (!instance.devToolsConfirmed) {
                        const confirmed =
                          (
                            await dialog.showMessageBox(webxdcWindow, {
                              type: 'warning',
                              buttons: [tx('open'), tx('cancel')],
                              defaultId: 1,
                              cancelId: 1,
                              title: tx('webxdc_devtools_dialog_title'),
                              message: tx('webxdc_devtools_dialog_title'),
                              detail: tx('webxdc_devtools_dialog_message'),
                            })
                          ).response === 0
```

packages/target-electron/src/deltachat/webxdc.ts
```diff
@@ -430,29 +440,53 @@ export default class DCWebxdc {
...
+                      }
+                      const instance = open_apps[appId]
+                      if (!instance.devToolsConfirmed) {
+                        const confirmed =
+                          dialog.showMessageBoxSync(webxdcWindow, {
+                            type: 'warning',
+                            buttons: [tx('open'), tx('cancel')],
+                            defaultId: 1,
+                            cancelId: 1,
+                            title: tx('webxdc_devtools_dialog_title'),
```
WofWca: Doesn't seem good to mix up the order of the buttons. One might intuitively reach for "cancel" in the left corner, but it's the "action" button in reality.

WofWca (commented):
Regarding the fact that the permission is per-WebXDC-app-session and not per-Delta-Chat-session: this might I guess be safer, but I imagine this could get annoying if you keep debugging a single app, or re-sending different versions of the same app as you're developing it. So, maybe a global module-level bool, or a `permissionGivenAt` timestamp would be better.

PR #6656, opened 2026-08-15
remove: move tauri edition to own repository

nicodh:
The tauri package lives now in it's own repository, keeping a reference to this repository in a submodule. It can easily be synced and builded with the frontend package of this repository see  https://github.com/deltachat/deltachat-tauri


packages/target-tauri/README.md
```diff
@@ -1,87 +1 @@
...
-| macOS    | ~/Library/Application Support/chat.delta.desktop.tauri |
-| windows  | %AppData%\chat.delta.desktop.tauri                     |
-
-## Funding
-
-This project is funded through [NGI0 Entrust](https://nlnet.nl/entrust), a fund established by [NLnet](https://nlnet.nl) with financial support from the European Commission's [Next Generation Internet](https://ngi.eu) program. Learn more at the [NLnet project page](https://nlnet.nl/project/DeltaTauri).
-
-[<img src="https://nlnet.nl/logo/banner.png" alt="NLnet foundation logo" width="20%" />](https://nlnet.nl)
-[<img src="https://nlnet.nl/image/logos/NGI0_tag.svg" alt="NGI Zero Logo" width="20%" />](https://nlnet.nl/entrust)
+The tauri package has moved to it's own dedicated repository at https://github.com/deltachat/deltachat-tauri
```
WofWca:
```suggestion
The Tauri package has moved to its own dedicated repository at https://github.com/deltachat/deltachat-tauri
```

.gitignore
```diff
@@ -64,8 +64,6 @@ deltachat-rpc-server-aarch64-macos
 default.nix
 
 target
```
WofWca:
```suggestion
```

`target` is also a Tauri-only thing, see 96383b9d2ecd31e74ec6f709ddc673d499f2a2e2

README.md
```diff
@@ -8,11 +8,17 @@
...
+| The default application. Based on Electron. Currently deployed in appstore and used by most users. | Highly experimental version with a webserver component and web-ui in the browser. At the moment only meant for developers and automated testing.                                              |
+| [Project Folder](./packages/target-electron) <br /> [Download Links](https://get.delta.chat)       | [Project Folder](./packages/target-browser) <br /> [Fediverse Thread](https://fosstodon.org/@treefit/113116480883632874) <br /> [Blog Post](https://delta.chat/en/2025-05-22-browser-edition) |
+
+There is a third, **WIP** edition based on [`Tauri`](https://tauri.app/) <img src="README_ASSETS/TAURI_Glyph_Color.svg" width="16px" height="16px" style="vertical-align:middle" /> instead of
+Electron. It lives in its own repository — [deltachat-tauri](https://github.com/deltachat/deltachat-tauri) — and
+consumes this repository's frontend through a pinned `upstream/` submodule.
+See [Fediverse Thread](https://fosstodon.org/@treefit/113578409177635057) and
+[Delta Tauri - nlnet project](https://nlnet.nl/project/DeltaTauri).
```
WofWca, line 21: IDK, why do we remove Tauri from the table but keep the Browser version? Just based on the fact that Tauri is in a different repo? Not sure if it matters that much.
The browser version isn't something production-ready either.

.prettierignore
```diff
@@ -12,5 +12,3 @@ packages/e2e-tests/data
 packages/target-browser/dist
 packages/target-browser/data
 target
```
WofWca:
```suggestion
```

See https://github.com/deltachat/deltachat-desktop/pull/6656/changes#r3795113741.

flake.nix
```diff
@@ -1,18 +1,13 @@
 {
-  description = "A Nix-flake-based Rust development environment";
+  description = "A Nix-flake-based development environment for Delta Chat Desktop";
```
WofWca, line 2: Do you have recommendations on how to review this file as someone not familiar with NixOS?

flake.nix
```diff
@@ -24,31 +19,15 @@
...
-            extensions = ["rust-src" "rustfmt"];
-          };
-    };
-
     devShells = forEachSupportedSystem ({pkgs}: {
       default = pkgs.mkShell rec {
         nativeBuildInputs = with pkgs; [
           pkg-config
           gobject-introspection
-          cargo
```
WofWca, line 50: Is it right to remove `cargo` given that `build_and_link_local_core.py` (which is also referenced in this file) depends on it?

WofWca (commented):
TL;DR good overall, but see checklist item and inline comments.

Regardless of whether it's good to stop maintaining the Tauri version, I think moving to another repo is sensible, same as it is sensible to have Android, iOS, Desktop, and Core all separated into their own repos. At the same time it's also fine to have Electron in this repo, together with the front-end, since it remains the main distribution and the one that's used for development of the front-end the most.

It's nice that these two are so easy to separate, great job on that @Simon-Laux, and great job @nicodh on actually proving it, I really like that in the new repo only the commits that are relevant to `target-tauri` are left (although maybe the downside is that not all are individually buildable but I didn't check).

Maybe we'd have to think twice if Tauri wasn't so easily separable, but it is. And fundamentally it's not a big deal if it's in a separate repo or not.

Now: regarding whether it's good to stop maintaining Tauri: I think for now the answer is yes.
- This is still an experiment.
- It already lags behind some features.
- Tauri still isn't mature enough IMO to replace Electron.
- Sometimes its build fails when upgrading Core and making releases.
- It makes sense to track its issues and have a separate development process for it.

I think it'd be nice to come back to Tauri from time to time to see how it's doing, assessing how far we are from it being worth replacing Electron with it. But, again, that shouldn't be something that we are _forced_ to do in this repo.

I have done a search for "Tauri" and "Cargo" in the code base, found no more stray mentions.
This MR keeps the support for Tauri in the front-end.

Regarding the new repo: I did not review it thoroughly yet, as I am lacking the details as to how it was created. Reviewing manually is not feasible as there is a lot of code and a lot of commits. I am assuming the process was largely automated.
A couple of things, however:
- [ ] the `webxdc-check-permissions-policy-count` GitHub workflow has not been preserved.
- The Tauri build works and the product works.
- The contents of `packages/target-tauri` is identical to the contents of `deltachat-tauri` at HEAD~2 (except for the removed `.gitignore`).
- The last 2 commits add files, but I believe those are just (possibly modified) versions of the files that already existed.

flake.nix
nicodh: The removal was a hint from Simon

README.md
nicodh: Mainly because Tauri is not in this repository anymore. I just removed/moved all Tauri related docs etc.

flake.nix
nicodh: But I wonder if we should not remove the whole file. None of the current developers uses nixos for development...

WofWca: > The removal was a hint from Simon

Well I mean if it works now, maybe we can keep it until next time.

WofWca (approved):
Alright, the question of git history of the new Tauri repo is a separate topic, not a blocker for this MR. If needed we can just re-make it and force-push. As I said, the contents of that repo is nearly identical, and the build works, so we know that this is working out.
Having the "browser" edition but not the Tauri edition is questionable, but whatever, not a big deal, maybe a topic for the next README overhaul.

flake.nix
Simon-Laux: Without such a file people using NixOS have a hard time contributing because some things like electron are build for normal Linux and don't find their libraries otherwise.

Also using it can make sense for normal Linux users too because with it, you can install the whole dependency list for a complete development environment with one command. Think of node version manager, just for all dependencies, even python and everything else can be pinned to a specific version with it.

flake.nix
```diff
@@ -62,13 +41,10 @@
...
-          libsoup_3
           pango
-          webkitgtk_4_1
           openssl
           libayatana-appindicator
           # Video/Audio data playback
-          gst_all_1.gstreamer
           gst_all_1.gst-plugins-base
           gst_all_1.gst-plugins-good
           gst_all_1.gst-plugins-bad
```
Simon-Laux:
```suggestion
```

These are just the codecs for `gstreamer`, remove them as well.

PR #6657, opened 2026-08-16
refactor: move "recent apps" code deeper

WofWca:
Cleans up `MainScreen` nicely IMO.
Originally this code had to be in `MainScreen`
because that was where the header was. See
3bc21041bfbafea2da9c297adcdffa28528e8480
(https://github.com/deltachat/deltachat-desktop/pull/5151).

This is not an entirely pure move. I also did the following:
- removed the conditions for handling `accountId` and `chatId`
  beind possibly `undefined`.
- Removed the comment about "why asyncThrottle", see below why.

With this commit, `asyncThrottle` does _not_ throttle
across different chats. This is because the place where `useFetch`
is behind `key={chatId}`.
But I think that overall the performance is improved:
we won't start loading the "recent apps"
until we have loaded the full chat,
which is good since "recent apps" is generally lower-priority thing.
This is how the JSON-RPC request log looks like now
when switching a chat:

```log
[JSONRPC] { method: 'marknoticed_chat', ... }
[JSONRPC] { method: 'set_config', ... }
[JSONRPC] { method: 'get_first_unread_message_of_chat', ... }
[JSONRPC] { method: 'get_message_list_items', ... }
[JSONRPC] { method: 'get_full_chat_by_id', ... }
[JSONRPC] { method: 'get_messages', ... }
[JSONRPC] { method: 'get_contact', ... }
[JSONRPC] { method: 'get_draft', ... }
[JSONRPC] { method: 'get_chat_media', ... }
[JSONRPC] { method: 'get_messages', ... }
[JSONRPC] { method: 'get_webxdc_info', ... }
```

Marking as draft because the base is not merged.

### DIFF
--- a/packages/frontend/src/components/ChatView/ChatView.tsx
+++ b/packages/frontend/src/components/ChatView/ChatView.tsx
@@ -1,5 +1,12 @@
 import styles from './styles.module.scss'
-import React, { useCallback, useContext, useId, useMemo, useRef } from 'react'
+import React, {
+  useCallback,
+  useContext,
+  useEffect,
+  useId,
+  useMemo,
+  useRef,
+} from 'react'
 import { C } from '@deltachat/jsonrpc-client'
 
 import MessageListAndComposer from '../message/MessageListAndComposer'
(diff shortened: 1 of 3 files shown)


WofWca:
Why I pushed the last change: it's not great to have `asyncThrottle` be global, because it makes it possible to cross-contaminate the state between different callers: previous calls to it resolve to the next call's result if throttled.

PR #6659, opened 2026-08-16
perf: wait for idle before loading "recent apps"

WofWca:
Again, this is a lower-priority widget, and `get_chat_media`
is not super cheap (takes ~50ms on my machine
with no other concurrent request).
Still this doesn't delay the load significantly,
no longer than half a second,
at least when the machine is idle otherwise.

Related:
- https://github.com/chatmail/core/pull/7325

Marking as draft because the base is not merged.

### DIFF
--- a/packages/frontend/src/components/ChatView/ChatView.tsx
+++ b/packages/frontend/src/components/ChatView/ChatView.tsx
@@ -576,6 +576,12 @@ function AppIcon({ accountId, app }: { accountId: number; app: T.Message }) {
 // since "recent apps" is a lower-priority widget.
 const throttledFetchLastUsedApps = asyncThrottle(
   async (accountId: number, chatId: number, smallScreenMode: boolean) => {
+    // We don't provide a timeout, so this can result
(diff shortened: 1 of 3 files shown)


r10s:
the PR description describes the "what", but not so much "why is that needed? which problem does it solve?"

r10s:
ftr, with upcoming "pinned messages" it may be that the "three recent app" become less important. so investing lots of resources/time just now may be "late to the party".

WofWca:
> the PR description describes the "what", but not so much "why is that needed? which problem does it solve?"

IDK, I would classify the first 4 characters of the title and the first sentence of the description as the "why".

We load this widget right after switching a chat, and when switching a chat we have a lot of work to do: load the chat, load the messages, mark the messages as seen, etc.

WofWca:
This MR is no longer relevant as I had to change the approach in the base MR: https://github.com/deltachat/deltachat-desktop/pull/6657#issuecomment-5313777935.

It's still possible to load on idle, but it has to be done differently now.

PR #6663, opened 2026-08-18
fix: add Arabic and Telugu languages

WofWca:
They are in the list of available languages but each has 0 strings.
The fact that the two languages' translations
don't have any strings that are plural
makes the script bail early as if this is an error.

image

This bug has existed for at least 8 years,
a63f319bafcff2282b159a98ceeeee8d24cd2b70
(https://github.com/deltachat/deltachat-desktop/pull/422).


nicodh (approved):
Reviewed by reading

PR #6666, opened 2026-08-18
feat: allow user to resize chat list

WofWca:
[resize.webm](https://github.com/user-attachments/assets/d8031ef2-9868-4731-af1c-e382a0c058bc)

Note that this affects the default relative sizes of the sections, 
and reduces `min-width`
so for existing users the width will change as soon as they upgrade.

Again, the fact that it jumps as you start resizing it is not nice,
but I don't know a nice way to fix this without JS,
while preserving the "specify width as a fraction" nature.

For comparison, Element Desktop uses some JS to set `flex-grow`
on the sections.

The `flex-grow` and `flex-shrink` values might need adjustments
according to good taste, but I think it's already good enough.

I also removed the "refuse to shrink" comment as I suspect
that `width: 0;` is not necessary anymore,
plus we now need to set a different `width` value.

This is based on top of https://github.com/deltachat/deltachat-desktop/pull/6671 due to conflicts,
but this doesn't strictly require that MR. Let me know if you need me to rebase this onto main.

### DIFF
--- a/packages/frontend/scss/chat/_chat-list-item.scss
+++ b/packages/frontend/scss/chat/_chat-list-item.scss
@@ -215,6 +215,7 @@
     linear-gradient(grey 40px, transparent 0),
     linear-gradient(grey 40px, transparent 0),
     linear-gradient(grey 40px, transparent 0);
+  // TODO style: make this adapt to actual chat list width.
   background-size:
     56px 56px,
     200px 18px,
--- a/packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
+++ b/packages/frontend/src/components/screens/MainScreen/MainScreen.tsx
@@ -11,7 +11,10 @@ import { C } from '@deltachat/jsonrpc-client'
 
 import ChatList from '../../chat/ChatList'
 import ConnectivityToast from '../../ConnectivityToast'
-import SettingsStoreInstance from '../../../stores/settings'
+import SettingsStoreInstance, {
+  DesktopSettingsStoreInstance,
+  useDesktopSettingsStore,
+} from '../../../stores/settings'
 import { BackendRemote, onDCEvent } from '../../../backend-com'
 import ChatListHeader from './ChatListHeader'
 import { ChatView } from '../../ChatView/ChatView'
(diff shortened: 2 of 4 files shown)


nicodh (changes requested):
1. it seems resizing is only possible at a very small (15px ?) at the bottom of the chat list. 
2. you wrote "but I don't know a nice way to fix this without JS" - si why not use JS? 

The current solution is not acceptable imo, mainly for the first issue

WofWca:
> 1

Yes, this is how it is natively. Is this a big problem? What's so bad about aiming not only horizontally but also vertically?

> 2

It's simpler if it's only CSS.

nicodh:
> Yes, this is how it is natively. Is this a big problem? What's so bad about aiming not only horizontally but also vertically?

I mean the width resizing should be possible on any point of the border between chatlist and message list. Not only at the lowest 15px. Maybe get some inspiration from https://github.com/signalapp/Signal-Desktop/blob/dd424a9642204dfb52584e8d625e48a50dc3d137/ts/components/NavSidebar.dom.tsx#L98 - at least the result looks as expected imo

WofWca:
> on any point of the border between chatlist and message list

Yes, I understood it. I'm asking why you think that the way it is in this MR is worse than nothing.

r10s:
first, thanks on pushing this feature!

> Yes, this is how it is natively. Is this a big problem? What's so bad about aiming not only horizontally but also vertically?

we should do what the user expects from comparable apps. i just tried the current implementation, and beside cursor position and jumping bugs, it is barely visible.

we should look at telegram/signal/finder/etc. and do what they're doing here, no need to discuss much.

and: at the same time as introducing the feature, we shall stop changing the width of the chatlist as a percentage of the window width as well. again, look at telegram/signal/finder/etc. to get the idea:

- the bar changes the width of the chatlist (making the message list smaller)

- the window resize width changes adapt the width of the message list (keeping the chatlist the same)

WofWca:
> we shall stop changing the width of the chatlist as a percentage of the window width as well. again, look at telegram/signal/finder/etc.

If "etc." here includes Element, then "stop changing the width of the chatlist as a percentage of the window width" does not apply. See MR description.

r10s:
> If "etc." here includes Element, then "stop changing the width of the chatlist as a percentage of the window width" does not apply. See MR description

of course, obviously, there are apps that do differently.

however, in general, element is usually not regarded as a role model. in special, it is more in the category "team collaboration software" than "messenger"

PR #6668, opened 2026-08-19
refactor: move `Runtime` desktopSettings closer

WofWca:
Just move the declarations, nothing else.


nicodh (approved):
Just moves code around...

PR #6670, opened 2026-08-19
fix: hide menubar setting on Mac

### LINKED ISSUE #6667: "Hide Menu Bar" on macOS should be hidden, not disabled
ui-wise, "disabled state" is used for things, that by normal usage gets to "enabled state", so that the user can select it. 

this is not true for the "hide menu bar" option on macOS, that will never be enabled, and can never be selected. therefore, the option should be "hidden", not only "disabled"

nicodh:
closes #6667


WofWca (approved):
I confirm that Electron docs say that `autoHideMenuBar` has no effect on macOS.
Also yep, this was how it was originally in the MR before the last-minute change.

PR #6671, opened 2026-08-19
perf: don't reload global settings on acc switch

WofWca:
Basically split off the `desktopSettings` part of the `SettingsStore`.

This addresses this comment https://github.com/deltachat/deltachat-desktop/issues/4200#issuecomment-4048983680.

We do `SettingsStoreInstance.effect.clear()` on every account switch,
which also makes `desktopSettings` inaccessible until the next
`.effect.load()`.
`desktopSettings` is not tied to a single account
so it doesn't need to be reloaded.

One visible effect of this is the flash of the chat background image
which falls back to the default while we're loading
the next account's settings.

This is gonna be more important when we introduce
chat list section resizing, where it would also flash
the default value.

Also I think this is nicer code-wise.

I used AI for updating all the referenced of `desktopSettings`,
since that part is trivial.
The `stores/settings.ts` changes are hand-written,
the rest is AI-generated, and skimmed through by me.

### DIFF
--- a/packages/frontend/src/stores/settings.ts
+++ b/packages/frontend/src/stores/settings.ts
@@ -31,7 +31,6 @@ export interface SettingsStoreState {
       team_profile: '0' | '1'
     }[P]
   }
-  desktopSettings: DesktopSettingsType
   rc: RC_Config
 }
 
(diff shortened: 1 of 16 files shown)


packages/frontend/src/stores/settings.ts
```diff
@@ -61,41 +60,76 @@ export const enum WhoCanCallMe {
...
+          [key]: value,
+        }
+      }, 'set')
+    },
+  }
+  effect = {
+    load: async () => {
+      this.reducer.setState(await runtime.getDesktopSettings())
+    },
+    setDesktopSetting: async <T extends keyof DesktopSettingsType>(
```
nicodh: nitpick: while SettingsStore.effect.setCoreSetting might make sense, DesktopSettingsStore.effect.setDesktopSetting sounds a bit weird.

Maybe use just effect.set and rename reducer.set to reducer.applyExternalChange (to make more obvious that the reducer does not persist the state?)

nicodh (approved):
To distinguish DesktopSettings and AccountSettings makes sense. Reviwed by reading & testing.

I never saw any flashing of the background image though.

packages/frontend/src/stores/settings.ts
WofWca: Hm, I renamed the `reducer` part but not the `effect` part. Perhaps I simply forgot, or wanted to be explicit? Or to be consistent with the Core settings store interface. Or to have less changes wherever it's used. But yes, it makes sense for the two to be named the same.
It renamed it in the force-push.

> rename reducer.set to reducer.applyExternalChange

I agree that some more clarity on this front would be nice, but I'm not sure that this is the right wording.
This also affects the original Core settings store, so this might be something for a separate commit / MR.

PR #6674, opened 2026-08-20
improvement: chat list item: shrink date if needed

WofWca:
- **refactor: chat list item: remove unnecessary `div``**
- **improvement: chat list item: shrink date if needed**


WofWca:
Actually I figured that it anyway might be better to reconsider the whole shrinking order: e.g. it's not great that we shrink the name to 0 and prefer to show the date in full.

image

Issue #6673, opened 2026-08-20
cannot read "Show full message..." window with a screen reader


fastfinge:
- Operating System (Linux/Mac/Windows/iOS/Android): Windows
- Delta Chat Version: 2.57.0
- Expected behavior: after pressing "Show Full Message…", the new window that opens should have the content of the message visible to screen readers
- Actual behavior: After pressing "Show Full Message…", the window that opens only contains the username, date, and a ⋮. No message text can be read by a screen reader.
- Steps to reproduce the problem: install NVDA. Receive a long message via deltachat. Press "show full message". Try to read the message with NVDA or any other screen reader.
- Screenshots:
- Logs:

WofWca:
Hey, thanks!

I am pretty sure that this has to do with the fact that there are multiple web views.

I also noticed that you can't focus the content web view by using just the Tab key. Only after you actually click on the contents, then the focus transfers there, and you can focus links and stuff.

WofWca:
If you want a workaround: Menu (press Alt) -> Edit -> Select All. This will focus the contents.

Related:
- d9bcb1bf553f19a5d13345545f8440ef8bd62cba

WofWca:
Guess we can juts focus the contents by default for now. Not a solution, but better than focusing the relatively nearly useless header.

fastfinge:
Thanks, that worked perfectly! Exactly what I needed.

nicodh:
I think we can close this now, since the solution to focus the content is fine imo.

The only thing not reachable then for screen readers is the option to show/load images which defaults to false which should be fine for screen reader users.

Any concerns?

WofWca:
> which should be fine for screen reader users
> Any concerns?

Images are still useful to screen reader users. They might want to pass images through OCR to recognize text or through AI to get a description, or to simply download for further action.

nicodh:
So maybe add a similar option to be accessible via menu?

WofWca:
Sounds like it should work as a workaround.

Although I'm wondering how really useful this "header" WebView is overall. It's just the subject and the timestamp (edit: and the "load remote content" button). And the subject is already present in the window's title. Though I wasn't the one who introduced it, so maybe there is a bigger idea behind it.

nicodh:
Maybe it's just about the content window does not allow javascript (for good reasons) so a 3dot menu would not work?

WofWca:
No, my point is not to move the header to the content's WebView but to remove it completely.

[...]

PR #6675, opened 2026-08-20
fix: unfocusable "Show Full Message..." contents

WofWca:
See https://github.com/deltachat/deltachat-desktop/issues/6673.
This doesn't really fix it, but improves the situation.


nicodh (approved):
According to the user feedback this is an improvement

PR #6676, opened 2026-08-21
fix: update "Reactions" dialog live

WofWca:
What can happen if a reaction is removed while the dialog is still open:

Screenshot_20260821

### DIFF
--- a/packages/frontend/src/components/dialogs/ReactionsDialog/index.tsx
+++ b/packages/frontend/src/components/dialogs/ReactionsDialog/index.tsx
@@ -1,10 +1,16 @@
-import React, { useEffect, useRef, useState } from 'react'
+import React, {
+  useEffect,
+  useEffectEvent,
+  useMemo,
+  useRef,
+  useState,
+} from 'react'
 import classNames from 'classnames'
 
 import Dialog, { DialogBody, DialogContent, DialogHeader } from '../../Dialog'
 import useTranslationFunction from '../../../hooks/useTranslationFunction'
 import { selectedAccountId } from '../../../ScreenController'
-import { BackendRemote } from '../../../backend-com'
+import { BackendRemote, onDCEvent } from '../../../backend-com'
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/dialogs/ReactionsDialog/index.tsx
```diff
@@ -34,26 +44,69 @@ type ContactWithReaction = T.Contact & {
...
+  /**
+   * The `originalMessage` prop will not update as reactions change,
+   * so we have to update it ourselves.
+   * Similar to {@linkcode BasicMessageInfo}.
+   */
+  const [originalIsOutdated, setOriginalIsOutdated] = useState(false)
+  const freshMessageFetch = useRpcFetch(
+    useMemo(
+      () =>
+        asyncThrottle(
```
nicodh, line 62: Throttling here seems not neccessary imo. If we skip that we also don't need originalIsOutdated state and less code is needed.

nicodh (commented):
Good improvement for an edge case (reactions updated while dialog is open)

Reviewed by reading & testing

packages/frontend/src/components/dialogs/ReactionsDialog/index.tsx
WofWca: > Throttling here seems not neccessary imo

It's nice for channels at least, where there might be a lot of reactions.

Maybe we ought to introduce a new `useRpcFetch` with this throttle built-in, so that it doesn't look so ugly.

> If we skip that we also don't need originalIsOutdated

No, that's a separate thing. `originalIsOutdated === false` ensures that we don't do another `rpc` request at all, and that the dialog renders immediately without waiting for an `rpc` round-trip.

WofWca:
(Also - not sure it was intentional - you didn't actually approve this MR)

nicodh (approved):

PR #6677, opened 2026-08-21
improvement: "Reactions" dialog title: show count

WofWca:
Screenshot_0-title

### DIFF
--- a/packages/frontend/src/components/dialogs/ReactionsDialog/index.tsx
+++ b/packages/frontend/src/components/dialogs/ReactionsDialog/index.tsx
@@ -82,9 +82,26 @@ export default function ReactionsDialog({
     ? freshMessageFetch.lingeringResult.value
     : originalMessage
 
+  const totalReactions =
+    message.reactions == null
+      ? 0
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/dialogs/ReactionsDialog/index.tsx
```diff
@@ -94,7 +111,7 @@ export default function ReactionsDialog({
           <DialogContent>
             {message.reactions == null ||
             message.reactions.reactions.length === 0 ? (
-              <p>{tx('n_reactions', (0).toLocaleString(), { quantity: 0 })}</p>
+              <></>
```
nicodh, line 114: Why not just null?

nicodh (approved):
Reviewed by reading & testing

packages/frontend/src/components/dialogs/ReactionsDialog/index.tsx
WofWca: No strong reason. Just though that here it's appropriate to be more explicit that "if this, then we don't render anything".

PR #6680, opened 2026-08-21
improvement: "Reactions" dialog: show pills & list

WofWca:
It's the same on Android:
https://github.com/deltachat/deltachat-android/pull/4560.

image


nicodh (approved):

I only review this PR as #6679 is included and I think these should be in one PR, no need to split the current changes into a very tiny separate PRs imo

PR #6681, opened 2026-08-21
test: split up group and channel tests

WofWca:
When developing, this makes you not have to wait
for the group tests to finish before the channel tests start.

This was an easy split, only had to adjust imports and copy-paste the `beforeAll` and similar stuff.

~~Marking as draft because the base is not merged yet.~~
Actually there is just a tiny possible conflict, so this is ready for review.


nicodh (approved):
Reviewed by reading

PR #6684, opened 2026-08-22
fix: do not always create a team profile

### LINKED ISSUE #6683: All new profiles are created as "team profiles"
I am testing commit bdd602375368fe2f2b3def1470d4968b6c3f173b, latest on `main`, and all profiles that i create are created as "team profiles" now. This is also a problem on tag v2.59.0.

I checked the database, `config` table has `team_profile` set to 1 for such profiles.

It is like this since the introduction of the setting in https://github.com/deltachat/deltachat-desktop/pull/6585

link2xt:
Since the introduction of team profiles,
all profiles were created with `team_profile` core setting set regardless of the user choice.
It is visible in the hover overlay on top of each profile as well.

Fixes #6683

### DIFF
--- a/packages/frontend/src/components/screens/WelcomeScreen/InstantOnboardingScreen.tsx
+++ b/packages/frontend/src/components/screens/WelcomeScreen/InstantOnboardingScreen.tsx
@@ -188,7 +188,8 @@ export default function InstantOnboardingScreen({
 
     try {
       await Promise.all([
-        SettingsStoreInstance.effect.setCoreSetting('team_profile', '1'),
+        isTeamProfile &&
+          SettingsStoreInstance.effect.setCoreSetting('team_profile', '1'),
         saveDisplayName(),
       ])
       // Automatically create a "chatmail" account


link2xt:
I tested manually that i can creat a normal profile, a team profile, and that setting and unsetting "team profile" during profile creation results in non-team profile.

packages/frontend/src/components/screens/WelcomeScreen/InstantOnboardingScreen.tsx
```diff
@@ -188,7 +188,10 @@ export default function InstantOnboardingScreen({
 
     try {
       await Promise.all([
-        SettingsStoreInstance.effect.setCoreSetting('team_profile', '1'),
+        SettingsStoreInstance.effect.setCoreSetting(
+          'team_profile',
+          isTeamProfile ? '1' : '0'
+        ),
```
WofWca: This will avoid the `setCoreSetting` completely if `!isTeamProfile` it's not a team profile.

```suggestion
        isTeamProfile &&
          SettingsStoreInstance.effect.setCoreSetting('team_profile', '1'),
```

WofWca (approved):
Oooof, I don't know how managed to make such an obvious mistake. Probably forgot to update something during refactoring. Sorry about that.

Thanks for the fix! Looks correct to me now.

nicodh (approved):

WofWca:
I re-tested this for good measure. Works.

PR #6685, opened 2026-08-22
chore: prepare release 2.59.1

nicodh:


WofWca (approved):
The MR title says "upgrade Core", but it's a Desktop version bump (but it's understandable with all the me-caused-rush).

PR #6689, opened 2026-08-24
feat: adding transport: progress bar and "Cancel"

### LINKED ISSUE #5874: no progressbar when adding relays
as long as you can only add chatmail relays and no classical relays you could argue that is not needed, because configuration for those is usually fast. But when you add a classical relay with a `dclogin:` qr code, then it may take a while for deltachat, especially if the password is wrong.
For the cases where it takes longer showing the configuration progressbar would be nice. (the configuration progress events are there, there is just no progress dialog for adding relays yet)

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/5874.

### DIFF
--- a/packages/frontend/src/hooks/dialog/useAddTransportDialog.ts
+++ b/packages/frontend/src/hooks/dialog/useAddTransportDialog.ts
@@ -1,9 +1,10 @@
 import { useCallback } from 'react'
 
-import { BackendRemote } from '../../backend-com'
 import useAlertDialog from './useAlertDialog'
 import useConfirmationDialog from './useConfirmationDialog'
 import useTranslationFunction from '../useTranslationFunction'
+import useDialog from './useDialog'
+import { ConfigureProgressDialog } from '../../components/dialogs/ConfigureProgressDialog'
 
 type AddTransportDialogFn = (
   accountId: number,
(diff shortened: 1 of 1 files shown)


packages/frontend/src/hooks/dialog/useAddTransportDialog.ts
```diff
@@ -37,16 +39,29 @@ export default function useAddTransportDialog(): AddTransportDialogFn {
...
+      const relayAdded = await new Promise<boolean>(r => {
+        openDialog(ConfigureProgressDialog, {
+          qrCode: transportString,
+          credentials: null,
+          onSuccess() {
+            r(true)
+          },
+          onFail(error) {
+            r(false)
+            openAlertDialog({
```
nicodh, line 51: nitpick: when the user canceled the dialog it shows:

Relay could not be added. Configuration failed. Error: Canceled

Which is a bit weird since canceling is not an error and the reason is not that configuration failed. But acceptable if not easy to fix since the cancelation was a user previous action

nicodh (approved):
Reviewed by reading & testing

Nice improvement!

packages/frontend/src/hooks/dialog/useAddTransportDialog.ts
WofWca: I agree, this is a bit weird. Though I believe it's not something that changed in this MR. This also goes for other uses of the `ConfigureProgressDialog` dialog.

PR #6690, opened 2026-08-24
refactor: improve `openDialog` types

WofWca:
c3993f28eda53c2816169b9f0fc85a626473397b
(https://github.com/deltachat/deltachat-desktop/pull/5825)
made `onClose` work with `openDialog`,
and we already use it in some places
(search for `openDialog(\n|.){1,300}?^\s+OnClose[:,]`).
However, `onClose` doesn't appear in IDEs and its type
is basically `any`.

For reasons not completely clear,
this also disallows setting unknown properties
when calling `openDialog`, which I also had to fix here,
otherwise TypeScript complains.
These already don't actually do anything,
but also there is no need to try to make them work:
- `setViewType` was introduced in the very commit that introduced
  the `CloneChat` component itself. The prop was never used.
- `title: tx('error')` for `AlertDialog` is probably insignificant.
  We already provide the `error_x` text.
- `confirmLabel: tx('ok')` is already the default.

### DIFF
--- a/packages/frontend/src/components/Settings/ExperimentalFeatures.tsx
+++ b/packages/frontend/src/components/Settings/ExperimentalFeatures.tsx
@@ -56,10 +56,7 @@ export function ExperimentalFeatures() {
 
     message +=
       '\n\n• If you want to quit the experimental feature, you can disable it at "Settings / Advanced"'
-    openDialog(AlertDialog, {
-      message,
-      confirmLabel: tx('ok'),
-    })
+    openDialog(AlertDialog, { message })
   }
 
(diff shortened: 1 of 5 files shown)


nicodh (approved):
Reviewed by reading

Maybe "refactor: cleanup some openDialog, onClose parameter" would be a more concise PR title

WofWca:
My idea was that the main point is the first commit, and the second one is simply necessary for the first one to work. But yep, that perhaps is not sensible if the second commit is significant.
So I changed the title to a more generic one.

PR #6691, opened 2026-08-28
style: remove `cursor: pointer` on non-clickable

WofWca:
The button doesn't actually fill the parent completely
for some reason, so sometimes it so happens that you click
but nothing happens.

Maybe it'd be better to make the button match the parent,
but let's keep that for later (or never).


nicodh (approved):
Reviewed by reading

PR #6692, opened 2026-08-29
fix: check for resume from sleep with setInterval

### LINKED ISSUE #5915: Restore from sleep doesn't work on deb 2.35?
- Operating System (Linux/Mac/Windows/iOS/Android): Linux, Debian
- Delta Chat Version: 2.35
- Expected behavior: Starts fetching messages after restore from sleep
- Actual behavior: needs restart
- Screenshots:
- Logs:

See https://support.delta.chat/t/higher-incidence-of-messages-out-of-order-in-2-35/4586?u=wofwca.

This is about a non-Chatmail server, IDK if it's related.

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/5915
and other similar issues.
This is a method independent of permissions.

I had to add the package directly from GitHub and not NPM
because the NPM version has a bug where it ignores
the `threshold` parameter.

- [ ] What I'm a bit worried about with this is performance.
  We'll be running something every 5 seconds.
  The code that runs is pretty small itself,
  but maybe it forces the engine to keep a lot of stuff loaded
  or something?
- [ ] Oops, also apparently the lib is buggy. `lasTickDidTrigger` only gets reset in the "diff is too big" situation, so it only works half of the time.

### DIFF
--- a/packages/target-electron/package.json
+++ b/packages/target-electron/package.json
@@ -87,6 +87,7 @@
     "esbuild": "^0.28.1",
     "mocha": "^10.7.0",
     "rc": "^1.2.8",
+    "sleeptime": "github:timrach/sleeptime",
     "typescript": "catalog:"
   }
 }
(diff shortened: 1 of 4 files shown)


WofWca:
So it was decided that it's fine to add the D-bus permission in Flatpak instead, https://github.com/flathub/chat.delta.desktop/pull/221.

PR #6695, opened 2026-08-29
fix: use native `<dialog>.closedby`

WofWca:
This is mostly a refactor, but also a fix to the bug
where doing `mousedown` on a dialog
and then `mouseup` on its backdrop would still close the dialog.

Also change `LogDialog.closedby` `closerequest` -> `any`,
as desired.

The mapping is simple:
- `closedby='any'` when
  `canOutsideClickClose === true && canEscapeKeyClose === true`
  (default)
- `closedby='closerequest'` when
  `canOutsideClickClose === false && canEscapeKeyClose === true`
- `closedby='none'` when
  `canOutsideClickClose === false && canEscapeKeyClose === false`

This also makes "Back" work as "Escape" (good for mobile devices).

I have tested that this doesn't re-introduce
the "double Escape always closes the dialog" bug
(2bc987abc41a542ccbc12a20029b0e42d5ee2f46,
https://github.com/deltachat/deltachat-desktop/issues/4397).
One can test this on the "progress" dialog when making a new account.

Note that `HTMLDialogElement.closedby` is not yet supported
by WebKit (Safari), which is used by Tauri:
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog#closedby

### DIFF
--- a/packages/frontend/src/components/AccountListSidebar/EditPrivateTagDialog.tsx
+++ b/packages/frontend/src/components/AccountListSidebar/EditPrivateTagDialog.tsx
@@ -29,7 +29,7 @@ export function EditPrivateTagDialog({
   }
 
   return (
-    <Dialog onClose={onClose} canOutsideClickClose={true}>
+    <Dialog onClose={onClose} closedby='any'>
       <DialogHeader title={tx('profile_tag')} />
       <DialogBody>
         <DialogContent>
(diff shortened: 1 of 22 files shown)


nicodh (approved):
Reviewed by reading and testing.

Good catch! Simpler and more robust.

A small lint fix is still needed.

Maybe adding a line to Tauri [upstream issue](https://github.com/deltachat/deltachat-tauri/issues/25) after the merge makes sense

PR #6697, opened 2026-08-31
refactor: unify account dir in target browser

nicodh:
It was possible to set DC_ACCOUNTS_DIR in browser .env which breaks e2e tests. So this PR removes that variable and makes only use of DATA_DIR


WofWca (approved):
I'm not really getting why the solution isn't simply not to specify this env var, but alright. I personally don't use this var.

There is also `PRIVATE_CERTIFICATE_PATH` which is similar to `DC_ACCOUNTS_DIR`, so I guess you might want to remove that too.

nicodh:
> There is also `PRIVATE_CERTIFICATE_PATH` which is similar to `DC_ACCOUNTS_DIR`, so I guess you might want to remove that too.

No it is not similar. That is the path to your certificates and is completely independant of the account dir.

PR #6698, opened 2026-08-31
fix: improve scheme detection in links

### LINKED ISSUE #6628: `example.com/fun_ftp_facts` link not clickable

nicodh:
Resolves #6628

Some of the tests were created with AI assistance

### DIFF
--- a/packages/frontend/src/utils/linkify/getLinkDestination.ts
+++ b/packages/frontend/src/utils/linkify/getLinkDestination.ts
@@ -1,0 +1,65 @@
+import type * as linkify from 'linkifyjs'
+
+export type PunycodeWarning = {
+  original_hostname_or_full_url: string
+  ascii_hostname: string
+  punycode_encoded_url: string
+}
+
+export type LinkDestination = {
+  target: string
+  hostname: null | string
+  punycode: null | PunycodeWarning
+  scheme: null | string
+  linkText?: string
+}
+
+function stripLastSlash(url: string): string {
+  return url.endsWith('/') ? url.slice(0, -1) : url
+}
+
+/**
+ * Turn a linkify `url` token into the destination of a link.
+ *
+ * @returns `null` if the token is not a parsable URL, in which case it should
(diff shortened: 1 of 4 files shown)


packages/frontend/src/utils/linkify/getLinkDestination.ts
```diff
@@ -0,0 +1,65 @@
...
+  return url.endsWith('/') ? url.slice(0, -1) : url
+}
+
+/**
+ * Turn a linkify `url` token into the destination of a link.
+ *
+ * @returns `null` if the token is not a parsable URL, in which case it should
+ * be rendered as plain text instead of as a link.
+ */
+export function getLinkDestination(
```
WofWca, line 27: nit:
I think turning this into `isSuspicousUrl` would be simpler. Makes it more obvious what the tests are actually covering.

WofWca (approved):
`example.com/fun_ftp_facts` is now rendered correctly, no error.
I also tested some other links. Haven't managed to find issues.

The approach looks simpler. Uses a more "down-to-earth" function.

The moved code isn't actually simply moved but is also rewritten a bit. I have checked that `stripLastSlash` and the return values are the same.
I didn't review the tests (apart from a sanity-check of imports). I think we can have them, although it's smelling a bit like we'd want to have all that in a separate library.

PR #6700, opened 2026-09-01
fix: replace user feedback with dialogs

nicodh:


packages/frontend/src/components/composer/Composer.tsx
```diff
@@ -944,11 +943,13 @@ function useMessageEditing(
...
-        type: 'error',
-        text: tx('chat_please_enter_message'),
-      })
       log.error('doEdit called, but newText is empty')
+      void openAlertDialog({
+        message: tx('chat_please_enter_message'),
+      }).then(() => {
+        editMessageInputRef.current?.focus()
+        editMessageInputRef.current?.moveCursorToTheEnd()
+      })
```
WofWca, line 952: I doubt that it's good to show an entire dialog for such a minor thing. Especially for touch screens. At least we don't show a dialog for when the user tries to *send* an empty message.

I think we might after all need a "toast" component (but a proper one), where a dialog is not appropriate.
I think it would also be nice in other places, e.g. "success" actions, such as adding a relay, copying text, and possibly even for having a "cancel destructive action" button, e.g. deleting a relay.

If you really really want to get rid of `userFeedback` without a proper replacement, then I'd say it's better to simply not show anything here.

packages/frontend/src/ScreenController.tsx
```diff
@@ -273,8 +270,6 @@ export default class ScreenController extends Component {
   }
 
   componentDidMount() {
-    BackendRemote.on('Error', this.onError)
```
WofWca, line 276: Is there a justification for removing this? [The docs](https://github.com/chatmail/core/blob/fd8c56894ab85a24f5301edea1ac4b290f1af326/src/events/payload.rs#L54-L64) state

> The library-user should report an error to the end-user.

packages/frontend/src/components/composer/Composer.tsx
WofWca, line 952: Offtopic:
huh, the fact that we have to manually re-focus the previously focused element means that our dialog implementation is crappy. I'll take a look.

WofWca (commented):
Apart from the two places where I left a comment I think it's fine to replace those with a dialog.

packages/frontend/src/ScreenController.tsx
WofWca: Ah I see, it's provided in a different MR: https://github.com/deltachat/deltachat-desktop/pull/6701.

So, does it mean then that the docs are wrong now? That is, such errors should not be reported to the user anymore?

packages/frontend/src/components/composer/Composer.tsx
nicodh: I consider this a first step. The decision where to show a dialog or not or to introduce a new "toast" like dialog should go into a follow up. For now I would say a dialog is better than the userFeedback

WofWca: > introduce a new "toast" like dialog

Functionally it's not new. This `userFeedback` thing is already a toast (a poorly implemented one).

WofWca: Done: https://github.com/deltachat/deltachat-desktop/pull/6704

nicodh: I think it makes sense to discuss what is appropriate with @r10s when he is back

WofWca (approved):

PR #6704, opened 2026-09-01
fix: return focus on programmatic dialog close

WofWca:
Reproduction steps:
1. Focus the "New Chat" (bug plus) button.
2. Press Space.
3. Focus the "Close" button of the newly opened dialog.

Expected: focus is on the "New Chat" button.
Actual: nothing is focused, pressing tab focuses account list items.

The same goes for other dialogs, such as:
- Reactions
- Scan QR
- Gallery
- Setting

The context menu dialog is not affected by this bug.

This was brought up here:
https://github.com/deltachat/deltachat-desktop/pull/6700#discussion_r3902123858.

### DIFF
--- a/packages/frontend/src/components/Dialog/Dialog.tsx
+++ b/packages/frontend/src/components/Dialog/Dialog.tsx
@@ -119,6 +119,7 @@ const Dialog = React.memo<Props>(
         }
       }
     }, [allowDefaultFocus])
+    // Make sure to properly close the dialog before it's removed from DOM.
     // This ensures that the focus is returned to the element which had focus
     // before the dialog was opened, not only on "Escape" key close
     // but also on programmatic `onClose`, which simply un-renders the <dialog>
--- a/packages/frontend/src/components/Dialog/Dialog.tsx
+++ b/packages/frontend/src/components/Dialog/Dialog.tsx
@@ -1,5 +1,5 @@
 import classNames from 'classnames'
-import React, { useEffect, useRef } from 'react'
+import React, { useEffect, useLayoutEffect, useRef } from 'react'
 
 import styles from './styles.module.scss'
 import { runtime } from '@deltachat-desktop/runtime-interface'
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/Dialog/Dialog.tsx
```diff
@@ -119,6 +119,22 @@ const Dialog = React.memo<Props>(
...
+    // This ensures that the focus is returned to the element which had focus
+    // before the dialog was opened, not only on "Escape" key close
+    // but also on programmatic `onClose`, which simply un-renders the <dialog>
+    // (see `DialogContext`).
+    // From docs:
+    // > Before your component is removed from the DOM,
+    // > React will run your cleanup function.
+    //
+    // See also
+    // https://github.com/whatwg/html/issues/5802#issuecomment-2207368917
```
nicodh, line 132: I would not point to that comment, but better explain (summarize) what it's about

IIUC the fix is: make sure dialog.close is called while the element is still in the DOM and focus is still inside it, since then the browser sets the focus to the element that had the focus when the dialog was opened

nicodh (approved):
Reviewed by reading & testing

packages/frontend/src/components/Dialog/Dialog.tsx
```diff
@@ -119,6 +119,22 @@ const Dialog = React.memo<Props>(
         }
       }
     }, [allowDefaultFocus])
+    // This ensures that the focus is returned to the element which had focus
```
WofWca, line 123:
```suggestion
    // Make sure to properly close the dialog before it's removed from DOM.
    // This ensures that the focus is returned to the element which had focus
```

packages/frontend/src/components/Dialog/Dialog.tsx
WofWca: The reason I'm pointing to the comment is not to simply explain what we're doing here but also to keep track of the spec's changes, so as to know when this "hack" can be removed (or when it needs to be adjusted).

Though you're right, I think my comment wasn't entirely clear, so I added a suggestion.

PR #6709, opened 2026-09-02
improvement: deleteTransport instead of unpublish

WofWca:
The "unpublush" API will be removed in the upcoming Core:
https://github.com/chatmail/core/commit/70a01a6813073b614be8d513af20ac3af6a70286.

Note that the `confirm_remove_relay_x` string still says
that the relay will be "gradually phased out", this is tracked in
https://github.com/deltachat/deltachat-android/issues/4637.

### DIFF
--- a/packages/frontend/src/components/dialogs/Transports/index.tsx
+++ b/packages/frontend/src/components/dialogs/Transports/index.tsx
@@ -135,11 +135,7 @@ export default function TransportsDialog(
         return
       }
 
-      await BackendRemote.rpc.setTransportUnpublished(
-        accountId,
-        transport.addr,
-        true
-      )
+      await BackendRemote.rpc.deleteTransport(accountId, transport.addr)
       getTransports()
     },
     [accountId, getTransports, openConfirmationDialog, tx]


packages/frontend/src/components/dialogs/Transports/index.tsx
```diff
@@ -135,11 +135,7 @@ export default function TransportsDialog(
         return
       }
 
-      await BackendRemote.rpc.setTransportUnpublished(
-        accountId,
-        transport.addr,
-        true
-      )
+      await BackendRemote.rpc.deleteTransport(accountId, transport.addr)
```
nicodh, line 138: did you consider to wrap this in a try catch to give feedback to the user if deleteTransport failed?

nicodh (approved):
Reviewed by reading

packages/frontend/src/components/dialogs/Transports/index.tsx
WofWca: This is something that applies to `setTransportUnpublished` as well. Sounds reasonable to me, but belongs to a separate MR IMO.
Related: https://github.com/deltachat/deltachat-desktop/issues/3883.

PR #6710, opened 2026-09-02
improvement: allow to remove primary relay

WofWca:
The upcoming Core will allow it:
https://github.com/chatmail/core/commit/70a01a6813073b614be8d513af20ac3af6a70286.
So currently this doesn't work.


nicodh (approved):
Reviewed by reading

PR #6713, opened 2026-09-03
refactor: fix "unchecked index access" in notifs

WofWca:
The lower two changes don't change the behavior at all.
The first two hypothetically might,
but that basically doesn't affect behavior I believe,
except that we won't print an error to the console.


packages/target-electron/src/notifications.ts
nicodh:
```suggestion
  const { chatId, accountId, msgId } = data
```

If we extract msgId here we do not need to define it 3 times

nicodh (approved):
Reviewed by reading

Suggestion is no blocker imo

PR #6714, opened 2026-09-03
feat: add popover toast

nicodh:
Toast for now is used in 2 places: aftery copying to clipboard and empty text on message edit.

More tbd

### DIFF
--- a/packages/frontend/src/components/Toast/ToastLayer.tsx
+++ b/packages/frontend/src/components/Toast/ToastLayer.tsx
@@ -1,0 +1,92 @@
+import React, { useEffect, useRef } from 'react'
+import classNames from 'classnames'
+
+import styles from './styles.module.css'
+
+import type { Toast, ToastPosition } from '../../contexts/ToastContext'
+
+type Props = {
+  position: ToastPosition
+  toasts: Toast[]
+  onExpire: (id: Toast['id']) => void
+}
+
+/**
+ * Renders all currently visible toasts in the
+ * [top layer](https://developer.mozilla.org/en-US/docs/Glossary/Top_layer),
+ * so that they are visible even while modal dialogs are open.
+ *
+ * The layer is never interactive (see `pointer-events` in the styles),
+ * because it overlaps whatever is below it, including dialog buttons.
+ */
+export default function ToastLayer({ position, toasts, onExpire }: Props) {
+  const layerRef = useRef<HTMLDivElement>(null)
+  const shownToasts = useRef(new Map<Toast['id'], Toast['shownAt']>())
+
+  useEffect(() => {
+    const layer = layerRef.current
+    if (layer == null) {
+      return
+    }
+
+    // A changed `shownAt` means an already displayed toast was shown anew,
+    // which should bring it back to the front just like a brand new one.
+    const hasNewToast = toasts.some(
+      toast => shownToasts.current.get(toast.id) !== toast.shownAt
+    )
+    shownToasts.current = new Map(
+      toasts.map(toast => [toast.id, toast.shownAt])
+    )
+
+    const isOpen = layer.matches(':popover-open')
+    if (isOpen && !hasNewToast) {
+      return
+    }
+    // Elements of the top layer are painted in the order in which they
+    // entered it, and dialogs enter it through `showModal()`.
+    // Re-entering it for every new toast is therefore what keeps toasts
+    // above dialogs that were opened after this component was mounted.
+    if (isOpen) {
+      layer.hidePopover()
+    }
+    layer.showPopover()
+  }, [toasts])
+
+  return (
+    <div
+      ref={layerRef}
+      // Not `"auto"`, because those get dismissed on `Escape`
+      // and by clicking anywhere.
+      popover='manual'
+      className={classNames(styles.layer, styles[position])}
+      // The layer is kept open (and thus in the accessibility tree)
+      // even while empty, so that added toasts are announced.
+      role='status'
+      aria-atomic='false'
+    >
+      {toasts.map(toast => (
+        <ToastItem key={toast.id} toast={toast} onExpire={onExpire} />
+      ))}
+    </div>
+  )
+}
+
+function ToastItem({
+  toast,
+  onExpire,
+}: { toast: Toast } & Pick<Props, 'onExpire'>) {
+  useEffect(() => {
+    const timeout = setTimeout(() => onExpire(toast.id), toast.durationMs)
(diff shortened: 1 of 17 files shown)


packages/frontend/src/hooks/useToast.ts
```diff
@@ -0,0 +1,7 @@
+import { useContext } from 'react'
+
+import { ToastContext } from '../contexts/ToastContext'
+
+export default function useToast() {
+  return useContext(ToastContext).showToast
+}
```
WofWca, line 7: Nitpick: I think we should avoid introducing any more of such one-line abstractions. It obscures the under the hood code. If we ever actually need to introduce more logic apart from `useContext(ToastContext)` then it's trivial to replace all the occurrences.

packages/frontend/src/components/Toast/ToastLayer.tsx
```diff
@@ -0,0 +1,92 @@
...
+}
+
+function ToastItem({
+  toast,
+  onExpire,
+}: { toast: Toast } & Pick<Props, 'onExpire'>) {
+  useEffect(() => {
+    const timeout = setTimeout(() => onExpire(toast.id), toast.durationMs)
+    return () => clearTimeout(timeout)
+  }, [toast.id, toast.shownAt, toast.durationMs, onExpire])
```
WofWca, line 106: This will restart the timer every time `onExpire` function changes, which might happen on every re-render. Consider utilizing `useEffectEvent`.

packages/frontend/src/components/Toast/ToastLayer.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  return (
+    <div
+      ref={layerRef}
+      // Not `"auto"`, because those get dismissed on `Escape`
+      // and by clicking anywhere.
+      popover='manual'
+      className={classNames(styles.layer, styles[position])}
+      // The layer is kept open (and thus in the accessibility tree)
+      // even while empty, so that added toasts are announced.
```
WofWca:
```suggestion
```

We have many similar places, and there we don't have these comments that explain how live regions work.

packages/frontend/src/components/Toast/ToastLayer.tsx
```diff
@@ -0,0 +1,92 @@
...
+  onExpire: (id: Toast['id']) => void
+}
+
+/**
+ * Renders all currently visible toasts in the
+ * [top layer](https://developer.mozilla.org/en-US/docs/Glossary/Top_layer),
+ * so that they are visible even while modal dialogs are open.
+ *
+ * The layer is never interactive (see `pointer-events` in the styles),
+ * because it overlaps whatever is below it, including dialog buttons.
```
WofWca: So this means that we can't add a "close" button or any other button to any of the popovers? Also that one can't copy their text?
Is it not possible to do things properly, by having the popovers only obscure what's behind them and not the entire screen? And the fact that the entire layer is a `popover` instead of each individual toast is odd.

WofWca (commented):
Some preliminary review

packages/frontend/src/components/Toast/ToastLayer.tsx
nicodh: Sure? onExpire is assigned removeToast which is memoized in a callback with no dependencies

packages/frontend/src/components/Toast/ToastLayer.tsx
nicodh: If interactivity is needed we should use a dialog.

I changed the entire screen layer so no every toast is an own popover now

packages/frontend/src/components/Toast/ToastLayer.tsx
WofWca: Sorry, I meant that it *could* change. If you forget `useCallback` or ever need it to have dependencies, you'll get a bug.

packages/frontend/src/components/Toast/ToastLayer.tsx
WofWca: > I changed the entire screen layer so no every toast is an own popover now

Looks better! So now `pointer-events: none` can be removed.
(Except that toasts are still not interactive when a dialog is open...)

> If interactivity is needed we should use a dialog.

For toasts it's not always necessarily needed, but a "close" button is nice.

packages/frontend/src/contexts/ToastContext.tsx
```diff
@@ -0,0 +1,109 @@
...
+        type,
+        durationMs,
+        position,
+        shownAt: Date.now(),
+      }
+      setToasts(toasts => {
+        // Repeating the action that caused a toast shouldn't show another
+        // toast with the same message. The one that is still displayed
+        // takes over the new toast's properties instead, which also
+        // restarts its display duration.
```
WofWca, line 74: When is this needed? Sounds weird to me. Makes it look like doing the action a second time didn't actually do anything. E.g. if you copied one proxy and then copied another. Inconsistent.

packages/frontend/src/components/Toast/ToastLayer.tsx
```diff
@@ -0,0 +1,120 @@
...
+    if (hasNewToast) {
+      // The top layer is painted in the order it was entered, and dialogs
+      // enter it through `showModal()`, so re-entering keeps toasts above
+      // dialogs that were opened after them.
+      for (const element of elements) {
+        if (element.matches(':popover-open')) {
+          element.hidePopover()
+        }
+        element.showPopover()
+      }
```
WofWca, line 60: Hmmm, sounds like an alternative would be to put the toast inside the dialog element? We might need to see the common practice.

WofWca (commented):
Alright, this is probably better than our current "toast" implementation.

But TBH I was expecting things to go smoother with the new Popover API, yet there is a bunch of hacks here. It feels like we'll eventually need to get rid of this implementation as well, especially since you can't manually close such a toast, so it will keep obscuring stuff behind it.

packages/frontend/src/contexts/ToastContext.tsx
nicodh: If you edit a message, remove all text and hit enter multiple times it would create a bunch of toasts with the same message which to me seems useless. If you copy multiple proxies (why would you do that?) you won't see multiple toasts, if the second copy is in the 4 second range. If the first toast disappered you will see a new one on copy.

I will go through the toasts with Björn anyway after his holidays

WofWca (approved):

PR #6715, opened 2026-09-03
feat: group notifications by chat

WofWca:
FYI this doesn't seem to do much on Windows 10 though.
Probably also need to specify `groupTitle`.
On Linux this also does nothing for now.

### DIFF
--- a/packages/target-electron/src/notifications.ts
+++ b/packages/target-electron/src/notifications.ts
@@ -57,6 +57,11 @@ function createNotification(data: DcNotification): Notification {
   }
 
   const notificationOptions: Electron.NotificationConstructorOptions = {
+    groupId:
+      data.chatId !== 0 && data.accountId !== 0
+        ? `account:${data.accountId}_chat:${data.chatId}`
+        : undefined,
+    // groupTitle
     title: data.title,
     // https://www.electronjs.org/docs/latest/tutorial/notifications#linux
     // says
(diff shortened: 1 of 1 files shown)


nicodh:
> Probably also need to specify groupTitle.

Is that the reason why you added the //groupTitle comment? Why don't you set in then?

WofWca:
> Is that the reason why you added the //groupTitle comment?

Well the main reason is to make it clear that it's available but not specified.

> Why don't you set in then?

I played around with it, but the thing is that it's only supported on Windows. If we use it then we probably should remove the `title` then so as to not duplicate it on Windows, but then we have to run some logic: if Windows then use `groupTitle`, otherwise use `title`. I decided to keep that for another time, or for when the API is supported by more platforms.

nicodh (approved):
Reviewed by reading

Concerning the groupTitle comment: if you plan to do a followup to this PR it's fine, otherwise there should be a comment to make understand why it's there

WofWca:
> if you plan to do a followup to this PR

Not now at least

packages/target-electron/src/notifications.ts
```diff
@@ -57,6 +57,11 @@ function createNotification(data: DcNotification): Notification {
   }
 
   const notificationOptions: Electron.NotificationConstructorOptions = {
+    groupId:
+      data.chatId !== 0 && data.accountId !== 0
+        ? `account:${data.accountId}_chat:${data.chatId}`
+        : undefined,
+    // groupTitle
```
WofWca:
```suggestion
```

PR #6716, opened 2026-09-03
improvement: notifications: extraneous app name

WofWca:
The app name is already included in the notification
(tested on Windows and Linux KDE),
so let's not say it twice.

Before / after:

Screenshot_20260903_before Screenshot_20260903_after

### DIFF
--- a/packages/shared/shared-types.d.ts
+++ b/packages/shared/shared-types.d.ts
@@ -166,7 +166,7 @@ export interface BuildInfo {
 }
 
 export interface DcNotification {
-  title: string
+  title?: string
   body: string
   /**
    * path to image that should be shown instead of icon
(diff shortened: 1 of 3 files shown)


packages/shared/shared-types.d.ts
```diff
@@ -166,7 +166,7 @@ export interface BuildInfo {
 }
 
 export interface DcNotification {
-  title: string
+  title?: string
```
r10s, line 169: are there other occurances left that use title?

WofWca: Yes, plenty.

https://github.com/deltachat/deltachat-desktop/blob/95f5a1e1d16c9e1c8a3e1d340122fae1761b02ad/packages/frontend/src/system-integration/notifications.ts#L243

https://github.com/deltachat/deltachat-desktop/blob/95f5a1e1d16c9e1c8a3e1d340122fae1761b02ad/packages/frontend/src/system-integration/notifications.ts#L291

https://github.com/deltachat/deltachat-desktop/blob/95f5a1e1d16c9e1c8a3e1d340122fae1761b02ad/packages/frontend/src/system-integration/notifications.ts#L307

nicodh (approved):
Reviewed by reading

PR #6718, opened 2026-09-05
chore: upgrade electron 43

nicodh:


packages/e2e-tests/tests/basic-tests.spec.ts
```diff
@@ -480,7 +480,16 @@ test.describe('custom app picker URL', () => {
 
     await page.getByRole('button', { name: 'Attach' }).click()
     await page.getByRole('menuitem', { name: 'App' }).click()
-    await page.getByRole('button', { name: 'Poll' }).first().click()
+    // Narrow down the list first, and then pick the app by its name only:
+    // the accessible name of a list entry also includes the description,
```
WofWca:
```suggestion
    // the name of a list entry also includes the description,
```

Not only "accessible" but in general.

WofWca (approved):
Just an upgrade. Nothing caught my eye in the list of breaking changes https://www.electronjs.org/blog/electron-43-0#breaking-changes.
FreeBSD should still be able to use v42.

Regarding the test: to be clear, the change that fixed the test is `.filter({ has: page.getByText('Poll', { exact: true }) })`. The rest does not affect whether it passes or not.

PR #6724, opened 2026-09-09
fix: guard for huge dimensions

nicodh:
If an image has (extrem) huge dimensions decoding it in the main process as NativeImage might cause memory issues since decoding allocates `width * height * 4` bytes and there is no memory limit in the main process

Using the renderer to get the natural dimensions is the cheapest way to check these values


WofWca (approved):
I tested this and found no bugs.

PR #6726, opened 2026-09-09
remove: webxdc drag-file-out

nicodh:
It was implemented as inofficial method anyway and apparently only for one app


WofWca (approved):
The MR where it has been introduced https://github.com/deltachat/deltachat-desktop/pull/3325.
The "store" WebXDC app is no longer recommended, and I think that the server is down. To those who don't know what it is, it was talked about here https://delta.chat/en/2023-08-11-xstore.

WofWca:
Also IIRC that was introduced even before we had the "share to chat API"? Anyway yep, I think even if some app wants something similar to this "drag out", we still want different API, or no API at all.

PR #6732, opened 2026-09-12
chore: upgrade vulnerable `adm-zip`

WofWca:
Which BTW is only used for tests.


WofWca:
The release is too fresh, CI will be green tomorrow

nicodh (approved):

r10s:
just out of curiosity: why does desktop need a zip dependency?

WofWca:
It's used for some migration tests here

https://github.com/deltachat/deltachat-desktop/blob/e75c0d712323734fd2ee2e8ad97c78e0bc915375/packages/target-electron/migration-tests/migration_function.test.ts#L30

nicodh:
But maybe that migration data should be cleaned up by the way. It holds data from v0.84 to v1.36 (which is from Nov 2019 to April 2023)

PR #6734, opened 2026-09-12
chore: upgrade pnpm to v12

WofWca:
Note that the lock file now contains 2 documents,
so `writeFlatDependencies.js` had to be adjusted.


WofWca:
The build ran, so I'm pretty sure this means that the adjusted script ran fine.

nicodh:
Is there a reason to upgrade just 3 weeks after it's released?

WofWca:
No.
Feel free to close.

PR #6737, opened 2026-09-13
fix: make the tray able to bring the window back (macOS)

korjavin:
When the app sits in the tray on macOS, the tray menu can end up offering only "Hide" and "Quit", with no way to bring the window back — and even when it offers "Activate", that entry does not always work.

Two separate causes, both addressed here:

- `mainWindowIsVisible()` on macOS is just `window.isVisible()`. A window that is merely occluded by another app's windows, or sits on another Space, still reports visible, so the menu renders "Hide" (a no-op from the user's point of view) and the only remaining entry is "Quit". This now includes `isFocused()`, exactly as the Linux branch already does, so an unreachable window gets "Activate".
- `showDeltaChat()` called `show()` on a minimized window, which does not un-minimize it, and did not bring the app in front of the currently active app on macOS. It now `restore()`s a minimized window, focuses it, and calls `app.focus({ steal: true })` on macOS.

It also refreshes the Linux context menu on show, mirroring what `hideDeltaChat()` already does on hide — otherwise "Minimize" stays disabled after showing the window from the tray.

No new strings.

I do not have a working build environment for this project, so this is reasoning over the code paths, not a tested change — a check on macOS would be welcome.

Separately, and happy to leave it out of this PR: on macOS a left click on the tray icon opens the menu, while on Windows/Linux it shows or hides the window directly. Would you take a change making the macOS left click activate the window (keeping the menu on right click)? From a user's point of view "click the icon, get the app" is what the tray icon is for.


packages/target-electron/src/tray.ts
```diff
@@ -67,7 +67,10 @@ export function showDeltaChat() {
   if (!mainWindow.window) {
     throw new Error('window does not exist, this should never happen')
   }
+  if (mainWindow.window.isMinimized()) mainWindow.window.restore()
   mainWindow.window.show()
+  mainWindow.window.focus()
```
WofWca: We have a lot of places where we do a `focus()` on a window. I think we should unify those.
Maybe something for a different MR.

packages/target-electron/src/tray.ts
WofWca, line 146: Looks like the if-else branches are nearly identical now, the only difference is a label and the `refreshTrayContextMenu` call.

packages/target-electron/src/tray.ts
```diff
@@ -99,28 +102,27 @@ function getTrayMenu() {
...
-          },
-
+      {
+        id: 'open_windows',
+        label: tx('activate'),
+        type: 'normal',
+        click() {
+          showDeltaChat()
+          // fix #3041
+          refreshTrayContextMenu()
```
WofWca: Also now that `showDeltaChat` also includes the "refresh", looks like there is no need to call these `refreshTrayContextMenu` a second time?

WofWca (commented):
Welcome, and thanks!

The change to `showDeltaChat` looks reasonable, at least the "refresh the menu" part.

> A minimized window still reports `isVisible() === true`

Are you sure that this is the case on macOS? You said you didn't test it. I'm guessing that there is a reason that we have two different menus for different OSes, i.e. we seem to be trying to handle that case.

korjavin:
You are right, thanks for catching it — I checked and my claim was wrong. electron/electron#38242 unified `isVisible()` in May 2023, and this project is on Electron 43, so a minimized window reports `false` on macOS and the menu does render "Activate" in that case. I dropped that claim and the menu restructuring.

What is left is narrower and does not depend on minimize semantics:

- macOS `mainWindowIsVisible()` is `isVisible()` only, so a window occluded by another app (or on another Space) counts as visible and the menu offers just "Hide" and "Quit" — no way to activate the app from the tray. It now also checks `isFocused()`, like the Linux branch does.
- `showDeltaChat()` did `show()` on a possibly minimized window and did not bring the app to the front on macOS; it now restores, focuses, and calls `app.focus({ steal: true })` there.
- Plus the context menu refresh on show that you found reasonable.

Still untested by me — I am on macOS but without a build environment for this project, so I am relying on your review for the actual behaviour. I also left a question at the end of the PR description about the macOS left-click behaviour.

packages/target-electron/src/tray.ts
nicodh, line 161: This is always false (at least on my Linux test). It seems opening the tray menu triggers onBlur on the window which sets isFocused to wrong.

When playing around I found that isVisble() together with  !mainWindow.window.isMinimized() is the better condition to enable this menu item

nicodh (changes requested):
When testing on MacOS Tahoe 26.6.2  I could not find any difference between the old tray menu behaviour and the new one. It always shows
Show all windows
Show or Hide
Quit

Show (second row) is only shown if the window was hidden using the Hide entry of the tray menu - not by minimizing the window from the window bar)

I could not reproduce what you describe here, only when minimizing the window using the yellow icon in the window bar. Then "Show all windows" does nothing "No windows available" and second tray item shows "Hide".

Left click on the try icon already opens the window immediately so even that I could not reproduce.

For linux it always shows "Minimized" disabled, which is an old bug but could be fixed (see comment)

In general: the pr title says (macOS) but you say you didn't test - so you can't even reproduce the bug on main, right?

What is you MacOS version?

korjavin:
image
My problem and what I am trying to solve is this menu on mac in tray. (not in dock).

There is no way to bring to focus/view deltachat window from tray icon.

nicodh:
> My problem and what I am trying to solve is this menu on mac in tray. (not in dock).
> There is no way to bring to focus/view deltachat window from tray icon.

Ok, interesting - I didn't even know that there is an additional icon on MacOS on the upper tray.

So now I can confirm that your PR fixes the MacOS problem.

nicodh (approved):
Solves the MacOS issue

For Linux I will do a followup

Thanks for contributing!

Issue #3986, opened 2024-06-28
deltachat-desktop on Linux misuses XDG configuration location for storing cached data, not just configuration


dkg:
I am testing DeltaChat 1.46.1 on Debian stable (12.5).

It looks to me like DeltaChat is storing configuration information, application state, application data, and ephemeral (cached) material all within `~/.config/DeltaChat`.

The [XDG basedir spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) identifies `~/.config` (as `$XDG_CONFIG_HOME`) specifically for "user-specific configuration".  However, i see at least two things stored here that are definitely not local configuration:

- logfiles written to `~/.config/DeltaChat/logs/` (these should probably be logged to the system's journal)
- cached data written to `~/.config/DeltaChat/{Cache,GPUCache,Code Cache,DawnWebGPUCache,DawnGraphiteCache}` (these should probably live in `$XDG_CACHE_HOME/DeltaChat`, or `~/.cache/DeltaChat`, if `$XDG_CACHE_HOME` is unset)

farooqkz:
Thanks even I didn't know that. Yeah it makes perfect sense that cache data should live in their own home not someone else's.

@Simon-Laux Who is in charge of Cache? I mean what part of Desktop?

Simon-Laux:
electron is, not sure yet how to change that, might be electron api that we call on startup (like a chrome cli switch, or standard electron api), might be an env var, or electron-builder config. or if we are unlucky it's an upstream issue in electron.

nicodh:
There is an old issue still unsolved but still active discussed:
https://github.com/electron/electron/issues/8124
and a project related to that problem: https://github.com/danisztls/ephemeral

Simon-Laux:
The tauri edition is better in this aspect.

- Logfile:
   - on macos it is stored in the log directory of the system: `~/Library/Logs/chat.delta.desktop.tauri`
   - but on linux it logs are stored together with userdata in `~/.local/share/chat.delta.desktop.tauri/`, though we can change this more easily than in electron. (and we probably should change it because of deltachat/deltachat-tauri#8, though I'm skeptical about writing them to the systems journal, because it would increase friction for users to send us the logfiles)
- Cache:
   - no cache in the data folder, it looks like it is stored in `~/.cache/deltachat-tauri/` instead

Not much we can do on electron. there is an api to set the file location of the browser session, but there is no api to set cache directory for these sessions separately.

[...]

PR #6742, opened 2026-09-15
fix: move Electron cache and skip protocol registration in portable mode

nicodh:
Reduce traces created by deltachat portable version on Windows

#public-preview


nicodh:
Tested on Win11

There is no AppData/Roaming/DeltaChat folder anymore!

Edit: so maybe this could be a fix for Chrome cache location #3986 also?

WofWca (approved):

John-foobar-Doe:
When & where can I download a new release or build containing this fix?

nicodh:
> When & where can I download a new release or build containing this fix?

We will have a new release soon - probably during next week

PR #6743, opened 2026-09-15
feat: add UI settings access points for native menu items (Language, Keybindings, DevTools) + redesign About dialog

Bushikot:
# Summary

Adds in-app graphical UI access points (in Settings and the About dialog) for items that previously lived only in the native window menu - Language, Developer Tools, Keybindings, and log-folder/file access. This makes them reachable when the menu bar is hidden, and groups them with related settings. The native menu is not modified; it remains an untouched fallback. Electron target only - Tauri is not touched.

What's new

### Settings - main list 

- Language entry (icon translate) → opens a new Language sub-dialog (radio list of available languages, sourced from the same data the native menu uses). Switching language applies immediately via the existing setLocale flow.
- Keybindings entry (icon keyboard) → opens the existing KeybindingCheatSheet dialog.

### Settings - Advanced

- Developer Tools action button (hidden on browser target - Electron-only). 
- New Logging group: View Log (existing), Open Log Folder (new, Electron-only), Open Current Log File (new, Electron-only). 

### About dialog - redesigned

- App logo + version (with git ref in devmode) + edition label for non-Electron targets. 
- Short description and license paragraph (GPL-3.0 + GitHub links via ClickableLink).
- Footer: Contribute (hand-heart) and Report an Issue (bug-outline) buttons. Removed the old Help/View Log buttons (Help stays in Settings list; View Log stays in Advanced). 

### Runtime / IPC (Electron)

Three new optional methods on the Runtime interface, implemented over new IPC channels - frontend calls them with optional chaining so Tauri needs no changes:
- getAvailableLanguages() - reuses the already-parsed languages array from menu.ts (no duplication). 
- toggleDevTools() - toggles devtools on the main window.
- openLogFolder() - opens the log folder in the OS file manager.

### Welcome screen 

Language selector on the Add-Profile / onboarding screen (reuses the Language component). 

### Misc 

Hides the native window menu bar by default (hideMenuBar: true) - the native menu is still reachable via the setting / Alt key.

Translation strings

New keys in _locales/_untranslated_en.json: pref_logging, about_description, about_license_text. Existing keys reused where listed (pref_language, keybindings, global_menu_view_developer_tools_desktop, log-folder/file labels, etc.).

Out of scope / intentional

- Native menu (menu.ts) - completely unchanged.
- Tauri - not touched.

Screenshots
- [Add profile](https://github.com/user-attachments/assets/cdfa6428-984b-4fbc-bbeb-2a7bda318b18)
- [About Delta Chat](https://github.com/user-attachments/assets/9bfa104f-0c62-4cbd-94a1-fe00926803f1)
- [Settings](https://github.com/user-attachments/assets/564d0022-9c07-4b91-b53a-eff6f83c58fd)
- [Settings -> Advanced](https://github.com/user-attachments/assets/27a7ccda-8c2a-4966-8267-38696a2dc7d9)


r10s (changes requested):
thanks for thinking about that, i cannot say anything for the code, but to to UI as such, we do not want to have quite some of the suggested changes:

- in settings, "Language" should go to "Appearance"; it is not important enough for a top-level option, it is already in the menu, also the selected language is a language the user speaks

- in the onboarding screen, we do not want to clutter UI with the "Language" option for the same reasons

- in settings, "Keybindings" is not important enough as well for a top-level option, it is in the "Help" menu. it is anyway for advanced user and they will find it there. we can consider to link it from the help window tho

- in advanced settings,  we do not want to have  "Developer Tools action button", we just removed that. it is super-advanced, and there, the menu is the correct place

- in advanced settings, we do not want to have a "View log" group. we can consider to have "open log folder/file" in a three dot menu in the log window tho. 
for the end user, a "share log to other app" might be interesting, but not sure how well this is supported on electron

- we just revamped and discussed the "About" screen. that should not be changed again. "Repot and issue" is in the menu, same "Contribute"

all in all, focus is end user. focus is not about debugging, logging, rare features. and we want to have fewer options, not more :)

nicodh:
Thanks for your contribution.

Please keep in mind that changes to the UI are generally implemented only after extensive consultation and with due consideration of the priorities mentioned above and the other Deltachat clients.

It is also preferable not to include too many changes in a PR, but rather to focus on individual changes that can be better discussed in the comments.

For example it would be a wanted change, to have an additional language switch in Settings -> Appearance

Bushikot:
Hi everyone! Thanks for taking the time to look at this and for the feedback. 

To be honest, I was a bit caught off guard by how quickly the PR was closed. I had hoped the PR itself could serve as a working draft and a discussion space where we could iterate on the changes, adjust placements, or drop the parts you felt strongly against. 

It’s a bit disappointing, as I invested two full evenings into this — not just writing the code, but researching references, UI best practices, and trying to find the minimal set of changes required to make hiding the native menu by default viable without stranding features. But I understand that my expectations regarding the process and scope didn't align with your workflow. 

While I see your points regarding UI priorities, I have no intention of stretching this into a drawn-out cycle of single-button PRs and endless back-and-forth reviews for every minor tweak. The overhead just isn't worth it.

Thanks again for your time and for Delta Chat!

PR #6747, opened 2026-09-16
fix: remove default setting for relays

nicodh:
As a preparation for the upcoming release where we have no default relay anymore


WofWca (approved):

link2xt:
"Messages are received on all relays." can likely be removed as well now:
image

It looks like a clarification that was needed because one of the relays was marked as relay for sending before.

WofWca:
It's from

https://github.com/deltachat/deltachat-android/blob/42fa72b92f127958cbe1d5bc50644e156f6ecb12/src/main/res/values/strings.xml#L739

I'll create an issue on Android first.

r10s:
@WofWca please keep showing the string `transport_list_hint`; it will be updated at https://github.com/deltachat/deltachat-android/pull/4680

Issue #6739, opened 2026-09-14
In the .deb version the opened app does not have the icon (and other hacks now required on distros)


account1009:
I've tested the .deb version with a Phosh DE and I have to manually change from StartupWMClass=DeltaChat to StartupWMClass=deltachat to have everything working.

It looks that some DE are case sensitive so a precise string is required.

I've searched upstream on Electron and they are providing a solution to avoid these issue:
https://www.electron.build/docs/linux/#window-association-desktopname

Electron doc is suggesting to specify "desktopName" in the `package.json`
So my suggested solution is to add `"desktopName": "deltachat-desktop",` to the `package.json` file

Apart Debian I'm packaging deltachat-desktop for Alpine Linux and I'll propose it to the distro when I'm happy about the result.
With the point of view of distro maintainer another downstream solution is to use "pkg set desktopName" as is doing Arch Linux to solve. But I think that it is a sort of hack because "desktopName" is not defined on package.json

Another thing that I've noted from the distro point of view trying to do the package for Alpine is that many distros use a separated electrum package (that is fine for distros) and put inside the package only the .asar file.

To generate this .asar now the distro has to call 3 separated commands:
```
pnpm --filter=@deltachat-desktop/target-electron pack:generate_config
pnpm --filter=@deltachat-desktop/target-electron build4production
pnpm --filter=@deltachat-desktop/target-electron pack:linux:dir
```

maybe it can be nice for distros to provide also a single command such as "buildasar" to have all the steps together

The last think that I've noticed that can be nice for distros is that the distro has a package for server separated from the deltachat-desktop

For this reason I'm using something like 
```
export DELTA_CHAT_RPC_SERVER=/usr/bin/deltachat-rpc-server
exec electron /usr/lib/deltachat-desktop/app.asar "$@" --allow-unsafe-core-replacement
```
to use the system version of the deltachat-rpc-server. In that case "allow-unsafe-core-replacement" has a name that seems unsafe but I'm just telling to use the system `deltachat-rpc-server` installed with another package. 
Maybe it could be nice to use a neutral name that does not have the "unsafe" word

For now I'm focusing on Alpine Linux but my goal it to make deltachat-desktop easy to add and to maintain on any linux distro
For these reasons I'm suggesting upstream edits on deltachat side to simplify :)

@nicodh For now I've not suggested merge requests because the edits are trivial but it is more to discuss if it is fine to solve some stuff upstream on deltachat-desktop instead of having hacks on distro level

[...]

PR #3567, opened 2023-11-25
use stdio binary instead of dc node & update electron to 30 & update min node to 20

Simon-Laux:
Todo:
- [ ] account migration function in core
- [X] port into own npm module that has all the prebuilds somehow
- [x] remove open multiple instances parameter
- [x] issue when closing deltachat 
- [ ] code review from "experts" -> fix linereader/performance
- [ ] optimise with regex or use `split2` package (https://github.com/mcollina/split2/blob/master/index.js) -> to make it more correct with `/\r?\n/`

This experiment came from a hacking session together with @Jikstra.

depends on:
- https://github.com/deltachat/deltachat-core-rust/pull/5332

PR #6754, opened 2026-09-17
change: rename allow-unsafe-core parameter

nicodh:
- rename allow-unsafe-core-replacement => allow-custom-rpc-server-path
- (!) this parameter now includes using deltachat-rpc-server from $PATH

related to #6739 

when looking into the code it turned out, that $PATH was never used (takeVersionFromPATH was not passed as argument)
Now this changes and does what the docs claim

tbd: there would be other possible names for that parameter, depending what has priority: the fact that not the shipped binary is used, the fact that the installed binary will be used, the fact that there is a possible risk. So _allow-custom-rpc-server-path_ is my suggestion


WofWca (approved):
Relevant MRs:
- https://github.com/chatmail/core/pull/5332
- https://github.com/deltachat/deltachat-desktop/pull/3567
  (https://github.com/deltachat/deltachat-desktop/pull/3567/commits/43003647d148e41ecf6cc55182f2e783247f6280)
- https://github.com/chatmail/core/pull/5332

PR #6755, opened 2026-09-17
remove: migration for very old versions

nicodh:
The migrations were only needed for account data of versions that were not updated since 5 years.
(The API has changed in d8cad99fa ) The newest data the tests were running on were 3 year old. So let's remove the migration code and the tests.


_locales/_untranslated_en.json
```diff
@@ -1,4 +1,7 @@
 {
+  "data_found_legacy_format_message": {
+    "message": "Your profiles were created by a very old version of Delta Chat (before 1.22) and cannot be opened by this version.\n\nTo keep them:\n1. Quit now\n2. Install Delta Chat Desktop 2.59.x and start it once, it converts your profiles to the current format\n3. Relaunch this version\n\nYour old data is stored at:\n%1$s\n\nContinue without your previous profiles?"
+  },
```
r10s: ftr, this string would probably never gets translated as it is too rarely used, more an error message, and also hard to translate, and/or will open questions.

so we could also hardcode them to the place it is actually used - similar to other error messages

WofWca (approved):

PR #6757, opened 2026-09-17
WIP: feat: add device message when an update available

### LINKED ISSUE #6749: feat: inform users about updates
Since https://github.com/chatmail/core/pull/8557 we have get_app_version in JSON-RPC which allows us to detect a new version without calling any url or service. We need to provide some AppVersion to https://github.com/chatmail/relay/blob/main/chatmaild/src/chatmaild/defaults/appversions.json which will be used to inform all clients.

Then the client should inform the user about it (preferable with a device messages)

We probably need a matrix how to process on the different platform. But let's start with the device message and a link.

in AppVersion Sources
{
     "sourceId": "apple-store",
     "versionInteger": 123,
     "versionString": "2.60.1"
},
{
     "sourceId": "microsoft-store",
     "versionInteger": 123,
     "versionString": "2.60.1",
},
{
     "sourceId": "flatpak",
     "versionInteger": 123,
     "versionString": "2.60.1",
}
{
     "sourceId": "appimage",
     "versionInteger": 123,
     "versionString": "2.60.1",
}

The download URLs should not be fetched from the relay (untrusted source) but should be hardcoded on the client. (see links in Status release issue) We could also provide a "linux" sourceId just pointing to get.delta.chat

WofWca:
Closes https://github.com/deltachat/deltachat-desktop/issues/6749.


packages/target-electron/src/index.ts
```diff
@@ -575,3 +575,98 @@ ipcMain.handle('restart_app', async _ev => {
...
+      //
+      // Ah or we could simply pass `null` as message for now.
+      `update_${lastReleaseInfo.versionString}`,
+      {
+        // TODO so I guess we need some new API
+        // that ensures that this is not linkified?
+        //
+        // TODO translated `update_available_msg`
+        // text: tx('update_available_msg', ),
+        text: `New update ${lastReleaseInfo.versionString}!!\nhttps://get.delta.chat/`,
```
WofWca, line 656: Until we have the "don't linkify" API, I think we can simply skip the version string, or to apply even more filtering to it, i.e. only allow dots and digits.

